PR #7343, opened 2021-09-10
feat: Implement Appsmith helm chart

geekup-legodevops:
## Description

- Implement Appsmith Helm chart (with the use of fat container image)
- Helm chart includes application deployment with basic default setup
- Support for custom configuration for persistent volume data, custom ingress controller and secure with TLS certificate

## Type of change

## Checklist:


deploy/helm/Chart.yaml
```diff
@@ -0,0 +1,18 @@
+annotations:
+  category: Application
+apiVersion: v2
+appVersion: "1.16.0"
+name: appsmith
+type: application
+dependencies:
+  - name: common
+    version: 0.3.1
+    repository: https://charts.bitnami.com/bitnami
```
mohanarpit, line 10: Is this a typo in the URL?

deploy/helm/templates/NOTES.txt
```diff
@@ -0,0 +1,22 @@
+1. Get the application URL by running these commands:
```
mohanarpit, line 1: Is the file `deploy/helm/charts/common-0.3.1.tgz` required or is it a build output?

mohanarpit (approved):

fernandocarletti:
Hello! Is there any reasoning going for a fat container image? It sounds off since it goes against the idea of isolating the processes in a different container. Is this chart supposed to be used in production?

sharat87:
Hey @fernandocarletti, thank you for writing to us. Yes, this is the production Docker image of Appsmith, and this is what we recommend for production deployments.

I'd agree on the point of isolation being one of the primary reasons for containerization. But that doesn't necessarily mean isolated _processes_. If we think in terms of isolated _applications_, this makes more sense. For instance, Appsmith needs a Java backend server, an NGINX server, a MongoDB etc. to be fully functional. Now, for a user who wants to deploy Appsmith and start building things with it, this is non-critical information. Appsmith could be running on magical beans for all they care. So in such cases, having Appsmith as a single container makes deployment much easier and simpler.

Additionally, upgrades become easy because of isolating at the application level, instead of process level. Consider, hypothetically, if we want to add a new microservice for Appsmith, or, say, want to replace MongoDB with PostgreSQL (cause, why not!). This would mean that, in the fat container, we swap MongoDB with PostgreSQL, include a migration path as part of the entrypoint, and none of our users should have to care or know. Contrast with a multi-container setup, we'll have to reach out to all our users, and everybody will have to change their `docker-compose.yml` files, carefully since YAMLs are so sensitive, and then perhaps migrate data, and then bring everything up. I'd assume a majority of our users would just prefer to not upgrade at all if this were the case.

Now, don't get me wrong, this is not a silver bullet. So, if your Appsmith fat container is configured to connect and use an external MongoDB database, the one inside the fat container _will not run_. This is true of using external Redis as well. Appsmith fat container is fat, but with an option to trim it down, for users that do want them as separate containers.

fernandocarletti:
Hello @sharat87. Thanks for the explanation. We need high availability, it seems we need to create a custom deployment of our scenario.

Although I understand the point, it still sounds off to me to use "fat containers by default", especially in production (and mostly in Kubernetes, as it removes the ability to set container requests/limits). Using Helm, both the database and the cache servers (maybe even the ingress, but I would not go that far) could be [Helm dependencies](https://helm.sh/docs/helm/helm_dependency/), which could still grant you guys the goal of having it all together. Of course, the downside would be maintaining a different "drop-in" setup for development, which sounds like a case for a versioned docker-compose file.

PR #7385, opened 2021-09-13
ci: Add CI workflow for building fat container

sharat87:
This PR adds a new GitHub Actions workflow to do the client and server tests, and if they pass, then build the fat container image and push to DockerHub.


.github/workflows/test-and-build-fat.yml
```diff
@@ -0,0 +1,445 @@
...
+          APPSMITH_MONGODB_URI: "mongodb://localhost:27017/appsmith"
+          APPSMITH_REDIS_URL: "redis://127.0.0.1:6379"
+          APPSMITH_ENCRYPTION_PASSWORD: "password"
+          APPSMITH_ENCRYPTION_SALT: "salt"
+          APPSMITH_IS_SELF_HOSTED: "false"
+          APPSMITH_CLOUD_SERVICES_BASE_URL: "https://release-cs.appsmith.com"
+          APPSMITH_CLOUD_SERVICES_USERNAME: ""
+          APPSMITH_CLOUD_SERVICES_PASSWORD: ""
+        run: |
+          ./build.sh -DskipTests
```
mohanarpit, line 174: The name of the step claims that we are executing the tests here while we are infact skipping the tests. Can we correct either the name or the build step?

.github/workflows/test-and-build-fat.yml
```diff
@@ -0,0 +1,445 @@
...
+            echo "Integration tests completed successfully!";
+          elif [[ "${{ needs.ui-test.result }}" == "skipped" ]]; then
+            echo "Integration tests were skipped";
+            exit 1;
+          else
+            echo "Integration tests have failed";
+            exit 1;
+          fi
+
+  package:
```
mohanarpit, line 393: From my understanding, only the `package` step changes between the CI workflow that we have today & this one. Do we need a completely separate workflow for building fat containers? Or should we just add another job to our current workflow to execute `fat-package` images as well.

.github/workflows/test-and-build-fat.yml
```diff
@@ -0,0 +1,445 @@
...
+        uses: actions/checkout@v2
+        with:
+          ref: refs/pull/${{ github.event.pull_request.number }}/merge
+
+      - name: Checkout the head commit of the branch
+        if: github.event_name == 'push' || github.event_name == 'workflow_dispatch'
+        uses: actions/checkout@v2
+
+      - name: Download the react build artifact
+        uses: actions/download-artifact@v2
```
mohanarpit, line 417: Do we not have to download the server jar as well? Am I missing something here?

.github/workflows/test-and-build-fat.yml
sharat87: Good catch. We do want the tests to run here.

.github/workflows/test-and-build-fat.yml
sharat87: I opted for a new one, since we have two separate workflows. Which one do we add to? We have the option of picking one, say client, and add fat container in that workflow. But I want the server tests to also run and pass before the fat container is run, so this workflow now needs to run server tests as well. At that point, we can do away with the server workflow, and just use the client workflow for client tests, server tests, Cypress tests, and then build the old-server-container, old-client-container, and the fat-container.

Both of those were viable options. I chose the path of a new workflow because this is less tampering with existing workflows, so, it felt like a more prudent approach is all. Does that make sense?

.github/workflows/test-and-build-fat.yml
sharat87: Yeah we do. Copy-pasta mistake. Thanks.

.github/workflows/test-and-build-fat.yml
mohanarpit: The problem with a new workflow is that the Cypress tests will run *again*. I'm trying to avoid that. If we move the `package` step from `Server Build Workflow` to the `Appsmith Test Workflow` then we can use the same workflow for everything. Ensures that we run specific steps on specific events but the workflow is the same & the steps aren't repeated. 

I'm trying to optimize on build time right now. Our current builds take too long.

sharat87: Understood. In that case, just so we're on the same page, here's what we'll do:

1. Remove this new workflow.
2. Change the client-test workflow, to run server tests as well, before Cypress tests.
3. Build the Docker images for client, server, and fat-container, after Cypress tests pass.
4. Remove (or disable) the server workflow.

Going forward it'll just be one workflow, for both server and client. Does that sound right?

mohanarpit: Minor modification to #4. We don't remove (or disable) the server workflow right now. We keep it for a few days to ensure that we don't disrupt pipelines. Once we are confident that this new flow is working, we disable the old server workflow.

sharat87: Oh yeah, that's not a _now_ thing sorry, that will happen a few weeks later. Should've phrased that better, but yes, agreed.

sharat87: @mohanarpit, done. Can you take a look please. I'm not getting the "re-request for review" button on GitHub, so posting here. 🤷

.github/workflows/server.yml
```diff
@@ -5,16 +5,16 @@ on:
   # This line enables manual triggering of this workflow.
   workflow_dispatch:
 
-  push:
```
mohanarpit, line 8: Sorry, I didn't understand this change. Why are we commenting out the server build workflow? This is run on each commit to the branch. Do we not want to do this any more? 

I'm asking because from my understanding, the `client-test.yml` workflow is only run when a PR is approved. When it is still under development, no tests will be executed if these build specific workflows are commented.

.github/workflows/test-build-docker-image.yml
```diff
@@ -474,27 +478,58 @@ jobs:
...
+      - name: Push server master image to Docker Hub with commit tag
+        if: success() && github.ref == 'refs/heads/master'
+        run: |
+          docker build --build-arg APPSMITH_SEGMENT_CE_KEY=${{ secrets.APPSMITH_SEGMENT_CE_KEY }} -t ${{ secrets.DOCKER_HUB_ORGANIZATION }}/appsmith-server:${GITHUB_SHA} .
+          docker build --build-arg APPSMITH_SEGMENT_CE_KEY=${{ secrets.APPSMITH_SEGMENT_CE_KEY }} -t ${{ secrets.DOCKER_HUB_ORGANIZATION }}/appsmith-server:nightly .
+          echo ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }} | docker login -u ${{ secrets.DOCKER_HUB_USERNAME }} --password-stdin
+          docker push ${{ secrets.DOCKER_HUB_ORGANIZATION }}/appsmith-server:${GITHUB_SHA}
+          docker push ${{ secrets.DOCKER_HUB_ORGANIZATION }}/appsmith-server:nightly
+
+      - name: Build and push release image to Docker Hub
```
mohanarpit, line 516: Do we not have to build RTS as well?

.github/workflows/server.yml
sharat87: This will be duplicating effort right? The combined workflow now builds all three images, server, client and fat. That's why I commented it out.

I'm not sure how we want to do this distinction that server tests need to run even for non-approved PRs. Should we add another condition there in that combined workflow? That client tests run only on approved PRs, but server tests run always. But Docker images are built only on approved PRs? Does that make sense?

.github/workflows/test-build-docker-image.yml
sharat87: Indeed. I'll pull in RTS workflow into this file as well. This, again, should run for non-approved PRs as well, right?

.github/workflows/server.yml
sharat87: Okay, so the combined workflow runs always, but the client tests won't run when the PR is not approved. Only the server tests will run. But since the client tests won't run when PR is not approved, the following jobs won't be triggered, which are the Cypress tests, and building of images. This is what we need right? Once the PR is approved, client tests also will run, and assuming they pass, Cypress tests will run, and assuming they pass, all four Docker images are built and pushed.

.github/workflows/test-build-docker-image.yml
sharat87: Done. Pulled RTS into the combined workflow.

sharat87:
@mohanarpit, just pushed changes as discussed. There's quite a few insignificant whitespace changes, so you may want to ignore them in the Diff settings.

.github/workflows/test-build-docker-image.yml
```diff
@@ -27,25 +29,18 @@ defaults:
 
 jobs:
   buildClient:
-    # If the build has been triggered manually via workflow_dispatch or via a push to protected branches
```
mohanarpit, line 30: Can we add this condition back? Else, workflow_dispatch event may not work.

.github/workflows/test-build-docker-image.yml
```diff
@@ -193,15 +188,14 @@ jobs:
...
-      - name: Build package
+
+      - name: Test and Build package
         env:
           APPSMITH_MONGODB_URI: "mongodb://localhost:27017/mobtools"
           APPSMITH_REDIS_URL: "redis://127.0.0.1:6379"
           APPSMITH_ENCRYPTION_PASSWORD: "password"
           APPSMITH_ENCRYPTION_SALT: "salt"
           APPSMITH_IS_SELF_HOSTED: false
-        working-directory: app/server
```
mohanarpit, line 204: I think the working-directory is required here. My hunch is that this step will not be able to find the build script without the working-directory parameter.

.github/workflows/test-build-docker-image.yml
```diff
@@ -212,10 +206,65 @@ jobs:
...
+          echo ::set-output name=tag::$(echo ${GITHUB_REF:11})
+
+      - name: Build
+        run: ./build.sh
+
+      # Upload the build artifact so that it can be used by the test & deploy job in the workflow
+      - name: Upload server build bundle
+        uses: actions/upload-artifact@v2
+        with:
+          name: build
```
mohanarpit, line 254: The output folder for buildServer, buildClient & buildRts steps is the same. Is this intentional? Can you please verify if the final Docker image has all the required services?

sharat87: Honestly, I'm not sure. This is the current workflow, where both client and server artifacts use the same artifact name. So, just to not have too many changes, I kept the status quo.

mohanarpit: Ok. Once it's merged into release, we'll find out. 😃

.github/workflows/test-build-docker-image.yml
sharat87: I _think_ it'll still run, though I don't have any proof to show for it. Brought the condition back. :)

.github/workflows/test-build-docker-image.yml
sharat87: I think it will. The current job has a default working directory set for `run` steps to this same folder. But, again, let's see, one step at a time. Reverted and pushed.

mohanarpit (approved):

PR #7387, opened 2021-09-13
fix: disable sorting in view mode

### LINKED ISSUE #5640: Disable sorting for end-user view mode(required when sorting will mess up the row order)

Tooluloope:
## Description

disable sorting in view mode

Fixes #5640

## Type of change

## Checklist:

    // Code coverage diff between base branch:release and head branch: fix/disable-sorting-in-view-mode 
 :green_circle: | total | 54.9 **(0.01)** | 36.6 **(0.02)** | 33.62 **(0.01)** | 55.44 **(0.01)**
 :green_circle: | app/client/src/utils/DSLMigrations.ts | 73.62 **(0.11)** | 65.14 **(-0.06)** | 70.21 **(0)** | 73.43 **(0.12)**
 :green_circle: | app/client/src/utils/migrations/TableWidget.ts | 84.43 **(0.69)** | 76.32 **(1.04)** | 92.59 **(0.59)** | 86.49 **(0.61)**
 :red_circle: | app/client/src/widgets/TableWidget/component/TableUtilities.tsx | 37.95 **(-0.23)** | 21.62 **(-1.24)** | 16.22 **(0)** | 36.18 **(-0.24)**
 :red_circle: | app/client/src/widgets/TableWidget/component/index.tsx | 27.27 **(0)** | 2.6 **(-0.03)** | 15.79 **(0)** | 25.68 **(0)**
 :green_circle: | app/client/src/widgets/TableWidget/widget/index.tsx | 26.49 **(0)** | 6.09 **(0.81)** | 28 **(0)** | 26.36 **(0)**</details>


app/client/src/widgets/TableWidget/component/TableUtilities.tsx
```diff
@@ -526,7 +527,7 @@ export function TableHeaderCell(props: {
     <div
       {...column.getHeaderProps()}
       className="th header-reorder"
-      onClick={handleSortColumn}
+      onClick={props.editMode !== false ? handleSortColumn : undefined}
```
SatishGandham, line 530: Can we not write this as       `onClick={props.editMode ? handleSortColumn : undefined}`

Also, won't this disable sorting in view mode completely?

SatishGandham (changes requested):

app/client/src/widgets/TableWidget/component/TableUtilities.tsx
Tooluloope: @SatishGandham , I think the issue is to disable sorting in view mode.

SatishGandham: Issue says when it can break the table, so there should be an option to disable sorting.
Do we store sorting state in the edit mode when we sort columns?

Tooluloope: Oh, So just a switch in the property pane.
Shouldn't we just tackle why the table breaks, cos users might not know the real reason for disabling sorting

somangshu:
@Tooluloope the sorting disabled needs to be a configurable option, The dev can choose to do so for the end user. Hence, we need to provide the option to do so from the property pane

app/client/src/widgets/TableWidget/component/TableUtilities.tsx
```diff
@@ -517,12 +519,15 @@ export function TableHeaderCell(props: {
       props.isAscOrder === undefined ? false : !props.isAscOrder;
     props.sortTableColumn(columnIndex, sortOrder);
   };
+  const disableSort = useMemo(() => {
```
SatishGandham, line 522: We don't need a memo here. It is only required it is going to return a new object with same value or if we are doing expensive calculations. 
Here it is a primitive value.

SatishGandham: I don't see any changes?

SatishGandham (changes requested):

app/client/src/widgets/TableWidget/component/TableUtilities.tsx
Tooluloope: @SatishGandham I have removed it.
Not sure why github still showing this on your end
Screenshot 2021-10-01 at 07 49 50

SatishGandham (dismissed):

Tooluloope:
@YogeshJayaseelan Fixed

YogeshJayaseelan:
@Tooluloope Verified the fix and the sorting toggle is working as expected

SatishGandham (dismissed):

app/client/src/utils/DSLMigrations.ts
```diff
@@ -975,6 +976,10 @@ export const transformDSL = (
     currentDSL = mapAllowHorizontalScrollMigration(currentDSL);
     currentDSL.version = LATEST_PAGE_VERSION;
```
riodeuno, line 977: This needs to change to 44 from `LATEST_PAGE_VERSION`

riodeuno (changes requested):

riodeuno:
@Tooluloope I have one change requested. I believe @SatishGandham has already reviewed, this so I'll remove myself from the reviewers list.

app/client/src/utils/DSLMigrations.ts
```diff
@@ -975,6 +976,10 @@ export const transformDSL = (
...
   }
+  if (currentDSL.version === 44) {
```
Tooluloope, line 979: @riodeuno we're using 44 here already

SatishGandham (approved):

PR #7397, opened 2021-09-14
feat: Smart Substitution for Google Sheets Plugin

### LINKED ISSUE #6638: [Feature] Smart Substitution for Google Sheets
## Summary

The smart substitution feature doesn't work for the google sheet plugin. A user is trying to bind a multi-line input to an insert query and it fails because of \n substitution

trishaanand:
Fixes #6638

### DIFF
--- a/app/server/appsmith-plugins/googleSheetsPlugin/src/main/resources/editor.json
+++ b/app/server/appsmith-plugins/googleSheetsPlugin/src/main/resources/editor.json
@@ -381,12 +381,48 @@
           "label": "Row Object",
           "configProperty": "actionConfiguration.pluginSpecifiedTemplates[9].value",
           "controlType": "QUERY_DYNAMIC_TEXT",
+          "evaluationSubstitutionType": "PARAMETER",
           "hidden": {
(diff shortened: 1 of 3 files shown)


app/server/appsmith-plugins/googleSheetsPlugin/src/main/resources/editor.json
```diff
@@ -381,12 +381,48 @@
           "label": "Row Object",
           "configProperty": "actionConfiguration.pluginSpecifiedTemplates[9].value",
           "controlType": "QUERY_DYNAMIC_TEXT",
+          "evaluationSubstitutionType": "PARAMETER",
```
nidhi-nair, line 384: Could you please explain why there are two of these?

trishaanand: For dictating the UI for the evaluated value. In case of TEMPLATE, the client would show the evaluated value as is. In case of SMART_SUBSTITUTE (was PARAMETER earlier which is used to display the prepared statement UI), the client would do the smart substitution to show the corrected JSON evaluation.

nidhi-nair (approved):

PR #7452, opened 2021-09-14
feat: set `modifiedBy` during create, clone, fork, import app

### LINKED ISSUE #7451: [Feature] Set `modifiedBy` in application during create, clone, duplicate and import
## Summary

We need to set `modifiedBy` attribute of application to current user when they create, duplicate, clone or import an application.

## Motivation
In order to show last updated by information of an application in user home page, we need this information to be set when an application is created.

## Additional Context
N/A

nayan-rafiq:
## Description

Adds `modifiedBy` field to application when an user create, duplicate, clone or import an application. This field is required by UI to show last who last updated an application in the new home page.

Fixes #7451 

## Type of change

## How Has This Been Tested?
- Unit tests
- Tested on local machine

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ApplicationPageServiceImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ApplicationPageServiceImpl.java
@@ -240,7 +240,12 @@ public class ApplicationPageServiceImpl implements ApplicationPageService {
         Mono<Application> applicationWithPoliciesMono = setApplicationPolicies(userMono, orgId, application);
 
         return applicationWithPoliciesMono
-                .flatMap(applicationService::createDefault)
+                .zipWith(userMono)
+                .flatMap(tuple -> {
+                    Application application1 = tuple.getT1();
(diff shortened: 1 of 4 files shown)


abhvsn (commented):
LGTM.

Does it makes sense not to include this change in clone app and in turn fork as well. Because the way I'm seeing this field is giving users a way to know who has modified the application which will involve updating/adding pages, actions, layouts etc. If we include this in cloning as well I'm little skeptical if we are providing the actual info to the user, any thoughts?

nayan-rafiq:
> LGTM.
> 
> Does it makes sense not to include this change in clone app and in turn fork as well. Because the way I'm seeing this field is giving users a way to know who has modified the application which will involve updating/adding pages, actions, layouts etc. If we include this in cloning as well I'm little skeptical if we are providing the actual info to the user, any thoughts?

Me and @albinAppsmith thought about this and we decided to treat clone and fork as edit events so that other users will be able to know who cloned or forked an application. We also thought that it might not make sense to some users. So we decided to implement this way in the first phase and update later if most people think something else.

abhvsn (approved):

PR #7480, opened 2021-09-15
feat: Server restart API, and restart detection API

sharat87:
This PR adds a `/api/v1/admin/restart` API that is available for superusers to initiate a restart of the server. This API isn't final yet.

A work-in-progress here is detection of whether a restart is required when admin config changes are applied.


sharat87:
Regarding hot-reloading changes in env variables, I’m able to do so for some variables like `APPSMITH_MAIL_ENABLED`, `APPSMITH_DISABLE_TELEMETRY`, `APPSMITH_RECAPTCHA_SITE_KEY` etc., but these have a problem. The client code also uses these variables, and I am able to hot-reload only on the server. If I want the client to learn about these changes, I need to restart the client nginx as well. We have two options here:

1. Just silently restart nginx. It takes a few hundred milliseconds and shouldn’t cause any downtime, unlike restarting server which takes a few seconds proper.
1. Have client not depend on these environment variables, but instead fetch them from the server on page load. So, to detect if mail is enabled, instead of relying on env variables, the server will tell whether mail is enabled or not. Similarly for the other env variables as well.

I’m personally leaning towards the second. Thoughts? Any other ideas? @mohanarpit

abhvsn:
> Regarding hot-reloading changes in env variables, I’m able to do so for some variables like `APPSMITH_MAIL_ENABLED`, `APPSMITH_DISABLE_TELEMETRY`, `APPSMITH_RECAPTCHA_SITE_KEY` etc., but these have a problem. The client code also uses these variables, and I am able to hot-reload only on the server. If I want the client to learn about these changes, I need to restart the client nginx as well. We have two options here:
> 
>     1. Just silently restart nginx. It takes a few hundred milliseconds and shouldn’t cause any downtime, unlike restarting server which takes a few seconds proper.
> 
>     2. Have client not depend on these environment variables, but instead fetch them from the server on page load. So, to detect if mail is enabled, instead of relying on env variables, the server will tell whether mail is enabled or not. Similarly for the other env variables as well.
> 
> 
> I’m personally leaning towards the second. Thoughts? Any other ideas? @mohanarpit

Considering the frequency superuser will be updating these env vars IMO first approach will be good enough in this case

sharat87:
> Considering the frequency with which the superuser will be updating these env vars IMO the first approach will be good enough in this case

Yeah, figured we take that a bit further than that 😄

Since this superuser config won't be used often, we've merged the save and restart. Not just nginx, but server will also be restarted right away, showing the user a spinner.

Speaking of which, I need to make this change in the API.

sharat87:
Now we always restart after saving changes in the admin config UI. The UI reflects this, in that, the CTA is termed along the lines of "Save and restart", and the UI shows a loading modal until the server starts to respond again.

sharat87:
@abhvsn, yeah, technically, we can implement so that restart won't always be required, and the server indeed reloads many env variables without needing a restart. But the thing is, many of these env variables are used on the client as well, and the nginx container today isn't capable of hot-reloading these env variables.

Additionally, the config UI is supposed to be used sparingly. The idea is to start with a simple and predictable definition for this UI like, _you make a change in admin settings, you restart_. From there, depending on usage, perhaps we move towards implementing hot-reloading on the client container as well.

Also, if changing some env variables require a restart a some don't, that means the server needs to inform the client whether a restart is required, _after_ the env changes are saved. Which means there's now an interim state, where env changes are saved, but a restart is pending. So we need to keep alerting the admin that a restart is pending, which is just a whole game in itself.

sharat87:
Just resolved merge conflict. Can I get an approval again please?

PR #7560, opened 2021-09-17
Introducing SaaS integrations as a plugin

nidhi-nair:
WIP

### DIFF
--- a/app/client/src/components/editorComponents/ActionCreator/index.tsx
+++ b/app/client/src/components/editorComponents/ActionCreator/index.tsx
@@ -391,7 +391,8 @@ function getIntegrationOptionsWithChildren(
   const apis = actions.filter(
     (action) =>
       action.config.pluginType === PluginType.API ||
-      action.config.pluginType === PluginType.SAAS,
+      action.config.pluginType === PluginType.SAAS ||
+      action.config.pluginType === PluginType.REMOTE,
   );
   const option = options.find(
     (option) => option.value === ActionType.integration,
--- a/app/client/src/components/editorComponents/CodeEditor/generateQuickCommands.tsx
+++ b/app/client/src/components/editorComponents/CodeEditor/generateQuickCommands.tsx
@@ -104,6 +104,7 @@ function Command(props: {
             DB: <DataSourcesColoredIcon />,
             API: <ApisIcon />,
             SAAS: <DataSourcesColoredIcon />,
+            REMOTE: <DataSourcesColoredIcon />,
             JS: <JsIcon />,
           }[props.pluginType]}
         {props.imgSrc && <img src={props.imgSrc} />}
--- a/app/client/src/components/editorComponents/LightningMenu/hooks.ts
+++ b/app/client/src/components/editorComponents/LightningMenu/hooks.ts
@@ -30,7 +30,11 @@ export const useActions = () => {
     .map((action) => action.config);
 
   const saas: Action[] = actions
-    .filter((action) => action.config.pluginType === PluginType.SAAS)
+    .filter(
+      (action) =>
+        action.config.pluginType === PluginType.SAAS ||
+        action.config.pluginType === PluginType.REMOTE,
+    )
     .map((action) => action.config);
 
   return { apis, queries, saas };
--- a/app/client/src/constants/AppsmithActionConstants/ActionConstants.tsx
+++ b/app/client/src/constants/AppsmithActionConstants/ActionConstants.tsx
@@ -119,6 +119,7 @@ export const defaultActionSettings: Record<PluginType, any> = {
   [PluginType.API]: apiActionSettingsConfig,
   [PluginType.DB]: queryActionSettingsConfig,
   [PluginType.SAAS]: saasActionSettingsConfig,
+  [PluginType.REMOTE]: saasActionSettingsConfig,
   [PluginType.JS]: [],
 };
 
@@ -126,6 +127,7 @@ export const defaultActionEditorConfigs: Record<PluginType, any> = {
   [PluginType.API]: apiActionEditorConfig,
   [PluginType.DB]: [],
   [PluginType.SAAS]: [],
+  [PluginType.REMOTE]: [],
   [PluginType.JS]: [],
 };
 
@@ -136,5 +138,6 @@ export const defaultActionDependenciesConfig: Record<
   [PluginType.API]: apiActionDependencyConfig,
   [PluginType.DB]: {},
   [PluginType.SAAS]: {},
+  [PluginType.REMOTE]: {},
   [PluginType.JS]: {},
 };
--- a/app/client/src/entities/Action/index.ts
+++ b/app/client/src/entities/Action/index.ts
@@ -7,6 +7,7 @@ export enum PluginType {
   DB = "DB",
   SAAS = "SAAS",
   JS = "JS",
+  REMOTE = "REMOTE",
 }
 
 export enum PaginationType {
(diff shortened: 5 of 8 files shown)


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/DatasourceConfiguration.java
```diff
@@ -38,6 +38,8 @@
 
     List<Property> headers;
 
+    String command;
```
trishaanand, line 41: Sorry for the stupid question, but what does command signify in datasource configuration?

app/server/appsmith-plugins/saasPlugin/src/main/java/com/external/helpers/BodyReceiver.java
```diff
@@ -0,0 +1,228 @@
...
+
+/**
+ * This receiver essentially instantiates a custom ClientHttpRequest that stores the request body via a subscriber
+ * The BodyInserter instance from our original request inserts into this subscriber that we can then retrieve
+ * using receiveValue.
+ *
+ * We do this so that the received value that we display to the user is exactly the same as
+ * the body tht is sent over the wire
+ */
+public class BodyReceiver {
```
trishaanand, line 41: QQ : If this is exactly the same as Rest Api plugin, do you think it might make sense to add this to interfaces as helper?

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Plugin.java
```diff
@@ -30,6 +30,8 @@
 
     String packageName;
 
+    String pluginName;
```
trishaanand, line 33: QQ : How is this different from `name`?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/OrganizationServiceImpl.java
```diff
@@ -340,4 +338,9 @@ public OrganizationServiceImpl(Scheduler scheduler,
                 });
     }
 
+    @Override
+    public Flux<Organization> getAll() {
+        return repository.findAll();
```
trishaanand, line 343: Would you like to filter out the deleted organizations? I know that today we don't support deletion of organizations, but very soon we should. When we do (and go with the soft delete as we do for other objects), we might end up here with organizations which are basically defunct.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/PluginServiceImpl.java
```diff
@@ -185,6 +185,20 @@ public PluginServiceImpl(Scheduler scheduler,
                 .switchIfEmpty(Mono.empty());
     }
 
+    @Override
+    public Flux<Organization> installDefaultPlugins(List<Plugin> plugins) {
+        final List<OrganizationPlugin> newOrganizationPlugins = plugins.stream()
```
trishaanand, line 190: Would you like to add a check for if the plugin is indeed default install or not and skip out in case it isn't? This ensures that a future developer who might use this service function and forgets to add a check before calling this doesn't end up installing a non required plugin to all the organizations.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/DatasourceConfiguration.java
nidhi-nair: Same as `actionConfiguration.formData.command`. We don't have that concept for datasource right now because we are only using it for authentication. I was debating whether we should introduce formData right now and map it to authentication or introduce command for now and do a migration if we extend datasources later. Chose the latter since the rest of the contributor experience is designed around not reusing shared config

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Plugin.java
nidhi-nair: `name` is what the end user will see, pluginName is the identifier for saas plugins internally (different from package name because package name for all of these is `saas-plugin`

trishaanand: Got it. Thanks!

app/server/appsmith-server/src/main/java/com/appsmith/server/services/OrganizationServiceImpl.java
nidhi-nair: Yeah I was wondering about that too. Then I figured we can do that when we support deleted organizations (because I wasn't sure whether we would support getting orgs back in that implementation and then we'd have to think about how to get default installed plugins into orgs that are reinstated like that). What do you think?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/Datasource.java
```diff
@@ -1,8 +1,5 @@
-package com.appsmith.server.domains;
+package com.appsmith.external.models;
```
sumitsum, line 2: QQ: Just for my understanding - why is the datasource moved to `appsmith-interface` from `appsmith-server` ?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/PluginServiceImpl.java
```diff
@@ -649,8 +675,21 @@ private Map loadPluginResourceGivenPlugin(Plugin plugin, String resourcePath) {
...
                 .map(plugin -> {
-                    if (resourcePath.equals("editor.json") && plugin.getUiComponent().equals(UQI_DB_EDITOR_FORM)) {
-                        return loadEditorPluginResourceUqi(plugin);
+                    if (resourcePath.equals("editor.json")) {
+                        // UI config will be available if this plugin is sourced from the cloud
+                        if (plugin.getActionUiConfig() != null) {
+                            return plugin.getActionUiConfig();
+                        }
+                        // For UQI, use another format of loading the config
+                        if (plugin.getUiComponent().equals(UQI_DB_EDITOR_FORM)) {
```
sumitsum, line 684: Nitpick: Can we change it to `UQI_DB_EDITOR_FORM.equasl(...)`. Same for the following `if` condition. In case of a time crunch, this can also be picked up as part of another PR - since we have a merge freeze deadline. Leaving this decision up to you.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/Datasource.java
nidhi-nair: It needs to be sent on the wire for saas plugins to identify individual datasources. But you're right, I was uncomfortable doing this as well. I've created a new DTO to share the id with the cloud.

trishaanand (dismissed):

sumitsum (approved):

arunvjn (approved):

arunvjn (approved):

arunvjn (approved):

app/client/src/selectors/entitiesSelector.ts
```diff
@@ -181,6 +181,13 @@ export const getDBPlugins = createSelector(getPlugins, (plugins) =>
   plugins.filter((plugin) => plugin.type === PluginType.DB),
 );
 
+export const getDBAndRemotePlugins = createSelector(getPlugins, (plugins) =>
+  plugins.filter(
+    (plugin) =>
+      plugin.type === PluginType.DB || plugin.type === PluginType.REMOTE,
```
trishaanand, line 187: I dont understand the utility of this function. What is the purpose that is being served here by fetching only DB and REMOTE plugins? What about new types of plugins which we would add?

trishaanand (approved):

app/client/src/selectors/entitiesSelector.ts
nidhi-nair: The function that this one replaces only used to fetch DB plugins (for populating in the dropdown of other available datasources). We have now started counting remote plugins as well so that those show up as well

trishaanand (approved):

PR #7590, opened 2021-09-18
feat: Switch Group widget

### LINKED ISSUE #3350: [Feature] Switch group widget
## Summary 
Allow users to create a list of switch widgets so that they can easily use them in a form.
Similar to #3344 

## Requirement 

| Title   | User Story | Importance | Notes |
| ------------- | ------------- | ------------- | ------------- |
| Add switch | Ability to add new switch & drag to reorder the sequence | High  | - |
| Inherit properties | Inherit all the properties of switch widget for each row | High  | - |
| Internal property - `selectedValues` | API reference to get all selected values from the switch group, will be an array  | High  | - |
| Action - `onChange` | API to trigger an action on toggle of a switch inside the group | High  | - |

wmdev0808:
## Description
Allow users to create a list of switch widgets so that they can easily use them in a form.

Fixes #3350

## Type of change

## How Has This Been Tested?

- Test A
- Test B

## Checklist:

## Figma link to the icon
https://www.figma.com/file/ATfTB5F9Q7ZiI0yYckuXW7/Deprecated-Design-System?node-id=72%3A8452

    // Code coverage diff between base branch:release and head branch: feature/3350-switch-group-widget 
 :red_circle: | total | 55.12 **(0)** | 37.04 **(-0.03)** | 34.01 **(0.01)** | 55.64 **(0.01)**
 :red_circle: | app/client/src/icons/WidgetIcons.tsx | 62.34 **(-0.33)** | 100 **(0)** | 19.44 **(-0.56)** | 62.34 **(-0.33)**
 :green_circle: | app/client/src/utils/AppsmithUtils.tsx | 57.21 **(2.02)** | 31.48 **(2.63)** | 36.17 **(2.84)** | 52.36 **(2.36)**
 :sparkles: :new: | **app/client/src/widgets/SwitchGroupWidget/index.ts** | **100** | **100** | **100** | **100**
 :sparkles: :new: | **app/client/src/widgets/SwitchGroupWidget/component/index.tsx** | **26.67** | **12.5** | **0** | **36.36**
 :sparkles: :new: | **app/client/src/widgets/SwitchGroupWidget/widget/index.tsx** | **52.38** | **0** | **60** | **52.5**</details>


app/client/src/widgets/SwitchGroupWidget/widget/index.tsx
```diff
@@ -0,0 +1,293 @@
...
+import SwitchGroupComponent from "../component";
+
+class SwitchGroupWidget extends BaseWidget<
+  SwitchGroupWidgetProps,
+  WidgetState
+> {
+  static getPropertyPaneConfig() {
+    return [
+      {
+        sectionName: "General",
```
riodeuno, line 18: @wmdev0808 Let's add a cypress test case to test each of the following properties.

app/client/src/widgets/SwitchGroupWidget/widget/index.tsx
```diff
@@ -0,0 +1,293 @@
...
+}
+
+export interface SwitchGroupWidgetProps extends WidgetProps {
+  isDisabled?: boolean;
+  isVisible?: boolean;
+  isInline?: boolean;
+  groupItems: Record<
+    string,
+    {
+      widgetId: string;
```
riodeuno, line 278: Let's create a new type called `SwtichItemConfig`. And let's configure this `groupItems: Record<string, SwitchItemConfig>`

riodeuno (changes requested):
@wmdev0808 Nice work with this widget! I've partially reviewed this. I'll review the rest of the files today.

somangshu:
@wmdev0808 feedback from the deploy preview:

- The switch widget does not function, cant change value (happens until we add a new option)
- The default widget size on drag and drop should be minimum rows
- Adding new item crashes the app (critical issue)
- The accessibility to the key / value pair should be similar compared to the checkbox group / radio group
- A default selected option should be provided
- The icon for the widget needs to be created
- Expose `required` API
- Maintain the property order (consistent to other group widgets)

app/client/src/widgets/SwitchGroupWidget/widget/index.tsx
```diff
@@ -0,0 +1,302 @@
...
+                    ],
+                  },
+                },
+              },
+            },
+            evaluationSubstitutionType:
+              EvaluationSubstitutionType.SMART_SUBSTITUTE,
+          },
+          {
+            helpText: "Selects values of the options checked by default",
```
somangshu, line 112: Add here > `Enter comma separated values for multiple selected`

somangshu (changes requested):

somangshu (dismissed):

momcilo-appsmith:
Here is the icon design: https://www.figma.com/file/i80YbQLLs5mXthjIHclPpX/Appsmith-Design-System?node-id=3172%3A13666

app/client/src/components/propertyControls/ItemsControl.tsx
```diff
@@ -0,0 +1,266 @@
...
+    index,
+  ]);
+  return (
+    <ItemWrapper>
+      <StyledDragIcon height={20} width={20} />
+      <StyledOptionControlInputGroup
+        dataType="text"
+        defaultValue={item.label}
+        onChange={(value: string) => {
+          debouncedUpdate(index, value);
```
riodeuno, line 90: Let's use `useCallback` here as well as the `deleteOption` on line 99.

app/client/src/components/propertyControls/ItemsControl.tsx
```diff
@@ -0,0 +1,266 @@
...
+        obj[each.id] = {
+          ...each,
+          index,
+        };
+        return obj;
+      },
+      {},
+    );
+    this.deleteProperties([`${this.props.propertyName}.${itemId}.isVisible`]);
+    this.updateProperty(this.props.propertyName, updatedObj);
```
riodeuno, line 224: The look and feel of this control is not consistent with the other controls. 
@momcilo-appsmith @somangshu Is there a reason why we're adding a new control here and not going with what we have in the select widget?
Screenshot 2021-09-22 at 11 21 13 AM

app/client/src/widgets/SwitchGroupWidget/component/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+  return (
+    <SwitchGroupContainer
+      inline={inline}
+      optionCount={options.length}
+      valid={valid}
+    >
+      {options &&
+        options.length > 0 &&
+        [...options].map((option: OptionProps) => (
```
riodeuno, line 80: Is there a reason we have to spread this here? I believe we can do without the spread.

app/client/src/widgets/SwitchGroupWidget/component/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+          <StyledSwitch
+            checked={(selectedValues || []).includes(option.value)}
+            disabled={disabled}
+            inline={inline}
+            key={generateReactKey()}
```
riodeuno, line 85: These keys will be different for every render. Can we make this consistent until the options change? For example, use the options value or any other value which will be unique among siblings but consistent across renders if the values don't change?

app/client/src/widgets/SwitchGroupWidget/component/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+    <SwitchGroupContainer
+      inline={inline}
+      optionCount={options.length}
+      valid={valid}
+    >
+      {options &&
+        options.length > 0 &&
+        [...options].map((option: OptionProps) => (
+          <StyledSwitch
+            checked={(selectedValues || []).includes(option.value)}
```
riodeuno, line 82: Is it possible to have a default for the `selectedValues`? That way we won't have to do `selectedValues || []`

app/client/src/widgets/SwitchGroupWidget/icon.svg
```diff
@@ -0,0 +1,9 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path fill-rule="evenodd" clip-rule="evenodd" d="M11 6C11 4.89543 10.1046 4 9 4H6C4.89543 4 4 4.89543 4 6C4 7.10457 4.89543 8 6 8H9C10.1046 8 11 7.10457 11 6ZM9 7C9.55228 7 10 6.55228 10 6C10 5.44772 9.55228 5 9 5C8.44772 5 8 5.44772 8 6C8 6.55228 8.44772 7 9 7Z" fill="#4B4848"/>
+<path d="M12.9998 5.80005C12.9998 5.38584 13.3355 5.05005 13.7498 5.05005L19.2498 5.05005C19.664 5.05005 19.9998 5.38584 19.9998 5.80005C19.9998 6.21426 19.664 6.55005 19.2498 6.55005L13.7498 6.55005C13.3355 6.55005 12.9998 6.21426 12.9998 5.80005Z" fill="#4B4848"/>
+<path fill-rule="evenodd" clip-rule="evenodd" d="M11 12C11 10.8954 10.1046 10 9 10H6C4.89543 10 4 10.8954 4 12C4 13.1046 4.89543 14 6 14H9C10.1046 14 11 13.1046 11 12ZM9 13C9.55228 13 10 12.5523 10 12C10 11.4477 9.55228 11 9 11C8.44772 11 8 11.4477 8 12C8 12.5523 8.44772 13 9 13Z" fill="#4B4848"/>
```
riodeuno, line 4: Please use an online SVG optimiser to create a smaller SVG file here.

riodeuno (changes requested):
A few changes requested. We need comprehensive cypress tests here for the following:
- All properties of the widget
- The new property control

app/client/src/components/propertyControls/ItemsControl.tsx
somangshu: @riodeuno for all group widgets this has been the default. In case we take a call to remove this control then it can be taken up separately considering all widgets

app/client/src/components/propertyControls/ItemsControl.tsx
wmdev0808: This control should be removed

app/client/src/components/propertyControls/ItemsControl.tsx
wmdev0808: This control should be removed because the current implementation does not use this control anymore

app/client/src/widgets/SwitchGroupWidget/component/index.tsx
wmdev0808: You're right, thank you

app/client/src/components/propertyControls/ItemsControl.tsx
somangshu: @wmdev0808 what do you mean by "current implementation does not use this control anymore"?

app/client/src/widgets/SwitchGroupWidget/component/index.tsx
wmdev0808: You're right, thank you

app/client/src/widgets/SwitchGroupWidget/component/index.tsx
wmdev0808: Agree with you, thank you

app/client/src/widgets/SwitchGroupWidget/component/index.tsx
wmdev0808: It was just my mess, selectedValues are always an array, thank you

app/client/src/widgets/SwitchGroupWidget/icon.svg
wmdev0808: I am just using the svg from the design which is provided by other guys.
That's why, it would be better if any designer can provide me with the new optimized one

somangshu:
@momcilo-appsmith i think we need the light version of the icon, the color doesn't match others, From the same file attached can you create the light version, Like done for others?

momcilo-appsmith:
Here is the light version: https://www.figma.com/file/i80YbQLLs5mXthjIHclPpX/Appsmith-Design-System?node-id=3253%3A13655. Attached to the issue as well.

app/client/src/widgets/SwitchGroupWidget/icon.svg
somangshu: This is done

somangshu (dismissed):
Icon replaced, Approving to run tests

app/client/src/widgets/SwitchGroupWidget/widget/index.tsx
```diff
@@ -0,0 +1,303 @@
...
+            controlType: "INPUT_TEXT",
+            isBindProperty: true,
+            isTriggerProperty: false,
+            validation: {
+              type: ValidationTypes.FUNCTION,
+              params: {
+                fn: defaultSelectedValuesValidation,
+                expected: {
+                  type: "Value or Array of values",
+                  example: `value1 | ['value1', 'value2']`,
```
riodeuno, line 126: @wmdev0808 I was wondering if the defaultSelectedValues would always be an array? Is there a scenario where `value1` in the example make sense?

If this is always going to be an array, then we don't need to use `ValidationTypes.FUNCTION` and we can use a standard `ValidationTypes.ARRAY` with detailed configurations.

I'm asking this because, we want to avoid `ValidationTypes.FUNCTION` where possible.

riodeuno (changes requested):
Everything looks great! We have cypress tests, the code hierarchy is good.

I have one comment about the `defaultSelectedValues` property, as well as a request for unit tests for a function.

app/client/src/widgets/SwitchGroupWidget/widget/index.tsx
wmdev0808: Users are allowed to input string or array for this field like this: 
- value1[, value2, ... value_n]
- [value1[, value2, ...value_n]
=> This means that neither ValidationTypes.TEXT nor ValidationTypes.ARRAY is fit for this scenario.

defaultSelectedValues are always evaluated as an array by that validation function, I think.

riodeuno: Hmm, maybe I'm missing something here. @somangshu I believe the Switch Widget should work with `defaultSelected` while, Switch Group Widget should work with only arrays `defaultSelectedValues`. Can you let me know if there is a reason for this to not be the case?

somangshu: I agree that the default can we the array in case of switch / checkbox group. However even comma separated string works. 
I will support the argument to have default as array here! @wmdev0808 what do you think? I dont see any specific use case of this to be string

wmdev0808: @somangshu I think it is just for UX
My concern is how we can handle the validation without using function.
We are required to refer to `options` property for validation.
@riodeuno , What's your solution for the above?

riodeuno: @wmdev0808 I was wondering if we can remove the need for the `defaultSelectedValues` to be a string. This way, we can use `ValidationTypes.ARRAY` to validate `defaultSelectedValues`.

@somangshu I believe the same applies for the select widget, however, i believe we still support arrays in the Select Widget for backwards compatibility. Here, we'd like to get this right from the get go.

wmdev0808: @riodeuno Please provide me with the validation config parameters for better understanding

riodeuno: This means that the expected value will be `['value1', 'value2']` or `Array<string>`
`{ type: ValidationTypes.ARRAY, params: { default: [], children: { type: ValidationTypes.TEXT } }` 
Something like this? This will make the expected value `Array<string>`

If we really want to validate by checking if the `defaultSelectedValues` exist in the `options`, at runtime, then we can still use `ValidationTypes.FUNCTION` but validate only array. This will make the expected value `['value1', 'value2']`

riodeuno: Please let me know if these will not work.

wmdev0808: @riodeuno Got it.
After all, you just want to validate Array<string>.
Now, am I required to revise the code base for your suggestion?

riodeuno: According to the example in the current validation function, we're validating `string[]` as the type. 
The question is, are we trying to validate if the `defaultSelectedValues` exist in the `options` as well, or are we simply validating if the value is a `string[]`. If it is the latter, then we can go with the validation configuration instead of function. If it is the former, we can update the function to make such checks, and still keep the expected value as `string[]`

somangshu:
@riodeuno kindly review the changes here

app/client/src/widgets/SwitchGroupWidget/widget/index.tsx
```diff
@@ -0,0 +1,248 @@
...
+            controlType: "SWITCH",
+            isJSConvertible: true,
+            isBindProperty: true,
+            isTriggerProperty: false,
+            validation: { type: ValidationTypes.BOOLEAN },
+          },
+          {
+            propertyName: "isRequired",
+            label: "Required",
+            helpText: "Makes input to the widget mandatory",
```
riodeuno, line 91: Does this property mean that this switch has to be turned "on" when inside a form? If that is the case, can we change the helpText like so:

`Turning this property on is mandatory for this field to be valid, inside a form widget`

riodeuno (changes requested):

riodeuno:
@vivekverma2312 we might want to start QA for this widget.

app/client/src/utils/AppsmithUtils.tsx
```diff
@@ -344,3 +344,12 @@ export const parseBlobUrl = (blobId: string) => {
   }/${blobId.substring(5)}`;
   return url.split("?type=");
 };
+
+export const getCamelCaseString = (sourceString: string) => {
```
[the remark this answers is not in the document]
riodeuno: I don't see the jest test for this. Can we add this?

app/client/src/widgets/SwitchGroupWidget/widget/index.tsx
```diff
@@ -0,0 +1,248 @@
...
+                  },
+                },
+              },
+            },
+            evaluationSubstitutionType:
+              EvaluationSubstitutionType.SMART_SUBSTITUTE,
+          },
+          {
+            helpText:
+              "Selects values of the options checked by default. Enter comma separated values for multiple selected",
```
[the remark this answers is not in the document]
wmdev0808: > We're validating an array here. Will comma separated values work?

No, it is just validating string[]

app/client/src/widgets/SwitchGroupWidget/widget/index.tsx
```diff
@@ -0,0 +1,248 @@
...
+          {
+            propertyName: "isRequired",
+            label: "Required",
+            helpText: "Makes input to the widget mandatory",
+            controlType: "SWITCH",
+            isJSConvertible: true,
+            isBindProperty: true,
+            isTriggerProperty: false,
+            validation: {
+              type: ValidationTypes.BOOLEAN,
```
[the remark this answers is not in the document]
wmdev0808: done

app/client/src/widgets/SwitchGroupWidget/widget/index.tsx
wmdev0808: I think so. @somangshu Please confirm this

app/client/src/widgets/SwitchGroupWidget/widget/index.tsx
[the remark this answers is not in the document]
Tooluloope: @riodeuno  Here the validation would pass, but the component would fail, cos the Component expects `Array<{ "label": "string", "value": "string" }>`

So it's either we handle it Validation Level or Handle it component level

Screenshot 2021-09-29 at 09 31 06
Screenshot 2021-09-29 at 09 30 25

app/client/src/widgets/SwitchGroupWidget/widget/index.tsx
```diff
@@ -0,0 +1,247 @@
...
+    return [
+      {
+        sectionName: "General",
+        children: [
+          {
+            helpText:
+              "Displays a list of options for a user to select. Values must be unique",
+            propertyName: "options",
+            label: "Options",
+            controlType: "OPTION_INPUT",
```
Tooluloope, line 25: This Control type still has this issue https://www.loom.com/share/f32e748f48754b768df2ccc22eb148d5 
@riodeuno, should we still use this?

app/client/src/widgets/SwitchGroupWidget/widget/index.tsx
somangshu: @riodeuno I believe the UX right now is currently inline to what already happens on Appsmith right now. I see that none of the widget inside the form widget automatically turn required in the form widget so the above message change should not be required.

riodeuno:
@wmdev0808 @somangshu Found some issues with the widget, some of these could exist in other widgets as well. These need triaging.

- [ ] When we have the same or empty values for two options, the widget goes blank
Screenshot 2021-10-21 at 12 18 24 PM
Screenshot 2021-10-21 at 12 19 42 PM

- [ ] When we start typing in the options' JS input, the widget crashes.
Screenshot 2021-10-21 at 12 21 16 PM


- [ ] Due to the string validation special chars are not accepted in the validation of the default values. We can either update the validation to allow for these characters or disallow these types of values to be set in the options.

Screenshot 2021-10-21 at 12 20 10 PM

- [ ] The icon needs to be updated
Screenshot 2021-10-21 at 12 16 20 PM

- [ ] There is some strange background in the options fields
Screenshot 2021-10-21 at 12 17 15 PM

wmdev0808:
https://github.com/appsmithorg/appsmith/actions/runs/1424854034

app/client/src/utils/AppsmithUtils.tsx
```diff
@@ -345,3 +345,38 @@ export const parseBlobUrl = (blobId: string) => {
...
+  return out;
+};
+
+/**
+ * Deeply compare two arrays
+ * @param array1 array1
+ * @param array2 array2
+ * @returns boolean
+ */
+export const isArrayEqual = (array1: any[] = [], array2: any[] = []) => {
```
riodeuno, line 375: @wmdev0808 These functions are available in `lodash` we can use that instead of maintaining custom functions in our codebase.

riodeuno (changes requested):

wmdev0808:
https://github.com/appsmithorg/appsmith/actions/runs/1425066219

app/client/src/widgets/SwitchGroupWidget/widget/index.tsx
```diff
@@ -0,0 +1,272 @@
...
+      if (optionsLength >= prevOptionsLength) {
+        const defaultSelectedValues = this.props.defaultSelectedValues;
+        selectedValues = selectedValues.concat(
+          defaultSelectedValues.filter((defaultSelectedValue: string) =>
+            diffOptions.includes(defaultSelectedValue),
+          ),
+        );
+      }
+
+      this.props.updateWidgetMetaProperty("selectedValues", selectedValues, {
```
riodeuno, line 204: @wmdev0808 I was wondering if we can make the `selectedValues` a derived property? For example, in the table widget we store `selectedRowIndices` in the meta, and use this to compute `selectedRows` array. 

There are a couple of reasons I was wondering about this.
- There is quite a bit of computations happening in `componentDidUpdate`. 
- `options` will be available correctly in the evaluations, which means that the `selectedValues` can be computed without figuring out if the options have changed. So, we will have smaller amount of code to maintain.

If we implement this, in the manner I am requesting --
- What will happen to the bindings to `selectedValues` when called in `onSelectionChange` trigger? Will it still give us the correct values?

app/client/src/widgets/SwitchGroupWidget/widget/index.tsx
```diff
@@ -0,0 +1,272 @@
...
+      const prevOptions = compact(prevProps.options).map(
+        (prevOption) => prevOption.value,
+      );
+      const options = compact(this.props.options).map((option) => option.value);
+
+      let diffOptions = prevOptions.filter(
+        (prevOption) => !options.includes(prevOption),
+      );
+
+      if (optionsLength >= prevOptionsLength) {
```
riodeuno, line 187: @wmdev0808 What happens if the `optionsLength` is smaller than `prevOptionsLength`?

app/client/src/widgets/SwitchGroupWidget/widget/index.tsx
```diff
@@ -0,0 +1,272 @@
...
+    return "SWITCH_GROUP_WIDGET";
+  }
+
+  componentDidUpdate(prevProps: SwitchGroupWidgetProps) {
+    const prevOptionsLength = prevProps.options.length;
+    const optionsLength = this.props.options.length;
+
+    // Update if an option is added or removed and any value inside options changes
+    if (
+      optionsLength !== prevOptionsLength ||
```
riodeuno, line 173: @wmdev0808 I find that this code is difficult to follow. Would you mind adding comments to the code to help me understand how we're dealing with the following scenarios:
- [ ] When new options are added.
- [ ] When one or more option value(s) have changed
- [ ] When a selected option is deleted.
- [ ] When `defaultValues` have a value which no longer exists in the `options`

riodeuno (changes requested):

wmdev0808:
https://github.com/appsmithorg/appsmith/actions/runs/1436590807

riodeuno:
The icon needs to be updated. @somangshu Please help out with the icon.

Screenshot 2021-11-16 at 11 01 10 AM

The code overall looks fine to me.

@YogeshJayaseelan Please test this widget out thoroughly. We will want to provide odd values to the options and default values and find expected results. Another approach for testing would be to bind the `defaultValues` of the switch widget with a multi-select or checkboxgroup widget, to see if the switch widget reacts accordingly.

somangshu:
@riodeuno please approve the code if it looks good

@momcilo-appsmith @UXFalcon we need to the icon for this

riodeuno:
@somangshu There are merge conflicts and the icon needs to be updated. These changes are needed before approval.

wmdev0808:
https://github.com/appsmithorg/appsmith/actions/runs/1471539634

momcilo-appsmith:
New link for the icon: https://www.figma.com/file/ATfTB5F9Q7ZiI0yYckuXW7/Deprecated-Design-System?node-id=72%3A8452

wmdev0808:
https://github.com/appsmithorg/appsmith/actions/runs/1471936874

somangshu:
@wmdev0808 we have the icon from the design system. Kindly replace the old one
https://www.figma.com/file/ATfTB5F9Q7ZiI0yYckuXW7/Deprecated-Design-System?node-id=72%3A8452

Let me know if you have a problem accessing

wmdev0808:
> @wmdev0808 we have the icon from the design system. Kindly replace the old one https://www.figma.com/file/ATfTB5F9Q7ZiI0yYckuXW7/Deprecated-Design-System?node-id=72%3A8452
> 
> Let me know if you have a problem accessing

@somangshu I've already replaced it in the previous commit

somangshu:
@wmdev0808 related test is failing, [check this out](https://dashboard.cypress.io/projects/eyxvp8/runs/12061/test-results/30783d6a-1a84-4bc1-b038-f93471af73da?statuses=%5B%7B%22value%22%3A%22FAILED%22%2C%22label%22%3A%22FAILED%22%7D%5D)

somangshu:
Icon change verified and looks good to me. 

@riodeuno I am moving this to needsQA, will need one final approval from you before merging

somangshu (dismissed):

YogeshJayaseelan:
@wmdev0808 - Started testing this feature and came up with the below issues and testing is in-progress 

Issue 1 - When the user starts typing in the Option JS the widget crashes and throws - **Cannot read properties of undefined (reading,'length') in the console log** 

when {{null}} value is passed in option - **Cannot read properties of null (reading,'length') in the console log** 

https://www.loom.com/share/2dabe9e331794625842dc1b3b94d5fef

Issue 2 - Icon is not displayed in the entity explorer when the switch widget is dragged and dropped to the canvas

https://loom.com/share/10d9b7a939194301b9ed4b7e393efe10

wmdev0808:
> @wmdev0808 - Started testing this feature and came up with the below issues and testing is in-progress
> 
> Issue 1 - When the user starts typing in the Option JS the widget crashes and throws - **Cannot read properties of undefined (reading,'length') in the console log**
> 
> when {{null}} value is passed in option - **Cannot read properties of null (reading,'length') in the console log**
> 
> [LOOM DEMO](https://www.loom.com/share/2dabe9e331794625842dc1b3b94d5fef)
> 
> Issue 2 - Icon is not displayed in the entity explorer when the switch widget is dragged and dropped to the canvas
> 
> https://loom.com/share/10d9b7a939194301b9ed4b7e393efe10
> 
> Issue 3 - The evaluated values are not aligned properly when the default option values ate not separated by a comma
> 
> Screenshot 2021-11-25 at 4 46 13 PM
> 
> Issue 4 - Disabling the switch button - the toggle appears in blue colour
> 
> [LOOM DEMO](https://www.loom.com/share/7bd9ed7392da4c4faf13f05105a5f671)

@YogeshJayaseelan , 
- Issue 3: I think Evaluated value is only aligned properly if the evaluation of the value is done without error.
If you don't agree on it, please provide any other widget example with the same use case
- Issue 4: I cannot reproduce it

YogeshJayaseelan:
@wmdev0808 For issue 4 - When you mouse hover on the switch toggle then the mouse pointer icon will be displayed then click on the toggle to observe the blue colour 

Issue 5 - Switch button left and Right Alignment functionality is not available  - comparing with switch widget

riodeuno (approved):
Looks good to me! It took some time, but we seem to have a robust widget. Nice work @wmdev0808 !

YogeshJayaseelan:
@wmdev0808 Verified the issue fix and they are working as expected.

Issue 1 - Fixed,
Issue 2 - Fixed,
Issue 3 - Fixed,
Issue 4 - Not in scope 
Issue 5 - Fixed

riodeuno (approved):

PR #7592, opened 2021-09-18
feat: Where condition helper library using H2 in memory database

trishaanand:
This helper library service provides a functionality where the data and conditions can be passed to the library. The library then returns back filtered data. This is implemented using in memory H2 database where the data is first inserted, where query is generated from the conditions, and the results from the executed where query is returned

### DIFF
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/InMemoryDatabase.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/InMemoryDatabase.java
@@ -12,6 +12,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.bson.types.ObjectId;
 import org.springframework.stereotype.Component;
 
+import java.io.IOException;
 import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.ResultSet;
@@ -52,7 +53,9 @@ public class InMemoryDatabase {
             ConditionalOperator.EQ, "=",
             ConditionalOperator.NOT_EQ, "<>",
             ConditionalOperator.GT, ">",
-            ConditionalOperator.GTE, ">="
+            ConditionalOperator.GTE, ">=",
+            ConditionalOperator.IN, "IN",
+            ConditionalOperator.NOT_IN, "NOT IN"
     );
 
     private static Connection connection;
@@ -117,9 +120,38 @@ public class InMemoryDatabase {
             sb.append(" ");
             sb.append(sqlOp);
             sb.append(" ");
-            sb.append("'");
-            sb.append(value);
-            sb.append("'");
+
+            // These are array operations. Convert value into appropriate format and then append
+            if (operator == ConditionalOperator.IN || operator == ConditionalOperator.NOT_IN) {
+
+                StringBuilder valueBuilder = new StringBuilder("(");
+
+                // The array could be an array of Strings
+                if (value.contains("\"")) {
+                    try {
+                        List<String> stringValues = objectMapper.readValue(value, List.class);
+                        List<String> updatedStringValues = stringValues.stream().map(stringValue -> "\'" + stringValue + "\'").collect(Collectors.toList());
+                        String finalValues = String.join(",", updatedStringValues);
+                        valueBuilder.append(finalValues);
+                    } catch (IOException e) {
+                        e.printStackTrace();
+                    }
+                } else {
+                    String trimmedValue = value.replaceAll("^\\[|]$", "");
+                    valueBuilder.append(trimmedValue);
+                }
+
+                valueBuilder.append(")");
+                value = valueBuilder.toString();
+                sb.append(value);
+
+            } else {
+                // Since the value is not an array, surround the same with single quotes and append
+                sb.append("'");
+                sb.append(value);
+                sb.append("'");
+            }
+
             iterator++;
         }
 
@@ -318,7 +350,7 @@ public class InMemoryDatabase {
          *      "field3" : "12"
          * }
          *
-         * The scheme generated would be a Map as follows :
+         * The schema generated would be a Map as follows :
          * {
          *      field1 : DataType.STRING
          *      field2 : DataType.BOOLEAN
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/InMemoryDatabase.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/InMemoryDatabase.java
@@ -19,12 +19,13 @@ import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.ArrayList;
-import java.util.HashMap;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
+import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static com.appsmith.external.helpers.DataTypeStringUtils.stringToKnownDataTypeConverter;
 import static com.appsmith.external.services.InMemoryDatabase.Condition.generateConditionList;
@@ -36,7 +37,7 @@ public class InMemoryDatabase {
     private static String url = "jdbc:h2:mem:filterDb";
     private static ObjectMapper objectMapper = new ObjectMapper();
 
-    private static Map<DataType, String> sqlDataTypeMap = Map.of(
+    private static final Map<DataType, String> SQL_DATATYPE_MAP = Map.of(
             DataType.INTEGER, "INT",
             DataType.LONG, "BIGINT",
             DataType.FLOAT, "REAL",
@@ -45,7 +46,7 @@ public class InMemoryDatabase {
             DataType.STRING, "VARCHAR"
     );
 
-    private static Map<ConditionalOperator, String> sqlOperatorMapping = Map.of(
+    private static final Map<ConditionalOperator, String> SQL_OPERATOR_MAP = Map.of(
             ConditionalOperator.LT, "<",
             ConditionalOperator.LTE, "<=",
             ConditionalOperator.EQ, "=",
@@ -83,24 +84,33 @@ public class InMemoryDatabase {
         insertData(tableName, items, schema);
 
         // Filter the data
-        List<Map<String, Object>> finalResults = selectQuery(tableName, conditions);
+        List<Map<String, Object>> finalResults = executeFilterQuery(tableName, conditions);
+
+        // Drop the table
+        dropTable(tableName);
 
         ArrayNode finalResultsNode = objectMapper.valueToTree(finalResults);
 
         return finalResultsNode;
     }
 
-    public static List<Map<String, Object>> selectQuery(String tableName, List<Condition> conditions) {
+    public static List<Map<String, Object>> executeFilterQuery(String tableName, List<Condition> conditions) {
         StringBuilder sb = new StringBuilder("SELECT * FROM " + tableName + " WHERE ");
 
+        int iterator = 0;
         for (Condition condition : conditions) {
             String path = condition.getPath();
             ConditionalOperator operator = condition.getOperator();
             String value = condition.getValue();
 
-            String sqlOp = sqlOperatorMapping.get(operator);
+            String sqlOp = SQL_OPERATOR_MAP.get(operator);
             if (sqlOp == null) {
                 // Operator not supported. Handle the same // TODO
+            }
+
+            if (iterator > 0) {
+                // This is not the first condition. Append an `AND` before adding the next condition
+                sb.append(" AND ");
             }
 
             sb.append(path);
@@ -109,12 +119,8 @@ public class InMemoryDatabase {
             sb.append(" ");
             sb.append("'");
             sb.append(value);
-            sb.append("' AND ");
-        }
-
-        if (!conditions.isEmpty()) {
-            // Trim the last " AND " aka last 5 characters
-            sb.setLength(sb.length() - 5);
+            sb.append("'");
+            iterator++;
         }
 
         sb.append(";");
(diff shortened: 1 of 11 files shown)


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/InMemoryDatabase.java
```diff
@@ -0,0 +1,347 @@
...
+        // Trim the trailing comma
+        int i = columnNamesBuilder.lastIndexOf(",");
+        columnNamesBuilder.deleteCharAt(i);
+
+        columnNamesBuilder.append(")");
+
+        mainBuilder.append(columnNamesBuilder);
+        mainBuilder.append(" VALUES ");
+
+        for (JsonNode item : items) {
```
mohanarpit, line 199: Suggestion to batch the number of rows. IIRC, the max number of rows that can be inserted in a single statement is 1000.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/InMemoryDatabase.java
```diff
@@ -0,0 +1,347 @@
...
+        }
+    }
+
+
+    // INSERT INTO tableName (columnName1, columnName2) VALUES (data1, data2)
+    public static void insertData(String tableName, ArrayNode items, Map<String, DataType> schema) {
+
+        List<String> columnNames = schema.keySet().stream().collect(Collectors.toList());
+
+        StringBuilder mainBuilder = new StringBuilder("INSERT INTO ");
```
mohanarpit, line 180: Nit: Suggest renaming this to `insertQueryBuilder`

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/InMemoryDatabase.java
```diff
@@ -0,0 +1,347 @@
...
+
+        String tableName = generateTable(schema);
+
+        // insert the data
+        insertData(tableName, items, schema);
+
+        // Filter the data
+        List<Map<String, Object>> finalResults = selectQuery(tableName, conditions);
+
+        ArrayNode finalResultsNode = objectMapper.valueToTree(finalResults);
```
mohanarpit, line 88: Once the data has been filtered & queried, please drop the table so that we don't keep any user data in memory.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/InMemoryDatabase.java
```diff
@@ -0,0 +1,347 @@
...
+    }
+
+    public static String generateTable(Map<String, DataType> schema) throws SQLException {
+
+        if (connection == null || connection.isClosed() || !connection.isValid(5)) {
+            connection = DriverManager.getConnection(url);
+        }
+
+        // Generate table name
+        String generateUniqueId = new ObjectId().toString().toUpperCase(Locale.ROOT);
```
mohanarpit, line 245: Nit: Suggestion to remove `Locale.ROOT`

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/InMemoryDatabase.java
```diff
@@ -0,0 +1,347 @@
...
+        String tableName = new StringBuilder("tbl_").append(generateUniqueId).toString();
```
mohanarpit, line 246: Please add comment here explaining the reasoning behind this table name construction.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/InMemoryDatabase.java
```diff
@@ -0,0 +1,347 @@
...
+import static com.appsmith.external.services.InMemoryDatabase.Condition.generateConditionList;
+
+@Component
+public class InMemoryDatabase {
+
+    private static String JDBC_DRIVER = "org.h2.Driver";
+    private static String url = "jdbc:h2:mem:filterDb";
+    private static ObjectMapper objectMapper = new ObjectMapper();
+
+    private static Map<DataType, String> sqlDataTypeMap = Map.of(
```
mohanarpit, line 39: Nit: As per naming conventions, please use CAPS separated by underscore for static constant variables. Also, since this variable will not change, suggest using the `final` keyword.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/InMemoryDatabase.java
```diff
@@ -0,0 +1,347 @@
...
+
+        Boolean columnsAdded = false;
+        for (Map.Entry<String, DataType> entry : schema.entrySet()) {
+            String fieldName = entry.getKey();
+            DataType dataType = entry.getValue();
+
+            String sqlDataType = sqlDataTypeMap.get(dataType);
+            if (sqlDataType == null) {
+                // or throw an error for unsupported data type
+                continue;
```
mohanarpit, line 262: If we don't understand the column data type, suggest defaulting to string data type. It's generic and hopefully will handle all other scenarios.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/InMemoryDatabase.java
```diff
@@ -0,0 +1,347 @@
...
+
+
+    private static Map<String, DataType> generateSchema(JsonNode jsonNode) {
+        Map<String, DataType> schema = new HashMap<>();
+
+        jsonNode.fieldNames().forEachRemaining(name -> {
+            String value = jsonNode.get(name).asText();
+            DataType dataType = stringToKnownDataTypeConverter(value);
+
+            schema.put(name, dataType);
```
mohanarpit, line 300: Suggestion to use `.map()` in the stream & collect the results into a HashMap. This way we allocate memory only 1 time. Micro performance improvement.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/InMemoryDatabase.java
```diff
@@ -0,0 +1,347 @@
...
+
+            List<Condition> conditions = new ArrayList<>();
+
+            for (Object condition : conditionList) {
+                String path = ((Map<String, String>) condition).get("path");
+                ConditionalOperator operator = ConditionalOperator.valueOf(((Map<String, String>) condition).get("operator"));
+                String value = ((Map<String, String>) condition).get("value");
+                DataType dataType = stringToKnownDataTypeConverter(value);
+
+                conditions.add(new Condition(path, operator, value, dataType));
```
mohanarpit, line 340: Nit: Can we use `collectList` in a streaming collector? This way we are protected against multiple threads writing to the array.

mohanarpit:
This is a really good PR! Extends Appsmith functionality in plugins with a lot of flexibility and lets us provide deeper features with very little effort in the future.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/QueryDataService.java
```diff
@@ -0,0 +1,412 @@
...
+            ConditionalOperator.NOT_IN, "NOT IN"
+    );
+
+    private static Connection connection;
+
+    public QueryDataService() {
+        try {
+            connection = DriverManager.getConnection(url);
+        } catch (SQLException e) {
+            e.printStackTrace();
```
mohanarpit, line 68: Can we replace this with `log.error`?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/QueryDataService.java
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+    public static ArrayNode filterData(ArrayNode items, List<Condition> conditionList) {
+
+        if (items == null || items.size() == 0) {
+            return items;
+        }
+
+        if (!validConditionList(conditionList)) {
+            return items; // or throw an exception
```
mohanarpit, line 80: Suggestion to throw an exception here instead.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/QueryDataService.java
```diff
@@ -0,0 +1,412 @@
...
+            }
+            columnsAdded = true;
+            sb.append(fieldName);
+            sb.append(" ");
+            sb.append(sqlDataType);
+            sb.append(",");
+        }
+
+        // Delete trailing comma
+        if (columnsAdded) {
```
mohanarpit, line 345: Suggestion to move this inside the loop.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/QueryDataService.java
```diff
@@ -0,0 +1,412 @@
...
+        insertQueryBuilder.append(valuesBuilder);
+        insertQueryBuilder.append(";");
+
+        String finalInsertQuery = insertQueryBuilder.toString();
+        System.out.println(finalInsertQuery);
+
+        return finalInsertQuery;
+    }
+
+    private static void checkAndSetConnection() {
```
mohanarpit, line 299: Suggestion to rename this function to `checkAndGetConnection` and return the `Connection` object back. This way we can extend this function to connection pools as well.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/QueryDataService.java
```diff
@@ -0,0 +1,412 @@
...
+                sb.append(" AND ");
+            }
+
+            String path = condition.getPath();
+            ConditionalOperator operator = condition.getOperator();
+            String value = condition.getValue();
+
+            String sqlOp = SQL_OPERATOR_MAP.get(operator);
+            if (sqlOp == null) {
+                // Operator not supported. Handle the same // TODO
```
mohanarpit, line 168: Can we handle this scenario by throwing an exception?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/QueryDataService.java
```diff
@@ -0,0 +1,412 @@
...
+                    try {
+                        List<String> stringValues = objectMapper.readValue(value, List.class);
+                        List<String> updatedStringValues = stringValues.stream().map(stringValue -> "\'" + stringValue + "\'").collect(Collectors.toList());
+                        String finalValues = String.join(",", updatedStringValues);
+                        valueBuilder.append(finalValues);
+                    } catch (IOException e) {
+                        e.printStackTrace();
+                    }
+                } else {
+                    String trimmedValue = value.replaceAll("^\\[|]$", "");
```
mohanarpit, line 192: Suggestion to add a comment explaining this regex.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/QueryDataService.java
```diff
@@ -0,0 +1,412 @@
...
+        insertQueryBuilder.append(" VALUES ");
+
+        StringBuilder valuesMasterBuilder = new StringBuilder();
+
+        int iterator = 0;
+        for (JsonNode item : items) {
+
+            // In the number of values inserted is greater than 1000, the insert would fail. Once we have reached 1000
+            // rows, execute the insert for rows so far and start afresh.
+            if (iterator == 1000) {
```
mohanarpit, line 233:
```suggestion
            if (iterator == 999) {
```

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/QueryDataService.java
```diff
@@ -0,0 +1,412 @@
...
+        StringBuilder columnNamesBuilder = new StringBuilder("(");
+        columnNamesBuilder.append(String.join(", ", columnNames));
+        columnNamesBuilder.append(")");
+
+        insertQueryBuilder.append(columnNamesBuilder);
+        insertQueryBuilder.append(" VALUES ");
+
+        StringBuilder valuesMasterBuilder = new StringBuilder();
+
+        int iterator = 0;
```
mohanarpit, line 228: Nit: Can we rename this variable to `count`? Iterator makes me believe that this is a Java iterator object.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/QueryDataService.java
```diff
@@ -0,0 +1,412 @@
...
+                    valuesBuilder.append("'");
+                    valuesBuilder.append(fieldNode.asText());
+                    valuesBuilder.append("'");
+                }
+                valuesBuilder.append(",");
+            }
+
+            // All the columns have been read
+            // Trim the trailing comma
+            int i = valuesBuilder.lastIndexOf(",");
```
mohanarpit, line 256: Suggestion to refactor the call to `lastIndexOf` function. We can make a decision based on a boolean value earlier in the lifecycle.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/QueryDataService.java
```diff
@@ -0,0 +1,412 @@
...
+            e.printStackTrace();
+            throw new AppsmithPluginException(AppsmithPluginError.PLUGIN_ERROR, "Filtering failure seen during insertion of data : " + e.getMessage());
+        }
+    }
+
+    private static String finalInsertQueryString(String partialInsertQuery, StringBuilder valuesBuilder) {
+
+        StringBuilder insertQueryBuilder = new StringBuilder(partialInsertQuery);
+
+        if (valuesBuilder.lastIndexOf(",") == valuesBuilder.length() - 1) {
```
mohanarpit, line 286: Suggestion to remove invocations to `lastIndexOf`. This call makes me believe that we are adding a trailing comma only to be removed later. We can make this cleaner by avoiding a trailing comma.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/QueryDataService.java
```diff
@@ -0,0 +1,412 @@
...
+        String dropTableQuery = "DROP TABLE " + tableName + ";";
+
+        executeDbQuery(dropTableQuery);
+    }
+
+
+    private static Map<String, DataType> generateSchema(JsonNode jsonNode) {
+
+        Iterator<String> fieldNamesIterator = jsonNode.fieldNames();
+        /**
```
mohanarpit, line 374:
```suggestion
        /*
```

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/QueryDataService.java
```diff
@@ -0,0 +1,419 @@
...
+        insertQueryBuilder.append(" VALUES ");
+
+        StringBuilder valuesMasterBuilder = new StringBuilder();
+
+        int counter = 0;
+        for (JsonNode item : items) {
+
+            // If the number of values inserted is greater than 1000, the insert would fail. Once we have reached 1000
+            // rows, execute the insert for rows so far and start afresh for the rest of the rows
+            if (counter == 999) {
```
mohanarpit, line 234: Sorry, my bad. This should have been 1000 for a batch size of 1000.

mohanarpit:
```suggestion
            if (counter == 1000) {
```

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/QueryDataService.java
```diff
@@ -0,0 +1,419 @@
...
+            ConditionalOperator.NOT_EQ, "<>",
+            ConditionalOperator.GT, ">",
+            ConditionalOperator.GTE, ">=",
+            ConditionalOperator.IN, "IN",
+            ConditionalOperator.NOT_IN, "NOT IN"
+    );
+
+    private static Connection connection;
+
+    public QueryDataService() {
```
mohanarpit, line 64: Since all the functions in the class are static, it may be a good reason to refactor this class into a Singleton class. This way we can reuse the Connection object while maintaining object function invocations across the code base.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/QueryDataService.java
```diff
@@ -0,0 +1,430 @@
...
+            ConditionalOperator.LTE, "<=",
+            ConditionalOperator.EQ, "=",
+            ConditionalOperator.NOT_EQ, "<>",
+            ConditionalOperator.GT, ">",
+            ConditionalOperator.GTE, ">=",
+            ConditionalOperator.IN, "IN",
+            ConditionalOperator.NOT_IN, "NOT IN"
+    );
+
+    private static Connection connection;
```
mohanarpit, line 64: Since this is a singleton class, this doesn't need to be static anymore. It can be a normal class level variable.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/QueryDataService.java
```diff
@@ -0,0 +1,430 @@
...
+
+        if (valuesMasterBuilder.length() > 0) {
+            String insertQueryString = finalInsertQueryString(insertQueryBuilder.toString(), valuesMasterBuilder);
+            executeDbQuery(insertQueryString);
+        }
+    }
+
+    private void executeDbQuery(String query) {
+
+        Connection connection = checkAndGetConnection();
```
mohanarpit, line 298: Nit: Suggest renaming the variable to `conn` so that it doesn't conflict with the class level variable.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/QueryDataService.java
```diff
@@ -0,0 +1,430 @@
...
+import java.util.stream.Stream;
+
+import static com.appsmith.external.helpers.DataTypeStringUtils.stringToKnownDataTypeConverter;
+import static com.appsmith.external.models.Condition.addValueDataType;
+
+@Component
+@Slf4j
+public class QueryDataService {
+
+    private static QueryDataService single_instance = null;
```
mohanarpit, line 38:
```suggestion
    private static QueryDataService instance = null;
```

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/QueryDataService.java
```diff
@@ -0,0 +1,430 @@
...
+
+    private static String JDBC_DRIVER = "org.h2.Driver";
+    private static String url = "jdbc:h2:mem:filterDb";
+    private static ObjectMapper objectMapper = new ObjectMapper();
```
mohanarpit, line 42: Instead of a static class variable, suggest initializing this inside the private constructor object.

mohanarpit (approved):

PR #7602, opened 2021-09-20
fix: Adding handling for columns with multiple words in name for filtering data in filtering library

### LINKED ISSUE #7599: [Bug] Google Sheets where clause does not support space separated column name (aka multiple words)
If the column name comprises of multiple words, the insertion of data into H2 database fails leading to errors.

trishaanand:
Fixes #7599

### DIFF
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/FilterDataService.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/FilterDataService.java
@@ -412,12 +412,18 @@ public class FilterDataService {
                 .takeWhile(x -> fieldNamesIterator.hasNext())
                 .map(n -> fieldNamesIterator.next())
                 .collect(Collectors.toMap(
-                        Function.identity(),
-                        name -> {
-                            String value = jsonNode.get(name).asText();
-                            DataType dataType = stringToKnownDataTypeConverter(value);
-                            return dataType;
-                        }));
+                            Function.identity(),
+                            name -> {
+                                String value = jsonNode.get(name).asText();
+                                DataType dataType = stringToKnownDataTypeConverter(value);
+                                return dataType;
+                            },
+                            (u, v) -> {
+                                throw new IllegalStateException(String.format("Duplicate key %s", u));
+                            },
+                            LinkedHashMap::new
+                        )
+                );
 
         return schema;
     }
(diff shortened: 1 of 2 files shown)


app/server/appsmith-interfaces/src/test/java/com/appsmith/external/services/FilterDataServiceTest.java
```diff
@@ -274,6 +274,53 @@ public void testFilterNotInConditionForNumbers() {
...
+            ArrayNode items = (ArrayNode) objectMapper.readTree(data);
+
+            List<Condition> conditionList = new ArrayList<>();
+
+            Condition condition = new Condition("orderAmount", "LT", "15");
+            conditionList.add(condition);
+
+            ArrayNode filteredData = filterDataService.filterData(items, conditionList);
+
+            assertEquals(filteredData.size(), 2);
```
nidhi-nair, line 321: Minor could you also add a check for the upper case conversion here?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/FilterDataService.java
```diff
@@ -298,6 +300,7 @@ private void executeDbQuery(String query) {
 
         Connection conn = checkAndGetConnection();
 
+        log.debug("{}", query);
```
nidhi-nair, line 303: Is this required?

trishaanand: Issh. Sorry. Removing this.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/FilterDataService.java
```diff
@@ -180,7 +180,7 @@ private String generateWhereClause(List<Condition> conditions) {
                         operator.toString() + " is not supported currently for filtering.");
             }
 
-            sb.append(path);
+            sb.append("\"" + path + "\"");
```
nidhi-nair, line 183: QQ: How would this work for anything with quotes in them?

trishaanand: No good solution that I can come up with right now. Throwing an exception if the column name contains quotes. Handling values via Prepared Statement for which there is a raised issue.

PR #7656, opened 2021-09-21
fix: Added a condition to json stringify values if they are js objects

### LINKED ISSUE #7539: [Bug] Mongo Evaluated Value does not match the actual value sent
## Description

The evaluated value in mongo does not match the actual data that is being sent.
- There are no quotes around the values
- The values are not in a JSON format
Screenshot 2021-09-16 at 5 46 11 PM

ayushpahwa:
## Description

When showing evaluated values for substitutionType `TEMPLATE` the evaluated value popup shows JSON which is not valid. If the type is changed to `SMART_SUBSTITUTE`, the evaluated values are shown correctly. This is because when replacing the values in the final step, proper JSON stringify was not happening. Fixed the same with this PR

Fixes #7539 

## Type of change

## Checklist:


app/client/src/workers/evaluationSubstitution.ts
```diff
@@ -113,6 +113,10 @@ export const templateSubstituteDynamicValues = (
     } catch (e) {
       // do nothing
     }
+
+    // When replacing the binding with the evaluated value, we need to stringify the value
+    if (b.includes("{{")) value = JSON.stringify(value);
```
hetunandu, line 118: Add a test case for this

hetunandu (changes requested):

Nikhil-Nandagopal:
@ayushpahwa can we get this merged?

app/client/src/workers/evaluationSubstitution.ts
ayushpahwa: Currently stuck in the new where clause implementation for UQI. Have raised a new [issue](8243) and assigned it to me for the same. Will work on it after the current tasks.

Nikhil-Nandagopal (dismissed):

PR #7707, opened 2021-09-22
ci: Use separate artifacts for separate services

sharat87:
Currently, all three services' artifact names are set to `build`. This leads to a single artifact with files from all three services as siblings. This has two problems:

1. If there's an overlap in a filename among the files in build folder of two services, one of them will be overwritten with the other.
2. The same large artifact is downloaded every time we do a `download-artifact`, once for server, once for client, and once for RTS, the same large artifact is downloaded thrice before building the Docker images.

Here's a preview of how the artifact looks like currently. The marked ones are from server and the rest are from client (and RTS as well I think).

Screenshot 2021-09-22 at 10 59 16

### DIFF
--- a/.github/workflows/test-build-docker-image.yml
+++ b/.github/workflows/test-build-docker-image.yml
@@ -534,7 +534,7 @@ jobs:
           name: server-build
           path: app/server/dist
 
-      - name: Download the server build artifact
+      - name: Download the rts build artifact
         uses: actions/download-artifact@v2
         with:
           name: rts-build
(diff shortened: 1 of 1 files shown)


.github/workflows/test-build-docker-image.yml
```diff
@@ -524,19 +525,19 @@ jobs:
...
           path: app/client/build
 
       - name: Download the server build artifact
         uses: actions/download-artifact@v2
         with:
-          name: build
+          name: server-build
           path: app/server/dist
 
       - name: Download the server build artifact
```
mohanarpit, line 537: Minor: Can we correct the name of the step as well?

sharat87: Oh nice catch. Fixed.

mohanarpit (approved):

PR #7735, opened 2021-09-22
feat: Announce that Appsmith is running!

sharat87:
Since the file diff on GitHub doesn't do justice to what this PR actually is about, here's a picture!

Screenshot 2021-09-22 at 18 22 08

This is printed after the server is up.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/InstanceConfig.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/InstanceConfig.java
@@ -35,6 +35,8 @@ public class InstanceConfig implements ApplicationListener<ApplicationReadyEvent
         configService.getByName(Appsmith.APPSMITH_REGISTERED)
                 .filter(config -> Boolean.TRUE.equals(config.getConfig().get("value")))
                 .switchIfEmpty(registerInstance())
+                .doOnSuccess(ignored -> this.printReady())
+                .doOnError(ignored -> this.printReady())
                 .subscribe(null, e -> {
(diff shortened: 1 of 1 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/InstanceConfig.java
```diff
@@ -73,4 +75,17 @@ public void onApplicationEvent(ApplicationReadyEvent applicationReadyEvent) {
                         .save(Appsmith.APPSMITH_REGISTERED, Map.of("value", true))
                 );
     }
+
+    private void printReady() {
+        System.out.println(
```
mohanarpit, line 80: Will log4j be available at this juncture? Can we use `log.info` instead?

sharat87:
@mohanarpit, _just_ added a desc, sorry.

Primarily targeted towards folks who'd be running `docker run` with the fat container, to give them a hint that Appsmith is now running, and that they can hit the port they used in the command.

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/InstanceConfig.java
sharat87: It is. I actually started out with `log.info` in fact. But changed to this at the last moment. I didn't want this to be tied to log level configuration. If the level was configured to warning, this message won't come out at all. But then, I don't want to use `log.error` either, so picked this as a compromise. What say?

mohanarpit (approved):

PR #7753, opened 2021-09-23
ci: Build and push fat container in release workflow

sharat87:
The GitHub release workflow currently doesn't build the fat container image. This PR adds support for this.

### DIFF
--- a/.github/workflows/github-release.yml
+++ b/.github/workflows/github-release.yml
@@ -40,15 +40,11 @@ jobs:
             echo "::set-output name=is_beta::false"
           fi
 
-  build-client:
+  build:
     needs:
       - prelude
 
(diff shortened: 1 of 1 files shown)


.github/workflows/github-release.yml
```diff
@@ -40,16 +40,12 @@ jobs:
             echo "::set-output name=is_beta::false"
           fi
 
-  build-client:
+  build:
```
mohanarpit, line 43: The build job is building in serial. Is it possible to build all services parallely?

sharat87: Sure, that would mean uploading and downloading the artifacts though, which basically means copy pasting the other workflow. :)

On it.

mohanarpit (approved):

PR #7762, opened 2021-09-23
fix: Adding Prepared Statement setting to the main query pane for Postgres, MySQL and MS SQL plugins

### LINKED ISSUE #6235: [Feature]Send a hint to turn off Prepared Statement if the error says syntax error at or near $x
## Summary

More often than not, users are not used to Prepared Statements semantics and don't understand why their queries are failing. Hint messages would be a solution in the short term to get them unblocked by turning off prepared statement

trishaanand:
Fixes #6235


app/server/appsmith-plugins/mssqlPlugin/src/main/resources/editor.json
```diff
@@ -27,6 +27,13 @@
             "comparison": "EQUALS",
             "value": true
           }
+        },
+        {
+          "label": "Use Prepared Statement",
+          "info": "Turning on Prepared Statement makes your queries resilient against bad things like SQL injections. This feature can be used when dynamic bindings replace values and not when they contain SQL keywords like 'SELECT', 'WHERE', 'AND', etc.",
+          "configProperty": "actionConfiguration.pluginSpecifiedTemplates[0].value",
```
sumitsum, line 34: This line is a bit confusing: `This feature can be used when dynamic bindings replace values and not when they contain SQL keywords like 'SELECT', 'WHERE', 'AND', etc.`. I had to re-read it a few times to understand it. Can we make it simpler ?

trishaanand: I tried to make the current language simpler. Do you have suggestions? Should I break it into two sentences (aka post the and for "not when they contain.." . What else could I do to impart this knowledge about PS in a manner that is easy to grok and understand?

sumitsum:
Not sure about this but this could be a small issue: Since the toggle exists on both the editor and settings page, the two toggles may be in conflicting states - ie. one `off` and another one `on`until the user refreshes the page. Should another issue be created to address this problem in future ?

trishaanand:
> Not sure about this but this could be a small issue: Since the toggle exists on both the editor and settings page, the two toggles may be in conflicting states - ie. one `off` and another one `on`until the user refreshes the page. Should another issue be created to address this problem in future ?

@sumitsum : I tested for this and the toggles were always in sync because the underlying data has a single copy. Are you seeing a different behaviour?

sumitsum:
> > Not sure about this but this could be a small issue: Since the toggle exists on both the editor and settings page, the two toggles may be in conflicting states - ie. one `off` and another one `on`until the user refreshes the page. Should another issue be created to address this problem in future ?
> 
> @sumitsum : I tested for this and the toggles were always in sync because the underlying data is a single copy. Are you seeing a different behaviour?

@trishaanand seems fine then. I didn't know it uses the same data object underneath. I have not tested it. I was just imagining the flow in my mind.

sumitsum (approved):

PR #7767, opened 2021-09-23
fix: Fixed installation for multiple plugins into organizations

nidhi-nair:
The default installs were failing for existing organizations because the findAll function was expecting a user context. This fix uses a query without such an expectation.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/CustomOrganizationRepositoryImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/CustomOrganizationRepositoryImpl.java
@@ -1,7 +1,6 @@
 package com.appsmith.server.repositories;
 
 import com.appsmith.server.acl.AclPermission;
-import com.appsmith.server.domains.Comment;
 import com.appsmith.server.domains.Organization;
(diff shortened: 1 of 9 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/CustomOrganizationRepositoryImpl.java
```diff
@@ -78,4 +77,9 @@ public CustomOrganizationRepositoryImpl(ReactiveMongoOperations mongoOperations,
                 )
                 .then();
     }
+
+    @Override
+    public Flux<Organization> findAllOrganizations() {
```
trishaanand, line 82: Wouldn't adding the function declaration `Flux<Organization> findAll()` in the file `OrganizationRepository` work out of the box?

nidhi-nair: You mean override that method to not use user context? Yeah but I figured I shouldn't mess with the expectation from `findAll` from all domain repositories, it might stump some developer in the future.
I hope that's what you're asking about!

trishaanand (approved):

PR #7769, opened 2021-09-23
fix: copy snippet

arunvjn:
## Description
Fixes copy snippet in customize tab and display toast messages based on the following criteria.
Returns success If results/triggers exists.
Returns error if errors exists.
Return success if result is undefined. (Some snippets evaluates to undefined)

## Type of change

## Checklist:

    // Code coverage diff between base branch:release and head branch: fix/copy_snippet 
 :red_circle: | total | 54.88 **(-0.01)** | 37.03 **(-0.02)** | 33.89 **(0)** | 55.49 **(-0.01)**
 :red_circle: | app/client/src/components/editorComponents/GlobalSearch/index.tsx | 52.48 **(-0.52)** | 14.4 **(-0.34)** | 37.31 **(0)** | 54.04 **(-0.61)**
 :red_circle: | app/client/src/constants/messages.ts | 74.52 **(-0.02)** | 100 **(0)** | 24.01 **(-0.06)** | 80.9 **(0.03)**
 :red_circle: | app/client/src/sagas/EvaluationsSaga.ts | 24.02 **(-0.36)** | 5.88 **(-0.37)** | 13.04 **(-0.6)** | 27.01 **(-0.64)**
 :red_circle: | app/client/src/utils/autocomplete/TernServer.ts | 50.47 **(-0.24)** | 38.26 **(-0.87)** | 36.21 **(0)** | 55.08 **(-0.27)**</details>


app/client/src/sagas/EvaluationsSaga.ts
```diff
@@ -400,23 +400,33 @@ export function* evaluateSnippetSaga(action: any) {
...
+      Following code handles that. Also, we set the results to "undefined" when the evaluated value is undefined.
+      */
       yield put(
         setEvaluatedSnippet(
           result
             ? JSON.stringify(result, null, 2)
             : errors && errors.length
             ? JSON.stringify(errors, null, 2)
-            : "",
+            : "undefined",
```
ApekshaBhosale, line 412: why is this string? why not just undefiend?

ApekshaBhosale (changes requested):

app/client/src/sagas/EvaluationsSaga.ts
arunvjn: We want the text "undefined" to show up in evaluated snippet when the evaluated snippet resolves to undefined.

app/client/src/sagas/EvaluationsSaga.ts
```diff
@@ -400,23 +400,33 @@ export function* evaluateSnippetSaga(action: any) {
...
         ),
       );
     }
     Toaster.show({
       text: createMessage(
-        result || triggers
+        result || triggers?.length > 0
```
rishabhsaxena, line 418: QQ: Just to get a bit more clarity, would it also mean the same if we just check for `results`, as in if it's possible we have `triggers` but the result is falsy?

app/client/src/sagas/EvaluationsSaga.ts
```diff
@@ -400,23 +400,33 @@ export function* evaluateSnippetSaga(action: any) {
         {},
```
rishabhsaxena, line 400: Noticed we just execute the first `trigger`, should we loop through in case we have multiple triggers?
Actually that's not a part of this PR, came across it,  (line: 398)

app/client/src/components/editorComponents/GlobalSearch/index.tsx
```diff
@@ -484,6 +484,12 @@ function GlobalSearch() {
 
   const handleSnippetClick = (event: SelectEvent, item: any) => {
     if (event && event.type === "click") return;
+    const snippetExecuteBtn = document.querySelector(
+      ".snippet-execute",
+    ) as HTMLButtonElement;
+    if (snippetExecuteBtn && !snippetExecuteBtn.disabled) {
+      return snippetExecuteBtn && snippetExecuteBtn.click();
+    }
```
rishabhsaxena, line 492: nit: might be more readable if we pass the function instead of using the btn ref

rishabhsaxena (approved):

rishabhsaxena:
We should also consider adding few cypress tests for the snippets feature

app/client/src/sagas/EvaluationsSaga.ts
arunvjn: Yes, it is possible to have triggers when the result is falsy

app/client/src/sagas/EvaluationsSaga.ts
arunvjn: Will make this change

app/client/src/components/editorComponents/GlobalSearch/index.tsx
rishabhsaxena: As an alternative if we'd not like to refactor this currently, can we try passing a ref to the snippets item, and attach the `handleSnippets` function to it?
Also should create an issue to refactor if we're not updating it in this PR

app/client/src/sagas/EvaluationsSaga.ts
```diff
@@ -382,41 +383,49 @@ export function* evaluateSnippetSaga(action: any) {
...
         setEvaluatedSnippet(
-          result
-            ? JSON.stringify(result, null, 2)
-            : errors && errors.length
+          errors?.length
             ? JSON.stringify(errors, null, 2)
-            : "",
+            : isUndefined(result)
+            ? "undefined"
+            : JSON.stringify(result),
```
rishabhsaxena, line 420: should we handle the case where the result is `undefined` where we are consuming the snippets instead of setting it to an `"undefined"` string here?

rishabhsaxena (approved):

PR #7774, opened 2021-09-23
ci: Push latest tag for RTS and fat

sharat87:


.github/workflows/github-release.yml
```diff
@@ -250,11 +250,23 @@ jobs:
...
       - name: Build and push fat image
         run: |
           docker build \
             --build-arg APPSMITH_SEGMENT_CE_KEY=${{ secrets.APPSMITH_SEGMENT_CE_KEY }} \
             --tag ${{ secrets.DOCKER_HUB_ORGANIZATION }}/appsmith-ce:${{needs.prelude.outputs.tag}} \
             .
+
+          # Only build & tag with latest if the tag doesn't contain beta
+          if [[ ! ${{needs.prelude.outputs.tag}} == *"beta"* ]]; then
+            docker build --tag ${{ secrets.DOCKER_HUB_ORGANIZATION }}/appsmith-ce:latest .
```
mohanarpit, line 268: I think we should push the docker image here as well. I think there was an issue with pushing `latest` tag when we try to push all the images. Hence, we added the step to push the latest tag separately.

sharat87: I _think_ the issue was a little different back then, but I'm in no mood to find out, so, fixed. 😄

mohanarpit (approved):

PR #7816, opened 2021-09-24
feat: Skip bot-comment when onboarding is skipped

### LINKED ISSUE #7281: [Bug]Skip bot comments if the comments tour is skipped
## Description
Skip bot comments if the comments tour is skipped

1. Signup with a new user
2. Click on toggle comment mode button
3. Click on the skip button, enter the name and submit the comments onboarding form
4. Add a new comment, a bot comment is added, if the tour is skipped the bot comment should also be skipped

nayan-rafiq:
## Description
Skips the bot comments if user has skipped the on-boarding flow during their first comment.

Fixes #7281

## Type of change

## How Has This Been Tested?
- Tested on local machine
- Unit tests

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/constants/CommentState.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/constants/CommentState.java
@@ -1,0 +1,8 @@
+package com.appsmith.server.constants;
+
+public enum CommentState {
+    SKIPPED, // skipped the on-boarding tutorial
+    ONBOARDED, // finished the on-boarding tutorial
+    COMMENTED, // added a comment
+    RESOLVED // resolved a comment
+}
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/UserController.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/UserController.java
@@ -146,8 +146,7 @@ public class UserController extends BaseController<UserService, User, String> {
     @GetMapping("/me")
     public Mono<ResponseDTO<UserProfileDTO>> getUserProfile() {
         return sessionUserService.getCurrentUser()
-                .zipWith(userDataService.getForCurrentUser())
-                .flatMap(objects -> service.buildUserProfileDTO(objects.getT1(), objects.getT2()))
+                .flatMap(service::buildUserProfileDTO)
                 .map(profile -> new ResponseDTO<>(HttpStatus.OK.value(), profile, null));
     }
 
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/UserController.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/UserController.java
@@ -146,7 +146,8 @@ public class UserController extends BaseController<UserService, User, String> {
     @GetMapping("/me")
     public Mono<ResponseDTO<UserProfileDTO>> getUserProfile() {
         return sessionUserService.getCurrentUser()
-                .flatMap(service::buildUserProfileDTO)
+                .zipWith(userDataService.getForCurrentUser())
+                .flatMap(objects -> service.buildUserProfileDTO(objects.getT1(), objects.getT2()))
                 .map(profile -> new ResponseDTO<>(HttpStatus.OK.value(), profile, null));
     }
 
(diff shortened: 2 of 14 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/UserController.java
```diff
@@ -206,4 +208,20 @@ public UserController(UserService service,
...
+    @PatchMapping("/comment/state")
+    public Mono<ResponseDTO<CommentState>> setCommentState(@RequestBody UserData userData) {
+        return userDataService.setCommentState(userData.getCommentState())
+                .map(savedUserData ->
+                        new ResponseDTO<>(HttpStatus.OK.value(), savedUserData.getCommentState(), null)
+                );
+    }
+
+    @GetMapping("/comment/state")
+    public Mono<ResponseDTO<CommentState>> getCommentState() {
```
sharat87, line 220: Is the client going to call this API on every page load? Can we put this information as part of user profile itself? That is anyway being fetched, so we save on an extra API call no?

Also, if we do that, then the client will need to make sure this field is **not** passed on to `appsmith.user` object, or folks will be able to see this in auto-complete when writing JS in Appsmith.

app/server/appsmith-server/src/main/java/com/appsmith/server/constants/CommentState.java
```diff
@@ -0,0 +1,8 @@
+package com.appsmith.server.constants;
+
+public enum CommentState {
```
sharat87, line 3: If this is going to be _only_ about onboarding, can we name it `CommentOnboardingState`? Right now this name feels like the _state of a comment_, with options like `RESOLVED` meaning that the comment is _in a resolved state_, etc.

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/UserController.java
nayan-rafiq: @sharat87 I thought about that but currently, we don't return any properties of `UserData` to any API e.g. `/me`. If I return `userData` contents as part of the `/me` API, other properties of the `UserData` will be exposed and many of them are for internal use. That's why I came up with a new API. After reading your comment, I think I can expose this property only in `me` API. Do you think that'll be a better idea? 

To avoid an API call each time, client will cache this value and only make an API call when cache is empty. Client will use this value to decide whether to show the on-boarding flow to user or not.

nayan-rafiq: @sharat87 I've updated the PR. Now the `commentOnboardingState` is returned in the response of `/me` API.

sharat87 (approved):
Looks good, so approving. However, note that once this goes in, the `appsmith.user` object will have a `commentOnboardingState` field for all users to see and use. The client code needs to not expose that field to user-written Javascript. That change can go along with this PR togather, or soon after, but needs to go soon.

sharat87 (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/UserController.java
```diff
@@ -144,7 +146,8 @@ public UserController(UserService service,
     @GetMapping("/me")
     public Mono<ResponseDTO<UserProfileDTO>> getUserProfile() {
         return sessionUserService.getCurrentUser()
-                .flatMap(service::buildUserProfileDTO)
+                .zipWith(userDataService.getForCurrentUser())
```
mohanarpit, line 149: Any particular reason that we are doing this logic in the controller & not the service method itself?

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/UserController.java
```diff
@@ -206,4 +209,12 @@ public UserController(UserService service,
                 .map(map -> new ResponseDTO<>(HttpStatus.OK.value(), map, null));
     }
 
+    @PatchMapping("/comment/state")
+    public Mono<ResponseDTO<CommentOnboardingState>> setCommentState(@RequestBody UserData userData) {
+        return userDataService.setCommentState(userData.getCommentOnboardingState())
```
mohanarpit, line 214: Quick question: Do these DB calls not require any ACL? Just checking.

nayan-rafiq: We don't set policies to `UserData` and there is no ACL check anywhere for `UserData`. Should we change that?

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/UserController.java
mohanarpit: Yes please. If we need to refactor, lets do that. Adding such logic in a controller makes it virtually untestable since we don't have any controller level tests in the code base.

sharat87 (changes requested):
Looks good. It looks like all tests have a `@WithUserDetails(value = "api_user")` above them. Considering the error we saw recently, can we add a test where no one is logged in, and ensure that we get a profile object, instead of an empty Mono?

nayan-rafiq:
> Looks good. It looks like all tests have a `@WithUserDetails(value = "api_user")` above them. Considering the error we saw recently, can we add a test where no one is logged in, and ensure that we get a profile object, instead of an empty Mono?

@sharat87 I've added a unit test to test profile is returned when user is anonymous.

sharat87 (approved):
Looks good to be merged. Please:

1. Edit the merge commit's detail to the PR description, or otherwise more details about the change.
1. Squash merge once the CI is finished. Thanks!

PR #7829, opened 2021-09-25
fix: on page load actions take into account widget dependencies as well to compute correct order of action computaion on page load

### LINKED ISSUE #7065: [Bug] Default value from select not picked up in query on page load.
## Description
There is a select widget, whose options depend on a query and it has a default value.
Value of this select is used in another query, which in turn is used to populate another select.

The second query is not getting the default value from the first select widget.

DataTree has the correct value for the first drop down.

Please check the video.

https://www.loom.com/share/dc4b047b21c44e8d89d38ec28b5628f2

### Important Details

- Version: [Cloud]
- OS: [e.g.MacOSX]
- Browser: chrome
- Environment:production

trishaanand:
Fixes #7065


mohanarpit:
This is a good change. Refactoring the extraction logic makes this flow more extensible for future scenarios. 🎉

PR #7880, opened 2021-09-28
fix: Actions not being deleted when app is deleted

### LINKED ISSUE #4609: [Bug] Unable to delete datasource even if all the applications in organisation are deleted
## Description

After deleting all the application in an organisation it's expected that datasources should not be attached with any action, but trying to delete datasources throws error : _(Cannot delete datasource since it has 1 action(s) using it.)_ 

1. Create new org
2. Create new application
3. Create DS (Created DB query using mongoDB as datasource)
4. Deleted the application so now there isn’t any application present
5. Create new application without any widget and actions
6. Try to delete the datasource

sharat87:
Fixes #4609.

Testing:

1. Create new organization, called `org1`.
2. Create an application in this, called `app1`.
3. Create a datasource called `ds1`, and an action for that datasource, in `app1`.
4. Now create another application in `org1`, called `app2`.
5. Now delete the application `app1`.
6. Go to `app2`, try to delete the datasource `ds1`.

- Before this PR: The deletion is not allowed because there's one action currently using it.
- After this PR: The deletion is successful because that other action is deleted when `app1` is deleted.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
@@ -3265,11 +3265,8 @@ public class DatabaseChangelog {
 
         for (final NewAction action : actions) {
             final String applicationId = action.getApplicationId();
-            if (StringUtils.isEmpty(applicationId)) {
-                continue;
-            }
 
-            final boolean isAppDeleted = mongockTemplate.exists(
+            final boolean shouldDelete = StringUtils.isEmpty(applicationId) || mongockTemplate.exists(
                     query(
                             where(fieldName(QApplication.application.id)).is(applicationId)
                                     .and(fieldName(QApplication.application.deleted)).is(true)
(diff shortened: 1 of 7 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
```diff
@@ -3251,4 +3251,40 @@ public void updatePluginPackageNameIndexToPluginNamePackageNameAndVersion(Mongoc
...
+
+        final List<NewAction> actions = mongockTemplate.find(
+                query(where(fieldName(QNewAction.newAction.deleted)).ne(true)),
+                NewAction.class
+        );
+
+        for (final NewAction action : actions) {
+            final String applicationId = action.getApplicationId();
+            if (StringUtils.isEmpty(applicationId)) {
+                continue;
```
trishaanand, line 3269: Why are we skipping the actions without an application id? Shouldn't these actions be orphan too?

trishaanand (approved):
LGTM. Minor question added.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
sharat87: Yep. Thanks for pointing out. Fixed and pushed 😁

trishaanand (approved):

PR #7892, opened 2021-09-28
fix: Changes for JS actions in the import/export feature

### LINKED ISSUE #7008: [Bug] Implement import/export for js collections

nidhi-nair:
Fixes #7008

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ImportExportApplicationService.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ImportExportApplicationService.java
@@ -6,28 +6,31 @@ import com.appsmith.external.models.AuthenticationResponse;
 import com.appsmith.external.models.BaseDomain;
 import com.appsmith.external.models.BasicAuth;
 import com.appsmith.external.models.DBAuth;
+import com.appsmith.external.models.Datasource;
 import com.appsmith.external.models.DatasourceConfiguration;
 import com.appsmith.external.models.DecryptedSensitiveFields;
 import com.appsmith.external.models.OAuth2;
 import com.appsmith.server.acl.AclPermission;
 import com.appsmith.server.constants.FieldName;
 import com.appsmith.server.constants.SerialiseApplicationObjective;
+import com.appsmith.server.domains.ActionCollection;
 import com.appsmith.server.domains.Application;
 import com.appsmith.server.domains.ApplicationJson;
 import com.appsmith.server.domains.ApplicationPage;
-import com.appsmith.external.models.Datasource;
 import com.appsmith.server.domains.NewAction;
 import com.appsmith.server.domains.NewPage;
-import com.appsmith.server.domains.PluginType;
 import com.appsmith.server.domains.User;
+import com.appsmith.server.dtos.ActionCollectionDTO;
 import com.appsmith.server.dtos.ActionDTO;
 import com.appsmith.server.dtos.PageDTO;
 import com.appsmith.server.exceptions.AppsmithError;
 import com.appsmith.server.exceptions.AppsmithException;
+import com.appsmith.server.repositories.ActionCollectionRepository;
 import com.appsmith.server.repositories.DatasourceRepository;
 import com.appsmith.server.repositories.NewActionRepository;
 import com.appsmith.server.repositories.NewPageRepository;
 import com.appsmith.server.repositories.PluginRepository;
+import com.appsmith.server.services.ActionCollectionService;
 import com.appsmith.server.services.ApplicationPageService;
 import com.appsmith.server.services.ApplicationService;
 import com.appsmith.server.services.DatasourceService;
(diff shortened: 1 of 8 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ImportExportApplicationService.java
```diff
@@ -97,187 +103,212 @@
...
+                    application.setOrganizationId(null);
+                    application.setPages(null);
+                    application.setPublishedPages(null);
+                    application.setModifiedBy(null);
+                    application.setUpdatedAt(null);
+                    application.setLastDeployedAt(null);
+                    application.setGitApplicationMetadata(null);
+                    examplesOrganizationCloner.makePristine(application);
+                    applicationJson.setExportedApplication(application);
+                    // TODO @Abhijeet please take a look at the ACL permissions
```
trishaanand, line 165: @nidhi-nair : Does MANAGE_PAGES does not look like the correct permission to use here?

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ImportExportApplicationService.java
```diff
@@ -97,187 +103,212 @@
...
-                    })
-                    .collectList()
-                    .map(newActionList -> {
-                        Set<String> concernedDBNames = new HashSet<>();
-                        newActionList.forEach(newAction -> {
-                            newAction.setPluginId(pluginMap.get(newAction.getPluginId()));
-                            newAction.setOrganizationId(null);
-                            newAction.setPolicies(null);
-                            newAction.setApplicationId(null);
-                            //Collect Datasource names to filter only required datasources
```
trishaanand, line 233: @abhvsn : QQ : What was the reason this check was added?

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ImportExportApplicationService.java
```diff
@@ -97,187 +103,212 @@
...
+                                    if (newAction.getUnpublishedAction() != null) {
+                                        ActionDTO actionDTO = newAction.getUnpublishedAction();
+                                        actionDTO.setPageId(pageIdToNameMap.get(actionDTO.getPageId() + PublishType.UNPUBLISHED));
+                                    }
+                                    if (newAction.getPublishedAction() != null) {
+                                        ActionDTO actionDTO = newAction.getPublishedAction();
+                                        actionDTO.setPageId(pageIdToNameMap.get(actionDTO.getPageId() + PublishType.PUBLISHED));
+                                    }
+                                });
+                                // This is where we're removing global datasources that are unused in this application
```
trishaanand, line 257: Thanks for writing the comment @nidhi-nair ! ❤️

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ImportExportApplicationService.java
```diff
@@ -378,275 +418,358 @@
...
+                                        .flatMap(actionId -> newActionRepository.findById(actionId, AclPermission.MANAGE_ACTIONS))
+                                        .map(newAction -> {
+                                            newAction.getUnpublishedAction().setCollectionId(
+                                                    unpublishedActionIdToCollectionIdMap.getOrDefault(newAction.getId(), null));
+                                            newAction.getPublishedAction().setCollectionId(
+                                                    publishedActionIdToCollectionIdMap.getOrDefault(newAction.getId(), null));
+                                            return newAction;
+                                        })
+                                        .flatMap(newAction -> newActionService.update(newAction.getId(), newAction))
+                                        .then()
```
trishaanand, line 732: Any reason for adding a then() considering you are calling a thenReturn right after?

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ImportExportApplicationService.java
abhvsn: I was not sure if `PluginType.JS` will break the flow in some scenario that's why put that check here. But we can safely remove this

trishaanand: Thanks!

abhvsn (approved):
LGTM

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ImportExportApplicationService.java
nidhi-nair: Mysterious webflux behaviour, it was failing without it

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ImportExportApplicationService.java
nidhi-nair: We seem to have a separate role for `export`, but we're using different kinds of permissions throughout this flow. We should stick to one ACL type so that any changes in that graph in the future does not break this flow.

trishaanand: Role is used to programmatically give permissions (which are of types manage, read, execute, etc.) on top of the domain object. So the export role would give manage page permissions to the user for all the apps the user can export. Hence, this would be the correct permission to use.

trishaanand (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ImportExportApplicationService.java
nidhi-nair: Oh I see. We should map these expected roles visually some place. I thought we'd have export at various levels as well. But yeah, @abhvsn you can ignore my comment I'll delete it in some other PR :P

PR #7902, opened 2021-09-28
fix: Effect the telemetry choice in welcome page

sharat87:
Solves for two things:

1. From the welcome page signup, depending on telemetry being enabled or disabled, we report.
2. When telemetry is changed from welcome page (which, it obviously will), we make the change persistent by writing to the env file.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/AnalyticsService.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/AnalyticsService.java
@@ -11,7 +11,6 @@ import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.codec.digest.DigestUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
-import org.springframework.util.CollectionUtils;
 import reactor.core.publisher.Mono;
 import reactor.core.scheduler.Schedulers;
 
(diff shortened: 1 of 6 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/services/AnalyticsService.java
```diff
@@ -77,7 +81,8 @@ public void sendEvent(String event, String userId, Map<String, Object> propertie
         Map<String, Object> analyticsProperties = properties == null ? new HashMap<>() : new HashMap<>(properties);
 
         // Hash usernames at all places for self-hosted instance
-        if (!commonConfig.isCloudHosting()
+        if (hashUserId
```
abhvsn, line 84: Considering hashUserId can be false in some scenarios 
`(!CollectionUtils.isEmpty(analyticsProperties) && commonConfig.isCloudHosting())` check should be handled as for self hosted instances `commonConfig.isCloudHosting()` will be false which may result in having null values for `analyticsProperties`

abhvsn (changes requested):
Tiny suggestion 

Looks good to be merged after that

app/server/appsmith-server/src/main/java/com/appsmith/server/services/AnalyticsService.java
sharat87: That's actually a good catch. I'm wondering why we even have that check at all at this point. No matter what, we should never send `null` to Segment, or it'll throw up right away.

Just pushed a commit to always clean that map up before moving on.

abhvsn (dismissed):

app/server/appsmith-server/src/main/java/com/appsmith/server/services/AnalyticsService.java
```diff
@@ -94,12 +99,10 @@ public void sendEvent(String event, String userId, Map<String, Object> propertie
...
         }
 
-        if (!CollectionUtils.isEmpty(analyticsProperties) && commonConfig.isCloudHosting()) {
-            // Segment throws an NPE if any value in `properties` is null.
-            for (final Map.Entry<String, Object> entry : analyticsProperties.entrySet()) {
-                if (entry.getValue() == null) {
-                    analyticsProperties.put(entry.getKey(), "");
-                }
+        // Segment throws an NPE if any value in `properties` is null.
+        for (final Map.Entry<String, Object> entry : analyticsProperties.entrySet()) {
```
abhvsn, line 103: I was thinking instead of checking this here can we do it just after `Map<String, Object> analyticsProperties = properties == null ? new HashMap<>() : new HashMap<>(properties);` that way we can make sure this will be executed only once in any case

sharat87: Hey I didn't notice this was being done twice. Just fixed and pushed. :)

abhvsn (approved):
LGTM

PR #7918, opened 2021-09-29
fix: Signposting when user signs up with oauth is not working

### LINKED ISSUE #7417: [Bug] Signposting does not work when user signs up using OAuth
## Description
Google and GitHub sign up does not trigger Sign Posting feature

1. Sign up using Google/Github account and observe that sign posting flow is not triggered.

### Important Details
- Version: Cloud
- OS: Win10
- Browser: Chrome
- Environment: Release

nayan-rafiq:
> Pull Request Template
>

## Description
In the signposting feature, we create a default application when someone signs up. Currently, it works for form based signups only. This PR adds oauth for signposting also. 

Fixes #7417

## Type of change

## How Has This Been Tested?
- Manually

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/handlers/AuthenticationSuccessHandler.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/handlers/AuthenticationSuccessHandler.java
@@ -102,9 +102,9 @@ public class AuthenticationSuccessHandler implements ServerAuthenticationSuccess
             if(isFromSignup) {
                 boolean finalIsFromSignup = isFromSignup;
                 redirectionMono = createDefaultApplication(user)
-                        .flatMap(defaultApplication->handleOAuth2Redirect(webFilterExchange, finalIsFromSignup));
+                        .flatMap(defaultApplication->handleOAuth2Redirect(webFilterExchange, defaultApplication, finalIsFromSignup));
             } else {
-                redirectionMono = handleOAuth2Redirect(webFilterExchange, isFromSignup);
+                redirectionMono = handleOAuth2Redirect(webFilterExchange, null, isFromSignup);
             }
         } else {
             boolean finalIsFromSignup = isFromSignup;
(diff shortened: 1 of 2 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/handlers/AuthenticationSuccessHandler.java
```diff
@@ -154,8 +181,14 @@
...
 
+        boolean addFirstTimeExperienceParam = false;
         if (isFromSignup) {
-            redirectUrl = buildSignupSuccessUrl(redirectUrl, false);
+            if(redirectUrl.endsWith(RedirectHelper.DEFAULT_REDIRECT_URL) && defaultApplication != null) {
+                addFirstTimeExperienceParam = true;
+                HttpHeaders headers = exchange.getRequest().getHeaders();
+                redirectUrl = redirectHelper.buildApplicationUrl(defaultApplication, headers);
+            }
+            redirectUrl = buildSignupSuccessUrl(redirectUrl, addFirstTimeExperienceParam);
```
sharat87, line 191: Can you elaborate a little on this please? If the user is coming from signup (checked by the `if (isFromSignup)` line, shouldn't we always take them to the first time experience? Shouldn't `addFirstTimeExperienceParam` here be always `true`? Can you add a comment on when we want it to be `false`, even though they came from signup?

sharat87 (commented):
Looks good. Wanted a small comment on something.

app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/handlers/AuthenticationSuccessHandler.java
nayan-rafiq: @sharat87 As per discussion with @sbalaji1192 we only redirect to default application if the default redirect url is empty. For example, user is trying to access an application using an URL `/applications/abcd` and they are taken to signup page with url `/signup?redirectUrl=/applications/abc`. In this case, we don't want to take user to newly created application. That's why we are checking if user has any redirect url other than default url.

sharat87: I see. So what's the idea behind using `.endsWith` here? Looks like we want to check if the redirect URL is _equal_ to the default redirect URL, that's when we want to do this no?

nayan-rafiq: I've used `endsWith` so that `https://app.appsmith.com/applications` and `/applications` both are matched.

sharat87 (approved):
Looks good to be merged. Please:

1. Edit the merge commit's detail to the PR description, or otherwise more details about the change.
1. Squash merge once the CI is finished. Thanks!

PR #7935, opened 2021-09-29
Set selectedRow for multiple row selection in table widget

### LINKED ISSUE #7696: [Bug][Deploy Preview] In case of the multi-select mode, the last selected row is not updating on Table.selectedRow

1. Drag and drop a Table widget
2. Drag a Text widget
3. Bind text widget`{{Table1.selectedRow}}`
4. Now enable Multi select for table
5. Now select multiple options 
6. Observe that text widget do not get updated

### Important Details

- Version: [Cloud / Self-Hosted vx.x]
- OS: [e.g.MacOSX]
- Environment [production, release, deploy preview]

vicky-primathon:
## Description
Set selectedRow for multiple row selection in table widget

Fixes #7696 

## Type of change

## Checklist:

### DIFF
--- a/app/client/src/widgets/TableWidget/widget/derived.js
+++ b/app/client/src/widgets/TableWidget/widget/derived.js
@@ -1,10 +1,15 @@
 /* eslint-disable @typescript-eslint/no-unused-vars*/
 export default {
   getSelectedRow: (props, moment, _) => {
+    const selectedRowIndices = Array.isArray(props.selectedRowIndices)
+      ? props.selectedRowIndices
+      : [props.selectedRowIndices];
     const selectedRowIndex =
       props.selectedRowIndex === undefined ||
       Number.isNaN(parseInt(props.selectedRowIndex))
-        ? -1
+        ? selectedRowIndices.length
+          ? selectedRowIndices[selectedRowIndices.length - 1]
+          : -1
         : parseInt(props.selectedRowIndex);
     const filteredTableData =
       props.filteredTableData || props.sanitizedTableData || [];


riodeuno (changes requested):
@vicky-primathon Seems like this is only selecting the `selectedRowIndex`, but it is not populating `selectedRow`. 

We need to be careful of the following scenarios:
- What happens to the `selectedRow` when switching from single select to multi select?
- What happens to `selectedRow` when switching from multi select to single select?
- What happens when we unselect from the middle of the array, then select it again, and do such operations in any order?

vicky-primathon:
> @vicky-primathon Seems like this is only selecting the `selectedRowIndex`, but it is not populating `selectedRow`.
> 
> We need to be careful of the following scenarios:
> 
> * What happens to the `selectedRow` when switching from single select to multi select?
> * What happens to `selectedRow` when switching from multi select to single select?
> * What happens when we unselect from the middle of the array, then select it again, and do such operations in any order?

Whenever its a multi selection we use last selection from selectedRowIndices or else selectedRowIndex to compute selectedRow.
Whenever selection type is changed from single select to multi selection, row selection is preserved.
Whenever selection type is changed from multi select to single selection, row selection is not preserved.
We use last selected row as selectedRow in case of multi row selection. So whatever row user selected last is made as selectedRow.

riodeuno:
@vicky-primathon I can't seem to see the `selectedRow` working. Attached is the screen recording, please review.

https://user-images.githubusercontent.com/103687/135846903-2b7252dc-ef19-496b-9b27-503d3de185cf.mov

vicky-primathon:
> @vicky-primathon I can't seem to see the `selectedRow` working. Attached is the screen recording, please review.
> 
>  Screen.Recording.2021-10-04.at.4.53.39.PM.mov

@riodeuno above case is handled and unit test added for the same

PR #7967, opened 2021-09-30
fix: Don't include the deleted page in applicationPages for export application flow

abhvsn:
## Description

> Check if the page is deleted (published and unpublished part) and then include to applicationPages list in application object

## Type of change

## How Has This Been Tested?

> Tested manually

## Checklist:


nidhi-nair (changes requested):
Hey instead of discarding the entire object, can you add the `deleted = true` boolean flag during import in case the page is missing? As we discussed yesterday, this might break other parts of the application, and it is best to retain the original application's structure as much as possible.

abhvsn:
Hey, I did start with the `deleted=true` but was still facing the issue where deleted unpublished page was still showing up in imported application, so modified the approach to remove reference from the applicationPage within the exported application object and keep the deleted page object as it is. This decision has been taken to retain the original structure because when we delete the unpublished page that reference is removed from the applicationPage, just trying to imitate that here.

nidhi-nair (approved):

PR #8015, opened 2021-09-30
fix: [Filter Library]Prepared Statements doesn't set value if the data type is not supported in H2. Defaulting to String for unsupported types.

trishaanand:


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/FilterDataService.java
```diff
@@ -538,10 +539,9 @@ private PreparedStatement setValueInStatement(PreparedStatement preparedStatemen
                     break;
                 }
                 case STRING: {
```
sumitsum, line 541: Should the `case STRING` case be removed because it is taken care of by the default case now ?

trishaanand: I have added this simply for code understanding than any purpose. I have done the same thing above for FLOAT and DOUBLE. Would it be okay if we let it be?

sumitsum: Nit: In that case, can we remove the curly braces from the `STRING` case ?

trishaanand: Done! Nice catch.

sumitsum (approved):

PR #8017, opened 2021-09-30
fix: Handling the empty object set for where clause value in new google sheets action

### LINKED ISSUE #8001: [Bug] Google sheet is broken
Unable to execute any query as it ends up displaying  an error

trishaanand:
Fixes #8001

### DIFF
--- a/app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/MethodConfig.java
+++ b/app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/MethodConfig.java
@@ -91,7 +91,7 @@ public class MethodConfig {
                         this.rowObjects = propertyValue;
                         break;
                     case "where":
-                        if (value != null) {
+                        if (value != null && value instanceof List) {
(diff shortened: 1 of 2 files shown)


mohanarpit (approved):

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/MethodConfig.java
```diff
@@ -89,8 +91,15 @@ public MethodConfig(List<Property> propertyList) {
                         this.rowObjects = propertyValue;
                         break;
                     case "where":
-                        if (property.getValue() != null && !((List<Object>) property.getValue()).isEmpty()) {
-                            this.whereConditions = Condition.generateFromConfiguration((List<Object>) property.getValue());
+                        if (value != null) {
```
nidhi-nair, line 94: Just to be sure can you add an instanceof check here as well? Not sure how far the UI issue goes at this point

nidhi-nair (approved):

PR #8158, opened 2021-10-04
feat: setInterval and clearInterval support

hetunandu:
## Description

> TODO


hiteshjoshi:
@hetunandu there is just one small bug for setInterval. 
If I delete the button which has setInterval triggered on onClick. The onClick event on the edit more does not deregister automatically. Not a big issue as its for current session only, the moment you reload its fine.

MichalisDBA:
When you set the clearInterval widget and add the ID it takes the form of {{clearInterval(,,'1234')}} you have to manually edit the JS and remove the 2 ,, to make the clearinterval work.

PR #8187, opened 2021-10-05
fix: Don't run test build docker for external PRs

nidhi-nair:
The test build Docker image job requires secrets and should not be triggered by external PRs.

Second commit is for this failure: https://github.com/appsmithorg/appsmith/runs/3799275555?check_suite_focus=true


.github/workflows/test-build-docker-image.yml
```diff
@@ -486,7 +492,12 @@ jobs:
 
   ui-test-result:
     needs: ui-test
-    if: always()
+    if: always() &&
+      (github.event_name == 'workflow_dispatch' ||
+      github.event_name == 'push' ||
+      (github.event_name == 'pull_request_review' &&
+      github.event.review.state == 'approved' &&
+      github.event.pull_request.head.repo.full_name == github.repository))
```
sharat87, line 500: If Cypress tests are skipped, this job used to fail. But with this change, if Cypress tests are skipped, this job will also be skipped. Is that what we want here? If yes, do we even still need this job? I'm okay with removing it in a future PR so as to not make big changes to the workflow in a single PR, just want to know the intention here.

nidhi-nair: The idea with all the filters is that none of these jobs get executed for external PRs. I would actually question if this entire workflow is required for every PR or whether we can live with building images only on release and master. For now though, we're skipping through so that the ok-to-test job can override these statuses instead.

sharat87 (approved):

PR #8207, opened 2021-10-05
fix: check with url path when redirecting to default application url

### LINKED ISSUE #8200: [Bug] Redirect to default application if default redirect url has query params
## Description

[What happened]
When an user signs up, we create a default application for them and redirect to that application when signup and login is finished. This is done if user has no defaul redirect url set or the default redirect url is `/applications`. This does not work if the default redirect url is `/applications?someVar=someVal` 

1. Go to `/signup?redirectUrl=/applications?error=true`
2. Enter email and password and Click signup
3. Check urls and redirects
4. You should be landed on homepage. You should've landed on the My first application

### Important Details

- Version: [Cloud / Self-Hosted]
- OS: [All]
- Browser [All]
- Environment [production, release, deploy preview]

nayan-rafiq:
## Description
When deciding whether to redirect to default application or the redirect url provided during signup, it'll compare with url path now. Earlier it was considering any query params added to the redirect url.

Fixes #8200 

## Type of change

## How Has This Been Tested?
- Manually

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/handlers/AuthenticationSuccessHandler.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/handlers/AuthenticationSuccessHandler.java
@@ -184,7 +184,8 @@ public class AuthenticationSuccessHandler implements ServerAuthenticationSuccess
 
         boolean addFirstTimeExperienceParam = false;
         if (isFromSignup) {
-            if(redirectUrl.endsWith(RedirectHelper.DEFAULT_REDIRECT_URL) && defaultApplication != null) {
+            URI redirectUri = URI.create(redirectUrl);
+            if(redirectUri.getPath().endsWith(RedirectHelper.DEFAULT_REDIRECT_URL) && defaultApplication != null) {
                 addFirstTimeExperienceParam = true;
                 HttpHeaders headers = exchange.getRequest().getHeaders();
                 redirectUrl = redirectHelper.buildApplicationUrl(defaultApplication, headers);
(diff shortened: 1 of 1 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/handlers/AuthenticationSuccessHandler.java
```diff
@@ -184,7 +184,8 @@
 
         boolean addFirstTimeExperienceParam = false;
         if (isFromSignup) {
-            if(redirectUrl.endsWith(RedirectHelper.DEFAULT_REDIRECT_URL) && defaultApplication != null) {
+            URI redirectUri = URI.create(redirectUrl);
+            if(redirectUri.getPath().endsWith(RedirectHelper.DEFAULT_REDIRECT_URL) && defaultApplication != null) {
```
sharat87, line 188: Going by the docs, the `.getPath()` method may return `null`. So, can we add a null-check, just in case?

```suggestion
            if(redirectUri.getPath() != null && redirectUri.getPath().endsWith(RedirectHelper.DEFAULT_REDIRECT_URL) && defaultApplication != null) {
```

app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/handlers/AuthenticationSuccessHandler.java
```diff
@@ -184,7 +184,8 @@
 
         boolean addFirstTimeExperienceParam = false;
         if (isFromSignup) {
-            if(redirectUrl.endsWith(RedirectHelper.DEFAULT_REDIRECT_URL) && defaultApplication != null) {
+            URI redirectUri = URI.create(redirectUrl);
```
sharat87, line 187: If the `redirectUrl` is set to a non-URL, this will throw a `IllegalArgumentException`, which will cause the user to see an internal server error. Can we catch this error and just if this is the case, just take the through the sign-posting flow perhaps?

app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/handlers/AuthenticationSuccessHandler.java
```diff
@@ -204,13 +205,16 @@
...
-                        addFirstTimeExperienceParam = true;
-                        HttpHeaders headers = exchange.getRequest().getHeaders();
-                        url = redirectHelper.buildApplicationUrl(defaultApplication, headers);
-                    }
                     if (isFromSignup) {
+                        URI redirectUri = URI.create(url);
+                        boolean addFirstTimeExperienceParam = false;
+
+                        // only redirect to default application if the redirectUrl contains no other url
+                        if(redirectUri.getPath().endsWith(RedirectHelper.DEFAULT_REDIRECT_URL) && defaultApplication != null) {
```
sharat87, line 213: Same here, as with the other two comments please.

sharat87 (changes requested):
Looks good, just a few edge cases and we should be ready.

app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/handlers/AuthenticationSuccessHandler.java
nayan-rafiq: Will do it. Thanks.

nayan-rafiq:
@sharat87 I've rebased from release because the UI tests were failing. Would you please approve again?

PR #8357, opened 2021-10-08
fix(Dropdown): remove empty dropdown in case of no options

### LINKED ISSUE #7921: [Bug] When options are empty, the dropdown has this small box
## Description
Screenshot 2021-09-29 at 10 26 04 AM
reproduce this behaviour, include console / network logs & screenshots

Open a select widget when there are no options

### Important Details

- Version: All
- OS: MacOSX
- Browser: Chrome
- Environment: All

bhavin-a:
## Description

The PR fixes a bug in case an empty options array is passed to the Select widget. Earlier, it used to render an empty wrapper.

Fixes #7921 

## Type of change

## How Has This Been Tested?

1. Run the client app
2. Drag a select widget from the side nav and pass empty options (`[]`)
3. Click on `-- Select --`
3. There **should be no** empty wrapper below the select 

## Checklist:

- [ ] I have made corresponding changes to the documentation - NA

### DIFF
--- a/app/client/src/widgets/DropdownWidget/component/index.tsx
+++ b/app/client/src/widgets/DropdownWidget/component/index.tsx
@@ -200,7 +200,7 @@ class DropDownComponent extends React.Component<DropDownComponentProps> {
             }
             itemRenderer={this.renderSingleSelectItem}
             items={this.props.options}
-            noResults={<MenuItem disabled text="No Results" />}
+            noResults={<MenuItem disabled text="No Results Found" />}
             onItemSelect={this.onItemSelect}
             onQueryChange={
               this.props.serverSideFiltering ? this.serverSideSearch : undefined
--- a/app/client/src/widgets/DropdownWidget/widget/index.test.tsx
+++ b/app/client/src/widgets/DropdownWidget/widget/index.test.tsx
@@ -77,6 +77,6 @@ describe("<DropdownWidget />", () => {
     const selectElement = screen.getByText("-- Select --");
     fireEvent.click(selectElement);
 
-    expect(screen.getByText("No Results")).toBeInTheDocument();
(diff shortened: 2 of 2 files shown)


Tooluloope:
> @Tooluloope While investigating the code, I found that at few instances, we are using `No Results` instead of `No Item Found`.
> 
> image
> 
> Should I use the same here as well? Below is how it looks when rendered with that. Thoughts?
> 
> image
> 
> With the above, we also won't need a snapshot test. Still not sure of the cypress though 😅. Thanks!

Okay lets go with `No Results`, Do create an issue to change Multiselect to `No result` as well

Bhavin789:
@Tooluloope , found some instances of `No Item Found` as well in the code. Should I create an issue for all of them? Or, do we need to go by `No Item Found` itself?

image


These two statements are equally distributed 😅. I personally found the style of `No results found` to be more in sync with the dropdown option design

image
image

Than `No Item Found`

image

Let me know whichever way we want to go. Will create an issue accordingly or will update this PR. Thanks!

Tooluloope:
@Bhavin789 Please do use No Results Found.
You can check if an issue exist for those others that don't use `No Results found`, if not, create it and tag this PR as a solution

Bhavin789:
@Tooluloope This is how the final message from this PR looks like. Let me know if some changes are required. Thanks!

image

Tooluloope:
> here



> > Sure @Tooluloope. Will do, thanks!
> 
> Hi @Tooluloope , I could not find any existing issues and thus have filed a bug [here](https://github.com/appsmithorg/appsmith/issues/8514) to replace all instances of `No Item Found`. Do let me know if I missed something here.

Would you like to work on this issue @Bhavin789

PR #8472, opened 2021-10-12
feat: JSON Form widget

### LINKED ISSUE #2504: [Epic] JSON Form: Generate forms from JSON Object/Queries/APIs
## Objective

As a user, I should be able to generate a form with input fields by using fields of a DB table schema that automatically generates the queries, bindings and child widgets. As a user, I should be able to configure changes/validations easily here.

## Features

- [ ] It should automatically identify the datatypes and associated widgets for the form
- [ ] It should be able to add basic validations that can be enhanced by the user
- [ ] It should be able to automatically connect the submit buttons to the relevant queries (PUT/POST/CREATE/UPDATE).

## Success Metrics

| Goal  | Metric |
| ------------- | ------------- |
| Simplify user experience  | One-click to generate a form from a single DB schema and be able to map it to a PUT/POST request which updates/saves data |
| Decrease the time needed to build a form  | Time that user spends creating a form |

## Assumptions

- Forms are the core building block of a web application
- It is currently difficult/time-taking to set up a working form comprising of multiple child widgets and bind them all the data query and finally be able to add them to a post request
- One of the most common operations in any DB is the insert/add/create entry operation. It would go a long way to save the user time by generating the scaffolding, if not the complete widget for users

## Requirements

| Requirement   | User Story | Importance | Notes |
| ------------- | ------------- | ------------- | ------------- |
| Connecting to Query/API  | As a user, I should be provided with all the option to connect to different queries and APIs. Furthermore, I should be able to select the fields I need inside the form and I should be able to change the datatype to others from what was automatically connected | High  | If this is not going to be based on the response object and depend on the DB schema, this will be removed from the scope |
| Create a form from a JSON object  | As a user, I should be provided with all the option to create a form from a JSON object apart from the Queries/APIs. All the other functionalities mentioned above will be extended here as well | High  | - |
| Identify datatypes and associated widgets  | As a user, after connecting a DB query to the form, the platform needs to identify the widget types based on the data types. For ex: `Array` maps to `Select` widget. All the widgets need to remain configurable | High  | related to the first requirement |
| Basic input validation  | As a user, I should be able to see basic input validations for the form I have created from an internal DB query and also be able to modify any of these easily | High  | related to the first requirement |
| Connect submit action to queries that update related DB schema | As a user, I should be able to easily configure the submit request which finally updates or saves my data. I should be shown all fields preloaded in my query and appropriate bindings against them  | High  | This might not be involved in this scope |
| Auto naming with schema / object keys | All fields & label in the form should be named based on the schema key they are mapped to | High | - |
| Reopen form config | Once import all fields from an object/datasource I should be able to reopen the config page view to alter any initial configurations. This will reset and changes made to the form elements individually. | High | The default connect data will be used for this, another CTA might be required to update the form data through the same workflow |

### Discussion 21-Sept-2021
- We will get rid of the modal which shows up in when setting up the form fields, Instead we show the fields under the field configuration directly
- Fields on the form will be deep linked to the property pane field configuration, i.e. the respective property pane field setting should open when the field on the form is clicked
- The form fields / view will reflect changes as the form data is updates
- We need to keep in mind a future scope while developing the feature. Nesting of form fields / handling nested form data objects `needs design` `**_pending_**`

### Discussion 27-Sept-2021

We decided that the following will be the type of widget taking into consideration, Additionally the component used in this widget will have there own set of properties (like the table widget) which should be covered by automated tests to avoid any regressions in the future. We also discussed the form properties which will need design review 

Types of fields:
- Text
- Number
- Phone
- Email
- Switch (boolean)
- Checkbox (boolean)
- RadioGroup (exclusive or)
- Select (one out of many)
- Multi-Select (many out of many)
- Files
- Date
- Time
​
Complex fields:
- Grouped fields
​
Additional (optional) fields:
- Rate (Range)
- Map (search)
- JSON object
- Audio

---------------- 

We figured Properties of fields:

Common:
- Type
- Default 
- Label
- Disabled
- Required
- Tooltip
- Visible
- helperText
​
Not common:
- Validation Fn
- Error
- Possible values (options)
- Placeholder
- onChange
​
---------------- 

`**_needs design_**`
Main properties of the form for it to work:
Form: data
Form: submit (action)
Form: source data
Form: Edit the field types and properties
​​
Features in the form:
Form -> input type -> possible properties
Form -> has state of the fields
Form -> onSubmit
Form -> onCancel
Form -> can hide cancel button
Form -> can disable submit button
Form -> can reset itself*
Form -> checks for validations -> knows which fields have errors
Form -> tab navigation works
Form -> can scroll to selected fields
Form -> button location (sticky | bottom)
Form -> field styling (borders, colors, label alignment, error colors, helper text, etc)
Form -> spacing between fields
Form -> all fields have full width

---------

Library to consider: react-hook-forms

------------

### Discussion 19-Oct-2021
`Files` component is being considered out of scope right now as the implementation of the widget component is not well constructed to be used in this widget and would need a lot of refactoring.
This can be considered in the next iteration.

## Out of Scope

- Dynamic width / auto layout of form elements
- Alignment templates/Layouts - As a dev, I should be able to choose the layout for aligning my input fields in the form. Templates can be 2 Column, Single column, Label & input side by side. Or, Drag to re-order elements side by side or one below the other

## Developer Handoff Document in Figma
Link: https://www.figma.com/file/pZuLD9mWXtk45TOjAeh32b/Form-Widget-Thu-26-Aug?node-id=2%3A2

## Questions

| Question  | Answer | Date Answered |
| ------------- | ------------- | ------------- |
| How to identify the submit request needs to be Create/Update? |  It will by default be POST the user can change to there choice |   |
| Can we add necessary formatting function ex: JSON.stringify() automatically to remove any invalid api attributes | | |

ashit-rath:
## Description
This widget generates forms with input fields automatically from any JSON data source. Generated form field's properties can be modified to cater to the form's functionality. 
It supports arrays and objects with n-level depth but the property pane can show properties of nested fields up to 5 levels of depth only, beyond that the configuration object has to be edited manually.

### Nomenclature
- Field - A field is an input entity on the form. It can be a multi-select, radio input, text input, etc.
- Schema - This is a representation of the form that is generated from the JSON. It holds every configuration for every field and it can be considered as the source of truth for the form. 
- SchemaItem - Each individual field's state in the schema is a schema item.
- Source data - It is the JSON input that is provided by the Appsmith user from which the form fields are generated.

### Sample source data and corresponding schema structure:
```JSON
{
  "name": "John",
  "age": 10
}
```

Schema
```js
{
  __root_schema__: {
    isDisabled: false,
    isRequired: false,
    label: "",
    isVisible: true,
    children: {
      name: {
        isDisabled: false,
        isRequired: false,
        label: "Name",
        isVisible: true,
        isSpellCheck: false,
        children: {},
        dataType: DataType.STRING,
        defaultValue:
          "{{((sourceData, formData, fieldState) => (sourceData.name))(JSONForm1.sourceData, JSONForm1.formData, JSONForm1.fieldState)}}",
        fieldType: FieldType.TEXT,
        sourceData: "Test name",
        isCustomField: false,
        name: "name",
        accessor: "name",
        identifier: "name",
        originalIdentifier: "name",
        position: 0,
      },
      age: {
        isDisabled: false,
        isRequired: false,
        label: "Age",
        isVisible: true,
        isSpellCheck: false,
        children: {},
        dataType: DataType.NUMBER,
        defaultValue:
          "{{((sourceData, formData, fieldState) => (sourceData.age))(JSONForm1.sourceData, JSONForm1.formData, JSONForm1.fieldState)}}",
        fieldType: FieldType.NUMBER,
        sourceData: 20,
        isCustomField: false,
        name: "age",
        accessor: "age",
        identifier: "age",
        originalIdentifier: "age",
        position: 1,
      },
    isCustomField: false,
    name: "",
    accessor: "",
    identifier: "",
    originalIdentifier: "",
    position: -1,
  },
}
```

Fixes #2504 

https://miro.com/app/board/uXjVOXio8tY=/

## Type of change

## How Has This Been Tested?

- Cypress Tests
- Jest Tests
- Manual testing

## Checklist:

    // Code coverage diff between base branch:release and head branch: feature/form-builder-widget 
 :red_circle: | total | 55.4 **(-0.23)** | 36.87 **(0.01)** | 34.94 **(0)** | 55.7 **(-0.24)**
 :red_circle: | app/client/src/components/ads/DraggableListCard.tsx | 13.46 **(-0.83)** | 3.28 **(-0.17)** | 0 **(0)** | 15.56 **(-1.11)**
 :green_circle: | app/client/src/components/ads/DraggableListComponent.tsx | 41.03 **(0)** | 18.18 **(0)** | 33.33 **(0)** | 44.12 **(1.7)**
 :sparkles: :new: | **app/client/src/components/editorComponents/Tooltip.tsx** | **100** | **100** | **100** | **100**
 :red_circle: | app/client/src/components/propertyControls/BaseControl.tsx | 75 **(-8.33)** | 44.44 **(19.44)** | 50 **(-16.67)** | 71.43 **(-10.39)**
 :sparkles: :new: | **app/client/src/components/propertyControls/ButtonControl.tsx** | **46.43** | **11.76** | **33.33** | **57.14**
 :green_circle: | app/client/src/components/propertyControls/DropDownControl.tsx | 89.29 **(0.4)** | 64.71 **(1.07)** | 71.43 **(0)** | 88 **(0.5)**
 :sparkles: :new: | **app/client/src/components/propertyControls/FieldConfigurationControl.tsx** | **18.8** | **8.33** | **10** | **21.82**
 :sparkles: :new: | **app/client/src/components/propertyControls/JSONFormComputeControl.tsx** | **24.47** | **9.76** | **11.76** | **24.1**
 :red_circle: | app/client/src/components/propertyControls/PrimaryColumnsControl.tsx | 36 **(0)** | 11.94 **(-0.37)** | 19.23 **(0)** | 34.97 **(0)**
 :green_circle: | app/client/src/components/propertyControls/index.ts | 86.49 **(1.2)** | 100 **(0)** | 0 **(0)** | 86.11 **(1.26)**
 :green_circle: | app/client/src/entities/Widget/utils.ts | 93 **(0.14)** | 89.13 **(0)** | 100 **(0)** | 92.86 **(0.15)**
 :red_circle: | app/client/src/icons/WidgetIcons.tsx | 58.76 **(-0.19)** | 100 **(0)** | 14.89 **(-0.33)** | 58.76 **(-0.19)**
 :green_circle: | app/client/src/pages/Editor/PropertyPane/PropertySection.tsx | 90.32 **(0.66)** | 56.25 **(-0.89)** | 83.33 **(0)** | 89.29 **(0.4)**
 :red_circle: | app/client/src/sagas/SnipingModeSagas.ts | 19.44 **(-0.56)** | 3.03 **(-0.1)** | 40 **(0)** | 20.19 **(-0.6)**
 :red_circle: | app/client/src/selectors/propertyPaneSelectors.tsx | 88.89 **(-0.97)** | 70.37 **(-1.63)** | 91.67 **(-8.33)** | 88.71 **(-1.12)**
 :green_circle: | app/client/src/utils/WidgetFactory.tsx | 74.47 **(0.56)** | 72.22 **(1.63)** | 64.71 **(0)** | 74.47 **(0.56)**
 :red_circle: | app/client/src/utils/autocomplete/EntityDefinitions.ts | 26.32 **(-0.71)** | 0 **(0)** | 12.5 **(-1.79)** | 25.71 **(-0.76)**
 :sparkles: :new: | **app/client/src/utils/hooks/useDeepEffect.ts** | **100** | **100** | **100** | **100**
 :green_circle: | app/client/src/widgets/WidgetUtils.ts | 86.27 **(6.9)** | 61.9 **(11.9)** | 62.5 **(8.65)** | 83.33 **(8.82)**
 :red_circle: | app/client/src/widgets/BaseInputWidget/component/index.tsx | 62.6 **(-0.97)** | 44.85 **(-1.11)** | 60 **(-1.36)** | 61.6 **(-1)**
 :red_circle: | app/client/src/widgets/CheckboxWidget/component/index.tsx | 95 **(-1.97)** | 65.71 **(-2.03)** | 92.31 **(1.4)** | 93.75 **(-2.4)**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/FormContext.tsx** | **100** | **100** | **100** | **100**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/constants.ts** | **100** | **100** | **100** | **100**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/helper.ts** | **97.03** | **91.23** | **100** | **96.47**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/index.ts** | **100** | **100** | **100** | **100**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/schemaParser.ts** | **98.17** | **90.3** | **100** | **98.74**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/schemaTestData.ts** | **100** | **100** | **100** | **100**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/component/Accordion.tsx** | **33.33** | **30** | **0** | **41.38**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/component/Field.tsx** | **27.78** | **16.67** | **0** | **29.41**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/component/FieldLabel.tsx** | **35.19** | **20.69** | **0** | **44.19**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/component/Form.tsx** | **29.35** | **20** | **0** | **30.68**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/component/NestedFormWrapper.tsx** | **55.56** | **33.33** | **0** | **55.56**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/component/index.tsx** | **27.08** | **33.33** | **0** | **27.66**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/component/styleConstants.ts** | **100** | **100** | **100** | **100**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/component/useFixedFooter.ts** | **19.44** | **0** | **0** | **20**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/fields/ArrayField.tsx** | **25.86** | **20** | **0** | **27.27**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/fields/BaseInputField.tsx** | **18.1** | **2.3** | **7.69** | **19.05**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/fields/CheckboxField.tsx** | **35.9** | **14.29** | **0** | **38.89**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/fields/CurrencyInputField.tsx** | **43.55** | **46.43** | **20** | **45.61**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/fields/DateField.tsx** | **50.63** | **13.33** | **40** | **50.68**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/fields/FieldRenderer.tsx** | **18.75** | **0** | **0** | **21.43**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/fields/InputField.tsx** | **55.41** | **57.89** | **40** | **55.56**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/fields/MultiSelectField.tsx** | **27.27** | **5.41** | **0** | **32.81**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/fields/ObjectField.tsx** | **31.25** | **14.29** | **0** | **34.09**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/fields/PhoneInputField.tsx** | **27.08** | **0** | **0** | **27.91**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/fields/RadioGroupField.tsx** | **40.74** | **0** | **0** | **42.31**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/fields/SelectField.tsx** | **29.31** | **4.55** | **0** | **32.08**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/fields/SwitchField.tsx** | **35.29** | **0** | **0** | **37.5**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/fields/index.ts** | **100** | **100** | **100** | **100**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/fields/useBlurAndFocusEvents.ts** | **12.5** | **0** | **0** | **12.5**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/fields/useObserveAccessor.ts** | **30.77** | **0** | **0** | **30.77**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/fields/useRegisterFieldValidity.ts** | **95.45** | **100** | **100** | **95**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/fields/useUpdateInternalMetaState.ts** | **35.71** | **0** | **0** | **35.71**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/widget/helper.ts** | **98.91** | **90.48** | **100** | **98.82**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/widget/index.tsx** | **29.81** | **0** | **22.22** | **31.91**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/widget/propertyConfig.ts** | **33.96** | **0** | **16.67** | **34.04**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/widget/propertyConfig/generatePanelPropertyConfig.ts** | **42.11** | **15.38** | **16.67** | **38.89**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/widget/propertyConfig/helper.ts** | **100** | **73.68** | **100** | **100**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/widget/propertyConfig/properties/array.ts** | **55.56** | **100** | **0** | **55.56**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/widget/propertyConfig/properties/checkbox.ts** | **45.45** | **100** | **0** | **45.45**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/widget/propertyConfig/properties/common.ts** | **22.22** | **0** | **11.11** | **23.26**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/widget/propertyConfig/properties/date.ts** | **30.77** | **100** | **0** | **30.77**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/widget/propertyConfig/properties/index.ts** | **100** | **100** | **100** | **100**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/widget/propertyConfig/properties/input.ts** | **10.42** | **0** | **0** | **10.42**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/widget/propertyConfig/properties/multiSelect.ts** | **71.19** | **90** | **46.67** | **69.81**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/widget/propertyConfig/properties/radioGroup.ts** | **55.56** | **100** | **0** | **55.56**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/widget/propertyConfig/properties/select.ts** | **59.46** | **85.71** | **22.22** | **61.11**
 :sparkles: :new: | **app/client/src/widgets/JSONFormWidget/widget/propertyConfig/properties/switch.ts** | **45.45** | **100** | **0** | **45.45**
 :red_circle: | app/client/src/widgets/MultiSelectWidget/component/index.tsx | 18.84 **(-0.56)** | 0 **(0)** | 0 **(0)** | 20 **(-0.63)**
 :red_circle: | app/client/src/widgets/MultiSelectWidgetV2/component/index.tsx | 11.82 **(-0.22)** | 0 **(0)** | 0 **(0)** | 13.27 **(-0.27)**
 :green_circle: | app/client/src/widgets/SwitchWidget/component/index.tsx | 94.12 **(0.37)** | 62.5 **(0)** | 50 **(0)** | 94.12 **(0.37)**
 :green_circle: | app/client/src/workers/evaluationUtils.ts | 62.01 **(3.43)** | 60.3 **(1.51)** | 67.19 **(1.56)** | 60.11 **(2.13)**
 :green_circle: | app/client/src/workers/validations.ts | 81.32 **(0)** | 71.69 **(1.12)** | 76.67 **(0)** | 82.39 **(0)**</details>


app/client/src/components/propertyControls/DropDownControl.tsx
```diff
@@ -9,13 +9,16 @@ class DropDownControl extends BaseControl<DropDownControlProps> {
       label: "No selection.",
       value: undefined,
     };
+
```
sbalaji1192, line 12: Need to move this to validations function itself

app/client/src/components/propertyControls/SwitchControl.tsx
```diff
@@ -1,22 +1,40 @@
 import React from "react";
+import { get } from "lodash";
 import BaseControl, { ControlProps } from "./BaseControl";
 import Switch from "components/ads/Switch";
 
 class SwitchControl extends BaseControl<ControlProps> {
+  getPropertyValue = () => {
+    // We are using the evaluated value as the JSONForm widget might
+    // pass a binding as propertyValue which might lead to unfamiliar behaviors.
+    const evaluatedValue = get(
```
sbalaji1192, line 10: Need to reevaluate this

app/client/src/entities/DataTree/dataTreeWidget.ts
```diff
@@ -61,25 +61,30 @@ export const generateDataTreeWidget = (
...
-    },
-    bindingPaths,
-    triggerPaths,
-    validationPaths,
-    ENTITY_TYPE: ENTITY_TYPE.WIDGET,
-    privateWidgets: {
-      ...widget.privateWidgets,
+
+  // Use _.merge here to merge all the deeply nested object properties.
+  return _.merge(
```
sbalaji1192, line 66: Add benchmarks

sbalaji1192: Need fe coders pod signoff. @ApekshaBhosale

app/client/src/entities/Widget/utils.ts
```diff
@@ -45,10 +45,13 @@ const checkPathsInConfig = (
   return { configBindingPaths, configTriggerPaths, configValidationPaths };
 };
 
+// "originalWidget" param here always contains the complete widget props
```
sbalaji1192, line 48: Need clarification

sbalaji1192 (commented):
Notes till widget/utils

app/client/src/widgets/JSONFormWidget/widget/propertyConfig/helper.ts
```diff
@@ -0,0 +1,153 @@
...
+import SchemaParser, {
+  sanitizeSchemaItemKey,
+} from "widgets/JSONFormWidget/schemaParser";
+import { FieldType, SchemaItem, ARRAY_ITEM_KEY, Schema } from "../../constants";
+import { getGrandParentPropertyPath, getParentPropertyPath } from "../helper";
+import { JSONFormWidgetProps } from "..";
+
+export type HiddenFnParams = [JSONFormWidgetProps, string];
+
+export const fieldTypeUpdateHook = (
```
sbalaji1192, line 12: Add function comments

app/client/src/entities/Widget/utils.ts
```diff
@@ -45,10 +45,13 @@ const checkPathsInConfig = (
...
+// as this function's widget parameter tends to change in each iteration
 const childHasPanelConfig = (
   config: any,
   widget: WidgetProps,
   basePath: string,
+  originalWidget: WidgetProps,
```
riodeuno, line 54: Please verify if the Table Widget, Menu Button Widget, Tabs Widget, Button Group widget, etc, are working correctly, w.r.t the `hidden` property. Because, we were passing a subset of the original props earlier. Now that we're passing the full widget, maybe the path computations would result in incorrect values and incorrect hidden states as a result.

riodeuno (changes requested):

app/client/src/entities/Widget/utils.ts
ashit-rath: @riodeuno Tested all the widgets where `panelConfig` is used and they all work fine.

vivekverma2312:
@ashit-rath Please look into few observations, I am still testing the feature.
1. Connect data message is displayed on widget for a sec before loading the default data
2. Removing sample data do not make the widget blank
https://www.loom.com/share/3e8e343aec264f33af079963960b9885
3. Writing source code for multi-select widget still shows default data in drop-down https://www.loom.com/share/83c03bae8ce6473bb9d8618d02dd73b1
4. After I make any change in the form like delete or add and then refresh the page the source code data is updated https://www.loom.com/share/47dcd16b8e5344c48d5b901d3c5c6ac4
5. Sequence of form changes when we make any change in source data https://www.loom.com/share/2ef1480fb3d54a66b013fc6caea50093
6. Changing widget type to Text input displays object by default in field https://www.loom.com/share/9deb7d0a66d04cc5b24a108fb162fa8e

ashit-rath:
@vivekverma2312 
>1 Connect data message is displayed on widget for a sec before loading the default data

I'll fix this

>2. Removing sample data do not make the widget blank

It's a safety mechanism, there would be cases where the `source data` might get undefiend or empty object from an API that is still fetching the data. If we try to process `undefined` or empty data then the user might lose field configuration.

>3. Writing source code for multi-select widget still shows default data in drop-down

The multi-select field gets default options, that have to be modified by the user. We do not process the default value as options as that might be destructive as the default value might change and we would have to update the options which can get messy.

>4. After I make any change in the form like delete or add and then refresh the page the source code data is updated

Whatever you enter in the form is metadata i.e it is only persisted until you are active on the page. It's similar to the input widget having some default data and upon refresh that is again set.
Similarly array field as some default value i.e 2 entries `curefit` and `appsmith`. 
So on refresh, the same default value of the array is applied.

>5 Sequence of form changes when we make any change in source data

Yeah, it is considered as a new entry (as the property name changes) and get's added at the end. You can reposition is in the property pane `Field Configuration`

>6 Changing widget type to Text input displays object by default in field

This seems to be a bug

vivekverma2312:
@ashit-rath 
7. After clicking on add new and deploy the new fields are not displayed. similar behaviour for delete. it deploys the source code data.
https://www.loom.com/share/17ca6ef8f58e4f03920d6e900c764b53
8. Date widget is getting highlighted in blue colour but all other fields ate highlighted in green colour.
Screenshot 2022-02-16 at 5 32 02 PM
Screenshot 2022-02-16 at 5 31 56 PM

ashit-rath:
> 7. After clicking on add new and deploy the new fields are not displayed. similar behaviour for delete. it deploys the source code data.
For arrays using the `Add New` button won't  define how many entries you would have in the deploy mode. It is similar to how entering some text in an input field won't define what the value of that input field would be in the deploy mode.

For changing how many entires show up, you would have to change the default value of the array field in the property pane.

>8. Date widget is getting highlighted in blue colour but all other fields ate highlighted in green colour.
This has to be fixed in the individual widgets itself.

app/client/src/widgets/JSONFormWidget/fields/MultiSelectField.tsx
```diff
@@ -0,0 +1,263 @@
...
+    propertyValue: filterText,
+  });
+
+  const { componentDefaultValue, fieldDefaultValue } = useMemo(() => {
+    let componentDefaultValue: LabelValueType[] = [];
+    let fieldDefaultValue: LabelValueType["value"][] = [];
+
+    const values: LabelValueType["value"][] | LabelValueType[] = (() => {
+      if (
+        schemaItem.defaultValue !== undefined &&
```
sbalaji1192, line 132: Suggestion: Can use lodash's isNil

app/client/src/widgets/JSONFormWidget/fields/RadioGroupField.tsx
```diff
@@ -0,0 +1,103 @@
...
+  options: [
+    { label: "Yes", value: "Y" },
+    { label: "No", value: "N" },
+  ],
+};
+
+const isValid = (
+  schemaItem: RadioGroupFieldProps["schemaItem"],
+  value?: string,
+) => (schemaItem.isRequired ? Boolean(value) : true);
```
sbalaji1192, line 35: Can this be moved to a common file and be used across multiple fields?

app/client/src/widgets/JSONFormWidget/fields/SelectField.tsx
```diff
@@ -0,0 +1,231 @@
...
+type MetaProps = {
+  filterText?: string;
+};
+
+type DefaultValue = string | number | DropdownOption | null | undefined;
+
+type SelectComponentProps = FieldComponentBaseProps &
+  MetaProps & {
+    placeholderText?: string;
+
```
sbalaji1192, line 30: remove empty line

app/client/src/widgets/JSONFormWidget/fields/useEvents.ts
```diff
@@ -0,0 +1,87 @@
...
+  "onfocus" | "onblur" | "addEventListener" | "removeEventListener"
+>;
+
+type UseEventsProps = {
+  fieldBlurHandler?: ControllerRenderProps["onBlur"];
+  onFocusDynamicString?: string;
+  onBlurDynamicString?: string;
+};
+
+function useEvents<TElement extends BaseEvents>({
```
sbalaji1192, line 18: Suggestion: Since its only handling focus and blur, we can rename accordingly.

app/client/src/widgets/JSONFormWidget/helper.ts
```diff
@@ -0,0 +1,148 @@
...
+
+export const validateOptions = (
+  values: unknown,
+): values is LabelValueType["value"][] | LabelValueType[] => {
+  if (!Array.isArray(values)) return false;
+
+  let hasPrimitive = false;
+  let hasObject = false;
+  for (const value of values) {
+    if (value === undefined || value === null || Number.isNaN(value)) {
```
sbalaji1192, line 99: use _.isNil

app/client/src/widgets/JSONFormWidget/schemaParser.ts
```diff
@@ -0,0 +1,711 @@
...
+
+type ParserOptions = {
+  currSourceData?: JSON | string;
+  fieldType?: FieldType;
+  isCustomField?: boolean;
+  prevSchema?: Schema;
+  schemaItem?: SchemaItem;
+  // - skipDefaultValueProcessing
+  // When an array type is detected, by default we want to process default value
+  // only for the array level and keep every every field below it as empty default.
```
sbalaji1192, line 38: typo `every`

app/client/src/widgets/JSONFormWidget/schemaParser.ts
```diff
@@ -0,0 +1,711 @@
...
+ * @example
+ *  getSourcePath("age", "sourceData.profile") -> "sourceData.profile.age"
+ *  getSourcePath(0, "sourceData.education") -> "sourceData.education[0]"
+ */
+
+export const getSourcePath = (name: string | number, basePath?: string) => {
+  let nameWithNotation = name;
+
+  if (typeof name === "string") {
+    const sanitizedName = typeof name === "string" && sanitizeKey(name);
```
sbalaji1192, line 88: no need to check again `typeof name === "string"`

app/client/src/widgets/JSONFormWidget/widget/propertyConfig/generatePanePropertyConfig.ts
```diff
@@ -0,0 +1,122 @@
+import { get, isEmpty } from "lodash";
```
sbalaji1192, line 1: typo in file name. panel is missing `l`

app/client/src/widgets/MultiSelectWidgetV2/component/index.tsx
```diff
@@ -76,15 +78,18 @@ function MultiSelectComponent({
...
   onFilterChange,
+  onFocus,
   options,
   placeholder,
   serverSideFiltering,
   value,
   widgetId,
   width,
 }: MultiSelectProps): JSX.Element {
+  console.log("MULTI WIDGET values", value);
```
sbalaji1192, line 92: remove

app/client/src/widgets/WidgetUtils.ts
```diff
@@ -174,3 +178,71 @@ export const escapeSpecialChars = (stringifiedJSONObject: string) => {
...
+      map[prefix] = suffix;
+    }
+  });
+
+  return map;
+};
+
+export const sanitizeKey = (key: string, options?: SanitizeOptions) => {
+  // Step 1 Replaces all spl. characters/spaces with _
+  let sanitizedKey = key.replace(/[^\w]/gi, "_");
```
sbalaji1192, line 216: We have started allowing (not merged to release yet) non - ASCII characters in the table column.  we need to start doing same thing in the JSON form as well.

sbalaji1192 (commented):
Great work !! Have provided some minor callouts and suggestions. please go through them !

app/client/src/widgets/JSONFormWidget/widget/propertyConfig/generatePanePropertyConfig.ts
ashit-rath: Nice  catch!

app/client/src/components/propertyControls/FieldConfigurationControl.tsx
```diff
@@ -0,0 +1,278 @@
+import React from "react";
+import { cloneDeep, isEmpty, maxBy, set, sortBy } from "lodash";
+
```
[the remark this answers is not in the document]
ashit-rath: This new line actually separates node_modules imports from local imports. I know we don't have such conventions but this really helps to identify where things are coming from with a quick glance.

app/client/src/components/propertyControls/JSONFormComputeControl.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+  return [];
+}
+
+function processSchemaItem(schemaItem: SchemaItem, defaultValue?: any) {
+  if (schemaItem.dataType === DataType.OBJECT) {
+    return processObject(schemaItem.children, defaultValue);
+  }
+
+  if (schemaItem.dataType === DataType.ARRAY) {
```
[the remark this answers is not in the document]
ashit-rath: I would differ a bit here. As we are returning from the first if condition, else-if is not necessary.

app/client/cypress/support/commands.js
```diff
@@ -2608,8 +2618,6 @@ Cypress.Commands.add("openPropertyPane", (widgetType) => {
   )
     .first()
     .click({ force: true });
```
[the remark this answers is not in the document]
ashit-rath: Ahh yes, my bad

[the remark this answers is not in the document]
ashit-rath: Tested by removing the `wait` in the original command. It doesn't load the panels properly so I think the wait seems to be necessary here.

app/client/src/widgets/JSONFormWidget/fields/BaseInputField.tsx
```diff
@@ -0,0 +1,326 @@
...
+        | React.KeyboardEvent<HTMLInputElement>,
+      fieldOnChangeHandler: (...event: any[]) => void,
+      isValueValid: boolean,
+    ) => {
+      const { onEnterKeyPress } = schemaItem;
+      const isEnterKey = e.key === "Enter";
+
+      if (isEnterKey && onEnterKeyPress && isValueValid) {
+        executeAction({
+          triggerPropertyName: "onEnterKeyPress",
```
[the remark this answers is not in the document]
ashit-rath: Actually the event name here is `onEnterKeyPress` which is different to `onSubmit` as in the context of JSONForm `onSubmit` would be a form level-event and `onEnterKeyPress` would be a field level event.
So in order to get proper logs we need to pass `onEnterKeyPress` as the triggerPropertyName, otherwise it might give out incorrect result.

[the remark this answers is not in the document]
sbalaji1192: sounds good!

app/client/src/widgets/JSONFormWidget/fields/MultiSelectField.tsx
```diff
@@ -0,0 +1,263 @@
...
+};
+
+const fieldValuesToComponentValues = (
+  values: LabelValueType["value"][],
+  options: LabelValueType[] = [],
+) => {
+  return values.map((value) => {
+    const option = options.find((option) => option.value === value);
+
+    if (option) return option;
```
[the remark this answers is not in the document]
ashit-rath: Makes sense!

vivekverma2312:
@ashit-rath 
9. Fixed footer in view mode is not proper. It is showing outside the form and at bottom of the page. In case of multiple forms also only one button is displayed.
Screenshot 2022-02-25 at 3 52 03 PM

10. Valid property is throwing error 
Screenshot 2022-02-28 at 10 42 53 AM

chandannkumar:
@ashit-rath I am also testing this PR and have few observations to share

1. Phone Number Input is accepting Characters
https://www.loom.com/share/8d41c59a86c248da9fa37c3320fcce67

2. There is no scroll in dropdown of Field type and Date format dropdown(Select field type as date and see option of Date format)
https://www.loom.com/share/f5aed183405e46e1ae3a750ee64a711f

3. Convert to ISO isn't working for Date Picker field type
https://www.loom.com/share/9dbce6206c9c485dac5eb35059e50154

4. While adding Date picker inside Array displays "Invalid date" by default
https://www.loom.com/share/fb982b15a35d46a49249e1c2cfa12ffe

5. Delete Button doesn't work when Array item of an Array is invisible
https://www.loom.com/share/5c1b9c9d1d094ac08da8759e7738a52b

ashit-rath:
> @ashit-rath 9. Fixed footer in view mode is not proper. It is showing outside the form and at bottom of the page. In case of multiple forms also only one button is displayed. Screenshot 2022-02-25 at 3 52 03 PM
> 
> 10. Valid property is throwing error
> 
> Screenshot 2022-02-28 at 10 42 53 AM

For 10. There is a `()` after name
Should the binding look like this?
```
{{formData.name.FirstName.length > 5}}
```

ashit-rath:
@chandannkumar 
>1. Phone Number Input is accepting Characters

I'll look into this

>2. There is no scroll in dropdown of Field type and Date format dropdown(Select field type as date and see option of Date format)

This is an issue with every dropdown field in the property pane and I think the UI builders pod are looking into it.

>3. Convert to ISO isn't working for Date Picker field type

The convert to ISO does not change the value that you see in the input to ISO but changes the value of that particular field in the `formData` property. Try binding `JSONForm1.formData` to a text widget and toggle this.
Do you think we should update the hint of this property in some way to help user understand this better.

>4 While adding Date picker inside Array displays "Invalid date" by default

I will check this

>5 Delete Button doesn't work when Array item of an Array is invisible

Nice catch! I'll look into this as well

app/client/src/widgets/WidgetUtils.ts
```diff
@@ -174,3 +180,74 @@ export const escapeSpecialChars = (stringifiedJSONObject: string) => {
...
+
+  return map;
+};
+
+export const sanitizeKey = (key: string, options?: SanitizeOptions) => {
+  // Step1 convert to ASCII characters
+  let sanitizedKey = punycode.toASCII(key);
+
+  // Step 2 Replaces all spl. characters/spaces with _
+  sanitizedKey = sanitizedKey.replace(/[^\w]/gi, "_");
```
sbalaji1192, line 221: both `$test` and `#test` will converted to `_test`. This was the original problem with sanitizeKey function. But I guess there is no way to avoid that situation.

app/client/src/widgets/WidgetUtils.ts
```diff
@@ -174,3 +180,74 @@ export const escapeSpecialChars = (stringifiedJSONObject: string) => {
...
+  if (sanitizedKey.length === 0) sanitizedKey = "_";
+
+  // Step 4 Check if key starts with number
+  const [firstCharacter] = sanitizedKey;
+  if (/\d/.test(firstCharacter)) sanitizedKey = `_${sanitizedKey}`;
+
+  // Step 5 handle checking with existing keys if present
+  const { existingKeys = [] } = options || {};
+  if (existingKeys.length) {
+    const keyToIndexMap = generateKeyToIndexMap(existingKeys);
```
sbalaji1192, line 233: Move this declaration after L236

app/client/src/widgets/JSONFormWidget/widget/propertyConfig/helper.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    return [
+      {
+        propertyPath: `${schemaItemPath}.children`,
+        propertyValue: newChildrenSchema,
+      },
+    ];
+  }
+
+  return;
```
sbalaji1192, line 139: not necessary

app/client/src/widgets/JSONFormWidget/widget/propertyConfig/helper.ts
```diff
@@ -0,0 +1,140 @@
...
+    fieldType === schemaItem.fieldType;
+
+  const fieldTypeNotMatches = (fieldType: FieldType) =>
+    fieldType !== schemaItem.fieldType;
+
+  const fieldTypeNotIncludes = (
+    fieldTypes: Readonly<FieldType[]> | FieldType[],
+  ) => !fieldTypes.includes(schemaItem.fieldType);
+
+  const then = (cb: (props: TSchemaItem) => any) => cb(schemaItem);
```
sbalaji1192, line 76: Could we come up with a better name instead of `then` ?

app/client/src/widgets/JSONFormWidget/helper.ts
```diff
@@ -0,0 +1,188 @@
...
+    } else if (isPrimitive(value)) {
+      hasPrimitive = true;
+    } else if (isPlainObject(value) && "label" in value && "value" in value) {
+      hasObject = true;
+    } else {
+      return false;
+    }
+  }
+
+  if (hasPrimitive && hasObject) return false;
```
sbalaji1192, line 150: return !(hasPrimitive && hasObject) should suffice.

app/client/src/widgets/WidgetUtils.ts
```diff
@@ -20,6 +20,12 @@ import {
 } from "components/constants";
 import tinycolor from "tinycolor2";
 
+const punycode = require("punycode/");
```
sbalaji1192, line 23: can't we use import syntax ?

app/client/src/components/propertyControls/JSONFormComputeControl.tsx
```diff
@@ -0,0 +1,267 @@
...
+ */
+function generateAutoCompleteStructure(
+  schema: Schema,
+  defaultValue?: any,
+): any {
+  let obj;
+
+  if (schema) {
+    Object.values(schema).forEach((schemaItem) => {
+      obj = processSchemaItem(schemaItem, defaultValue);
```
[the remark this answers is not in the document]
ashit-rath: Removed the function, was unecessary

app/client/src/components/ads/DraggableListCard.tsx
```diff
@@ -101,6 +101,32 @@ export function DraggableListCard(props: RenderComponentProps) {
     }
   };
 
+  const renderVisibilityIcon = () => {
```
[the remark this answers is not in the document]
ashit-rath: Why do we need to use `useMemo` here? It's not being used for referential equality and there is no heavy computation. Can you think of a reason of using it?

app/client/src/widgets/JSONFormWidget/fields/ArrayField.tsx
```diff
@@ -0,0 +1,318 @@
...
+    defaultValue,
+  );
+
+  useUpdateAccessor({ accessor: schemaItem.accessor });
+
+  const { setMetaInternalFieldState } = useContext(FormContext);
+
+  const basePropertyPath = `${propertyPath}.children.${ARRAY_ITEM_KEY}`;
+
+  const add = () => {
```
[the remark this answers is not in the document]
ashit-rath: Same here as well, I don't see the point of using `useCallback` here, won't seem to improve any perf gain as this is not being passed as a prop to a heavy component or a part of another effect or callback. Please do let me know if I'm missing something here.

app/client/src/widgets/JSONFormWidget/component/FieldLabel.tsx
```diff
@@ -0,0 +1,163 @@
...
+      fontSize: TEXT_SIZES[labelTextSize],
+      fontWeight: styles?.includes(FontStyleTypes.BOLD) ? "bold" : "normal",
+      fontStyle: styles?.includes(FontStyleTypes.ITALIC) ? "italic" : "",
+      textDecoration: styles?.includes(FontStyleTypes.UNDERLINE)
+        ? "underline"
+        : "",
+    };
+  }, [labelStyle, labelTextColor, labelTextSize]);
+
+  // If field and label are to be displayed horizontally then we consider the alignField
```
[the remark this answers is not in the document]
ashit-rath: Force of habit 😅. I'll make sure to use multi-line

app/client/src/widgets/WidgetUtils.ts
ashit-rath: Yes, it will get converted but the second `_test` would get converted to `_test1` if we pass our `existingKeys` with correct previous generated keys.

app/client/src/widgets/WidgetUtils.ts
ashit-rath: There is a soft deprecation warning in the esm syntax import and they recommend 
```
Use require('punycode/') to import userland modules rather than core modules.
```

app/client/src/widgets/WidgetUtils.ts
sbalaji1192: The problem is finding which one is $test and #test.

sbalaji1192: But We got to live with it.

app/client/src/widgets/WidgetUtils.ts
sbalaji1192: Sounds good. Could you add comment for future references?

app/client/src/widgets/WidgetUtils.ts
ashit-rath: Yeah, we have to. It's a trade-off

app/client/src/entities/DataTree/dataTreeWidget.ts
```diff
@@ -116,31 +116,48 @@ export const generateDataTreeWidget = (
...
+   *
+   * {
+   *  foo: { baz: 200 } // bar in "a" object got overridden by baz in "b"
+   * }
+   *
+   * Therefore spread is replaced with "merge" which merges objects recursively.
+   */
+  return _.merge(
+    {},
+    defaultMetaProps,
```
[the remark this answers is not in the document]
sbalaji1192: Any Updates on this?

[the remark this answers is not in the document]
ashit-rath: Yeah, I'm actually looking into this right now. 
I re-arranged it for a reason but I do now feel like it's an unnecessary change but need to verify.
I'll update once I check

app/client/src/components/ads/DraggableListCard.tsx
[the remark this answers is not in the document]
ashit-rath: I don't think the perf gain that we would get here would be significant as we would pass some dependencies to the `useMemo` which we need to make sure that they have referentially equal. 
With this ordeal, the number of comparisons we would be doing would probably nullify the difference we would have otherwise.

app/client/src/widgets/JSONFormWidget/fields/ArrayField.tsx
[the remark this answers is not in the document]
ashit-rath: I would differ here we shouldn't just do such optimizations as they come at a cost.  React is very fast and we shouldn't pre-optimize it like this, eventually, it would just cause degradation in performance due to the no of comparisons that would be happening all over the place.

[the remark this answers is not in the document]
ashit-rath: There is a great article for when using useMemo and useCallback makes sense
https://kentcdodds.com/blog/usememo-and-usecallback

app/client/src/entities/DataTree/dataTreeWidget.ts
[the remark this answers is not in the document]
ashit-rath: @sbalaji1192 Reverted the order change.

chandannkumar:
@ashit-rath Please look into these observations also:

12. App crash right after adding Json form widget onto canvas
https://www.loom.com/share/10950135f2444eb6a49fe42b84d5cf43

13. App crash when using array at starting of the source code in view and edit mode
https://www.loom.com/share/bb7b5420861744d48a28d7e88ee46349

14. App crash after changing the existing Input type field to Array field
https://www.loom.com/share/2f203737cf9f4bf29fe6162a313ca10f

15. Behavior on adding multiple objects in an array. In video it shows: when we add objects with same name, its creates 1 Array items and only 1 field configuration, but when i tried to add new object with same name but in lower case, a new field got added inside the array item
https://www.loom.com/share/449d0f20c1b24d26a9607283afc99371 

16.  Unable to view icon dropdown after click on icon field of submit and reset button
image

ashit-rath:
@chandannkumar 
> 15. Behavior on adding multiple objects in an array. In video it shows: when we add objects with same name, its creates 1 Array items and only 1 field configuration, but when i tried to add new object with same name but in lower case, a new field got added inside the array item

Yes the keys are case sensitive so the behaviour you are seeing is valid.

> 16. Unable to view icon dropdown after click on icon field of submit and reset button

This is being fixed in another PR #11736

chandannkumar:
@ashit-rath 
This feature PR has been Tested and working as expected.
Also all the reported bugs are fixed and verified.

sbalaji1192 (approved):
Great Work!. Had few minor comments.

app/client/src/widgets/JSONFormWidget/widget/propertyConfig.ts
```diff
@@ -0,0 +1,534 @@
...
+const MAX_NESTING_LEVEL = 5;
+
+const panelConfig = generatePanelPropertyConfig(MAX_NESTING_LEVEL);
+
+export const sourceDataValidationFn = (
+  value: any,
+  props: JSONFormWidgetProps,
+  _?: any,
+) => {
+  if (_.isNil(value)) {
```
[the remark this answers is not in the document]
ashit-rath: JSON.parse in line 45 would throw an error for an empty value and hence we would show an error.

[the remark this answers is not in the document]
sbalaji1192: Let's explicitly handle the empty value.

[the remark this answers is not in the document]
ashit-rath: Okay

sbalaji1192 (approved):
LGTM!

app/client/src/components/ads/DraggableListCard.tsx
```diff
@@ -101,6 +101,32 @@ export function DraggableListCard(props: RenderComponentProps) {
...
+        width={20}
+      />
+    ) : (
+      <StyledHiddenIcon
+        className="t--show-column-btn"
+        height={20}
+        onClick={() => {
+          setVisibility(!visibility);
+          toggleVisibility && toggleVisibility(index);
+        }}
```
Rishabh-Rathod, line 122: shouldn't we avoid each time re-creation of a function instead use `useCallback` here, this also makes sure the children won't re-render due to a change of the `onClick` function reference.
@ashit-rath

app/client/src/entities/DataTree/dataTreeWidget.ts
```diff
@@ -116,31 +116,48 @@ export const generateDataTreeWidget = (
...
-    propertyOverrideDependency,
-    overridingPropertyPaths,
-    bindingPaths,
-    triggerPaths,
-    validationPaths,
-    ENTITY_TYPE: ENTITY_TYPE.WIDGET,
-    privateWidgets: {
-      ...widget.privateWidgets,
+  /**
+   * Spread operator does not merge deep objects properly.
```
Rishabh-Rathod, line 120: The spread operator doesn't recursively merge objects, I believe that is known behavior and worked as expected. Did we face any issue for which we added lodash merge? 

@ashit-rath

ashit-rath: With the JSONForm, we started dealing we deeply nested objects and during an update, it failed to merge the old properties with the new ones.

app/client/src/components/ads/DraggableListCard.tsx
ashit-rath: I'm not 100% sure if that would give us a significant perf gain as this is just an icon component (pretty light). Do we see a perf drop here?

app/client/src/entities/DataTree/dataTreeWidget.ts
Rishabh-Rathod: Got it. Thanks.

app/client/src/widgets/JSONFormWidget/fields/BaseInputField.tsx
```diff
@@ -0,0 +1,387 @@
...
+
+      if (isNil(value)) {
+        isNilSetByField.current = true;
+      }
+
+      fieldOnChangeHandler(value);
+      setInputText(text);
+
+      if (onTextChanged && executeAction) {
+        executeAction({
```
Rishabh-Rathod, line 278: Any reason why we trigger `executeAction` onTextChange here asking as normally in input widget we trigger `updateWidgetMetaProperty` which triggers `onTextChange` with debounce.

ashit-rath: @Rishabh-Rathod If you see the Input widget we have a similar `onTextChange` event and the json form's input field replicates that behavior. This will only run if the user has set this action up in the property pane.

Regarding how the value updates, we use the `updateWidgetMetaProperty` but in a different way. You won't find it in this file though.

PR #8829, opened 2021-10-26
feat: remove region requirement from s3 plugin

### LINKED ISSUE #6313: [Feature] s3 plugin: auto fill the region field in datsource page
## Summary
From slack:

> QQ, when an S3 datasource is configured with the wrong region, and we execute a query with that datasource, we get the error saying “this is not the right region, this other one is the right region”, so it looks like we have information on what the right region is. Can we simplify this for the user by not asking for the region at all, and just infer it instead?

sumitsum:
## Description
- Remove `Region` field from S3 datasource editor page for AWS S3, Upcloud, Digital Ocean Spaces, Dream Objects, Wasabi. 
  - Use SDK provided property for AWS S3 to delegate region selection to the SDK. 
  - Extract region info from endpoint URL for Upcloud, Digital Ocean Spaces, Dream Objects and Wasabi, since the SDK property does not work for these service providers.
- Removed some redundant checks from `datasourceCreate` that were already part of `validateDatasource`
- Fix show clause in `list.json`

Fixes #6313, #8835

## Type of change

## How Has This Been Tested?
- JUnit TCs
- Manual

## Checklist:


app/server/appsmith-plugins/amazons3Plugin/src/main/resources/form.json
```diff
@@ -6,14 +6,14 @@
       "children": [
         {
           "label": "S3 Service Provider Key",
-          "configProperty": "datasourceConfiguration.properties[1].key",
+          "configProperty": "datasourceConfiguration.properties[0].key",
```
trishaanand, line 9: Could we leave the index untouched and unused to ensure that we dont have to run the migrations?

trishaanand:
What a great PR @sumitsum ! I loved how well the code has been commented. Leaving 1 minor comment to not do migrations unless required.

trishaanand (approved):

Issue #8846, opened 2021-10-27
[Bug]: Env variable changes don't reflect after changing from admin UI


sharat87:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The admin config UI is currently in `release` and we found an issue. Here’s how it works:

1. User makes a change in env variable in the UI.
2. Client sends the changes to server.
3. Server writes the changes to `docker.env` file.
4. Server calls upon Supervisor to restart the `backend`, `editor` and `rts`, by running `supervisorctl restart backend editor rts`.

This is fairly simple and works for the most part. The problem is with nginx configuration in the `editor` process. The environment variable placeholders in the nginx configuration are written in the entrypoint script, before supervisord takes control. So, since we only restart `backend`, `editor` and `rts`, the nginx configuration is not updated. For the nginx configuration to be updated, we have to restart the whole Docker container.

https://github.com/appsmithorg/appsmith/blob/4487fee83e2d99514f37883ec915931f759acb42/deploy/docker/entrypoint.sh#L123

To fix this, I’m thinking we need the `editor` supervisor process, should run a small script instead of directly running `nginx`, and that script will update the nginx config before starting nginx. Something like:

```bash
#!/usr/bin/env bash
cat /etc/nginx/conf.d/nginx_app.conf.template | envsubst "$(printf '$%s,' $(env | grep -Eo '^APPSMITH_[A-Z0-9_]+'))" | sed -e 's|\${\(APPSMITH_[A-Z0-9_]*\)}||g' > /etc/nginx/sites-available/default
exec nginx -g 'daemon off;'
```

But of course, it should work with or without SSL enabled, and we need to test how Supervisor behaves when the process is run like this. I’ve had trouble with supervisor when the processes start child processes, so I’m using `exec` in the example above. I’m _hoping_ supervisor will be okay with that, but need to check this.

[...]

PR #8922, opened 2021-10-29
feat: Config for editor program run script instead of nginx

geekup-legodevops:
## Description

- Add script to re-generate nginx config file & start nginx 
- Move function init_ssl_cert into separated file and reuse it in new script and entrypoint
- Change command config & add new config to kill all sub-process of a supervisord child program in editor program configure file

Fixes # (issue)

[[Bug]: Env variable changes don't reflect after changing from admin UI #8846](https://github.com/appsmithorg/appsmith/issues/8846)

## Type of change

## Checklist:


deploy/docker/scripts/run-nginx.sh
```diff
@@ -0,0 +1,26 @@
...
+echo "Generating nginx configuration"
+cat /etc/nginx/conf.d/nginx_app.conf.template | envsubst "$(printf '$%s,' $(env | grep -Eo '^APPSMITH_[A-Z0-9_]+'))" | sed -e 's|\${\(APPSMITH_[A-Z0-9_]*\)}||g' >/etc/nginx/sites-available/default
+
+# Check exist certificate with given custom domain
+if ! [[ -e "/etc/letsencrypt/live/$APPSMITH_CUSTOM_DOMAIN" ]]; then
+  source "/opt/appsmith/init_ssl_cert.sh"
+  init_ssl_cert "$APPSMITH_CUSTOM_DOMAIN"
+fi
+
+nginx -g "daemon off;"
```
sharat87, line 26: If we use `exec` here, perhaps there won't be a child process and we may not need the `killasgroup` option in supervisor config? Does that work?

```suggestion
exec nginx -g "daemon off;"
```

deploy/docker/scripts/run-nginx.sh
```diff
@@ -0,0 +1,26 @@
...
+echo "Re-generating nginx config template with domain"
+bash "/opt/appsmith/templates/nginx_app.conf.sh" "$NGINX_SSL_CMNT" "$APPSMITH_CUSTOM_DOMAIN" >"/etc/nginx/conf.d/nginx_app.conf.template"
+
+echo "Generating nginx configuration"
+cat /etc/nginx/conf.d/nginx_app.conf.template | envsubst "$(printf '$%s,' $(env | grep -Eo '^APPSMITH_[A-Z0-9_]+'))" | sed -e 's|\${\(APPSMITH_[A-Z0-9_]*\)}||g' >/etc/nginx/sites-available/default
+
+# Check exist certificate with given custom domain
+if ! [[ -e "/etc/letsencrypt/live/$APPSMITH_CUSTOM_DOMAIN" ]]; then
+  source "/opt/appsmith/init_ssl_cert.sh"
+  init_ssl_cert "$APPSMITH_CUSTOM_DOMAIN"
```
sharat87, line 23: Is there a reason we are doing `init_ssl_cert` both in the entrypoint script as well as when starting nginx itself?

deploy/docker/scripts/run-nginx.sh
```diff
@@ -0,0 +1,26 @@
...
+fi
+
+echo "Re-generating nginx config template"
+bash "/opt/appsmith/templates/nginx_app.conf.sh" "$NGINX_SSL_CMNT" "$APPSMITH_CUSTOM_DOMAIN" >"/etc/nginx/conf.d/nginx_app.conf.template"
+
+echo "Generating nginx configuration"
+cat /etc/nginx/conf.d/nginx_app.conf.template | envsubst "$(printf '$%s,' $(env | grep -Eo '^APPSMITH_[A-Z0-9_]+'))" | sed -e 's|\${\(APPSMITH_[A-Z0-9_]*\)}||g' >/etc/nginx/sites-available/default
+
+# Check exist certificate with given custom domain
+if ! [[ -e "/etc/letsencrypt/live/$APPSMITH_CUSTOM_DOMAIN" ]]; then
```
sharat87, line 21: Can we check if custom domain is present as well here? Like on line 10?

deploy/docker/entrypoint.sh
```diff
@@ -3,196 +3,125 @@
...
-	unset APPSMITH_RECAPTCHA_ENABLED
+  unset APPSMITH_RECAPTCHA_SITE_KEY # If this field is empty is might cause application crash
+  unset APPSMITH_RECAPTCHA_SECRET_KEY
+  unset APPSMITH_RECAPTCHA_ENABLED
 fi
 
 # Main Section
 init_mongodb
-configure_ssl
+mount_letsencrypt_folder
```
sharat87, line 124: Looks like the name is `mount_letsencrypt_directory` in the definition. Or is this function defined somewhere else?

deploy/docker/scripts/init_ssl_cert.sh
```diff
@@ -0,0 +1,59 @@
...
+
+  echo "Re-generating nginx config template with domain"
+  bash "/opt/appsmith/templates/nginx_app.conf.sh" "$NGINX_SSL_CMNT" "$APPSMITH_CUSTOM_DOMAIN" >"/etc/nginx/conf.d/nginx_app.conf.template"
+
+  echo "Generating nginx configuration"
+  cat /etc/nginx/conf.d/nginx_app.conf.template | envsubst "$(printf '$%s,' $(env | grep -Eo '^APPSMITH_[A-Z0-9_]+'))" | sed -e 's|\${\(APPSMITH_[A-Z0-9_]*\)}||g' >/etc/nginx/sites-available/default
+
+  local live_path="/etc/letsencrypt/live/$domain"
+  if [[ -e "$live_path" ]]; then
+    echo "Existing certificate for domain $domain"
```
sharat87, line 29: If we are not generating the certificate, let's also not have the cron job to renew the cert either. Can we disable that part of the process as well?

geekup-legodevops: Hi @sharat87 Actually, this is the auto-generated certificate using Let's Encrypt which is generated by container and in the control. We think we should have the cron job to renew certificate for this one and only disable the cron job in case of user use custom certificate (another PR). Is it okay ?

sharat87 (approved):

sharat87 (approved):

PR #8930, opened 2021-10-31
test: Make Cypress tests resilient to changes in app texts

### LINKED ISSUE #8389: [Task] Appsmith applicaiton Copy

### SubTasks

- [ ] Migrate orphaned messages from the code base into `messages.ts`
- [ ] Create static list of messages which can be used by `cypress` and the Appsmith application in isolation.
- [ ] Consider the feasibility of translations using this list of messages and make changes (optional)
- [ ] Utilise the static messages in cypress to achieve one source of messages in the Appsmith codebase

rafaeelaudibert:
We are now resorting to fetching the texts from `constants.ts` instead of hardcoding the messages expected from the app.

Fixes #8389

## Checklist:

### DIFF
--- a/app/client/src/widgets/FilePickerWidgetV2/component/index.tsx
+++ b/app/client/src/widgets/FilePickerWidgetV2/component/index.tsx
@@ -5,6 +5,7 @@ import "@uppy/dashboard/dist/style.css";
 import "@uppy/webcam/dist/style.css";
 import { BaseButton } from "widgets/ButtonWidget/component";
 import { Colors } from "../../../constants/Colors";
+import { FILEPICKER_WIDGET_FILES_SELECTED } from "constants/messages";
 
 class FilePickerComponent extends React.Component<
   FilePickerComponentProps,
@@ -26,7 +27,7 @@ class FilePickerComponent extends React.Component<
   render() {
     let label = this.props.label;
     if (this.props.files && this.props.files.length) {
-      label = `${this.props.files.length} files selected`;
+      label = FILEPICKER_WIDGET_FILES_SELECTED(this.props.files.length);
     }
     return (
       <BaseButton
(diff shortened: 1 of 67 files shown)


app/client/src/widgets/FilePickerWidgetV2/component/index.tsx
```diff
@@ -26,7 +27,7 @@ class FilePickerComponent extends React.Component<
   render() {
     let label = this.props.label;
     if (this.props.files && this.props.files.length) {
-      label = `${this.props.files.length} files selected`;
+      label = FILEPICKER_WIDGET_FILES_SELECTED(this.props.files.length);
```
mohanarpit, line 30: Minor: Instead of invoking these constant functions directly in the source code, can we please replace it with `createMessage(FILEPICKER_WIDGET_FILES_SELECTED, this.props.files.length)`

I'm requesting this because when we implement internationalization later on, only the function `createMessage` needs to be modified a little. Nothing else in the codebase will be required to change.

mohanarpit:
@rafaeelaudibert This is an awesome change! Thank you so much 🙏🏽 

This will make our test suite a lot more resilient to text changes that happen often & improve the quality of our code as well.

mohanarpit:
Test result: https://github.com/appsmithorg/appsmith/runs/4058552736

rafaeelaudibert:
@mohanarpit Hmm, so that's the intention behind `createMessage`, thanks for clarifying it! How would we go about using it on Cypress, though? Should we just call the same `createMessage` function too?

rafaeelaudibert:
> Test result: https://github.com/appsmithorg/appsmith/runs/4058552736

Also, just pushed a commit to fix this :).

mohanarpit:
/ok-to-test sha=5e7be04

> @mohanarpit Hmm, so that's the intention behind `createMessage`, thanks for clarifying it! How would we go about using it on Cypress, though? Should we just call the same `createMessage` function too?

Yes please. We can invoke the same function `createMessage` from the Cypress tests as well. That should work well.

mohanarpit:
Test result: https://github.com/appsmithorg/appsmith/actions/runs/1405572962

sbalaji1192:
Reopening as this is important.

@rafaeelaudibert Could you please make the requested changes above?

PR #8986, opened 2021-11-02
fix: Added support for multiple files in multipart REST

### LINKED ISSUE #8985: [Feature]: Add support for multiple files in the same key for multipart REST APIs

### Summary

Users would like to attach multiple files to the same key in multipart file uploads

We're currently only accepting one file per key and the spec is not limited to such behaviour.

nidhi-nair:
## Description
This fix retains support for single uploads as well as adds support for an array of files. This means that existing actions do not have to be migrated and future actions continue to have the option of using an array of files or just a single file.

Fixes #8985 

## Type of change

## How Has This Been Tested?
- Tested manually
- JUnit case

## Checklist:

### DIFF
--- a/app/server/appsmith-plugins/restApiPlugin/src/main/java/com/external/helpers/DataUtils.java
+++ b/app/server/appsmith-plugins/restApiPlugin/src/main/java/com/external/helpers/DataUtils.java
@@ -155,6 +155,7 @@ public class DataUtils {
                             try {
                                 populateFileTypeBodyBuilder(bodyBuilder, property, outputMessage);
                             } catch (JsonProcessingException e) {
+                                e.printStackTrace();
                                 throw new AppsmithPluginException(
                                         AppsmithPluginError.PLUGIN_DATASOURCE_ARGUMENT_ERROR,
                                         "Unable to parse content. Expected to receive an array or object of multipart data"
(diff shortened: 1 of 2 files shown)


app/server/appsmith-plugins/restApiPlugin/src/main/java/com/external/helpers/DataUtils.java
```diff
@@ -150,26 +152,11 @@ public String parseFormData(List<Property> bodyFormData, Boolean encodeParamsTog
...
                         } else if (MultipartFormDataType.FILE.equals(multipartFormDataType)) {
-
-                            MultipartFormDataDTO multipartFormDataDTO = null;
                             try {
-                                multipartFormDataDTO = objectMapper.readValue(
-                                        String.valueOf(property.getValue()),
-                                        MultipartFormDataDTO.class);
+                                populateFileTypeBodyBuilder(bodyBuilder, property, outputMessage);
                             } catch (JsonProcessingException e) {
                                 e.printStackTrace();
```
sumitsum, line 158: Hey sorry for this, I know this is not a part of this change, still since it is related I wanted to bring this up - the try catch here silently suppresses the exception from reading file data. Can we do something else here ? - like maybe return an `AppsmithExeception` or a warning message to the client that the file upload has failed ?

sumitsum (approved):

PR #8989, opened 2021-11-03
chore: Removes incorrect trigger paths from list widget via database migrations

trishaanand:


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
```diff
@@ -3631,4 +3647,161 @@ public void setSlugToApplicationAndPage(MongockTemplate mongockTemplate) {
...
+    private DslUpdateDto updateListWidgetTriggerPaths(DslUpdateDto dslUpdateDto) {
+        JSONObject dsl = dslUpdateDto.getDsl();
+        Boolean updated = dslUpdateDto.getUpdated();
+
+        if (dsl == null) {
+            // This isn't a valid widget configuration. No need to traverse this.
+            return dslUpdateDto;
+        }
+
+        String widgetType = dsl.getAsString(FieldName.WIDGET_TYPE);
```
sumitsum, line 3660: I encountered a `null` value for `wigetType` identifier when `dsl` is:
```
{
  "key": "dsl",
  "value": {
    "name": "MainContainer",
    ...... 
    "type": ....
  }
}
```

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
```diff
@@ -3631,4 +3647,161 @@ public void setSlugToApplicationAndPage(MongockTemplate mongockTemplate) {
...
+                    dslUpdateDto = updateListWidgetTriggerPaths(new DslUpdateDto(layout.getDsl(), dslUpdateDto.getUpdated()));
+                    layout.setDsl(dslUpdateDto.getDsl());
+                }
+            }
+
+            if (dslUpdateDto.getUpdated().equals(TRUE)) {
+                mongockTemplate.save(page);
+            }
+        }
+    }
```
sumitsum, line 3806: QQ: This migration change seems to remove the trigger paths which are no longer valid. Is it because the widget has changed ? If so, then what would happen to the JS code written in the older trigger paths ?

sumitsum (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
```diff
@@ -3631,4 +3647,161 @@ public void setSlugToApplicationAndPage(MongockTemplate mongockTemplate) {
             );
         }
     }
+
+    private DslUpdateDto updateListWidgetTriggerPaths(DslUpdateDto dslUpdateDto) {
+        JSONObject dsl = dslUpdateDto.getDsl();
+        Boolean updated = dslUpdateDto.getUpdated();
```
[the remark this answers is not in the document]
trishaanand: Quoting from line 3802 :
```
            if (dslUpdateDto.getUpdated().equals(TRUE)) {
                mongockTemplate.save(page);
            }

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
trishaanand: Is that a valid DSL?  And in case of invalid DSL, and there is no widget type, the check `if (widgetType.equals("LIST_WIDGET"))` ensures that we dont do anything for such a scenario/

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
trishaanand: We are removing trigger paths which either dont exist or no longer have JS code written in it. This is because of a client bug where they werent computing the dynamic trigger paths correctly. The client has fixed the bug for future edits. This ensures that the existing pages have correct paths without requiring any new edits.

sumitsum (approved):

PR #8993, opened 2021-11-03
chore: Adding evaluation version in application for new evaluation logic.

### LINKED ISSUE #2862: [Bug] Modify dynamic binding evaluation to support escaped characters without double escapes
Escape characters (like `\t`, `\n` etc) are currently not read as such during evaluation. They need to be written as a double escaped `\\n` for the character to make sense while evaluation. This is non-intuitive since the editor space is not represented as a string literal.

trishaanand:
Fixes #2862

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ApplicationPageServiceImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ApplicationPageServiceImpl.java
@@ -78,6 +78,8 @@ public class ApplicationPageServiceImpl implements ApplicationPageService {
     private final GitFileUtils gitFileUtils;
     private final CommentThreadRepository commentThreadRepository;
 
+    public static final Integer EVALUATION_VERSION = 2;
+
     public Mono<PageDTO> createPage(PageDTO page) {
         if (page.getId() != null) {
             return Mono.error(new AppsmithException(AppsmithError.INVALID_PARAMETER, FieldName.ID));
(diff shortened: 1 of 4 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ApplicationPageServiceImpl.java
```diff
@@ -247,6 +249,9 @@ private Boolean isDuplicatePage(Application application, String pageId) {
 
         application.setPublishedPages(new ArrayList<>());
 
+        // For all new applications being created, set it to use the latest evaluation version.
+        application.setEvaluationVersion(EVALUATION_VERSION);
```
nidhi-nair, line 253: @trishaanand uneducated question, is this flow also used for clone applications, fork applications, import/export etc?

trishaanand: Aah yes. Great catch! Let me look at the code and figure out how to differentiate between these flows.

trishaanand: Checked. We are good. All of these flows manually handle the objects for the application and do not use the create application flow <phew> @nidhi-nair

trishaanand:
Test run at : https://github.com/appsmithorg/appsmith/runs/4089854695?check_suite_focus=true

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ApplicationPageServiceImpl.java
nidhi-nair: Got it, so then version null will be duplicated as version null and version 2 will be duplicated as version 2 automatically?

trishaanand: Thats correct.

nidhi-nair (approved):

PR #9021, opened 2021-11-06
fix: Add support for content type in create file for S3 plugin

### LINKED ISSUE #7308: [Feature] Ability to set the content-type when uploading a file to S3
## Summary

Currently, when uploading a file to S3 from the S3 plugin, we cannot set an explicit content-type, and so all files are uploaded as the content-type `application/octet-stream`. This causes problems because when we upload an SVG file, accessing the URL will download the file instead of showing the image, because the `Content-Type` header is `application/octet-stream` instead of `image/svg+xml`.

If we were able to set the content type when uploading, we could've set it to `image/svg+xml` and so the file's URL would be working right away. Currently, we have go to S3 console and update the content type of these files before we can use them.

## Additional Context

On [this line in the code](https://github.com/appsmithorg/appsmith/blob/6061206970e8b7fb405b2775922e767c81023bf6/app/server/appsmith-plugins/amazons3Plugin/src/main/java/com/external/plugins/AmazonS3Plugin.java#L226), we are just passing an empty `ObjectMetadata` object. This object has a `setContentType` method which we need to call with `"image/svg+xml"` before passing it on.

nidhi-nair:
Fixes #7308 

## Type of change

## How Has This Been Tested?
- JUnit test case
- Manually tested


app/server/appsmith-plugins/amazons3Plugin/src/main/java/com/external/plugins/AmazonS3Plugin.java
```diff
@@ -196,15 +197,28 @@ String uploadFileFromBody(AmazonS3 connection,
                 throws InterruptedException, AppsmithPluginException {
 
             byte[] payload;
+            MultipartFormDataDTO multipartFormDataDTO = null;
```
sumitsum, line 200: I think there is one corner case that is not handled here that we have been supporting so far i.e. when user wants to upload a file - not via `filepicker` but by writing the content in the body. In this case the body content will not map to a `MultipartFormDataDTO` object but will remain as a normal string. Whether `filepicker` is used or not should be decided by the boolean param `usingFilePicker`.

app/server/appsmith-plugins/amazons3Plugin/src/test/java/com/external/plugins/AmazonS3PluginTest.java
```diff
@@ -302,7 +302,7 @@ public void testCreateFileFromBodyWithFalseCredentialsAndNonNullDuration() {
         AmazonS3Plugin.S3PluginExecutor pluginExecutor = new AmazonS3Plugin.S3PluginExecutor();
 
         ActionConfiguration actionConfiguration = new ActionConfiguration();
-        String dummyBody = "";
+        String dummyBody = "{\"data\": \"\"}";
```
sumitsum, line 305: Could be related to the first comment.

app/server/appsmith-plugins/amazons3Plugin/src/test/java/com/external/plugins/AmazonS3PluginTest.java
```diff
@@ -345,7 +345,7 @@ public void testFileUploadFromBodyWithMissingDuration() {
...
```
sumitsum, line 348: Could be related to the first comment.

app/server/appsmith-plugins/amazons3Plugin/src/test/java/com/external/plugins/AmazonS3PluginTest.java
```diff
@@ -375,14 +375,63 @@ public void testFileUploadFromBodyWithMissingDuration() {
...
+        actionConfiguration.setBody(dummyBody);
+
+        String dummyPath = "path";
+        actionConfiguration.setPath(dummyPath);
+
+        Map<String, Object> configMap = new HashMap<>();
+        setValueSafelyInFormData(configMap, COMMAND, "UPLOAD_FILE_FROM_BODY");
+        setValueSafelyInFormData(configMap, BUCKET, "bucket_name");
+        setValueSafelyInFormData(configMap, CREATE_DATATYPE, "YES");
+        setValueSafelyInFormData(configMap, CREATE_EXPIRY, "100000");
```
sumitsum, line 394: I think we should also set the `using file picker` attribute to `yes` here.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
```diff
@@ -3631,4 +3631,48 @@ public void setSlugToApplicationAndPage(MongockTemplate mongockTemplate) {
             );
         }
     }
+
+    @ChangeSet(order = "095", id = "update-s3-action-configuration-for-type", author = "")
```
sumitsum, line 3635: Please add a comment explaining why this migration is done and what it does.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
```diff
@@ -3631,4 +3631,48 @@ public void setSlugToApplicationAndPage(MongockTemplate mongockTemplate) {
...
+    public void updateS3ActionConfigurationForType(MongockTemplate mongockTemplate) {
```
sumitsum, line 3636: Can you please update the method name as it is a bit confusing ?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
```diff
@@ -3631,4 +3631,48 @@ public void setSlugToApplicationAndPage(MongockTemplate mongockTemplate) {
...
+        // Find all S3 actions
```
sumitsum, line 3637: Can you please move this comment to the next line ? - since the current code fetchs S3 plugin object.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
```diff
@@ -3631,4 +3631,48 @@ public void setSlugToApplicationAndPage(MongockTemplate mongockTemplate) {
...
+            // First migrate the plugin specified templates to form data
+            ActionDTO unpublishedAction = s3Action.getUnpublishedAction();
+
+            if (unpublishedAction == null || unpublishedAction.getActionConfiguration() == null) {
+                // No migrations required
+                continue;
+            }
+
+            final String oldUnpublishedBody = unpublishedAction.getActionConfiguration().getBody();
+            final String newUnpublishedBody = "{\n\t\"data\": \"" + oldUnpublishedBody + "\"\n}";
```
sumitsum, line 3661: Should we check for `using filepicker` attribute value as well ? - please check the first comment for context.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
```diff
@@ -3631,4 +3631,48 @@ public void setSlugToApplicationAndPage(MongockTemplate mongockTemplate) {
...
+                NewAction.class
+        );
+
+        List<NewAction> actionsToSave = new ArrayList<>();
+
+        for (NewAction s3Action : s3Actions) {
+            // First migrate the plugin specified templates to form data
+            ActionDTO unpublishedAction = s3Action.getUnpublishedAction();
+
+            if (unpublishedAction == null || unpublishedAction.getActionConfiguration() == null) {
```
sumitsum, line 3655: not sure about this - should we check for `publishedAction` even if the `unpublishedAction` does not exist ?

sumitsum (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
nidhi-nair: Published actions would never exist if unpublished actions do not. It is mandated by our creation and publish flow.

app/server/appsmith-plugins/amazons3Plugin/src/main/java/com/external/plugins/AmazonS3Plugin.java
nidhi-nair: @sumitsum I've maintained the same type so that users are made to manually create the filepicker-like objet even when they don't use the file picker. This makes sure that they are using the content type functionality. And even if they don't, the minimum criteria is for their data to be in a prop called data. It makes for similarity in behaviour.

app/server/appsmith-plugins/amazons3Plugin/src/test/java/com/external/plugins/AmazonS3PluginTest.java
nidhi-nair: @sumitsum Sorry, but I did not understand this

sumitsum (dismissed):

sumitsum:
Tracking the required documentation changes here: https://github.com/appsmithorg/appsmith/issues/9056

sumitsum (approved):

nidhi-nair:
https://github.com/appsmithorg/appsmith/actions/runs/1439308229

PR #9051, opened 2021-11-09
feat: UQI where clause support

trishaanand:


trishaanand:
Test run to be found here : https://github.com/appsmithorg/appsmith/actions/runs/1443296775

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/FilterDataService.java
```diff
@@ -118,14 +118,104 @@ public ArrayNode filterData(ArrayNode items, List<Condition> conditionList) {
         return finalResultsNode;
     }
 
-    public List<Map<String, Object>> executeFilterQuery(String tableName, List<Condition> conditions, Map<String, DataType> schema) {
+    public ArrayNode filterDataNew(ArrayNode items, Condition condition) {
+
+        if (items == null || items.size() == 0) {
+            return items;
```
nidhi-nair, line 124: If it's null do you want to return an empty array instead?

app/server/appsmith-interfaces/src/test/java/com/appsmith/external/helpers/PluginUtilsTest.java
```diff
@@ -0,0 +1,95 @@
...
+import static com.appsmith.external.helpers.PluginUtils.parseWhereClause;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class PluginUtilsTest {
+
+    private final ObjectMapper objectMapper = new ObjectMapper();
+
+    @Test
+    public void parseWhereClauseTest() {
+        String whereJson = "{\n" +
```
nidhi-nair, line 22: Shouldn't the test case have a parent conjunction operator here?

nidhi-nair, line 22: Some test cases to add to check for behaviour:
1. In with empty array and array with values
2. Any of the conditions with type mismatch
3. Any of the conditions with evaluated value evaluating to null
4. Different data types (especially date)

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/constants/ConditionalOperator.java
```diff
@@ -61,4 +61,16 @@ public String toString() {
             return "not in";
         }
     },
+    AND {
```
sumitsum, line 64: Should we create different enum type for `AND` and `OR` operators - since these are logical operators where as the rest are comparison operators. This may help us avoid errors of the kind where we might end up using a logical operator in place of a comparison operator and vice-versa.

app/server/appsmith-plugins/amazons3Plugin/src/main/java/com/external/plugins/AmazonS3Plugin.java
```diff
@@ -535,6 +548,17 @@ String readFile(AmazonS3 connection, String bucketName, String path, Boolean enc
                                     null, null));
                         }
 
+                        // Check if where condition is configured
+                        Object whereFormObject = getValueSafelyFromFormData(formData, LIST_WHERE);
+
+                        if (whereFormObject != null) {
```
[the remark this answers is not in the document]
trishaanand: An object can't be used with the above. Resolving this.

app/server/appsmith-plugins/amazons3Plugin/src/main/resources/editor/list.json
```diff
@@ -29,6 +29,52 @@
...
+            },
+            {
+              "label": "OR",
+              "value": "OR"
+            }
+          ],
+          "comparisonTypes": [
+            {
+              "label": "==",
+              "value": "EQ"
```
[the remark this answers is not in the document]
trishaanand: Thanks! That makes sense. I have removed operators like "<" ">", etc as well since they dont make sense for file names. Resolving this!

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/PluginUtils.java
```diff
@@ -197,4 +204,35 @@ else if(!CollectionUtils.isEmpty(datasourceConfiguration.getEndpoints())) {
 
         return message;
     }
+
+    public static Condition parseWhereClause(Map<String, Object> whereClause) {
+        Condition condition = new Condition();
+
+        Object unparsedOperator = whereClause.getOrDefault(CONDITION, ConditionalOperator.EQ.toString());
+        ConditionalOperator operator = ConditionalOperator.valueOf(((String) unparsedOperator).trim().toUpperCase());
```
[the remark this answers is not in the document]
trishaanand: Good catch! Added.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/PluginUtils.java
```diff
@@ -197,4 +204,35 @@ else if(!CollectionUtils.isEmpty(datasourceConfiguration.getEndpoints())) {
...
+
+        if (operator != null) {
```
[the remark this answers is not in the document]
trishaanand: Have added catching of the exception if there is no match for the enum. The code would exit before reaching here. Resolving this!

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/constants/ConditionalOperator.java
trishaanand: @sumitsum : Would it be possible for you to take this up in a refactor post this?

sumitsum: Sure. Resolving this for now.

app/server/appsmith-plugins/amazons3Plugin/src/main/java/com/external/plugins/AmazonS3Plugin.java
```diff
@@ -100,6 +107,12 @@ public AmazonS3Plugin(PluginWrapper wrapper) {
     @Extension
     public static class S3PluginExecutor implements PluginExecutor<AmazonS3>, SmartSubstitutionInterface {
         private final Scheduler scheduler = Schedulers.elastic();
+        private ObjectMapper objectMapper = new ObjectMapper();
```
sumitsum, line 110: Can we re-use the `objectMapper` defined in `BasePlugin` class ?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/FilterDataService.java
```diff
@@ -118,14 +119,104 @@ public ArrayNode filterData(ArrayNode items, List<Condition> conditionList) {
...
+        // Now that the data has been filtered. Clean Up. Drop the table
+        dropTable(tableName);
+
+        ArrayNode finalResultsNode = objectMapper.valueToTree(finalResults);
+
+        return finalResultsNode;
+    }
+
+    private List<Map<String, Object>> executeFilterQueryNew(String tableName, Condition condition, Map<String, DataType> schema) {
+        Connection conn = checkAndGetConnection();
```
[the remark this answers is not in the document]
trishaanand: We only have one connection at a time. `checkAndGetConnection` checks for the validity of the connection and returns the same (or creates a new one and then returns that one). 
I agree that connection pool is a good direction to take. What you bring up is a valid concern which could cause performance degradation without a connection pool.

Can we add an issue for this task and take it up?

app/server/appsmith-plugins/amazons3Plugin/src/main/java/com/external/plugins/AmazonS3Plugin.java
trishaanand: Good catch! Removing this!

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/Condition.java
```diff
@@ -45,19 +45,39 @@ public Condition(String path, String operator, String value) {
...
+                    .map(subCondition -> addValueDataType(subCondition))
+                    .collect(Collectors.toList());
+            condition.setValue(updatedConditions);
+        }
+        return condition;
+    }
+
     public static Boolean isValid(Condition condition) {
 
         if (StringUtils.isEmpty(condition.getPath()) ||
```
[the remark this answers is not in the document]
trishaanand: Good catch! Thought this was more to ensure that NPE are not thrown. Anyways this code would not be hit in UQI flows and once we migrate GSheets to UQI, we can remove this flow completely.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/FilterDataService.java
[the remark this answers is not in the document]
sumitsum: Sure, resolving this for now. I will create an issue for this and add it to UQI epic.

sumitsum (approved):

app/server/appsmith-interfaces/src/test/java/com/appsmith/external/helpers/PluginUtilsTest.java
trishaanand: Added these test cases

sumitsum (approved):

PR #9103, opened 2021-11-11
ci: Remove unused package for DP step

sharat87:
This step in the workflow would actually never run because of the way the workflow is triggered and the conditions set on it. So this PR removes this unused step.


nidhi-nair:
QQ: How are we handling package for DP in that case @sharat87 ?

sharat87:
@nidhi-nair CodeBuild 😎

nidhi-nair (approved):

PR #9104, opened 2021-11-11
chore: Let the server start automatically on boot

sharat87:
This change will ensure the Appsmith server container and watchtower container will start automatically if the EC2 instance is rebooted.


mohanarpit (approved):

mohanarpit:
@sharat87 I'm merging this in. But we'll have to create a new AWS AMI for this to take effect right? 

@geekup-legodevops Can you please take this into consideration & publish a new AMI?

sharat87:
@mohanarpit, I just created a new instance with an AMI and the `restart: unless-stopped` is already there. This is already done on AMI :)

PR #9127, opened 2021-11-12
ci: Refactor to deploy Heroku using new Docker image

geekup-legodevops:
## Description

- Refactor entrypoint script in Docker image to deploy Appsmith on Heroku
- Add environment variables into docker.env if exist, use default value if variable not exist
- Remove unnecessary resource in Heroku folder
- Add sub-path for supervisor process UI

## Type of change

## Checklist:


deploy/docker/entrypoint.sh
```diff
@@ -2,197 +2,227 @@
...
+  unset APPSMITH_RECAPTCHA_SECRET_KEY
+  unset APPSMITH_RECAPTCHA_ENABLED
 fi
 
+echo "Generating Basic Authentication file"
+printf "$APPSMITH_BASIC_AUTH_USER:$(openssl passwd -apr1 $APPSMITH_BASIC_AUTH_PASSWORD)" > /etc/nginx/passwords
+
 # Main Section
-init_mongodb
+if [[ -z $DYNO ]]; then
```
sharat87, line 220:
```suggestion
if [[ -z $DYNO ]]; then
  # Don't run MongoDB if running in a Heroku dyno.
```

deploy/docker/templates/nginx_app.conf.sh
```diff
@@ -7,67 +7,93 @@ CUSTOM_DOMAIN="$2"
...
+    sub_filter __APPSMITH_INTERCOM_APP_ID__ '\${APPSMITH_INTERCOM_APP_ID}';
+    sub_filter __APPSMITH_MAIL_ENABLED__ '\${APPSMITH_MAIL_ENABLED}';
+    sub_filter __APPSMITH_DISABLE_TELEMETRY__ '\${APPSMITH_DISABLE_TELEMETRY}';
+    sub_filter __APPSMITH_RECAPTCHA_SITE_KEY__ '\${APPSMITH_RECAPTCHA_SITE_KEY}';
+    sub_filter __APPSMITH_RECAPTCHA_SECRET_KEY__ '\${APPSMITH_RECAPTCHA_SECRET_KEY}';
+    sub_filter __APPSMITH_RECAPTCHA_ENABLED__ '\${APPSMITH_RECAPTCHA_ENABLED}';
+
+    if (\$http_referer ~ "^.*/supervisor"){
+      return 301 /supervisor/\$request_uri;
+    }
```
sharat87, line 71: Can you add a comment on what this block is doing here please? Isn't the `/supervisor/` route defined above be sufficient? If it is needed, should we add it to https configuration also?

deploy/docker/templates/nginx_app.conf.sh
```diff
@@ -7,67 +7,93 @@ CUSTOM_DOMAIN="$2"
...
+    sub_filter __APPSMITH_RECAPTCHA_ENABLED__ '\${APPSMITH_RECAPTCHA_ENABLED}';
+
+    if (\$http_referer ~ "^.*/supervisor"){
+      return 301 /supervisor/\$request_uri;
+    }	
+  }
+
+  location /f {
+    proxy_pass https://cdn.optimizely.com/;
+  }
```
sharat87, line 76: This is deprecated and unused. Not needed today.

```suggestion
```

deploy/docker/entrypoint.sh
```diff
@@ -2,197 +2,227 @@
...
+  init_mongodb
+fi
 configure_ssl
+get_maximum_heap
+setup_backend_cmd
```
sharat87, line 225: Can you describe or add a comment on the reason we need to set the heap size here? I'm guessing it's a Heroku specific thing, but a comment would be nice to clarify that.

deploy/docker/templates/nginx_app.conf.sh
```diff
@@ -7,67 +7,93 @@ CUSTOM_DOMAIN="$2"
...
+  location = /supervisor {
+    return 301 /supervisor/;
+  }
+
+  location /supervisor/ {
+    proxy_http_version 1.1;
+    proxy_buffering     off;
+    proxy_max_temp_file_size 0;
+    proxy_redirect     off;
+    proxy_set_header   Host             \$http_host/supervisor/;
```
sharat87, line 32: The `Host` header shouldn't contain the path right? The `/supervisor/` part? Shouldn't this be just `\$http_host;` instead? Like:

```suggestion
    proxy_set_header   Host             \$http_host;
```

geekup-legodevops: Hi @sharat87 this is configuration for setting host in the control request (action like stopping process, restarting process,...) from Supervisor UI. When performing the action, a request will be sent to supervisor to perform. Then, it return a URL to reload the UI. There is a case that if user use another port rather than 80, the `http_host` variable leads to the missing of the port in the Host in the header of request and it route user to the the wrong URL.

sharat87: Yeah, I understand the port path, which is why we are using `$http_host` instead of `$host`, since `$host` doesn't have port in it.

My question is regarding the _path_. The `/supervisor/` part. Host header value should never have the path. See the Host header syntax at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Host#syntax, it can only contain the hostname and the port.

deploy/docker/templates/nginx/nginx-app-https.conf.template.sh
```diff
@@ -50,7 +68,7 @@ server {
   }
 
   proxy_set_header X-Forwarded-Proto \$origin_scheme;
-  proxy_set_header X-Forwarded-Host \$host;
+  proxy_set_header X-Forwarded-Host \$http_host;
```
sharat87, line 71: Same as above. Can we move this to inside the `location /supervisor/` block?

deploy/docker/entrypoint.sh
```diff
@@ -162,28 +186,41 @@ configure_supervisord() {
...
-init_replica_set
+if [[ -z "${DYNO}" ]]; then
+  # Don't run MongoDB if running in a Heroku dyno.
+  init_mongodb
+  init_replica_set
+fi
 mount_letsencrypt_directory
+# These functions are used to limit heap size for Backend process when deployed on Heroku
+get_maximum_heap
+setup_backend_cmd
```
sharat87, line 216: Should we not wrap these two lines inside the `if [[ -z $DYNO ]]; then` block as well? We don't need to do this when not running on Heroku, right?

geekup-legodevops: For this one, I think we will need to run these 2 functions in both case because these are taking responsibility in processing the command for Java backend process which will be used in the supervisord backend process config

sharat87 (approved):

PR #9166, opened 2021-11-15
fix: add migration for application going public event it has isPublic…

### LINKED ISSUE #8999: [Bug]: App is publicly accessible even when the toggle is turned off during app share

### Current Behavior

App is still publicly accessible when toggle is turned off during app share. Tried to repro this on a new app and this worked  as expected (see video). Only this [old app of mine](https://app.appsmith.com/applications/615c82c9ea18372f05103a0c/pages/616717781e01bd5e20880d55) shows up publicly.
 (https://app.appsmith.com/applications/615c82c9ea18372f05103a0c/pages/616717781e01bd5e20880d55)

[LOOM DEMO](https://www.loom.com/share/e4989e4e32d14e31b1e8beadb599a26a)

### Steps To Reproduce

1. Get publicly available app link from **Share**
2. Disable the _public app_ toggle and the app should display a login page - which usually happens but this month old app still displays all content

### Environment

Production

### Version

Cloud

nayan-rafiq:
## Description
Added migration to fix the issue: applications are publicly accessible event it has isPublic=false set

Fixes #8999 

## Type of change

## How Has This Been Tested?

- Manually

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
@@ -3854,4 +3854,24 @@ public class DatabaseChangelog {
             mongockTemplate.save(s3Action);
         }
     }
+
+    /**
+     * This migration fixes the data due to issue #8999 Due to this bug, public applications have isPublic=false
+     * when they are edited but in policies anonymousUser still have read application permission.
+     * This migration will set isPublic=true to those applications which have isPublic=false but anonymousUser has
+     * read:applications permission in policies
+     * @param mongockTemplate
+     */
+    @ChangeSet(order = "097", id = "fix-ispublic-is-false-for-public-apps", author = "")
+    public void fixIsPublicIsSetFalseWhenAppIsPublic(MongockTemplate mongockTemplate) {
+        Query query = query(
+                where("isPublic").is(false)
+                        .and("deleted").is(false)
+                        .and("policies").elemMatch(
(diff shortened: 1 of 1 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
```diff
@@ -3854,4 +3854,24 @@ public void updateS3ActionConfigurationBodyForContentTypeSupport(MongockTemplate
...
+    @ChangeSet(order = "097", id = "fix-ispublic-is-false-for-public-apps", author = "")
+    public void fixIsPublicIsSetFalseWhenAppIsPublic(MongockTemplate mongockTemplate) {
+        Query query = query(
+                where("isPublic").is(false)
+                        .and("deleted").is(false)
+                        .and("policies").elemMatch(
+                                where("permission").is("read:applications").and("users").is("anonymousUser")
+                        )
+        );
+        Update update = new Update().set("isPublic", true);
```
mohanarpit, line 3874: Should we be removing the access for the `anonymousUser` instead of marking `isPublic=true` here? I'm worried about private applications suddenly being marked as public.

nayan-rafiq: No, those applications are actually public. Users have made them public by themselves. Otherwise the policies would not have anonymous users. 
We've set `isPublic=false` as part of the bug so we should reset the flag again.

mohanarpit (approved):

PR #9177, opened 2021-11-16
chore: Update Validations

riodeuno:
## Description

Fixes # (issue)

- [x] If number value is out of bounds, return minmax || parsed || 0
- [x] If text has regex, match RegExp - earlier we were matching with
  string, which was always failing
- [x] Prioritise  expected  value in TEXT validation RegExp case.

## Type of change

- Bug fix (breaking change which fixes an issue)

## How Has This Been Tested?

Unit tests added

## QA notes

- Maps, Charts, Table validations should work correctly (out of bounds value should provide the nearest bound value)
- Response for regex based validations should be readable - iframe, video

## Checklist:

    // Code coverage diff between base branch:release and head branch: fix/validation-updates 
 :green_circle: | total | 55.05 **(0.01)** | 37.03 **(0.09)** | 33.78 **(0)** | 55.53 **(0.01)**
 :red_circle: | app/client/src/utils/autocomplete/TernServer.ts | 52.04 **(-0.22)** | 40.42 **(-0.83)** | 34.48 **(0)** | 55.96 **(-0.26)**
 :green_circle: | app/client/src/workers/validations.ts | 74.92 **(0.77)** | 70.66 **(2.9)** | 64.71 **(0)** | 76.92 **(0.49)**</details>


app/client/src/workers/validations.ts
```diff
@@ -452,7 +456,7 @@ export const VALIDATORS: Record<ValidationTypes, Validator> = {
       if (parsed < Number(config.params.min)) {
         return {
           isValid: false,
-          parsed,
+          parsed: config.params.min || parsed || 0,
```
hetunandu, line 459: If min or parsed is negative it would return 0

riodeuno: Are you talking about the cases in the screenshot below? Could you elaborate

Screenshot 2021-11-16 at 1 49 50 PM
?

hetunandu: Ahh, looks like I misunderstood. Never mind

hetunandu (dismissed):

aswathkk (dismissed):
LGTM

aswathkk (commented):
@riodeuno The regex used for the test is having a bug when I checked it using regex101.com

riodeuno:
@aswathkk Thanks for pointing out the issue!
- What is the bug? Could you explain?
- Do you know of any alternatives? 
- The idea here to make sure regex can be used to validate correctly, does the issue in the regex effect that?

aswathkk:
@riodeuno There's an unescaped `/` in the last group.
Solution would be to escape it
```diff
- regex: /(http(s)?:\/\/.)?(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/,
+ regex: /(http(s)?:\/\/.)?(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&\/=]*)/,
```
> The idea here to make sure regex can be used to validate correctly, does the issue in the regex effect that?

This issue is not affecting the test at all. Randomly tested it using Regex101. This is just a nitpick

aswathkk:
@riodeuno My bad!
This is not an issue in Javascript. I tested in regex101 by selecting the default language available which was PHP

aswathkk (approved):

PR #9178, opened 2021-11-16
fix: PageLoad - action.data implicitly depends on action.actionConfiguration. Added this relationship to dag

trishaanand:


sumitsum (dismissed):

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ActionServiceTest.java
```diff
@@ -2282,4 +2285,97 @@ public void suggestWidget_ArrayListDataEmpty_SuggestTextWidget() {
...
+                    return layoutActionService.updateLayout(page1.getId(), layout.getId(), newLayout);
+
+                });
+        StepVerifier
+                .create(testMono)
+                .assertNext(layout -> {
+                    assertThat(layout).isNotNull();
+                    assertThat(layout.getId()).isNotNull();
+                    assertThat(layout.getLayoutOnLoadActions()).hasSize(1);
+                    // TODO: This is an incorrect expectation, it should be 8 actions in the first set
```
nidhi-nair, line 2368: This doesn't seem valid

app/server/appsmith-server/src/test/java/com/appsmith/server/services/LayoutServiceTest.java
```diff
@@ -508,32 +508,33 @@ public void getActionsExecuteOnLoad() {
...
-                            // TODO: Due to an existing bug the below two actions are detected in the first set
-                            // TODO: Fix the logic to move these actions to the second set and then update the test
-                            "Collection.anAsyncCollectionActionWithoutCall",
-                            "Collection.aSyncCollectionActionWithoutCall"
+                            "hiddenAction4"
                     );
 
                     Set<String> secondSetPageLoadActions = Set.of(
-                            "aPostActionWithAutoExec"
+                            "aTableAction",
```
nidhi-nair, line 522: I know we've been through this 100 times, but do you think there's a scenario where Api1 gets pushed here to the second set because of that logical gap, and an Api2 from another dependency chain also ends up in the second set. The result of this being that the chronology as the user's server expects it to be is affected and breaks something that works today?

nidhi-nair (changes requested):
Just a few comments related changes and a big one 😛

app/server/appsmith-server/src/test/java/com/appsmith/server/services/LayoutServiceTest.java
trishaanand: Because of the addition of actionConfiguration -> data relationship, only a single level would exist for action.data and hence the above scenario would not exist. But I have added a comment in the code where this gets handled.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/LayoutServiceTest.java
```diff
@@ -508,32 +508,33 @@ public void getActionsExecuteOnLoad() {
...
-                            "aTableAction",
-                            "anotherDBAction",
                             "hiddenAction1",
                             "hiddenAction2",
-                            "hiddenAction4",
-                            // TODO: Due to an existing bug the below two actions are detected in the first set
-                            // TODO: Fix the logic to move these actions to the second set and then update the test
-                            "Collection.anAsyncCollectionActionWithoutCall",
-                            "Collection.aSyncCollectionActionWithoutCall"
+                            "hiddenAction4"
```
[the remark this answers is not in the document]
trishaanand: Added comment in the code.

nidhi-nair (approved):

PR #9179, opened 2021-11-16
fix: [Fix] Set user email to lowercase before creating the user and in reset password

### LINKED ISSUE #9091: Store email as lowercase when user signs up
When user signs up with email and password - we make sure this email does not exists for any other user event in other cases. Once this check is successful, we store the email address in the case user provided.

For example, if user signed up with `aBc@example.com`, we store it exactly. This can create a problem when someone invites this user with lower case version to an organization.

The email should be converted to lowercase and stored in lowercase. The user should be able to login with any case.

nayan-rafiq:
## Description

If an user signs up with an email which is not in lowercase, we store it as it is. Instead, we should store the email in lowercase. User should be able to login with case insensitive emails. Also the reset password should work with case insensitive emails.

Fixes #9091 

## Type of change

## How Has This Been Tested?

- Manually
- JUnit tests

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/UserServiceImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/UserServiceImpl.java
@@ -194,8 +194,8 @@ public class UserServiceImpl extends BaseService<UserRepository, User, String> i
      * This function creates a one-time token for resetting the user's password. This token is stored in the `passwordResetToken`
      * collection with an expiry time of 48 hours. The user must provide this one-time token when updating with the new password.
      *
-     * @param resetUserPasswordDTO
-     * @return
+     * @param resetUserPasswordDTO DTO object containing the request params from form
+     * @return True if email is sent successfully
      */
     @Override
     public Mono<Boolean> forgotPasswordTokenGenerate(ResetUserPasswordDTO resetUserPasswordDTO) {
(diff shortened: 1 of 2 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/services/UserServiceImpl.java
```diff
@@ -212,34 +212,34 @@ public UserServiceImpl(Scheduler scheduler,
...
-                    return Mono.just(passwordResetToken);
-                }))
-                .map(resetToken -> {
-                    // check the validity of the token
-                    validateResetLimit(resetToken);
-                    resetToken.setTokenHash(passwordEncoder.encode(token));
-                    return resetToken;
-                });
+        return repository.findByEmail(email)
+                .switchIfEmpty(repository.findByCaseInsensitiveEmail(email))
```
abhvsn, line 218: @nayan-rafiq Why do we need this change little confused here? Do we have run the migration to change all emails to lowercase but I guess we are not doing that as well because with that change we will be merging accounts which we are not doing deliberately?

nayan-rafiq: If an user signs up with an uppercase email e.g. `NaYan@appsmith.com`, we'll store them as `nayan@appsmith.com`. The user can login with both `NaYan@appsmith.com` and `nayan@appsmith.com`. As they can login, they should be able to reset their password using any version of their email address. Earlier it would return user not found.

abhvsn (approved):
LGTM

PR #9247, opened 2021-11-18
ci: Upload server logs if Cypress tests fail

mohanarpit:
This will help us debug some Cypress tests that fail because of an API that has returned erroneously.


yatinappsmith (approved):
LGTM

PR #9311, opened 2021-11-24
fix: google sheets plugin input fields for numbers parse issue

### LINKED ISSUE #8196: [Bug] Google sheets plugin: Row limit does not parse properly
## Description

In case the row limit property in the google sheet plugin has an extra new line the query return an error saying that the row limit is not valid, We should trim such characters from the property value

1. Use google sheet get query
2. While defining the row limit add an additional new line
3. Run the query to see the error

Note: We will also need to test this for the rest for the fields.
 
### Important Details

- Version: [Cloud]
- OS: [MacOSX]
- Browser [chrome]
- Environment [production]

leotom2000:
## Description

> Google sheets plugin: Row limit does not parse properly
> When We give a space or enter in the input where an Integer is expected, it is throwing error.

Fixes #8196

>By adding a trim() where an integer value is expected, this error is handled
>This fix resolved in 5 input fields, and also handles enter keys before and after values

## Type of change

## How Has This Been Tested?

>Use google sheet get query
-While defining the row limit add an additional new line after input value - Error was found and fixed now
-While defining the row limit add an additional new line before input value- Error was found and fixed now
-While defining the row limit add a space before input value- Error was found and fixed now
-While defining the row limit add a space before input value- Error was found and fixed now
 
>Tested for the following input fields
(1) Table Heading Row Index   - Error was found and fixed now
(2) Row Offset   - Error was found and fixed now
(3) Row Limit   - Error was found and fixed now
(4) Cell Range   - Error was found and fixed now

## Checklist:

### DIFF
--- a/app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/MethodConfig.java
+++ b/app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/MethodConfig.java
@@ -58,25 +58,25 @@ public class MethodConfig {
                         }
                         break;
                     case "range":
-                        this.spreadsheetRange = propertyValue;
+                        this.spreadsheetRange = propertyValue.trim();
                         break;
                     case "spreadsheetName":
                         this.spreadsheetName = propertyValue;
                         break;
                     case "tableHeaderIndex":
-                        this.tableHeaderIndex = propertyValue;
+                        this.tableHeaderIndex = propertyValue.trim();
                         break;
                     case "queryFormat":
                         this.queryFormat = propertyValue;
                         break;
                     case "rowLimit":
-                        this.rowLimit = propertyValue;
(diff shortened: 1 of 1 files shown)


app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/MethodConfig.java
```diff
@@ -58,25 +58,25 @@ public MethodConfig(List<Property> propertyList) {
                         }
                         break;
                     case "range":
-                        this.spreadsheetRange = propertyValue;
+                        this.spreadsheetRange = propertyValue.trim();
```
nidhi-nair, line 61: @leotom2000 Do you think we could handle these by trimming all property values at L#47 instead?

leotom2000:
Raised another pull request, due to some version conflicts

PR #9325, opened 2021-11-24
fix: Load env variables at start of backend and rts

sharat87:


deploy/docker/scripts/run-with-env.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+ENV_PATH="/appsmith-stacks/configuration/docker.env"
+echo 'Load environment configuration'
+set -o allexport
+. "$ENV_PATH"
+set +o allexport
+
+if [[ -z "${APPSMITH_MAIL_ENABLED}" ]]; then
+  unset APPSMITH_MAIL_ENABLED # If this field is empty is might cause application crash
```
mohanarpit, line 10: Quick question: Why will the application crash if these fields are empty? Aren't we handling this scenario in our server code base?

mohanarpit (approved):

deploy/docker/scripts/run-with-env.sh
sharat87: Yeah it crashes unfortunately 😞

PR #9364, opened 2021-11-25
fix: Google Sheet Plugin Numeric inputs Whitespace issue.

### LINKED ISSUE #8196: [Bug] Google sheets plugin: Row limit does not parse properly

leotom2000:
Fixes #8196
> By adding a trim() where an integer value is expected, this error is handle
>This fix resolves 5 input fields, and also handles enter keys before and after values

## Type of change
- Unit Test has been added for this issue

## How Has This Been Tested?
Use google sheet get query
-While defining the row limit add an additional new line after input value - Error was found and fixed now
-While defining the row limit add an additional new line before input value- Error was found and fixed now
-While defining the row limit add a space before input value- Error was found and fixed now
-While defining the row limit add a space before input value- Error was found and fixed now

Tested for the following input fields
(1) Table Heading Row Index - Error was found and fixed now
(2) Row Offset - Error was found and fixed now
(3) Row Limit - Error was found and fixed now
(4) Cell Range - Error was found and fixed now

## Checklist:

### DIFF
--- a/app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/MethodConfig.java
+++ b/app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/MethodConfig.java
@@ -58,25 +58,25 @@ public class MethodConfig {
                         }
                         break;
                     case "range":
-                        this.spreadsheetRange = propertyValue;
+                        this.spreadsheetRange = propertyValue.trim();
                         break;
                     case "spreadsheetName":
                         this.spreadsheetName = propertyValue;
                         break;
                     case "tableHeaderIndex":
-                        this.tableHeaderIndex = propertyValue;
+                        this.tableHeaderIndex = propertyValue.trim();
                         break;
                     case "queryFormat":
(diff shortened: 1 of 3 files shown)


leotom2000:
Reply to @nidhi-nair query : Do you think we could handle these by trimming all property values at L#47 instead?

Line 47 is processing almost 10 input data, out of this we need validation only for 4 items, so considering unnecessary processing ( trim() ), I have place the trim() in the required 5 places.

nidhi-nair:
@leotom2000 I think we can safely add this check for all other inputs as well. It is valid to expect this validation to exist in all other fields as well. Could you please add the check as common for everything instead?

nidhi-nair (approved):

PR #9392, opened 2021-11-26
fix: Added client side execution logic and on page load logic for js

### LINKED ISSUE #9332: Enable onPageLoad js functions

nidhi-nair:
Fixes #9332


app/server/envs/dev.env.example
```diff
@@ -12,6 +12,7 @@ APPSMITH_ENCRYPTION_SALT=abcd
 APPSMITH_CODEC_SIZE=10
 
 #APPSMITH_CLOUD_SERVICES_BASE_URL="https://release-cs.appsmith.com"
+APPSMITH_CLOUD_SERVICES_BASE_URL="http://localhost:8090"
```
trishaanand, line 15: Should this change be ignored? Folks won't be running cloud services on their local usually.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/LayoutServiceTest.java
```diff
@@ -519,14 +519,12 @@ public void getActionsExecuteOnLoad() {
...
                             "hiddenAction2",
-                            "hiddenAction4"
+                            "hiddenAction4",
+                            "Collection.anAsyncCollectionActionWithoutCall",
+                            "Collection.aSyncCollectionActionWithoutCall"
                     );
 
-                    // TODO : Once the client implements on page load execution of JS, uncomment the lines below.
                     Set<String> secondSetPageLoadActions = Set.of(
-//                            "Collection.anAsyncCollectionActionWithoutCall",
```
trishaanand, line 527: Is this correct? I think the dependency calculation might be incorrect if this is getting recognized as first set of actions for page load instead of the second set.

nidhi-nair: Why would it be second set, it's a direct reference right?

nidhi-nair: Oh yes you're right, I forgot about hiddenAction being referenced by js actions. I'll fix this.

trishaanand: Actually this is the bug that we shall catch up for on Monday. Let me resolve this and not take this up as part of on page load JS changes. Apologies. Could you just add a comment that the action needs to be moved to the second set post the bug fix?

nidhi-nair: Ah I'd missed this comment. I just understood context. :)

trishaanand (approved):

Issue #9096, opened 2021-11-10
[Bug]: Widget doesnt evaluate; Dependency Map Issue


Tooluloope:
### Is there an existing issue for this?

- [ ] I have searched the existing issues

### Current Behavior

[![LOOM DEMO](http://cdn.loom.com/sessions/thumbnails/2051e9bc9a2b464a894e1a7531ecf924-00001.gif)](https://www.loom.com/share/2051e9bc9a2b464a894e1a7531ecf924)

From the loom video, I have this function in those widgets, and they all evaluate to undefined on page load, but when I remove their dependency from `input1.text` and make `const index = "IN"`, the evaluate correctly.
```js
{{function() {
    const index = Input1.text

    const obj = {
        "IN":[{
            "label":"Ezetap (BETA !!)",
            "value":"ACCESS_EZETAP"
            }], 
            "ZA":[{
            "label": "Yoco",
            "value": "ACCESS_YOCO"
            }],
            "NZ":[{
            "label": "Smartpay",
            "value":"ACCESS_SMARTAY"
            }],
            "AU":[{
            "label": "Tyro",
            "value": "ACCESS_TYRO"
            },{
            "label": "Smartpay",
            "value":"ACCESS_SMARTAY"
            }],
            "JP":[{
            "label": "Coiney",
            "value":"ACCESS_COINEY"
            },{"label": "CpayPro",
            "value":"ACCESS_ONEPAY"
            },{"label": "Softbank",
            "value":"ACCESS_SOFTBANK"
            },{"label": "Paygate",
            "value":"ACCESS_PAYGATE"
            }],
            "KR":[{
            "label": "Kiccgate",
            "value": "ACCESS_KICC"
            },{"label": "Nice",
            "value":"ACCESS_NICEPAY"
            }],
            "CA":[{
            "label": "Worldpay",
            "value":"ACCESS_VANTIV"
            }],
            "US":[
            {"label": "Worldpay",
            "value":"ACCESS_VANTIV"
            },{ "label": "iZettle",
            "value":"ACCESS_IZETTLE"
            },{ "label": "Sumup",
            "value": "ACCESS_SUMUP"
            },{
            "label":"Cardconnect",
            "value":"ACCESS_FIRSTDATA"
            }],
            "BR":[{
            "label": "Sumup",
            "value":"ACCESS_SUMUP"
            },{ "label": "iZettle",
            "value":"ACCESS_IZETTLE"}],
            "FI":[{
[...]

PR #9418, opened 2021-11-28
fix: Add default props dependency in dependency map

### LINKED ISSUE #9334: [Bug]: Select query fails on crud when the user switches between the pages

### Current Behavior

The Select query fails on crud when the user switches between the pages
[LOOM DEMO](https://www.loom.com/share/df79137264b341f2955bdee6542bd2a8)

### Steps To Reproduce

1. Generate a crud page
2. Now create a new page and navigate back to the Previously created Generate page
3. The select query fails

### Environment

Release

### Version

Cloud

Rishabh-Rathod:
Fixes #9334 #9096 

### Problem

When we have a SelectWidget named `Select1` on canvas and `createFirstTree` runs, it was noticed that `Select1.defaultValue` is getting evaluated after `Select1.selectedOptionValue` 

This is due to an incorrect sorted dependency map.
Logically, `Select1.selectedOptionValue` is dependent on `defaultValue` but in dependencyMap current there is no such relation.
 ```
 selectedOptionValue: `{{(()=>{const index = _.findIndex(this.options, { value: this.defaultValue }); return this.options[index]?.value; })()}}`
 ```
 
But when we see the actual dependencyMap, there is no relation between `selectedOptionValue` and `defaultValue`.

Screenshot 2021-11-29 at 12 12 33 AM

#### Why does this doesn't happen `onPageLoad` when it happens `onPageSwitch` if it is a depenedencyMap issue?
This doesn’t happen onPageLoad because there we also run updateDataTree,
1. In createDataTree’s evaluation itself defaultValue is evaluated and stored in dataTree 
2. hence. when executing updateDataTree and evaluating `Select1.selectedOptionValue`  before `Select1.defaultValue`  doesn’t seem a problem as `Select1.defaultValue` got defined.

Though, logically it is an issue because we need to always evaluate `Select1.defaultValue` before `Select1.selectedOptionValue`.

## Solution

Make sure property with `undefined` value doesn't get omitted.

When forming a dependency map, getAllKeys will have defaultValue as key and this will result `Select1.selectedOptionValue` to depend on `Select1.defaultValue`.

This will fix the sortedDependencyMap and also the evaluation sequence.

## Type of change

## Checklist:

### DIFF
--- a/app/client/src/workers/DataTreeEvaluator.ts
+++ b/app/client/src/workers/DataTreeEvaluator.ts
@@ -93,6 +93,7 @@ export default class DataTreeEvaluator {
   }
 
   createFirstTree(unEvalTree: DataTree) {
+    debugger;
     const totalStart = performance.now();
     let localUnEvalTree = JSON.parse(JSON.stringify(unEvalTree));
     let jsUpdates: Record<string, JSUpdate> = {};
@@ -1541,6 +1542,7 @@ const extractReferencesFromBinding = (
   const identifiers = dependentPath.match(/[a-zA-Z_$][a-zA-Z_$0-9.\[\]]*/g) || [
     dependentPath,
   ];
+  debugger;
   identifiers.forEach((identifier: string) => {
     // If the identifier exists directly, add it and return
     if (all.hasOwnProperty(identifier)) {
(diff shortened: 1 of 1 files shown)


eco-monk:
LGTM. Nice catch.! 👍 

Since this is a problem with defaultProps in general, I suggest testing with other widgets too.
e.g. input(DefaultText), checkBox(DefaultChecked) etc...

ApekshaBhosale (approved):

eco-monk (approved):

RakshaKShetty (approved):

PR #9423, opened 2021-11-29
fix: added support for timePrecision

### LINKED ISSUE #9100: [Feature]: Allow users to insert hour:time:seconds - improved datetimepicker

### Summary

Improve the datepicker, isn't possible insert the seconds even if the format datetime have the seconds

Maybe having three widgets, datepicker, datetimepicker, timepicker... otherwise maybe improve the actual datepicker so can be possible choose also the seconds
image

techbhavin:
## Description

> Allow users to insert hour:time: seconds in date-time-picker widget

Fixes #9100 

## Type of change

## Checklist:

    // Code coverage diff between base branch:release and head branch: feature/improved-datetimepicker 
 :green_circle: | total | 55.23 **(0.01)** | 37.02 **(0.01)** | 34.08 **(0.01)** | 55.75 **(0.02)**
 :sparkles: :new: | **app/client/src/widgets/DatePickerWidget2/constants.ts** | **100** | **100** | **100** | **100**
 :red_circle: | app/client/src/widgets/DatePickerWidget2/component/index.tsx | 27.78 **(0)** | 3.08 **(-0.09)** | 6.25 **(0)** | 27.14 **(0)**
 :green_circle: | app/client/src/widgets/DatePickerWidget2/widget/index.tsx | 84.62 **(0.62)** | 0 **(0)** | 75 **(0)** | 87.5 **(0.54)**
 :green_circle: | app/client/src/widgets/TableWidget/component/CascadeFields.tsx | 28.68 **(0.55)** | 24.66 **(0)** | 3.13 **(0)** | 30.33 **(0.58)**</details>


jsartisan (dismissed):

shwetha-ramesh:
@techbhavin tested this PR
1. the time slot input box appears only when the input is dropped on the canvas. When the user changes from the default format to any other format and moved back to the default format the time slot input disappears. (refer video)
2. Time slot input not available for any of the format that.

https://loom.com/share/406044fb99a4403489afcc487dd83ed2

jsartisan (dismissed):

shwetha-ramesh:
@techbhavin tested this PR. Below are the points.
1) The seconds field must not be shown for _YYYY-MM-DD HH:mm_ format or any format that does not have _ss_ mentioned, irrespective of what 'time precision' is set
https://loom.com/share/0f9adc6e148d42fda0c96a5a7842c068

2) _Clear_ must clear the timeslot values also.
image

https://loom.com/share/b10f5565f2b74c7ea760de1d7db4c492

3) Since the _Time precision_ field is placed at the end of General settings, the user might not know a field as such exists and when changes the format which has hh:mm:ss and not be able to change the seconds field.

techbhavin:
> The seconds field must not be shown for `YYYY-MM-DD HH:mm` format or any format that does not have ss mentioned, irrespective of what 'time precision' is set

In a current implementation, we show `hh:mm`  in every format. did we want to do as per @shwetha-ramesh  mention?

ALSO NEED FEEDBACK about point 2 , 3
  
@somangshu  @sbalaji1192  @riodeuno

sbalaji1192:
@techbhavin @shwetha-ramesh @Tooluloope 
WRT,
#1, Current implementation seems good as it's not leading to any bug. Having two configs controlling the same field is bit much to maintain.
#2, Agreed on the callout. we need to clear the time as well. since closing and opening the popup resets the time.
#3, Let's move the time precision right below to the time format.

techbhavin:
@sbalaji1192  @shwetha-ramesh  @Tooluloope 

point 3: updated 
point 2: here we have a limitation in the library we can’t reset the time picker value. 

image

app/client/src/widgets/DatePickerWidget2/widget/index.tsx
```diff
@@ -124,6 +124,41 @@ class DatePickerWidget extends BaseWidget<DatePickerWidget2Props, WidgetState> {
...
+              },
+            ],
+            isJSConvertible: true,
+            isBindProperty: true,
+            isTriggerProperty: false,
+            validation: {
+              type: ValidationTypes.TEXT,
+              params: {
+                allowedValues: [
+                  "None",
```
riodeuno, line 154: Let's include this in the TimePrecision enum. `TimePrecision.NONE`. 
One way to achieve this is to create a TimePrecision enum by ourselves.
This helps with maintaining integrity across the codebase regarding time precision values.

riodeuno (changes requested):

app/client/src/widgets/DatePickerWidget2/constants.ts
```diff
@@ -1 +1,10 @@
 export type DatePickerType = "DATE_PICKER" | "DATE_RANGE_PICKER";
+
+export declare const TimePrecision: {
```
riodeuno, line 3: Is there a reason we used `declare const` instead of an `enum`?

riodeuno (dismissed):

riodeuno:
@sbalaji1192 I've checked the code. Please verify the issue resolution in your review. @shwetha-ramesh Moved this to the QA pipeline.

shwetha-ramesh:
> @sbalaji1192 @shwetha-ramesh @Tooluloope
> 
> point 3: updated point 2: here we have a limitation in the library we can’t reset the time picker value.
> 
> image

@techbhavin @sbalaji1192 Are we not clearing the time ?

shwetha-ramesh:
@techbhavin tested this PR. Found the below issue.

1. When the user clicks on _Clear_ , the field gets empty and no selection is seen on the calendar however the previous time still remains on the time slot field. Only when the user clicks on the canvas and then back on the widget the time slot field will be cleared (Refer video)
https://loom.com/share/6933df7b7d7f44b085884950866af58c

2. After _Clear_ the timeslot field must display 00:00:00 / 00:00 and not 0:00:00 / 0:00
image

Ideal behavior would be : On _Clear_ both time and date must be cleared without the user having to go back and forth on the widget and 00:00:00 format must be displayed in timeslot field after clear.

Moving to Inprogress.

techbhavin:
Above issue 1 and 2 is Out of scope issues. for reference, 
  I am please refer [blueprintjs-datepicker](https://blueprintjs.com/docs/#datetime/datepicker) 
  
@shwetha-ramesh  @sbalaji1192  @Tooluloope  @somangshu

somangshu:
@techbhavin can we solve for the issue 1 reported above by @shwetha-ramesh. I think its valid to see the correct state, Can we set it on clear? Cause rerendering the widget cause it to show default

techbhavin:
@somangshu  this is not possible reason is time picker is controlled by the date-picker. here showing `0:00` reason is component is re-render on date-picker popover closed that why the state is re-set. 

we don't have a clear `onClick` handler.

somangshu:
@jsartisan can you think of a way to re-render the component so that we see the update

jsartisan:
@somangshu We can't do anything if library is not giving any callback handler :/

cc: @techbhavin

somangshu:
@shwetha-ramesh @techbhavin we can proceed then

shwetha-ramesh:
As agreed upon with above comments. Moving to DONE.

app/client/src/widgets/DatePickerWidget2/index.ts
```diff
@@ -2,6 +2,7 @@ import Widget from "./widget";
 import IconSVG from "./icon.svg";
 import moment from "moment";
 import { GRID_DENSITY_MIGRATION_V1 } from "widgets/constants";
+import { TimePrecision } from "@blueprintjs/datetime";
```
sbalaji1192, line 5: Lets use our TimePrecision

sbalaji1192 (approved):

PR #9460, opened 2021-11-30
fix: Fix Appsmith image name in Helm chart values

sharat87:
In Helm charts' values.yaml file, we are pointing to the slim editor image instead of the fat Appsmith image. This PR fixes this.


sub1983 (approved):
approved

PR #9473, opened 2021-11-30
fix: Modified action execution endpoint

### LINKED ISSUE #8431: [Bug]: Add support for raw binary data in actions

### Current Behavior

Currently the REST API is unable to post raw binary data to endpoints. We only support base64 encoded data for such operations. Some endpoints mandatorily need binary data though, so our proxy server should handle such data as well.

### Steps To Reproduce

```markdown
Create a REST API and try to use multipart/form-data with raw data
```

### Environment

Production

### Version

Cloud

### LINKED ISSUE #9395: [Feature] Client side changes to support raw data for action execution

### LINKED ISSUE #11510: [Bug]: Raw data execution keys have encoded quotes

### Current Behavior

With the new [multipart form based action executions](https://github.com/appsmithorg/appsmith/pull/9473), we're passing mustache bindings as the keys for the data. However, the client is passing quotes in these keys in their URL encoded form (as `%22`).

Investigation so far has shown that Postman handles this well, and replicating the request with axios on a node server works as well. We need to look at whether the browser usage is affecting the request creation in any way.

### Steps To Reproduce

On test.appsmith.com,
1. Create a REST API to a mock server
2. Create a mustache binding with the value `{{ "Test string" }}`
3. Run the API and check the network console

### Environment

Deploy Preview

### Version

Cloud

nidhi-nair:
Fixes #8431
Fixes #9395
Fixes #11510 

This change modifies the interface for action execution to a multipart. The idea is to be able to send binary data as is, and enhance support. for larger evaluated values. A caveat to be noted with this change is that we're always going to be sending the mustache binding keys as keys in the multipart request, and these keys will always be in their encoded format.

Additional info:
- Each request has an `executeActionDTO` key that hold meta properties for execution
- Binary data is always converted into a UTF-8 String. Since we have not solved for mustache replacement logic with streams, this remains a String. Once we have support for streaming, we can start using the byte stream directly.
- The change only allows us to use binary data in REST APIs for now, since each plugin will now have to add logic for supporting binary data. We can follow this PR up with individual plugin level handling to add this support.

    // Code coverage diff between base branch:release and head branch: fix/binary-data-for-exec 
 :green_circle: | app/client/src/sagas/ActionExecution/PluginActionSaga.ts | 18.92 **(0.07)** | 2.5 **(0.02)** | 11.11 **(0)** | 21.62 **(0)**
 :green_circle: | app/client/src/utils/autocomplete/TernServer.ts | 52.94 **(0.23)** | 41.67 **(0.84)** | 36.21 **(0)** | 56.99 **(0.25)**</details>


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/plugins/PluginExecutor.java
```diff
@@ -170,7 +172,7 @@ default void variableSubstitution(ActionConfiguration actionConfiguration,
                             // We also add a backslash before every double-quote or backslash character
                             // because we apply the template replacing in a JSON-stringified version of
                             // these properties, where these two characters are escaped.
-                            p -> p.getKey().trim(), // .replaceAll("[\"\n\\\\]", "\\\\$0"),
+                            p -> URLDecoder.decode(p.getKey().trim(), StandardCharsets.UTF_8), // .replaceAll("[\"\n\\\\]", "\\\\$0"),
```
sumitsum, line 175: QQ: do we need a similar decode before comparing `param.getKey()` in `smartSubstitutionOfBindings` method ?

sumitsum (approved):

PR #9535, opened 2021-12-02
chore: Send telemetry option in user profile to client

sharat87:

    // Code coverage diff between base branch:release and head branch: fix/telemetry-from-welcome-page 
 :red_circle: | total | 55.15 **(-0.01)** | 37.19 **(0)** | 33.99 **(0)** | 55.66 **(-0.01)**
 :green_circle: | app/client/src/configs/index.ts | 95.12 **(-0.53)** | 86.14 **(1.52)** | 100 **(0)** | 97.3 **(-0.32)**
 :green_circle: | app/client/src/sagas/userSagas.tsx | 16.17 **(0.15)** | 1.23 **(-0.07)** | 5 **(0)** | 19.39 **(0.12)**
 :green_circle: | app/client/src/utils/AnalyticsUtil.tsx | 29.17 **(0)** | 19.05 **(7.94)** | 30.77 **(0)** | 29.35 **(0)**
 :red_circle: | app/client/src/utils/AppsmithUtils.tsx | 56.52 **(0.15)** | 30.61 **(0)** | 33.33 **(-0.76)** | 51.41 **(-0.02)**</details>

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/UserProfileDTO.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/UserProfileDTO.java
@@ -41,7 +41,7 @@ public class UserProfileDTO {
 
     String useCase;
 
-    boolean disableTelemetry = true;
+    boolean enableTelemetry = false;
 
     public boolean isAccountNonExpired() {
         return this.isEnabled;
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/UserProfileDTO.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/UserProfileDTO.java
@@ -41,6 +41,8 @@ public class UserProfileDTO {
 
     String useCase;
 
+    boolean disableTelemetry = true;
+
     public boolean isAccountNonExpired() {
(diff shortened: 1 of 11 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/UserProfileDTO.java
```diff
@@ -41,6 +41,8 @@
 
     String useCase;
 
+    boolean enableTelemetry = false;
```
mohanarpit, line 44: Since this is an instance level check, any reason we are adding this to the user profile?

sharat87: The client needs this information as early as possible. The user profile happens to be the first request from the client so we put it there. After this, the client won't look at the telemetry env variable at all. The server will do that. The client will just get the value from the server instead.

rishabhsaxena (dismissed):
FE changes LGTM!

rishabhsaxena (approved):
Good catch for sentry and smartlook!

rishabhsaxena (approved):

mohanarpit (approved):

PR #9565, opened 2021-12-04
fix: Use port from mock db details when creating mock datasource

sharat87:
The port information from cloud services is not being used when creating a new mock datasource. This isn't a problem today since all mock datasources use the default ports. Fixing this now so it doesn't become a headache later.

Marking it as draft since we want this merged _after_ a release made after 5-Dec-2021.


AnaghHegde (approved):
LGTM

PR #9705, opened 2021-12-10
Allow runtime env vars to create docker.env file

amirhmoradi:
This allows to install appsmith with docker compose with all vars set in the "environment" key and helps avoid the user to modify the internal docker.env file  after first install or edit it manually.

## Description

This allows to install appsmith with docker compose with all vars set in the "environment" key and helps avoid the user to modify the internal docker.env file  after first install or edit it manually.

Fixes # (issue)

All env vars are set at first deploy, and user does not need to edit manually the docker.env file after the first install.

## Type of change

## How Has This Been Tested?

Tested by setting vars in docker compose 'environment' key and check that content is generated in /app-stacks/configuration/docker.env correctly on first install/launch.
 Example docker compose file:
```yaml
services:
  appsmith:
    image: index.docker.io/appsmith/appsmith-ce
    labels:
      com.centurylinklabs.watchtower.enable: "true"
    volumes:
      - "/data/stacks/:/appsmith-stacks"
    ## Testing here my PR:
      - "/data/docker.env.sh:/opt/appsmith/templates/docker.env.sh"
    environment:
      - APPSMITH_CUSTOM_DOMAIN=backoffice.example.com
      - APPSMITH_SIGNUP_DISABLED=true
      - APPSMITH_ADMIN_EMAILS=admin@example.com
      - APPSMITH_MAIL_ENABLED=true
      - APPSMITH_MAIL_SMTP_AUTH=true
#...
```

## Checklist:

- [Y] My code follows the style guidelines of this project
- [Y] I have performed a self-review of my own code
- [N/A] I have commented my code, particularly in hard-to-understand areas
- [N/A] I have made corresponding changes to the documentation
- [Y] My changes generate no new warnings
- [Y] I have added tests that prove my fix is effective or that my feature works
- [Y] New and existing unit tests pass locally with my changes


mohanarpit:
@amirhmoradi Thank you for opening a PR & improving the Appsmith project. 🎉 

Unfortunately, I don't think this is a good idea for Appsmith to implement. While I understand the underlying pain point, we are introducing an Admin UI page within Appsmith to manage all the environment variables. This Admin page syncs with the `docker.env` file & writes changes back to the file. This way, as a user, you can always be sure that the values you are seeing within Appsmith are the same in the docker.env file & vice-versa. If we override values via Docker compose's environment section, then discrepancies will start arising. This will get confusing for users. 

I hope you understand where I'm coming from. Once again, I thank you for your PR contribution & look forward to other improvements/suggestions for Appsmith. 😄

PR #9785, opened 2021-12-15
fix: Fix action collection not being deleted when the deleted collection in edit mode is published

### LINKED ISSUE #9722: [Bug]: Action collection is not gettting deleted even if we publish deleted actionCollection

### Current Behavior

When I'm trying to delete the actionCollection entry I'm unable to do so. Basically `delted` or `deletedAt` fields are not getting updated after we delete the actionCollection from UI. This is causing the issues in import-export flow as this might get exported the deleted actionCollection. 

### Steps To Reproduce

1. Create actionCollection
2. Create binding to a widget on the canvas
3. Publish app
4. Delete the action collection from edit mode
5. `deletedAt` field gets updated in unpublishedCollection as expected
6. Publish the application again
7. Also `deleted` field remains false

### Environment

Release

### Version

Cloud

abhvsn:
## Description

> This will fix the issue related to action collection not being deleted when the deleted collection is published. This fix will fetch the collection by appId and then filter as per the view mode  

Fixes #9722 

## Type of change

## How Has This Been Tested?

> Junit TC
> Manual test

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ActionCollectionServiceImpl.java
```diff
@@ -73,7 +73,7 @@ public ActionCollectionServiceImpl(Scheduler scheduler,
 
     @Override
     public Flux<ActionCollection> findAllByApplicationIdAndViewMode(String applicationId, Boolean viewMode, AclPermission permission, Sort sort) {
-        return repository.findByApplicationIdAndViewMode(applicationId, viewMode, permission)
+        return repository.findByApplicationId(applicationId, permission, sort)
```
mohanarpit, line 76: Why aren't we considering the `viewMode` any more here? Will this have any other impact?

nidhi-nair (dismissed):

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ActionCollectionServiceImpl.java
abhvsn: We are filtering the collection inside the flatmap for the `viewMode`.

mohanarpit: That doesn't seem to be the case. In the `flatMap` below, we are only checking for `viewMode` value to return Mono.empty() for empty collection sizes. We aren't performing any filtering there.

nidhi-nair: It's in the upstream method that the filtering is happening @mohanarpit. I think we'll need to go through a cycle of renaming these methods, they are no longer clearly indicative of their purpose.

nidhi-nair (approved):

PR #9804, opened 2021-12-15
chore: Update local git repo path

abhvsn:
## Description
To support volumes for git-sync this PR update default path. 

## Type of change

### DIFF
--- a/app/server/docker-compose.yml
+++ b/app/server/docker-compose.yml
@@ -15,7 +15,7 @@ services:
     networks:
       - appsmith
     volumes:
-      - ./container-volumes/git-storage:/data/git-storage
+      - /container-volumes/git-storage:/data/git-storage
 
   mongo:
     image: mongo:4.4.6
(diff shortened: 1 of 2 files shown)


app/server/docker-compose.yml
```diff
@@ -15,7 +15,7 @@ services:
     networks:
       - appsmith
     volumes:
-      - ./container-volumes/git-storage:/data/git-storage
+      - /container-volumes/git-storage:/data/git-storage
```
sharat87, line 18: We need the `.` here since we want the path on the host system to be relative to the `docker-compose.yml` file, instead of at the root `/` path. We want a root based path only _inside_ the container, which is `/data/git-storage`.

```suggestion
      - ./container-volumes/git-storage:/data/git-storage
```

sharat87 (changes requested):

sharat87 (approved):

PR #9821, opened 2021-12-16
fix: Use Appsmith user profile as a fallback git profile

abhvsn:
## Description

> If the git-profile is not present for user, Appsmith user profile will be used as a fallback value. Also we will be supporting a null values in authorName and authorEmail when for gitProfile when user intends to use globalProfile. This PR introduces a boolean field in gitProfile to use repo specific 

## Type of change

## How Has This Been Tested?

> Junit TC
> Manual testing 

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
@@ -164,8 +164,7 @@ public class GitServiceCEImpl implements GitServiceCE {
 
         return sessionUserService.getCurrentUser()
                 .flatMap(user -> userService.findByEmail(user.getEmail()))
-                .flatMap(user -> userDataService
-                        .getForUser(user.getId())
+                .flatMap(user -> userDataService.getForUser(user.getId())
                         .flatMap(userData -> {
                             // GitProfiles will be null if the user has not created any git profile.
                             GitProfile savedProfile = userData.getGitProfileByKey(defaultApplicationId);
@@ -179,7 +178,6 @@ public class GitServiceCEImpl implements GitServiceCE {
                             }
                             return Mono.just(userData.getGitProfiles());
                         })
-                        .filter(profiles -> !CollectionUtils.isNullOrEmpty(profiles))
                         .switchIfEmpty(Mono.defer(() -> {
                                     // If profiles are empty use Appsmith's user profile as git default profile
                                     GitProfile profile = new GitProfile();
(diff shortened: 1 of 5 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -194,19 +192,48 @@ private String getReason() {
...
                         )
+                        .filter(profiles -> !CollectionUtils.isNullOrEmpty(profiles))
                 );
     }
 
     @Override
     public Mono<Map<String, GitProfile>> updateOrCreateGitProfileForCurrentUser(GitProfile gitProfile) {
-        gitProfile.setUseGlobalProfile(true);
+        gitProfile.setUseGlobalProfile(null);
         return updateOrCreateGitProfileForCurrentUser(gitProfile, DEFAULT);
```
AnaghHegde, line 202: Why are we setting useGlobalProfile flag to null.

abhvsn: `useGlobalProfile` is not useful for the defaultGitProfile for the user. Consider a situation where user want to use the deaultProfile instead of repo specific profile this setting will be done at repo level by setting `useGlobalProfile` flag to true. This is the reason now we started supporting null values for repo level authorName and authorEmail as the flag can be toggled from the client side.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -216,6 +253,13 @@ private String getReason() {
...
                     if (gitProfile != null && gitProfile.getUseGlobalProfile() == null) {
                         gitProfile.setUseGlobalProfile(true);
+                    } else if (gitProfile == null) {
+                        // If the profile is requested for repo specific using the applicationId
+                        GitProfile gitProfile1 = new GitProfile();
+                        gitProfile1.setAuthorName("");
+                        gitProfile1.setAuthorEmail("");
+                        gitProfile1.setUseGlobalProfile(true);
+                        return gitProfile1;
                     }
```
AnaghHegde, line 263: Why are we modifying the flag while fetching the profile ? I am little confused here. Shouldn't this be done during update or create or connectToGit flow ?

AnaghHegde (approved):

AnaghHegde (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
abhvsn: Here we are not updating the DB just sending empty values for name and email.

Thanks for pointing around the names of the methods!

PR #9900, opened 2021-12-21
fix: Add CSV validation for font-style

### LINKED ISSUE #6797: [Bug]-[688]: JS code on text style does not clear when user turns off the JS mode on text widget causing unexpected behaviour
## Description
When JS code on text style is input with an incorrect code and JS toggle is turned off, then the error still stays on debugger and turning on text style via buttons (**bold, italics**) only adds to the existing JS incorrect JS code.

1. Add text widget on to canvas and turn on JS code for text colour
2. Input a code such that an error appears on the debugger
3. Now turn off the JS code and use bold/italics buttons to define text style - observe the JS error still present on the debugger and any bold/italics value being added to the invalid JS code
4. toggle back on to JS code and observe that the code was not cleared and selected bold/italics are added to the code

### Important Details

- Version: Cloud
- OS: Win10
- Browser: Chrome
- Environment: Release

[LOOM DEMO](https://www.loom.com/share/9e173c960b3d4f01a65a38ffbcb93503)

aswathkk:
## Description

Added CSV validation to fix a couple of issues with the Font Style property in multiple widgets. The plan is to avoid the CSV validation in the future. Refer the Slack thread [here](https://theappsmith.slack.com/archives/CNQ9Q91C0/p1640150925236800).

Fixes #6797

Widgets whose font style property got modified:
- TextWidget
- Table Widget
- Select widget
- multi select widget
- Single select tree
- multi select tree
- input widget
## Type of change

## How Has This Been Tested?

Tested manually

## Checklist:

    // Code coverage diff between base branch:release and head branch: fix/font-style-issue 
 :green_circle: | total | 55.68 **(0.01)** | 37.1 **(0.04)** | 35.35 **(0.01)** | 55.94 **(0)**
 :green_circle: | app/client/src/utils/validation/common.ts | 62.3 **(3.28)** | 60.87 **(3.37)** | 33.33 **(0)** | 61.67 **(3.34)**
 :green_circle: | app/client/src/workers/validations.ts | 80.31 **(-0.08)** | 70.63 **(-0.7)** | 75.76 **(1.57)** | 81.09 **(-0.26)**</details>


app/client/src/workers/validations.ts
```diff
@@ -291,6 +291,8 @@ export function getExpectedType(config?: ValidationConfig): string | undefined {
       return `base64 encoded image | data uri | image url`;
     case ValidationTypes.SAFE_URL:
       return "URL";
+    case ValidationTypes.CSV:
+      return "CSV";
```
riodeuno, line 295: I was wondering if we should check for `allowedValues` first before returning CSV?

app/client/src/widgets/TableWidget/widget/propertyConfig.ts
```diff
@@ -603,6 +603,15 @@ export default [
                   ],
                   isBindProperty: true,
                   isTriggerProperty: false,
+                  validation: {
```
riodeuno, line 606: How about adding an `expected` `type` and `example`? We can then use that as a default to display in the expected example and structure popups. We've done the same in other validations.

riodeuno (commented):
Code looks fine to me. I have a few comments on the details of the validation.

app/client/src/workers/validations.ts
aswathkk: @riodeuno I didn't get you.
If you were saying about checking for `allowedValues`, I'm doing it [here](https://github.com/appsmithorg/appsmith/pull/9900/files#diff-e7e2b5776f9525df854356addbab33804e25601da70d2e77697b22ba8f1e1016R954)

app/client/src/widgets/TableWidget/widget/propertyConfig.ts
aswathkk: I'm having a generic expected type and example [here](https://github.com/appsmithorg/appsmith/pull/9900/files#diff-4882093df074ae9b4277fb1d65962247233ca768313db4e4c676bccc8b55970eR161)

app/client/src/workers/validations.ts
riodeuno: @aswathkk This is regarding what is shown in the Expected structure and Expected example popup in the property pane. If we can use the `allowedValues` we can increase the accuracy of the information shared with the user regarding what values can be provided for the property. We can fallback to `CSV` if `allowedValues` aren't available.

aswathkk: @riodeuno Got it.
I'm just thinking about the format which needs to be shown.
How about this format: `CSV<'BOLD' | 'ITALIC'>`

app/client/src/widgets/TableWidget/widget/propertyConfig.ts
riodeuno: Yes. I was wondering about:
- Do we need to add an `expected` property in the validation configurations which can specify that `bold`, `italic` and `underline` are the available options.
- If we add an `expected` property in the `ValidationTypes.CSV`, will it pick it up in the expected structure popup correctly?

app/client/src/workers/validations.ts
riodeuno: Hmm, this is a good point. How do we represent this? Shall we try getting inputs from the team?

app/client/src/widgets/TableWidget/widget/propertyConfig.ts
aswathkk: > Do we need to add an expected property in the validation configurations which can specify that bold, italic and underline are the available options.

We don't have to do that here since it will be picked up by the generic validation

> If we add an expected property in the ValidationTypes.CSV, will it pick it up in the expected structure popup correctly?
No. This needs to be added. I will make the necessary changes.

app/client/src/workers/validations.ts
aswathkk: Created a thread in Slack to get feedback.

riodeuno (dismissed):

vivekverma2312:
@aswathkk In table widget sub menu Font style if we switch between JS option the selected option gets reset
https://watch.screencastify.com/v/Qd8M6OWbuKrgmYYBtyC1

riodeuno (approved):
Looks fine to me! One note is that we should discourage `ValidationTypes.CSV` from being used henceforth.

@sbalaji1192 Let's use starndard `ValidationTypes.ARRAY` instead of `CSV` in Table Widget V2. Can you also comment on whether we need `ValidationTypes.CSV` in the InputWidget?

vivekverma2312:
@aswathkk For drop-down widgets error is not coming for Label Font style for random text
Screenshot 2022-03-30 at 11 53 25 AM

sbalaji1192:
> Looks fine to me! One note is that we should discourage `ValidationTypes.CSV` from being used henceforth.
> 
> @sbalaji1192 Let's use standard `ValidationTypes.ARRAY` instead of `CSV` in Table Widget V2. Can you also comment on whether we need `ValidationTypes.CSV` in the InputWidget?

@riodeuno We are introducing a Select type column and the developer can give options in CSV format as well. But we use the validation function to validate. Regarding the Input widget, In v2 we're not using CSV format. (ideally, we should be using it for label style property, but in V2 it's not a js property somehow. will check with @dilippitchika .)

app/client/src/workers/validations.test.ts
```diff
@@ -1368,6 +1368,77 @@ describe("Validate Validators", () => {
...
   });
+
+  it("correctly validates CSV validation", () => {
+    const inputs = [
+      "a",
+      ["a", "b"],
+      "x",
+      ["a", "b", "x"],
+      "a,b",
+      "a,b,x",
```
Rishabh-Rathod, line 1379: 1. This seems to me an invalid Array but we used to allow this earlier and maybe hence we allow it now, i believe.
Screenshot 2022-03-30 at 1 29 51 PM

2. This seems to me a valid Array but we do not allow this now and earlier we allowed it. Are we helping users in migrating here?

Screenshot 2022-03-30 at 1 30 00 PM


3. Earlier an extra comma was valid but now it is invalid. 
Screenshot 2022-03-30 at 1 30 38 PM

4. Earlier font style having invalid values with valid values, used to work. Now that is not the case, Will we help users to migrate here?
Screenshot 2022-03-30 at 1 30 47 PM

Rishabh-Rathod (commented):
Can you please look into the below comments, are we intentionally making these changes?

app/client/src/workers/validations.test.ts
riodeuno: All of these are good catches @Rishabh-Rathod ! @aswathkk let's include these in the unit tests as well. We should not break existing apps, migrated where it is necessary and keep things working as they are where we can't migrate.

Rishabh-Rathod: 1. As the user might not be always typing an array in such fields, The error message "Value is not allowed in this array" doesn't seem to be correct for all the cases.

2. **Expected structure** mentions only Array but we also allow "BOLD" or "ITALIC" string literal values.

Screenshot 2022-03-30 at 1 54 25 PM

Hence information in the `expected structure` is not accurate.

3. If the error message could be more helpful for users to correct their mistakes easily, that would be a better experience. ( General discussion )

As a user POV, my first question would be. 
1. What are the allowed values for this field?
We know that the expected structure has all allowed values. Maybe when there is an validation error we should show expected structure and example both in non-hidden state. ( currently, we keep hidden )

aswathkk:
Thanks for pointing out these issues, @Rishabh-Rathod 

Migration is not possible. Because this field used to accept anything!
For example, the following value in the Font style field of the Text widget will bold and italicise the text. But, we can’t migrate this:
```
Anything string that contains BOLD in it. If you wanna make it {{String.fromCharCode(73)}}{{"TA"}}LIC include it as a substring
```

  @riodeuno Now, I feel that the field accepting CSV values and showing as an array in the Expected structure might confuse the users.  
  
  What if we only accept array values?
  
   To not break existing applications, we will simply pass through the incoming values as it is to the widget rather than passing the default value on validation failure.  This will show errors on edit mode and allow users to manually migrate to the newly accepted format without breaking the look of their application.

Rishabh-Rathod:
> What if we only accept array values?
> To not break existing applications, we will simply pass through the incoming values as it is to the widget rather than passing the default value on validation failure

This was the behavior even i was expecting for this case. 

Ideally, any Appsmith new changes should not start showing errors to users in their apps that had no errors earlier. 

If we want to go with stricter validations and show an error to the user for old valid values (which won't be valid anymore) at least we should inform users why they see such errors now and what they could do to fix it.
cc: @aswathkk

aswathkk:
Closing this since we are disabling the JS toggle button when the value can't be converted from JS code to UI.

PR #9909, opened 2021-12-21
chore: Add migration for new MongoDB mock endpoint

sharat87:


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
```diff
@@ -4677,4 +4677,19 @@ public void migrateFirestorePaginationData(MongockTemplate mongockTemplate) {
             mongockTemplate.save(firestoreAction);
         }
     }
+
+    @ChangeSet(order = "106", id = "update-mongodb-mockdb-endpoint", author = "")
+    public void updateMongoMockdbEndpoint(MongockTemplate mongockTemplate) {
+        mongockTemplate.updateMulti(
+                query(where("datasourceConfiguration.endpoints.host").is("mockdb.swrsq.mongodb.net")),
+                update("datasourceConfiguration.endpoints.$.host", "mockdb.kce5o.mongodb.net"),
```
mohanarpit, line 4685: We aren't using CNAME aliases?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
```diff
@@ -4677,4 +4677,19 @@ public void migrateFirestorePaginationData(MongockTemplate mongockTemplate) {
...
+                Datasource.class
+        );
+        mongockTemplate.updateMulti(
+                query(where("datasourceConfiguration.properties.value").is("mongodb+srv://mockdb_super:****@mockdb.swrsq.mongodb.net/movies")),
```
mohanarpit, line 4689: Is the `****` correct? Asking for my information if this will update the password in the database?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
sharat87: Nope, not possible with MongoDB Atlas SRV endpoints. See: <https://jira.mongodb.org/browse/CDRIVER-2556?focusedCommentId=2722051&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2722051>.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
sharat87: Yes, it's used as a marker to mask the password in the URI that the user sees/edits. The password is stored in a separate field in the datasource configuration object. Encrypted, of course.

mohanarpit (approved):

PR #9916, opened 2021-12-21
fix: Switched reference from package name to plugin name for import export

### LINKED ISSUE #9876: [Bug]: Airtable breaks on import ,export and Duplicate

### Current Behavior

When the application has an Airtable datasource and we import the same application the API breaks 

### Steps To Reproduce

1) Add an Airtable datasource 
2) Bind a widget 
3) Export an application 
4) Import an application 
and observe the configuration mismatch

Screenshot 2021-12-20 at 3 41 14 PM

### Environment

Release

### Version

Self-Hosted

nidhi-nair:
Fixes #9876


abhvsn (approved):
LGTM

PR #9947, opened 2021-12-22
fix: Fix for on page load action in forked application

### LINKED ISSUE #9945: [Bug]: The Forked app fails to load data on page load

### Current Behavior

 The Forked app fails to load data on page load
[LOOM DEMO](https://www.loom.com/share/170155be60a04a0ea01dc69d75181c80)

### Steps To Reproduce

1. fork any of the application 
2. On the canvas verify the  data binded for the widget

### Environment

Release

### Version

Cloud

abhvsn:
## Description

> Because of the default actionId was missing in layoutOnLoad action inside page layout for the forked application, forked application on page action was not getting executed. This PR assigns the newly created action id as the defaultActionId. 

Fixes #9945 

## Type of change

## How Has This Been Tested?

> Junit TC
> Manual test

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ExamplesOrganizationClonerCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ExamplesOrganizationClonerCEImpl.java
@@ -385,7 +385,9 @@ public class ExamplesOrganizationClonerCEImpl implements ExamplesOrganizationClo
         for (final Set<DslActionDTO> actionSet : layout.getLayoutOnLoadActions()) {
             for (final DslActionDTO actionDTO : actionSet) {
                 if (actionIdsMap.containsKey(actionDTO.getId())) {
-                    actionDTO.setId(actionIdsMap.get(actionDTO.getId()));
+                    final String srcActionId = actionDTO.getId();
+                    actionDTO.setId(actionIdsMap.get(srcActionId));
+                    actionDTO.setDefaultActionId(actionIdsMap.get(srcActionId));
                     shouldSave = true;
                 } else {
                     log.error(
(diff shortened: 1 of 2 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ExamplesOrganizationClonerCEImpl.java
```diff
@@ -385,7 +385,9 @@ private boolean updateOnLoadActionsWithNewActionIds(Map<String, String> actionId
         for (final Set<DslActionDTO> actionSet : layout.getLayoutOnLoadActions()) {
             for (final DslActionDTO actionDTO : actionSet) {
                 if (actionIdsMap.containsKey(actionDTO.getId())) {
-                    actionDTO.setId(actionIdsMap.get(actionDTO.getId()));
+                    final String srcActionId = actionDTO.getId();
+                    actionDTO.setId(actionIdsMap.get(srcActionId));
+                    actionDTO.setDefaultActionId(actionIdsMap.get(srcActionId));
```
abhvsn, line 390: Earlier this statement was missing which was causing the issue as defaultActionId was not getting updated correctly

sharat87: Sorry, a little out of the loop here, what's a default action id?

abhvsn: When we implemented the git, every resource is going to have a default resource Id from where this is originated. This is useful for 2 reasons,
1. To minimise the DB calls (In current case if the defaultActiuonId is not saved in layoutOnLoadAction then we have to make a call to newActionService to find out what is the defaultActionId for this)
2. Connect resources across branches

In git client is going to be aware of single ID for each resource.

abhvsn: So while responding to client we just replace the actionId with defaultActionId in current case, considering client don't have to worry about different IDs we have in server side

sharat87 (approved):

PR #9961, opened 2021-12-23
chore: Fixed dot replacement policy for mongo

### LINKED ISSUE #5868: [Bug] MappingException: Map key Sq.Yds. Per Carton contains dots but no replacement was configured! Make sure map keys do...
Sentry Issue: [APPSMITH-VA](https://sentry.io/organizations/appsmith/issues/2424076937/?referrer=github_integration)

```
MappingException: Map key Sq.Yds. Per Carton contains dots but no replacement was configured! Make sure map keys don't contain dots in the first place or configure an appropriate replacement!
    at org.springframework.data.mongodb.core.convert.MappingMongoConverter.potentiallyEscapeMapKey(MappingMongoConverter.java:873)
```

nidhi-nair:
Fixes #5868


app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/MongoConfig.java
```diff
@@ -107,6 +107,7 @@ public MappingMongoConverter mappingMongoConverter(DefaultTypeMapper<Bson> typeM
         MappingMongoConverter converter = new MappingMongoConverter(NoOpDbRefResolver.INSTANCE, context);
         converter.setTypeMapper((MongoTypeMapper) typeMapper);
         converter.setCustomConversions(mongoCustomConversions());
+        converter.setMapKeyDotReplacement("-APPSMITH-DOT-REPLACEMENT-");
```
mohanarpit, line 110: I'm slightly wary of this change. Since we'll be replacing all the dots in Mongo keys with a placeholder, the Mongo converter will traverse all keys of all objects in order to find dot characters & replace them. Will this cause a performance issue?

nidhi-nair: The mongo converter already does these steps, apart from the replacement itself. If you look at the stack trace of the issue, the converter was attempting to find if this key has any dots, identified that it does, and failed to find a replacement which is why it threw an error. By setting this replacement we're simply allowing the string replace to happen. Same thing happens with unescaping.

mohanarpit (approved):

PR #9963, opened 2021-12-23
fix: add query params from datasource to URI

### LINKED ISSUE #9962: [Bug]: REST API plugin: query params in datasource is not taking effect

### Current Behavior

Query params in datasource is not taking effect.

### Steps To Reproduce

1. Create REST API datasource.
2. Add a query query param to the datasource.
3. Execute an API action with the datasource and check that the query param is not passed over to the endpoint.

### Environment

Production

### Version

Cloud

sumitsum:
## Description
- Add query params from datasource to final URI used for REST API query.

Fixes #9962 

## Type of change

## How Has This Been Tested?
- JUnit TC
- Manual

## Checklist:

### DIFF
--- a/app/server/appsmith-plugins/restApiPlugin/src/test/java/com/external/plugins/RestApiPluginTest.java
+++ b/app/server/appsmith-plugins/restApiPlugin/src/test/java/com/external/plugins/RestApiPluginTest.java
@@ -929,4 +929,34 @@ public class RestApiPluginTest {
                 })
                 .verifyComplete();
     }
+
+    @Test
+    public void testQueryParamsInDatasource() {
+        DatasourceConfiguration dsConfig = new DatasourceConfiguration();
+        dsConfig.setUrl("https://postman-echo.com/post");
+
+        ActionConfiguration actionConfig = new ActionConfiguration();
+        actionConfig.setHeaders(List.of(new Property("content-type", "application/json")));
+        actionConfig.setHttpMethod(HttpMethod.POST);
+        String requestBody = "body";
+        actionConfig.setBody(requestBody);
+        actionConfig.setEncodeParamsToggle(true);
+
+        List<Property> queryParams = new ArrayList<>();
+        queryParams.add(new Property("query_key", "query val")); /* encoding changes 'query val' to 'query+val' */
+        dsConfig.setQueryParameters(queryParams);
+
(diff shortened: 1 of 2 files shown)


app/server/appsmith-plugins/restApiPlugin/src/test/java/com/external/plugins/RestApiPluginTest.java
```diff
@@ -929,4 +929,34 @@ public void testHintMessageForDuplicateParamWithOneInstanceEachInActionAndDsConf
...
+        dsConfig.setQueryParameters(queryParams);
+
+        Mono<ActionExecutionResult> resultMono = pluginExecutor.executeParameterized(null, new ExecuteActionDTO(), dsConfig, actionConfig);
+
+        StepVerifier.create(resultMono)
+                .assertNext(result -> {
+                    assertTrue(result.getIsExecutionSuccess());
+                    assertNotNull(result.getBody());
+
+                    String expected_url = "\"https://postman-echo.com/post?query_key=query+val\"";
```
nidhi-nair, line 956: Shouldn't this be encoded to %20? There's this whole set of opinions on how Java handles this but I think we handle this as %20 some other place. At least, especially when we do the url encoding from client side. Could you please check this?

nidhi-nair (changes requested):

app/server/appsmith-plugins/restApiPlugin/src/test/java/com/external/plugins/RestApiPluginTest.java
sumitsum: @nidhi-nair thanks for pointing this out. It seems that it could be either based on the use case (came as a surprise to me). Please check out this link: https://stackoverflow.com/questions/1634271/url-encoding-the-space-character-or-20 

Quoting from w3schools:
```
URLs cannot contain spaces. URL encoding normally replaces a space with a plus (+) sign or with %20.
```
Ref: https://www.w3schools.com/tags/ref_urlencode.ASP

nidhi-nair: TIL! :) Nice clarification of expectations.

nidhi-nair (approved):

PR #9975, opened 2021-12-23
fix: 9648 The Update query fails to update for all matching Document in mongo 

### LINKED ISSUE #9648: [Bug]: The Update query fails to update for all matching Document in mongo

### Current Behavior

When user try to update all the matching document at once the update happens only for 1 data in the mongo database
[LOOM DEMO](https://www.loom.com/share/ad32bd8ae7484ac3aeeed9ca7f052eef)

### Steps To Reproduce

1. go to the database
2. add the mongo database
3.  Select the update query 
4. add the query in such a way that the user is trying to update 2 records with the same name
5. Now select the All matching documents in the limit dropdown 
6. Run the query 
7. The update happens for only 1 data

### Environment

Production

### Version

Cloud

leotom2000:
## Description

The Update query fails to update for all matching Document in mongo 
 It works fine for single and fails for multiple records.
The issue is related to the Mongo version change syntax of handling multi records update.
Implemented the updates to fix the issue

Fixes #9648 

## Type of change

## How Has This Been Tested?

> Manually verified by verifying the data in DB during development
> Unit Test cases was modified to take up the new change

## Checklist:

### DIFF
--- a/app/server/appsmith-plugins/mongoPlugin/src/test/java/com/external/plugins/MongoPluginTest.java
+++ b/app/server/appsmith-plugins/mongoPlugin/src/test/java/com/external/plugins/MongoPluginTest.java
@@ -1201,7 +1201,7 @@ public class MongoPluginTest {
         setValueSafelyInFormData(configMap, COMMAND, "UPDATE");
         setValueSafelyInFormData(configMap, COLLECTION, "users");
         setValueSafelyInFormData(configMap, UPDATE_QUERY, "{ name: \"Alden Cantrell\" }");
-        setValueSafelyInFormData(configMap, UPDATE_OPERATION, "{ age: 31 }}");
+        setValueSafelyInFormData(configMap, UPDATE_OPERATION, "{ $set: { age: 31 }}}");
         setValueSafelyInFormData(configMap, UPDATE_LIMIT, "SINGLE");
 
         actionConfiguration.setFormData(configMap);
@@ -1235,7 +1235,7 @@ public class MongoPluginTest {
         setValueSafelyInFormData(configMap, COLLECTION, "users");
         // Query for all the documents in the collection
         setValueSafelyInFormData(configMap, UPDATE_QUERY, "{}");
-        setValueSafelyInFormData(configMap, UPDATE_OPERATION, "{ updatedByCommand: true }}");
+        setValueSafelyInFormData(configMap, UPDATE_OPERATION, "{ $set: { updatedByCommand: true }}}");
         setValueSafelyInFormData(configMap, UPDATE_LIMIT, "ALL");
 
         actionConfiguration.setFormData(configMap);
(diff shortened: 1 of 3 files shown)


app/server/appsmith-plugins/mongoPlugin/src/test/java/com/external/plugins/MongoPluginTest.java
```diff
@@ -1235,7 +1235,7 @@ public void testUpdateManyFormCommand() {
         setValueSafelyInFormData(configMap, COLLECTION, "users");
         // Query for all the documents in the collection
         setValueSafelyInFormData(configMap, UPDATE_QUERY, "{}");
-        setValueSafelyInFormData(configMap, UPDATE_OPERATION, "{ $set: { updatedByCommand: true }}}");
+        setValueSafelyInFormData(configMap, UPDATE_OPERATION, "{ updatedByCommand: true }}");
```
nidhi-nair, line 1238: Are we changing how users need to enter data into the query here @leotom2000 ?

leotom2000: Yes, the syntax of update command needs $set attribute, this has been handled in the unit test earlier,
since we have implemented in the main code now, this became a duplicate in the generated query, so we have removed from here.

nidhi-nair: Will this not require users to change the way they write their actions? Could you please give an example of what the end user would end up using?

leotom2000: The below syntax will work, as $set is added in code,
 but if we need to consider other update options, we need to decide,
 I tested with the existing $inc, it is not working in this scenario even for single document.
 Will discuss before closing this issue


image

leotom2000: If we may need to utilise other update operators also, we need to follow the below syntax.
update operators : ($currentDate, $inc, $min, $max, $mul, $rename, $set, $setOnInsert, $unset)


image

leotom2000: A change to one constants fixed this issue, the issue was it was not able to resolve whether the update is going to be multi or single. After the change to the constant it worked comparing the dynamic input.
The unit tests for single and multi updates were already handled. The new change did not affect Tests is because, the same variable was used in both places like the simulated input and in the place where it is used to evaluate if it is multi or single. so even after the code change (constant) it continues to work.

nidhi-nair (approved):

Issue #9513, opened 2021-12-02
[Bug]: Using external MongoDB will still try to connect and configure local MongoDB


sharat87:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When a fat container is configured with an external MongoDB URI, the entrypoint script should not try to connect and initialize a local MongoDB Replica Set. Instead, it should connect to this external MongoDB, check that it's a replica set already, and if not, do an `rs.initiate()` on it.

User report: https://discord.com/channels/725602949748752515/760761686549463060/915781460810666095

### Steps To Reproduce

Start  fat container with a `docker.env` containing a MongoDB URI pointing to a locally running MongoDB server. The fat container's entrypoint will still try to initialize a separate MongoDB instance and have the server try to connect to this external MongoDB server.

### Environment

Production

### Version

Self-Hosted

sum35h:
@sharat87  I've tested and verified this with appsmith/appsmith-ce:latest. The issue is resolved now.

PR #10010, opened 2021-12-27
fix: Refactor init mongodb flow to check and enable replica set for external mongodb 

geekup-legodevops:
## Description
- Add flow to check if MongoDB URI pointing to local service or not. Perform local mongodb init as normal
- Remove unnecessary environment variables for MongoDB credentials and directly extract credentials from URI
- Add prefix /api/v1 in the API_BASE_URL 
- Add flow to check external MongoDB is set as replica set or not. If yes, initiate the replica set. If not, print out error message and stop installation.

Fixes # (issue)

[[Bug]: Using external MongoDB will still try to connect and configure local MongoDB](https://github.com/appsmithorg/appsmith/issues/9513)

## Type of change

## Checklist:


deploy/docker/entrypoint.sh
```diff
@@ -2,7 +2,35 @@
...
 set -e
 
-check_initialized_db() {
+get_variable_mongodb() {
+  echo "Get variable mongodb"
+  MONGODB_PROTOCOL=$(echo "$APPSMITH_MONGODB_URI" | grep -oP "[a-z+]+(?=\:\/\/)") 
+  MONGODB_USERNAME=$(echo "$APPSMITH_MONGODB_URI" | grep -oP "(?<=\:\/\/)(.*\s?)(?=\:)")
+  MONGODB_PASSWORD=$(echo "$APPSMITH_MONGODB_URI" | grep -oP "\w+(?=@)")
+  MONGODB_HOST=$(echo "$APPSMITH_MONGODB_URI" | grep -oP "[a-z0-9.]+(?=\/)")
+  MONGODB_DATABASE=$(echo "$APPSMITH_MONGODB_URI" | grep -oP "\w+(?!\S)")
```
sharat87, line 11: Can you confirm if this works when the URI has query params with `?authSource=admin` as well as when there are no query params at all?

deploy/docker/entrypoint.sh
```diff
@@ -12,42 +40,47 @@ check_initialized_db() {
...
     sleep 10
     mongo "$APPSMITH_MONGODB_URI" --eval 'rs.initiate()'
     mongod --dbpath "$MONGO_DB_PATH" --shutdown || true
   fi
+
+  if [[ $isUriLocal -gt 0 ]]; then
+    # Check mongodb cloud replica set
+    echo 'Check mongodb cloud replica set'
+    responseStatus=$(mongo "$APPSMITH_MONGODB_URI" --eval "rs.status()" | grep ok | xargs)
+    okString='ok : 1'
```
sharat87, line 72: I just ran this command and the output shows `ok: 1`, not `ok : 1`, is that okay? Will it still match?

deploy/docker/entrypoint.sh
```diff
@@ -67,7 +100,7 @@ configure_supervisord() {
   cp -f "$SUPERVISORD_CONF_PATH/application_process/"*.conf /etc/supervisor/conf.d
 
   # Disable services based on configuration
-  if [[ "$APPSMITH_MONGODB_URI" = "mongodb://appsmith:$MONGO_INITDB_ROOT_PASSWORD@localhost/appsmith" ]]; then
+  if [[ "$APPSMITH_MONGODB_URI" = "mongodb://appsmith:$MONGODB_PASSWORD@$MONGODB_HOST/appsmith" && $isUriLocal -eq 0 ]]; then
```
sharat87, line 103: Isn't checking `isUriLocal` enough here? In the first check, we are basically checking is username is `appsmith` and database name is `appsmith`. The other two things, password and host, are basically extracted from `APPSMITH_MONGODB_URI` only, so they'll anyway match.

sharat87 (changes requested):
In addition to these comments, I _think_ there's a conflict as well. Please merge release back into this branch just in case.

deploy/docker/entrypoint.sh
geekup-legodevops: Hi @sharat87 have you run the whole command with the 'xargs' command ? Last time we checked, the 'xargs' command will make the `ok: 1` string to `ok : 1`

sharat87: Oh I see, cool, thanks for confirming.

sharat87 (approved):

MichaelLeeHobbs:
#11418 Bug on line 125

```
```

Should be

```
  if [[ $isUriLocal -eq 0 ]]; then
```

PR #10036, opened 2021-12-29
chore: Don't take client cancellation for git flows into account to keep the repo and DB in sane state   

### LINKED ISSUE #10018: [Bug] Handle timeout error gracefully for git-sync
Today we are directly throwing the error after the timeout, this needs to be handled as cloning, pushing (communication with remote) might take more time

abhvsn:
## Description

> Git APIs are slow, because of the interaction with remote repo and working with the file system which includes the serialisation and de-serialisation part. During these operations if client cancels the flow, DB or git repo may end up in bad state as execution on server side takes client subscription into account. To resolve this we are leveraging the MonoSink which don't take subscription cancellation into account. 

Fixes #10018 

## Type of change

## How Has This Been Tested?

> Manual testing
> Junit TC

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1977,9 +1987,6 @@ private boolean isInvalidDefaultApplicationGitMetadata(GitApplicationMetadata gi
                                                    String errorType,
                                                    String errorMessage,
                                                    Boolean repoType) {
-        if (!analyticsService.isActive()) {
```
abhvsn, line 1980: We are already doing this check in analytics service

app/server/appsmith-server/src/main/java/com/appsmith/server/services/GitServiceImpl.java
```diff
@@ -32,11 +31,12 @@ public GitServiceImpl(UserService userService,
...
-                          CloudServicesConfig cloudServicesConfig,
                           AnalyticsService analyticsService,
+                          GitCloudServicesUtils gitCloudServicesUtils,
                           GitDeployKeysRepository gitDeployKeysRepository) {
-        super(userService, userDataService, sessionUserService, applicationService, applicationPageService, newPageService, newActionService, actionCollectionService, fileUtils, importExportApplicationService, gitExecutor, responseUtils, emailConfig, commonConfig, configService, cloudServicesConfig, analyticsService, gitDeployKeysRepository);
+
+        super(userService, userDataService, sessionUserService, applicationService, applicationPageService,
+                newPageService, newActionService, actionCollectionService, fileUtils, importExportApplicationService,
+                gitExecutor, responseUtils, emailConfig, analyticsService, gitCloudServicesUtils, gitDeployKeysRepository);
     }
```
AnaghHegde, line 41: This is not indented. Can we move back to the way it was existing ?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -408,17 +413,7 @@ private String getReason() {
...
-                                branchedApplication.getOrganizationId(),
-                                defaultApplicationId,
-                                branchedApplication.getId(),
-                                AppsmithError.INVALID_GIT_CONFIGURATION.getTitle(),
-                                AppsmithError.INVALID_GIT_CONFIGURATION.getMessage("Unable to find the git " +
-                                        "configuration, please configure the your application to use version control service"),
-                                branchedApplication.getGitApplicationMetadata().getIsRepoPrivate()
-                        ).flatMap(user -> Mono.error(new AppsmithException(AppsmithError.INVALID_GIT_CONFIGURATION, "Unable to find the git " +
-                                "configuration, please configure the your application to use version control service")));
+                        return Mono.error(new AppsmithException(AppsmithError.INVALID_GIT_CONFIGURATION, GIT_CONFIG_ERROR));
```
AnaghHegde, line 416: Why are we removing this state from the analytics ?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -444,27 +439,28 @@ private String getReason() {
...
-                                    childApplication.getOrganizationId(),
-                                    defaultApplicationId,
-                                    childApplication.getId(),
-                                    ((RepositoryNotFoundException) e).getClass().getName(),
-                                    e.getMessage(),
-                                    childApplication.getGitApplicationMetadata().getIsRepoPrivate()
-                            ).flatMap(user -> Mono.error(new AppsmithException(AppsmithError.GIT_ACTION_FAILED, "commit", e)));
-                        }
-                        return Mono.error(new AppsmithException(AppsmithError.GIT_FILE_SYSTEM_ERROR, e.getMessage()));
+                        return Mono.error(e);
```
AnaghHegde, line 447: Same as above. Why are we removing this error state from the analytics ?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/GitServiceImpl.java
abhvsn: Can you elaborate on the issue here?

AnaghHegde: Either we should declare all the parameters in same line or each of them in a new line. Does that sounds good ?

AnaghHegde (commented):
Couple of minor nits. Rest all looks good to me.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/GitServiceImpl.java
abhvsn: IMO this is a special case I tried doing both of the mentioned ways but both looked cluttered to me so finally opted for this option but if this is absolutely required then I will go with one param on each line what say?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
abhvsn: IMO this is bad state the application currently in and will not be useful in analytics funnel. We should track these on sentry if required though

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
abhvsn: catch will never get called we will have to use `onErrorResume` for this. But again is this useful to track in analytics funnel or just a bug in our setup?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: yeah agreed.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/GitServiceImpl.java
AnaghHegde: Yeah i agree. Lets keep it this way.

AnaghHegde (approved):

PR #9944, opened 2021-12-22
fix: Users unable to drag a column outside of viewport of a property pane to rearrange column order

vicky-primathon:
## Description
Allow users to drag a column inside of viewport of property pane to rearrange column order

Fixes #4266

## Type of change
- Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

## Test coverage results :test_tube:
<details><summary>:red_circle: Total coverage has decreased</summary>

    // Code coverage diff between base branch:release and head branch: fix/drag-and-scroll 
Status | File | % Stmts | % Branch | % Funcs | % Lines 
 -----|-----|---------|----------|---------|------ 
 :red_circle: | total | 55.2 **(-0.02)** | 36.99 **(-0.03)** | 34.07 **(0)** | 55.72 **(-0.01)**
 :red_circle: | app/client/src/components/ads/DraggableList.tsx | 37.97 **(-6.29)** | 19.05 **(-14.28)** | 21.43 **(0)** | 41.79 **(-9.23)**
 :red_circle: | app/client/src/utils/autocomplete/TernServer.ts | 52.71 **(-0.23)** | 40.83 **(-0.84)** | 36.21 **(0)** | 56.74 **(-0.25)**</details>

PR #10056, opened 2021-12-29
fix: Users unable to drag a column outside of viewport of a property pane to rearrange column order

vicky-primathon:
Reverts appsmithorg/appsmith#9944


somangshu (approved):
LGTM, verified the revert

PR #10157, opened 2022-01-05
feat: Add Helm configMap to update application configuration

geekup-legodevops:
## Description
- Adding Helm ConfigMap for users who deploy Appsmith by Helm Chart to change application configuration
- Refactor the flow that being used to generate environment variables for supporting predefined environment variables when initializing the container
## Type of change

## Checklist:


deploy/docker/templates/docker.env.sh
```diff
@@ -2,88 +2,79 @@
...
 set -o nounset
 
-MONGO_PASSWORD="$1"
-ENCRYPTION_PASSWORD="$2"
-ENCRYPTION_SALT="$3"
-
 cat <<EOF
 # Sentry
-APPSMITH_SENTRY_DSN=
+APPSMITH_SENTRY_DSN=${APPSMITH_SENTRY_DSN:-\$APPSMITH_SENTRY_DSN}
```
sharat87, line 7: Do these values need to be wrapped in `"` characters? If the current value of this coming from ConfigMap contains a special character like `'` or a space character, does it not get interpreted differently here?

deploy/docker/entrypoint.sh
```diff
@@ -2,6 +2,102 @@
...
+      echo ''
+    )
+    bash "$TEMPLATES_PATH/docker.env.sh" > "$TEMPLATES_PATH/docker.env.temp"
+    
+    if [[ -z "${APPSMITH_DISABLE_TELEMETRY}" ]]; then
+      export APPSMITH_DISABLE_TELEMETRY=false
+    fi
+
+    if [[ -z "${APPSMITH_MONGODB_URI}" ]]; then
+      export APPSMITH_MONGODB_URI=mongodb://appsmith:$AUTO_GEN_MONGO_PASSWORD@localhost/appsmith
```
sharat87, line 33: The password might contain special characters. Wrapping in `"` won't fix the case where the password contains `"` itself, but it solves 90% of the cases, including `'` or the space character.

```suggestion
      export APPSMITH_MONGODB_URI="mongodb://appsmith:$AUTO_GEN_MONGO_PASSWORD@localhost/appsmith"
```

deploy/docker/entrypoint.sh
```diff
@@ -2,6 +2,102 @@
...
+    fi
+
+    if [[ -z "${APPSMITH_REDIS_URL}" ]]; then
+      export APPSMITH_REDIS_URL=redis://127.0.0.1:6379
+    fi
+
+    if [[ -z "${APPSMITH_ENCRYPTION_PASSWORD}" ]]; then
+      export APPSMITH_ENCRYPTION_PASSWORD=$AUTO_GEN_ENCRYPTION_PASSWORD
```
sharat87, line 41: Same thing as above.

```suggestion
      export APPSMITH_ENCRYPTION_PASSWORD="$AUTO_GEN_ENCRYPTION_PASSWORD"
```

deploy/docker/entrypoint.sh
```diff
@@ -2,6 +2,102 @@
...
+    fi
+
+    if [[ -z "$APPSMITH_ENCRYPTION_SALT" ]]; then
+      export APPSMITH_ENCRYPTION_SALT=$AUTO_GEN_ENCRYPTION_SALT
```
sharat87, line 45: Same thing as above.

```suggestion
      export APPSMITH_ENCRYPTION_SALT="$AUTO_GEN_ENCRYPTION_SALT"
```

deploy/docker/templates/docker.env.sh
sharat87: Also, isn't this the same as:

```sh
APPSMITH_SENTRY_DSN=${APPSMITH_SENTRY_DSN:-}
```

That is to say, if the variable isn't defined already, we set it to empty instead of the default value. But if the variable isn't set already, the currently value is also empty so this should behave the same, right?

deploy/docker/entrypoint.sh
```diff
@@ -2,6 +2,102 @@
...
+    fi
+    cat "$TEMPLATES_PATH/docker.env.temp" | envsubst "$(printf '$%s,' $(env | grep -Eo '^APPSMITH_[A-Z0-9_]+'))" | sed -e 's|\$APPSMITH_[A-Z0-9_]*||g' > "$ENV_PATH"
```
sharat87, line 47: Can you elaborate why we need to replace the env variable values into the `docker.env` file? Can we not have all the variables look like `APPSMITH_ENV_VAR="${APPSMITH_ENV_VAR:-}"`?

deploy/helm/README.md
```diff
@@ -163,13 +163,58 @@ helm install \
...
+| `applicationConfig.APPSMITH_OAUTH2_GOOGLE_CLIENT_SECRET`| `""`									|
+| `applicationConfig.APPSMITH_OAUTH2_GITHUB_CLIENT_ID`		| `""`									|
+| `applicationConfig.APPSMITH_OAUTH2_GITHUB_CLIENT_SECRET`| `""`									|
+| `applicationConfig.APPSMITH_SEGMENT_KEY`								| `""`									|
+| `applicationConfig.APPSMITH_RAPID_API_KEY_VALUE`				| `""`									|
+| `applicationConfig.APPSMITH_MARKETPLACE_ENABLED`				| `""`									|
+| `applicationConfig.APPSMITH_OPTIMIZELY_KEY`							| `""`									|
+| `applicationConfig.APPSMITH_ALGOLIA_API_ID`							| `""`									|
+| `applicationConfig.APPSMITH_ALGOLIA_API_KEY`						| `""`									|
+| `applicationConfig.APPSMITH_ALGOLIA_SEARCH_INDEX_NAME`	| `""`									|
```
sharat87, line 189: Let's remove these from the docs. These are kinda internal variables.

deploy/helm/README.md
```diff
@@ -163,13 +163,58 @@ helm install \
...
 ```
 
 *Tip: You can use the default [values.yaml](https://github.com/appsmithorg/appsmith/blob/release/deploy/helm/values.yaml)*
 
+### Appsmith configuration
+To change Appsmith configurations, you can use configuration UI in application or update value in values.yaml(The available configurations is listed below). 
+|	Name																										|	Value									|
+|	----------------------------------------------------		|	---------------------	|
+|	`applicationConfig.APPSMITH_SENTRY_DSN`									| `""`									|
+| `applicationConfig.APPSMITH_SMART_LOOK_ID`							| `""`									|
```
sharat87, line 178: Let's remove these, they are internal.

deploy/helm/values.yaml
```diff
@@ -271,3 +271,38 @@ autoupdate:
...
+  APPSMITH_OAUTH2_GOOGLE_CLIENT_SECRET: ""
+  APPSMITH_OAUTH2_GITHUB_CLIENT_ID: ""
+  APPSMITH_OAUTH2_GITHUB_CLIENT_SECRET: ""
+  APPSMITH_SEGMENT_KEY: ""
+  APPSMITH_RAPID_API_KEY_VALUE: ""
+  APPSMITH_MARKETPLACE_ENABLED: ""
+  APPSMITH_OPTIMIZELY_KEY: ""
+  APPSMITH_ALGOLIA_API_ID: ""
+  APPSMITH_ALGOLIA_API_KEY: ""
+  APPSMITH_ALGOLIA_SEARCH_INDEX_NAME: ""
```
sharat87, line 288: Let's remove internal variables here as well.

```suggestion
  APPSMITH_OAUTH2_GOOGLE_CLIENT_ID: ""
  APPSMITH_OAUTH2_GOOGLE_CLIENT_SECRET: ""
  APPSMITH_OAUTH2_GITHUB_CLIENT_ID: ""
  APPSMITH_OAUTH2_GITHUB_CLIENT_SECRET: ""
```

sharat87 (changes requested):

deploy/docker/entrypoint.sh
```diff
@@ -2,6 +2,67 @@
...
+      echo ''
+    )
+    AUTO_GEN_ENCRYPTION_SALT=$(
+      tr -dc A-Za-z0-9 </dev/urandom | head -c 13
+      echo ''
+    )
+    bash "$TEMPLATES_PATH/docker.env.sh" "$AUTO_GEN_MONGO_PASSWORD" "$AUTO_GEN_ENCRYPTION_PASSWORD" "$AUTO_GEN_ENCRYPTION_SALT" > "$ENV_PATH"
+  fi
+
+  printenv | grep -Eo '^APPSMITH_[A-Z0-9_]+=.*|^MONGO_[A-Z0-9_]+=.*' > "$TEMPLATES_PATH/pre-define.env"
```
sharat87, line 29:
```suggestion
  printenv | grep -E '^APPSMITH_|^MONGO_' > "$TEMPLATES_PATH/pre-define.env"
```

deploy/docker/templates/docker.env.sh
```diff
@@ -1,89 +1,88 @@
...
+APPSMITH_ALGOLIA_API_KEY=${APPSMITH_ALGOLIA_API_KEY:-}
+APPSMITH_ALGOLIA_SEARCH_INDEX_NAME=${APPSMITH_ALGOLIA_SEARCH_INDEX_NAME:-}
 
 #Client log level (debug | error)
-APPSMITH_CLIENT_LOG_LEVEL=
+APPSMITH_CLIENT_LOG_LEVEL=${APPSMITH_CLIENT_LOG_LEVEL:-}
 
 # GOOGLE client API KEY
-APPSMITH_GOOGLE_MAPS_API_KEY=
+APPSMITH_GOOGLE_MAPS_API_KEY=${APPSMITH_GOOGLE_MAPS_API_KEY:-}
```
sharat87, line 42: Since we now have the `pre-define.env` file, do we still need this current value interpolation here? Like, can't we just have `APPSMITH_GOOGLE_MAPS_API_KEY=` instead?

sharat87 (commented):
Looks much better, thanks guys. Just a couple of minor questions, otherwise we are ready to merge this in!

sharat87 (approved):

deploy/docker/entrypoint.sh
```diff
@@ -2,6 +2,67 @@
...
+    )
+    bash "$TEMPLATES_PATH/docker.env.sh" "$AUTO_GEN_MONGO_PASSWORD" "$AUTO_GEN_ENCRYPTION_PASSWORD" "$AUTO_GEN_ENCRYPTION_SALT" > "$ENV_PATH"
+  fi
+
+  printenv | grep -E '^APPSMITH_|^MONGO_' > "$TEMPLATES_PATH/pre-define.env"
+ 
+  echo 'Load environment configuration'
+  set -o allexport
+  . "$ENV_PATH"
+  . "$TEMPLATES_PATH/pre-define.env"
```
WernerDreierB, line 34: Hi all, 
seems to me, this line, does not work as expcected on our kubernetes 1.21 environment. 
only default-env-vars from $ENV_PATH are actually used within the script, following "pre-define.env"-files are not picked up, even though the file exists and contains values (e.g. for mongo-db). 

See attached screenshot, where I entered some debugging points to point out env-variables while running entrypoint.sh: 
error_on_override

WernerDreierB: @sharat87 could you have a look at this?

sharat87: Hey, while I'm not sure of this without looking at the `entrypoint_wdr.sh` script, I think you are dealing with the issue fixed in this PR: https://github.com/appsmithorg/appsmith/pull/11799. We should be publishing it soon.

As a workaround in the meantime, could you try adding a backslash just before the `?` and see if that makes a difference?

WernerDreierB: @sharat87 Thank you for pointing out https://github.com/appsmithorg/appsmith/pull/11799 to me. Actually escaping Question-Marks (?) and Ampersands (&) with a backslash (or in my case double back-slash to get it through our own scripting-pipeline) worked for me. Looking forward to the PR to be released, as other special-chars might be problematic as well (as described in the PR)
Thank you

PR #10173, opened 2022-01-05
fix: ssh to http url conversion

### LINKED ISSUE #10216: Url conversion is wrong for bitbucket clients

AnaghHegde:
## Description

> This fixes the issue with the bitbucket url. In the git connected apps when there is a conflict error we show the browser supported(https) git client urls so that the user can go to the git repo and resolve the conflicts. But for the bitbucket this conversion is wrong and this PR fixes the conversion issue.  

Fixes #10216 #10217

## Type of change

## How Has This Been Tested?

- Locally 

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/GitUtils.java
```diff
@@ -28,6 +28,7 @@ public static String convertSshUrlToHttpsCurlSupportedUrl(String sshUrl) {
         return sshUrl
                 .replaceFirst("git@", "https://")
```
abhvsn, line 29: Hey we have to update the regex here to cover other cases as well https://git-scm.com/docs/git-clone#_git_urls.

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/GitUtils.java
```diff
@@ -28,6 +28,7 @@ public static String convertSshUrlToHttpsCurlSupportedUrl(String sshUrl) {
...
                 .replaceFirst("\\.com:", ".com/")
+                .replaceFirst("\\.org", ".org/")
```
abhvsn, line 31: Here **\\.org`:`** is missing. But I started to think should we address other placeholders as well?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1576,12 +1576,18 @@ private String getReason() {
 
                         localBranch.removeAll(remoteBranches);
```
abhvsn, line 1577: Can we filter the `defaultBranch` in DB from the list so as not to delete this even the same is deleted from remote? We will come up with a better approach in future though

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/GitUtils.java
AnaghHegde: Do git clients have extensions such as .in, .io kind ? There are not many git clients. We can use regex to support them but is it really required ?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: Yep this would lead to bad state. Good catch.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1575,15 +1571,29 @@ private String getReason() {
                                 .collect(Collectors.toList());
 
                         localBranch.removeAll(remoteBranches);
+                        localBranch.remove(gitApplicationMetadata.getBranchName());
+
+                        // Remove the branches which are not in remote from the list before sending
+                        gitBranchListDTOS = gitBranchListDTOS.stream()
+                                .filter(gitBranchDTO -> !localBranch.contains(gitBranchDTO.getBranchName()))
```
abhvsn, line 1578: Sorry I should have reported this earlier but why are we checking with contains, so consider there are 2 branches `feature/f1` and `feature/f10`.

abhvsn (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: I am using `.contains` on list of strings and it looks for exact match.

abhvsn: `.equals` looks for the exact match

AnaghHegde: `list.contains()` looks for the exact match. `String.contains()` looks for partial match.

abhvsn (approved):

PR #10249, opened 2022-01-07
chore: Add env variable for git filesystem in CI

abhvsn:
## Description

> Set a file path for the git to store files during the CI runs. As this was not set earlier TCs were failing with the error `Creating a directory failed`   

## Type of change


mohanarpit (approved):

sharat87:
@abhvsn if the folder doesn't exist, does the server create it and it's parents on startup automatically?

abhvsn:
> @abhvsn if the folder doesn't exist, does the server create it and it's parents on startup automatically?

@sharat87 Yes in local setup when we run the `java -jar` command and if env is present it creates the directory if not present alredy

PR #10257, opened 2022-01-10
feat: Mongo plugin: add `limit` field to `aggregate` command to allow users to define `batchSize` 

### LINKED ISSUE #2290: [Enhancement] Mongo Atlas is returning limited records
## Description

The user reported that their query is returning only 101 records when there are multiple records in the DB

sumitsum:
## Description
- This PR adds a `limit` field which sets the `batchSize` for `aggregate` command. This way the user can set the number of documents they want in their Mongo query result. 
- `limit` is set to `10` initially. However, for pre-existing Mongo actions, migration has been added to set this field to `101` - which is the default value that has been used by Mongodb in the absence of any user set value.  Ref: https://docs.mongodb.com/manual/tutorial/iterate-a-cursor/

Fixes #2290

## Type of change

## How Has This Been Tested?
- JUnit TC
- Manual

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
@@ -148,6 +148,8 @@ import static org.springframework.data.mongodb.core.query.Update.update;
 public class DatabaseChangelog {
 
     public static ObjectMapper objectMapper = new ObjectMapper();
+    private static final String AGGREGATE_LIMIT = "aggregate.limit";
+    private static final Object DEFAULT_BATCH_SIZE = "101";
     public static final String FIRESTORE_PLUGIN_NAME = "firestore-plugin";
     public static final String CONDITION_KEY = "condition";
     public static final String CHILDREN_KEY = "children";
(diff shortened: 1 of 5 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
```diff
@@ -4767,4 +4769,108 @@ public void createSystemThemes(MongockTemplate mongockTemplate) throws IOExcepti
...
+     * It iterates over each action id one by one to avoid out of memory error.
+     * @param mongoActions
+     * @param mongockTemplate
+     */
+    private void updateLimitFieldForEachAction(List<NewAction> mongoActions, MongockTemplate mongockTemplate) {
+        mongoActions.stream()
+                .map(NewAction::getId) // iterate over one action id at a time
+                .map(actionId -> fetchActionUsingId(actionId, mongockTemplate)) // fetch action using id
+                .filter(this::hasUnpublishedActionConfiguration)
+                .forEachOrdered(mongoAction -> {
```
nidhi-nair, line 4784: Do you think we could also filter by action configurations that do not use the `aggregate` command?

sumitsum: @nidhi-nair this thought had crossed my mind, but I thought we should do this for all the actions because this scenario might occur - say a user has a `find` query that he then changes to a `aggregate` query - in this case the client would expect some data for `limit` field from the database since it has an initial value set in `aggregate.json` file - so any persisted mongo actionConfiguration is expected to have `aggregate.limit` field set even if the actual cmd is something else - so basically I am not sure how the client would handle a missing initial value in case we choose to skip it for non-aggregate cmds and then user switches to aggregate cmd later.

nidhi-nair (approved):

PR #10282, opened 2022-01-10
feat: Git import api

### LINKED ISSUE #5969: Import app flow into the org via GIT
- Import link at organisation context menu at the home page
- Import modal at the home page. Steps:
    - Generate a standalone ssh key 
    - Connect step: Import api that supports local git config, repo URL or flag to use local / global
    - Message: datasources won’t contain secrets
    - On completion should navigate to the newly created app with the default branch checked out

Edge cases to consider:
- On quitting the flow after generating the key should a new app be created or only on completion?
- Datasources should need to to be updated only if the app is being imported into a new org which doesn't contain the datasources

### Tasks
- [x] #9755
  - [x] Make the SSH generation API independent of application id
  - [x] API for connect to git should work independent of application id  
- [x] #9754
- [ ] #9753
- [x] #10700
- [x] #10882
- [x] #10700 
- [x] #10884 
- [ ] #10950

AnaghHegde:
## Description

Allow users to import application from git repositories. 

Fixes #5969 

## Type of change

## How Has This Been Tested?

- Unit Tests
- Locally 

## Checklist:


AnaghHegde:
@abhvsn ready for review :)

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCE.java
```diff
@@ -57,7 +57,7 @@
 
     Mono<String> createConflictedBranch(String defaultApplicationId, String branchName);
 
-    Mono<Application> importApplicationFromGit();
+    Mono<Application> importApplicationFromGit(String organisationId, GitConnectDTO gitConnectDTO);
```
abhvsn, line 60: Can we change this to `organizationId` instead of `organisationId`

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1958,8 +1959,166 @@ private String getReason() {
...
+                                            }
+                                            return Mono.just(gitAuth).zipWith(applicationMono);
+                                        });
+                            });
+                })
+                .flatMap(tuple -> {
+                    Application application = tuple.getT2();
+                    GitAuth gitAuth = tuple.getT1();
+
+                    String repoName = GitUtils.getRepoName(gitConnectDTO.getRemoteUrl());
```
abhvsn, line 2019: As this is used in a couple of places can we create a method level variable for repoName

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1958,8 +1959,166 @@ private String getReason() {
...
+                        if (error instanceof InvalidRemoteException) {
+                            return addAnalyticsForGitOperation(
+                                    AnalyticsEvents.GIT_IMPORT.getEventName(),
+                                    application.getOrganizationId(),
+                                    application.getId(),
+                                    application.getId(),
+                                    error.getClass().getName(),
+                                    error.getMessage(),
+                                    false
+                            ).flatMap(user -> deleteApplicationMono
```
abhvsn, line 2052: As the `addAnalyticsForGitOperation` is same in both iff statements can we please club these.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1958,8 +1959,166 @@ private String getReason() {
...
+                                    );
+                                } catch (IOException e) {
+                                    gitApplicationMetadata.setIsRepoPrivate(true);
+                                    log.debug("Error while checking if the repo is private: ", e);
+                                }
+
+                                Mono<ApplicationJson> applicationJsonMono;
+                                try {
+                                    applicationJsonMono = fileUtils.reconstructApplicationFromGitRepo(organizationId, application.getId(), repoName, defaultBranch);
+                                } catch (GitAPIException | IOException e) {
```
abhvsn, line 2088: If I'm not wrong `catch` will never get called. We will have to use onError** methods to catch the errors

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: We are using this in two places but yeah good to be used as method level argument.

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/GitImportDTO.java
```diff
@@ -0,0 +1,16 @@
...
+import lombok.Setter;
+
+import java.util.List;
+
+@Getter
+@Setter
+public class GitImportDTO {
+    Application application;
+
+    List<Datasource> emptyDatasourceList;
```
abhvsn, line 15: Can we have a more relevant fieldName here something like `unconfiguredDatasources`

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -2084,6 +2090,28 @@ private String getReason() {
                 .thenReturn(gitAuth);
     }
 
+    private Mono<List<Datasource>> findNonConfiguredDatasourceByApplicationId(String applicationId,
```
abhvsn, line 2098: Can we move this method to `datasourceService`

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -2084,6 +2090,28 @@ private String getReason() {
...
+                                                                             Mono<List<Datasource>> dataListMono) {
```
abhvsn: Please rename this to `datasourceListMono`? Also can't we have a flux here if possible

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -2084,6 +2090,28 @@ private String getReason() {
...
+        Mono<List<NewAction>> newActionList = newActionService.findAllByApplicationIdAndViewMode(applicationId, false, AclPermission.READ_ACTIONS, null).collectList();
+
+        return Mono.zip(dataListMono, newActionList)
+                .flatMap(objects -> {
+                    List<Datasource> datasourceList = objects.getT1();
+                    List<NewAction> actionList = objects.getT2();
+                    List<String> usedDatasource = actionList.stream()
```
abhvsn, line 2103: Does it make sense to have a `Set<String>` instead of `List<String>` to make the filter more performant?

AnaghHegde:
@abhvsn We will be removing the logic to the un-configured datasource list from the import flow. Per latest design we need to provide an option to the user to configure it post import as well. So i will be adding this logic to the existing get Datasource for org API itself.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/GitServiceTest.java
```diff
@@ -175,6 +176,11 @@ public void setup() throws IOException {
         }
         Organization testOrg = organizationRepository.findByName("Another Test Organization", AclPermission.READ_ORGANIZATIONS).block();
         orgId = testOrg.getId();
+        MockedStatic<GitUtils> gitUtilsMockedStatic = Mockito.mockStatic(GitUtils.class);
+        gitUtilsMockedStatic.when(() -> GitUtils.isRepoPrivate(Mockito.anyString()))
+                .thenReturn(Boolean.FALSE);
+        gitUtilsMockedStatic.when(() -> GitUtils.convertSshUrlToBrowserSupportedUrl(Mockito.anyString()))
+                .thenReturn("https://test.com");
```
abhvsn: I don't think we need to mock GitUtils at all . Can we revert back this change to the earlier implementation

app/server/appsmith-server/src/test/java/com/appsmith/server/services/GitServiceTest.java
```diff
@@ -286,7 +292,7 @@ public void connectApplicationToGit_EmptyRemoteUrl_ThrowInvalidParameterExceptio
     @WithUserDetails(value = "api_user")
     public void connectApplicationToGit_EmptyOriginHeader_ThrowInvalidParameterException() {
 
-        GitConnectDTO gitConnectDTO = getConnectRequest("git@github.com:test/testRepo.git", testUserProfile);
+        GitConnectDTO gitConnectDTO = getConnectRequest("sshUrl", testUserProfile);
```
abhvsn: Why are we touching connect TC?

app/server/appsmith-server/src/test/java/com/appsmith/server/services/GitServiceTest.java
```diff
@@ -306,7 +312,7 @@ public void connectApplicationToGit_InvalidGitApplicationMetadata_ThrowInvalidGi
         testApplication.setOrganizationId(orgId);
         Application application1 = applicationPageService.createApplication(testApplication).block();
 
-        GitConnectDTO gitConnectDTO = getConnectRequest("git@github.com:test/testRepo.git", testUserProfile);
+        GitConnectDTO gitConnectDTO = getConnectRequest("test.url.git", testUserProfile);
```
abhvsn: Same here

AnaghHegde: Yes i agree. But the values used were not per standard and found it while checking for something else. Hence i just replaced the values. Its a string value and it does not affect the test cases.

abhvsn: Sorry I didn't get what are we expecting here. For git URLs following are the standard format which seems to be correct with earlier URLs:

Github :
* git@github.com:user/repo.git

Gitlab:
* git@gitlab.com:user/repo.git

Bitbucket:
* git@bitbucket.org:user/repo.git

AnaghHegde: This is outdated, my bad. I was referring to the same, if you look at the changes from latest commits, you can see that old values were random strings and i replaced those with standard ssh urls.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/Datasource.java
```diff
@@ -59,6 +59,14 @@
     @JsonIgnore
     DatasourceStructure structure;
 
+
+    /*
+     * This field is introduced as part of git sync feature, for the git import we will need to identify the datasource's
+     * which are not configured. This way user can configure those datasource, which may have been introduced as part of git import.
+     */
+    @Transient
+    Boolean isEmpty;
```
abhvsn, line 68: Can we rename this to `isConfigured` as `isEmpty` can convey different meaning

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/GitImportDTO.java
abhvsn: Why are we calling this as `emptyDatasource` if only the config is missing?

abhvsn (approved):
LGTM

PR #10290, opened 2022-01-10
fix: Added null check on eval output and started using standard call for pluginId

### LINKED ISSUE #10286: [Bug]: The infinite loading state is displayed in Query Param

### Current Behavior

The infinite Loading state is displayed in the query parm
image

### Steps To Reproduce

This issue i am facing mainly in this app
https://release.app.appsmith.com/applications/616d26cd8cc2642df8e19f51/pages/616d26cd8cc2642df8e19f59/edit/queries/616d26cd8cc2642df8e19f5d

### Environment

Release

### Version

Cloud

ayushpahwa:
## Description

We were using 2 different paths for fetching pluginID (`action.pluginId` and `action.datasource.pluginId`) for fetching the query forms. While the first one is a standard call, we used the second one in some places causing issues. While most of the plugins had both of the values set, in some cases the second one was not set. This was causing the evaluations to not run and hence leading to an infinite loading state since the form was waiting on evaluation output. This PR fixes the same. Also, added a null check on the output of the evaluation output to prevent setting the whole state as null.

Fixes #10286 

## Type of change

## How Has This Been Tested?

- Test A
- Test B

## Checklist:

### DIFF
--- a/app/client/src/sagas/QueryPaneSagas.ts
+++ b/app/client/src/sagas/QueryPaneSagas.ts
@@ -66,14 +66,10 @@ function* changeQuerySaga(actionPayload: ReduxAction<{ id: string }>) {
     return;
   }
 
-  const currentEditorConfig: any[] = yield select(
-    getEditorConfig,
-    action.datasource.pluginId,
-  );
-  const currentSettingConfig: any[] = yield select(
-    getSettingConfig,
-    action.datasource.pluginId,
-  );
+  // fetching pluginId and the consequent configs from the action
+  const pluginId = action.pluginId;
+  const currentEditorConfig: any[] = yield select(getEditorConfig, pluginId);
+  const currentSettingConfig: any[] = yield select(getSettingConfig, pluginId);
 
   // Update the evaluations when the queryID is changed by changing the
   // URL or selecting new query from the query pane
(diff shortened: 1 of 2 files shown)


mohanarpit:
By any chance, do we know why the value is not being set in certain places or for certain actions?

rashmi-rai (approved):

ayushpahwa:
> By any chance, do we know why the value is not being set in certain places or for certain actions?

@mohanarpit , the way new actions are created is done from different places in the frontend code. So while most of the places are setting the pluginId, my assumption is that some places are missing the same. Since the `action.datasource` object isn't been protected by TS, I guess the null value is slipping out. I talked with Trisha and she mentioned that `action.pluginId` is the correct way to get this data and the FE code also has defined type for this. Thus concluded with this solution.

mohanarpit:
@ayushpahwa While we can merge this in as an emergency fix, we should fix the root cause of the issue. This is a tech debt problem where the actions are getting created via 2 flows. We should ideally have a single flow to perform a single action.

ayushpahwa:
@mohanarpit while there is a single function that creates the action, the issue is the missing TS definitions. I think fixing that is going to help us the most. Will create an issue for the same.

PR #10307, opened 2022-01-11
fix: Change SSH key generation protocol for git 

### LINKED ISSUE #10302: [Bug]: SSH keys generation issue

### Current Behavior

We use Jsch library to generate SSH keys which used by git sync feature to connect to a git client. The Jsch library generates the RSA SHA-1 types keys. But the github made some changes to improve protocol security when you push or pull Git data. This change includes removing support for RSA SHA-1 type signature keys. 
More can be found here - https://github.blog/2021-09-01-improving-git-protocol-security-github/

### Steps To Reproduce

All the git related operations will fail because of the security change done by github as the support for RSA SHa-1 type keys is stopped from today. 

### Environment

Production

### Version

Cloud

AnaghHegde:
## Description
We use Jsch library to generate SSH keys which used by git sync feature to connect to a git client. The Jsch library generates the RSA SHA-1 types keys. But the github made some changes to improve protocol security when you push or pull Git data. This change includes removing support for RSA SHA-1 type signature keys.
More can be found here - https://github.blog/2021-09-01-improving-git-protocol-security-github/. 
As a fix we are moving away from RSA-SHA1 to ECDSA-SHA2 protocol for key generation. 

Fixes #10302 

## Type of change

## How Has This Been Tested?

- Locally

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/GitDeployKeyGenerator.java
```diff
@@ -10,6 +10,8 @@
 import com.jcraft.jsch.JSchException;
 import com.jcraft.jsch.KeyPair;
 
+import java.security.KeyPairGenerator;
+import java.security.NoSuchAlgorithmException;
```
abhvsn, line 14: Can we remove these imports if not required?

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/GitDeployKeyGenerator.java
```diff
@@ -19,10 +21,11 @@
...
         JSch jsch = new JSch();
         KeyPair kpair;
+
         try {
-            kpair = KeyPair.genKeyPair(jsch, KeyPair.RSA, 2048);
+            kpair = KeyPair.genKeyPair(jsch, KeyPair.ECDSA, 256);
         } catch (JSchException e) {
-            log.error("failed to generate RSA key pair", e);
+            log.error("failed to generate ECDSA key pair", e);
             throw new AppsmithException(AppsmithError.GENERIC_BAD_REQUEST, "Failed to generate SSH Keypair");
```
abhvsn, line 29: Also can we add more relevant exception here so that we can log this in sentry as well

abhvsn (commented):
LGTM just a minor comment so as to log these errors in sentry

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/GitDeployKeyGenerator.java
AnaghHegde: The failure could be because of any reason. The error message contains the detailed message of that. I don't see any benefits in classifying the error as its library method and we don't have any control over this.

abhvsn: IIUC this error is related to key generation then why can't we add this in AppsmithError list and pass the exact error msg as a param

abhvsn (approved):

PR #10312, opened 2022-01-11
chore: Remove CORS configuration since we don't need it today

sharat87:


mohanarpit (approved):

sharat87:
/ok-to-test sha=
253520f

Issue #5890, opened 2021-07-15
[Bug]Migrations are unaccounted during import/export of applications


trishaanand:
When an application is exported from a very old version and imported in the newest code versions, more often than not, this breaks. This is because the code expects the data in a certain format which database migrations usually take care of but import/export flow does not account for the same.

PR #10339, opened 2022-01-12
feat: Apply JSON schema migration while importing the application using JSON file  

abhvsn:
## Description

> When we import the application using the JSON file it does not take DB migration into consideration. This PR provides the structure to write a JSON schema migration so that developers will have a way to write a migration for both DB and exported json files.

Fixes https://github.com/appsmithorg/appsmith/issues/5890

## Type of change

## How Has This Been Tested?

> JUnit TC
> Manual Test

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -390,6 +396,20 @@
...
+                    JsonObject json = gson.fromJson(data, JsonObject.class);
+                    JsonObject update = new JsonObject();
+                    update.addProperty("slug", "update_name");
+                    update.addProperty("name", "update name");
+
+                    ((JsonObject) json.get("exportedApplication")).add("name", update);
+                    json.get("random") == null => true
+
+                    ((JsonArray) json.get("pageList"))
+                    */
```
abhvsn, line 412: These are some of the examples I tried to work with Json string itself before doing any type of de-serialization and we can use this strategy when we completely remove the field from the `domain` instead of just deprecating the field. 
But if this clutters we can remove this

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/JsonSchemaMigration.java
```diff
@@ -0,0 +1,54 @@
+package com.appsmith.server.migrations;
+
+import com.appsmith.server.domains.ApplicationJson;
+import com.appsmith.server.exceptions.AppsmithError;
+import com.appsmith.server.exceptions.AppsmithException;
+
+public class JsonSchemaMigration {
+    private static Boolean checkCompatibility(ApplicationJson applicationJson) {
```
sharat87, line 8: Prefer primitive `boolean` here since we never intend to return `null` here.

```suggestion
    private static boolean checkCompatibility(ApplicationJson applicationJson) {
```

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/JsonSchemaMigration.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public static ApplicationJson migrateApplicationToLatestSchema(ApplicationJson applicationJson) {
+        // Check if the schema versions are available and set to initial version if not present
+        Integer serverSchemaVersion = applicationJson.getServerSchemaVersion() == null ? 0 : applicationJson.getServerSchemaVersion();
+        Integer clientSchemaVersion = applicationJson.getClientSchemaVersion() == null ? 0 : applicationJson.getClientSchemaVersion();
+
+        applicationJson.setClientSchemaVersion(clientSchemaVersion);
+        applicationJson.setServerSchemaVersion(serverSchemaVersion);
+        if (Boolean.FALSE.equals(checkCompatibility(applicationJson))) {
```
sharat87, line 20: If the return type is changed to `boolean`, this can just be:

```suggestion
        if (!checkCompatibility(applicationJson)) {
```

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/JsonSchemaVersions.java
```diff
@@ -0,0 +1,17 @@
...
+ * migration which is going to affect the underlying DB structure for different collection used in the JSON file, we
+ * have to increment the corresponding version here by 1 unit. We have 2 different fields for client and server to
+ * address the issue of linear incremental migration on both server and client side as both run their own migrations.
+ * Having said that during import server will have to check the compatibility for both client
+ * (widget dsl) and server-side resources, so that imported application will be in sane state after import is successful
+ */
+@Getter
+public class JsonSchemaVersions {
+    public static Integer serverVersion = 1;
+    public static Integer clientVersion = 1;
```
sharat87, line 16: Can we make them constants? Do we expect them to change without a server restart? If not, something like `public static final Integer SERVER_SCHEMA_VERSION = 1;`?

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/ApplicationJson.java
```diff
@@ -18,8 +19,19 @@
 @Setter
 public class ApplicationJson {
 
+    // This field will be useful when we migrate fields within JSON files (currently this will be useful for Git feature)
     @JsonIgnore
-    Integer version = 1;
+    Integer fileSystemVersion = 1;
```
sharat87, line 24: Can we keep this as `version`, or perhaps `fileFormatVersion`? The word [_file system_](https://en.wikipedia.org/wiki/File_system) gives a very different meaning, which can easily trip people up. :)

abhvsn: `fileFormatVersion` looks more relevant I'll update this

sharat87 (approved):

PR #10383, opened 2022-01-13
fix: table column name migration update

### LINKED ISSUE #6671: [Bug]The user cant bind data in the input widget  if the column name are numeric
## Description

[What happened]The user cant bind data in the input widget  if the column name are numeric 

### Steps to reproduce the behavior:

1.Generate a page using google sheet
2.Select a sheet that has a numeric value as its column name 
3.now very the display of the autogenerated template
image
If the column names are numeric then that value is not displayed in the autosuggest
image

### Important Details

- Version: [Cloud ]
- OS: [MacOS]
- Browser [chrome]
- Environment [ deploy preview->6301]

techbhavin:
## Description

> remove label from table migration 

Fixes #6671

## Type of change

## Checklist:

    // Code coverage diff between base branch:release and head branch: fix/numeric-column-name-migration 
 :green_circle: | total | 55.14 **(0.05)** | 36.93 **(0.05)** | 35.05 **(0.04)** | 55.62 **(0.05)**
 :green_circle: | app/client/src/utils/DSLMigrations.ts | 73.3 **(0.12)** | 69.56 **(-0.06)** | 71.13 **(0)** | 73.1 **(0.11)**
 :green_circle: | app/client/src/utils/migrations/TableWidget.ts | 85.66 **(9.43)** | 75.22 **(6.09)** | 91.18 **(11.77)** | 87.23 **(10.63)**
 :green_circle: | app/client/src/widgets/TableWidget/component/TableHelpers.ts | 100 **(3.12)** | 78.57 **(14.28)** | 100 **(16.67)** | 100 **(3.85)**</details>


app/client/yarn.lock
```diff
@@ -6248,6 +6248,11 @@ cypress-real-events@^1.5.1:
   version "1.5.1"
   resolved "https://registry.yarnpkg.com/cypress-real-events/-/cypress-real-events-1.5.1.tgz#5eeb86d2a7aad9aa6d5271e288a23e46373915cd"
 
+cypress-wait-until@^1.7.2:
```
marks0351, line 6251: are these yarn changes relevant?

marks0351 (changes requested):

marks0351:
@techbhavin do we have test cases written for this migration?

techbhavin:
@marks0351  no we don’t have a test cases for this migration

marks0351:
> @marks0351 no we don’t have a test cases for this migration

can u pls add test cases.

app/client/src/utils/migrations/TableWidget.ts
```diff
@@ -509,7 +509,6 @@ export const migrateTableWidgetNumericColumnName = (currentDSL: DSLWidget) => {
           child.primaryColumns[columnId] = {
             ...column,
             id: columnId,
```
sbalaji1192, line 511: Don't we need to change the label as well for the columns other than custom columns?

app/client/src/utils/DSLMigrations.ts
```diff
@@ -1001,6 +1002,11 @@ export const transformDSL = (
 
   if (currentDSL.version === 48) {
     currentDSL = migrateRecaptchaType(currentDSL);
+    currentDSL.version = 49;
```
sbalaji1192, line 1005: We need to add a comment on DSL migration version 48, that the skipped migration is added as version 50. for future reference.

sbalaji1192 (changes requested):

app/client/src/utils/migrations/TableWidget.ts
techbhavin: here we remove label properties  from migration, so will not change for any type of columns 

@sbalaji1192

marks0351 (commented):
@techbhavin what will happen to these column bindings that might be used in other widgets.
Table1.selectedRow.1 will it be migrated to Table1.selectedRow._1 as well?
cc: @sbalaji1192

marks0351:
@techbhavin also another case that i observed is, when data has a number column, then it creates a _ prefixed column name, but it is not restricted i can still change it to number. is this expected?

techbhavin:
> @techbhavin what will happen to these column bindings that might be used in other widgets. Table1.selectedRow.1 will it be migrated to Table1.selectedRow._1 as well? cc: @sbalaji1192

Before fixing this issue  `Table1.selectedRow.1` format was not accessible due to bad syntax. if still there is a way to bind this syntax. we have to handle it. 

cc: @marks0351

techbhavin:
> 



> @techbhavin also another case that I observed is when data have a number column, then it creates a _ prefixed column name, but it is not restricted I can still change it to number. is this expected?

@marks0351  here we are adding _prefixed in a column id property while generating a table column and the user doesn’t have control to modify it.  but users can change their label Property so I think we can’t restrict users to change column label

marks0351 (approved):
I am approving the changes but still looks a bit inconsistent experience coz we allow users to set a numeric label but the user has to use an _ if he wants to bind it in another widget.
cc: @techbhavin @sbalaji1192 @somangshu

YogeshJayaseelan:
@techbhavin Tested this PR and please find the status below,

Auto suggestion and the numeric bindings are working fine when the table data is binded with the google sheet api with numeric columns.

Issue 1 - Channing the column name manually to numeric and on binding with inout widget auto suggestion are not displayed for numeric columns

https://loom.com/share/5fa28ec46e6e445883a88a2fa4fd9615

sbalaji1192 (approved):

YogeshJayaseelan:
@techbhavin Tested this PR and working as expected

app/client/src/utils/migrations/TableWidget.test.ts
```diff
@@ -2010,3 +2012,487 @@ describe("Table Widget selectedRow bindings update", () => {
...
+          widgetName: "Table1",
+          parentColumnSpace: 74,
+          defaultPageSize: 0,
+          columnOrder: [
+            "_1",
+            "_2",
+            "_user_",
+            "_client_",
+            "rowIndex",
+            "customColumn1",
```
sbalaji1192, line 2291: Could we add one more custom column with a custom label and check the label is not lost during migration?

sbalaji1192 (approved):

PR #10430, opened 2022-01-17
feat: Remove action and actionCollection ids during export application

### LINKED ISSUE #10429: [Feature]: Remove action and actionCollection ids from the exported JSON file

### Summary

When we export the application using the JSON file, actionId and actionCollectionIds are exported along with this JSON. For security reasons we are converting these Ids to a human-readable identifiers.

This will help in reducing id leaks when we export the application, also this will be helpful to avoid un-necessary conflicts as we are using the same mechanism for git-synk.

abhvsn:
## Description

> Today when we export the application action ids and actionCollection ids are exported along with the JSON file. This may result in sharing the DB ids which can be a security concern. This PR converts the these ids to human readable identifiers which prevents sharing the actual DB Ids.

Fixes #10429 

## Type of change

## How Has This Been Tested?

> JUnit TC
> Manual test

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/GitFileUtils.java
```diff
@@ -268,6 +268,7 @@ private void removeUnwantedFieldFromActionCollection(ActionCollection actionColl
         }
         if (publishedCollection != null) {
             publishedCollection.setDefaultResources(null);
+            publishedCollection.setDefaultToBranchedActionIdsMap(null);
```
abhvsn, line 271: This change is not specific to the description but added this because it was a small and earlier we missed this to add when we implemented git

AnaghHegde (approved):

PR #10467, opened 2022-01-18
Feature: Refactor k8s installation to use fat container

geekup-legodevops:
## Description
- Remove unnecessary k8s templates (templates being used for multiple containers)
- Create new deployment template that use the fat container
- Refactor the k8s installation script to fetch the new template for fat container deployment on k8s
## Type of change
## Checklist:


deploy/k8s/scripts/appsmith-configmap.yaml.sh
```diff
@@ -1,38 +1,43 @@
...
-  # APPSMITH_MAIL_SMTP_TLS_ENABLED: ""
-  # APPSMITH_MAIL_USERNAME: ""
-  # APPSMITH_MAIL_PASSWORD: ""
-  # APPSMITH_MAIL_SMTP_AUTH: ""
-  # APPSMITH_OAUTH2_GOOGLE_CLIENT_ID:  ""
-  # APPSMITH_OAUTH2_GOOGLE_CLIENT_SECRET: ""
-  # APPSMITH_OAUTH2_GITHUB_CLIENT_ID: ""
-  # APPSMITH_OAUTH2_GITHUB_CLIENT_SECRET:  ""
+  APPSMITH_SENTRY_DSN: ""
+  APPSMITH_SMART_LOOK_ID: ""
```
sharat87, line 13: These are internal variables, not be configurable by the user.

```suggestion
```

deploy/k8s/scripts/appsmith-configmap.yaml.sh
```diff
@@ -1,38 +1,43 @@
...
+  APPSMITH_OAUTH2_GOOGLE_CLIENT_SECRET: ""
+  APPSMITH_OAUTH2_GITHUB_CLIENT_ID: ""
+  APPSMITH_OAUTH2_GITHUB_CLIENT_SECRET: ""
+  APPSMITH_SEGMENT_KEY: ""
+  APPSMITH_RAPID_API_KEY_VALUE: ""
+  APPSMITH_MARKETPLACE_ENABLED: ""
+  APPSMITH_OPTIMIZELY_KEY: ""
+  APPSMITH_ALGOLIA_API_ID: ""
+  APPSMITH_ALGOLIA_API_KEY: ""
+  APPSMITH_ALGOLIA_SEARCH_INDEX_NAME: ""
```
sharat87, line 24: These are internal variables, not be configurable by the user.

```suggestion
```

deploy/k8s/scripts/appsmith-configmap.yaml.sh
```diff
@@ -1,38 +1,43 @@
...
+  APPSMITH_MAIL_SMTP_AUTH: ""
+  APPSMITH_MAIL_SMTP_TLS_ENABLED: ""
   APPSMITH_DISABLE_TELEMETRY: "$disable_telemetry"
   APPSMITH_RECAPTCHA_SITE_KEY: ""
   APPSMITH_RECAPTCHA_SECRET_KEY: ""
   APPSMITH_RECAPTCHA_ENABLED: "false"
   APPSMITH_DISABLE_INTERCOM: "false"
-  # APPSMITH_PLUGIN_MAX_RESPONSE_SIZE_MB=5
+  APPSMITH_MONGODB_URI: $mongo_uri
+  APPSMITH_REDIS_URL: redis://127.0.0.1:6379
```
sharat87, line 42: Let's wrap these in double-quotes as well.

```suggestion
  APPSMITH_MONGODB_URI: "$mongo_uri"
  APPSMITH_REDIS_URL: "redis://127.0.0.1:6379"
```

deploy/k8s/scripts/appsmith-ingress.yaml.sh
```diff
@@ -19,80 +19,37 @@ if [[ "$ssl_enable" == "true" ]]; then
...
-              servicePort: 80
           - path: /
             pathType: Prefix
             backend:
-              serviceName: appsmith-editor
-              servicePort: 80
+              service:
+                name: appsmith-service
+                port:
+                  number: 80
```
sharat87, line 32: If a custom domain is configured, then Appsmith would be listening on 443 as well right? Shouldn't we route to that as well?

deploy/k8s/templates/appsmith-template.yaml
```diff
@@ -0,0 +1,50 @@
...
+  serviceName: "appsmith-service"
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: appsmith
+    spec:
+      containers:
+      - name: appsmith
+        image: legodevops/appsmith:helm-20220105
```
sharat87, line 32: Looks like a dev image? Can you check once?

sharat87 (changes requested):

deploy/k8s/scripts/appsmith-ingress.yaml.sh
geekup-legodevops: Hi @sharat87 , when user access to Appsmith deployed on the K8S, all requests will be routed to Ingress on HTTP or HTTPS connection. Then, these requests will be routed to internal services by the Ingress which not require a secure protocol like HTTPS (internal communication in K8S cluster). Therefore, we will only configure the ingress to route request to port 80 which is the default port of service (container)

sharat87: So in that case if we configure the custom domain in `APPSMITH_CUSTOM_DOMAIN`, will Appsmith not attempt to provision its own certificate and start listening on port 443 if it successfully got a certificate? In that case port 80 will just be redirecting to the https server on port 80 no?

geekup-legodevops: Yes @sharat87 The context for the K8S is different from the Docker container. While user connects directly to the port 80 or 443 of the container when we start it with Docker, K8S exists with many other components and more like a whole system rather than a single running container. Therefore, in K8S, a request will need to go through another layer (ingress) instead of reaching the container port itself

sharat87: I think we are talking about different things. Let me rephrase my question. After deploying Appsmith using this install script, if I go set the `APPSMITH_CUSTOM_DOMAIN` variable to `something.appsmith.com`, what will happen? Won't the entrypoint script attempt to provision an SSL certificate because this env variable is set?

PR #10478, opened 2022-01-19
feat: Add logic to sync the repo if the path is not updated in env file

### LINKED ISSUE #10469: [Task] Sync the repo with remote when the repo is deleted from the filesystem
Docker volumes does not get picked up until user’s don’t have env variable `APPSMITH_GIT_ROOT` in their `docker.env`
This issue is faced when the user restarts their container. 
So to keep local git-repo and DB in sync when the local repo is deleted, we should do the below set of actions from BE. 
- Clone remote repo
- Check apps in DB which are linked to the repo and pull these branches in local

AnaghHegde:
## Description

> 
1. At present, we are cloning the git repo the local file system and if this repo is deleted by the user(mostly on self-hosted) accidentally we don't have any mechanism to sync the repo again.  
2. Another issue is where if the user has not set the env variable `APPSMITH_GIT_ROOT` while configuring - if the container is restarted all the git related files are lost. 

This change set will address this issue by cloning the repo when it is deleted from the local file system. This check is dynamic and user does not notice any difference from the UX point of view. 

Fixes #10469 

## Type of change

## How Has This Been Tested?

- Locally

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1932,6 +1940,37 @@ private boolean isInvalidDefaultApplicationGitMetadata(GitApplicationMetadata gi
...
+                    GitAuth gitAuth = gitApplicationMetadata.getGitAuth();
+                    return gitExecutor.cloneApplication(repoPath, gitApplicationMetadata.getRemoteUrl(), gitAuth.getPrivateKey(), gitAuth.getPublicKey())
+                            .flatMap(defaultBranch -> gitExecutor.listBranches(
+                                    repoPath,
+                                    gitApplicationMetadata.getRemoteUrl(),
+                                    gitAuth.getPrivateKey(),
+                                    gitAuth.getPublicKey(),
+                                    false))
+                            .flatMap(gitBranchDTOList -> {
+                                List<String> branchList = gitBranchDTOList.stream().map(GitBranchDTO::getBranchName).collect(Collectors.toList());
```
abhvsn, line 1961: Should we filter local branch, at this point I suppose only `default branch` will be present in local after the clone is successful. This way we can prevent 1 DB call 🙂

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1932,6 +1940,37 @@ private boolean isInvalidDefaultApplicationGitMetadata(GitApplicationMetadata gi
...
+                                return Flux.fromIterable(branchList)
+                                        .flatMap(branchName -> applicationService.findByBranchNameAndDefaultApplicationId(branchName, application.getId(), READ_APPLICATIONS)
+                                                .onErrorResume(throwable -> {
+                                                    GitBranchDTO gitBranchDTO = new GitBranchDTO();
+                                                    gitBranchDTO.setBranchName(branchName);
+                                                    return createBranch(application.getId(), gitBranchDTO, gitApplicationMetadata.getDefaultBranchName());
+                                                }))
+                                        .then(Mono.just(application));
```
abhvsn, line 1969: Instead of sending a application can we send the `updatedBranchList`, this way we don't have to query again for the `branchList` in the upstream method.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1436,34 +1436,42 @@ private String getReason() {
...
-                                false)
-                                .onErrorResume(error -> Mono.error(new AppsmithException(
+                                false);
+                    }
+                    return Mono.zip(gitBranchDTOMono, Mono.just(application), Mono.just(repoPath))
+                            .onErrorResume(error -> {
+                                if (error instanceof RepositoryNotFoundException) {
+                                    return handleRepoNotFoundException(defaultApplicationId)
+                                            .flatMap(s -> {
+                                                Mono<List<GitBranchDTO>> branchListMono = gitExecutor.listBranches(
```
abhvsn, line 1460: If we get the updated branchList from `handleRepoNotFoundException` we can altogether avoid this call.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1932,6 +1940,37 @@ private boolean isInvalidDefaultApplicationGitMetadata(GitApplicationMetadata gi
...
+                                    gitAuth.getPublicKey(),
+                                    false))
+                            .flatMap(gitBranchDTOList -> {
+                                List<String> branchList = gitBranchDTOList.stream().map(GitBranchDTO::getBranchName).collect(Collectors.toList());
+                                return Flux.fromIterable(branchList)
+                                        .flatMap(branchName -> applicationService.findByBranchNameAndDefaultApplicationId(branchName, application.getId(), READ_APPLICATIONS)
+                                                .onErrorResume(throwable -> {
+                                                    GitBranchDTO gitBranchDTO = new GitBranchDTO();
+                                                    gitBranchDTO.setBranchName(branchName);
+                                                    return createBranch(application.getId(), gitBranchDTO, gitApplicationMetadata.getDefaultBranchName());
```
abhvsn, line 1967: Correct me if I'm wrong here but the branchName will have the prefix as `origin/` which is not allowed in current implementation of `createBranch` method. 

Also `createBranch` creates a branch from some `sourceBranch` but in this case, we need to create a branch from the local file system

abhvsn (changes requested):
Added some comments please take a look, we might need to tackle some cases here

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: Good point :)

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: I was thinking of removing the createBranch flow, because 
1. All the data in the db will be present. So it's just the git repo and anyway we are handling this case in `checkoutBranch` flow. 
2. And createBranch flow does not add any value here.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1932,6 +1932,49 @@ private boolean isInvalidDefaultApplicationGitMetadata(GitApplicationMetadata gi
...
+                            .flatMap(defaultBranch -> gitExecutor.listBranches(
+                                    repoPath,
+                                    gitApplicationMetadata.getRemoteUrl(),
+                                    gitAuth.getPrivateKey(),
+                                    gitAuth.getPublicKey(),
+                                    false))
+                            .flatMap(gitBranchDTOList -> {
+                                List<String> branchList = gitBranchDTOList.stream()
+                                        .filter(gitBranchDTO -> gitBranchDTO.getBranchName().contains("origin"))
+                                        .map(gitBranchDTO -> gitBranchDTO.getBranchName().replace("origin/", ""))
```
abhvsn, line 1955: With this approach we may end up in a situation where we have some branches in local but not present in DB. Consider a situation: 
1. We have `master, F1` present in local along with there remote counterparts (both the branches will be present in local repo and DB)
2. We create another branch in remote through other Appsmith instance `F2`
3. Now we try to synch our first instance 
4. We will end up in having `master, F1 and F2` in local repo but only `master and F1` will be present in DB

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1932,6 +1932,49 @@ private boolean isInvalidDefaultApplicationGitMetadata(GitApplicationMetadata gi
...
+                                        .collect(Collectors.toList());
+                                // Remove the default branch of Appsmith
+                                branchList.remove(gitApplicationMetadata.getBranchName());
```
abhvsn, line 1958: Any particular reason why we are removing the `defaultBranch`?

AnaghHegde: Since the defaultBranch is already checked out in local, we are removing this from the list

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: @abhvsn We will not be ending up in this situation because before checking out a local branch we are first checking in the DB(Line No 1960-69), so the branch will be checked out only if it exists in DB.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1932,6 +1932,49 @@ private boolean isInvalidDefaultApplicationGitMetadata(GitApplicationMetadata gi
...
+                                                .flatMap(application1 -> {
+                                                    // Add the locally checked out branch to the branchList
+                                                    GitBranchDTO gitBranchDTO = new GitBranchDTO();
+                                                    gitBranchDTO.setBranchName(branchName);
+                                                    gitBranchDTO.setDefault(false);
+                                                    gitBranchDTOList.add(gitBranchDTO);
+                                                    return gitExecutor.checkoutRemoteBranch(repoPath, branchName);
+                                                })
+                                                // Return empty mono when the branched application is not in db
+                                                .onErrorResume(throwable -> Mono.empty()))
```
abhvsn, line 1972: To avoid onErrorStates we can consider `findAllApplicationsByDefaultApplicationId` to find all the applications and then checking out in local if the branch is present in `branchList`

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
abhvsn: My bad you are correct!

abhvsn (approved):

PR #10491, opened 2022-01-19
fix: [Rest API Plugin] Only do generic mustache replacement if smart substitution is turned off

### LINKED ISSUE #10489: [Bug]: Literal mustache syntax getting truncated in API request

### Current Behavior

I've got a TextWidget with `text` property value as follows:
```
name: "Abcd"
age: 12
templates:
    - en: "Hello {{"{{ name }}"}}"
    - fr: "Bonjour {{"{{ name }}"}}"
```

Since there's a literal mustache syntax, I've disabled "Smart JSON substitution" as well.
I'm sending it to an API via POST method with the data as follows:
```
{
	content: {{input1.text}}
}
```

In the API end, what I receive is a truncated value:
```
name: "Abcd"
age: 12
templates:
    - en: "Hello {{name
```

### Steps To Reproduce

Do the steps shown in the above description

### Environment

Production

### Version

Cloud

trishaanand:
Fixes #10489 

Today, if smart substitution is turned on, mustache replacement happens twice. First, as per required for smart substitution, and then the generic mustache replacement which is meant to get hit when smart substitution is turned off. This leads to mustache substitution twice in case of smart substitution turned on. This wouldnt have been a problem because in usual circumstances because all the mustaches get replaced in the first round with nothing happening in the second round. In this particular instance, the evaluated value contains mustaches which when going through the erroneous second replacement leads to incorrect request body being sent out.

### DIFF
--- a/app/server/appsmith-plugins/restApiPlugin/src/main/java/com/external/plugins/RestApiPlugin.java
+++ b/app/server/appsmith-plugins/restApiPlugin/src/main/java/com/external/plugins/RestApiPlugin.java
@@ -174,9 +174,9 @@ public class RestApiPlugin extends BasePlugin {
 
                     actionConfiguration.setBody(updatedBody);
                 }
+            } else {
+                prepareConfigurationsForExecution(executeActionDTO, actionConfiguration, datasourceConfiguration);
             }
-
-            prepareConfigurationsForExecution(executeActionDTO, actionConfiguration, datasourceConfiguration);
 
             // If the action is paginated, update the configurations to update the correct URL.
             if (actionConfiguration.getPaginationType() != null &&


app/server/appsmith-plugins/restApiPlugin/src/main/java/com/external/plugins/RestApiPlugin.java
```diff
@@ -174,10 +174,10 @@ public RestApiPluginExecutor(SharedConfig sharedConfig) {
 
                     actionConfiguration.setBody(updatedBody);
                 }
+            } else {
+                prepareConfigurationsForExecution(executeActionDTO, actionConfiguration, datasourceConfiguration);
```
sumitsum, line 178: Not sure about this - just thinking out loud - if say we have a mustache binding in any other field except `actionConfiguration.body` and smart JSON substitution is turned `ON` - wouldn't that flow break because we wouldn't substitute the mustache binding with the values ?

app/server/appsmith-plugins/restApiPlugin/src/main/java/com/external/plugins/RestApiPlugin.java
```diff
@@ -174,10 +174,10 @@ public RestApiPluginExecutor(SharedConfig sharedConfig) {
...
             }
```
sumitsum, line 180: General: If possible, can you please add a JUnit TC for this ?

sumitsum (changes requested):

app/server/appsmith-plugins/restApiPlugin/src/main/java/com/external/plugins/RestApiPlugin.java
trishaanand: Agreed. Good catch.

Issue #10557, opened 2022-01-22
[Task]: `flattenObject` cannot handle circular/self-referential objects


IAmAnubhavSaini:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When an object is passed to `flattenObject` and that object has circular dependency, it fails to flatten the object and stack is overrun.

### Steps To Reproduce

It's about code quality. I am not sure if one can pass such an object from UI.

### Environment

Release

### Version

Self-Hosted

PR #10558, opened 2022-01-22
fix: flattenObject handles circular objects

IAmAnubhavSaini:
## Description

`app/client/src/utils/helpers.tsx` has a function `flattenObject` which is used to (well) flatten an object. This function however didn't handle any recursive/self-referential objects. This might crash the program.

Fixes # (issue)

#10557 10557

## Type of change

## How Has This Been Tested?

- Added a test in `app/client/src/utils/helpers.test.ts`

## Checklist:

- [ x ] My code follows the style guidelines of this project
- [ x ] I have performed a self-review of my own code
- [ x ] I have added tests that prove my fix is effective or that my feature works
- [ x ] New and existing unit tests pass locally with my changes

### DIFF
--- a/app/client/src/utils/helpers.test.ts
+++ b/app/client/src/utils/helpers.test.ts
@@ -6,6 +6,45 @@ import {
 } from "./helpers";
 
 describe("flattenObject test", () => {
+  it("Check if recursively referenced object is returend correctly", () => {
+    type InputType = {
+      s: string;
+      n: number;
+      b: boolean;
+      o1: { s: string };
+      o2: InputType;
+      a1: number[];
+      a2: InputType[];
+    };
+    const input: InputType = {
+      s: "string",
+      n: 1,
+      b: true,
+      o1: { s: "str" },
+      o2: {} as InputType,
+      a1: [1, 2, 3],
+      a2: [] as InputType[],
+    };
+
+    input.o2 = input;
+    input.a2[0] = input;
+    input.a2[1] = input;
+
+    const output = {
+      s: "string",
+      n: 1,
+      b: true,
+      "o1.s": "str",
+      o2: input,
+      "a1[0]": 1,
+      "a1[1]": 2,
+      "a1[2]": 3,
+      "a2[0]": input,
+      "a2[1]": input,
+    };
+
+    expect(flattenObject(input)).toStrictEqual(output);
+  });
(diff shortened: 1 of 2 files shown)


app/client/src/utils/helpers.test.ts
```diff
@@ -6,6 +6,45 @@ import {
 } from "./helpers";
 
 describe("flattenObject test", () => {
+  it("Check if recursively referenced object is returend correctly", () => {
```
mohanarpit, line 9: Possibly the smallest nit possible. :) 

```suggestion
  it("Check if recursively referenced object is returned correctly", () => {
```

app/client/src/utils/helpers.tsx
```diff
@@ -428,6 +428,7 @@ export const scrollbarWidth = () => {
 // To { isValid: false, settings.color: false}
 export const flattenObject = (data: Record<string, any>) => {
```
rahulramesha, line 429: Suggestion NIT: For Circular Objects, since the returned object is no longer flat, It might be worth accepting a second flag something like `shouldHandleCircularReferences` with default `true`. If `false` could just throw an error instead of a non-flat object. This might be good for use cases where you absolutely require flattenedObjects and are willing to catch an error if there are circular references.

IAmAnubhavSaini: This change specifically protects from a crash, which will happen if a circular objects ends up here. Earlier, calling code didn't care to know if the object was circular or interested in catching an error thrown by this function. 

If there is provision for catching such an error, then let me know error type, message and where it will be caught, I can surely add that.

rahulramesha: @IAmAnubhavSaini, Agreed. This will help in Handling circular objects for calling methods, It was a suggestion for future uses of this method. It is definitely not a blocker for these changes. I will leave the decision on implementing it upto you. : )

rahulramesha (approved):

IAmAnubhavSaini:
@mohanarpit Can we merge this PR? This has been waiting for long.

mohanarpit:
@IAmAnubhavSaini This is because you haven't run the `/ok-to-test sha=<last commit sha id>` command here. Hence, the tests haven't run. This is why the PR hasn't been merged.

PR #10572, opened 2022-01-24
feat: add JSON smart substitution feature to Firestore plugin actions

### LINKED ISSUE #7110: Add support for "smart substitution" for Firestore plugin

sumitsum:
## Description
- This PR adds the JSON smart substitution feature to `Firestore` plugin actions. 
  - This smart substitution will work exactly the same way as it works for `REST API` actions. 
  - Currently, smart substitution only gets applied to the `body` section of `Create`, `Set`, `Add` or `Update` commands as this is the only input field that expects a `JSON` input type.
  - It can be toggled `On/Off` via the `Settings` tab.
  - It is set to `On` by default. However, data migration has been added to set it to `Off` for all existing Firestore actions so that they are not impacted by this PR's changes.

Fixes #7110 

## Type of change

## How Has This Been Tested?
- JUnit TC
- Manual

## Checklist:


trishaanand (approved):

app/server/appsmith-plugins/firestorePlugin/src/main/java/com/external/plugins/FirestorePlugin.java
```diff
@@ -104,13 +110,63 @@ public FirestorePlugin(PluginWrapper wrapper) {
...
+
         @Override
         public Mono<ActionExecutionResult> executeParameterized(
                 Firestore connection,
                 ExecuteActionDTO executeActionDTO,
                 DatasourceConfiguration datasourceConfiguration,
                 ActionConfiguration actionConfiguration) {
 
+            Object smartSubstitutionObject = actionConfiguration.getFormData().getOrDefault(SMART_SUBSTITUTION, TRUE);
+            Boolean smartBsonSubstitution = TRUE;
```
trishaanand, line 133: Should we rename the variable to json instead of bson which is a mongo construct?

trishaanand (approved):

app/server/appsmith-plugins/firestorePlugin/src/main/java/com/external/plugins/FirestorePlugin.java
trishaanand: No problem. I also approved because in case it was urgent it could go in since this isn't an urgent change. Do fix this whenever you are free and taking a break :)

sumitsum: Thanks :) addressing it as part of this PR: https://github.com/appsmithorg/appsmith/pull/12204

PR #10600, opened 2022-01-24
feat: Controls for labels in widgets to align the widgets in forms and other places

### LINKED ISSUE #7064: [Enhancement] Controls for labels in widgets to align the widgets in forms and other places
## Description

[What happened]
When there are multiple input widgets with different label lengths then the input box looks miss aligned

1. Go to 'canvas
2. Drag and drop 2 to 3 input widget 
3. Now change the label name of all 3 input box
4. Give the first input box label as name and second as age and third as address 
5. Now the Label value area looks misaligned
image

### Important Details

- Version: [Cloud ]
- OS: [MacOS]
- Browser [chrome]
- Environment [release]

### LINKED ISSUE #7215: [Bug] Long label name on input widget does not adjust to resize, only input box does
## Description
When an input box with a long name is reduced in size, only the input box size reduces and any input is not visible whereas the widget label retains it's size.

[LOOM DEMO](https://www.loom.com/share/32513be9f0a94ab7ace6750d971d6d11)

### Important Details
- Version: Cloud
- OS: Win10
- Browser: Chrome
- Environment: Release

### LINKED ISSUE #9781: [Bug]: Input widget expansion cause UI discrepancy

### Current Behavior

Stretching Input widget to move label to the top causes widget to overlap with another widget if there's less space and highlighted boundary is not aligned to the actual boundary

### Steps To Reproduce

[LOOM DEMO](https://www.loom.com/share/69d295d5e80f48c6a3927bb82e0587eb)

### Environment

Production

### Version

Cloud

### LINKED ISSUE #9908: [Feature]: Add option for "Radio Group" widget to be presented in horizontal position

### Summary

Currently Radio Group widget is presented only in vertical position. This should works like for Switch Group widget, in responsive manner:
Zrzut ekranu 2021-12-21 o 10 35 01
Zrzut ekranu 2021-12-21 o 10 35 08

There are cases where from UX perspective it's required presentation.

### LINKED ISSUE #10119: [Bug]: The label text truncates on resizing the input widget

### Current Behavior

It is observed that when the user resizes the input widget the label text truncates when the text size is chosen as Heading 1

[LOOM DEMO](https://www.loom.com/share/a1a0ce96a6754d8b9f50b2db34b9cfc5)

### Steps To Reproduce

1.Launch appmsith app in release/Production environment 
2.Drag and drop Input widget
3.Add label text
4.Resize the widget position - Top to bottom 
5.Choose the text size as heading 1 and observe

### Environment

Production

### Version

Cloud

### LINKED ISSUE #11844: [Bug]: Text overlap from actual widget size when Label name is long text in Select family widgets

### Description

[LOOM DEMO](https://www.loom.com/share/7476a5f7784c4b409f26afda6d913156)

### Steps To Reproduce

1. Add any select widget to canvas
2. Add label name as long text
3. Observe the size of the widget and see the long text. It is overlapping the canvas too

### Public Sample App

_No response_

### Version

Cloud - Release

### LINKED ISSUE #12355: [Bug]: Label does not show on top of Password text inputs when resized

### Description

Normally When a Text input is resized the Label moves to its top. But for Password Text inputs this behaviour does not happen.

https://user-images.githubusercontent.com/37867493/160695912-bfcbd25b-258b-4f6b-9cdd-92cc3a17d612.mov

### Steps To Reproduce

1. Add an Input widget on canvas
2. Add a Label to it
3. Drag the input down and notice that the Label aligns to the top
4. Switch Data type to Password
5. Watch this alignment fail.

### Public Sample App

_No response_

### Version

1.6.16

wmdev0808:
## Description
- Today the users have to manually align labels for the input fields they are working on. With the default settings we provide it’s hard for a user to keep aligning them manually. We wish to improve this experience by providing the user will controls for editing the label’s position, alignment and width.
- Enhanced widgets: Input, Date picker, Select/Multi select, Tree select/Multi tree select, Rich text editor, Checkbox group, Switch group, and Radio group. 

> Side notes for QA:
- We will not have any scrollbar across all widgets in edit mode, which means we allowed contents to overflow beyond widgets bounds.
- Apart from testing core functionalities, you guys also have to test if old widgets including Input, select related widgets will not break its original design.

Fixes #7064 
Fixes #9781 
Fixes #7215 
Fixes #10119 
Fixes #9908 
Fixes #11844
Fixes #12355 

## Type of change

## How Has This Been Tested?

- Test A
- Test B

## Checklist:

    // Code coverage diff between base branch:release and head branch: feat/7064-label-enhancements-for-widgets 
 :green_circle: | total | 56.51 **(0.01)** | 38.12 **(0.15)** | 36.3 **(0.03)** | 56.73 **(0.01)**
 :sparkles: :new: | **app/client/src/components/ads/LabelWithTooltip.tsx** | **71.58** | **53.54** | **58.33** | **73.85**
 :sparkles: :new: | **app/client/src/components/propertyControls/LabelAlignmentOptionsControl.tsx** | **66.67** | **33.33** | **40** | **72.22**
 :sparkles: :new: | **app/client/src/components/propertyControls/NumericInputControl.tsx** | **50** | **40** | **50** | **50**
 :green_circle: | app/client/src/components/propertyControls/index.ts | 87.18 **(0.69)** | 100 **(0)** | 0 **(0)** | 86.84 **(0.73)**
 :red_circle: | app/client/src/constants/DefaultTheme.tsx | 81.29 **(-15.26)** | 52.78 **(-37.22)** | 68.75 **(-16.44)** | 82.35 **(-13.69)**
 :green_circle: | app/client/src/utils/DSLMigrations.ts | 73.68 **(0.17)** | 69.02 **(-0.06)** | 71.13 **(0)** | 73.57 **(0.14)**
 :sparkles: :new: | **app/client/src/utils/migrations/RadioGroupWidget.ts** | **100** | **83.33** | **100** | **100**
 :green_circle: | app/client/src/widgets/BaseWidget.tsx | 79.84 **(0.31)** | 46.81 **(2.37)** | 76.47 **(0.71)** | 79.03 **(0.34)**
 :red_circle: | app/client/src/widgets/BaseInputWidget/component/index.tsx | 63.13 **(0.53)** | 37.67 **(-7.18)** | 58.14 **(-1.86)** | 61.07 **(-0.53)**
 :green_circle: | app/client/src/widgets/BaseInputWidget/widget/index.tsx | 70.45 **(2.95)** | 7.14 **(0)** | 66.67 **(5.13)** | 73.81 **(2.76)**
 :green_circle: | app/client/src/widgets/CheckboxGroupWidget/component/index.tsx | 31.58 **(2.55)** | 8.89 **(3.33)** | 0 **(0)** | 37.5 **(1.5)**
 :red_circle: | app/client/src/widgets/CheckboxGroupWidget/widget/index.tsx | 47.62 **(0.12)** | 20.59 **(0)** | 34.78 **(-3.32)** | 48.1 **(0.1)**
 :green_circle: | app/client/src/widgets/CurrencyInputWidget/widget/index.tsx | 45.61 **(0.48)** | 16.98 **(0)** | 43.75 **(0)** | 45.54 **(0.49)**
 :red_circle: | app/client/src/widgets/DatePickerWidget2/component/index.tsx | 26.74 **(-3.26)** | 5.71 **(2.72)** | 5.88 **(-0.37)** | 26.19 **(-2.8)**
 :red_circle: | app/client/src/widgets/DatePickerWidget2/widget/index.tsx | 68.29 **(-1.15)** | 0 **(0)** | 50 **(-10)** | 69.23 **(-1.36)**
 :red_circle: | app/client/src/widgets/DropdownWidget/component/index.tsx | 83.33 **(0.66)** | 53.13 **(2.41)** | 64.71 **(-5.29)** | 81.94 **(0.51)**
 :red_circle: | app/client/src/widgets/DropdownWidget/widget/index.tsx | 69.84 **(-1.35)** | 36.84 **(0)** | 56.25 **(-8.04)** | 68.97 **(-1.4)**
 :red_circle: | app/client/src/widgets/InputWidget/component/index.tsx | 57.23 **(1.88)** | 26 **(-10.7)** | 47.73 **(-5.33)** | 55.35 **(1.16)**
 :red_circle: | app/client/src/widgets/InputWidget/widget/index.tsx | 35.29 **(0.51)** | 16.25 **(0)** | 26.92 **(-2.25)** | 36.28 **(0.5)**
 :green_circle: | app/client/src/widgets/InputWidgetV2/widget/index.tsx | 80.8 **(0)** | 67.78 **(0.39)** | 78.95 **(0)** | 80.33 **(0)**
 :green_circle: | app/client/src/widgets/JSONFormWidget/fields/RadioGroupField.tsx | 42.86 **(2.12)** | 0 **(0)** | 0 **(0)** | 44.44 **(2.13)**
 :green_circle: | app/client/src/widgets/MultiSelectTreeWidget/component/index.styled.tsx | 50 **(2.38)** | 45.16 **(20.16)** | 0 **(0)** | 58.06 **(6.63)**
 :green_circle: | app/client/src/widgets/MultiSelectTreeWidget/component/index.tsx | 16.28 **(0.43)** | 0 **(0)** | 0 **(0)** | 18.18 **(0.37)**
 :red_circle: | app/client/src/widgets/MultiSelectTreeWidget/widget/index.tsx | 40.54 **(0.54)** | 0 **(0)** | 35.29 **(-4.71)** | 41.43 **(0.52)**
 :green_circle: | app/client/src/widgets/MultiSelectWidget/component/index.styled.tsx | 56.52 **(13.66)** | 55.56 **(32.23)** | 0 **(0)** | 70.59 **(22.31)**
 :green_circle: | app/client/src/widgets/MultiSelectWidget/component/index.tsx | 19.18 **(0.34)** | 0 **(0)** | 0 **(0)** | 20.29 **(0.29)**
 :red_circle: | app/client/src/widgets/MultiSelectWidget/widget/index.tsx | 48.21 **(0.13)** | 0 **(0)** | 42.86 **(-7.14)** | 49.06 **(0.08)**
 :green_circle: | app/client/src/widgets/MultiSelectWidgetV2/component/index.styled.tsx | 52.78 **(2.78)** | 58.06 **(26.81)** | 0 **(0)** | 62.07 **(7.52)**
 :red_circle: | app/client/src/widgets/MultiSelectWidgetV2/component/index.tsx | 10.45 **(-0.95)** | 0 **(0)** | 0 **(0)** | 11.76 **(-1.24)**
 :red_circle: | app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx | 73.74 **(-1)** | 48.57 **(0)** | 57.14 **(-6.02)** | 73.4 **(-1.04)**
 :green_circle: | app/client/src/widgets/PhoneInputWidget/widget/index.tsx | 40.59 **(0.59)** | 6.52 **(0)** | 43.75 **(0)** | 40.4 **(0.6)**
 :red_circle: | app/client/src/widgets/RadioGroupWidget/component/index.tsx | 47.37 **(-21.05)** | 28.57 **(-2.2)** | 0 **(-25)** | 50 **(-25)**
 :red_circle: | app/client/src/widgets/RadioGroupWidget/widget/index.tsx | 28.85 **(-4.08)** | 5.13 **(0)** | 50 **(-8.33)** | 29.29 **(-4.48)**
 :red_circle: | app/client/src/widgets/RichTextEditorWidget/widget/index.tsx | 57.41 **(0.27)** | 8.7 **(0)** | 43.75 **(-6.25)** | 57.69 **(0.24)**
 :green_circle: | app/client/src/widgets/SelectWidget/component/index.styled.tsx | 53.57 **(5.09)** | 48 **(22.19)** | 0 **(0)** | 60 **(8.39)**
 :green_circle: | app/client/src/widgets/SelectWidget/component/index.tsx | 16.8 **(0.27)** | 0 **(0)** | 4.76 **(0)** | 17.54 **(0.27)**
 :red_circle: | app/client/src/widgets/SelectWidget/widget/index.tsx | 52.44 **(-0.12)** | 20.97 **(0)** | 42.11 **(-4.95)** | 53.16 **(-0.17)**
 :green_circle: | app/client/src/widgets/SingleSelectTreeWidget/component/index.styled.tsx | 52.63 **(2.63)** | 48.28 **(22.47)** | 0 **(0)** | 58.06 **(5.12)**
 :green_circle: | app/client/src/widgets/SingleSelectTreeWidget/component/index.tsx | 16.47 **(0.42)** | 0 **(0)** | 0 **(0)** | 18.42 **(0.36)**
 :red_circle: | app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx | 46.97 **(0.2)** | 0 **(0)** | 40 **(-6.15)** | 48.39 **(0.11)**
 :green_circle: | app/client/src/widgets/SwitchGroupWidget/component/index.tsx | 45 **(18.33)** | 21.05 **(9.29)** | 0 **(0)** | 50 **(13.64)**
 :red_circle: | app/client/src/widgets/SwitchGroupWidget/widget/index.tsx | 41.27 **(-6.73)** | 0 **(0)** | 46.15 **(-8.4)** | 40.98 **(-6.94)**
 :x: | ~~app/client/src/widgets/DropdownWidget/component/index.styled.tsx~~ | ~~50~~ | ~~11.76~~ | ~~0~~ | ~~50~~</details>


shwetha-ramesh:
**Generic Issues**
1.	What would be the general sequence of fields in property pane, since there is no regular pattern and every widget differs? 
 image
image

2.     The field ```Position``` height and width does not match with other fields in the property pane.
image

 3.	Without setting ```Width``` Alignment does not make any changes. 
https://loom.com/share/eb7a53de03d1466da2aa8cf510bfe256

4.	```Position``` and ```Alignment``` varies from Figma design
 image

5.	When Position = Top is selected, Alignment  & Width options are not present.
image


**### Input Widget**

1.	Bug #7064 – not fixed.
When a user adds 3 input widgets on a form, there is no definitive way to set the width for all the 3 widgets for it to match for the same length; it’s more of a trial and error approach to set the width. In the below video though the Width is set to 10 for all the 3 widgets it still looks misaligned. 

https://loom.com/share/e1ed7ae69b5a444abe5ad54366a27398

2.	Bug #9781 – fixed
3.	Bug #10119 – fixed
4.	Bug #7215 – 

5.	Text Size shows ‘No selection’ by default
 image

6.	When ```Data Type``` is set to Password and when the widget height is resized the icons remain in the original position where as the text moves as the widget is resized

https://loom.com/share/c5d5becd6645437b91aba096cfba4735
 
7.

riodeuno:
@wmdev0808 I am moving this PR to draft as there are QA comments from @shwetha-ramesh which need to be addressed.

app/client/src/components/propertyControls/LabelAlignmentOptionsControl.tsx
```diff
@@ -0,0 +1,79 @@
...
+    const { propertyValue } = this.props;
+
+    return (
+      <StyledButtonGroup className={replayHighlightClass} fill>
+        <StyledButton
+          active={
+            propertyValue === Alignment.LEFT || propertyValue === undefined
+          }
+          icon={<ControlIcons.LEFT_ALIGN color="#979797" />}
+          onClick={() => this.handleAlign(Alignment.LEFT)}
```
riodeuno, line 63: Since we can't use `useCallback`, can we try this: 
`onClick = {this.handleAlign.bind(this, Alignment.LEFT)}`

app/client/src/components/propertyControls/LabelAlignmentOptionsControl.tsx
```diff
@@ -0,0 +1,79 @@
...
+        />
+        <StyledButton
+          active={propertyValue === Alignment.RIGHT}
+          icon={<ControlIcons.RIGHT_ALIGN color="#979797" />}
```
riodeuno, line 67: Will we be getting anything for the colors in this control from the appsmith design system theme?
@jsartisan @albinAppsmith @sbalaji1192

app/client/src/utils/DSLMigrations.ts
```diff
@@ -91,7 +92,7 @@ const addLogBlackListToAllListWidgetChildren = (
  * @param currentDSL
  * @returns
  */
-const addPrivateWidgetsToAllListWidgets = (
+export const addPrivateWidgetsToAllListWidgets = (
```
riodeuno, line 95: @ohansFavour Did we revert this migration? If so, shall we remove this function?

app/client/src/utils/DSLMigrationsUtils.test.ts
```diff
@@ -1156,14 +1155,13 @@ describe("correctly migrate dsl", () => {
...
       type: "CANVAS_WIDGET",
-      version: nextVersion,
       widgetId: "0",
       widgetName: "MainContainer",
       renderMode: RenderModes.CANVAS,
       isLoading: false,
     };
 
-    const actualNextDsl = transformDSL(currentDSL, false);
+    const actualNextDsl = addPrivateWidgetsToAllListWidgets(currentDSL);
```
riodeuno, line 1164: @ohansFavour Once again, did we revert this change? Is this update valid?

@wmdev0808 This is very important, we need confirmation from @ohansFavour before moving ahead with this change. This can potentially break applications.


I was also wondering if this change is necessary for your PR?

app/client/src/utils/migrations/RadioGroupWidget.ts
```diff
@@ -0,0 +1,20 @@
+import { Alignment } from "@blueprintjs/core";
+import { WidgetProps } from "widgets/BaseWidget";
+import { DSLWidget } from "widgets/constants";
+
+export const migrateInlineAndAlignmentProperties = (currentDSL: DSLWidget) => {
+  currentDSL.children = currentDSL.children?.map((child: WidgetProps) => {
+    if (child.type === "RADIO_GROUP_WIDGET") {
+      if (!("isInline" in child)) {
```
riodeuno, line 8: @wmdev0808 What do you think about `!child.hasOwnProperty("isInline")`? Is this more readable?

riodeuno: Also, does the radio group widget crash if it doesn't have these properties defined?

app/client/src/utils/migrations/RadioGroupWidget.ts
```diff
@@ -0,0 +1,20 @@
+import { Alignment } from "@blueprintjs/core";
+import { WidgetProps } from "widgets/BaseWidget";
+import { DSLWidget } from "widgets/constants";
+
+export const migrateInlineAndAlignmentProperties = (currentDSL: DSLWidget) => {
```
riodeuno, line 5: Reviewer's understanding:
This function simple adds two new properties, `isInline` and `alignment` with default values  ( `false` and `Left` respectively) to all existing radio group widgets.

@wmdev0808 Let's add a simple jest test, which verifies that the DSL is correctly updated. Please make sure to use a fairly complex DSL as an input.

app/client/src/widgets/BaseInputWidget/component/index.tsx
```diff
@@ -66,8 +72,22 @@ const InputComponentWrapper = styled((props) => (
   allowCurrencyChange?: boolean;
   disabled?: boolean;
   inputType: InputType;
+  compactMode: boolean;
+  labelPosition: LabelPosition;
 }>`
-  flex-direction: ${(props) => (props.compactMode ? "row" : "column")};
+  flex-direction: ${(props) =>
+    props.labelPosition === LabelPositionTypes.Left
```
riodeuno, line 79: @wmdev0808 Let us avoid nested ternary operators. It will be difficult to maintain this code.
Let's use simple `if` and `else` clauses here.

app/client/src/widgets/BaseInputWidget/component/index.tsx
```diff
@@ -315,9 +382,74 @@ class BaseInputComponent extends React.Component<
...
+      prevProps.tooltip !== this.props.tooltip
+    ) {
+      this.setState({ hasLabelEllipsis: this.checkHasLabelEllipsis() });
+    }
+  }
+
+  componentWillUnmount() {
+    if (isNumberInputType(this.props.inputHTMLType) && this.props.onStep) {
+      const element: any = document.querySelectorAll(
+        `.appsmith_widget_${this.props.widgetId} .bp3-button`,
```
riodeuno, line 425: @sbalaji1192  The selector used when adding event listeners was `.bp3-button-group`. Will this work and remove the appropriate listeners?

@sbalaji1192 Please review this particular file, as it seems to effect the Input Widget.

app/client/src/widgets/BaseInputWidget/component/index.tsx
```diff
@@ -315,9 +382,74 @@ class BaseInputComponent extends React.Component<
 > {
   constructor(props: BaseInputComponentProps) {
     super(props);
-    this.state = { showPassword: false };
+    this.state = { showPassword: false, hasLabelEllipsis: false };
+  }
+
+  componentDidMount() {
+    if (isNumberInputType(this.props.inputHTMLType) && this.props.onStep) {
+      const element: any = document.querySelector(
```
riodeuno, line 390: @sbalaji1192 We are using query selector here, is it possible to use React.ref? If we've made a decision to use query selectors, can we add an elaboration on the reason?

riodeuno (changes requested):
Some comments on the code.
@wmdev0808 We will want to add cypress tests for the new properties added to the widgets.

app/client/src/utils/DSLMigrationsUtils.test.ts
riodeuno: I just noticed that we have a merge conflict with DSLMigrationsUtils. Maybe this is why we see reverted changes in this PR. Please verify.

app/client/src/utils/DSLMigrations.ts
ohansFavour: @riodeuno, we did not revert this Migration. This function is used in `DSLMigrations.test.ts` to unit test the privateWidgets Migration.

app/client/src/utils/DSLMigrationsUtils.test.ts
ohansFavour: @wmdev0808 could you pull the latest release code?, these changes are no longer needed in your PR.

app/client/src/utils/migrations/RadioGroupWidget.ts
wmdev0808: > @wmdev0808 What do you think about `!child.hasOwnProperty("isInline")`? Is this more readable?

I agree with you, thank you

wmdev0808: > Also, does the radio group widget crash if it doesn't have these properties defined?

No, it doesn't crash.
The purpose of this migration is just for consistency, for example, the similar widgets(checkbox group, switch group) have isInline property, of which default value is true

riodeuno:
@sbalaji1192 Could you review the input widget v2 files which have been updated?
@wmdev0808 Let's resolve the merge conflicts in this PR.

sbalaji1192 (commented):
Reviewed the Input related files. added a few callouts. @wmdev0808 please check.

riodeuno:
@wmdev0808 The code looks good to me. If all the callouts by @sbalaji1192 have been resolved, we can move this back into the QA pipeline.

@shwetha-ramesh These changes, as you know, are related to all input type widgets and their property panes. As a result, this can effect our experience in a Form Widget as well.

app/client/src/utils/DSLMigrationsUtils.test.ts
riodeuno: @wmdev0808 @ohansFavour Has this been updated. I can't tell. Could you let me know?

app/client/src/components/propertyControls/LabelAlignmentOptionsControl.tsx
albinAppsmith: We should use colors from Colors.tsx for now

app/client/src/utils/DSLMigrationsUtils.test.ts
ohansFavour: Yes, it has

app/client/src/widgets/BaseInputWidget/component/index.tsx
```diff
@@ -309,15 +370,103 @@ export const isNumberInputType = (inputHTMLType: InputHTMLType = "TEXT") => {
...
+      if (this.state.hasLabelEllipsis !== hasLabelEllipsis) {
+        this.setState({ hasLabelEllipsis });
+      }
+    }
+  }
+
+  componentWillUnmount() {
+    if (isNumberInputType(this.props.inputHTMLType) && this.props.onStep) {
+      const element: any = document.querySelectorAll(
+        `[class*=StyledNumericInput] .bp3-button`,
```
sbalaji1192, line 442: should be `.appsmith_widget_${this.props.widgetId} .bp3-button-group`

sbalaji1192 (approved):
Input changes looks good to me. Just had one callout on the BaseInput widget.

Tooluloope:
@dilippitchika @momcilo-appsmith 
Could you review the DP, it's really hard to know what is selected.
Screenshot 2022-03-23 at 09 40 28

wmdev0808:
@Tooluloope ,
- Issue 1: The label is not centered.
When the input widget is not in compact mode, its underlying component is converted to textarea.
So from the UX favor, I've aligned the label to the top.
- Issue 2: Label alignment does not work
It's just because you are using long text(truncated by ellipsis), meaning labels with ellipsis cannot be aligned

dilippitchika:
Enhancements needed @wmdev0808 

1. Button group
    1. Selected color - Can you please use the same hover and selected colors from this field 
Screenshot 2022-03-25 at 3 06 46 PM
    2. Font size - Can you please change the font size to be 14px
    3. Button size - Can you make the button normal height?
Screenshot 2022-03-25 at 3 08 51 PM

2. Width in columns in the label - Can you please add "(in columns)" in the width label
Screenshot 2022-03-25 at 3 10 50 PM


3. Default values - Can you add the logic that for all new widgets being dragged, the default is set to these values
Screenshot 2022-03-25 at 3 34 25 PM

Tooluloope:
@dilippitchika  @wmdev0808 

Is there a way to also align this vertically?

I feel it would be needed. 
Screenshot 2022-03-25 at 11 45 59

wmdev0808:
> @dilippitchika @wmdev0808
> 
> Is there a way to also align this vertically?
> 
> I feel it would be needed. It seems more needed than the left and right alignment Screenshot 2022-03-25 at 11 45 59

@Tooluloope ,
For that, you can use `Position`: `Top`

Tooluloope:
Using Top turns the Arrangement like this which isn't what I want.

Screenshot 2022-03-25 at 13 00 48

wmdev0808:
> Using Top turns the Arrangement like this which isn't what I want.
> 
> Screenshot 2022-03-25 at 13 00 48

Well, please add a screenshot for your desirable thing

Tooluloope:
This way,
It will be helpful if we could get it aligned this way

Screenshot 2022-03-25 at 13 04 35

wmdev0808:
> Enhancements needed @wmdev0808
> 
> 1. Button group
>    
>    1. Selected color - Can you please use the same hover and selected colors from this field
>       Screenshot 2022-03-25 at 3 06 46 PM
>    2. Font size - Can you please change the font size to be 14px
>    3. Button size - Can you make the button normal height?
>       Screenshot 2022-03-25 at 3 08 51 PM
> 2. Width in columns in the label - Can you please add "(in columns)" in the width label
>    Screenshot 2022-03-25 at 3 10 50 PM
> 3. Default values - Can you add the logic that for all new widgets being dragged, the default is set to these values
>    Screenshot 2022-03-25 at 3 34 25 PM

@dilippitchika ,
Could you please add the resulting screenshot of `1. Button group`?
I can't fully understand 2, 3 there

app/client/src/widgets/MultiSelectWidget/component/index.tsx
```diff
@@ -72,11 +86,61 @@ function MultiSelectComponent({
...
+      setIsSelectAll(true);
+    }
+    if (isSelectAll && options.length !== value.length) {
+      setIsSelectAll(false);
+    }
+  }, [options, value]);
+
+  useEffect(() => {
+    if (labelText && !isLabelTooltipEnabled) {
+      addLabelTooltipEventListeners(
```
ashit-rath, line 114: A lot of code is being repeated everywhere, for example adding of listners/ useEffects/ defining mouseleave and enter functions across all components. This is getting repeated.
```

        <TextLabelWrapper
          alignment={labelAlignment}
          compactMode={compactMode}
          position={labelPosition}
          ref={labelRef}
          width={labelWidth}
        >
          <StyledTooltip
            content={labelText}
            hoverOpenDelay={200}
            isOpen={isLabelTooltipOpen}
            position={Position.TOP}
          >
```
If I were to create a new widget I would have to copy paste the entire logic from some other widget into this.
Every component should not be responsible for the rendering/handling of the label. There should just be 1 component as `Label` and it should deal with everything eg. the label wrapper, showing of tooltip, etc.

app/client/src/widgets/MultiSelectWidget/component/index.tsx
```diff
@@ -72,11 +86,61 @@ function MultiSelectComponent({
...
+      setIsLabelTooltipEnabled(true);
+    } else if (!labelText && isLabelTooltipEnabled) {
+      setIsLabelTooltipEnabled(false);
+    }
+  }, [labelText]);
+
+  useEffect(() => {
+    return () =>
+      removeLabelTooltipEventListeners(
+        `.appsmith_widget_${widgetId} .multiselect-label`,
```
ashit-rath, line 128: Rather than passing class names and then using document.querySelector use refs.

app/client/src/widgets/MultiSelectWidget/component/index.tsx
```diff
@@ -72,11 +86,61 @@ function MultiSelectComponent({
...
+      );
+      setIsLabelTooltipEnabled(true);
+    } else if (!labelText && isLabelTooltipEnabled) {
+      setIsLabelTooltipEnabled(false);
+    }
+  }, [labelText]);
+
+  useEffect(() => {
+    return () =>
+      removeLabelTooltipEventListeners(
```
ashit-rath, line 127: Is there any particular reason to use native event handlers rather than React's synthetic events like `onMouseEnter` etc?

app/client/src/widgets/SelectWidget/widget/index.tsx
```diff
@@ -217,6 +209,67 @@ class SelectWidget extends BaseWidget<SelectWidgetProps, WidgetState> {
           },
         ],
       },
+      {
+        sectionName: "Label",
```
ashit-rath, line 213: @riodeuno @wmdev0808 @sbalaji1192 
I think slowly we are moving towards the state where all of our widgets would have a bunch of common properties. Would this be the perfect opportunity to move these to a commonplace and not repeat it in every widget property config? 
If yes, then where should we keep this?

ashit-rath (changes requested):

app/client/src/widgets/CheckboxGroupWidget/component/index.tsx
```diff
@@ -1,27 +1,38 @@
-import React from "react";
+import React, { useCallback, useEffect, useState } from "react";
 import styled from "styled-components";
+import { Alignment, Label, Position } from "@blueprintjs/core";
 
 import { Classes } from "@blueprintjs/core";
```
Tooluloope, line 5: `@blueprintjs/core"` is being repeated

app/client/src/widgets/CheckboxGroupWidget/component/index.tsx
```diff
@@ -1,27 +1,38 @@
...
 import { StyledCheckbox } from "widgets/CheckboxWidget/component";
 import { OptionProps, SelectAllState, SelectAllStates } from "../constants";
+import {
+  addLabelTooltipEventListeners,
+  hasLabelEllipsis,
+  removeLabelTooltipEventListeners,
+} from "widgets/WidgetUtils";
 
-export interface CheckboxGroupContainerProps {
+export interface InputContainerProps {
```
Tooluloope, line 28: Could we leave this as `CheckboxGroupContainerProps` and not Input

app/client/src/widgets/MultiSelectTreeWidget/component/index.tsx
```diff
@@ -192,90 +250,108 @@ function MultiTreeSelectComponent({
...
-            {labelText}
-          </StyledLabel>
+            <StyledLabel
+              $compactMode={compactMode}
+              $disabled={disabled}
+              $labelStyle={labelStyle}
+              $labelText={labelText}
+              $labelTextColor={labelTextColor}
+              $labelTextSize={labelTextSize}
+              className={`multitree-select-label ${
```
Tooluloope, line 285: It will be best to use Ref.

app/client/src/widgets/JSONFormWidget/fields/MultiSelectField.tsx
```diff
@@ -199,6 +199,7 @@ function MultiSelectField({
           filterText={filterText}
           isFilterable={schemaItem.isFilterable}
           isValid={isDirty ? isValueValid : true}
+          labelText=""
```
Tooluloope, line 202: @ashit-rath would you be able to confirm the JSONForm Changes?

Tooluloope (changes requested):

app/client/src/widgets/BaseInputWidget/component/index.tsx
```diff
@@ -312,13 +381,92 @@ export const isNumberInputType = (inputHTMLType: InputHTMLType = "TEXT") => {
...
+interface LabelComponentProps {
+  isLoading: boolean;
+  label: string;
+}
+
+function LabelComponent({ isLoading, label }: LabelComponentProps) {
+  return (
+    <Label
+      className={`t--input-widget-label ${
+        isLoading ? Classes.SKELETON : Classes.TEXT_OVERFLOW_ELLIPSIS
```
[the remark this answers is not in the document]
sbalaji1192: @Tooluloope This got carried over from the Inputwidget v1. 
@dilippitchika can answer the question from a product perspective.

app/client/src/widgets/SelectWidget/widget/index.tsx
riodeuno: @ashit-rath Yes! This will really help out with re-usability. The onus on the implementation of these properties would still be on the widgets, so there is a concern there. 
We have something similar coming up for the dynamic height project. 

I would suggest we create a marker of some type, which when added to the property pane configuration, are automatically replaced with the full form property/properties. For example, `..., __section_Label__, ...`(just an example, not the exact type of marker) would be replaced by `..., { sectionName: "Label", ..other Label properties..}, ...`

app/client/src/widgets/BaseInputWidget/component/index.tsx
ashit-rath: @sbalaji1192 Follow-up question: Out of curiosity; why do we need to handle step changes explicitly like this? I see that Currency Input uses the `onStep` but wouldn't the `onValueChange` method do the same thing if we remove the step-change implementation?

app/client/src/widgets/SelectWidget/widget/index.tsx
ashit-rath: @riodeuno That's a good idea. This seems like a bigger job and app theming is also introducing a bunch of common properties. Do you think we should take this up as a separate issue as more discussion is needed and it would just weigh down the timelines of this feature?
cc: @jsartisan @keyurparalkar

app/client/src/widgets/JSONFormWidget/fields/MultiSelectField.tsx
ashit-rath: @Tooluloope Yes, these changes are fine. JSONForm does not use component labels.

app/client/src/components/propertyControls/LabelAlignmentOptionsControl.tsx
```diff
@@ -0,0 +1,77 @@
...
+    background-color: #ffffff !important;
+  }
+  &:hover {
+    background-color: #ffffff !important;
+  }
+`;
+
+export interface LabelAlignmentOptionsControlProps extends ControlProps {
+  propertyValue: Alignment | undefined;
+  onChange: (align: Alignment) => void;
```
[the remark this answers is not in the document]
wmdev0808: Oh, it can be removed, thank you

app/client/src/widgets/BaseInputWidget/component/index.tsx
```diff
@@ -312,13 +381,92 @@ export const isNumberInputType = (inputHTMLType: InputHTMLType = "TEXT") => {
...
 > {
   constructor(props: BaseInputComponentProps) {
     super(props);
-    this.state = { showPassword: false };
+    this.state = { showPassword: false, isLabelTooltipOpen: false };
+  }
+
+  componentDidMount() {
+    if (isNumberInputType(this.props.inputHTMLType) && this.props.onStep) {
+      const element: any = document.querySelector(
```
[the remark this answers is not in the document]
wmdev0808: > Could we avoid using `any`, lets be as strict as possible with our types

It was written by @sbalaji1192 , so we need to confirm with him

app/client/src/widgets/CheckboxGroupWidget/component/index.tsx
wmdev0808: Every widget consists of Label and Input section and CheckboxGroupContainerProps are whole container for the widget
So its name comes after that purpose

app/client/src/widgets/BaseInputWidget/component/index.tsx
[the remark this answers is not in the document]
sbalaji1192: @wmdev0808 Must be a mistake from my side. Agree with @Tooluloope

app/client/src/widgets/BaseInputWidget/component/index.tsx
ashit-rath: @sbalaji1192 Got it, thanks for the detailed explanation.

app/client/src/components/propertyControls/LabelPositionOptionsControl.tsx
```diff
@@ -0,0 +1,80 @@
...
+  border: ${(props) =>
+    props.active ? `1px solid #6A86CE` : `1px solid #A9A7A7`};
+  border-radius: 0;
+  box-shadow: none !important;
+  background-image: none !important;
+  background-color: #ffffff !important;
+  & > div {
+    display: flex;
+  }
+  &.bp3-active {
```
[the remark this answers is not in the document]
aswathkk: I would suggest re-using ads/ButtonTabComponent instead. Because they are already made keyboard accessible.

aswathkk (changes requested):
@wmdev0808 I see that you introduced new property controls.
Before adding new property controls, check whether you can re-use existing ones.
If you are introducing new ones, make sure to make them keyboard accessible.

dilippitchika:
@aswathkk this design was done as per the new designs for property pane. Something which you are yet to implement overall in the property pane. I don't know if we can add keyboard accessibility here directly unless we change the component itself. 
Please check the design here - 
https://www.figma.com/file/hlOcM5LasxcRqEwHPXe5lm/UI-%7C-Content-Property-Pane?node-id=2777%3A64824

@wmdev0808 please use the component @aswathkk suggested and update the DP once. Let's see how this looks like.

dilippitchika:
After discussion, @wmdev0808 please update the following
- Label position to use a "dropdown" component instead of the component you have used right now. Order of items in the dropdown is "Left, Top, Auto"
- Label alignment to use ads/ButtonTabComponent as suggested by Aswath. 


No changes to the label width component

app/client/src/widgets/SwitchGroupWidget/index.ts
```diff
@@ -10,8 +11,8 @@ export const CONFIG = {
   isCanvas: false, // Defines if this widget has a canvas within in which we can drop other widgets
   defaults: {
     widgetName: "SwitchGroup",
-    rows: 6,
-    columns: 18,
+    rows: 4,
```
ashit-rath, line 14: Why are we reducing the size of this widget?

aswathkk (commented):
Changes to LabelAlignmentOptionsControl looks good to me.

app/client/src/components/ads/LabelWithTooltip.tsx
```diff
@@ -0,0 +1,233 @@
...
+  }, []);
+
+  const handleMouseLeave = useCallback(() => {
+    setTooltipOpen(false);
+  }, []);
+
+  return (
+    <LabelContainer
+      alignment={alignment}
+      className="label-container"
```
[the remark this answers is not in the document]
wmdev0808: It is currently used only for input widgets which need specific CSS rule, `flex-grow: 0;`

app/client/src/components/ads/LabelWithTooltip.tsx
```diff
@@ -0,0 +1,233 @@
...
+
+    ${!inline && optionCount && optionCount > 1 && `align-self: flex-start;`}
+  `}
+`;
+
+export const StyledTooltip = styled(Tooltip)`
+  overflow: hidden;
+`;
+
+export interface StyledLabelProps {
```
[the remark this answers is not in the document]
wmdev0808: Ok, thanks. I will

app/client/src/components/propertyControls/LabelAlignmentOptionsControl.tsx
```diff
@@ -0,0 +1,79 @@
...
+}
+
+class LabelAlignmentOptionsControl extends BaseControl<
+  LabelAlignmentOptionsControlProps
+> {
+  constructor(props: LabelAlignmentOptionsControlProps) {
```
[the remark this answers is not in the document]
wmdev0808: @ashit-rath ,
The comments from @riodeuno has already been fixed.
If you have any doubt, please specify here

app/client/src/components/propertyControls/LabelPositionOptionsControl.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { LabelPosition } from "components/constants";
+import { replayHighlightClass } from "globalStyles/portals";
+import { StyledButton, StyledButtonGroup } from "./LabelButton";
+
+export interface LabelPositionOptionsControlProps extends ControlProps {
+  propertyValue: LabelPosition | undefined;
+  options: any[];
+}
+
+class LabelPositionOptionsControl extends BaseControl<
```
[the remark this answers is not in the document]
wmdev0808: According to @dilippitchika 's new requirements, we don't use this control.
There exists the underlying component for the purpose, `ButtonTabComponent`
For your requirements, we need to change this underlying component, but it is out of scope at the moment

app/client/src/widgets/RadioGroupWidget/component/index.tsx
```diff
@@ -2,83 +2,194 @@ import React from "react";
...
-      align-self: flex-start;
-      max-width: calc(30% - ${WIDGET_PADDING}px);
-    }
-  }
+export interface RadioGroupContainerProps {
+  compactMode: boolean;
+  labelPosition?: LabelPosition;
+}
+
+export const RadioGroupContainer = styled.div<RadioGroupContainerProps>`
```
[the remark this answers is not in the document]
wmdev0808: I don't agree on your proposal.
> The reason:
- These container components are specific to the widget, not to label
- However, The common css part  can be extracted

app/client/src/utils/migrations/RadioGroupWidget.ts
```diff
@@ -0,0 +1,17 @@
+import { Alignment } from "@blueprintjs/core";
+import { WidgetProps } from "widgets/BaseWidget";
+import { DSLWidget } from "widgets/constants";
+
+export const migrateRadioGroupAlignmentProperty = (currentDSL: DSLWidget) => {
+  currentDSL.children = currentDSL.children?.map((child: WidgetProps) => {
+    if (child.type === "RADIO_GROUP_WIDGET") {
+      if (!child.hasOwnProperty("alignment")) {
+        child.alignment = Alignment.LEFT;
```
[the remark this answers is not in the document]
wmdev0808: It is not for label alignment, it is another alignment for the radio group widget

app/client/src/widgets/SwitchGroupWidget/index.ts
wmdev0808: It was for showing the widget in compact mode for the first time

wmdev0808: I will just revert it at the moment

aswathkk (approved):

app/client/src/components/propertyControls/LabelButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+    &.bp3-active {
+      background: ${Colors.GREY_3};
+    }
+
+    & > div {
+      display: flex;
+      cursor: pointer;
+    }
+  }
+`;
```
Tooluloope, line 31: I don't think this is necessary anymore.

app/client/src/components/propertyControls/LabelPositionOptionsControl.tsx
```diff
@@ -0,0 +1,49 @@
...
+      </StyledButtonGroup>
+    );
+  }
+
+  private toggleOption = (option: LabelPosition) => {
+    this.updateProperty(this.props.propertyName, option);
+  };
+}
+
+export default LabelPositionOptionsControl;
```
Tooluloope, line 49: I don't think this is necessary anymore.

app/client/src/components/propertyControls/NumericInputControl.tsx
```diff
@@ -0,0 +1,77 @@
...
+}
+
+export interface NumericInputControlProps extends ControlProps {
+  propertyValue: string;
+  min?: number;
+  max?: number;
+  minorStepSize?: number | null;
+  majorStepSize?: number | null;
+  placeholderText?: string;
+  stepSize?: number;
```
Tooluloope, line 74: Are these properties needed?

app/client/src/utils/DSLMigrations.ts
```diff
@@ -1064,6 +1066,11 @@ export const transformDSL = (
 
   if (currentDSL.version === 53) {
     currentDSL = migrateScrollTruncateProperties(currentDSL);
+    currentDSL.version = 54;
+  }
+
+  if (currentDSL.version === 54) {
+    currentDSL = migrateRadioGroupAlignmentProperty(currentDSL);
```
Tooluloope, line 1073: @wmdev0808 could you tell me what the migration / Alignment does for the RadioGroup Widget.

@somangshu @dilippitchika was there an agreed Migration

Tooluloope (changes requested):

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/FormWidgets/Select_spec.js
```diff
@@ -0,0 +1,7 @@
+const dsl = require("../../../../fixtures/multiSelectDsl.json");
+
+describe("Select Widget", () => {
+  before(() => {
+    cy.addDsl(dsl);
+  });
+});
```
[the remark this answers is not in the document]
wmdev0808: Oh, it is unnecessary, I will remove it

app/client/src/utils/DSLMigrations.ts
somangshu: not sure about this, I am not fully aware of the context, @dilippitchika might be able to answer

app/client/src/components/propertyControls/NumericInputControl.tsx
wmdev0808: Yes, These can be used if user configure them in property config.
Even though these props are not directly used in the current use for label width, it will be used for fine-grained control in the future.
So I think it is worth having these props, as well

app/client/src/utils/DSLMigrations.ts
wmdev0808: @Tooluloope ,
It is just for alignment of the indicator within container.
alignment-left
alignment-right

somangshu: That looks straightforward, one more question, Is this JS convertible?

wmdev0808: No, it is a dropdown

Tooluloope (changes requested):

app/client/src/utils/DSLMigrations.ts
Tooluloope: @dilippitchika Could you check this out,
The migration is simple as @somangshu said, So most likely nothing would break.

Just a design review, the Radio buttons don't look aligned

dilippitchika: I think there is a change in the radio buttons happening, previously the distance between 2 radio buttons was more compared to what it is now vertically. Upon resize the distance doesn't change. @wmdev0808 what is happening here?

OLD
Screenshot 2022-04-01 at 3 14 35 PM
NEW
Screenshot 2022-04-01 at 3 14 27 PM

Tooluloope (dismissed):
LGTM,

@appsmithorg/qa this would need design audit as well

app/client/src/utils/DSLMigrations.ts
wmdev0808: @dilippitchika ,
It is because `inline` prop is `true`.
So NEW is the result of being wrapped of inline version

dilippitchika: Screenshot 2022-04-01 at 4 42 36 PM
@wmdev0808 the behavior is different for checkbox group and radio group then when inline is turned on. We should follow what checkbox group is doing for radio group as well. Please check the image above from the dp.

app/client/src/widgets/RadioGroupWidget/component/index.tsx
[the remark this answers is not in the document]
ashit-rath: If components have specific styles and every component needs to push around the label to get it to its perfect position then the components are just not written correctly and we need to refactor every one of them to get it right. Imagine a UI library like material UI or ant design, when using their component you can add labels/tooltip/error messages without any extra configuration; they just work out of the box so we don't have to think about it. That is how things should be, continuous layers of abstraction. That being said, I think this refactor would be a bigger task as we need to verify each component; so we'll take this up in widget reskinning.
cc: @jsartisan

app/client/src/widgets/SwitchGroupWidget/component/index.tsx
```diff
@@ -1,143 +1,209 @@
...
     selected,
     valid,
+    width,
   } = props;
 
+  const [scrollable, setScrollable] = useState(false);
+
+  const containerRef = React.createRef<HTMLDivElement>();
+
+  useEffect(() => {
```
ashit-rath, line 116: Using useEffect with setState would cause unnecessary re-renders. Use `useMemo` 

```jsx
const scrollable = useMemo(() => {
    return (
      containerElement &&
      containerElement.scrollHeight > containerElement.clientHeight
    )
  }
), [
    height,
    width,
    inline,
    JSON.stringify(options),
    labelText,
    labelPosition,
    labelWidth,
  ])
```

app/client/src/widgets/SwitchGroupWidget/component/index.tsx
```diff
@@ -1,143 +1,209 @@
...
+          text={labelText}
+          width={labelWidth}
+        />
+      )}
+      <InputContainer
+        alignment={alignment}
+        compactMode={compactMode}
+        inline={inline}
+        optionCount={optionCount}
+        scrollable={scrollable}
```
ashit-rath, line 166: What is the purpose of `scrollable`?

app/client/src/components/ads/LabelWithTooltip.tsx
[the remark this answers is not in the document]
wmdev0808: It's just my way of deal with such a case.
Do you have better solution?

app/client/src/widgets/RichTextEditorWidget/component/index.tsx
```diff
@@ -66,37 +106,57 @@ export function RichtextEditorComponent(props: RichtextEditorComponentProps) {
...
     }
   };
+
   return (
     <StyledRTEditor
       className={`container-${props.widgetId}`}
+      compactMode={compactMode}
+      data-testid="rte-container"
       isValid={props.isValid}
+      labelPosition={labelPosition}
```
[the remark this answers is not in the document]
wmdev0808: Yes, I've extract the common styles as `labelLayoutStyles` css method

app/client/src/widgets/RadioGroupWidget/component/index.tsx
```diff
@@ -2,83 +2,216 @@ import React from "react";
...
-                ? Classes.SKELETON
-                : Classes.TEXT_OVERFLOW_ELLIPSIS
-            }
-          >
-            {this.props.label}
-          </Label>
+      <RadioGroupContainer
+        compactMode={compactMode}
+        data-testid="radiogroup-container"
+        labelPosition={labelPosition}
```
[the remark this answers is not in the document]
wmdev0808: The same answer as above

app/client/src/widgets/SwitchGroupWidget/component/index.tsx
wmdev0808: It is not required anymore

app/client/src/components/ads/LabelWithTooltip.tsx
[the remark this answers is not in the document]
ashit-rath: I think the problem I had is that, when I look at the input widget and see the style as 
```css
& .label-container {
 ...
}
```
I would need to find where this class is and which element I am changing.

I think here we can take an approach from blueprintjs where they export classNames as constants
So we can export this className as a constant and use it in the input component.

```js
// LabelWithTooltip.tsx
export const LABEL_CONTAINER_CLASS = "label-container"

<LabelContainer
      alignment={alignment}
      className={LABEL_CONTAINER_CLASS}

// Input component
& ${LABEL_CONTAINER_CLASS} {
 ...
}
```

ashit-rath (approved):

app/client/src/components/ads/LabelWithTooltip.tsx
```diff
@@ -0,0 +1,261 @@
...
+      {helpText && (
+        <Tooltip
+          content={helpText}
+          hoverOpenDelay={LABEL_TOOLTIP_OPEN_DELAY}
+          position={Position.TOP}
+        >
+          <ToolTipIcon
+            color={Colors.SILVER_CHALICE}
+            compact={compact}
+            height={14}
```
ashit-rath, line 248: Can we put this in a const like `TOOLTIP_ICON_SIZE` and keep it in this file as it only belongs to this.

app/client/src/components/constants.ts
```diff
@@ -87,3 +87,24 @@ export enum ButtonPlacementTypes {
...
+
+export enum LabelPosition {
+  Auto = "Auto",
+  Top = "Top",
+  Left = "Left",
+}
+/**
+ * Max width of the label, specified in percentage(%)
+ */
+export const LABEL_MAX_WIDTH_RATE = 70;
```
ashit-rath, line 99: I don't this these constants `LABEL_MAX_WIDTH_RATE` `LABEL_DEFAULT_GAP`  `LABEL_TOOLTIP_OPEN_DELAY` belong here as they won't be used by any component other than the Label. So let's move these to the LabelWithTooltip component.

Whereas the `LabelPosition` belongs to this file as it would be pulled by almost all the components.

Great job on adding comments.

app/client/src/components/ads/LabelWithTooltip.tsx
```diff
@@ -0,0 +1,261 @@
...
+            position={position}
+            width={14}
+          >
+            <HelpIcon className={helpTextClassName} />
```
ashit-rath, line 252: Can we rename this as `cyHelpTextClassName`?

dilippitchika:
@wmdev0808 for the top alignment of labels did we change something, I can now see a scroll bar by default for anything I top align in the release.app.appsmith.com and open it in the DP, attached is the image for reference

Release
Screenshot 2022-04-05 at 10 58 34 AM

DP
Screenshot 2022-04-05 at 10 58 44 AM

wmdev0808:
> @wmdev0808 for the top alignment of labels did we change something, I can now see a scroll bar by default for anything I top align in the release.app.appsmith.com and open it in the DP, attached is the image for reference
> 
> Release Screenshot 2022-04-05 at 10 58 34 AM
> 
> DP Screenshot 2022-04-05 at 10 58 44 AM

That is inevitable because we should keep consistency across affected widgets and the original ones are not normalized one

ashit-rath:
@wmdev0808 @dilippitchika If we are still under discussion for the scrolling issue, can we move this card to `Needs PRD/Design`?

dilippitchika:
@ashit-rath we will have to resolve this today. No point moving it there and back here again for this.

app/client/src/components/ads/LabelWithTooltip.tsx
```diff
@@ -0,0 +1,256 @@
...
+  ${({ alignment, compact, inline, optionCount, position, width }) => `
+    ${
+      position !== LabelPosition.Top &&
+      (position === LabelPosition.Left || compact)
+        ? `&&& {margin-right: 5px; flex-shrink: 0;} max-width: ${LABEL_MAX_WIDTH_RATE}%;`
+        : `width: 100%;`
+    }
+    ${position === LabelPosition.Left &&
+      `
+      ${!width && `width: 33%`};
```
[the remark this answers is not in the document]
ashit-rath: Nice catch! Paul please take of this.

app/client/src/widgets/MultiSelectTreeWidget/component/index.styled.tsx
```diff
@@ -929,3 +907,11 @@ export const inputIcon = (): JSX.Element => (
     />
   </svg>
 );
+
+export const InputContainer = styled.div<{
+  compactMode: boolean;
+  labelPosition?: LabelPosition;
+}>`
+  ${multiSelectInputContainerStyles}
+  ${({ labelPosition }) => labelPosition && "height: 32px"};
```
ashit-rath, line 916: `32px` is repeated across all the widgets. Please put it in the constants

app/client/src/widgets/MultiSelectWidgetV2/component/index.styled.tsx
```diff
@@ -574,36 +585,10 @@ export const inputIcon = (): JSX.Element => (
...
-      ? Colors.GREY_8
-      : "inherit"};
-  font-size: ${(props) =>
-    props.$labelTextSize ? TEXT_SIZES[props.$labelTextSize] : "14px"};
-  font-weight: ${(props) =>
-    props?.$labelStyle?.includes(FontStyleTypes.BOLD) ? "bold" : "normal"};
-  font-style: ${(props) =>
-    props?.$labelStyle?.includes(FontStyleTypes.ITALIC) ? "italic" : ""};
+  ${multiSelectInputContainerStyles}
+  ${({ labelPosition }) => labelPosition && "height: 32px"};
```
ashit-rath, line 593: Here as well

app/client/src/widgets/SelectWidget/component/index.styled.tsx
```diff
@@ -246,15 +219,20 @@ ${({ dropDownWidth, id }) => `
...
-
-  label.select-label {
-    margin-bottom: ${(props) => (props.compactMode ? "0px" : "5px")};
-    margin-right: ${(props) => (props.compactMode ? "10px" : "0px")};
+  ${labelLayoutStyles}
+  & .${LABEL_CONTAINER_CLASS} {
+    label {
+      ${({ labelPosition }) => {
+        if (!labelPosition) {
+          return "margin-bottom: 5px";
```
ashit-rath, line 232: this margin bottom value can also be put in the constants

Tooluloope: https://github.com/appsmithorg/appsmith/blob/1d27c951de17a830320841fb8f59fcc50a01e68d/app/client/src/widgets/WidgetUtils.ts#L78

Do help correct the comment on the constant

ashit-rath: This should be used but I think we should move this to the `constants.ts` file rather than the WidgetUtils.

ashit-rath (approved):

ashit-rath (dismissed):

YogeshJayaseelan:
@wmdev0808  Testing is in progress for the for and please find the status below 

**Status of Bugs that are linked to this PR:**

Issue 1 - #7064 - Testing in progress

Issue 2 - #9781 - **Issue still exists - Reopen** - https://loom.com/share/37068f042f5841858628d1d31718412e

Issue 3 - #7215 - **Fixed**

Issue #10119 - **Fixed**

Issue #9908 - **Fixed**

Issue #11844 - **Fixed**

Issue #12355 - **Verified in the issue #7064**

**New issues on Enhancement 7064**

1. When any of the widths of the widget is set to 0 it consumes the area of width which is equivalence to width 6 

https://loom.com/share/ed33ccf587614d888805f2367c86257e

2. By default the outer boundary breaks for the Switch group and Checkbox group widget 

https://www.loom.com/share/046b97c7615448be85ef08412cc826cb

3. When the label position is set to top the outer boundary shows UI discrepancies and in case if there are any widgets that are placed below overlaps 

https://www.loom.com/share/2a1dd7029a464454afe4b9dc4b8fdd9b

4. By default the label component is not separated for Switch group/ Checkbox Group and Radio group widget

https://www.loom.com/share/55f9eac969444f5e91b890853a5b50b0


**Note: We will update the status of bugs that are mentioned in the comments earlier on this PR...**

YogeshJayaseelan:
> * We need to fix item 2 by increasing the default width, we will not be fixing the rest
> * @YogeshJayaseelan did not understand new issue 4

@dilippitchika If you compare the label component will be shown separated for other widgets except 'Switch group,Radio group  and CheckboxGroup'

Screenshot 2022-04-12 at 6 26 05 PM

YogeshJayaseelan:
@wmdev0808 @dilippitchika Tested this PR and please find the status below,

Status of Bugs that are linked to this PR:

Issue 1 - https://github.com/appsmithorg/appsmith/issues/7064 - **Fixed**

Issue 2 - https://github.com/appsmithorg/appsmith/issues/9781 - **Will not be Fixed**

Issue 3 - https://github.com/appsmithorg/appsmith/issues/7215 - **Fixed**

Issue https://github.com/appsmithorg/appsmith/issues/10119 - **Fixed**

Issue https://github.com/appsmithorg/appsmith/issues/9908 - **Fixed**

Issue https://github.com/appsmithorg/appsmith/issues/11844 - **Fixed**

Issue https://github.com/appsmithorg/appsmith/issues/12355 - **Verified in the issue https://github.com/appsmithorg/appsmith/issues/7064**

**Status of New issues raised on 12/04/2022**

Issue 1 - When any of the widths of the widget is set to 0 it consumes the area of width which is equivalence to width 6 - **Will not be fixed**

Issue 2 - By default the outer boundary breaks for the Switch group and Checkbox group widget - **Fixed**

Issue 3 - When the label position is set to top the outer boundary shows UI discrepancies and in case if there are any widgets that are placed below overlaps - **Will not be Fixed**

Issue 4 - By default the label component is not separated for Switch group/ Checkbox Group and Radio group widget - **This is not an issue** 

**New issues Dated 13/04/2022:**

Issue 1 - The widget value fonts slightly differ from other fonts in the property pane - Slightly bigger and bold 

Screenshot 2022-04-13 at 2 27 41 PM

Question 1: what is the purpose of field Auto under position? will it show the complete label text(70%) and the widget to (30%)

dilippitchika:
Auto is for maintaining the release like experience, if you resize the widget the label will move to top/left @YogeshJayaseelan

YogeshJayaseelan:
@wmdev0808 Tested this PR and it is working as expected. We can proceed further

somangshu (approved):
After the last approval, I found 2 small changes that look acceptable. In the essence of time, moving ahead here. @ashit-rath @Tooluloope if you see any concerns when you get back, Please let us know urgently, and we shall look into it

dilippitchika:
@YogeshJayaseelan @wmdev0808 this issue is still happening. How was this tested?
Issue 2 - By default the outer boundary breaks for the Switch group and Checkbox group widget - Fixed

Loom - https://www.loom.com/share/c030793964954a96ab72f2e8cbbfb721

YogeshJayaseelan:
> @YogeshJayaseelan @wmdev0808 this issue is still happening. How was this tested? Issue 2 - By default the outer boundary breaks for the Switch group and Checkbox group widget - Fixed
> 
> Loom - [LOOM DEMO](https://www.loom.com/share/c030793964954a96ab72f2e8cbbfb721)

@dilippitchika @wmdev0808 I did not experience the above problem during the second iteration of testing. Below is the screenshot which highlights the default state of Switch and Checkbox Group 

Screenshot 2022-04-15 at 11 40 43 AM

wmdev0808:
@dilippitchika , I can also reproduce the issue

YogeshJayaseelan:
@dilippitchika @wmdev0808 I also verified in release and I'm not able to reproduce. Any specific steps to reproduce?

https://loom.com/share/d1e194ba8ea241aea1ad67eab1eeadc4

dilippitchika:
@YogeshJayaseelan i guess you are using a monitor, you need to check in a smaller screen. Either resize the canvas width to tablet or try using a smaller screen

dilippitchika:
@wmdev0808 the video i shared is from release. I am using a monitor with 1366x768 resolution.

dilippitchika:
Satyajit and Abhishek also saw this issue in release - Please check the conversation here https://theappsmith.slack.com/archives/C02TWLJ26PM/p1650003769229109?thread_ts=1649920090.459319&cid=C02TWLJ26PM

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/FormWidgets/Inputv2_spec.js
```diff
@@ -22,7 +22,7 @@ describe("Input widget V2 - ", () => {
       ".t--property-control-onsubmit .t--open-dropdown-Select-Action",
     ).click();
     cy.selectShowMsg();
-    cy.addSuccessMessage("Submitted!!");
+    cy.addSuccessMessage("Submitted!!", ".t--property-control-onsubmit");
```
Rishabh-Rathod, line 25: This leads to adding `"Submitted!!"` to the label instead of the message field of the action creator.
Though the test passed this was not expected as now the message field remains empty and onSubmit shows an error.

Rishabh-Rathod: Also, I see this PR was huge and still wasn't reviewed by developers. ( looking at PR approvals )

Can I understand why we didn’t get this PR reviewed by multiple devs? 
( i don't mean that the PR code has issues, trying to understand the process here. )

cc: @somangshu

PR #10618, opened 2022-01-25
fix: Table Filter pane overflows when Table is at the top of the canvas

### LINKED ISSUE #7996: [Bug] App header overlaps the Table filter pane when table is placed at top of the canvas
Screenshot 2021-09-30 at 4 31 04 PM

yaldram:
## Description
- When the Table is placed at the top of the Canvas the Filterplane overflows / gets out of the canvas. Please check the issue below.

Fixes #7996

## Type of change

## How Has This Been Tested?
- Tested it Manually on the Canvas.
- Tested it Manually by deploying the app.

## Checklist:

    // Code coverage diff between base branch:release and head branch: fix/7996-table-filter-pane-overflow 
 :red_circle: | total | 55.7 **(0)** | 37.04 **(-0.01)** | 35.75 **(0)** | 56.05 **(0)**
 :green_circle: | app/client/src/pages/Editor/Popper.tsx | 68.18 **(0.74)** | 47.62 **(0.12)** | 50 **(0)** | 68.18 **(0.74)**
 :red_circle: | app/client/src/pages/Editor/utils.ts | 21.74 **(-0.19)** | 14.12 **(-1.07)** | 40 **(0)** | 19.81 **(-0.19)**
 :red_circle: | app/client/src/widgets/TableWidget/component/TableFilterPane.tsx | 59.52 **(-1.46)** | 18.18 **(-4.04)** | 12.5 **(0)** | 60 **(-1.54)**</details>


app/client/src/widgets/TableWidget/component/TableFilterPane.tsx
```diff
@@ -75,8 +75,20 @@ class TableFilterPane extends Component<Props> {
...
         generateClassName(this.props.tableFilterPane.widgetId);
       const el = document.getElementsByClassName(className)[0];
+
+      /*
+        Prevent the FilterPane from overflowing the canvas when the 
+        table widget is on the very top.
+        1.  document.getElementById("canvas-0") - edit mode parent
+        2. document.querySelector('[type="CANVAS_WIDGET"]') - preview mode parent
+      */
+      const boundaryParent =
```
marks0351, line 85: can u pls use the MAIN_CONTAINER_WIDGET_ID constant instead of using 0  to provide more context and readability.

app/client/src/widgets/TableWidget/component/TableFilterPane.tsx
```diff
@@ -75,8 +75,20 @@ class TableFilterPane extends Component<Props> {
...
+        document.getElementById("canvas-0") ||
+        document.querySelector('[type="CANVAS_WIDGET"]');
```
marks0351, line 87: I am guessing that this will always return the main canvas coz the first element of type canvas_widget is the main container.

app/client/src/widgets/TableWidget/component/TableFilterPane.tsx
```diff
@@ -75,8 +75,20 @@ class TableFilterPane extends Component<Props> {
...
+
       return (
         <Popper
+          boundaryParent={boundaryParent || "viewport"}
```
marks0351, line 91: Not a big comment: no need to send viewport coz i see that you have handled that path inside the popper.

marks0351 (changes requested):

app/client/src/widgets/TableWidget/component/TableFilterPane.tsx
yaldram: Thanks, `canvas_widget` can be used as the container for both edit and deployed apps.

app/client/src/widgets/TableWidget/component/TableFilterPane.tsx
yaldram: We don't need this anymore : - 
```javascript
const boundaryParent = document.querySelector('[type="CANVAS_WIDGET"]');
```
Do you want me to move `type="CANVAS_WIDGET"` to a constant ?

app/client/src/widgets/TableWidget/component/TableFilterPane.tsx
yaldram: Actually `boundaries` element requires a type - `Element | PopperJS.Boundary`. Before our fix we were passing `viewport` as a value. Now `document.querySelector()` can also return undefined so we handled that case here `boundaryParent={boundaryParent || "viewport"`. If you are using Propper anywhere else you don't need to pass the `boundaryParent` prop we have given it a default value of "viewport" in the Popper.tsx file. Thanks

app/client/src/widgets/TableWidget/component/TableFilterPane.tsx
marks0351: this is fine

marks0351 (changes requested):
Changes look good. can u pls add cypress test cases if possible.

yaldram:
I am new to cypress, I did follow this example for testing drag and drop scenarios - https://github.com/cypress-io/cypress-example-recipes/tree/master/examples/testing-dom__drag-drop/cypress/integration. But I am clueless about how should I test the `Filterpane overflow`. I did add - 
```javascript
cy.get(publish.tableFilterPopup).then((element) => {
      Cypress.dom.isVisible(element);
 });
```
But with the above we are just making sure that the widget isVisible which is true even if it goes under the Pages Tab. I think you can add better test cases @marks0351. Thanks.

app/client/src/pages/Editor/utils.ts
```diff
@@ -45,8 +45,8 @@ export const draggableElement = (
     if (calculatedLeft <= 0) {
       calculatedLeft = 0;
     }
-    if (calculatedTop <= 30) {
-      calculatedTop = 30;
+    if (calculatedTop <= 70) {
```
marks0351, line 48: this wont work when there is no page tab, try creating an app with just one page. we need to get this value in real time based on the popper's parent top offset wrt the window.

marks0351 (changes requested):

app/client/src/pages/Editor/utils.ts
yaldram: Yes, It will push the popper a bit more. In align with your suggestion I got this working - 
```javascript
const top =document.querySelector('[type="CANVAS_WIDGET"]')?.getBoundingClientRect() .top || 70;

if (calculatedTop <= top) {
  calculatedTop = top;
}
```
Just like the Popper.tsx I will send the boundaryParent as a parameter to the `draggableElement` function in utils.tsx.

app/client/src/pages/Editor/utils.ts
```diff
@@ -36,28 +36,41 @@ export const draggableElement = (
...
     calculatedLeft: number,
     calculatedTop: number,
   ) => {
     const bottomBarOffset = 34;
 
+    /*
+      Default to 70 for a save offset that can also
+      handle the pagination Bar.
+    */
+    const canvasTopOffset =
```
marks0351, line 50: can u get the parent as an input parameter.

marks0351 (changes requested):

app/client/src/pages/Editor/utils.ts
yaldram: Yes, I have added another prop to the Popper component, I thought of using `boundaryParent`, but it is of type - `boundaryParent?: Element | PopperJS.Boundary;` and our parent is of type `Element | null`, therefore I had to add another prop name `parentElement`. Also I have made this prop as optional, so that if you popper is not draggable no need to pass this prop.

marks0351 (approved):

YogeshJayaseelan:
@yaldram Tested this PR and it is working as expected

PR #10620, opened 2022-01-25
feat: Add reference doc url for error responses for git APIs

abhvsn:
## Description

> Earlier we missed adding the ref doc for pull conflicts and push error where we have upstream changes which needs to be pulled in local to sync local and remote.

## Type of change

## How Has This Been Tested?

> Manual test

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/GitFileUtils.java
```diff
@@ -158,7 +158,7 @@
     public Mono<ApplicationJson> reconstructApplicationFromGitRepo(String organisationId,
                                                              String defaultApplicationId,
                                                              String repoName,
-                                                             String branchName) throws GitAPIException, IOException {
+                                                             String branchName) {
```
AnaghHegde, line 162: Any specific reason on why we are removing this now ?

AnaghHegde (commented):
The PR name and the change set does not match. Can you rename the PR? Most of the changes in the PR are improving the error handling messages.

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/GitFileUtils.java
abhvsn: This is not thrown at all so though of removing this. Earlier when we had this method as synchronous at that time this made sense

AnaghHegde (approved):

PR #10638, opened 2022-01-26
fix: Bugfix for curl import in the child branch when application is connected to git

### LINKED ISSUE #10637: [Bug]: Curl import fails for the sub-branch application

### Current Behavior

When user connects the app to git and tries to import the curl this works fine for default branched application but fails when tries to import the same curl in sub-branch app.

### Steps To Reproduce

1. Connect app to git
2. You will get the default branched application
3. Import the curl works as expected
4. Create a new branch from the base/default branch
5. Try to import the curl

### Environment

Release

### Version

Cloud

abhvsn:
## Description

> When the application is connected to git and user tries to import a curl it's working fine for default branched application but fails if user creates a sub-branch from the default branch and then tries to import the curl.

Fixes #10637

## Type of change

## How Has This Been Tested?

> Junit TC
> Manual test

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/CurlImporterServiceCE.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/CurlImporterServiceCE.java
@@ -7,7 +7,7 @@ import java.util.List;
 
 public interface CurlImporterServiceCE extends ApiImporterCE{
 
-    ActionDTO curlToAction(String command, String pageId, String name) throws AppsmithException;
+    ActionDTO curlToAction(String command, String name) throws AppsmithException;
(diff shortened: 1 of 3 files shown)


AnaghHegde:
@abhvsn Can you please elaborate on why we are removing the `id` and just using the page name?

AnaghHegde (approved):

abhvsn:
> @abhvsn Can you please elaborate on why we are removing the `id` and just using the page name?

We are not removing the `pageId` but setting this after we have the branched page id from where the request has been made

AnaghHegde:
So the page details are stored underCollection which is what is used depending on the view mode. Is that correct assumption ?

Issue #10650, opened 2022-01-26
[Feature]: Only commit unpublished version for resources in git


abhvsn:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

Currently to be consistent with the import-export service we are pushing the unpublished as well as published versions to git for all the resources. We can avoid pushing published versions for the following resources with the expectation that both will have the same data.

### Why should this be worked on?

With the mentioned change user will only have to resolve merge conflicts, for the unpublished version for the particular resource.

PR #10776, opened 2022-01-31
chore: Commit only unpublished resources to git repo which includes queries, JSObjects, pages to avoid data duplication

abhvsn:
## Description

> As we are publishing the application and then making a commit to git we expect the published and unpublished versions for different resources like page, action, actionCollection etc will be the same, so We can avoid pushing published versions for the these resources with the expectation that both will have the same data. This will avoid data duplication in git repos which will improve the merge conflict resolution experience as we have to resolve merge conflicts only a single place.

Fixes https://github.com/appsmithorg/appsmith/issues/10650

## Type of change

## How Has This Been Tested?

> Manual Test
> TC (WIP)

## Checklist:


AnaghHegde (approved):
LGTM

PR #10810, opened 2022-02-02
fix: Delete the pages post merge operation from application when they are removed from file system

### LINKED ISSUE #10773: [Bug]: When user delete a resource form the child branch and merge it back to parent branch, still the deleted resource will show up in the newly created branch

### Current Behavior

When we try to delete the resource by using the git branching, we can see the resource if we create the new branch from the parent branch even after we merged the sub-branch where the resource was deleted.

### Steps To Reproduce

Let's take an example with page as a resource:
1. We have default branch main which have 2 pages: page1 and page2
2. User creates F1 branch and deletes page2
3. Merge F1 -> master
4. Page2 will get deleted from master as expected
5. Create F2 branch from master, you will see Page2 in F2 (not expected)

### Environment

Release

### Version

Cloud

AnaghHegde:
## Description

> When we try to delete the resource by using the git branching, we can see the resource if we create the new branch from the parent branch even after we merged the sub-branch where the resource was deleted.

Fixes #10773

## Type of change

## How Has This Been Tested?

- Locally
- JUnit Tests

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -759,7 +768,31 @@
                                 return applicationPages;
                             })
                             .then()
-                            .thenReturn(applicationPages);
+                            .thenReturn(Mono.empty())
```
abhvsn, line 771: Why do we need this `thenReturn`?

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -759,7 +768,31 @@
...
+                                           for (NewPage newPage : existingPagesList) {
+                                               if(!validPageIds.contains(newPage.getId())) {
+                                                   inValidIds.add(newPage.getId());
+                                               }
+                                           }
+
+                                           // Delete the pages which were removed during git merge operation
+                                           // This does not apply to the traditional import via file approach
+                                           return Flux.fromIterable(inValidIds)
+                                                   .flatMap(newPageService::archiveById)
```
abhvsn, line 791: We need to delete all the actions and JS object attached to this page. Can we use `deleteUnpublishedPage` instead of `archive` method

abhvsn (changes requested):
Added minor comments for improvements

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -759,7 +768,31 @@
                                 return applicationPages;
                             })
                             .then()
-                            .thenReturn(applicationPages);
+                            .thenReturn(Mono.empty())
+                            .flatMap(unused -> {
```
abhvsn, line 772: Should we avoid executing this block for traditional import-export as this is valid only for git-sync as you commented below?
FYI `applicationId` will be null for traditional import-export.

AnaghHegde:
@abhvsn ready for review :)

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
abhvsn: When I meant `applicationId` will not be null for git-sync I was referring to the one passed in for the global method 
```
 public Mono<Application> importApplicationInOrganization(String organizationId,
                                                             ApplicationJson applicationJson,
                                                             String applicationId,
                                                             String branchName)
```

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
AnaghHegde: Same reason as it was used before adding this logic. We had `.then` & `.thenReturn` so that the next block executes post this Flux.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -771,7 +771,7 @@
                             .then()
                             .thenReturn(applicationPages)
                             .flatMap(unused -> {
-                                if (!Optional.ofNullable(importedApplication.getId()).isEmpty()) {
+                                if (!Optional.ofNullable(applicationId).isEmpty()) {
```
abhvsn, line 774: Hey also why are we not using the `StringUtils.isEmpty(applicationId)` for checking the availability of `applicationId` as it has a String type instead of more generic `Optional.ofNullable`

app/server/appsmith-server/src/test/java/com/appsmith/server/solutions/ImportExportApplicationServiceTests.java
```diff
@@ -1173,11 +1202,16 @@ public void importApplicationIntoOrganization_pageAddedInBranchApplication_Succe
         StepVerifier
                 .create(pageList)
                 .assertNext(newPages -> {
+                    // Check before import we had both the pages
+                    assertThat(applicationPageIdsBeforeImport).hasSize(2);
                     assertThat(newPages.size()).isEqualTo(3);
                     List<String> pageNames = newPages.stream().map(newPage -> newPage.getUnpublishedPage().getName()).collect(Collectors.toList());
                     assertThat(pageNames.contains("Page1"));
                     assertThat(pageNames.contains("Page2"));
                     assertThat(pageNames.contains("Page3"));
```
abhvsn, line 1211: We are not checking any condition here this should be 
```
assertThat(pageNames).contains("Page3");
or
assertThat(pageNames.contains("Page3")).isTrue();
```

abhvsn (approved):

PR #10858, opened 2022-02-03
fix: Fix NPE during import when page ids are missing for published action and actionCollection 

abhvsn:
## Description

> When the application is imported from the JSON file with null pageId for action and actionCollection, internal server error is thrown. This PR will use the unpublished resource pageId as a fallback when the published one don't have one for actions and/or action collection and then import the application.
 
Fixes https://github.com/appsmithorg/appsmith/issues/10857

## Type of change

## How Has This Been Tested?

> Junit TC
> Manual testing

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/test/resources/test_assets/ImportExportServiceTest/valid-application.json
+++ b/app/server/appsmith-server/src/test/resources/test_assets/ImportExportServiceTest/valid-application.json
@@ -517,11 +517,42 @@
         "userPermissions": []
       },
       "publishedAction": {
+        "name": "api_wo_auth",
         "datasource": {
+          "id": "api_ds_wo_auth",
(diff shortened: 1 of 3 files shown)


app/server/appsmith-server/src/test/resources/test_assets/ImportExportServiceTest/valid-application.json
```diff
@@ -517,11 +517,42 @@
         "userPermissions": []
       },
       "publishedAction": {
```
abhvsn, line 519: `pageId` is absent in the publishedAction. This was causing the NPE earlier. We are now using the pageId from the unpublishedVersion to fix this

AnaghHegde (approved):

PR #10862, opened 2022-02-03
fix: Added JSON migration for UQI actions conversion to support form/json mode

nidhi-nair:


nidhi-nair:
Closing in favour of #11534

PR #10869, opened 2022-02-03
chore: Add brotli compression

### LINKED ISSUE #7307: [Task] Add brotli compression
As appsmith's current frontend build chuck size is more than an mb. Compression methods like brotli can help to reduce load time for users.

Refer - https://tech.groww.in/enable-brotli-compression-in-webpack-with-fallback-to-gzip-397a57cf9fc6

- [x] Add compression in the build process
- [ ] #10889

Rishabh-Rathod:
## Description

Added brotli compression in build step.

Screenshot 2022-02-04 at 3 07 59 AM

Fixes #7307

## Type of change

Enhancement

## How Has This Been Tested?

Manually in DP

## Checklist:

    // Code coverage diff between base branch:release and head branch: test-brotli-perf 
 :red_circle: | total | 54.94 **(0)** | 36.4 **(-0.01)** | 34.61 **(0)** | 55.35 **(0)**


riodeuno (approved):
Nice initiative @Rishabh-Rathod ! This should save us some bandwidth.
Please sync up with the backend team to see if we need to update headers when serving these files via nginx

marks0351:
@Rishabh-Rathod @riodeuno it says this compression doesn't support http, will that be a problem?

riodeuno:
@marks0351 I believe we're falling back to gzip if the browser doesn't support it. @Rishabh-Rathod Please confirm.

Rishabh-Rathod:
> @marks0351 I believe we're falling back to gzip if the browser doesn't support it. @Rishabh-Rathod Please confirm.

From the build process step, we are creating both brotli and gzip compressed chunks and according to the browser's compatibility, the server would respond by looking at request headers.

mohanarpit:
@Rishabh-Rathod Really good initiative! 

If I understand this correctly, the build step now creates 2 copies of the files; gzip version & brotli version. This means that our build size will increase effectively double in size. But this will help us when serving the data via nginx. Am I understanding this correctly?

Also, if I'm not mistaken, we'll have to make changes to the Nginx Docker image that serves this data in order to take advantage of the brotli compressed files.

Rishabh-Rathod:
> If I understand this correctly, the build step now creates 2 copies of the files; gzip version & brotli version. This means that our build size will increase effectively double in size. But this will help us when serving the data via nginx. Am I understanding this correctly?

@mohanarpit 
Yes, it creates js, gzip, and brotli versions of all chunks which will lead to total storage of build files size increase by 25% approx. As there is a considerable difference between the size of normal js files vs brotli file size for bigger files.

> Also, if I'm not mistaken, we'll have to make changes to the Nginx Docker image that serves this data in order to take advantage of the brotli compressed files.

Yes, had a discussion with @sharat87 regarding the same. The current way we support gzip is via NGINX and adding brotli is not simple in that route.

Maybe the compression part should be handled by webpack. I don't have much knowledge on this part though but yes without that I assume we will have those files but won't be serving them to users.

PR #10886, opened 2022-02-04
fix: limit validation errors in debugger

### LINKED ISSUE #10938: [Perf]: Too many errors slows debugger rendering

### Current Behavior

When we have large, improperly formatted, data being validated; large number of error messages are rendered in the debugger.
For example, if we're trying to bind an array with 20000 entries (objects) in a chart widget, and these entries do not have x and y properties, validations return 20000 error messages (one for each entry). When the debugger tries to render these many messages, the application freezes for a few seconds.

### Steps To Reproduce

1. Create an API which results in 50K+ entries
2. Bind the result as is (without matching the expected value) in a chart widget's data property

In the edit mode, the page slows down drastically, as it is trying to render 50K errors in the debugger.

### Environment

Production

### Version

Cloud

riodeuno:
## Description

When we have large, improperly formatted, data being validated; large number of error messages are rendered in the debugger.
For example, if we're trying to bind an array with 20000 entries (objects) in a chart widget, and these entries do not have `x` and `y` properties, validations return 20000 error messages (one for each entry). When the debugger tries to render these many messages, the application freezes for a few seconds. 

The solution here, is to bail out of validations early **_IF_** we have already encountered a fixed number (in this case 10) of validation errors and show these messages in the debugger.

Does it negatively effect DX?
Fixing the issues shown so far will trigger another validation. As a result,
- If there are more validation errors, they will be displayed. 
- If there aren't any more validation errors, the fix added earlier has fixed all validation errors.

Also, this only effects individual property validations. Each property with errors in validation, will have their errors displayed nevertheless.

There will not be any lack of relevant information displayed to an Appsmith developer. 
Hence, there shouldn't be any negative effects to the DX.

Fixes #10938

## Type of change

## How Has This Been Tested?

Unit tests added and some updated

## Checklist:

    // Code coverage diff between base branch:release and head branch: fix/limit-validation-errors 
 :green_circle: | total | 55.03 **(0.04)** | 36.36 **(0.04)** | 34.72 **(0.04)** | 55.41 **(0.04)**
 :green_circle: | app/client/src/constants/messages.ts | 75 **(0.17)** | 100 **(0)** | 25.9 **(0.49)** | 80.47 **(0.16)**
 :red_circle: | app/client/src/utils/WorkerUtil.ts | 88.98 **(-0.78)** | 68.63 **(-3.92)** | 100 **(0)** | 92.38 **(-0.95)**
 :sparkles: :new: | **app/client/src/workers/helpers.ts** | **100** | **100** | **100** | **100**
 :green_circle: | app/client/src/workers/validations.ts | 82.01 **(5.22)** | 71.23 **(1.74)** | 76.67 **(9.02)** | 83.44 **(4.85)**</details>

### DIFF
--- a/app/client/src/workers/validations.ts
+++ b/app/client/src/workers/validations.ts
@@ -25,6 +25,12 @@ import evaluate from "./evaluate";
 import getIsSafeURL from "utils/validation/getIsSafeURL";
 import * as log from "loglevel";
 import { QueryActionConfig } from "entities/Action";
+import {
+  VALIDATION_ARRAY_DISALLOWED_VALUE,
+  VALIDATION_ARRAY_DUPLICATE_PROPERTY_VALUES,
+  VALIDATION_ARRAY_INVALID_ENTRY,
+  VALIDATION_ARRAY_UNIQUE,
+} from "constants/messages";
 export const UNDEFINED_VALIDATION = "UNDEFINED_VALIDATION";
 export const VALIDATION_ERROR_COUNT_THRESHOLD = 10;
 
@@ -145,7 +151,7 @@ function validateArray(
 
   // Should we validate against disallowed values in the value array?
   const shouldVerifyAllowedValues =
-    Array.isArray(allowedValues) && allowedValues.length;
+    Array.isArray(allowedValues) && !!allowedValues.length;
 
   // Do we have validation config for array children?
   const shouldValidateChildren = !!childrenValidationConfig;
@@ -153,8 +159,7 @@ function validateArray(
   // Should array values be unique? This should applies only to primitive values in array children
   // If we have to validate children with their own validation config, this should be false (Needs verification)
   // If this option is true, shouldArrayValuesHaveUniqueValuesForKeys will become false
-  const shouldArrayHaveUniqueEntries =
-    config.params?.unique === true && !shouldValidateChildren;
+  const shouldArrayHaveUniqueEntries = config.params?.unique === true;
 
   // Should we validate for unique values for properties in the array entries?
   const shouldArrayValuesHaveUniqueValuesForKeys =
@@ -171,7 +176,7 @@ function validateArray(
       return {
         isValid: false,
         parsed: config.params?.default || [],
-        messages: [`Array must be unique. Duplicate values found`],
+        messages: [VALIDATION_ARRAY_UNIQUE()],
       };
     }
   }
@@ -194,7 +199,7 @@ function validateArray(
         isValid: false,
         parsed: config.params?.default || [],
         messages: [
-          `Duplicate values found for the following properties, in the array entries, that must be unique -- ${uniqueKeys.join(
+          `${VALIDATION_ARRAY_DUPLICATE_PROPERTY_VALUES()} ${uniqueKeys.join(
             ",",
           )}.`,
         ],
@@ -206,7 +211,7 @@ function validateArray(
   value.every((entry, index) => {
     // Validate for allowed values
     if (shouldVerifyAllowedValues && !(allowedValues || []).includes(entry)) {
-      _messages.push(`Disallowed value: ${entry}`);
+      _messages.push(`${VALIDATION_ARRAY_DISALLOWED_VALUE()}: ${entry}`);
       _isValid = false;
     }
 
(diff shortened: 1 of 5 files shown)


app/client/src/workers/validations.ts
```diff
@@ -218,11 +248,13 @@ export const validate = (
   value: unknown,
   props: Record<string, unknown>,
 ): ValidationResponse => {
+  const start = performance.now();
```
riodeuno, line 251: @riodeuno Remove this line.

app/client/src/workers/validations.ts
```diff
@@ -106,105 +108,133 @@ function validateArray(
...
-      }
-    });
+    if (uniqueEntries.length !== value.length) {
+      // Bail out early
+      // Because, we don't want to re-iterate, if this validation fails
+      return {
+        isValid: false,
+        parsed: config.params?.default || [],
+        messages: [
+          `Duplicate values found for the following properties, in the array entries, that must be unique -- ${uniqueKeys.join(
```
riodeuno, line 197: @riodeuno Move these messages to `messages.ts`

app/client/src/workers/validations.ts
```diff
@@ -106,105 +108,133 @@ function validateArray(
...
     config.params?.children?.type === ValidationTypes.OBJECT &&
-    (config.params.children.params?.allowedKeys || []).length > 0
+    Array.isArray(allowedKeyConfigs) &&
+    allowedKeyConfigs.length
   ) {
-    const allowedKeysCofigArray =
-      config.params.children.params?.allowedKeys || [];
+    uniqueKeys = compact(
+      allowedKeyConfigs.map((allowedKeyConfig) => {
+        // TODO(abhinav): This is concerning, we now have two ways,
```
SatishGandham, line 127: What is the other way to pass unique keys?

app/client/src/workers/validations.ts
```diff
@@ -106,105 +114,134 @@ function validateArray(
...
+
+  // Should we validate for unique values for properties in the array entries?
+  const shouldArrayValuesHaveUniqueValuesForKeys =
+    !!uniqueKeys.length && !shouldArrayHaveUniqueEntries;
+
+  // Verify if all values are unique
+  if (shouldArrayHaveUniqueEntries) {
+    // Loop
+    if (
+      uniq(value.map((entry) => JSON.stringify(entry))).length !== value.length
```
aswathkk, line 172: If we replace `_.uniq` with our own implementation using loops, we could detect duplicate value and return error early rather than comparing the unique's length with `value.length`.
@riodeuno Do you think we will be gaining significant performance if we do so?

app/client/src/workers/validations.ts
```diff
@@ -106,105 +114,134 @@ function validateArray(
...
-        };
-      }
-    }
-  }
-  // Check uniqueness of object elements in an array
+  let _isValid = true; // Let's first assume that this is valid
+  const _messages: string[] = []; // Initialise messages array
+
+  // Values allowed in the array
+  const allowedValues = config.params?.allowedValues;
```
aswathkk, line 121: We could make this a javascript `Set` so that we could replace O(n) `Array.includes` with O(1) `Set.has`.

@riodeuno what do you think?

riodeuno: Yes. 
Although, we will have to iterate the array to generate the Set, once. 
I believe that is how the  `new Set( array )` works internally. Let me know if there is a more performant way.

Then we call `array.includes` multiple times, where we might get performance improvements by using `set.has`. It might be really small, but I'm willing to take that half a millisecond.

I'll update the code.

app/client/src/workers/validations.ts
SatishGandham: Good point @aswathkk 
Not a significant performance gain, but will be notable with large data.
We won't know when JSON.stringify will comeback to bite us, any way to reduce its calls will be good.

If we are not going with our implemetaion, we should use `Set`, it is twice as fast.
https://www.measurethat.net/Benchmarks/Show/3889/0/lodash-uniq-vs-set

riodeuno: Sounds good. 
If we go ahead with Set, do you recommend we convert the value into a set at the beginning of the validation and convert it back into an array for use in the widgets, or use Sets only for this particular uniq validation? We have seen data with 50K entries.

riodeuno: So, I'm going ahead with Satish's suggestion of using Set instead of _.uniq and stringify. 

@aswathkk I agree that it would be helpful to identify and inform an Appsmith developer regarding the index of the duplicate entry.

I am assuming a standard scenario of a happy path, where the validation passes, in this case the Set approach seems more performant than the loop approach.
Based on this, let's take a call in the interest of performance and circle back when the need to identify and inform arises.

Does that sound good?

SatishGandham: @riodeuno ,
Even with set approach, it has to go through the whole array in happy path.
It would be nice if we can fail early, if we are not doing it put a note in the code about possible improvement.

riodeuno: Yeah, one iteration is inevitable. Given the discussion, we seem to conclude that converting a large array into a Set, would be faster than iterating through a standard array (which is the type of the original input to the validations), in the happy path.

I find getting that performance improvement more valuable than the DX improvement, in this scenario.

I will add a comment.

riodeuno: @SatishGandham Just confirmed this. Seems like a major difference in performance.

https://www.measurethat.net/Benchmarks/Show/17083/0/create-set-vs-loop

aswathkk (approved):

PR #10895, opened 2022-02-04
fix: handle empty column name in UQI sort component

### LINKED ISSUE #10896: [Bug]: handle empty column name in UQI sort component

### Current Behavior

The sort component introduced as part of UQI currently fails if any of the column names are empty. This is a problem because if a sort component is added to a query editor form then at least one sort component with an empty column gets added by default, which can make the query fail.

### Steps To Reproduce

On release:
- create S3 list query
- don't edit sort component
- execute query will show failure

### Environment

Release

### Version

Cloud

sumitsum:
## Description
- The sort component introduced as part of UQI currently fails if any of the column names are empty. This is a problem because if a sort component is added to a query editor form then at least one sort component with an empty column gets added by default, which can make the query fail.
- This PR adds changes to ignore all the empty column names (or ignore the sort conditon if all column names are empty). 

Fixes #10896

## Type of change

## How Has This Been Tested?
- JUnit TC
- Manual

## Checklist:

### DIFF
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/ce/FilterDataServiceCE.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/ce/FilterDataServiceCE.java
@@ -310,12 +310,19 @@ public class FilterDataServiceCE implements IFilterDataServiceCE {
      * @throws AppsmithPluginException
      */
     private void addSortCondition(StringBuilder sb, List<Map<String, String>> sortBy) throws AppsmithPluginException {
-        if (CollectionUtils.isEmpty(sortBy)) {
+
+        /**
+         * Checks if:
+         *  o `sortBy` condition list is null or empty
+         *  o all column names in the sortBy list are empty
+         */
+        if (isSortConditionEmpty(sortBy)) {
             return;
         }
 
         sb.append(" ORDER BY");
         sortBy.stream()
+                .filter(sortCondition -> !isBlank(sortCondition.get(SORT_BY_COLUMN_NAME_KEY)))
                 .forEachOrdered(sortCondition -> {
                     String columnName = sortCondition.get(SORT_BY_COLUMN_NAME_KEY);
                     SortType sortType;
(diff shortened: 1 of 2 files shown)


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/ce/FilterDataServiceCE.java
```diff
@@ -310,12 +310,19 @@ private void addProjectionCondition(StringBuilder sb, List<String> projectionCol
...
+         *  o `sortBy` condition list is null or empty
+         *  o all column names in the sortBy list are empty
+         */
+        if (isSortConditionEmpty(sortBy)) {
             return;
         }
 
         sb.append(" ORDER BY");
         sortBy.stream()
+                .filter(sortCondition -> !isBlank(sortCondition.get(SORT_BY_COLUMN_NAME_KEY)))
```
nidhi-nair, line 325: Is this not redundant? If we get to this point we'll already have appended the `ORDER BY` clause.

sumitsum: @nidhi-nair , this is not redundant because at this point we are assured that at least one of the sort component is non-empty so we will process that sort componet while skipping all the other ones.

sumitsum: @nidhi-nair , this is not redundant because at this point we are assured that at least one of the sort component is non-empty so we will process that sort componet while skipping all the other ones.

sumitsum: @nidhi-nair , this is not redundant because at this point we are assured that at least one of the sort component is non-empty so we will process that sort componet while skipping all the other ones.

sumitsum: @nidhi-nair , this is not redundant because at this point we are assured that at least one of the sort component is non-empty so we will process that sort componet while skipping all the other ones.

nidhi-nair (approved):

PR #10933, opened 2022-02-07
fix: Sort unordered set elements before committing to git repo so as to have same serialised output string for identical set of elements

### LINKED ISSUE #10932: [Bug]: Convert unordered Set to ordered one before committing the file to repo

### Current Behavior

As the `layoutOnLoadAction` have `List<Set<DslActionDTO>>` datatype which includes the unordered set, this shows unnecessary file changes when we serialize the objects to the JSON files. 

### Steps To Reproduce

1. Connect application to git
2. Create multiple `onPageLoadAction`
3. Try changes on canvas like widget displacement, changing the size of a widget which have nothing to do with the binding 
4. Check the committed files on remote repo where we can observe the changes in `layoutOnLoadActions`

### Environment

Release

### Version

Cloud

abhvsn:
## Description

> As Set is an unordered datatype, this is causing unnecessary changes to the file system during the commit operation even if the user has not altered anything since the last commit.

Fixes #10932

## Type of change

## How Has This Been Tested?

> JUnit TC
> Manual test

## Checklist:

### DIFF
--- a/app/server/appsmith-git/src/main/java/com/appsmith/git/converters/GsonUnorderedToOrderedSetConverter.java
+++ b/app/server/appsmith-git/src/main/java/com/appsmith/git/converters/GsonUnorderedToOrderedSetConverter.java
@@ -1,0 +1,37 @@
+package com.appsmith.git.converters;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonSerializationContext;
+import com.google.gson.JsonSerializer;
+import org.springframework.util.CollectionUtils;
+
+import javax.lang.model.type.PrimitiveType;
+import java.lang.reflect.Type;
+import java.util.Collection;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class GsonUnorderedToOrderedSetConverter implements JsonSerializer<Set> {
+    @Override
+    public JsonArray serialize(Set src, Type typeOfSrc, JsonSerializationContext context) {
+        // Sort the set so that same elements will not end up in merge conflicts
+        return (JsonArray) new Gson().toJsonTree(getOrderedResource(src));
+    }
+
+    /**
+     * Sort the primitive datatype objects and string so that we will have predictable output
+     * @param objects set of objects which needs to be sorted before serialisation
+     * @param <T>
+     * @return sorted collection
+     */
+    private <T> Collection<T> getOrderedResource(Set<T> objects) {
+        if (!CollectionUtils.isEmpty(objects)) {
+            T element = objects.iterator().next();
+            if (element instanceof String || element instanceof PrimitiveType) {
+                return objects.stream().sorted().collect(Collectors.toList());
+            }
+        }
+        return objects;
+    }
+}
--- a/app/server/appsmith-git/src/main/java/com/appsmith/git/converters/GsonUnorderedToOrderedSetConverter.java
+++ b/app/server/appsmith-git/src/main/java/com/appsmith/git/converters/GsonUnorderedToOrderedSetConverter.java
@@ -20,7 +20,10 @@ public class GsonUnorderedToOrderedSetConverter implements JsonSerializer<Set> {
     }
 
     /**
-     * Sort the primitive datatype objects and string so that we will have predictable output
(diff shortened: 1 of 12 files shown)


app/server/appsmith-git/src/main/java/com/appsmith/git/converters/GsonUnorderedToOrderedSetConverter.java
```diff
@@ -0,0 +1,37 @@
...
+public class GsonUnorderedToOrderedSetConverter implements JsonSerializer<Set> {
+    @Override
+    public JsonArray serialize(Set src, Type typeOfSrc, JsonSerializationContext context) {
+        // Sort the set so that same elements will not end up in merge conflicts
+        return (JsonArray) new Gson().toJsonTree(getOrderedResource(src));
+    }
+
+    /**
+     * Sort the primitive datatype objects and string so that we will have predictable output
+     * @param objects set of objects which needs to be sorted before serialisation
```
AnaghHegde, line 24: I felt the comment is little confusing. Can we add an example in the end of the current line?

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CompareDslActionDTO.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Comparator;
+
+public class CompareDslActionDTO implements Comparator<DslActionDTO>, Serializable {
+    // Method to compare DslActionDTO based on id
+    @Override
+    public int compare(DslActionDTO action1, DslActionDTO action2)
+    {
+        return action1.getId().compareTo(action2.getId());
+    }
+}
```
AnaghHegde, line 15: Would it not work if we override the `comaprator` in DslActionDTO class ?

AnaghHegde, line 15: Would it not work if we override the `comaprator` in DslActionDTO class ?

AnaghHegde (commented):
Overall looks good to me other than the small nits. 
On a side thought, we are using the TreeSet for actions in a sorted manner. But being the TreeSet implementation asynchronous, wouldn't it cause any issues when there are multiple users trying to do git commit and status at the same time ?

abhvsn:
> Overall looks good to me other than the small nits. On a side thought, we are using the TreeSet for actions in a sorted manner. But being the TreeSet implementation asynchronous, wouldn't it cause any issues when there are multiple users trying to do git commit and status at the same time?

TreeSet does not have reactive asynchronous implementation but it's not thread-safe. But I suppose we are working with the single thread here

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CompareDslActionDTO.java
abhvsn: I tried this but was getting following error 
```
java.lang.ClassCastException: class com.appsmith.server.dtos.DslActionDTO cannot be cast to class java.lang.Comparable (com.appsmith.server.dtos.DslActionDTO is in unnamed module of loader 'app'; java.lang.Comparable is in module java.base of loader 'bootstrap')
```

abhvsn:
> Overall looks good to me other than the small nits. On a side thought, we are using the TreeSet for actions in a sorted manner. But being the TreeSet implementation asynchronous, wouldn't it cause any issues when there are multiple users trying to do git commit and status at the same time ?

TreeSet does not have reactive asynchronous implementation but it’s not thread-safe. But I suppose we are working with the single thread here, so it’s safe to assume this will work as expected. The other point you mentioned if multiple users make a call to commit, getStatus etc. `JGIT` handles the file locks but this can be improvised by taking control of file locks, nonetheless, this implementation doesn’t concern the above-mentioned issue.

AnaghHegde (approved):
LGTM

app/server/appsmith-git/src/main/java/com/appsmith/git/converters/GsonUnorderedToOrderedSetConverter.java
abhvsn: Provided a sample example for ease of understanding
understanding

PR #11053, opened 2022-02-09
chore: Add a new section in PR template for how-to apps

pranavkanade:
Adding this section to proactively create related community how-to apps for obvious cases.

@mohanarpit can we add a label to the PR automatically, if contributor has listed some scenarios?

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -28,6 +28,11 @@ Fixes # (issue)
 - Test A
 - Test B
 
+## Suggested How-to Apps
+
+> Please suggest couple ideas for how-to apps, to help us better explain the usability of this feature/bug to the community.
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -28,6 +28,11 @@ Fixes # (issue)
 - Test A
 - Test B
 
+## Suggested How-to Apps
```
mohanarpit, line 31: Sorry, I don't understand this addition. Can you help me understand how this will help us, our users or contributors?

pranavkanade: Yes, that is the idea. Creating a separate issue for maintaining a list of how to apps makes sense in case of Epics. But, seems overkill for incremental improvements. Any other suggestions?

PR #11058, opened 2022-02-09
fix: Increase supported header size for multipart data

### LINKED ISSUE #10660: DataBufferLimitException: Part headers exceeded the memory usage limit of 8192 bytes
Sentry Issue: [APPSMITH-1J4](https://sentry.io/organizations/appsmith/issues/2911835000/?referrer=github_integration)

```
DataBufferLimitException: Part headers exceeded the memory usage limit of 8192 bytes
    at org.springframework.http.codec.multipart.MultipartParser$HeadersState.onNext(MultipartParser.java:365)
```

abhvsn:
## Description

> Today we are using Spring's default multipart headers size of 8kB for file uploads, this works for most of the cases but lately, we are getting errors in sentry `DataBufferLimitException: Part headers exceeded the memory usage limit of 8192 bytes`, but we are not able to reproduce this on our end for import flow. If we try to import the application from incognito for the folks who are experiencing this behavior this works fine. This PR just increases the supported multipart header size from 8kB to 128kB, with the hope that this will minimize these issues and give us some headroom to log these outliers.   

Relevant discussion around this issue can be found at https://github.com/spring-projects/spring-boot/issues/25650

Fixes #10660 

## Type of change

## How Has This Been Tested?

> Junit TC

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationControllerCE.java
```diff
@@ -186,7 +186,10 @@ public ApplicationControllerCE(
                                                                     @PathVariable String orgId) {
         log.debug("Going to import application in organization with id: {}", orgId);
         return fileMono
-                .flatMap(file -> importExportApplicationService.extractFileAndSaveApplication(orgId, file))
+                .flatMap(file -> {
+                    log.debug("File part header for imported file: {}", file.headers());
```
trishaanand, line 190: QQ : Do we need this to be logged on our prod? If not, we can remove this.

app/server/appsmith-server/src/main/java/com/appsmith/server/exceptions/AppsmithError.java
```diff
@@ -135,6 +135,7 @@
     GIT_PULL_CONFLICTS(400, 4047, "Merge conflicts found during the pull operation: {0}", AppsmithErrorAction.DEFAULT, null, ErrorType.GIT_ACTION_EXECUTION_ERROR, ErrorReferenceDocUrl.GIT_PULL_CONFLICT),
     SSH_KEY_GENERATION_ERROR(500, 5015, "Failed to generate SSH keys, please contact Appsmith support for more details", AppsmithErrorAction.DEFAULT, null, ErrorType.GIT_CONFIGURATION_ERROR, null),
     GIT_UPSTREAM_CHANGES(400, 4048, "Looks like there are pending upstream changes. To prevent you from losing history, we will pull the changes and push them to your repo.", AppsmithErrorAction.DEFAULT, null, ErrorType.GIT_ACTION_EXECUTION_ERROR, ErrorReferenceDocUrl.GIT_UPSTREAM_CHANGES),
+    GENERIC_JSON_IMPORT_ERROR(400, 4049, "Unable to import application: {0}", AppsmithErrorAction.DEFAULT, null, ErrorType.BAD_REQUEST, null),
```
trishaanand, line 138: If possible, could we also add orgId as a parameter for this error?

app/server/appsmith-server/src/test/java/com/appsmith/server/controllers/ApplicationControllerTest.java
```diff
@@ -0,0 +1,122 @@
...
+                .expectStatus()
+                .isEqualTo(500)
+                .expectBody()
+                .json("{\n" +
+                        "    \"responseMeta\": {\n" +
+                        "        \"status\": 500,\n" +
+                        "        \"success\": false,\n" +
+                        "        \"error\": {\n" +
+                        "            \"code\": 5000,\n" +
+                        "            \"message\": \"Internal server error while processing request\"\n" +
```
trishaanand, line 98: Would it be possible to throw the correct error, aka, the file size is too large to be imported? Could we possibly catch the exact error in `GlobalExceptionHandler` and give the user a useful error?

If not possible, or too much work, this can be discarded as well.

trishaanand (approved):

trishaanand:
LGTM. Some very very minor comments added.

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationControllerCE.java
abhvsn: As I mentioned in the description we are not able to reproduce this issue on our side, so as a part of debugging this we are going to increase the supported filepart header size from 8kB (default) to 128kB and log these headers on server. So that we can have a RCA why this is happening, also this will be a temporary thing till we fix this from client side

PR #11080, opened 2022-02-10
feat: Return page slug for default pages in get applications API in home page

### LINKED ISSUE #10779: Return page slugs in response of applications list API
The FE calls `GET /api/applications/new` API to get list of organizations and applications. Currently the list of pages inside each application object does not have slugs. We need page slugs to generate the link of an application.

When user clicks on a application, we'll redirect them to this page: `/apps/application-slug/page/page-slug/pageid` where page-slug and pageid are from the default page. Hence we need page slug to generate this URL.

nayan-rafiq:
## Description
The `GET /application/new` API is used to show list of organizations, applications in home page. The list of pages under each application contains only page id. As we're introducing page slugs in URLs, we need the slug of the default page inside the response of this API. This PR adds default page's slug to the response of `GET /application/new` API.

Fixes #10779 

## Type of change

## How Has This Been Tested?

- JUnit tests
- Manually

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ApplicationFetcherCEImpl.java
```diff
@@ -159,6 +168,26 @@
...
+                            ).flatMap(Collection::stream).collect(Collectors.toList());
+
+                    // fetch the page slugs for the applications
+                    return newPageService.findSlugsByApplicationIds(applicationIds, READ_PAGES)
+                            .collectMultimap(NewPage::getApplicationId)
+                            .map(applicationPageMap -> {
+                                for (OrganizationApplicationsDTO orgApps : userHomepageDTO.getOrganizationApplications()) {
+                                    for (Application application : orgApps.getApplications()) {
+                                        setDefaultPageSlug(application, applicationPageMap, Application::getPages, NewPage::getUnpublishedPage);
+                                        setDefaultPageSlug(application, applicationPageMap, Application::getPublishedPages, NewPage::getPublishedPage);
```
abhvsn, line 185: Today this might not be required as we don't make this call from view mode, but I don't think there is any harm keeping this

abhvsn (approved):
LGTM

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ApplicationFetcherCEImpl.java
nayan-rafiq: I've checked this. Actually both edit mode and published mode default pages are required here. User can be app viewer for an organization and we need to redirect to the published default page for that case.

Issue #10575, opened 2022-01-24
[Task]: Refactor Select and MultiSelect Widget to have a base Shared widget


Tooluloope:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SubTasks

- Create a Base Select Widget
- Use the Base Select widget to configure Select and Multiselect widget
- improve UX of serverside filtering

Tooluloope:
Last DIscussion on Filtering

## Questions on Serverside Filtering
In select widgets we have the toggle for server side filtering which when enabled shows the onFilterUpdate action. Do we need this toggle? Can we show the onFilterUpdate directly instead of users having to switch on server side filtering?

## 
I am liking the idea to call it `onSearch`, we should confirm with @Nikhil-Nandagopal  once
Filterable is a very different property from actually searching and filtering, I recommend we dont mix up the two, Though your logic plays well, i.e. where filterable is false the onFilterUpdate is not required.
Filterable toggle is needed, Your use case pointed is valid, and actually was the original community request

dilippitchika:
@Tooluloope something to ask here, can we also use this base for tree select and multi tree select widget?

Tooluloope:
No we can't use this for those.
But we can create a Base for both TreeSelect widgets

dilippitchika:
Thanks, created a new issue - https://github.com/appsmithorg/appsmith/issues/13674

PR #11082, opened 2022-02-10
fix: select widget accept string and object in the defaultOption property

### LINKED ISSUE #11074: [Bug][DX] Use both string and Object values for the default values in select and Multiselect widget

### Current Behavior

It is difficult to bind the default value in a select and multi-select widget when we need to provide both label and value.

### Steps To Reproduce

The default value property in the Select Widget.

### Environment

Production

### Version

Cloud

### LINKED ISSUE #11081: [Bug]: Select Widget does not clear after submit or reset but is actually empty

### Current Behavior

Apologies if this is a dup. I did a search for "is:Bug select widget" and got 9+ pages of results. The select widget does not visually clear after submit or reset but the state does clear.

### Steps To Reproduce

Following the "Building a Store Catalog Manager" tutorial.
1. Using forms
2. Using a "Select" widget in place of the dropdown widget.
3. Complete the form and select something.
4. After submitting the form the dropdown widget does not visually clear/reset.
5. Fill out the form again but do not select anything just leave the select widget on what it was.
6. Submit the form.
7. Check the DB. You will see the first submit recorded data for the select but not the second.

### Environment

Deploy Preview

### Version

Cloud

Tooluloope:
## Description

- Accept both string and object for defaultOption in MultiSelect and Select to improve existing DX
- Reset Select Widget onSubmit or onReset
- Add `value` property to be able to get Select value in a form widget

Fixes #11074
Fixes #11081 

## Type of change

## How Has This Been Tested?

- Unit test to check validate defaultOption
- Unit test for Evaluations 

## Checklist:

    // Code coverage diff between base branch:release and head branch: fix/SelectV2-default-options-type 
 :green_circle: | total | 55.58 **(0.04)** | 37.06 **(0.06)** | 35.79 **(0.04)** | 55.91 **(0.04)**
 :green_circle: | app/client/src/entities/DataTree/dataTreeWidget.ts | 98.08 **(0)** | 83.33 **(8.33)** | 100 **(0)** | 97.83 **(0)**
 :red_circle: | app/client/src/utils/WorkerUtil.ts | 89.76 **(0)** | 70.59 **(-1.96)** | 100 **(0)** | 93.33 **(0)**
 :green_circle: | app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx | 80.95 **(14.28)** | 53.06 **(53.06)** | 70.59 **(10.59)** | 81.01 **(13.36)**
 :green_circle: | app/client/src/widgets/SelectWidget/component/index.styled.tsx | 45.45 **(1.7)** | 21.88 **(0)** | 0 **(0)** | 50 **(1.72)**
 :green_circle: | app/client/src/widgets/SelectWidget/component/index.tsx | 27.78 **(2.78)** | 0 **(0)** | 6.67 **(-0.47)** | 27.27 **(1.87)**
 :green_circle: | app/client/src/widgets/SelectWidget/widget/index.tsx | 54.93 **(5.91)** | 22.22 **(22.22)** | 50 **(3.85)** | 55.88 **(4.82)**
 :green_circle: | app/client/src/workers/evaluationUtils.ts | 58.37 **(0)** | 57.14 **(0.53)** | 64.52 **(0)** | 57.7 **(0)**</details>


app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -12,6 +15,106 @@ import {
...
+    });
+
+    if (_valid) {
+      return {
+        isValid: true,
+        parsed: values,
+      };
+    }
+  }
+  return {
```
sbalaji1192, line 112: This validation allows all object values.

app/client/src/widgets/SelectWidget/widget/index.tsx
```diff
@@ -5,9 +5,87 @@ import { EventType } from "constants/AppsmithActionConstants/ActionConstants";
...
+  let values: Record<string, unknown> = {};
+  if (value === undefined || value === null) {
+    return {
+      isValid: false,
+      parsed: "",
+      messages: ["This value does not evaluate to type: string"],
+    };
+  }
+  // if JSON, parse data and Check if it's an Object
+  try {
```
sbalaji1192, line 41: need to check if its a string before JSON.parse

sbalaji1192 (changes requested):

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -12,6 +15,115 @@ import {
 } from "rc-select/lib/interface/generator";
 import { Layers } from "constants/Layers";
 import { MinimumPopupRows, GRID_DENSITY_MIGRATION_V1 } from "widgets/constants";
+import { AutocompleteDataType } from "utils/autocomplete/TernServer";
+import _ from "lodash";
```
riodeuno, line 19: Let us tree shake this import. We can potentially reduce the build size by making sure we tree shake where-ever possible. I realise that lodash is being used in the application where the root import is necessary. However, I think it will help us in the long run if we use this approach as a practice.

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -12,6 +15,115 @@ import {
...
+      obj.hasOwnProperty("value") &&
+      _.isString(obj.label) &&
+      (_.isString(obj.value) || _.isFinite(obj.value))
+    );
+  };
+
+  /*
+   * Function to check for duplicate values in array
+   */
+  const hasUniqueValues = (arr: []) => {
```
riodeuno, line 46: Two things I would suggest here:
1. We move this function out of this validation function and put it in workers/utils. Not sure if this will work, but we can try and see.
2. We don't need to do an `Array.from`. We can probably check this using `set.size < arr.length`

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -12,6 +15,115 @@ import {
...
+    const uniqueValues = Array.from(new Set(arr));
+
+    return uniqueValues.length === arr.length;
+  };
+
+  /*
+   * When value is "['green', 'red']", "[{label: 'green', value: 'green'}]" and "green, red"
+   */
+  if (_.isString(value) && value !== "") {
```
riodeuno, line 55: This will probably pass strings like `"   "` as `true`. Let's make sure to trim the value, and them check for length.

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -12,6 +15,115 @@ import {
...
+      /*
+       * when value is "green, red", JSON.parse throws error
+       */
+      const splitByComma = (value as string).split(",") || [];
+
+      value = splitByComma.map((s) => s.trim());
+    }
+  }
+
+  if (value === "") {
```
riodeuno, line 71: Do we want to trim here and check for this condition?

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
riodeuno: Also, shall we type this as `arr: Array<string>`

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -328,7 +421,13 @@ class MultiSelectWidget extends BaseWidget<
     const options = isArray(this.props.options) ? this.props.options : [];
     const dropDownWidth = MinimumPopupRows * this.props.parentColumnSpace;
     const { componentWidth } = this.getComponentDimensions();
-
+    const values = this.props.selectedOptions
+      ? this.props.selectedOptions.map((o) =>
+          _.isNil(o.value)
+            ? { laebl: o, value: o }
```
riodeuno, line 427: There is a typo here. `label`. Let's fix this and see if there is any input we could have added to the test cases to have caught this.

app/client/src/widgets/SelectWidget/component/index.styled.tsx
```diff
@@ -12,6 +12,7 @@ import {
   BlueprintCSSTransform,
   createGlobalStyle,
 } from "constants/DefaultTheme";
+import _ from "lodash";
```
riodeuno, line 15: Same suggestion as above regarding tree-shaking.

app/client/src/widgets/SelectWidget/widget/index.tsx
```diff
@@ -300,6 +349,18 @@ class SelectWidget extends BaseWidget<SelectWidgetProps, WidgetState> {
   }
   componentDidUpdate(prevProps: SelectWidgetProps): void {
     // removing selectedOptionValue if defaultValueChanges
+    if (
```
riodeuno, line 352: Let's look at a solution in a separate PR for this:
I am curious if this can be caught in the `derviedProperty`? Instead of using componentDidUpdate and componentDidMount.

riodeuno (changes requested):
A few notes and suggestions.

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
Tooluloope: We currently allow empty values,
Since the values are Unique, You can only have one option with an Empty string and most users use this as a placeholder

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
Tooluloope: I think just adding types to Values would have caught this

app/client/src/widgets/SelectWidget/widget/index.tsx
Tooluloope: I'll add that to the list of task here
https://github.com/appsmithorg/appsmith/issues/10575

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
Tooluloope: We currently allow empty values,
Since the values are Unique, You can only have one option with an Empty string and most users use this as a placeholder

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
Tooluloope: Didn't work in utils.
is there any reason why you advise I put this in utils?

app/client/src/widgets/MultiSelectWidgetV2/widget/index.test.tsx
```diff
@@ -0,0 +1,249 @@
...
+      defaultOptionValueValidation(input, {} as MultiSelectWidgetProps, _),
+    ).toEqual({
+      isValid: true,
+      parsed: ["green"],
+      messages: [""],
+    });
+  });
+
+  it("should get tested with simple string", () => {
+    const input = `{"green"`;
```
riodeuno, line 96: I'm not sure if this is a typo or do we pass through any string values in default values? Or does this match one of the values in the options?

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
riodeuno: Yes, I believe as we're defining a function within another function -- so each time the parent function runs JS has to deal with defining the child function, this should come with a performance expense. 

If we can't move it out, it's ok, let's move on with this.

app/client/src/widgets/MultiSelectWidgetV2/widget/index.test.tsx
sbalaji1192: @riodeuno We pass string values in defaultValues, if there is no binding. This test case tests that invalid json strings don't throw error.

riodeuno (approved):
Looks good to me.

riodeuno:
@sbalaji1192 @Tooluloope Although we have added jest tests to validate our updates to the validation function -- I can see that we've made a few updates to the components to handle certain values and use-cases. We should look into adding/updating cypress tests for these scenarios, if we don't already have them in place.

sbalaji1192:
> @sbalaji1192 @Tooluloope Although we have added jest tests to validate our updates to the validation function -- I can see that we've made a few updates to the components to handle certain values and use-cases. We should look into adding/updating cypress tests for these scenarios, if we don't already have them in place.

+1

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -12,6 +15,114 @@ import {
...
+    try {
+      /*
+       * when value is "['green', 'red']", "[{label: 'green', value: 'green'}]"
+       */
+      value = JSON.parse(value as string);
+    } catch (e) {
+      /*
+       * when value is "green, red", JSON.parse throws error
+       */
+      const splitByComma = (value as string).split(",") || [];
```
ashit-rath, line 64: @riodeuno @Tooluloope  Should this check be enough for checking if it is a `CSV`?
ex - If I pass the value `"['green', 'red]"`, (notice `red` does not have closing single quotes). It would go to the catch clause and the value would come out to be `["['green', "'red]"]`.
This would a mistake by the Appsmith developer (missing quotes) but we are not informing the developer and treating it as a CSV.

Tooluloope: @ashit-rath, So when the user sees that in the Popup, they would be able to correct the value they inputted.

It's like having `'Green` in the PropertyPane for any width that accepts sting, it would still render as a string `"'Green". The developer can then correct  and add the extra quote

riodeuno: @ashit-rath I see your point. Nice catch! @Tooluloope could you look into the logic here to catch this case?

sbalaji1192: @riodeuno @ashit-rath There could be many such cases involving invalid JSON strings. For eg. the user forgetting to close the square bracket or curly bracket. Are we gonna cover that case as well? 
I guess as @Tooluloope  pointed out, the user can see the evaluated value popup and debug the issue. rather than us trying to catch all invalid cases and throw an error.

ashit-rath: @sbalaji1192 You are right Balaji, we should have a limit as to how far we should validate. IMO we should only check for valid CSV.

Found this well-explained answer in [stack overflow](https://stackoverflow.com/questions/8493195/how-can-i-parse-a-csv-string-with-javascript-which-contains-comma-in-data). It does not fully comply with [RFC 4180](https://datatracker.ietf.org/doc/html/rfc4180#page-2) but is a good enough case for us.

ashit-rath: BTW what was the original reason to have comma-separated strings? Are we trying to support values coming from google sheets?

Tooluloope: We were supporting such values earlier when Select and Multiselect were 1 widget, and continue it support till now

Tooluloope: Also, the value you dropped in the First comment wasn't a CSV

app/client/src/widgets/SelectWidget/widget/index.tsx
```diff
@@ -3,11 +3,83 @@ import BaseWidget, { WidgetProps, WidgetState } from "../../BaseWidget";
...
+  } else if (hasLabelValue(value)) {
+    /*
+     * When value is {label: "green", value: "green"}
+     */
+    isValid = true;
+    parsed = value;
+  } else {
+    isValid = false;
+    parsed = {};
+    message = `value does not evaluate to type: string | { "label": "label1", "value": "value1" }'`;
```
ashit-rath, line 63: @Tooluloope There is an extra `'` at the end of this line.

riodeuno (approved):

vivekverma2312:
@Tooluloope Object value is not working when I use sniping mode.
https://www.loom.com/share/6d5872d293fe43929beb1ee097dc5b67

shwetha-ramesh:
@Tooluloope Co tested this PR with @YogeshJayaseelan. DP working fine for issues  #11081 & #11074. 
Moving to Done.

PR #11097, opened 2022-02-11
fix: Remove default values for error message in analytics

### LINKED ISSUE #11096: [Bug]: Git syncs analytics

### Current Behavior

We are not sending gitHosting provider to the analytics
Error data points should be null for success states

### Steps To Reproduce

We are not sending gitHosting provider to the analytics
Error data points should be null for success states

### Environment

Production

### Version

Cloud

AnaghHegde:
## Description

> Git-sync analytics we were sending empty string for error data points. This was causing issues in filtering out the data between error events and success events. This change set removes the default values for the error data points in case of success events and pass the null value to the analytics.

Fixes #11096

## Type of change

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -2245,8 +2245,8 @@ private boolean isInvalidDefaultApplicationGitMetadata(GitApplicationMetadata gi
                             "applicationId", defaultApplicationId,
                             "organizationId", defaultIfNull(application.getOrganizationId(), ""),
                             "branchApplicationId", defaultIfNull(application.getId(), ""),
-                            "errorMessage", defaultIfNull(errorMessage, ""),
-                            "errorType", defaultIfNull(errorType, ""),
+                            "errorMessage", errorMessage,
+                            "errorType", errorType,
```
abhvsn, line 2249: This will still get registered can we remove these from the Map itself?
Check implementation in AnalyticsService
```
// Segment throws an NPE if any value in `properties` is null.
        for (final Map.Entry<String, Object> entry : analyticsProperties.entrySet()) {
            if (entry.getValue() == null) {
                analyticsProperties.put(entry.getKey(), "");
            }
        }
```

AnaghHegde: good point i was not aware of this.

abhvsn (approved):

PR #11138, opened 2022-02-14
fix: handled to hide widget in preview mode

### LINKED ISSUE #10228: [Bug]: App in preview mode shows hidden widgets

### Current Behavior

Hidden widgets should not be visible in **preview mode**

### Steps To Reproduce

1. Turn off `Visible` property of a widget and switch to **preview mode**. Observe hidden widget is still visible

### Environment

Production

### Version

Cloud

techbhavin:
## Description

>  Add Wrapper component for render preview mode. this component is responsible for hiding and showing content 
>   components base on isVisible 

Fixes #10228 

## Type of change

## Checklist:

    // Code coverage diff between base branch:release and head branch: fix/preview-mode-hide-widget 
 :green_circle: | total | 55.71 **(0.01)** | 37.06 **(0)** | 35.75 **(0.01)** | 56.07 **(0.01)**
 :sparkles: :new: | **app/client/src/components/editorComponents/PreviewModeComponent.tsx** | **90** | **50** | **100** | **88.89**
 :red_circle: | app/client/src/utils/WorkerUtil.ts | 88.98 **(-0.78)** | 70.59 **(-1.96)** | 100 **(0)** | 92.38 **(-0.95)**
 :green_circle: | app/client/src/widgets/BaseWidget.tsx | 79.53 **(0.67)** | 44.44 **(0)** | 75.76 **(0.76)** | 78.69 **(0.72)**</details>


app/client/src/components/editorComponents/PreviewModeComponent.tsx
```diff
@@ -0,0 +1,23 @@
...
+};
+
+/**
+ * render only visible components in preview mode
+ */
+function PreviewModeComponent({ children, isVisible }: Props) {
+  const isPreviewMode = useSelector(previewModeSelector);
+
+  if (isPreviewMode && !isVisible) {
+    return null as any;
```
sbalaji1192, line 17: do we need `as any` ?

sbalaji1192 (commented):
@riodeuno Could you please have a look at these changes?

app/client/src/components/editorComponents/PreviewModeComponent.tsx
riodeuno: @techbhavin I agree with @sbalaji1192 we don't need to return anything here if the condition is false, I believe.

We can simply say `if(!(isPreviewMode && !isVisible)) return children;`. Please see if this approach works.

riodeuno (changes requested):
This needs a proper cypress test for the following:
- A successful case, where an invisible widget in preview mode doesn't show up
- An unsuccessful case, where a visible widget in preview mode shows up (We probably have this one)

app/client/src/components/editorComponents/PreviewModeComponent.tsx
sbalaji1192: we can use a much simpler condition `if (!isPreviewMode || isVisible) return children;`

sbalaji1192:
> This needs a proper cypress test for the following:
> 
> * A successful case, where an invisible widget in preview mode doesn't show up
> * An unsuccessful case, where a visible widget in preview mode shows up (We probably have this one)

And a case to make sure that the invisible widgets show up in the edit mode.

app/client/src/components/editorComponents/PreviewModeComponent.tsx
```diff
@@ -0,0 +1,19 @@
...
+  isVisible?: boolean;
+};
+
+/**
+ * render only visible components in preview mode
+ */
+function PreviewModeComponent({ children, isVisible }: Props) {
+  const isPreviewMode = useSelector(previewModeSelector);
+  if (!(isPreviewMode && !isVisible)) return children;
+  return null as any;
```
riodeuno, line 16: @techbhavin I am curious as to why would we need `any` here? Can we try returning a `ReactNode` or something like that? Can we return a Fragment?

riodeuno (approved):
Looks good to me! @techbhavin let's move this to the QA pipeline.

shwetha-ramesh:
@techbhavin tested this PR. DP works fine. Widgets when ```Visible``` is off doesn't show up in Preview mode

image

Moving to Done.

PR #11144, opened 2022-02-14
feat: Add api to check the git connection after the key is regenerated

### LINKED ISSUE #10299: [Task] SSH key management
A place to manage ssh keys. 
- Add another ssh key and switch that as the primary key
- Remove/Replace existing key etc.

AnaghHegde:
## Description

> A user should have an option to regenerate the ssh keys for the git connected applications in case the key is compromised. This change set adds an API to test the connection after the key is regenerated by the user. 

Fixes #10299

## Type of change

## How Has This Been Tested?

- Locally
- TC's on the way

## Checklist:

### DIFF
--- a/app/server/appsmith-git/src/main/java/com/appsmith/git/service/GitExecutorImpl.java
+++ b/app/server/appsmith-git/src/main/java/com/appsmith/git/service/GitExecutorImpl.java
@@ -48,6 +48,7 @@ import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
@@ -667,6 +668,21 @@ public class GitExecutorImpl implements GitExecutor {
                 return git.getRepository().getBranch();
             }
         }).subscribeOn(scheduler);
+    }
+
+    @Override
+    public Mono<Boolean> testConnection(String publicKey, String privateKey, String remoteUrl) {
+        return Mono.fromCallable(() -> {
+            TransportConfigCallback transportConfigCallback = new SshTransportConfigCallback(privateKey, publicKey);
+            Git.lsRemoteRepository()
+                    .setTransportConfigCallback(transportConfigCallback)
+                    .setRemote(remoteUrl)
+                    .setHeads(true)
+                    .setTags(true)
+                    .call();
+            return true;
+        }).timeout(Duration.ofMillis(Constraint.REMOTE_TIMEOUT_MILLIS))
+                .subscribeOn(scheduler);
     }
 
 
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
@@ -2087,25 +2087,44 @@ public class GitServiceCEImpl implements GitServiceCE {
         return getApplicationById(defaultApplicationId)
                 .flatMap(application -> {
                     GitApplicationMetadata gitApplicationMetadata = application.getGitApplicationMetadata();
-                    if(isInvalidDefaultApplicationGitMetadata(gitApplicationMetadata)) {
+                    if (isInvalidDefaultApplicationGitMetadata(gitApplicationMetadata)) {
                         return Mono.error(new AppsmithException(AppsmithError.INVALID_GIT_SSH_CONFIGURATION));
                     }
                     return gitExecutor.testConnection(
                             gitApplicationMetadata.getGitAuth().getPublicKey(),
                             gitApplicationMetadata.getGitAuth().getPrivateKey(),
                             gitApplicationMetadata.getRemoteUrl()
-                    ).onErrorResume(error -> {
-                        if (error instanceof TransportException) {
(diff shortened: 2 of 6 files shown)


app/server/appsmith-git/src/main/java/com/appsmith/git/service/GitExecutorImpl.java
```diff
@@ -669,6 +670,21 @@ private Path createRepoPath(Path suffix) {
...
+    @Override
+    public Mono<Boolean> testConnection(String publicKey, String privateKey, String remoteUrl) {
+        return Mono.fromCallable(() -> {
+            TransportConfigCallback transportConfigCallback = new SshTransportConfigCallback(privateKey, publicKey);
+            Git.lsRemoteRepository()
+                    .setTransportConfigCallback(transportConfigCallback)
+                    .setRemote(remoteUrl)
+                    .setHeads(true)
+                    .setTags(true)
+                    .call();
```
abhvsn, line 682: If I'm not wrong `git ls-remote` fetches the list of `commit ids` along with the `refs`, does this checks the write access as well?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -2082,6 +2082,33 @@ else if (error.getMessage().contains("Nothing to fetch")) {
...
+                .flatMap(application -> {
+                    GitApplicationMetadata gitApplicationMetadata = application.getGitApplicationMetadata();
+                    if(isInvalidDefaultApplicationGitMetadata(gitApplicationMetadata)) {
+                        return Mono.error(new AppsmithException(AppsmithError.INVALID_GIT_SSH_CONFIGURATION));
+                    }
+                    return gitExecutor.testConnection(
+                            gitApplicationMetadata.getGitAuth().getPublicKey(),
+                            gitApplicationMetadata.getGitAuth().getPrivateKey(),
+                            gitApplicationMetadata.getRemoteUrl()
+                    ).onErrorResume(error -> {
```
abhvsn, line 2097: Can we please log the error here

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -2082,6 +2082,33 @@ else if (error.getMessage().contains("Nothing to fetch")) {
...
+                        if (error instanceof InvalidRemoteException) {
+                            return Mono.error(new AppsmithException(AppsmithError.INVALID_GIT_CONFIGURATION, error.getMessage()));
+                        }
+                        if (error instanceof TimeoutException) {
+                            return Mono.error(new AppsmithException(AppsmithError.GIT_EXECUTION_TIMEOUT));
+                        }
+                        return Mono.error(new AppsmithException(AppsmithError.GIT_GENERIC_ERROR, error.getMessage()));
+                    });
+                });
+    }
```
abhvsn, line 2110: Should we log the analytics event as well?

abhvsn (changes requested):
Minor comments added for improvements

app/server/appsmith-git/src/main/java/com/appsmith/git/service/GitExecutorImpl.java
AnaghHegde: No to check write access we need to do push operation.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: Thats a good point. I will add this for analytics.

app/server/appsmith-git/src/main/java/com/appsmith/git/service/GitExecutorImpl.java
abhvsn: AFAIK we should be checking the write access by pushing the logfile, am I missing something here

AnaghHegde: Yeah agreed. But that would involves committing and pushing to remote repo. What is the user has uncommitted changes during the operation. That is why i resorted to this way. I believe the user will update the keys and not giving the write access chance is less. But i am open to other alternatives.

AnaghHegde: I can edit the readme file in that case and do commit+push

abhvsn: Can you please bring this discussion on pod so that we can get the inputs from other folks, IMO committing another file won't be a big deal from implementation point of view. Also I'm not inclined in making changes to readme file as this change is not related to it and this may confuse users

abhvsn: @AnaghHegde are we going ahead with the log file or just the read access will be enough at this point?

AnaghHegde: @abhvsn We can do that, but i just don't think we need that at the movement. Because user would already gone through updating keys.

abhvsn (approved):

Issue #10684, opened 2022-01-27
[Bug]: The application cant be connected to private repo even after changing the visibility of the old private repo


RakshaKShetty:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The application cant be connected to private repo even after changing the visibility of the old private repo
![image](https://user-images.githubusercontent.com/45958978/151341400-d8317fcd-62f2-4d56-86bb-868a7184bbc5.png)
![image](https://user-images.githubusercontent.com/45958978/151341764-ef4f93e1-4f89-4874-93b5-4af0cbc110f3.png)

### Steps To Reproduce

1. Go to the application
2. connect the 3 applications to a private repo
3. Now try to connect the 4th application to a private repo
4. It will ask for  contact sales or disconnect the old app
5. Now Change the visibility from private to public to any one of the old app
6. Now try to connect the new private repo
7. The application will not connect and the user is displayed with contact sales modal

### Environment

Release

### Version

Cloud

PR #11151, opened 2022-02-14
fix: Take previously connected git applications accessibility into consideration to accurately calculate private repo count 

abhvsn:
## Description

> When we are connecting a application to git-repo we are not taking into account the fact that users can change the accessibility for the previously connected application from the remote repo, this leads to the incorrect number of currently connected private repo count. This PR makes sure to check the accessibility for all the private repo in an org and update the application git data if user alters the accessibility from remote.

Fixes https://github.com/appsmithorg/appsmith/issues/10684

## Type of change

## How Has This Been Tested?

> Junit TC
> Manual test

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -411,7 +411,7 @@ private String getReason() {
                                     .flatMap(application -> gitCloudServicesUtils.getPrivateRepoLimitForOrg(application.getOrganizationId(), false)
                                             .flatMap(limitCount -> {
                                                 //get git connected apps count from db
-                                                return applicationService.getGitConnectedApplicationCount(application.getOrganizationId())
+                                                return applicationService.getGitConnectedApplicationsCountWithPrivateRepoByOrgId(application.getOrganizationId())
```
AnaghHegde, line 414: IIUC, say there are 3 connected Application(a1, a2, a3-public-repo ). Now if i connect a 4th repo (private) and then make the a3 private repo then commit+push operation will go through right?
We should call the `getGitConnectedApplicationsWithPrivateRepoByOrgId` in commit operation right?

abhvsn: In the commit application, we check if the repo is public and also if the user has toggled accessibility to private, if yes then again follow the same flow to check the available repo count limit before making the commit. On the other hand for private repo we commit the application directly.

AnaghHegde: We have the same check present in connectFlow as well. We need the count only if the repo is private. Else any of the above conditions won't be present in the flow. So i don't understand how is the scenario i mentioned before is handled?

AnaghHegde (approved):
LGTM

PR #11193, opened 2022-02-16
chore: Updated Typescript types.

### LINKED ISSUE #9134: [Task]: Improve Typescript support for Sagas

### SubTasks

We are not using typescript efficiently when it comes to Sagas. If we are using it effectively, we could've avoided some run time bugs.
I've come across [typed-redux-saga](https://github.com/agiledigital/typed-redux-saga) project and think this library can help us to improve typescript in the case of Sagas.

### LINKED ISSUE #11091: [Task]: Update Typescript to 4.2.4

### SubTasks

- [x] Upgrade Typescript
- [ ] Fix type errors
- [ ] Add instability logging to identify data integrity issues

yaldram:
## Description
- Update typescript to `4.2.4` & fix typescript types.

Fixes #11091
Fixes #9134

## Type of change

## How Has This Been Tested?

- Run `yarn start` the code should compile without any typescript errors in the console.

## Checklist:


app/client/src/sagas/OnboardingSagas.ts
```diff
@@ -66,10 +66,14 @@ import { selectWidgetInitAction } from "actions/widgetSelectionActions";
...
 import { updateWidgetName } from "actions/propertyPaneActions";
 import AnalyticsUtil from "utils/AnalyticsUtil";
+import { DataTree } from "entities/DataTree/dataTreeFactory";
+import { CanvasWidgetsReduxState } from "reducers/entityReducers/canvasWidgetsReducer";
+import { User } from "constants/userConstants";
 
 function* createApplication() {
   const userOrgs: Organization[] = yield select(getOnboardingOrganisations);
-  const currentUser = yield select(getCurrentUser);
+  const currentUser: User | undefined = yield select(getCurrentUser);
```
arunvjn, line 75: For such selectors or function where we know that return types cannot be undefined/null, we should wrap the return values of these functions/selectors and reduce type. You could use a function like below to reduce the type.

```
function shouldBeDefined<T>(argument: T | undefined | null, message: string = 'some error message to default to'): T {
  if (argument === undefined || argument === null) {
    throw new TypeError(message);
  }

  return argument;
}
```
Here the return value of the getCurrentUser can be wrapped with shouldBeDefined to get the type we expect.

app/client/src/sagas/ModalSagas.ts
```diff
@@ -161,8 +164,11 @@ export function* closeModalSaga(
     let widgetIds: string[] = [];
     // If modalName is provided, we just want to close this modal
     if (modalName) {
-      const widget = yield select(getWidgetByName, modalName);
-      widgetIds = [widget.widgetId];
+      const widget: FlattenedWidgetProps | undefined = yield select(
```
arunvjn, line 167: We should write and use a utility type here like `type MayNotExist<T> = T | undefined;`

app/client/src/sagas/ActionExecution/NavigateActionSaga.ts
```diff
@@ -37,8 +37,8 @@ const isValidUrlScheme = (url: string): boolean => {
 export default function* navigateActionSaga(
   action: NavigateActionDescription["payload"],
 ) {
-  const pageList = yield select(getPageList);
-  const applicationId = yield select(getCurrentApplicationId);
+  const pageList: PageListPayload = yield select(getPageList);
```
arunvjn, line 40: We should remove PageListPayload type from codebase and start using Array<Page> or Page[]. It's a lot easier to understand while reading.

app/client/src/sagas/ActionSagas.ts
```diff
@@ -16,8 +16,8 @@ import {
 } from "redux-saga/effects";
 import { Datasource } from "entities/Datasource";
 import ActionAPI, { ActionCreateUpdateResponse } from "api/ActionAPI";
-import { GenericApiResponse } from "api/ApiResponses";
-import PageApi from "api/PageApi";
+import { ApiResponse, GenericApiResponse } from "api/ApiResponses";
```
arunvjn, line 19: Let's remove GenericApiResponse and make ApiResponse a generic. 

```
export type ApiResponse<T = any> = {
    data: T;
    responseMeta: ResponseMeta;
    code?: string;
}
```

app/client/src/sagas/ActionSagas.ts
```diff
@@ -448,7 +452,7 @@ function* moveActionSaga(
 ) {
   const actionObject: Action = yield select(getAction, action.payload.id);
   try {
-    const response = yield ActionAPI.moveAction({
+    const response: AxiosResponse<any> = yield ActionAPI.moveAction({
```
arunvjn, line 455: Can we add stricter type here?

app/client/src/sagas/PluginSagas.ts
```diff
@@ -73,10 +76,16 @@ function* fetchPluginFormConfigsSaga() {
       pluginIdFormsToFetch.add(apiPlugin.id);
     }
     for (const id of pluginIdFormsToFetch) {
-      pluginFormRequests.push(yield call(PluginsApi.fetchFormConfig, id));
+      const response: GenericApiResponse<PluginFormPayload> = yield call(
```
arunvjn, line 79: We need to standardize the names for request and response types. This needs to done across the code base. Naming convention needs to be discussed and decided. I vote for ${api}RequestData and ${api}ResponseData. So in this case, the types would be `FetchFormConfigRequestData` and `FetchFormConfigResponseData`

app/client/src/sagas/WidgetOperationUtils.ts
```diff
@@ -749,6 +749,10 @@ export function* createWidgetCopy(widget: FlattenedWidgetProps) {
   };
 }
 
+export type WidgetsInTree = (WidgetProps & {
+  children?: string[] | undefined;
```
arunvjn, line 753: Redundant undefined here.

app/client/src/selectors/editorSelectors.tsx
```diff
@@ -142,7 +142,7 @@ export const getCurrentPageName = createSelector(
   getPageListState,
   (pageList: PageListReduxState) =>
     pageList.pages.find((page) => page.pageId === pageList.currentPageId)
-      ?.pageName,
+      ?.pageName || "PAGE_NAME_NOT_FOUND",
```
arunvjn, line 145: Throw error here when pageName is not found. Makes sure the selector always return string. Refer
https://github.com/appsmithorg/appsmith/pull/11193/files#r811332675

app/client/src/sagas/PluginSagas.ts
yaldram: Good suggestion. @riodeuno please provide your inputs.

app/client/src/sagas/OnboardingSagas.ts
yaldram: Thanks for your suggestions. As discussed yesterday we have many selectors that use `.find()` and therefore return `T | undefined`. We can either wrap all our selectors with the above function or we can improve the types like making the `actionId` of type `keyof typeof state.actions`.

app/client/src/sagas/ModalSagas.ts
yaldram: Thanks make sense, we will first try to handle the `T | undefined` cases in the selectors if a selector is returning `T | undefined` I will replace all the above type with `type MayNotExist<T> = T | undefined;`

app/client/src/sagas/OnboardingSagas.ts
arunvjn: @yaldram That might not work since the return type of .find() itself,  is `T | undefined`. 
Also, we should only wrap the return  where we are sure that it cannot be undefined.

yaldram: @arunvjn Thanks, yes we need to implement `shouldbeDefined` like solution. @riodeuno

yaldram:
@arunvjn @riodeuno I am leaving behind some of my findings : - 
1. I have implemented `shouldBeDefined` locally thanks to @arunvjn's suggestions a lot of the type issues were resolved.
2. But there are some types in our code which do return `undefined` and we have handled them in places and we have some type mismatches also, for now I have `@ts-expect-error` but we can resolve these issues by talking to the authors.
3. We should implement `type MayNotExist<T>`  at many places in our code we have `T | undefined` (across 158 files).
4. We should also type the `unknown` Api Responses if possible.

yaldram:
Got some answers for typing redux-saga - https://github.com/redux-saga/redux-saga/issues/2251

app/client/src/api/ActionAPI.tsx
```diff
@@ -35,10 +35,10 @@ export interface QueryConfig {
   queryString: string;
 }
 
-export interface ActionCreateUpdateResponse extends ApiResponse {
+export type ActionCreateUpdateResponse = ApiResponse & {
```
riodeuno, line 38: @yaldram Could you explain why we have changed `interface` to `type` here? 
The reason for asking, is that if we have a standard reason for deciding one over the other, we can make it a practice in all of our codebase.

app/client/src/api/ActionAPI.tsx
```diff
@@ -61,21 +60,19 @@ export interface ActionApiResponseReq {
...
-  responseMeta: ResponseMeta;
-  data: {
-    body: Record<string, unknown> | string;
-    headers: Record<string, string[]>;
-    statusCode: string;
-    isExecutionSuccess: boolean;
-    request: ActionApiResponseReq;
-    errorType?: string;
-  };
+export type ActionExecutionResponse = ApiResponse<{
```
riodeuno, line 63: I am wondering if this could be achieved by using `interface` `extends ApiResponse`.
@yaldram I believe we had discussed the advantage of this approach. Could you mention it here so that we can consider this consistently further into the future.

app/client/src/api/JSActionAPI.tsx
```diff
@@ -1,12 +1,14 @@
...
 import { AxiosPromise } from "axios";
 import { JSCollection } from "entities/JSCollection";
-import { ApiResponse, GenericApiResponse } from "./ApiResponses";
+import { ApiResponse } from "./ApiResponses";
 import { Variable, JSAction } from "entities/JSCollection";
 import { PluginType } from "entities/Action";
-export interface JSCollectionCreateUpdateResponse extends ApiResponse {
+
+export type JSCollectionCreateUpdateResponse = ApiResponse & {
   id: string;
```
riodeuno, line 9: Can you confirm if this is supposed to be something like `ApiResponse<{id: string}>` instead?

app/client/src/api/ActionAPI.tsx
yaldram: Actually we have our `ApiResponse` as a generic type here, before it was an interface so I have changed some interfaces to types like : - 
```typescript
type TemplateResponse = ApiResponse<Template>
```
and if you want to add more params to the type just extend it using `&` like - 
```typescript
type JSCollectionCreateUpdateResponse = ApiResponse & { id: string } 
```

app/client/src/api/ActionAPI.tsx
yaldram: I had a discussion with @arunvjn where he suggested to use `ApiResponse<T>` as a Generic type we had a `GenericApiResponse` previously but it had the same fields as `ApiResponse` with an additional field that can be undefined. We are now replacing all our `extends ApiResponse` to use the generic version of it.

old version : - 
```typescript
export interface ActionExecutionResponse {
  responseMeta: ResponseMeta;
  data: {
    body: Record<string, unknown> | string;
    headers: Record<string, string[]>;
    statusCode: string;
    isExecutionSuccess: boolean;
    request: ActionApiResponseReq;
    errorType?: string;
  };
  clientMeta: {
    duration: string;
    size: string;
  };
}
```

now : - 
```typescript
export type ApiResponse<T = unknown> = {
  responseMeta: ResponseMeta;
  data: T;
  code?: string;
};

export type ActionExecutionResponse = ApiResponse<{
  body: Record<string, unknown> | string;
  headers: Record<string, string[]>;
  statusCode: string;
  isExecutionSuccess: boolean;
  request: ActionApiResponseReq;
  errorType?: string;
}> & {
  clientMeta: {
    duration: string;
    size: string;
  };
};
```

app/client/src/api/JSActionAPI.tsx
yaldram: Yes, it should be : - 
```typescript
export type JSCollectionCreateUpdateResponse = ApiResponse & {
  id: string;
}
```
The above type has all the fields of ApiResponse, plus the `id: string` field.
old version : - 
```typescript
export interface JSCollectionCreateUpdateResponse extends ApiResponse {
  id: string;
}
```

app/client/src/api/ActionAPI.tsx
arunvjn: All API responses are of the type APIResponse. We are doing something wrong here.
Did you mean to do something like ApiResponse<Template & { id: string }> ?

app/client/src/api/JSActionAPI.tsx
arunvjn: Yes, this should be `ApiResponse<{id: string}>`

yaldram: Hey @arunvjn actually we have an additional field here `id of type string` which will not be part of the data.

app/client/src/api/ActionAPI.tsx
yaldram: Hey @arunvjn no actually this `id field of type string` is an additional field on the response and not part of the data field.

riodeuno:
@yaldram Shall we close this without deleting the branch in favour of #13195?

yaldram:
> @yaldram Shall we close this without deleting the branch in favour of #13195?

Sure we can.

riodeuno:
Closing in favour of #13195

PR #11204, opened 2022-02-16
fix: NPE when application has no page

### LINKED ISSUE #10779: Return page slugs in response of applications list API

nayan-rafiq:
## Description

This PR fixes a NPE when application has no page or page are missing.

Fixes #10779

## Type of change

## How Has This Been Tested?
- Manually

## Checklist:


sharat87 (approved):

nayan-rafiq:
> @nayan-rafiq Can you help me understand the bug & the fix better? Why would an application not have any pages in it? Wouldn't that become an invalid application?

Home page was not loading for @sharat87 in release because there was an app that has no page in `newPage`. It seemed that this app was created by cypress quite a long time ago. As this somehow happened, we suspect this may happen for other users also. That's why we added a check to make sure list of pages is not null. 

You are right, the app will be an invalid one.

PR #11217, opened 2022-02-16
fix: unresponsive TextWidget when bound to large data

### LINKED ISSUE #10771: [Bug]: TextWidget crashes when bound to large data, like JSON from api response

### Current Behavior

Whenever you bind large data to text widget, it crashes the application.
Usually users bind api response to a text widget for debugging purposes, and this puts the application in unrecoverable state.

### Why this happens
Unlike table or list widget, we don't restrict the amount of data we show in a text widget. 
When we have lot of nested data, that results in a high amount of dom nodes.
Response from this innocent looking API of about 700KB creates a million dom nodes
https://restcountries.com/v3.1/all

Half a million among those are BR tags

Screenshot 2022-01-31 at 8 16 20 PM

### Suggested fix?
Truncate the text when its large. Show a small button/ellipsis to load more text.
We can estimate the number of characters we have to show based on area of the text widget.

```
// Sample code
  getText() {
    const { text } = this.props;
    if (text && text?.length > 10000) {
      return text?.slice(0, 5000);
    }
    return text;
  }
```

#### Things to consider

- Since we allow HTML tags in the text widget, this may remove the closing tags and mess with rest of the application. 

### Steps To Reproduce

- Make an api pointing to https://restcountries.com/v3.1/all
- Bind it to a text widget

This will put the application in unrecoverable state.

### Environment

Production

### Version

Cloud

prsidhu:
## Description

Cause - When TextWidget is bound to large data (JSON / html) through an api, then the screen becomes unresponsive due to creation to large number of DOM nodes.

Fix - 
1. Enable ```shouldTruncate``` by default.
2. In TextComponent, if truncation is enabled; then calculate the maximum length of text that can be displayed in the associated text box ( = widget area / rough area covered by a single character). Store the substring as a state variable ```displayText```.

Fixes #10771

## Type of change

## How Has This Been Tested?
1. Pull the branch.
2. Make an api pointing to https://restcountries.com/v3.1/all
3. Bind it to a text widget.
4. Upon refresh the TextWidget should render the truncated text without becoming unresponsive.
5. Click on ellipses icon to launch the modal component. Observe that the complete text is rendered in the modal.

## Checklist:

    // Code coverage diff between base branch:release and head branch: fix/10771 
 :green_circle: | total | 55.74 **(0.03)** | 37.13 **(0.05)** | 35.8 **(0.02)** | 56.09 **(0.02)**
 :green_circle: | app/client/src/widgets/TextWidget/component/index.tsx | 86.67 **(13.59)** | 50.23 **(5.2)** | 65.38 **(7.69)** | 85.71 **(13.34)**</details>


somangshu:
@allcontributors add @prsidhu as a contributor for code, bug

somangshu:
@prsidhu this will also require a documentation change based on your change.

cc @dilippitchika @Pranay105

somangshu:
Feedback from the DP:
- Truncate text is enables by default, Which is different from the original behaviour. TBD
- The modal when bound to the API takes a while to open and seems like there are spikes in the memory usage in the performance tab cc @SatishGandham 

**Suggestion**
- If we go about enabling this internally only when big text is entered, then we also need to disable it again once the text size is under the acceptable limit. We could consider the the max height of the component to decide this too. I am not sure how much the product agrees to it though. cc @Nikhil-Nandagopal @dilippitchika

prsidhu:
@somangshu Based on this change, truncation is enabled by default but the text will only be truncated (along with ellipses icon render) if the text length is more than what can fit in the given area; which matches the given suggestion.

somangshu:
> @somangshu Based on this change, truncation is enabled by default but the text will only be truncated (along with ellipses icon render) if the text length is more than what can fit in the given area; which matches the given suggestion.

Ill try to explain what I am talking about @prsidhu, Check out the below images from DP and Release env.
Screenshot 2022-02-17 at 11 36 17 AM


Screenshot 2022-02-17 at 11 36 56 AM

The basic difference is that the default behaviour changes for the user i.e. the text is placed differently by default, and this might seem unexpected to the user.

prsidhu:
ah understood. Fixing this.

prsidhu:
@dilippitchika @somangshu Fixed the alignment issue.

somangshu:
@prsidhu I can still see the default text to be aligned at top

Screenshot 2022-02-17 at 4 15 04 PM

somangshu:
@prsidhu i check the app again, The original issue is fixed, i.e. the text is not centered based on the amount of content that can fit inside the widget boundaries. 

However I was able to find a scenario which crashed the app. 
- use the api to bind to the text widget
- disable truncate
- you app is now stuck

cc @SatishGandham 

Another point is the the truncate modal still takes a while to open up

somangshu:
That does make sense @prsidhu 

@SatishGandham what's your suggestion

prsidhu:
We could catch the issue and suggest alternatives - pagination of api or using different widgets.

somangshu:
@prsidhu starting an internal async discussion, lets have more feedback before we can continue. 

I am in agreement with the user education bit

somangshu:
> How about showing a warning at higher threshold.

@SatishGandham I didn't get this part
- what error are we showing
- is it for the end user or the dev
- what's the trigger

SatishGandham:
@somangshu,
Lets say we show modal when there are 50 lines. If there are more than 100 lines, we show a warning on the modal in edit modal or a toast message.

- It's a warning, that it might break the app.
- Dev
- I think trigger is a wrong word, I'm talking about the threshold sizes we decide.

prsidhu:
Attempted a UI update for truncation. Thoughts? @Nikhil-Nandagopal @somangshu 

Screen Shot 2022-02-21 at 6 05 54 PM
Screen Shot 2022-02-21 at 6 08 44 PM

Nikhil-Nandagopal:
@prsidhu Why are we using an icon with ... and not relying on a natural truncation with `text-overflow: ellipsis;`?

somangshu:
@dilippitchika sorry about pushing this, Do we have a conclusion here from product

PR #11218, opened 2022-02-16
fix: Updated drop down control memo usage

### LINKED ISSUE #10762: [Feature]: Update the dynamic value fetch config to integrate post requests and query params

### Summary

This is regarding the new query editor forms' feature. This will allow for the plugin developers to pass in some static params and the datasource IDs.data source

This is a small issue blocking the release of entity selector and dynamic fetch value feature for UQI.

ayushpahwa:
## Description

We recently added memo to form control components to reduce the number of re-renders. This caused the functionality of dynamically fetching values. To overcome this, I connected the drop-down control directly to the redux state. This is a much better alternative than the initial method of passing everything down via the props and doing a deep diff on the complex props state. The selector is also called only if the particular option to enable dynamic value fetching is turned on for the component.

Fixes #10762 

## Type of change

## How Has This Been Tested?

- Test A
- Test B

## Checklist:

### DIFF
--- a/app/client/src/components/formControls/BaseControl.tsx
+++ b/app/client/src/components/formControls/BaseControl.tsx
@@ -1,10 +1,7 @@
 import { Component } from "react";
 import { ControlType } from "constants/PropertyControlConstants";
 import { InputType } from "components/constants";
-import {
-  ConditonalObject,
-  DynamicValues,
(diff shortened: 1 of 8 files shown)


Irongade (approved):
Kindly add more comments where necessary.

PR #11225, opened 2022-02-17
chore: Remove theme ids during git commit

abhvsn:
## Description

> This PR removes the theme ids during the git commit as this info is already present in `editModeTheme` and `publishedTheme`

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
@@ -189,17 +189,7 @@ public class ImportExportApplicationServiceCEImpl implements ImportExportApplica
 
                     // Refactor application to remove the ids
                     final String organizationId = application.getOrganizationId();
-                    application.setOrganizationId(null);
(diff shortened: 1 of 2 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -1553,4 +1543,20 @@ private Theme exportTheme(Theme srcTheme) {
...
+            application.setModifiedBy(null);
+            application.setUpdatedAt(null);
+            application.setLastDeployedAt(null);
+            application.setLastEditedAt(null);
+            application.setUpdatedAt(null);
+            application.setGitApplicationMetadata(null);
+            application.setPolicies(null);
+            application.setUserPermissions(null);
+            application.setEditModeThemeId(null);
+            application.setPublishedModeThemeId(null);
```
abhvsn, line 1560: For simple import-export this is not required as we have annotated this field with `JsonIgnore`, but will be required for git-sync as we are converting the document to object first and then serialise because we need some fields for git which are otherwise JsonIgnored

AnaghHegde (approved):

PR #11238, opened 2022-02-17
fix: Fixed test MongoDB datasource shows ok even if there is no database matching the name in the 'Default Database Name' input

### LINKED ISSUE #11214: [Bug]: Test MongoDB datasource shows ok even if there is no database matching the name in the 'Default Database Name' input

### Current Behavior

Test MongoDB datasource shows ok even if there is no database matching the name in the 'Default Database Name' input.
as mentioned in item number 4 in this comment https://github.com/appsmithorg/appsmith/issues/5005#issuecomment-1002032009

### Steps To Reproduce

1. Fill a valid MongoDB datasource 
2. Enter the name of a database that doesn't exist in MongoDB in the _Default Database Name_ input.
3. Click the _Test_ datasource button and it shows Ok instead of showing an error that the default database name does not exist.

### Environment

Release

### Version

Self-Hosted

mojtab23:
## Description

Test MongoDB datasource shows ok even if there is no database matching the name in the 'Default Database Name' input.

Fixes #11214

## Type of change

## How Has This Been Tested?

- Test case [`MongoPluginTest.testDatasourceFailWithInvalidDefaultDatabaseName()`](https://github.com/appsmithorg/appsmith/blob/979bd9b2d066f1ab028001fa8e91827569786698/app/server/appsmith-plugins/mongoPlugin/src/test/java/com/external/plugins/MongoPluginTest.java#L189)  is added
- Tested manually

## Checklist:

### DIFF
--- a/app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
+++ b/app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
@@ -897,7 +897,6 @@ public class MongoPlugin extends BasePlugin {
                                 .filter(d -> d.getString("name").trim().equals(defaultDatabaseName))
                                 .findFirst();
 
-                        log.debug("list databases:{}",document.toJson());
                         if (defaultDB.isEmpty()) {
                             // value entered in default database name is invalid
                             return Mono.just(new DatasourceTestResult("Default Database Name is invalid, no database found with this name."));
@@ -928,11 +927,7 @@ public class MongoPlugin extends BasePlugin {
                                                         "Please provide the default database name."));
 
                                     })
-                                    .flatMap(tuple2 -> {
-                                        Document document = tuple2.getT2();
-                                        log.debug("list collections:{}",document.toJson());
-                                        return Mono.just(tuple2.getT1());
-                                    })
+                                    .flatMap(tuple2 -> Mono.just(tuple2.getT1()))
                                     .doOnSuccess(MongoClient::close).then(Mono.just(new DatasourceTestResult()))
                                     .timeout(Duration.ofSeconds(TEST_DATASOURCE_TIMEOUT_SECONDS))
                                     .onErrorMap(TimeoutException.class, timeoutExceptionThrowableFunction)
--- a/app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
+++ b/app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
@@ -879,22 +879,24 @@ public class MongoPlugin extends BasePlugin {
                     "Connection timed out. Please check if the datasource configuration fields have " +
                             "been filled correctly."
             );
-            final String defaultDatabaseName = datasourceConfiguration.getConnection().getDefaultDatabaseName();
+
+            final String defaultDatabaseName;
+            if (datasourceConfiguration.getConnection() != null) {
+                defaultDatabaseName = datasourceConfiguration.getConnection().getDefaultDatabaseName();
+            } else defaultDatabaseName = null;
+
             return datasourceCreate(datasourceConfiguration)
                     .flatMap(mongoClient -> {
-                        Publisher<Document> result = mongoClient.getDatabase("admin").runCommand(new Document(
-                                "listDatabases", 1));
-                        return Mono.zip(Mono.just(mongoClient), Mono.from(result));
+                        final Publisher<String> result = mongoClient.listDatabaseNames();
+                        final Mono<List<String>> documentMono = Flux.from(result).collectList().cache();
+                        return documentMono.doFinally(ignored -> mongoClient.close()).then(documentMono);
                     })
-                    .flatMap(tuple2 -> {
-                        // close the connection
-                        tuple2.getT1().close();
-
-                        // this lambda checks if default database name is valid or not
-                        Document document = tuple2.getT2();
-                        List<Document> databases = document.getList("databases", Document.class);
-                        Optional<Document> defaultDB = databases.stream()
-                                .filter(d -> d.getString("name").trim().equals(defaultDatabaseName))
+                    .flatMap(names -> {
+                        if (defaultDatabaseName == null || defaultDatabaseName.isBlank()) {
+                            return Mono.just(new DatasourceTestResult());
+                        }
+                        final Optional<String> defaultDB = names.stream()
+                                .filter(name -> name.equals(defaultDatabaseName))
                                 .findFirst();
 
                         if (defaultDB.isEmpty()) {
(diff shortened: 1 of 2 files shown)


app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
```diff
@@ -874,23 +873,37 @@ private boolean isHostStringConnectionURI(Endpoint endpoint) {
...
+                        tuple2.getT1().close();
+
+                        // this lambda checks if default database name is valid or not
+                        Document document = tuple2.getT2();
+                        List<Document> databases = document.getList("databases", Document.class);
+                        Optional<Document> defaultDB = databases.stream()
+                                .filter(d -> d.getString("name").trim().equals(defaultDatabaseName))
+                                .findFirst();
+
+                        log.debug("list databases:{}",document.toJson());
```
nidhi-nair, line 900: Can we remove the logging here? We should only log things that will be required for debugging.

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
```diff
@@ -874,23 +873,37 @@ private boolean isHostStringConnectionURI(Endpoint endpoint) {
...
         @Override
         public Mono<DatasourceTestResult> testDatasource(DatasourceConfiguration datasourceConfiguration) {
-            final Consumer<Tuple2<MongoClient, Document>> closeClient = tuple -> tuple.getT1().close();
 
             Function<TimeoutException, Throwable> timeoutExceptionThrowableFunction = error -> new AppsmithPluginException(
                     AppsmithPluginError.PLUGIN_DATASOURCE_TIMEOUT_ERROR,
                     "Connection timed out. Please check if the datasource configuration fields have " +
                             "been filled correctly."
             );
+            final String defaultDatabaseName = datasourceConfiguration.getConnection().getDefaultDatabaseName();
```
nidhi-nair, line 882: Should we be using `datasourceConfiguration.authentication.databaseName` here instead of `datasourceConfigurations.connection.defaultDatabaseName`?

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
```diff
@@ -874,23 +873,37 @@ private boolean isHostStringConnectionURI(Endpoint endpoint) {
...
+                        Document document = tuple2.getT2();
+                        List<Document> databases = document.getList("databases", Document.class);
+                        Optional<Document> defaultDB = databases.stream()
+                                .filter(d -> d.getString("name").trim().equals(defaultDatabaseName))
+                                .findFirst();
+
+                        log.debug("list databases:{}",document.toJson());
+                        if (defaultDB.isEmpty()) {
+                            // value entered in default database name is invalid
+                            return Mono.just(new DatasourceTestResult("Default Database Name is invalid, no database found with this name."));
```
nidhi-nair, line 903: Should we not be throwing an error at this point?

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
```diff
@@ -915,7 +927,13 @@ private boolean isHostStringConnectionURI(Endpoint endpoint) {
...
                                                         "Please provide the default database name."));
 
-                                    }).doOnSuccess(closeClient).then(Mono.just(new DatasourceTestResult()))
+                                    })
+                                    .flatMap(tuple2 -> {
+                                        Document document = tuple2.getT2();
+                                        log.debug("list collections:{}",document.toJson());
+                                        return Mono.just(tuple2.getT1());
+                                    })
+                                    .doOnSuccess(MongoClient::close).then(Mono.just(new DatasourceTestResult()))
```
nidhi-nair, line 936: Will the doOnSuccess prevent the connection from closing if the default DB is empty?

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
```diff
@@ -904,8 +917,7 @@ private boolean isHostStringConnectionURI(Endpoint endpoint) {
                                 ((MongoCommandException) error).getErrorCodeName().equals("Unauthorized")) {
                             return datasourceCreate(datasourceConfiguration)
                                     .flatMap(mongoClient -> {
-                                        if (datasourceConfiguration.getConnection() != null && datasourceConfiguration.getConnection().getDefaultDatabaseName() != null) {
-                                            final String defaultDatabaseName = datasourceConfiguration.getConnection().getDefaultDatabaseName();
+                                        if (datasourceConfiguration.getConnection() != null && defaultDatabaseName != null) {
                                             Publisher<Document> result = mongoClient.getDatabase(defaultDatabaseName).runCommand(new Document(
                                                     "listCollections", 1).append("nameOnly", TRUE));
                                             return Mono.zip(Mono.just(mongoClient), Mono.from(result));
```
nidhi-nair, line 923: Instead of zipping the two, you can chain them right here, so you're only left with the connection at the end of this flatMap

nidhi-nair (changes requested):
Needs a few changes to the flow, also need confirmation on the field being manipulated.

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
mojtab23: Sure, I removed the logs.

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
mojtab23: The problem we try to solve here is to make sure that the _default database name_ field in datasource configuration is valid.  `datasourceConfiguration.authentication.databaseName` is used by mongoDB to verify the credentials, it might be usable if we want to entirely change the process for _testing datasource_. but with this flow I used the correct `databaseName`.

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
mojtab23: I've tested it manually and it's working as expected. what for replace do you have in mind?

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
mojtab23: I'm not sure, its behavior will be the same as before. You are right, there is no guarantee to call the client.close() if there is an error.

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
mojtab23: I'm not sure what to do here. can you help me?

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
nidhi-nair: Oh I see. In that case we should switch out the `admin` on L#885 with authentication database. I missed this in the previous change during review as well. This is my bad. :)

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
nidhi-nair: Can we use a `.then` in that case? This will make sure that the closing happens regardless of the result of the previous flatMap

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
```diff
@@ -904,8 +916,7 @@ private boolean isHostStringConnectionURI(Endpoint endpoint) {
                                 ((MongoCommandException) error).getErrorCodeName().equals("Unauthorized")) {
                             return datasourceCreate(datasourceConfiguration)
                                     .flatMap(mongoClient -> {
-                                        if (datasourceConfiguration.getConnection() != null && datasourceConfiguration.getConnection().getDefaultDatabaseName() != null) {
-                                            final String defaultDatabaseName = datasourceConfiguration.getConnection().getDefaultDatabaseName();
+                                        if (datasourceConfiguration.getConnection() != null && defaultDatabaseName != null) {
```
nidhi-nair, line 919: If getConnection can return a null, we will already have failed at L#882 where you are initializing `defaultDatabaseName`

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
nidhi-nair: How are we checking if result is valid here? That is, how do we know is `result` returned with an error?

PR #11242, opened 2022-02-17
chore: Save data of all performance runs

SatishGandham:
We are saving only the last performance run data when there are multiple runs.
This PR fixes it.


riodeuno (approved):
Looks good to me!

PR #11251, opened 2022-02-18
fix: Set comment authors email address as reply to field in comment notification email

### LINKED ISSUE #11016: [Bug]: The from email id is displayed wrong in the commenting emails

### Current Behavior

The wrong email id is displayed when user 1 tag user 2 and add a comment in the comment section
image

### Steps To Reproduce

1. Go to commenting feature
2. Tag Raksha@appsmith.com and add a new comment
3. Now when the user Raksha receives a email the receiver email is displayed as "hello@appsmith.com" instead of user email id

### Environment

Production

### Version

Cloud

nayan-rafiq:
## Description

When someone adds a comment to a thread, we send a email notification to the subscribers of that thread. The email has resply-to set to appsmith email address. If someone replies to that email, it's sent to an appsmith email address. This PR sets comment author's email address in the reply-to field so that replying to that email will be sent to the authors inbox.
Fixes #11016

## Type of change

## How Has This Been Tested?

- Unit tests
- Manually

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/notifications/EmailSender.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/notifications/EmailSender.java
@@ -38,11 +38,11 @@ public class EmailSender {
         REPLY_TO = makeReplyTo();
     }
 
-    public Mono<Boolean> sendMail(String to, String subject, String text) {
-        return sendMail(to, subject, text, null);
+    public Mono<Boolean> sendMail(String to, String subject, String text, Map<String, ? extends Object> params) {
+        return sendMail(to, subject, text, params, null);
     }
 
-    public Mono<Boolean> sendMail(String to, String subject, String text, Map<String, ? extends Object> params) {
+    public Mono<Boolean> sendMail(String to, String subject, String text, Map<String, ? extends Object> params, String replyTo) {
 
         /**
          * Creating a publisher which sends email in a blocking fashion, subscribing on the bounded elastic
@@ -59,7 +59,7 @@ public class EmailSender {
                     }
                 })
                 .doOnNext(emailBody -> {
-                    sendMailSync(to, subject, emailBody);
+                    sendMailSync(to, subject, emailBody, replyTo);
                 })
                 .subscribeOn(Schedulers.boundedElastic())
                 .subscribe();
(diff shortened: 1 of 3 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/notifications/EmailSender.java
```diff
@@ -37,11 +38,11 @@ public EmailSender(JavaMailSender javaMailSender, EmailConfig emailConfig) {
         REPLY_TO = makeReplyTo();
     }
 
-    public Mono<Boolean> sendMail(String to, String subject, String text) {
-        return sendMail(to, subject, text, null);
+    public Mono<Boolean> sendMail(String to, String subject, String text, Map<String, ? extends Object> params) {
+        return sendMail(to, subject, text, params, null);
```
sharat87, line 42: Looks like this signature is not used anywhere. Can we remove it, if it's unused code?

sharat87 (commented):
Minor refactor, otherwise looks good to be merged.

app/server/appsmith-server/src/main/java/com/appsmith/server/notifications/EmailSender.java
nayan-rafiq: No, the method with no replyTo is used in a lot of places. The one with replyTo is used in only the comment notification emails.

sharat87: Oh but the signature of this method is changing in this PR right? So then all those usages also should be updated otherwise we'll have a compilation error no?

nayan-rafiq: Understood. Let me explain the changes here: 
Earlier we had these two methods: 
1. `public Mono<Boolean> sendMail(String to, String subject, String text)` 
2. `public Mono<Boolean> sendMail(String to, String subject, String text, Map<String, ? extends Object> params)` 

The first method was used no where. It also called the second method with last param as null. I've removed the first method and added another one like the following:
3. `public Mono<Boolean> sendMail(String to, String subject, String text, Map<String, ? extends Object> params, String replyTo)` 

The second method now calls the new one with last param as null. The actual change is not reflected by the diff.

sharat87 (approved):

PR #11295, opened 2022-02-20
fix: curl import tries to guess the content-type from the body if content-type header is not specified

### LINKED ISSUE #7900: [Bug] when the header is not added to cURL command default must be empty
Steps to reproduce:
1) Add a cURL by removing a header 
2) Click on the import 
and observe

Observation:
It is observed that when the cURL is converted to API by default the header is considered as  "Form URL" even when the data sent is Json and retrieving of data is also not displayed properly. Even on changing the body selection, the data retrieving is. not appropriate 

Expectation:
When the header is not specified by the user we have to allow the user to specify and move the data body to align appropriately.
Possibly we can read the data imported to API and specify appropriate data section 

Video:
[LOOM DEMO](https://www.loom.com/share/534c91a990024889ae0b283975f7dc14)

mojtab23:
## Description

Before if the content-type was not specified and there was a body value, curl import (in the new datasource page) was treating the request as form-urlencoded, but now it tries to guess the content-type for the request.
it checks for `x-www-form-urlencoded` or `json` or sends it as raw without any content-type header.

Fixes #7900

## Type of change

## How Has This Been Tested?

- Tested manually
- Unit tests are fixed for it

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/test/java/com/appsmith/server/services/CurlImporterServiceTest.java
+++ b/app/server/appsmith-server/src/test/java/com/appsmith/server/services/CurlImporterServiceTest.java
@@ -311,33 +311,24 @@ public class CurlImporterServiceTest {
         );
         assertMethod(action, HttpMethod.POST);
         assertUrl(action, "http://loc");
-        assertEmptyBody(action);
-        assertBodyFormData(
-                action,
-                new Property("", "all of this exactly, but url encoded ")
-        );
+        assertBody(action,"all of this exactly, but url encoded ");
+        assertEmptyBodyFormData(action);
 
         action = curlImporterService.curlToAction(
                 "curl --data-urlencode 'spaced name=all of this exactly, but url encoded' http://loc"
         );
         assertMethod(action, HttpMethod.POST);
         assertUrl(action, "http://loc");
-        assertEmptyBody(action);
-        assertBodyFormData(
-                action,
-                new Property("spaced name", "all of this exactly, but url encoded")
-        );
+        assertBody(action,"spaced name=all of this exactly, but url encoded");
+        assertEmptyBodyFormData(action);
 
         action = curlImporterService.curlToAction(
                 "curl --data-urlencode 'awesome=details, all of this exactly, but url encoded' http://loc"
         );
         assertMethod(action, HttpMethod.POST);
         assertUrl(action, "http://loc");
-        assertEmptyBody(action);
-        assertBodyFormData(
-                action,
-                new Property("awesome", "details, all of this exactly, but url encoded")
-        );
+        assertBody(action,"awesome=details, all of this exactly, but url encoded");
+        assertEmptyBodyFormData(action);
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


app/server/appsmith-server/src/test/java/com/appsmith/server/services/CurlImporterServiceTest.java
```diff
@@ -311,33 +311,24 @@ public void testUrlEncodedData() throws AppsmithException {
         );
         assertMethod(action, HttpMethod.POST);
         assertUrl(action, "http://loc");
-        assertEmptyBody(action);
-        assertBodyFormData(
-                action,
-                new Property("", "all of this exactly, but url encoded ")
-        );
+        assertBody(action,"all of this exactly, but url encoded ");
```
leotom2000, line 314: Code formatting is expected, may be some automation tool can discover it later
Seems code formatting is not done, pls. do

app/server/appsmith-server/src/test/java/com/appsmith/server/services/CurlImporterServiceTest.java
```diff
@@ -311,33 +311,24 @@ public void testUrlEncodedData() throws AppsmithException {
...
                 "curl --data-urlencode 'spaced name=all of this exactly, but url encoded' http://loc"
         );
         assertMethod(action, HttpMethod.POST);
         assertUrl(action, "http://loc");
-        assertEmptyBody(action);
-        assertBodyFormData(
-                action,
-                new Property("spaced name", "all of this exactly, but url encoded")
-        );
+        assertBody(action,"spaced name=all of this exactly, but url encoded");
```
leotom2000, line 322: a space before and after = will be good, 
seems it is not going to affect the functionality
(in other similar places too )

app/server/appsmith-server/src/test/java/com/appsmith/server/services/CurlImporterServiceTest.java
```diff
@@ -311,33 +311,24 @@ public void testUrlEncodedData() throws AppsmithException {
...
                 "curl --data-urlencode 'awesome=details, all of this exactly, but url encoded' http://loc"
         );
         assertMethod(action, HttpMethod.POST);
         assertUrl(action, "http://loc");
-        assertEmptyBody(action);
-        assertBodyFormData(
-                action,
-                new Property("awesome", "details, all of this exactly, but url encoded")
-        );
+        assertBody(action,"awesome=details, all of this exactly, but url encoded");
```
leotom2000, line 330: space around equal,
leave it if affects the intended meaning

leotom2000 (commented):
Few Code formatting issues, check for similar issues in other places too you modified

app/server/appsmith-server/src/test/java/com/appsmith/server/services/CurlImporterServiceTest.java
mojtab23: I fixed this in the new commit.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/CurlImporterServiceTest.java
mojtab23: I can't do this because it's related to the code and business logic (form-urlencoded). I think it's better to keep it like this.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/CurlImporterServiceTest.java
mojtab23: I can't do this because it's related to the code and business logic (form-urlencoded). I think it's better to keep it like this.

leotom2000 (approved):

PR #11297, opened 2022-02-20
feat: Add new API to return list of default plugins

### LINKED ISSUE #11121: [Task]: Api to get list of plugins without having to specify an organisation id

### SubTasks

Currently an organisation id is required to get the list of plugins. The templates list view requires to show the datasources associated with it but this is not associated with any organisation initially. 
The frontend workaround currently is to get an organisation id where the user has permission to create an application and use the same to call the fetch plugins api.
The other alternative is to have an api which gets the list of all plugins without requiring to specify an organisation id.

nayan-rafiq:
## Description
Added a new API that'll return a list of plugins which are available to install. In other words, it'll return list of default/system plugins. This returns the package name, name and icon only - which will be used by templates to show datasource name and icons.

Fixes #11121 

## Type of change

## How Has This Been Tested?
- Manually
- JUnit test

## Checklist:


abhvsn (approved):
LGTM

PR #11341, opened 2022-02-22
fix: removing unnecessary paths to evaluate

### LINKED ISSUE #10799: JSON Form Performance improvements
When we have a large JSON object for the source data, we are facing a very high-performance degradation for the first time when the form renders. This performance hit seems to be coming from the evaluations.

The above logs are for a set of 150 odd fields with nested array and object sections.

ApekshaBhosale:
## Description
Removing internal paths from evaluations by making it 'NOOP' during translated diff creation. Paths that will be removed are 
all the paths which are inside dynamicBindingPathList, bindingPaths, validationPaths & triggerPaths

Fixes #10799 

## Type of change

## How Has This Been Tested?

- Test A
- Test B

## Checklist:

    // Code coverage diff between base branch:release and head branch: perf-removing-internal-paths 
 :green_circle: | total | 55.75 **(0.01)** | 37.08 **(0.01)** | 35.83 **(0.01)** | 56.11 **(0.01)**
 :green_circle: | app/client/src/workers/DataTreeEvaluator.ts | 77.39 **(0.13)** | 58.45 **(0.23)** | 75.58 **(0.29)** | 77.17 **(0.15)**
 :green_circle: | app/client/src/workers/evaluationUtils.ts | 58.73 **(0.36)** | 57.07 **(-0.07)** | 65.08 **(0.56)** | 58.13 **(0.43)**</details>


app/client/src/workers/evaluationUtils.ts
```diff
@@ -106,6 +106,16 @@ export const translateDiffEventToDataTreeDiffEvent = (
   if (!difference.path) {
     return result;
   }
+
+  //we do not need evaluations for these paths as these are internal paths
+  if (
```
hetunandu, line 111: This is not a well thought out change. Please think of a better long term solution and refactor to make this function generic. 
- What are the performance improvements by adding this change?
- Why not move these paths under `__evaluation__` so that we can just ignore a single key?
- Why does JSON form editor face this problem with these keys? Why are they getting evaluated because of that?
- Why not `dynamicTriggerPathList` ?

hetunandu (changes requested):

app/client/src/workers/evaluationUtils.ts
ApekshaBhosale: What is the consequence of this? 
1.  it reduces the number of paths to be evaluated hence the time taken is less
2.  why move them under __evaluation__? isn't that where we store evaluated data and errors but this is neither
3. JSON form widget has a schema. and each property in JSON is a widget so if there are 10 properties, it has 10 widgets and these 10 widgets have properties of their own. As a result of this, it's creating too many paths to be evaluated. JSON widget just increases the number of paths for evaluation. and if you check for each new diff we check references in other entities which again are too many. 
4. dynamicTriggerPathList is internal for us to create a dependency map etc. but what is the use of actually evaluating dynamicTriggerPathList? we don't evaluate or validate that path. eg - Table1 -> dynamicTriggerPathList - Table1.dynamicBindingPathList.primaryColumns.id.computedValue

Let me know if I am missing something

hetunandu: - I would like to see some numbers of how much of an impact does these paths have on performance
- We can move all these paths that do not need to be evaluated under `__evaluation__` to keep them all in the same place. That way you don't have to check the `indexOf` of multiple paths. indexOf is an O!n computation
- We might not be evaluating `dynamicTriggerPathList` today but we may end up coming in the differences in the future, since we will never need to evaluate it, it should also get a NOOP.
- This solution only solves for `updateDataTree`. `createFirstTree` will still evaluate all these paths, that means the first load would still be slower

ApekshaBhosale: - it used to take 7sec for updateDependencyMap for JSON widget it now takes 0.7 sec
- I made changes during translatedDiffs, not while eval
- createFirst does not eval all paths it evaluates dependency maps all paths and this does not contain in dependency map hence no problem during the first eval

ApekshaBhosale: cc @trishaanand

hetunandu: So, why can't we move these "path lists" inside a single key and remove them from the dataTree even before taking differences?

ApekshaBhosale: I am planning to create 2 different trees for eval and as meta tree. and this is what it's going to do remove them from eval and validation. That is a bigger change and till then we can ship as basic funda is same.

hetunandu: Is there an issue for this?

ApekshaBhosale: I will be creating the issue

ApekshaBhosale: https://github.com/appsmithorg/appsmith/issues/11351 i have created it. i will add more details to it @hetunandu

ApekshaBhosale: @hetunandu I haven't moved the change you suggested because we need differences to apply to the eval tree which goes for evaluation.

hetunandu (approved):

PR #11411, opened 2022-02-24
chore: Added zipy.ai for analytics

arunvjn:
## Description
Adds zippy.ai for analytics 

## Type of change
- New analytics tool

## How Has This Been Tested?
- Manual

## Checklist:


app/client/public/index.html
```diff
@@ -41,6 +41,19 @@
...
+    if(CLOUD_HOSTING && ZIPY_KEY) {
+      const script = document.createElement('script');
+      script.crossOrigin = "anonymous";
+      script.defer = true;
+      script.src = "https://cdn.zipy.ai/sdk/v1.0/zipy.min.umd.js";
+      script.onload = () => {
+        window.zipy && window.zipy.init(ZIPY_KEY);
+      }
+      const head = document.getElementsByTagName('head')[0];
+      head && head.appendChild(script);
```
sharat87, line 54: Wouldn't `document.head && document.head.appendChild(script)` work just as well here? 🤔

app/client/public/index.html
```diff
@@ -41,6 +41,19 @@
       return result;
     }
     const CLOUD_HOSTING = parseConfig("__APPSMITH_CLOUD_HOSTING__");
+    const ZIPY_KEY = parseConfig("__APPSMITH_ZIPY_SDK_KEY__");
+    if(CLOUD_HOSTING && ZIPY_KEY) {
+      const script = document.createElement('script');
```
sharat87, line 46: @SatishGandham, will adding a script like this here make this script a dependency for the `window.onload` event? If yes, should we wrap this code block inside a `window.addEventListener("load", () => {})`?

app/client/public/index.html
arunvjn: Older browsers do not support document.head (IE < 9). I am guessing we document.head should work for us since we don't support them anyway.

sharat87: Ah I see, did not know that. Thanks!

app/client/public/index.html
riodeuno: @sharat87 This won't be required, as the purpose of listening to the `load` event is to 
1. Make sure we complete the loader progress
2. Remove the reload flag, which when enabled reloads on chunk load failure. ( I am not so sure if this still works 🤔 )

riodeuno: Also, it is about time, we move our scripts to a different static `js` file. This index.html is getting unmanageable.

riodeuno (approved):

app/client/public/index.html
```diff
@@ -41,6 +41,19 @@
...
+      script.crossOrigin = "anonymous";
+      script.defer = true;
```
riodeuno, line 48: ❤️

sharat87 (approved):

PR #11414, opened 2022-02-24
feat: Provide a way to distinguish if the application is updated manually or through DB migration

### LINKED ISSUE #10644: [Feature]: When migrations take place this updates application resources, which results inaccurate tracking changes for git-sync

### Summary

When we run the migration during the release cycle more often than not we update resources attached to applications like pages, widget DSL, actions, JSObjects, etc. and users don't have any clue what migrations took place in a particular release. With the git-sync, we are providing a number of entities updated since the last commit to give more overview over updates to their apps, but with the migration we end up conveying incorrect updates.

This is defeating the purpose of displaying the number of resources updates since the last commit 

image

abhvsn:
## Description

> When we run the DB migration and then try to commit the application, we are not providing any context as to why some changes are getting reflected even if the user hasn't updated any resource attached to the application. This leads to confusion if the commit is safe. To address this we have decided to provide a friendly message which will give context to users that we have run the migrations between the commits

Fixes #10644

## Type of change

## How Has This Been Tested?

> JUnit TC
> Manual test

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
@@ -36,6 +36,7 @@ import com.appsmith.server.helpers.GitDeployKeyGenerator;
 import com.appsmith.server.helpers.GitFileUtils;
 import com.appsmith.server.helpers.GitUtils;
 import com.appsmith.server.helpers.ResponseUtils;
+import com.appsmith.server.migrations.JsonSchemaVersions;
 import com.appsmith.server.repositories.GitDeployKeysRepository;
 import com.appsmith.server.services.ActionCollectionService;
 import com.appsmith.server.services.AnalyticsService;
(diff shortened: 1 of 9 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -547,13 +548,19 @@ private String getReason() {
...
-                            application,
-                            application.getGitApplicationMetadata().getIsRepoPrivate()
-                    )
-                    .thenReturn(status);
+                    Application childApplication = tuple.getT2();
+                    // Update json schema versions so that we can detect if the next update was made by DB migration or
+                    // by the user
+                    Application update = new Application();
+                    update.setClientSchemaVersion(JsonSchemaVersions.clientVersion);
+                    update.setServerSchemaVersion(JsonSchemaVersions.serverVersion);
```
abhvsn, line 556: As we are only concerned with the migrations related to the resources which are going to be committed to git we are reutilizing `JsonSchemaVersions` again

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -1556,5 +1556,7 @@ private void removeUnwantedFieldsFromApplicationDuringExport(Application applica
             application.setUserPermissions(null);
             application.setEditModeThemeId(null);
             application.setPublishedModeThemeId(null);
+            application.setClientSchemaVersion(null);
+            application.setServerSchemaVersion(null);
```
sharat87, line 1560: Shouldn't we keep the version information in the exported application? Without this, how will we know the version during import, to know whether it's the latest schema or if it needs to be migrated? Or am I misunderstanding this change?

abhvsn: Yes that will still be present in the `ApplicationJson` which is being used as a DTO for import-export.

sharat87 (approved):

PR #11432, opened 2022-02-24
feat: Add API for delete branch for git connected application

### LINKED ISSUE #11272: [Feature]: Provide option to delete branch with in the Application

### Summary

Just like any git hosting provider we should allow the users to delete branch. 

Very basic and must needed feature for git
User will create multiple branches for testing developing branches and they would end up with stale branches

AnaghHegde:
## Description
Present the user has to delete the branch in the remote repo first and then come back to Appsmith UI and trigger the sync branches CTA. This is a bad UX and in some scenarios this CTA does not work as expected. 
This PR change set will solve this by exposing an API for the client for the delete action

Fixes #11272

## Type of change

## How Has This Been Tested?

- Locally
- TCs are on the way

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/GitControllerCE.java
```diff
@@ -207,5 +207,12 @@ public GitControllerCE(GitService service) {
                 .map(result -> new ResponseDTO<>((HttpStatus.OK.value()), result, null));
     }
 
+    @PostMapping("/branch/delete/{defaultApplicationId}")
```
abhvsn, line 210: Can we use DELETE method here with the same signature like `GET {baseUrl}/branch/defaultAppId`

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -2135,6 +2135,46 @@ else if (error.getMessage().contains("Nothing to fetch")) {
...
+    @Override
+    public Mono<Application> deleteBranch(String defaultApplicationId, String branchName) {
+        /**
+         *
+         * Cannont delete current branch
+         * Stale branch no app in db
+         * Branch does not exist in repo
+         * Branch with origin/
+         */
+        return getApplicationById(defaultApplicationId)
```
abhvsn, line 2147: Considering the fact that even if the application is not present in DB the branch might be available in local repo in some errorneos situations, this can be our way to clean up the repo and get both DB and local repo in sync again selectively

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -2135,6 +2135,46 @@ else if (error.getMessage().contains("Nothing to fetch")) {
...
+                                        .onErrorResume(throwable -> {
+                                            log.warn("Unable to find branch with name ", throwable);
+                                            return addAnalyticsForGitOperation(
+                                                    AnalyticsEvents.GIT_DELETE_BRANCH.getEventName(),
+                                                    application,
+                                                    throwable.getClass().getName(),
+                                                    throwable.getMessage(),
+                                                    gitApplicationMetadata.getIsRepoPrivate()
+                                            ).flatMap(application1 -> {
+                                                if( throwable instanceof GitAPIException ) {
```
abhvsn, line 2166: This block will actually never get executed in current scenario if I'm not mistaken because `applicationService.findByBranchNameAndDefaultApplicationId(branchName, defaultApplicationId, MANAGE_APPLICATIONS)` will always going to have `AppsmithException`

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: This is to get the defaultApplication so that the path to git repo can be formed before calling the git APIs. I am handling this scenario is L.2157

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: There is separate error handler for the gitAPI. My bad i missed removing the check for GitAPIException.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -2135,6 +2135,41 @@ else if (error.getMessage().contains("Nothing to fetch")) {
...
+                                        .onErrorResume(throwable -> {
+                                            log.warn("Unable to find branch with name ", throwable);
+                                            return addAnalyticsForGitOperation(
+                                                    AnalyticsEvents.GIT_DELETE_BRANCH.getEventName(),
+                                                    application,
+                                                    throwable.getClass().getName(),
+                                                    throwable.getMessage(),
+                                                    gitApplicationMetadata.getIsRepoPrivate()
+                                            ).flatMap(application1 -> Mono.just(application1));
+                                        })
```
abhvsn, line 2166: If we are not throwing the exception here would this be passed on to the subsequent chain and delete the default application altogether?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -2135,6 +2135,41 @@ else if (error.getMessage().contains("Nothing to fetch")) {
...
+                                        .flatMap(applicationPageService::deleteApplicationByResource);
+                            });
+                })
+                .flatMap(application -> addAnalyticsForGitOperation(AnalyticsEvents.GIT_DELETE_BRANCH.getEventName(), application, application.getGitApplicationMetadata().getIsRepoPrivate()));
```
abhvsn, line 2170: Can we please sanitise the resources from `ResponseUtils` so that the client will have the `defaultResourceIds` always

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/GitControllerCE.java
```diff
@@ -207,5 +208,12 @@ public GitControllerCE(GitService service) {
                 .map(result -> new ResponseDTO<>((HttpStatus.OK.value()), result, null));
     }
 
+    @DeleteMapping("/branch/delete/{defaultApplicationId}")
```
abhvsn, line 211: We don't require the delete keyword in the URL as this is a DELETE method only

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -2135,6 +2135,42 @@ else if (error.getMessage().contains("Nothing to fetch")) {
...
+         * Branch does not exist in repo
+         * Branch with origin/
+         */
+        return getApplicationById(defaultApplicationId)
+                .flatMap(application -> {
+                    GitApplicationMetadata gitApplicationMetadata = application.getGitApplicationMetadata();
+                    Path repoPath = Paths.get(application.getOrganizationId(), defaultApplicationId, gitApplicationMetadata.getRepoName());
+                    return gitExecutor.deleteBranch(repoPath, branchName)
+                            .flatMap(gitBranch -> {
+                                if(gitBranch.equals(Boolean.FALSE)) {
```
abhvsn, line 2153: Nit: Instead we should use `Boolean.FALSE.equals(gitBranch)`. Also can we please rename the variable to `isBranchDeleted` so as to convey this is a boolean

abhvsn (approved):
Minor nits but LGTM

PR #11490, opened 2022-02-28
fix: filter empty/undefined values  in Multipart

### LINKED ISSUE #10877: [Feature]: Do not send undefined values in multipart

### Summary

When making a JSON request, it would be common to send an `undefined` value for a key. Servers would know to ignore it. For example in a PATCH operation, `undefined` would mean do not make a change, while `null` would indicate to set the value as `null`.

When doing multipart there is no such distinction, as values are either text, either binary. Therefore we should prevent from sending any data for a given key if the value is `undefined`, as sending an empty value for the key (or worse the string `undefined`) does not look like the proper behavior here.

Also currently if the field type is set as `file`, and the value `undefined`. The internal-server would log the following line.
```
The connection observed an error
com.appsmith.external.exceptions.pluginExceptions.AppsmithPluginException: Unable to parse content. Expected to receive an array or object of multipart data
```

This would increase the ease of use of the application with multipart APIs

danieldare:
## Description

FIlters out empty/undefined data from the bodyFormData object.

Fixes #10877

## Type of change

## How Has This Been Tested?
Manually 

## Checklist:

    // Code coverage diff between base branch:release and head branch: feat/undefined-values-in-multipart 
 :red_circle: | total | 55.57 **(0)** | 36.82 **(-0.01)** | 34.86 **(0)** | 55.87 **(0)**
 :green_circle: | app/client/src/sagas/ActionSagas.ts | 12.73 **(0.4)** | 1.49 **(-0.01)** | 8 **(0)** | 15.69 **(0.51)**
 :red_circle: | app/client/src/transformers/RestActionTransformer.ts | 96.77 **(0)** | 85.29 **(-0.82)** | 100 **(0)** | 100 **(0)**</details>


app/client/src/sagas/ActionSagas.ts
```diff
@@ -308,7 +310,9 @@ export function* updateActionSaga(
     if (!action) throw new Error("Could not find action to update");
 
     if (isAPIAction(action)) {
-      action = transformRestAction(action);
+      // Redux form returns the form values unsantinized, transformRestAction then transforms the action
+      const apiFormAction = yield select(getFormValues(API_EDITOR_FORM_NAME));
```
AmanAgarwal041, line 314: Adding a specific form type `API_EDITOR_FOR_NAME` for the form values will make it difficult to track conditions for newer forms. Any reason this was added ?

danieldare: For newer actions we can add for example . `isGRAPHQLAction` to check for specific its use case. 

The new line was added to get the latest form values from the redux form.

Irongade (dismissed):

PR #11501, opened 2022-02-28
chore: Focus widget name in property pane.

### LINKED ISSUE #11199: [Task]: Opening property pane should focus the widget name
From @somangshu 
> When you select a widget you need to use the mouse to start with keyboard navigation, I believe we have to reduce this click simply by focusing on the widget name as soon as the widget is selected.

Final decision:
- Focus on the first property whenever we click on the widget.
- When we add a new widget or when we paste a widget, focus on the widget name. (it should be editable only if we press space or enter otherwise it will be hard to other operations like delete, copy, paste etc,)

yaldram:
## Description
When you select a widget you need to use the mouse to start with keyboard navigation, we have to reduce this click simply by focusing on the widget name as soon as the widget is selected.

Fixes #11199

## Type of change

## How Has This Been Tested?

- Tested Manually on the Canvas

## Checklist:

    // Code coverage diff between base branch:release and head branch: chore/focus-widget-name 
 :green_circle: | total | 55.72 **(0)** | 37.06 **(0.01)** | 35.79 **(0)** | 56.08 **(0)**
 :green_circle: | app/client/src/pages/Editor/PropertyPaneTitle.tsx | 74.58 **(0.9)** | 37.5 **(2.5)** | 75 **(2.27)** | 71.7 **(1.11)**</details>

### DIFF
--- a/app/client/src/pages/Editor/PropertyPaneTitle.test.tsx
+++ b/app/client/src/pages/Editor/PropertyPaneTitle.test.tsx
@@ -9,21 +9,30 @@ import { Provider } from "react-redux";
 import store from "../../store";
 
 describe("<PropertyPaneTitle />", () => {
-  it("should focus when f2 is pressed", async () => {
-    const getTestComponent = () => (
-      <Provider store={store}>
-        <ThemeProvider theme={lightTheme}>
-          <PropertyPaneTitle
-            actions={[]}
-            isPanelTitle
-            // title="test"
-            widgetId="1"
-          />
-        </ThemeProvider>
-      </Provider>
-    );
+  const getTestComponent = () => (
+    <Provider store={store}>
+      <ThemeProvider theme={lightTheme}>
+        <PropertyPaneTitle
+          actions={[]}
+          isPanelTitle
+          title="test"
+          widgetId="1"
+        />
+      </ThemeProvider>
+    </Provider>
+  );
+
+  it("should focus on the component mount", () => {
     const component = getTestComponent();
     const renderResult = render(component);
+    expect(renderResult.container.querySelector("input")).toHaveFocus();
+  });
+
+  it("should focus when f2 is pressed", async () => {
+    const component = getTestComponent();
+    const renderResult = render(component);
+    await userEvent.keyboard("{Tab}");
+    await userEvent.keyboard("{Tab}");
     await userEvent.keyboard("{F2}");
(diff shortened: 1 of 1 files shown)


app/client/src/pages/Editor/PropertyPaneTitle.test.tsx
```diff
@@ -9,21 +9,30 @@ import { Provider } from "react-redux";
...
-          />
-        </ThemeProvider>
-      </Provider>
-    );
     const component = getTestComponent();
     const renderResult = render(component);
+    await userEvent.keyboard("{Tab}");
+    await userEvent.keyboard("{Tab}");
     await userEvent.keyboard("{F2}");
     expect(renderResult.container.querySelector("input")).toBeVisible();
```
aswathkk, line 37: This should've been `toHaveFocus()` rather than `toBeVisible()`

app/client/src/pages/Editor/PropertyPaneTitle.test.tsx
```diff
@@ -9,21 +9,30 @@ import { Provider } from "react-redux";
...
-            // title="test"
-            widgetId="1"
-          />
-        </ThemeProvider>
-      </Provider>
-    );
     const component = getTestComponent();
     const renderResult = render(component);
+    await userEvent.keyboard("{Tab}");
+    await userEvent.keyboard("{Tab}");
```
aswathkk, line 35: I found that the input is having focus even if we do `{Tab}` twice and not doing `{F2}`.

A better thing to do would be do `renderResult.container.querySelector("input")?.blur();`

aswathkk (changes requested):
The test needs some change.
Also, no need to use await while firing `userEvent`

Otherwise, looks good to me!

app/client/src/pages/Editor/PropertyPaneTitle.test.tsx
yaldram: Actually when we click `Tab` we focus on the icons if we `Shift + Tab` we now foucs on the widget not the `widget name input` and can therefore copy and delete the widget.

app/client/src/pages/Editor/PropertyPaneTitle.test.tsx
yaldram: Thanks will change this.

PR #11531, opened 2022-03-01
feat: Apply env variables in NGINX config before startup

sharat87:
Instead of doing env substitutions at request processing time, do it before NGINX even starts.

This is a step towards making our static files more independent of NGINX, and potentially towards having them on S3 and/or by a CDN.

The NGINX config that is used during dev-time still uses `sub_filter`, since the compilation doesn't produce files there for us to apply substitutions. However, since we fallback to `process.env.VARNAME` in the client code, `sub_filter` shouldn't be needed there anyway. I need to test this theory and verify though.

## Type of change

- New feature, non-user-facing (non-breaking change which adds functionality)

## How Has This Been Tested?

- Manually.


deploy/docker/scripts/run-nginx.sh
```diff
@@ -1,28 +1,40 @@
...
+import re
+import os
+index_path = Path("/opt/appsmith/editor/index.html")
+content = re.sub(
+	r"\b__(APPSMITH_[A-Z0-9_]+?)__\b",
+	lambda match: [print(match.group(0), os.environ.get(match.group(1), None)), os.environ.get(match.group(1), match.group(0))][-1],
+	index_path.read_text(),
+)
+index_path.write_text(content)
+'
```
sharat87, line 38: I'm doing the substitution in Python here, since doing this with `sed` means too many special characters to be escaped and potentially cause problems. This felt safer and easier for this. Let me know if we should prefer shell utils for this.

sharat87: Disregard. Ported to NodeJS instead.

deploy/docker/scripts/run-nginx.sh
```diff
@@ -1,28 +1,37 @@
...
 
-echo "Re-generating nginx config template"
-bash "$APP_TEMPLATE" "$APPSMITH_CUSTOM_DOMAIN" >"/etc/nginx/conf.d/nginx_app.conf.template"
+bash "$APP_TEMPLATE" "$APPSMITH_CUSTOM_DOMAIN" > /etc/nginx/sites-available/default
 
-echo "Generating nginx configuration"
-cat /etc/nginx/conf.d/nginx_app.conf.template | envsubst "$(printf '$%s,' $(env | grep -Eo '^APPSMITH_[A-Z0-9_]+'))" | sed -e 's|\${\(APPSMITH_[A-Z0-9_]*\)}||g' >/etc/nginx/sites-available/default
+node -e '
+const fs = require("fs")
+const indexPath = "/opt/appsmith/editor/index.html"
```
mohanarpit, line 29: This means that all env variables will only be replaced in the index.html. We need to be careful to not include any environment variables in any other file.

mohanarpit (approved):

PR #11534, opened 2022-03-01
chore: Migrations to enable JSON and form mode

### LINKED ISSUE #11240: [Task]: Modify form/json migration to include S3 delete many

### SubTasks

Since the S3 plugin has had another command implemented before the form/json changes went in, that command will also have to be added to the migration.

nidhi-nair:
Had to redo this since my previous changes got overridden. Have merged the JSON migration in this PR as well to avoid redundancy. Will keep this branch separate to avoid any overwrite again.

Also has the changes for all server data structure updates that have happened since the last migration was written. (S3 delete many, S3 paginate and sort, S3 deprecating unused data points)

Fixes #11240 

    // Code coverage diff between base branch:release and head branch: fix/form-data-server-migration 
 :green_circle: | app/client/src/utils/WorkerUtil.ts | 89.76 **(0)** | 72.55 **(1.96)** | 100 **(0)** | 93.33 **(0)**</details>


nidhi-nair:
Moving to draft for two reasons:

1. Needs client to go in first.

2. From a conversation with @sumitsum , we either need to figure out a way to retain default values for components when switching commands, or needs to lose the ability of sharing data between commands so that the initial value of components makes sense at all.

sumitsum:
@nidhi-nair can you please share any writeup / link that has an example of both what the older schema was and what the new schema looks like - for future reference and debugging.

sumitsum (dismissed):

Issue #11223, opened 2022-02-17
[Feature]: Provide a way to modify the default branch for git


abhvsn:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

Today if user wants to update the default branch and delete the older branch there is no way to do that from Appsmith. Even though user can modify the same on remote the same will not be reflected in Appsmith UI. Also, deleting the default branch is not allowed today in Appsmith because we have default branch ids referenced nested in each resource.   

### Why should this be worked on?

This will be useful in scenarios when the default branch ended up in bad state due and user is unable to revert these changes somehow.

hiteshjoshi:
We will need to communicate with the user about git default branch and changing that will also change your default branch inside the App. Syncing branch after git default branch change is necessary.

hiteshjoshi:
- [ ] Better communication for updates on the default branch after sync etc.

AnaghHegde:
Issues covered part of this 
- #11651 
- #11541

AnaghHegde:
Closed via https://github.com/appsmithorg/appsmith/pull/11549

PR #11549, opened 2022-03-02
fix: Add logic to change default branch for git connected application

### LINKED ISSUE #11541: [Feature]: For git connected application always open the default branch application in edit mode

### Summary

We are opening the appsmith default application when the user clicks on edit. But when the app is connected to git we should always bring up the default application of the default branch.

We are opening the appsmith default application when the user clicks on edit. But when the app is connected to git we should always bring up the default application of the default branch.

AnaghHegde:
## Description
We are opening the appsmith default application when the user clicks on edit. But when the app is connected to git we should always bring up the default application of the default branch.

Fixes #11541 #11223 #11651 #12931 

## Type of change

## How Has This Been Tested?

- Locally
- Junit tests

## Checklist:

    // Code coverage diff between base branch:release and head branch: feat/fetch-app-by-default-branch 
 :red_circle: | total | 56.49 **(0)** | 37.98 **(0)** | 36.27 **(0)** | 56.7 **(-0.01)**
 :green_circle: | app/client/src/utils/WorkerUtil.ts | 89.76 **(0.78)** | 72.55 **(3.92)** | 100 **(0)** | 93.33 **(0.95)**</details>


abhvsn:
I did not follow this can you please elaborate on this further.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomApplicationRepositoryCEImpl.java
```diff
@@ -190,6 +196,24 @@ protected Criteria getIdCriteria(Object id) {
         return queryAll(List.of(organizationIdCriteria, repoCriteria, gitAuthCriteria), AclPermission.MANAGE_APPLICATIONS);
     }
 
+    @Override
+    public Mono<Application> getApplicationByDefaultApplicationIdAndDefaultBranch(String defaultApplicationId) {
+        String gitApplicationMetadata = fieldName(QApplication.application.gitApplicationMetadata);
+
+        Criteria applicationIdCriteria = where(gitApplicationMetadata + "." + fieldName(QApplication.application.gitApplicationMetadata.defaultApplicationId)).is(defaultApplicationId);
+        Criteria deletionCriteria = where(fieldName(QApplication.application.deleted)).ne(true);
+        boolean branchCriteria = where("this." + gitApplicationMetadata + "." + fieldName(QApplication.application.gitApplicationMetadata.branchName))
```
abhvsn, line 205: What is the use of `this.`, can you please point me to the right documentation as I'm not aware of this yet

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomApplicationRepositoryCEImpl.java
```diff
@@ -190,6 +196,24 @@ protected Criteria getIdCriteria(Object id) {
...
+                .equals("this." +gitApplicationMetadata + "." + fieldName(QApplication.application.gitApplicationMetadata.defaultBranchName));
+
+        Query query = new Query();
+        query.addCriteria(where(gitApplicationMetadata + "." + fieldName(QApplication.application.gitApplicationMetadata.defaultApplicationId)).is(defaultApplicationId));
+        query.addCriteria(where(fieldName(QApplication.application.deleted)).ne(true));
+        query.equals(where("this." + gitApplicationMetadata + "." + fieldName(QApplication.application.gitApplicationMetadata.branchName))
+                .equals("this." +gitApplicationMetadata + "." + fieldName(QApplication.application.gitApplicationMetadata.defaultBranchName)));
+
+        return mongoOperations.findOne(query,Application.class);
```
abhvsn, line 214: Whay are we not using the `queryOne` method here?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationServiceCEImpl.java
```diff
@@ -111,9 +112,21 @@ public ApplicationServiceCEImpl(Scheduler scheduler,
...
+    public Mono<Application> findByIdAndBranchName(String defaultApplicationId, String branchName) {
+            return findByBranchNameAndDefaultApplicationId(branchName, defaultApplicationId, READ_APPLICATIONS)
+                    // Check if the Appsmith application and the default branch are same
+                    .flatMap(application -> {
+                        if (StringUtils.isEmpty(branchName)) {
+                            if(Optional.ofNullable(application.getGitApplicationMetadata()).isEmpty()
+                                    || application.getGitApplicationMetadata().getBranchName().equals(application.getGitApplicationMetadata().getDefaultBranchName())) {
+                                return Mono.just(application);
+                            }
+                            return repository.getApplicationByGitBranchAndDefaultApplicationId(defaultApplicationId, application.getGitApplicationMetadata().getDefaultBranchName(), READ_APPLICATIONS)
```
abhvsn, line 124: Do we want to query here for `getApplicationByDefaultApplicationIdAndDefaultBranch`? Also why are we looking for application with defaultBranch when `branchName` is empty? Can you please elaborate what is going on here?

abhvsn (commented):
Some questions:
1. Have we tested the flow for already available applications where defaultBranch in remote is different than what we have in DB? Asking this because there are certain fields that we only store in the `defaultApplication` but with this implementation, we are requesting the `remoteDefaultBranchedApplication` always
2. Should we apply migration so that applications where defaultApplication and remote default branch are not same and insert the fields which are present in the gitApplicationMetadata of DB default application like gitAuth etc?
3. When we are running the sync flow are we creating the `remoteDefaultBranchedApplication` if the default branch is altered from remote?
4. Are we maintaining the `gitApplicationMetadata` for both `DBDefaultApplication` and `remoteBranchedDefaultApplication` henceforth? Is there any need for this, asking this because we have `findApplicationByDefaultApplicationIdAndBranchName` method present in quite a lot of places which needs to be addressed

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomApplicationRepositoryCEImpl.java
AnaghHegde: This is to compare two fields in the same collection where the query is being run. There are two ways to do this from v3.6+ one is to use aggregate with $expr inside the pipeline. Other one is to use this in the find query. Since this was a simple check on branchName and defaultBranchName i didn't feel the need to add a aggregate query here. 
https://docs.mongodb.com/v4.0/reference/operator/query/expr/
https://stackoverflow.com/questions/4442453/mongodb-query-condition-on-comparing-2-fields

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomApplicationRepositoryCEImpl.java
AnaghHegde: Query one internally calls the findOne. And the reason is i cannot achieve the expected output via queryOne as the way it forms the query object via List of criteria.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationServiceCEImpl.java
AnaghHegde: This is because the client does not send the branchName on the first call in edit mode. This is to determine that condition and fetch the defaultBranchName application on the load.

AnaghHegde:
> Some questions:
> 
> 1. Have we tested the flow for already available applications where defaultBranch in remote is different than what we have in DB? Asking this because there are certain fields that we only store in the `defaultApplication` but with this implementation, we are requesting the `remoteDefaultBranchedApplication` always
> 2. Should we apply migration so that applications where defaultApplication and remote default branch are not same and insert the fields which are present in the gitApplicationMetadata of DB default application like gitAuth etc?
> 3. When we are running the sync flow are we creating the `remoteDefaultBranchedApplication` if the default branch is altered from remote?
> 4. Are we maintaining the `gitApplicationMetadata` for both `DBDefaultApplication` and `remoteBranchedDefaultApplication` henceforth? Is there any need for this, asking this because we have `findApplicationByDefaultApplicationIdAndBranchName` method present in quite a lot of places which needs to be addressed

1. Yes this tested on the scenarios you are referring to.
2. No we do not need migration here, as the logic takes care of updating the these fields on refresh/sync workflow. On the change in defaultBranch all the details needed are copied to the respective application.
3. Yes we are doing that. There is oneError handler at L.1574 which is doing the same
4. I am not fully confident on removing the gitApplicationMetadata from the defaultBranch app yet, for the same reason you mentioned. But i don't see any harm in doing so as the majority cases are in GitService class and that is handled already via the newly created method in ApplicationService to handle this scenario.

AnaghHegde:
@abhvsn 
1. Yes this tested on the scenarios you are referring to.
2. No we do not need migration here, as the logic takes care of updating the these fields on refresh/sync workflow. On the change in defaultBranch all the details needed are copied to the respective application.
3. Yes we are doing that. There is oneError handler at L.1574 which is doing the same
4. I am not fully confident on removing the gitApplicationMetadata from the defaultBranch app yet, for the same reason you mentioned. But i don't see any harm in doing so as the majority cases are in GitService class and that is handled already via the newly created method in ApplicationService to handle this scenario.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationServiceCEImpl.java
abhvsn: IMO we should update the client-side implementation for this instead of doing this hack. This might confuse folks working on this feature in the future.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1295,7 +1295,7 @@ private String getReason() {
     }
 
     Mono<Application> getApplicationById(String applicationId) {
-        return applicationService.findById(applicationId, MANAGE_APPLICATIONS)
+        return applicationService.getApplicationByDefaultApplicationIdAndDefaultBranch(applicationId)
```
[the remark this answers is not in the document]
AnaghHegde: This would make sense if we go with older approach. I have updated logic per latest approach.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomApplicationRepositoryCEImpl.java
```diff
@@ -190,6 +196,24 @@ protected Criteria getIdCriteria(Object id) {
...
+        Criteria deletionCriteria = where(fieldName(QApplication.application.deleted)).ne(true);
+        boolean branchCriteria = where("this." + gitApplicationMetadata + "." + fieldName(QApplication.application.gitApplicationMetadata.branchName))
+                .equals("this." +gitApplicationMetadata + "." + fieldName(QApplication.application.gitApplicationMetadata.defaultBranchName));
+
+        Query query = new Query();
+        query.addCriteria(where(gitApplicationMetadata + "." + fieldName(QApplication.application.gitApplicationMetadata.defaultApplicationId)).is(defaultApplicationId));
+        query.addCriteria(where(fieldName(QApplication.application.deleted)).ne(true));
+        query.equals(where("this." + gitApplicationMetadata + "." + fieldName(QApplication.application.gitApplicationMetadata.branchName))
+                .equals("this." +gitApplicationMetadata + "." + fieldName(QApplication.application.gitApplicationMetadata.defaultBranchName)));
+
```
abhvsn, line 213: Why are we duplicating the criteria here?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1407,46 +1407,28 @@ private String getReason() {
...
-                                        .onErrorResume(throwable -> {
-                                            log.warn(" No application exists in DB for the local branch of file system", throwable);
-                                            return Mono.empty();
-                                        })
-                                        .then(gitExecutor.deleteBranch(repoPath, gitBranch)))
-                                .then(Mono.just(gitBranchListDTOS));
-
                         if(defaultBranchRemote.equals(dbDefaultBranch)) {
-                            return monoBranchList.zipWith(Mono.just(application));
+                            return Mono.just(gitBranchListDTOS).zipWith(Mono.just(application));
```
abhvsn, line 1411: Are we not pruning the branches in DB if both DB and local repo have the same default branch?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1407,46 +1407,28 @@ private String getReason() {
...
+                                    .onErrorResume(throwable -> checkoutRemoteBranch(defaultApplicationId, defaultBranchRemote))
+                                    // Copy the gitApplicationMetadata to the app
+                                    .flatMap(application1 -> {
+                                        application1.setGitApplicationMetadata(gitApplicationMetadata);
+                                        application1.getGitApplicationMetadata().setBranchName(defaultBranchRemote);
+                                        application1.getGitApplicationMetadata().setDefaultBranchName(defaultBranchRemote);
+                                        return applicationService.save(application1);
+                                    })
+                                    // Update the default branch name in all the branched applications
+                                    .flatMap(application1 -> applicationService.findAllApplicationsByDefaultApplicationId(defaultApplicationId)
```
abhvsn, line 1425: I don't think that we are storing the default branch in each application. Just the default application's gitApplicationMetadata has this info so no need to update this

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomApplicationRepositoryCEImpl.java
AnaghHegde: I was testing and forgot to remove the unused part. My bad

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: No with the delete branchs, it does not make sense to delete branch's in the sync flow. To keep it simple lets just delete the branch in the new delete branch API.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: I will add a migration for this

AnaghHegde: That makes sense. I had added this logic in the past where the idea was to delete the Appsmith default Application as well.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
abhvsn: I still think this can be useful in scenarios where user has lot of stale branches, but let us present this change in a bigger forum as I'm not aware if we are making this change from product point of view.

abhvsn: @AnaghHegde this PR is pending as we have made changes in the sync branch flows. Can we please add this change as a separate PR and let's merge this one? Asking this because both these changes are independent of each other and we are blocked on frontend to roll out delete branch feature

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1409,46 +1409,22 @@ public Mono<List<GitBranchDTO>> listBranchForApplication(String defaultApplicati
...
-                            return monoBranchList.zipWith(applicationService.save(application));
+                            // Get the application from DB by new defaultBranch name
+                            return applicationService.findByBranchNameAndDefaultApplicationId(defaultBranchRemote, defaultApplicationId, MANAGE_APPLICATIONS)
+                                    // Check if the branch is already present, If not follow checkout remote flow
+                                    .onErrorResume(throwable -> checkoutRemoteBranch(defaultApplicationId, defaultBranchRemote))
+                                    // Copy the gitApplicationMetadata to the app
+                                    .flatMap(application1 -> {
+                                        application1.setGitApplicationMetadata(gitApplicationMetadata);
+                                        application1.getGitApplicationMetadata().setBranchName(defaultBranchRemote);
+                                        application1.getGitApplicationMetadata().setDefaultBranchName(defaultBranchRemote);
```
abhvsn, line 1423: Are we storing the default branchName whenever user updates the default branch? Should we keep the defaultBranch is the defaultApplicationMetadata only, otherwise we will have to update previous defaultBranched application gitMetadata.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -2071,7 +2047,7 @@ public Mono<Application> deleteBranch(String defaultApplicationId, String branch
                     GitApplicationMetadata gitApplicationMetadata = application.getGitApplicationMetadata();
                     Path repoPath = Paths.get(application.getOrganizationId(), defaultApplicationId, gitApplicationMetadata.getRepoName());
                     if(branchName.equals(gitApplicationMetadata.getDefaultBranchName())) {
-                        return Mono.error(new AppsmithException(AppsmithError.GIT_ACTION_FAILED, " delete branch", "Cannot delete default branch"));
+                        return Mono.error(new AppsmithException(AppsmithError.GIT_ACTION_FAILED, " delete branch.", " Cannot delete default branch"));
```
abhvsn, line 2050:
```suggestion
                        return Mono.error(new AppsmithException(AppsmithError.GIT_ACTION_FAILED, "delete branch", " Cannot delete default branch"));
```

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ApplicationFetcherCEImpl.java
```diff
@@ -113,8 +113,12 @@ public Mono<UserHomepageDTO> getAllApplications() {
                     Flux<Application> applicationFlux = applicationRepository
                             .findByMultipleOrganizationIds(orgIds, READ_APPLICATIONS)
                             .filter(application -> application.getGitApplicationMetadata() == null
-                                    || (StringUtils.equals(application.getId(), application.getGitApplicationMetadata().getDefaultApplicationId()))
+                                    || (!Optional.ofNullable(application.getGitApplicationMetadata().getDefaultBranchName()).isEmpty()
```
abhvsn, line 116:
```suggestion
                                    || (!StringUtils.isEmpty(application.getGitApplicationMetadata().getDefaultBranchName()))
```

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ApplicationFetcherCEImpl.java
```diff
@@ -113,8 +113,12 @@ public Mono<UserHomepageDTO> getAllApplications() {
...
+                                    && application.getGitApplicationMetadata().getBranchName().equals(application.getGitApplicationMetadata().getDefaultBranchName()))
                             )
+                            /*.filter(application -> application.getGitApplicationMetadata() == null
+                                    || (StringUtils.equals(application.getId(), application.getGitApplicationMetadata().getDefaultApplicationId()))
+                            )*/
```
abhvsn, line 121: Can we please remove the commented code here if not needed

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
abhvsn: I wanted to keep the data confined to `defaultApplicationMetadata` because there are certain other fields except `defaultBranchName` like `isRepoPrivate` which is currently stored only in defaultApplicationGitMetadata object so that whenever we receive the changes from remote we only update the defaultApplicationMetadata

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ApplicationServiceTest.java
```diff
@@ -2546,4 +2547,37 @@ public void cloneApplication_WithCustomSavedTheme_ThemesAlsoCopied() {
...
+
+
+        Mono<Application> getApplication = applicationService.findByIdAndBranchName(gitConnectedApp.getId(), "release");
+        StepVerifier.create(getApplication)
+                .assertNext(application1 -> {
+                    assertThat(application1).isNotNull();
+                    assertThat(application1.getGitApplicationMetadata().getBranchName()).isNotEqualTo(gitConnectedApp.getGitApplicationMetadata().getBranchName());
+                    assertThat(application1.getGitApplicationMetadata().getBranchName()).isEqualTo(application.getGitApplicationMetadata().getBranchName());
+                    assertThat(application1.getId()).isEqualTo(gitConnectedApp.getId());
+                    assertThat(application1.getName()).isEqualTo(application.getName());
```
abhvsn, line 2578: Should we test this flow with listBranch API as that is the place where we are altering the defaultBranchName 🤔

abhvsn:
If possible can you also add a testcase for getAllApplications `/new` API?

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ApplicationServiceTest.java
AnaghHegde: Checking this here would be enough.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1266,7 +1267,15 @@ private Mono<Application> checkoutRemoteBranch(String defaultApplicationId, Stri
...
                                             .zipWith(Mono.just(application1))
-                            );
+                            )
+                            .onErrorResume(throwable -> {
+                                if (throwable instanceof DuplicateKeyException) {
+                                    return fileUtils.reconstructApplicationJsonFromGitRepo(srcApplication.getOrganizationId(), defaultApplicationId, srcApplication.getGitApplicationMetadata().getRepoName(), branchName)
+                                            .zipWith(Mono.just(tuple.getT2()));
+                                }
+                                log.error(" Git checkout remote branch failed", throwable.getMessage());
+                                return Mono.error(new AppsmithException(AppsmithError.INTERNAL_SERVER_ERROR));
```
abhvsn, line 1277: Internal server error is too generic can we use git error so that it will be easy to detect the root cause

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1266,7 +1267,15 @@ private Mono<Application> checkoutRemoteBranch(String defaultApplicationId, Stri
                             .flatMap(application1 ->
                                     fileUtils.reconstructApplicationJsonFromGitRepo(srcApplication.getOrganizationId(), defaultApplicationId, srcApplication.getGitApplicationMetadata().getRepoName(), branchName)
                                             .zipWith(Mono.just(application1))
-                            );
+                            )
+                            .onErrorResume(throwable -> {
+                                if (throwable instanceof DuplicateKeyException) {
+                                    return fileUtils.reconstructApplicationJsonFromGitRepo(srcApplication.getOrganizationId(), defaultApplicationId, srcApplication.getGitApplicationMetadata().getRepoName(), branchName)
```
abhvsn, line 1273: Let's add a comment on why the handling is different for duplicate key exceptions than in the other scenarios.

abhvsn (approved):
LGTM

abhvsn (approved):

PR #11555, opened 2022-03-02
chore: Soft delete resources across Appsmith board for delete operation

### LINKED ISSUE #11308: [Bug]: All the resources present in DB should be soft deleted

### Current Behavior

Currently, there are some inconsistencies in how we delete the resources in Appsmith, to name a few 
1. When we delete the action individually and then publish the application it will be hard deleted from DB but for delete application this is soft deleted.
2. actionCollections are not at all getting deleted **soft/hard** when we delete the application, this results in orphan action collections which might be bloating our DB

### Steps To Reproduce

1. Create application
2. Add action
3. delete action and publish, you will find the action is hard deleted from DB
4. Create another action
5. Delete the application
6. Action is soft deleted this time

### Environment

Production

### Version

Cloud

abhvsn:
## Description

> Currently, there are some inconsistencies in how we delete the resources in Appsmith as follows:
> 1. When we delete the action or actionCollection individually and then publish the application it will be hard deleted from DB but for delete application this is soft deleted.
> 2. actionCollections are not at all getting deleted soft/hard when we delete the application, this results in orphan action collections which might be bloating our DB
> 3. While deleting the page if the action and actionCollection are not published these are getting hard deleted

This PR fixes the above issues by soft deleting the documents. 

Fixes #11308 

## Type of change

## How Has This Been Tested?

> Manual test
> JUnit TC

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
@@ -1,7 +1,10 @@
 package com.appsmith.server.migrations;
 
+import com.appsmith.server.domains.NewAction;
 import com.appsmith.server.domains.Plugin;
+import com.appsmith.server.domains.QNewAction;
 import com.appsmith.server.domains.QPlugin;
+import com.appsmith.server.dtos.ActionDTO;
 import com.github.cloudyrock.mongock.ChangeLog;
 import com.github.cloudyrock.mongock.ChangeSet;
 import com.github.cloudyrock.mongock.driver.mongodb.springdata.v3.decorator.impl.MongockTemplate;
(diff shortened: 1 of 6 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCE.java
```diff
@@ -33,6 +33,4 @@
     <T extends BaseDomain> Mono<T> sendDeleteEvent(T object, Map<String, Object> extraProperties);
 
     <T extends BaseDomain> Mono<T> sendDeleteEvent(T object);
-
-    <T extends BaseDomain> Mono<T> sendArchiveEvent(T object);
```
nidhi-nair, line 37: Do we want to retain the archive event separately from the delete event?

nidhi-nair (commented):
LGTM, just a question around analytics @abhvsn

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCE.java
abhvsn: I thought as we are not using this anymore we can safely remove this and avoid any possible confusion in future, as when to use deleteEvent against this one.

nidhi-nair: Ah okay. We're basically saying there is essentially no difference between a delete and an archive, right?

nidhi-nair (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCE.java
abhvsn: Yes, in future if we wanted to differentiate these we can always revert to this implementation.

PR #11561, opened 2022-03-02
chore: Modified plugin config aggregation logic

### LINKED ISSUE #11536: [Task]: Modify UQI editor JSON creation logic

### SubTasks

- Move root json file out of editor folder
- Use root json as pure definition of selectors to determine action json
- Populate all jsons within editor json irrespective of reference. This will allow us to add complex logic for visibility.

nidhi-nair:
## Description

The existing logic for aggregating plugin editor config is on the basis of a single dropdown that determines where the remainder of the logic for the form lies. This is limiting because we are unable to create a `SELECTOR` for rendering a form based on multiple selections. 

For example, we might want to split the selectors for Google Sheets from the single commands drop down to two drop downs, one for selecting entity type (Spreadsheet file, sheet or rows), and another to select what action we want to perform on each.

The changes here disassociate the files to be loaded for aggregation from the components that determine the logic for rendering the form. As a result, the `SELECTOR` section can have any form of complicated logic/components, and the files that are loaded in the next section will all be responsible for making sure they fulfil the right set of conditions to display themselves.

Fixes #11536 

## Type of change

## How Has This Been Tested?
- Manually tested for regression
- JUnit test case

## Checklist:


mojtab23 (commented):
@nidhi-nair the code looks fine. Because I'm not very familiar with code I have some questions.
 Do these changes require some changes in the frontend or not?
Do these changes are compatible with other plugins or have you covered all the plugins that work this way? I mean this change from editor.json to editor/root.json is a new feature or you just somehow changed or refactored the code.

nidhi-nair:
All good questions, @mojtab23.

1. The changes are purely on the backend, because we're simply changing the way we aggregate the JSON configs. The end result that goes out to the client is still the same. We're simply increasing the number of possibilities that the individual configs can have.
2. These changes affect all the UQI ported plugins. Since we only have 3 of these today, those are all the plugins I had to touch in the PR (S3, Firestore, Mongo)

Issue #11539, opened 2022-03-02
Call user profile, and features request in parallel  to optimise first load


SatishGandham:
### This might not be required after we fix the router. 

Right now, we call the features request after the user/me request is done. Doing this adds to delay in loading the application.

We should make them in parallel. Before that, investigate if combining these two api's into a single one is more appropriate.

arunvjn:
API to fetch feature flag returns 401 for an anonymous user. We need to change the response to 200 ok with empty object as response. 
@nayan-rafiq Can you help us here ?

PR #11604, opened 2022-03-03
fix: Prevent page reload when launching an application from the applications dashboard

### LINKED ISSUE #11537: Fix the react router routing on the application list
On the applications list screen, when you click on any application it is served as a new page, meaning we serve the html document again from the server.

This means we spend time in doing two unnecessary things, 

- load the html document and 
- Parse, evaluate the scripts and html again.

This is due to using anchor tag for the application edit button. Ideally we should be using the <Link> component provided by react router, but for now use whatever the pages and apis use in the entity explorer.

With this change we will eliminate the extra working from these two steps. 

Screenshot 2022-03-02 at 10 20 50 AM

Fix react router

arunvjn:
## Description
Prevent page reload when launching an application from the applications dashboard

Fixes #11537

## Type of change
- Performance fix

## How Has This Been Tested?
- Manual

## Checklist:

### DIFF
--- a/app/client/src/pages/Applications/ApplicationCard.tsx
+++ b/app/client/src/pages/Applications/ApplicationCard.tsx
@@ -55,6 +55,7 @@ import { Variant } from "components/ads/common";
 import { getExportAppAPIRoute } from "@appsmith/constants/ApiConstants";
 import { Colors } from "constants/Colors";
 import { CONNECTED_TO_GIT, createMessage } from "@appsmith/constants/messages";
+import history from "utils/history";
 
 type NameWrapperProps = {
   hasReadPermission: boolean;
@@ -773,9 +774,9 @@ export function ApplicationCard(props: ApplicationCardProps) {
                     <EditButton
                       className="t--application-edit-link"
                       fill
-                      href={editApplicationURL}
                       icon={"edit"}
                       iconPosition={IconPositions.left}
+                      onClick={(e) => history.push(editApplicationURL)}
                       size={Size.medium}
                       text="Edit"
                     />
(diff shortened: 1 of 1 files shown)


app/client/src/pages/Applications/ApplicationCard.tsx
```diff
@@ -773,9 +774,12 @@ export function ApplicationCard(props: ApplicationCardProps) {
                     <EditButton
                       className="t--application-edit-link"
                       fill
-                      href={editApplicationURL}
                       icon={"edit"}
                       iconPosition={IconPositions.left}
+                      onClick={(e) => {
```
SatishGandham, line 779: Button component uses `tag` prop to decide between link and button. We are not passing it here, and this will render as an anchor tag without href.
```

function Button(props: ButtonProps) {
  return props.tag === "button" ? (
    <ButtonComponent {...props} />
  ) : (
    <LinkButtonComponent {...props} />
  );
}
```

Can we remove the tag prop and decide it based on href property. If you think this is out of scope of this PR, please raise an issue for that change.

Why are we stopping propagation? Whats happening with out it?

SatishGandham (changes requested):

app/client/src/pages/Applications/ApplicationCard.tsx
arunvjn: Will set the tag prop as a "button" here. Added stop propagation only as a safety net to not trigger other events.
Will check and remove this if necessary.

SatishGandham (approved):
Looks good, I can see a noticeable improvement in page load time.

~~We may no longer need on #11539 , but working on this might save 100-300ms of our /applications load time.~~

We will need it when the user directly lands on the editor page

Issue #11418, opened 2022-02-24
[Bug]: If MongoDB user doesn't have permissions to see replicaset status, server doesn't start


sharat87:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Currently, we check if replicaset is enabled on the external MongoDB, and if not, we exit and the fat container doesn't start.

But if replicaset is enabled, but the user being used doesn't have permission to do `rs.status()`, this failure is a false-positive.

So, can we check if we have the admin or just `clusterMonitor` role before doing the check? If we don't have permission to do so, can we try an alternate method to see if replicaset is enabled? Like, checking if the `changeStream` API is working? Or worst case, just print a warning message that we couldn't check for replicaset configuration on the MongoDB server?

### Steps To Reproduce

1. Start up Appsmith fat container.
2. Connect it with an external MongoDB database, but use a user credential that's not an admin, and has just read/write permissions on the database and nothing else.
3. Try to restart the container.

You'll see the following error and the container won't start.

```
appsmith-ce    | ********************************************************************
appsmith-ce    | *          Mongodb cloud replica set is not enabled                *
appsmith-ce    | ********************************************************************
appsmith-ce exited with code 1
```

### Environment

Production

### Version

Self-Hosted

MichaelLeeHobbs:
[appsmith](https://github.com/appsmithorg/appsmith/tree/7a950004db2f46ce3ff0db162f7b3382947bc219)/[deploy](https://github.com/appsmithorg/appsmith/tree/7a950004db2f46ce3ff0db162f7b3382947bc219/deploy)/[docker](https://github.com/appsmithorg/appsmith/tree/7a950004db2f46ce3ff0db162f7b3382947bc219/deploy/docker)/entrypoint.sh 
There is a bug on line 125

```
  if [[ $isUriLocal -gt 0 ]]; then
```

Should be

```
  if [[ $isUriLocal -eq 0 ]]; then
```

MichaelLeeHobbs:
I have confirmed this is the fix. Here is a work around example.

In the compose file add a volume `- ./patch/entrypoint.sh:/opt/appsmith/entrypoint.sh`
```
  appsmith:
    image: index.docker.io/appsmith/appsmith-ce
    restart: always
    env_file: env/appsmith.dev.env
#    entrypoint: 'sleep 500'
    ports:
      - "9001:9001"
    volumes:
      - appsmith:/appsmith-stacks
[...]

PR #11613, opened 2022-03-04
fix: mongodb permission issue when checking replica set status

khoahuynhdev:
## Description

- Use changeStream API from mongoDB driver to check if replica set is enabled
- Create new appsmithctl command `appsmith check-replica-set` or `appsmithctl crs`

Fixes # (issue)
- issue: https://github.com/appsmithorg/appsmith/issues/11418

## Type of change

#### admin user with replica set

1. Start Appsmith with docker/docker-compose

#### Non-admin user with replica set

1. Prepare non-admin user (read/write or read-only permission) with replica set enable (using cloud mongo, container image) and EXTERNAL\_URI
2. Run Appsmith with docker/docker-compose
3. Update mongoDB connection string to EXTERNAL\_URI
4. Restart container
    [](https://collab.geekup.vn/uploads/images/gallery/2022-03/9hyvVm2l2hIFI94O.png)

#### Non-admin user without replica set

1. Prepare non-admin user(read/write or read-only permission) without replica set enable (using cloud mongo, container image,etc) and EXTERNAL\_URI
2. Run Appsmith with docker/docker-compose
3. Update mongoDB connection string to EXTERNAL\_URI
4. Restart container
    [](https://collab.geekup.vn/uploads/images/gallery/2022-03/0xuLf65hACO0FUkV.png)

## Checklist:

### DIFF
--- a/deploy/docker/entrypoint.sh
+++ b/deploy/docker/entrypoint.sh
@@ -102,7 +102,7 @@ init_replica_set() {
       break
     fi
   done
-  
+
   if [[ $shouldPerformInitdb -gt 0 && $isUriLocal -eq 0 ]]; then
     echo "Initializing Replica Set for local database"
     # Start installed MongoDB service - Dependencies Layer
@@ -126,10 +126,9 @@ init_replica_set() {
   if [[ $isUriLocal -gt 0 ]]; then
     # Check mongodb cloud Replica Set
     echo "Checking Replica Set of external MongoDB"
-    responseStatus=$(mongo "$APPSMITH_MONGODB_URI" --eval "rs.status()" | grep ok | xargs)
-    okString="ok : 1"
+    cmd="appsmithctl check-replica-set"
 
-    if [[ $responseStatus == *$okString* ]]; then
+    if $cmd; then
       echo "Mongodb cloud Replica Set is enabled"
       mongo "$APPSMITH_MONGODB_URI" --eval 'rs.initiate()'
     else
--- a/deploy/docker/entrypoint.sh
+++ b/deploy/docker/entrypoint.sh
@@ -127,7 +127,7 @@ init_replica_set() {
     # Check mongodb cloud Replica Set
     echo "Checking Replica Set of external MongoDB"
 
-    if $(appsmithctl check_replica_set); then
+    if appsmithctl check_replica_set; then
       echo "Mongodb cloud Replica Set is enabled"
       mongo "$APPSMITH_MONGODB_URI" --eval 'rs.initiate()'
     else
--- a/deploy/docker/entrypoint.sh
+++ b/deploy/docker/entrypoint.sh
@@ -126,9 +126,8 @@ init_replica_set() {
   if [[ $isUriLocal -gt 0 ]]; then
     # Check mongodb cloud Replica Set
     echo "Checking Replica Set of external MongoDB"
-    cmd="appsmithctl check-replica-set"
 
-    if $cmd; then
+    if $(appsmithctl check_replica_set); then
       echo "Mongodb cloud Replica Set is enabled"
       mongo "$APPSMITH_MONGODB_URI" --eval 'rs.initiate()'
     else
--- a/deploy/docker/utils/bin/check_replica_set.js
+++ b/deploy/docker/utils/bin/check_replica_set.js
@@ -28,7 +28,7 @@ async function exec() {
 
 async function checkReplicaSet(client) {
   await client.connect();
-  return new Promise((resolve) => {
+  return await new Promise((resolve) => {
     try {
       client
         .watch()
--- a/deploy/docker/utils/bin/check_replica_set.js
+++ b/deploy/docker/utils/bin/check_replica_set.js
@@ -5,19 +5,25 @@ async function exec() {
     useNewUrlParser: true,
     useUnifiedTopology: true,
   });
-  checkReplicaSet(client).then((res) => {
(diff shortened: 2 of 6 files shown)


deploy/docker/entrypoint.sh
```diff
@@ -126,10 +126,9 @@ init_replica_set() {
   if [[ $isUriLocal -gt 0 ]]; then
     # Check mongodb cloud Replica Set
     echo "Checking Replica Set of external MongoDB"
-    responseStatus=$(mongo "$APPSMITH_MONGODB_URI" --eval "rs.status()" | grep ok | xargs)
-    okString="ok : 1"
+    cmd="appsmithctl check-replica-set"
 
-    if [[ $responseStatus == *$okString* ]]; then
+    if $cmd; then
```
sharat87, line 131: Is there a reason we are using a `cmd` variable here? Wherever possible, let's avoid variable interpolation without double-quotes. Can we just change this to `if appsmithctl check-replica-set; then` instead?

deploy/docker/utils/bin/check_replica_set.js
```diff
@@ -0,0 +1,46 @@
...
+      client.close();
+      process.exit(0);
+    }
+
+    // not support replica set
+    if (res === 1) {
+      client.close();
+      process.exit(1);
+    }
+  });
```
sharat87, line 20: Should we do the `client.close()` call in a `.finally`instead of inside the `.then`? So that if there's an error in the `checkReplicaSet` function, the client is still closed, right?

deploy/docker/utils/bin/check_replica_set.js
```diff
@@ -0,0 +1,46 @@
...
+}
+
+async function checkReplicaSet(client) {
+  await client.connect();
+  return new Promise((resolve) => {
```
sharat87, line 25: I think this will result in an unintentional nested-promise. We likely need an `await` here, like:

```suggestion
  return await new Promise((resolve) => {
```

deploy/docker/utils/bin/utils.js
```diff
@@ -8,5 +8,6 @@ function showHelp() {
   console.log('\tex, export_db\t\tExport interal database.\r');
   console.log('\tim, import_db\t\tImport interal database.\r');
   console.log('\tmi, migrate\t\tMigrate new server.\r');
+  console.log('\tcrs, check_replica_set\t\tcheck replica set mongoDB.\r');
```
sharat87, line 11: Shouldn't this be `check-replica-set` instead of `check_replica_set`, like the command in the entrypoint.sh script?

sharat87 (changes requested):

deploy/docker/utils/bin/utils.js
geekup-legodevops: The commands in the 'appsmithctl' are following the under_score syntax, so i think the check replica set command should follow the same.

geekup-legodevops: It's a typo, i will update it

deploy/docker/utils/bin/check_replica_set.js
khoahuynhdev: updated

deploy/docker/entrypoint.sh
khoahuynhdev: i have changed the condtion as suggested

deploy/docker/utils/bin/check_replica_set.js
khoahuynhdev: - I have added the `catch` phase to exit 1 when there is an error. 
- The `finally` phase will **always run** after `then` and `catch` phase but we exit the process in the `then` and `catch` so the client won't be closed in the `finally`.

deploy/docker/utils/bin/check_replica_set.js
sharat87: I don't see the `await` keyword in the return statement still, do you need to push your local commits perhaps? 🤔

deploy/docker/entrypoint.sh
```diff
@@ -126,10 +126,8 @@ init_replica_set() {
   if [[ $isUriLocal -gt 0 ]]; then
     # Check mongodb cloud Replica Set
     echo "Checking Replica Set of external MongoDB"
-    responseStatus=$(mongo "$APPSMITH_MONGODB_URI" --eval "rs.status()" | grep ok | xargs)
-    okString="ok : 1"
 
-    if [[ $responseStatus == *$okString* ]]; then
+    if $(appsmithctl check_replica_set); then
```
sharat87, line 130: I'm not sure this would work. For this to work, the command `appsmitctl check_replica_set` should _print_ `true` or `false`. But from the implementation it doesn't seem to do that.

Instead the command exits with 0 or 1 status code. To make use of that, this should be written without interpolation, like this:

```suggestion
    if appsmithctl check_replica_set; then
```

sharat87 (changes requested):

deploy/docker/utils/bin/check_replica_set.js
geekup-legodevops: Sorry, my bad. I have updated from the local

deploy/docker/entrypoint.sh
geekup-legodevops: thanks for the suggestion. We updated it.

sharat87 (approved):

PR #11626, opened 2022-03-04
feat: Add applicationVersion property to Application

### LINKED ISSUE #11471: [Task]: Add a new applicationVersion attribute to application object
We will require the org admins to manually update their old applications to the new URL format. We would require a 'applicationVersion' attribute that defaults to 1 for old applications to achieve this. Every new application that is created post Clean URLs will have an applicationVersion value of 2. We will need either migration to set this property of old applications to 1 or have the APIs return 1 when there is not applicationVersion property in the object.

nayan-rafiq:
## Description
Added a new property to Application object - `applicationVersion`. This property can be used to identity when there is a breaking change and can not be solved with migration. FE will use this property to detect such conditions. Another API is added to migrate the `applicationVersion` to latest version when user migrates the breaking changes.

Fixes #11471 

## Type of change

## How Has This Been Tested?

- JUnit tests
- Manually

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationControllerCE.java
```diff
@@ -217,4 +217,11 @@ public ApplicationControllerCE(
         return themeService.changeCurrentTheme(themeId, applicationId)
                 .map(theme -> new ResponseDTO<>(HttpStatus.OK.value(), theme, null));
     }
+
+    @PatchMapping("{applicationId}/application-version/upgrade")
```
mohanarpit, line 221: QQ: Why are we introducing a new API to update the application version? Can we not use the updateApplication endpoint to update this field? I'm asking because apart from updating the version, I can't seem to see this function doing anything special.

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Application.java
```diff
@@ -34,6 +34,9 @@
 
     String organizationId;
 
+    /*
```
mohanarpit, line 37: Is this TODO still relevant? If not, can we please remove it?

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationControllerCE.java
nayan-rafiq: This API adds the latest version to `applicationVersion`. The update Application API also can do this but I thought about adding a new one to to make sure latest value is always set. As the FE will have to manage the applicationVersion also, now I realize it would be better to use the update application API instead of adding this one.

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Application.java
nayan-rafiq: Yes, it's still relevant. We've used default value for the `isPublic` property. Need to remove this default initialization in future.

arunvjn:
@nayan-rafiq Can you change the target branch to feat/better_urls ? It needs to go along with the front-end changes.

nayan-rafiq:
> @nayan-rafiq Can you change the target branch to feat/better_urls ? These changes needs to go into release along with the front-end changes.

@arunvjn I can do that but in that case I need to change the latest version number as well. Currently I've kept the old version number and new version numbers same. Should I change the PR and set the latest app version number to 2?

arunvjn:
> > @nayan-rafiq Can you change the target branch to feat/better_urls ? These changes needs to go into release along with the front-end changes.
> 
> @arunvjn I can do that but in that case I need to change the latest version number as well. Currently I've kept the old version number and new version numbers same. Should I change the PR and set the latest app version number to 2?

Yes, that would allow me to test the changes behavior on front-end as well.

PR #11676, opened 2022-03-07
fix: Modified root json for SMTP to comply with UQI aggregation rules

nidhi-nair:
SMTP is registered as a UQI plugin, and hence needs to follow the format per UQI aggregation.


trishaanand (approved):

nidhi-nair:
Skipping tests since this is a config change and we don't have tests for SMTP yet.

PR #11695, opened 2022-03-08
fix: Fix links to old docs in client

sharat87:
## Description

Fixes all links to old docs with new ones, ones that don't contain a hardcoded version number in them. These old links are misleading users to pages that contain old and often incorrect information.

## Type of change

    // Code coverage diff between base branch:release and head branch: fix/docs-links-in-client 

### DIFF
--- a/app.json
+++ b/app.json
@@ -34,7 +34,7 @@
         "value": ""
       },
       "APPSMITH_DISABLE_TELEMETRY": {
-        "description" : "We want to be transparent and request that you share anonymous usage data with us. This data is purely statistical in nature and helps us understand your needs & provide better support to your self-hosted instance. You can read more about what information is collected in our documentation https://docs.appsmith.com/v/v1.2.1/setup/telemetry",
+        "description" : "We want to be transparent and request that you share anonymous usage data with us. This data is purely statistical in nature and helps us understand your needs & provide better support to your self-hosted instance. You can read more about what information is collected in our documentation https://docs.appsmith.com/telemetry",
         "value": "false"
       }
    }
--- a/app/client/src/components/editorComponents/GlobalSearch/githubHelper.ts
+++ b/app/client/src/components/editorComponents/GlobalSearch/githubHelper.ts
@@ -3,28 +3,28 @@ import { DocSearchItem } from "./utils";
 export const defaultDocsConfig = [
   {
     link:
-      "https://raw.githubusercontent.com/appsmithorg/appsmith-docs/v1.2.1/tutorials/building-a-store-catalog-manager/README.md",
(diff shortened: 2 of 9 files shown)


app.json
```diff
@@ -34,7 +34,7 @@
         "value": ""
       },
       "APPSMITH_DISABLE_TELEMETRY": {
-        "description" : "We want to be transparent and request that you share anonymous usage data with us. This data is purely statistical in nature and helps us understand your needs & provide better support to your self-hosted instance. You can read more about what information is collected in our documentation https://docs.appsmith.com/v/v1.2.1/setup/telemetry",
+        "description" : "We want to be transparent and request that you share anonymous usage data with us. This data is purely statistical in nature and helps us understand your needs & provide better support to your self-hosted instance. You can read more about what information is collected in our documentation https://docs.appsmith.com/telemetry",
```
Nikhil-Nandagopal, line 37: @sharat87 we need to hardcode the version in all the URLs because if we ever change the URLs, these would break. That is why we publish a new version when we break the URL structure

Nikhil-Nandagopal:
@sharat87 the versions should be bumped up but not removed because changing the URLs in the future would break all these old links

sharat87:
@Nikhil-Nandagopal, I actually just went to docs.appsmith.com, searched for the topic and copied the link in the browser. This didn't have a version at all. Wouldn't it be confusing if the link from docs.appsmith.com, and the link from within Appsmith's UI are different? As in, one include a hard-coded version number and the other doesn't?

In fact, if I use the Omnibar to search the docs for something, say `email`, I get the link without a version number. The version-less links are what the user sees everywhere, so I figured that's what we want the users to see here as well, no?

Nikhil-Nandagopal:
@sharat87 omnibar search is not a problem since it's a search but if we don't use absolute paths to links, we run the risk of the link being broken which is much worse than outdated. Generally even outdated information is fairly relevant expect for a few UI tweaks. It is a bit of a problem that the product and docs can diverge but we can just update the version whenever to solve this, the bigger problem is the docs change and the links get broken which happened in the past and we didn't even know it

sharat87:
@Nikhil-Nandagopal, those are two separate problems. Broken links can happen even if we hardcode the version number. Example: <https://docs.appsmith.com/v/v1.2.1/framework-reference/geolocation>, which is one of the things in this PR.

Broken links should be fixed by applying redirects when we are changing the link. Hard coded version numbers doesn't make the links work forever.

I'm okay to change the links to use the `v1.3` instead of not having a version number, sure, but I just don't think that means the links will work forever. We need separate monitoring there to ensure all links work.

Nikhil-Nandagopal:
@sharat87 that happened because of a process miss where we changed the URL in an existing version instead of publishing a new version. We don't have control over URL re-directs here.
Whenever a latest version is set as default, it is accessible without the version number and that is what caused this. So we ensured that we always use a version number for all links going forward

PR #11770, opened 2022-03-10
fix: Update admin settings permission from users when they are added or re…

### LINKED ISSUE #11118: [Bug]: The option to add admin emails does not grant other users access to admin setting

### Current Behavior

The option to add admin emails does not grant other users access to admin setting
image

### Steps To Reproduce

1. In the admin setting page 
2. under general tab
3. In Admin Email when user adds any email address along with the admin email 
4. The newly added users list doesn't get the admin access

### Environment

Production

### Version

Self-Hosted

nayan-rafiq:
## Description

Update user permission when email address is added as admin email from settings page.

Fixes #11118 

## Type of change

## How Has This Been Tested?
- Manually

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/EnvManagerCEImpl.java
```diff
@@ -253,6 +261,71 @@
...
+        }
+
+        if (!CollectionUtils.isEmpty(newAdminEmails)) {
+            for (String email : newAdminEmails) {
+                if (oldAdminEmails != null && !oldAdminEmails.contains(email)) {
+                    // need to add permission for this user
+                    newUsers.add(email);
+                }
+            }
+        }
```
sharat87, line 297: Could this be replaced with the `removeAll` method of Java's Set class? Something like:

```
        Set<String> removedUsers = new HashSet<>(oldAdminEmails);
        removedUsers.removeAll(newAdminEmails);
        Set<String> addedUsers = new HashSet<>(newAdminEmails);
        addedUsers.removeAll(oldAdminEmails);
```

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/EnvManagerCEImpl.java
```diff
@@ -253,6 +261,71 @@
...
+     * Adds or removes admin user policy from users.
+     * If an email is removed from admin emails, it'll remove the policy from that user.
+     * If a new email is added as admin email, it'll add the policy to that user
+     * @param oldAdminEmailsCsv comma separated email addresses that was set as admin email earlier
+     */
+    private void updateAdminUserPolicies(String oldAdminEmailsCsv) {
+        Set<String> oldAdminEmails = null;
+        if (!StringUtils.isEmpty(oldAdminEmailsCsv)) {
+            oldAdminEmails = Set.of(oldAdminEmailsCsv.trim().split("\\s*,\\s*"));
+        }
```
sharat87, line 274: What happens if the value of `oldAdminEmailsCsv` is just a space, like `" "`? Then the resulting set of emails will contain one string, which is the empty string, `""`. Then we subsequently attempt to find user with this email... which isn't harmless I think, but is there something we can do to avoid? Check out also inputs like `"a@one.com, , , b@two.com"`, which might also screw this up. I think if we use the split regex as `"(\\s*,\\s*)+"` perhaps.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/EnvManagerCEImpl.java
```diff
@@ -253,6 +261,71 @@
...
+                    Map<String, Policy> policyMap =  policyUtils.generatePolicyFromPermission(
+                            Set.of(AclPermission.MANAGE_INSTANCE_ENV), user.getUsername()
+                    );
+                    policyUtils.addPoliciesToExistingObject(policyMap, user);
+                    return userRepository.save(user);
+                });
+
+        /*
+         * we need to run these two flux immediately because server will be restarted and these changes
+         * should be persisted to DB before that
```
sharat87, line 319: Calling `.subscribe` on them doesn't mean they'll be called immediately. In fact, it's a little closer to the opposite. By calling this, we are telling the Reactive framework, that these can be done a separate thread, so just queue them up. When a thread in bounded thread pool becomes free, then schedule these jobs on those free threads.

To ensure they are finished before the restart, have this method return a `Mono<Void>` or something, and change this to something like:

```java
return Mono.whenDelayError(
        removedUserFlux.then(),
        newUsersFlux.then()
);
```

This will run the two `Flux` chains in parallel, and the returned `Mono` will only emit when _both_ of them are finished.

Now, in the caller of this method, ensure this returned `Mono` is finished before restart. Something like:

```java
updateAdminsMono.then(Mono.delay(Duration.ofSeconds(1))).flatMap(ignored -> restart())
```

sharat87 (changes requested):
Few changes, but overall looks good.

Are we not handling the case of checking for email in admin list on signup? Will that be a separate PR?

nayan-rafiq:
> separate

Good point. I really missed that. I'll handle the case and update this PR.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/EnvManagerCEImpl.java
```diff
@@ -151,8 +157,31 @@
...
+                return Mono.error(new AppsmithException(AppsmithError.INVALID_PARAMETER, "Admin Email"));
+            } else { // make sure user is not removing own email
+                Set<String> adminEmails = TextUtils.csvToSet(emailCsv);
+                if(!adminEmails.contains(user.getEmail())) { // user can not remove own email address
+                    return Mono.error(new AppsmithException(
+                            AppsmithError.GENERIC_BAD_REQUEST, "Removing own email from Admin Email is not allowed"
+                    ));
+                } else {
+                    // set the clean value to changes
+                    changes.put(APPSMITH_ADMIN_EMAILS.name(), String.join(",", adminEmails));
```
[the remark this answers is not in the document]
nayan-rafiq: @sharat87 We're removing the empty strings, duplicate emails during the validation. We've to do that again when storing the values. For example `b@example.com,,` is a valid input here so we'll end up storing this. As there will not be a lot of emails in this field, I think loosing the order will be ok.

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ValidationUtils.java
```diff
@@ -1,10 +1,17 @@
...
+import org.springframework.util.StringUtils;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public final class ValidationUtils {
     public static final int LOGIN_PASSWORD_MIN_LENGTH = 6;
     public static final int LOGIN_PASSWORD_MAX_LENGTH = 48;
+    private static final Pattern EMAIL_CSV_PATTERN = Pattern.compile(
+            "^[\\W]*([\\w+\\-.%]+@[\\w\\-.]+\\.[A-Za-z]+[\\W]*,{1}[\\W]*)*([\\w+\\-.%]+@[\\w\\-.]+\\.[A-Za-z]+)[\\W]*$"
```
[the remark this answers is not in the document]
sharat87: Ah okay. Let's add with a link to that SO answer as well then. If there's improvements on that answer, we have a linkback. 😄

nayan-rafiq:
@sharat87 Thanks a lot for your valuable review. I've updated the PR as per the points you mentioned. Would you please review it once again?

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ValidationUtils.java
```diff
@@ -1,10 +1,19 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 public final class ValidationUtils {
     public static final int LOGIN_PASSWORD_MIN_LENGTH = 6;
     public static final int LOGIN_PASSWORD_MAX_LENGTH = 48;
+    private static final String EMAIL_PATTERN = "[\\w+\\-.%]+@[\\w\\-.]+\\.[A-Za-z]+";
+
+    private static final Pattern EMAIL_CSV_PATTERN = Pattern.compile(
+            "^[\\s]*(" + EMAIL_PATTERN + "[\\s]*,[\\s]*)*(" + EMAIL_PATTERN + ")[\\s]*$"
```
sharat87, line 15: Isn't `[\\s]` the same as just `\\s` here? 🤔

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/TextUtils.java
```diff
@@ -42,4 +45,29 @@ public static String makeSlug(String inputText) {
...
+//        Set<String> stringSet = new HashSet<>();
+//        if(inputStringCsv != null) {
+//            String[] strings = inputStringCsv.split("(\\s*,\\s*)+");
+//            for(String str : strings) {
+//                if(str.length() > 0) {
+//                    stringSet.add(str.trim());
+//                }
+//            }
+//        }
+//        return stringSet;
```
sharat87, line 71: Can we remove commented code please?

sharat87 (changes requested):
Really sorry for going back-and-forth on this, just a couple of super-minor things and we should merge this in. Thanks for your patience man 🙏

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ValidationUtils.java
nayan-rafiq: Yeah, it's the same. Let me remove the extra square brackets. Thanks a lot man for pointing this out.

nayan-rafiq:
@sharat87 Thanks a lot for your time and patience. The PR has improved a lot for your great observations. Would you please have a look once again?

sharat87 (approved):

PR #11809, opened 2022-03-11
fix: Match iframe by object not origin.

### LINKED ISSUE #11779: [Bug]: Iframe onMessageReceived not getting triggered in case of srcDoc

### Description

Hi team, we recently started using Appsmith, and we are running into an issue with embedded iframe widgets within appsmith.

Specifically, the onMessageReceived hook never seems to be called.

### Steps To Reproduce

1. Add an iframe widget.
2. Set srcDoc to
```
<html>
  <script>
    setTimeout(() => window.parent.postMessage("hello",'*'), 1000);
  </script>
</html>
```

set onMessageReceived to `{{showAlert(Iframe1.message)}}`

Expectation: I see an alert with "hello".
Actual: Nothing happens.

Setup: Self hosted, version: v1.6.13

### Public Sample App

_No response_

### Version

Self hosted, version: v1.6.13

pallavagarwal07:
## Description

This PR moves to comparing the actual iframe contentwindow object with the postmessage source to invoke the onMessageReceived callback. This allows multiple iframes with same origin to keep their messages separate, as well as allows srcDocs to have postmessage in their code as well.

Fixes #11779 

## Type of change

Please note that even though #11779 only describes the srcDoc use case, it seems from the code that having multiple iframes with same origin will also be buggy with the existing implementation.

## How Has This Been Tested?

Brought up locally connecting to appsmith staging server, and tested the examples given in #11779 as well as the case where we have multiple iframes.

## Checklist:

### DIFF
--- a/app/client/src/widgets/IframeWidget/component/index.tsx
+++ b/app/client/src/widgets/IframeWidget/component/index.tsx
@@ -75,16 +75,25 @@ function IframeComponent(props: IframeComponentProps) {
     widgetId,
   } = props;
 
+  const frameRef = useRef<HTMLIFrameElement>(null);
+
   const isFirstRender = useRef(true);
 
   const [message, setMessage] = useState("");
 
   useEffect(() => {
+    const handler = (event: MessageEvent) => {
+      const iframeWindow =
+        frameRef.current?.contentWindow ||
+        frameRef.current?.contentDocument?.defaultView;
+      // Accept messages only from the current iframe
+      if (event.source !== iframeWindow) return;
+      onMessageReceived(event);
+    };
     // add a listener
-    window.addEventListener("message", onMessageReceived, false);
+    window.addEventListener("message", handler, false);
     // clean up
-    return () =>
-      window.removeEventListener("message", onMessageReceived, false);
+    return () => window.removeEventListener("message", handler, false);
   }, []);
 
   useEffect(() => {
@@ -129,9 +138,9 @@ function IframeComponent(props: IframeComponentProps) {
       {message ? (
         message
       ) : srcDoc ? (
-        <iframe src={source} srcDoc={srcDoc} title={title} />
+        <iframe ref={frameRef} src={source} srcDoc={srcDoc} title={title} />
       ) : (
-        <iframe src={source} title={title} />
+        <iframe ref={frameRef} src={source} title={title} />
       )}
     </IframeContainer>
   );
--- a/app/client/src/widgets/IframeWidget/widget/index.tsx
+++ b/app/client/src/widgets/IframeWidget/widget/index.tsx
@@ -162,11 +162,6 @@ class IframeWidget extends BaseWidget<IframeWidgetProps, WidgetState> {
   };
 
   handleMessageReceive = (event: MessageEvent) => {
-    // Accept messages only from the current iframe
-    if (!this.props.source?.includes(event.origin)) {
-      return;
-    }
-
     this.props.updateWidgetMetaProperty("message", event.data, {
       triggerPropertyName: "onMessageReceived",
       dynamicString: this.props.onMessageReceived,


wmdev0808 (approved):
@pallavagarwal07, thank you for your contribution

somangshu:
Tested the deploy preview and its working fine with all other older use cases.

somangshu:
> Please note that even though https://github.com/appsmithorg/appsmith/issues/11779 only describes the srcDoc use case, it seems from the code that having multiple iframes with same origin will also be buggy with the existing implementation.

@wmdev0808 @ashit-rath need your comments on this as well, What do we need to do additionally?

wmdev0808:
> > Please note that even though #11779 only describes the srcDoc use case, it seems from the code that having multiple iframes with same origin will also be buggy with the existing implementation.
> 
> @wmdev0808 @ashit-rath need your comments on this as well, What do we need to do additionally?

@somangshu , It means we will exclude any other messages from the other iframes targeting to our app.
That is, we will only deal with the messages from the actually embedded iframe into the widget

somangshu:
@wmdev0808 how can we support having multiple iframes messages with same origin, Does this sound like a valid use case? cc @dilippitchika @ashit-rath

wmdev0808:
> @wmdev0808 how can we support having multiple iframes messages with same origin, Does this sound like a valid use case? cc @dilippitchika @ashit-rath

Suppose that we have 2 iframes in our app(Both of them will send post messages to our app).
If we set onMessageReceived actions for both widgets, it results in setting listeners on the same windows object.
That is, every widget handler for a post message will be triggered by other messages from other widgets, as well.

pallavagarwal07:
@wmdev0808 Just to clarify, the current implementation would trigger both onMessageReceived when *either* iframe sends a message. This to me would be a bug.

With this PR, the first iframe onmessage would be triggered when the first iframe sends a message, and the second iframe onmessage would be triggered if the second iframe sends a message - which I think is the intended behavior - even if they both have the same origin.

Do you think you may have clients depending on the old (buggy(?)) behavior? Or is there some other concern here that I'm missing?

wmdev0808:
> @wmdev0808 Just to clarify, the current implementation would trigger both onMessageReceived when _either_ iframe sends a message. This to me would be a bug.
> 
> With this PR, the first iframe onmessage would be triggered when the first iframe sends a message, and the second iframe onmessage would be triggered if the second iframe sends a message - which I think is the intended behavior - even if they both have the same origin.
> 
> Do you think you may have clients depending on the old (buggy(?)) behavior? Or is there some other concern here that I'm missing?

I think you're right

ashit-rath:
> @wmdev0808 how can we support having multiple iframes messages with same origin, Does this sound like a valid use case? cc @dilippitchika @ashit-rath

@somangshu There is a way to support this is by having another property called `whitelisted domains` and let the user define what origins they want to mark as valid. Once they have this set, if any one of the iframes is sending the message, all the widgets would pick this message to check if the origin domain is one in the `whitelisted domains` and then proceed with the action.

Otherwise, this PR changes look good!

ashit-rath (approved):

somangshu:
@ashit-rath I am not sure if I am able to follow, Right now with what is implemented can we proceed without breaking anything, In that case I would suggest we create a separate issue around this while we move ahead and close this. Please confirm.

Otherwise we will figure out the solution here itself, Which will also take a while and in that case @pallavagarwal07 we will take it from here, Implement this internally.

ashit-rath:
@somangshu As @pallavagarwal07 pointed out, the current implementation allows multiple iframes to receive messages when any one of the iframes sends a message. I'm not sure if this is something that we are actively supporting or this is just a leak and users are just leveraging it. So, with this fix, it's going to stop working.
 
If it's a feature then we need to support it.
If not then this fix is good to merge.

somangshu:
@ashit-rath I think its a little different
In the current implementation multiple iframe can receive message when the origin is different
The problem is I think with the same origin iframes, all of them will trigger the onMessageReceived together. 

I might not be getting this right cc @pallavagarwal07 @wmdev0808

pallavagarwal07:
I'm confused about your understanding @somangshu.

@ashit-rath got it exactly right in https://github.com/appsmithorg/appsmith/pull/11809#issuecomment-1067549584 . I'm 100% certain the same origin issue is a bug, and not an intentional thing. BUT that doesn't mean you won't have clients depending on it (https://www.hyrumslaw.com/).

If you have a flag gating procedure, the ideal way to roll this out would be to put the fix behind a configuration option (say "iframe legacy mode"). Set it to legacy mode for all clients who are using multiple iframes. And move everyone else to the new behavior.

Let the legacy mode clients know about the change and the configuration option.

I don't know how strict your policy about backward incompatible bug fixes is, so please take a call accordingly. The bug however needs to be fixed (even if its a config option), since we are unable to add more than one custom component as all components get triggered.

somangshu:
Understood, We will take this up as a separate issue, meanwhile go ahead with this fix.

@ashit-rath let me know if you need any help merging this, @vivekverma2312 @YogeshJayaseelan @shwetha-ramesh can one of you give us a QA signoff

YogeshJayaseelan:
@pallavagarwal07 Tested this issue and it is working as expected now

pallavagarwal07:
@somangshu what's the next step?

dilippitchika:
@pallavagarwal07 it will automatically be in production in the next release we do next week.

ashit-rath:
Hey @pallavagarwal07, we have certain tests failing in our CI pipeline, and in order to resolve them can you merge the latest release into your branch and update this PR so that we can get a green signal from CI and merge it ASAP?

ashit-rath (approved):

pallavagarwal07:
Hi, is this PR blocked on something?

somangshu:
> Hi, is this PR blocked on something?

Thats right @pallavagarwal07, This is blocked because some tests in the CI workflow are failing continuously, Our automation team is looking into it, If we feel its not related and ok to merge, ill go ahead and do it, Apologies for the delay here

somangshu:
We are going ahead and merging this, the test failure is also seen in the release env.

somangshu:
@allcontributors add @pallavagarwal07 as a contributor for code, bug

PR #11812, opened 2022-03-12
fix: 10771 show text widget warning on exceeding line breaks limit

### LINKED ISSUE #10771: [Bug]: TextWidget crashes when bound to large data, like JSON from api response

prsidhu:
## Description

Case: Attempting to render a large json object in a TextWidget. Currently, it makes the app unresponsive due large number of line breaks.
Fix: If line breaks > 5000; then render the text without line breaks and show a warning to the user about the same.

Fixes #10771 

https://github.com/appsmithorg/appsmith/issues/10771

## Type of change

## How Has This Been Tested?

1. Add a text widget to your app.
2. Fetch data for the text widget using this api: ```https://restcountries.com/v3.1/all```.
3. Observe that the text is rendered without line breaks and that user is shown an appropriate warning in the property pane. (See image attached)

Screen Shot 2022-03-11 at 8 07 11 PM
Screen Shot 2022-03-12 at 11 17 44 AM

## Checklist:

    // Code coverage diff between base branch:release and head branch: fix-10771_text_widget_warning 
 :red_circle: | total | 55.7 **(-0.01)** | 37.09 **(-0.01)** | 35.28 **(-0.01)** | 55.95 **(-0.02)**
 :red_circle: | app/client/src/workers/helpers.ts | 41.94 **(-58.06)** | 12.5 **(-87.5)** | 50 **(-50)** | 36 **(-64)**
 :red_circle: | app/client/src/workers/validations.ts | 80.39 **(-0.93)** | 71.33 **(-0.36)** | 74.19 **(-2.48)** | 81.35 **(-1.04)**</details>


somangshu:
@dilippitchika can you check this as well? the DP I mean!

prsidhu:
cc: @rohitagarwal88

app/client/src/workers/validations.ts
```diff
@@ -352,9 +352,25 @@ export const VALIDATORS: Record<ValidationTypes, Validator> = {
...
+        const MAX_ALLOWED_LINE_BREAKS = 5000; // Rendering performance deteriorates beyond this number.
+        const lineBreakCount: number =
+          parsed.match(/[^\n]*\n[^\n]*/gi)?.length || 0;
+        if (lineBreakCount > MAX_ALLOWED_LINE_BREAKS) {
+          parsed = JSON.stringify(value); // Parse without line breaks
+          return {
+            isValid: false,
+            parsed,
+            messages: [
+              `Warning: Line breaks in text exceed ${MAX_ALLOWED_LINE_BREAKS}. Text will be visible without line breaks.`,
```
SatishGandham, line 366: Can we say number of lines instead of line breaks? Line breaks feels too technical, and the user is not putting these line breaks.

@somangshu, @dilippitchika  thoughts?

app/client/src/workers/validations.ts
```diff
@@ -352,9 +352,25 @@ export const VALIDATORS: Record<ValidationTypes, Validator> = {
...
 
     if (isObject(value)) {
+      let parsed: string = JSON.stringify(value, null, 2);
+      if (config.params && config.params.limitLineBreaks) {
+        const MAX_ALLOWED_LINE_BREAKS = 5000; // Rendering performance deteriorates beyond this number.
+        const lineBreakCount: number =
+          parsed.match(/[^\n]*\n[^\n]*/gi)?.length || 0;
+        if (lineBreakCount > MAX_ALLOWED_LINE_BREAKS) {
+          parsed = JSON.stringify(value); // Parse without line breaks
+          return {
```
SatishGandham, line 362: @riodeuno Is returning isValid false for valid value the right thing to do?
Any better way to do it? Is it going to break anything?

Also, should we start supporting warnings along with errors at the platform level?

app/client/src/workers/validations.ts
```diff
@@ -352,9 +352,25 @@ export const VALIDATORS: Record<ValidationTypes, Validator> = {
     let parsed = value;
 
     if (isObject(value)) {
+      let parsed: string = JSON.stringify(value, null, 2);
+      if (config.params && config.params.limitLineBreaks) {
+        const MAX_ALLOWED_LINE_BREAKS = 5000; // Rendering performance deteriorates beyond this number.
+        const lineBreakCount: number =
+          parsed.match(/[^\n]*\n[^\n]*/gi)?.length || 0;
```
SatishGandham, line 359: @prsidhu Since validations run several times on the editor, can you explore a faster alternate provided here
https://stackoverflow.com/questions/4009756/how-to-count-string-occurrence-in-string

@riodeuno , Do all the validations run in every eval cycle, or only for the changed paths and the ones dependent on it?

SatishGandham (changes requested):

app/client/src/workers/validations.ts
riodeuno: According to this [line of code](https://github.com/appsmithorg/appsmith/blob/aef4001fdeb676e1444dc71df290db7804fa1b30/app/client/src/workers/DataTreeEvaluator.ts#L663), it happens only if it belongs to the subtree which has been re-computed.

app/client/src/workers/validations.ts
riodeuno: @SatishGandham As far as the property is concerned, a value can be valid or invalid. If we consider a string with more than MAX_ALLOWED_LINE_BREAKS,  a value that we cannot safely work with, then I would say it is an invalid value.

I was wondering about this scenario -- Is it possible for users to resolve the issue, or does the page freeze when we have more than the MAX_ALLOWED_LINE_BREAKS? If it freezes, then we can consider this an `invalid` value, and replace it with defaults during validation. If not, then we can "warn" the user -- 

We have been doing something akin to warnings so far, in some validations, we flag them as `isValid: false` however, we simply pass through the original value to the widget, without resorting to defaults. This is a case, where we do inform the user that there is something wrong with the value, but we don't modify or intercept the value itself.

SatishGandham: @riodeuno , The page doesn't freeze, we format the text to not include line breaks when the line breaks are beyond the threshold.
This will be a warning.

app/client/src/workers/validations.ts
```diff
@@ -352,9 +352,25 @@ export const VALIDATORS: Record<ValidationTypes, Validator> = {
     let parsed = value;
 
     if (isObject(value)) {
+      let parsed: string = JSON.stringify(value, null, 2);
```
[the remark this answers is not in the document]
prsidhu: @SatishGandham I'll add this as a part of the validation at the end. However, I'm unsure about preserving 2k line breaks; it'll make for an odd user experience. @dilippitchika thoughts?

app/client/src/workers/validations.ts
dilippitchika: @SatishGandham can we use line breaks and number of lines interchangeably here? if so then yes we can update it.

app/client/src/workers/validations.ts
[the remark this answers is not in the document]
dilippitchika: Agreed with Preet, it's a weird experience to have 2k line breaks in a way and the rest in another way.

dilippitchika:
Note to self on this, need to update the text widget documentation to help users render these with line-breaks in an iframe widget's srcDoc property with the following code. Adding this as a snippet will also help - 

```
{{
`<pre><code>
${JSON.stringify(GetCountries.data, undefined, 2)}
</code></pre>`
}}
```

app/client/src/workers/validations.ts
dilippitchika: @prsidhu @SatishGandham does this error message work better? - 
"New lines in the text exceed X. The text displayed will not contain any new lines."

SatishGandham (approved):
LGTM

chandannkumar:

PR #11890, opened 2022-03-16
fix: App visibility of git connected application is reset after git merge

### LINKED ISSUE #11682: [Bug]: App visibility is set to private after git merge operation

### Current Behavior

After merging the branches with in Appsmith if the app is public it is reset to private. 

### Steps To Reproduce

1. Connect an app to git repo
2. Create branch from the main branch 
3. Make the app public
4. Add a widget or page or make any changes and then do commit + push
5. Now merge the branch back to main branch
6. Try to access the app in view mode and notice that it is not public and user is redirected to login screen
### Environment

Production

### Version

Cloud

AnaghHegde:
## Description

For git connected application, app visibility is not working correctly. There are two parts
1. Visibility is reset after merging
2. When we change the visibility to Public/Private from the UI, it should be applied to all the branch applications.
This change set will fix both the cases. 

Fixes #11682 

## Type of change

## How Has This Been Tested?

- Locally
- JUnit Tests

## Checklist:

- [X] I have made corresponding changes to the documentation


app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -699,38 +700,48 @@
...
+                                                                                                existingApplication,
+                                                                                                existingApplication.getName(),
+                                                                                                0
+                                                                                        );
+                                                                            }
+                                                                            throw error;
+                                                                        });
                                                             });
                                                 });
+                                    } else {
```
abhvsn, line 722: We can get away from the else as are returning from the if block

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -699,38 +700,48 @@
...
-                                                                                    0
-                                                                            );
-                                                                }
-                                                                throw error;
+                                                    return applicationService.findById(existingApplication.getGitApplicationMetadata().getDefaultApplicationId())
+                                                            .flatMap(application1 -> {
+                                                                // Set the policies from the defaultApplication
+                                                                existingApplication.setPolicies(application1.getPolicies());
+                                                                importedApplication.setPolicies(application1.getPolicies());
+                                                                return applicationService.save(existingApplication)
```
abhvsn, line 708: Why are we not reutilizing the `changeViewAccess` with `isPublic` flag that we have and get rid away with if-else block?

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
AnaghHegde: I added this so that we can easily differentiate between file and git operations.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
AnaghHegde: For to use `changeViewAccess` we need to know the value of `isPublic` flag which will not be part of git/app json. Just like how we are not committing policies we should not commit the flag as well. Since they depend on each other and should use DB as source of truth.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
abhvsn: Can we please remove this to be consistent with other implementation?

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ApplicationServiceTest.java
```diff
@@ -792,6 +792,16 @@ public void makeApplicationPublic_applicationWithGitMetadata_success() {
...
 
+        // Create a branch
+        Application testApplication = new Application();
+        testApplication.setName("branch1");
+        testApplication.setOrganizationId(orgId);
+        GitApplicationMetadata gitApplicationMetadata = new GitApplicationMetadata();
+        gitApplicationMetadata.setDefaultApplicationId(gitConnectedApp.getId());
+        gitApplicationMetadata.setBranchName("test");
+        testApplication.setGitApplicationMetadata(gitApplicationMetadata);
+        Application application = applicationPageService.createApplication(testApplication).block();
```
abhvsn, line 803: I'm not sure if I get this correctly, what are we trying to test with this testcase?

AnaghHegde: This is ti check if the app is made public, all the branched application policy also needs to be updated. I am testing the same scenario.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
AnaghHegde: This else block makes it easy to read and differentiate the flow for file and git import.  Since the method is huge it would be good to have the else block in this scenario.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationServiceCEImpl.java
```diff
@@ -266,9 +266,12 @@ public ApplicationServiceCEImpl(Scheduler scheduler,
     public Mono<Application> changeViewAccess(String defaultApplicationId,
                                               String branchName,
                                               ApplicationAccessDTO applicationAccessDTO) {
-        return this.findByBranchNameAndDefaultApplicationId(branchName, defaultApplicationId, MAKE_PUBLIC_APPLICATIONS)
+        // For git connected application update the policy for all the branch's
+        return findAllApplicationsByDefaultApplicationId(defaultApplicationId)
```
abhvsn, line 270: Should we include this in git-docs that visibility will be applicable at repo level and not the branch level?

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
abhvsn: I still feel this should be consistent with other places where if we are returning from the if block no need to include else part, but yes this is not a blocker though

abhvsn (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationServiceCEImpl.java
AnaghHegde: It is intuitive right? Most of the users complained that it is not working correctly across branches. But there is no harm in adding this to doc.

abhvsn: Reason I asked is because we are sharing branches in git corresponds to app in appsmith. Also today app visibility is at app level that's why confusion might arise in some situations

abhvsn (dismissed):

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationServiceCEImpl.java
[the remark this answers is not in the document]
AnaghHegde: @trishaanand No we are passing the permission in the repository implementation for this method. If you look at the implementation of the method `getApplicationByGitDefaultApplicationId`, we are passing the policy while querying the db.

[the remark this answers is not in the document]
trishaanand: But the permission used by the method `getApplicationByGitDefaultApplicationId` is actually `MANAGE_APPLICATIONS` instead of `MAKE_PUBLIC_APPLICATIONS`.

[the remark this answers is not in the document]
AnaghHegde: I thought `MANAGE_APPLICATION` is a super set and `MAKE_PUBLIC_APPLICATIONS` is part of that.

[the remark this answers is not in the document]
trishaanand: Today it is, but tomorrow, this would be controlled via RBAC with more granular access controls and this flow would break. It is easier to stick to the correct permissions at the application logic level.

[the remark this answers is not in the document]
AnaghHegde: Got it, thanks for highlighting this. I will change this.

trishaanand (approved):

AnaghHegde:
@trishaanand @abhvsn Can you guys please approve the PR again? Just merge release into my branch.

abhvsn (approved):

PR #11925, opened 2022-03-17
fix: Combined API returns branched application id instead of root application id

### LINKED ISSUE #11870: [Bug] Incorrect applicationId in the fetch application and pages API when switching branches
The new API to fetch applications and page list returns an invalid application ID different from actual application ID when switching git branches

nayan-rafiq:
## Description
We've added a new API to combine the get application details and get page list APIs as part of the clean URL project. This API was returning root application instead of branched application when application is in separate branches. This PR fixes that problem.

Fixes #11870 

## Type of change

## How Has This Been Tested?

- Manually

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ResponseUtils.java
```diff
@@ -103,6 +103,10 @@ public ApplicationPagesDTO updateApplicationPagesDTOWithDefaultResources(Applica
             }
             page.setId(page.getDefaultPageId());
         }
+        // need to update the application also if it's present
+        if(applicationPages.getApplication() != null) {
+            applicationPages.setApplication(updateApplicationWithDefaultResources(applicationPages.getApplication()));
```
abhvsn, line 108: Is this the actual solve?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/NewPageServiceCEImpl.java
```diff
@@ -575,9 +581,9 @@ public Layout createDefaultLayout() {
         if(StringUtils.hasLength(applicationId)) {
             return findApplicationPagesByApplicationIdViewModeAndBranch(applicationId, branchName, isViewMode, true);
         } else if(StringUtils.hasLength(pageId)) {
-            return findBranchedApplicationIdFromNewPage(branchName, pageId)
-                    .flatMap(branchedApplicationId -> findApplicationPagesByApplicationIdViewMode(branchedApplicationId, isViewMode, true))
-                    .map(responseUtils::updateApplicationPagesDTOWithDefaultResources);
+            return findRootApplicationIdFromNewPage(branchName, pageId)
+                    .flatMap(rootApplicationId -> findApplicationPagesByApplicationIdViewModeAndBranch(rootApplicationId, branchName, isViewMode, true))
```
abhvsn, line 585: Not able to understand why we need to change the code here?

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ResponseUtils.java
nayan-rafiq: @abhvsn This solves one problem - returning branched application with having id=default application id. Earlier we were returning branched application with their id. Another problem was how application id was fetched. Earlier, I was fetching branched application ID. In order to make is consistent, now I'm fetching default application id.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/NewPageServiceCEImpl.java
nayan-rafiq: @abhvsn This change is not mandatory. I though it would be better to make it consistent with the case when we've `applicationId`. 
When `applicationId` is present, it must be default application id. But when we're fetching page first, we were fetching application with branched application id. I made this change so that in both places we're using default application id.

abhvsn (approved):
LGTM

PR #11928, opened 2022-03-17
Replace deprecated String.prototype.substr()

CommanderRoot:
## Description

[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) or [String.prototype.substring()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring) which work similarily but aren't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

## Type of change

## How Has This Been Tested?

- I have tested the return of each change to make sure it's still the same as before

## Checklist:


mohanarpit:
Apologies for the delay in picking this PR up. Since there was no reviewer assigned, it slipped through the cracks. We will review & merge it in soo.

rahulramesha (approved):

PR #11947, opened 2022-03-17
fix: Delete stale application which is getting created for the failed API in case of fork, duplicate and import

### LINKED ISSUE #10490: [Bug]-[224]:App card is created even when app import fails and no action can be performed on this app card

### Current Behavior

App card is created with no timestamp details when app import fails. This app card is a placeholder and performing any action like edit or launch throws a resource not found error.

[LOOM DEMO](https://www.loom.com/share/71cebf292b384a6293f0767805417ae3)

### Steps To Reproduce

1. Try to import a corrupt app and observe an `internal server error` displayed when import fails
2. Refresh the page and observe an app card created for this app and have no timestamp details.
3. Clicking on `Edit` or `Launch` does not yield expected behaviour instead shows resource not found

### Environment

Production

### Version

Cloud

abhvsn:
## Description

> APIs for import, fork and duplicate application can fail at various stages like cloning pages, actions or JSObjects. This results in stale application getting created in this process. This PR makes sure if API fails for any reason the newly created application is also deleted so that we won't have any stale apps in DB. 

Fixes #10490 

## Type of change

## How Has This Been Tested?

> Manual test 

## Checklist:


nayan-rafiq (approved):
LGTM

nayan-rafiq:
@abhvsn It would be great if we can add few tests for this.

abhvsn:
@nayan-rafiq I was looking for a way to add the test cases but was unable to do so, the reason being the above code kicks in only when we end up in an erroneous state for the mentioned APIs. Even for manual testing, I had to deliberately remove some fields from DB for the parent application so that these APIs will result in errors. But please let me know if there is any way to do this, I'll add those test cases 🙂

nayan-rafiq:
@abhvsn I think mocking some services will help here. For example, you can mock the few services that are called when importing application and raise error from the mocks. Then check whether partial application is created or not.

abhvsn:
Closing this as this is being now handled with transactions, which is much better implementation than the current one.
Ref PR: https://github.com/appsmithorg/appsmith/pull/14939

PR #11961, opened 2022-03-18
fix: Unable to use IN or ANY clause with prepared statements fix

### LINKED ISSUE #11259: [Bug]: Unable to use IN or ANY clause with prepared statements turned on for MySQL

### Current Behavior

https://discord.com/channels/725602949748752515/760761686549463060/941054952711806976

### Steps To Reproduce

Create a select statement using MySQL plugin

### Environment

Production

### Version

Cloud

leotom2000:
## Unable to use IN or ANY clause with prepared statements

Resolved SQL IN clause when used with component mapping.
ANY clause works fine with SQL query, but not with data 

**Issue Resolution :**
When using a widget like Multiselect the IN clause is substituted with a array of values. This causes issues with the way prepared statement was handled earlier. Updated code by adding one method which can parse the multiple inputs and convert into multiple params.
eg:    
```
Earlier :
IN ( [1988, 1973, 1999, 2015] ) is identified with one param (?)   // Causes problem
Now :
IN ( 1988, 1973, 1999, 2015 )  with n params like (?, ?, ?, ?)   // Rest of the code can handle as before.
```

Fixes #11259

## Type of change

## How Has This Been Tested?

**Manual Tests Done**
| Test Case | Actions | Result | Status (Stmt , PreparedStmt) |
| --- | -----------| --- | :---: |
| Simple Select query| Select * from ...| As expected |  Pass, Pass |
| Simple insert query| Insert into...| As expected |  Pass, Pass |
| Simple update query| Update tabel...| As expected |  Pass, Pass |
| Qry with where condition| Params in query | As expected|  Pass, Pass |
| Qry with where condition| Params mapped from widgets | As expected|  Pass, Pass |
| Qry with where condition| Params mapped from multiple widgets | As expected|  Pass, Pass |
| Qry with IN clause| Params set in query | As expected|  Pass, Pass |
| Qry with IN clause| Params mapped from widgets | As expected|  Pass, Pass |
| Qry with IN clause| Params mapped from one multiselect widget | As expected|  Pass, Pass |
| Qry with IN clause| Params mapped from two multiselect widget | As expected|  Pass, Pass |
| Qry with IN clause| Params mapped from two multiselect widget and text widget | As expected|  Pass, Pass |
| Qry with IN clause| Params mapped from two multiselect widget and text widget and manual in query | As expected|  Pass, Pass |
| Qry with ANY clause| values set in query  | Error as expected (No support) |  Pass, Pass |
| Qry with ANY clause| Values mapped from widgets  | Error as expected (No support) |  Pass, Pass |
| Qry with ANY clause| Values from subquery  | As expected |  Pass, Pass |

## Checklist:

### DIFF
--- a/app/server/appsmith-plugins/mysqlPlugin/src/main/java/com/external/plugins/MySqlPlugin.java
+++ b/app/server/appsmith-plugins/mysqlPlugin/src/main/java/com/external/plugins/MySqlPlugin.java
@@ -15,6 +15,7 @@ import com.appsmith.external.models.DatasourceConfiguration;
 import com.appsmith.external.models.DatasourceStructure;
 import com.appsmith.external.models.DatasourceTestResult;
 import com.appsmith.external.models.Endpoint;
+import com.appsmith.external.models.Param;
 import com.appsmith.external.models.Property;
 import com.appsmith.external.models.PsParameterDTO;
 import com.appsmith.external.models.RequestParamDTO;
@@ -206,9 +207,15 @@ public class MySqlPlugin extends BasePlugin {
 
             // In case of non prepared statement, simply do binding replacement and execute
             if (FALSE.equals(isPreparedStatement)) {
+                cleanupParamsForSQLStatements(executeActionDTO);
                 prepareConfigurationsForExecution(executeActionDTO, actionConfiguration, datasourceConfiguration);
                 return executeCommon(connection, actionConfiguration, FALSE, null, null, requestData);
             }
+
+            //Convert array param input to multiple query param
+            convertArrayToSeparateParamsForPreparedStatement(executeActionDTO, actionConfiguration);
+
+            query = actionConfiguration.getBody();
 
             //This has to be executed as Prepared Statement
             // First extract all the bindings in order
@@ -218,6 +225,80 @@ public class MySqlPlugin extends BasePlugin {
             // Set the query with bindings extracted and replaced with '?' back in config
             actionConfiguration.setBody(updatedQuery);
             return executeCommon(connection, actionConfiguration, TRUE, mustacheKeysInOrder, executeActionDTO, requestData);
+        }
+
+        /**
+         * To cleanup queries to handle array inputs
+         * @param executeActionDTO
+         */
+        private void cleanupParamsForSQLStatements(ExecuteActionDTO executeActionDTO) {
+            executeActionDTO.getParams().stream().forEach(
+                    p -> {
+                        if (p.getValue().contains("[")) {
+                            p.setValue(QueryUtils.removeSquareBraces(p.getValue()));
+                        } else {
+                            p.setValue(QueryUtils.addDoubleQuotes(p.getValue()));
+                        }
+                    }
+            );
+        }
+
+        /**
+         * To convert each parameter in an array as separate prepared statement parameter.
+         * eg:To handle Widget like Multiselect dropdown sends parameter as an Array along with a square bracket
+         * @param executeActionDTO
+         * @param actionConfiguration
+         */
+        private void convertArrayToSeparateParamsForPreparedStatement(ExecuteActionDTO executeActionDTO, ActionConfiguration actionConfiguration) {
+
+            String[] modifiedQuery = {actionConfiguration.getBody().trim()};
+            List<Param> paramLists = executeActionDTO.getParams();
+            List<Param> newParamLists = new ArrayList<>();
+
+            paramLists.stream().forEach(
+                    paramItem -> {
+                        String paramValue = paramItem.getValue();
+                        String bodyQuery = modifiedQuery[0];
+
+                        if (paramValue.contains("[") || paramValue.contains("]")) {
+                            Param newParamObj = null;
+                            boolean firstIteration = true;
+                            StringBuilder sb = new StringBuilder();
+                            String paramKey = paramItem.getKey().trim();
+
+                            paramValue = paramValue
+                                    .replaceAll("\\[", "")
+                                    .replaceAll("\\]", "")
+                                    .replaceAll("\"", "");
+                            String[] paramValueItems = paramValue.split(",");
+                            String[] queryParts = bodyQuery.split(paramKey);
+                            sb.append(queryParts[0].substring(0, queryParts[0].length() - 2));
+
+                            for (int i = 0; i < paramValueItems.length; i++) {
+                                String newParamKey = paramKey + "_" + (i + 1);
+                                String newParamValue = paramValueItems[i];
+
+                                newParamObj = new Param();
+                                newParamObj.setKey(newParamKey);
+                                newParamObj.setValue(newParamValue);
+                                newParamLists.add(newParamObj);
+
+                                if (!firstIteration) sb.append(", ");
+                                sb.append("{{");
+                                sb.append(newParamKey);
+                                sb.append("}}");
+
+                                firstIteration = false;
+                            }
+                            sb.append(queryParts[1].trim().substring(2));
+                            modifiedQuery[0] = sb.toString();
+                        }
+                    }
+            );
+            actionConfiguration.setBody(modifiedQuery[0]);
+            newParamLists.stream().forEach(param ->
+                    executeActionDTO.getParams().add(param)
+            );
         }
 
         public Mono<ActionExecutionResult> executeCommon(Connection connection,
--- a/app/server/appsmith-plugins/mysqlPlugin/src/main/java/com/external/utils/QueryUtils.java
+++ b/app/server/appsmith-plugins/mysqlPlugin/src/main/java/com/external/utils/QueryUtils.java
@@ -45,4 +45,33 @@ public class QueryUtils {
         }
         return sb.toString().trim();
     }
+
+    /**
+     * To remove Square braces if any in the query.
+     * In a scenario like Multiselect widget, the params are passed as an array which carries a square bracket.
+     * This method will fix it
+     * @param query
+     * @return
+     */
+    public static String removeSquareBraces(String query) {
+        if (query.contains("[") || query.contains("]")) {
+            query = query.replaceAll("\\[", "").replaceAll("\\]","");
+        }
+        return query;
+    }
+
+    /**
+     * Add double quotes for params missing double quotes.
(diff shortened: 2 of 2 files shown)


leotom2000:
@leotom2000 had meeting with @Sumit and @Nidhi yesterday( 28/03/2022)
Need to discuss again as the array inputs could be a JSON input or Widget Input.

leotom2000:
Recording the conversation we had in Slack Channel regarding this issue for reference :
-------------------------------------------------------------------------------------------------------------

`
Sumit Mar 22nd at 7:31 PM
Need some input on this issue since the resolution of this issue would set a precedent for such kind of issues in future: https://github.com/appsmithorg/appsmith/issues/11259 .
This issue is about MySQL prepared statements not working for IN clause. Although the issue looks like a prepared statement issue, it is actually a smart substitution issue. The prepared statements implementation that we currently have supports the WHERE IN clause properly, but in this bug report the user was either not aware or is not able to construct the query in the right way. e.g.
(1) right format:
select * from ... where myVal in ("{{binding1}}", "{{binding2}}", ...);

(2)  wrong format:
select * from ... where myVal in {{myBinding}};
// myBinding = ["val1", "val2", ...]
As shown above in (1), each value in the array of values following the IN keyword needs to be a prepared statement parameter individually (as per the supported syntax). However the user wants to supply all values as an array using just one binding as shown in (2). If we were to say re-format the binding value in (2) to make it appear like (1) then what we are doing is actually somewhat similar to smart substitution. AFAIK, we haven't done this kind of smart substitution implementation for any of the prepared statement code flows. Hence, wanted to check if we want to go in this direction. cc: @Arpit Mohan @Nikhil N @Trisha Anand @Shri @Nidhi @Rohan Arthur @Ajinkya @Leo Thomas (edited) 
:eyes:
1

26 replies

Nidhi  7 days ago
The approach from #1 does not allow users to insert conditions dynamically, because the expectation there is to provide the exact numbers of inputs expected before hand. We'd talked about one solution for this to be a form based input that understands arrays as an input type based on the operator.
We should support dynamic inputs for the IN clause in the raw mode as well though. I remember we'd talked about SQL parsing as the solution for that a while ago, do you want to be revisiting that idea?

Sumit  7 days ago
@Nidhi agree with you that the recommend format i.e. (1) will not allow dynamically sized array and that it would be nice to have dynamically sized arrays supported. However, when I think about it from security POV, IMO, the core purpose of prepared statements is to provide security - there are two parts to it - (1) to know the structure of the SQL query before hand (2) user's trust that the queries are not getting modified by the application. If we manipulate the queries this way then we may lose both the advantages here.

Nidhi  7 days ago
Hmm .. The security aspect remains unaffected no? Because we'd still be using prepared statements for each value individually, meaning we're guaranteed that bad actors can't replace the individual values by anything fishy. Predictability should also be okay, because in our solve, the evaluated values should show what the PS will look like eventually, like it does right now for other parameters.

Sumit  7 days ago
@Nidhi IMO, the security gets a bit relaxed with this - e.g. say the array of values should ideally have at max or strictly 6 choices, but since it is dynamic it is possible to inject a 7th choice as well. Regarding predictability, IMO, its more about the fact that we have modified user's query - which can take away the feeling of security (not sure though how strongly users would feel about it).

Nikhil N  7 days ago
@Sumit @Nidhi
Do prepared statements not take in an array?
if not then if a user is writing a complex SQL, they cannot achieve a secure query without us generating the query for them right? So we need to eventually be able to modify the query at run time. Is there any other option?
If we've made peace with that then this can either happen by us generating it via the form that simplified crud queries are meant to achieve or by us performing smart substitution.
Even if we do the form, if a user converts it to text then we need to show the SQL and in that case, we would need to do smart substitution here regardless
So then it seems like smart substitution is a necessity if the prepared statements doesn't support an array right?
:+1:
1


Trisha Anand  7 days ago
This is a rock and a hard place kind of a situation. I would say lets go with the dynamic generation of the prepared statement for IN for MySQL:
Atleast the user inputs would be sanitized as opposed to the developer letting this query run without any sanitizations and hence safer.
With Prepared Statement, there's a chance of creating performance bottlenecks on the database and may be bring it down as well if too many unique prepared statements exist and make the db run out of pool space. One of the things that I was thinking of in the future for query performance was to re-use the Prepared Statements hence bring in performance to our query executions. This approach does not give us that. But atleast since we close the PS after every execution, we are simply letting go of the above future optimization.
I was wondering if we can limit this only for databases which dont support ARRAY data types for PS like MySQL. With Postgres, since ARRAY is supported, may be we can still bring in query performance by re-using Prepared Statements in the future. (edited) 
:+1:
1


Shri  6 days ago
Probably missing some context but option 2 seems like an obvious way forward for me. If MySQL supports arrays for placeholders in prepared statements, let’s give the array and let MySQL do what it’s best at. I don’t think security considerations on this are any different from those with just string placeholders. If a bad actor can get more stuff into arrays, they can get more stuff into strings as well.
:heavy_plus_sign:
1


Nidhi  6 days ago
To me the more problematic part is what the UX for this should look like. If we're saying that the user needs to be able to write a proper SQL query, how do we justify using an array in the bindings? Since the binding value as an array could also legitimately be a single value in the table.

Shri  6 days ago
That’s… okay right? How the interpolation will work, whether it’s an array or not, let the Database decide that. Just like it already decides how to interpolate string values vs numeric values vs date values, it’ll be the judge for array values as well. The UX should be just like the UX with string/numeric/date values, that is, just a placeholder in a prepared statement. That consistency can actually add to our UX right?
:thinking_face:
1


Nidhi  6 days ago
Let me use an example to explain what I'm talking about here. Say we have a table Table1 with one row as this:
{
  "k1": "v1",
  "k2": ["v2"]
}
Now, there are two ways in which I might want to use an in clause here.
SELECT * FROM Table1 WHERE k1 IN ({{["v1"]}})
OR
SELECT * FROM Table1 WHERE k2 IN ({{["v2"]}})
In the first case, smart substitution needs to be splitting v1 into an individual parameter as IN (?) and substitute it with "v1".
In the second case, the expectation is IN (?) with ["v2"].
My question is, given these two queries how do we convey to users what the expectation for usage is? It feels like it can get confusing, because even if we use data structure types to introduce the intelligence to do this ourselves, the end user experience is not helping any.
edit: Honestly this might not even be a use case for some DBs, like Postgres's notation for arrays is entirely different. But I'm raising this in case there are DBs out there that work with this notation. But apart from that, the same question still applies if we were to treat k2 as a varchar column. (edited) 

Shri  6 days ago
How about we start with saying “The binding given to the IN operator, should always be an array”, since, honestly, it doesn’t make a whole lot of sense to give a number or a string to the IN operator. With that, we can provide intelligence, error checking by ensuring the type of that binding is indeed an array, and the prepared statement interpolation can just be left to the database, whatever syntax the particular DB uses.
Now the query might be select * from tab where k1 in ({{[selectedRow.k1]}}) or select * from tab where k2 in ({{selectedRow.k2}}). The users just have to make sure they are giving an array to the IN operator, just like they have to make sure to give a date for date columns, number for number columns etc. Would this make it simpler?
:+1::skin-tone-5:
1


Rohan Arthur  6 days ago
This seems like the most understandable way to present this.
Is there a chance that we break users' running queries with this change?

Nidhi  6 days ago
Migrations are going to have loads of edge cases, think it'll be easier to just create a new category of queries for these going forward. I like what Shri is suggesting as well. Just a tiny clarification on the examples. The usage would then be: ({{[selectedRow.k1]}})  and ({{[selectedRow.k2]}}) , yes? Since we need the first to be array of string, and the second to be array of arrays?

Shri  6 days ago
Oh yeah, sure if the datatype of column k2 is itself an array, then yes. Otherwise it’ll turn into a nested array, which probably isn’t what the user wants. Which is why they might then need just ({{selectedRow.k2}}) instead.
Also does interpolation of arrays here need the (), in standard prepared statement syntax? If not, can we also drop them?

Nikhil N  6 days ago
@Nidhi glad you're thinking about the UX here, I don't see another option that makes it easier for a user than saying the IN operator always expects an array of strings like what Shri mentioned. The only way to improve this would be to eventually abstract this behind a form right?

Trisha Anand  6 days ago
Agreed with @Nidhi here.
But re-iterating again. Let's only do this smart substitution for databases where arrays are not handled in PS (Like MySQL) and not to be making this a generic change for DBs like Postgres. The reason I say this is because Prepared Statement gives us an opportunity of performance boost. In Postgres, we can re-use PS instead of closing them after every execution, and get performance boost on our queries. This would not be possible when we are doing dynamic generation of the statement.

Nikhil N  6 days ago
@Trisha Anand so for postgres what would a user need to type to achieve this same query?

Trisha Anand  5 days ago
@Nikhil N: https://docs.appsmith.com/datasource-reference/querying-postgres#using-arrays-in-prepared-statement

Nikhil N  5 days ago
@Trisha Anand but this basically says use ANY and don't use IN

Nikhil N  5 days ago
How would anybody know that?

Trisha Anand  5 days ago
Should we solve for this problem using linting so that this is discoverable for raw queries with prepared statements turned on?
In case of UQI with forms, since we would be generating the raw query, this would not be a problem.

Nikhil N  5 days ago
Yeah forms will solve this and then the user can learn that the form generated the code as ANY.

Leo Thomas  1 day ago
Hi @Sumit can we meet/discuss at 12:00pm , have sent meeting invite

Sumit  1 day ago
@Leo Thomas DMed you regarding the meeting.
:+1:
1


Sumit  24 hours ago
@Shri @Nidhi summarising my understanding here. Please let me know if corrections are required:
If the user types in .... IN {{binding}}  then we check that {{binding}}  is an array of values and we do the re-formatting of the query to bring it to the correct format.
if the user types in ... IN ( {{binding}} ) then we don't reformat the {{binding}} value and substitute the value as it is.

Shri  22 hours ago
Can we have the first behaviour irrespective of whether parenthesis were used? If {{binding}} resolves to a string or a number, then {{binding}} and ({{binding}}) behave the same right? The parens don’t make a difference there, for all I know. So, I’d think it should make no difference here either. What say?`

leotom2000:
@leotom2000 had meeting with @Sumit and @Nidhi yesterday( 28/03/2022)
Need to discuss again as the array inputs could be a JSON input or Widget Input.

sumitsum:
Closing it since we haven't been able to arrive at the desired implementation idea.

PR #11969, opened 2022-03-18
fix: Fix replicaset checking with limited permissions

sharat87:
When running the fat container, with an external MongoDB, where the user only has the following permission:

```js
            {
                "role" : "readWrite",
                "db" : "appsmith"
            }
```

The container fails to start. Although this permission is sufficient to work with `changeStream`s, we get the following error instead:

```
Checking Replica Set of external MongoDB
********************************************************************
*          MongoDB Replica Set is not enabled                      *
********************************************************************
```

This PR fixes the problem by checking for `changeStream` only in the DB that we have access to, and only on a specific collection, the `user` collection. This way, even if the user doesn't have permissions on the whole database, we can still work with it.

### DIFF
--- a/deploy/docker/utils/bin/check_replica_set.js
+++ b/deploy/docker/utils/bin/check_replica_set.js
@@ -6,12 +6,12 @@ async function exec() {
     useUnifiedTopology: true,
   });
 
-  let isReplicaSetEnabled = false
+  let isReplicaSetEnabled = false;
 
   try {
     isReplicaSetEnabled = await checkReplicaSet(client);
   } catch (err) {
-    console.error("Error trying to check replicaset", err)
+    console.error("Error trying to check replicaset", err);
   } finally {
     client.close();
   }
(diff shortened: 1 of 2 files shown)


deploy/docker/utils/bin/check_replica_set.js
```diff
@@ -5,43 +5,41 @@ async function exec() {
...
+  process.exit(isReplicaSetEnabled ? 0 : 1);
 }
 
 async function checkReplicaSet(client) {
   await client.connect();
   return await new Promise((resolve) => {
     try {
       client
+        .db()
+        .collection("user")
```
sharat87, line 28: Adding these two lines makes MongoDB client try to get a `changeStream` on a single collection, as opposed to the whole cluster, which the user might not have access to.

mohanarpit (approved):

sharat87:
Verified on release image.

PR #12011, opened 2022-03-21
fix: Remove unnecessary variables in Helm chart & refactor templates

geekup-legodevops:
## Description
- Remove unnecessary variables in Helm chart
- Refactor Helm templates to remove usage of unnecessary variables (e.g `useStatefulSet` , Appsmith Helm chart will go along with persistent data - configuration, MongoDB data,..., so, its kind should be `StatefulSet` all the times instead of checking variable and assign kind to it)
- Rename template `deployment` to `statefulset`
- Update Helm chart README
## Type of change

- Fix (non-breaking change which fixes an issue)

## Checklist:


deploy/helm/values.yaml
```diff
@@ -174,13 +167,6 @@ resources:
   limits: {}
   requests: {}
 
-autoscaling:
```
sharat87, line 177: Can you elaborate on how we can configure autoscaling if this is removed?

deploy/helm/templates/statefulset.yaml
```diff
@@ -85,10 +76,6 @@ spec:
   {{- if not .Values.persistence.enabled }}
       - name: data
         emptyDir: {}
-  {{- else if not .Values.useStatefulSet }}
-      - name: data
-        persistentVolumeClaim:
```
sharat87, line 90: What disk will be used if a PV claim is not setup? Will it mean loss of data when the pod is restarted due to an auto-update?

sharat87 (approved):

PR #12013, opened 2022-03-21
chore: Ignore actions that are already migrated

nidhi-nair:


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
```diff
@@ -138,14 +140,18 @@ public void updateActionFormDataPath(MongockTemplate mongockTemplate) {
...
+            try {
+                if (pluginMap.get(uqiAction.getPluginId()).equals("firestore-plugin")) {
+                    migrateFirestoreActionsFormData(uqiAction);
+                } else if (pluginMap.get(uqiAction.getPluginId()).equals("amazons3-plugin")) {
+                    migrateAmazonS3ActionsFormData(uqiAction);
+                } else {
+                    migrateMongoActionsFormData(uqiAction);
+                }
+            } catch (AppsmithException e) {
+                // This action is already migrated, move on
```
sharat87, line 152: @nidhi-nair Can we at least log the error, instead of silently ignoring it?

PR #12034, opened 2022-03-21
fix: Reconnect application showing for existing datasources due to missing flag value

### LINKED ISSUE #12041: [Bug]: Reconnect Application is shown existing datasources

### Description

Open any app in release and notice the `Reconnect Application` in active datasources tab. 
image

### Steps To Reproduce

Open any app in release and notice the `Reconnect Application` in active datasources tab. 

### Public Sample App

Any app on release env.

### Version

Release

AnaghHegde:
# Description

> As part of git import we have introduced reconnect option to inform the user that datasources needs to be reconnected. Since there is no migration for the existing datasources all the datasource are shown with reconnect CTA. 

Fixes #12041

## Type of change

## How Has This Been Tested?

- Locally

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
```diff
@@ -5072,6 +5072,17 @@ public void createSystemThemes2(MongockTemplate mongockTemplate) throws IOExcept
...
     }
+
+    @ChangeSet(order = "118", id = "add-isCOnfigured-flag-for-all-datasources", author = "")
+    public void updateIsConfiguredFlagForAllTheExistingDatasources(MongockTemplate mongockTemplate) {
+        Query query = new Query();
+        query.addCriteria(Criteria.where("deleted").is(FALSE));
+
+        for (Datasource datasource : mongockTemplate.find(query, Datasource.class)) {
+            datasource.setIsConfigured(TRUE);
+            mongockTemplate.save(datasource);
```
abhvsn, line 5083: Instead of save can we please use the update method and use `fields().include({requiredField})` for the above query. This way we can avoid the OOM exception that might probably occur on prod as we have huge structures inside the datasource documents.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
```diff
@@ -5075,12 +5075,19 @@ public void createSystemThemes2(MongockTemplate mongockTemplate) throws IOExcept
 
     @ChangeSet(order = "118", id = "add-isCOnfigured-flag-for-all-datasources", author = "")
     public void updateIsConfiguredFlagForAllTheExistingDatasources(MongockTemplate mongockTemplate) {
-        Query query = new Query();
-        query.addCriteria(Criteria.where("deleted").is(FALSE));
+        final Query datasourceQuery = query(where(fieldName(QDatasource.datasource.deleted)).ne(true));
```
abhvsn, line 5078: I'm not sure but does it make sense to take `invalids` input to decide if the datasource is correctly configured or not?

abhvsn (approved):

PR #12043, opened 2022-03-21
chore: Upgrading boxen dependency

mohanarpit:
Upgrading the boxen dependency because a transitive dependency ansi-regex needs to be upgraded. If the boxen dependency is not being used, we can remove it from the dependency list as well.

### DIFF
--- a/deploy/docker/utils/bin/index.js
+++ b/deploy/docker/utils/bin/index.js
@@ -1,6 +1,6 @@
 #! /usr/bin/env node
 
-const yargs = require('yargs');
+const process = require('process');
 const utils = require('./utils');
 const export_db = require('./export_db.js');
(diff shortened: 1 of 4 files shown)


sharat87 (changes requested):
@mohanarpit yeah looks like it's unused. Let's remove it.

Also, I'd suggest we remove `chalk` and `yargs` as well since they aren't exactly critical for the usecase and the fewer dependencies we have the better. I can remove those in a separate PR if preferred.

sharat87:
@mohanarpit `yargs` and `chalk` are actually used in the code, although superficially. So I removed the usages as well.

I think we are ready to commit, but just have a look at the latest commit and we're good.

sharat87 (approved):

PR #12138, opened 2022-03-22
fix: Allow Multiselect DefaultValue to accept numbers

### LINKED ISSUE #11714: [Bug]: Multi select dropdown does not pre populate single values .

### Current Behavior

I  have  a multi  dropdown  list with some integer values . I am able to select and save single as well as multi values . But while am trying to pre populate these values from my  db query , the single value is not prepopulating to the list.

Screenshot (26)

### Steps To Reproduce

1. Create a page  with multiselect dropdown widget  and set values into it.
2. Select single value from the dropdown and save to  database.
3. Create  a modal for  edit  these values  and set default values from the database query.

### Environment

Production

### Version

Self-Hosted

Tooluloope:
## Description
Allow number to be passed as default Value in Multiselect

Fixes #11714

## Type of change

## How Has This Been Tested?

- Unit Test

## Checklist:

    // Code coverage diff between base branch:release and head branch: fix/multiselect-prepopulate-value-with-number 
 :green_circle: | total | 55.77 **(0.01)** | 37.22 **(0.02)** | 35.36 **(0)** | 56.02 **(0)**
 :green_circle: | app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx | 80.46 **(0.46)** | 56.67 **(6.67)** | 70.59 **(0)** | 80.49 **(0.49)**</details>


app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -22,8 +22,8 @@ export function defaultOptionValueValidation(
   props: MultiSelectWidgetProps,
   _: any,
```
ashit-rath, line 23: Instead of `any` here try using `LoDashStatic`. This would help remove any `as string` or `as []` type assertion which we should try to avoid as much as possible.

After making this change please do go through the whole validation function and see if all the `as` assertions can be removed.

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -67,11 +67,10 @@ export function defaultOptionValueValidation(
     }
   }
 
-  if (_.isString(value) && (value as string).trim() === "") {
-    isValid = true;
-    parsed = [];
-    message = "";
-  } else if (Array.isArray(value)) {
+  /*
+   * When value is "['green', 'red']", "[{label: 'green', value: 'green'}]" and "green, red"
```
ashit-rath, line 71: This block won't deal with strings, right? If not, can we update the comment?

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -107,6 +106,35 @@ export function defaultOptionValueValidation(
...
+  }
+
+  /*
+   * When value is an empty string
+   */
+  if (_.isString(value) && (value as string).trim() === "") {
+    isValid = true;
+    parsed = [];
+    message = "";
+  } else if (_.isNumber(value)) {
```
ashit-rath, line 124: Can we add a comment in[ line 52 ](https://github.com/appsmithorg/appsmith/pull/12138/files#diff-c321b0cc1a03b818cf65267b761ad882f25a18957c398e7671f7a61434f3a9dcR52)stating that when we get a number as a value like `"100"`, it's a valid JSON and gets parsed to a number whereas `"test"` is not a valid JSON and we treat it as a CSV?

I feel like this particular value is an edge case and I think we should explicitly mention that.

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -107,6 +106,35 @@ export function defaultOptionValueValidation(
...
+    /*
+     * When value is a number
+     */
+    isValid = true;
+    parsed = [value];
+    message = "";
+  } else if (_.isString(value)) {
```
ashit-rath, line 131: We are checking `_.isString` at 2 places can we combine both into one?

app/client/src/widgets/MultiSelectWidgetV2/widget/index.test.tsx
```diff
@@ -26,6 +26,18 @@ describe("defaultOptionValueValidation - ", () => {
     });
   });
 
+  it("should get tested with a number", () => {
```
ashit-rath, line 29: Can we add test cases for singular values like `"100"` and `"test"` if not present?

ashit-rath (commented):
Looks good Tolu!
Wanted to know have we made sure that this takes care of all the cases for default value we were discussing the other day?

ashit-rath:
@Tooluloope Can we make the title more descriptive? Would help with the release notes
Also, could you also make sure the checklist in the PR description is also marked correctly?

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
Tooluloope: it actually does, 

```
"['green', 'red']", "[{label: 'green', value: 'green'}]" and "green, red" 
is converted to

['green', 'red'] , [{label: 'green', value: 'green'}] , ['green', 'red']

```

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
Tooluloope: Should we parse "100" to a number or leave it as a string?

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -67,51 +71,64 @@ export function defaultOptionValueValidation(
...
        */
-      if (hasUniqueValues(value.map((val) => val.value) as [])) {
+      if (hasUniqueValues(value.map((val) => val.value))) {
         isValid = true;
         parsed = value;
-        message = "";
       } else {
-        isValid = false;
         parsed = [];
         message = "path:value must be unique. Duplicate values found";
```
ashit-rath, line 98: The message here is a bit different from the one in line 87. Can we move this message to a constant like 

```js
const UNIQUENESS_MESSAGE = "..."
```
Also, can we do the same thing for the messages in line 105 and line 134?
 Putting it in constant makes it DRY and less prone to inconsistency.

Tooluloope: The reason behind this is that The first one is an array of strings, while the other is an array of objects hence the use of `path`.

Also for consistency, `path: value` is also used in Radio Widget.

ashit-rath: Got it!

ashit-rath (approved):

YogeshJayaseelan:
@Tooluloope Tested this PR and it is working as expected

PR #12173, opened 2022-03-23
fix: Check Redis URL for `localhost` as well

sharat87:
Currently, to check if Redis URI is local or not, we only check for `127.0.0.1` and not for `localhost`, and we also check for the port to be 6379, which doesn't actually have to be.

This change is actually copied from line 71, where we perform a similar check for MongoDB URI.


sharat87:
@mohanarpit @sub1983 I'll need this to go in for the fat-container migration document.

mohanarpit (approved):

PR #12174, opened 2022-03-23
 fix: Add file system locks after the git operation

### LINKED ISSUE #10856: Feature : (POC) Implement File lock mechanism for git operations
- [x] Research on JGIT about lock needs to be done.
- [x] Implement a redis based request lock. 

Findings - 

```
JGit implements the same locking protocol native git implements in order to ensure interoperability
between different git processes acting on the same repository concurrently. If there's a stale index lock file
remaining after a crash you need to ensure there's no other git process having an open file handle on the
index lock file then the stale lock file can be deleted.

```

Sentry Issue: [APPSMITH-1QN](https://sentry.io/organizations/appsmith/issues/2984738943/?referrer=github_integration)

### LINKED ISSUE #14366: JGitInternalException: Exception caught during execution of reset command. Cannot lock /data/git-storage/628e37967901344...
Sentry Issue: [APPSMITH-1SC](https://sentry.io/organizations/appsmith/issues/3006439004/?referrer=github_integration)

```
LockFailedException: Cannot lock /data/git-storage/628e37967901344ba8d296e3/629f4d18360bee1cd628fe46/suzuya-studio-rental/.git/index. Ensure that no other process has an open file handle on the lock file /data/git-storage/628e37967901344ba8d296e3/629f4d18360bee1cd628fe46/suzuya-studio-rental/.git/index.lock, then you may delete the lock file and retry.
    at org.eclipse.jgit.dircache.DirCache.lock(DirCache.java:221)
    at org.eclipse.jgit.dircache.DirCache.lock(DirCache.java:292)
    at org.eclipse.jgit.dircache.DirCache.lock(DirCache.java:258)
    at org.eclipse.jgit.lib.Repository.lockDirCache(Repository.java:1282)
    at org.eclipse.jgit.api.ResetCommand.checkoutIndex(ResetCommand.java:407)
...
(12 additional frame(s) were not displayed)

JGitInternalException: Exception caught during execution of reset command. Cannot lock /data/git-storage/628e37967901344ba8d296e3/629f4d18360bee1cd628fe46/suzuya-studio-rental/.git/index. Ensure that no other process has an open file handle on the lock file /data/git-storage/628e37967901344ba8d296e3/629f4d18360bee1cd628fe46/suzuya-studio-rental/.git/index.lock, then you may delete the lock file and retry.
    at org.eclipse.jgit.api.ResetCommand.call(ResetCommand.java:209)
```

AnaghHegde:
## Description

> Add file locking using redis with retry. Helpful when multiple users are trying to perform git actions in the Appsmith UI.

Fixes #10856 
Fixes #14366

## Type of change

## How Has This Been Tested?

- Test A
- Test B

## Checklist:


app/server/appsmith-git/src/main/java/com/appsmith/git/service/GitExecutorImpl.java
```diff
@@ -110,6 +110,7 @@
                         .setCommitter(finalAuthorName, finalAuthorEmail)
                         .call();
                 processStopwatch.stopAndLogTimeInMillis();
+                git.close();
```
abhvsn, line 113: I don't think we need to close the methods when the git object is initialized in try with resource statement. Here is the official docs around close method in AutoCloseable class states:
```
Closes this resource, relinquishing any underlying resources. This method is invoked automatically on objects managed by the try-with-resources statement.
```

app/server/appsmith-git/src/main/java/com/appsmith/git/service/GitExecutorImpl.java
```diff
@@ -243,6 +245,7 @@ private Path createRepoPath(Path suffix) {
             String branchName = git.getRepository().getBranch();
 
             repositoryHelper.updateRemoteBranchTrackingConfig(branchName, git);
+            git.getRepository().close();
```
abhvsn, line 248: Why are we explicitly closing the repository here? Can you share with me the doc for this if any?

app/server/appsmith-git/src/main/java/com/appsmith/git/service/GitExecutorImpl.java
```diff
@@ -312,7 +311,8 @@ private Path createRepoPath(Path suffix) {
             // We can safely assume that repo has been already initialised either in commit or clone flow and can directly
             // open the repo
             Path baseRepoPath = createRepoPath(repoSuffix);
-            try (Git git = Git.open(baseRepoPath.toFile())) {
+            Git git = Git.open(baseRepoPath.toFile());
```
abhvsn, line 314: As requested earlier can we please use try with resource which will leverage the autocloseable functionality under the hood so that we don't have to worry about closing the `Git` object

app/server/appsmith-git/src/main/java/com/appsmith/git/service/GitExecutorImpl.java
```diff
@@ -624,6 +627,7 @@ private Path createRepoPath(Path suffix) {
                     mergeStatus.setReferenceDoc(ErrorReferenceDocUrl.GIT_MERGE_CONFLICT);
                 }
                 mergeStatus.setStatus(mergeResult.getMergeStatus().name());
+                git.close();
```
abhvsn, line 630: No need to add this here as per previous comment

app/server/appsmith-git/src/main/java/com/appsmith/git/service/GitExecutorImpl.java
```diff
@@ -691,6 +696,7 @@ private Path createRepoPath(Path suffix) {
             Ref ref = git.reset().setMode(ResetCommand.ResetType.HARD).call();
             // Remove untracked files
             git.clean().setForce(true).setCleanDirectories(true).call();
+            git.close();
```
abhvsn, line 699: Don't close the repo object here as the parent method from which the method is called might still needs open repo.

abhvsn (commented):
Should we discuss on this more because the implementation doesn't look like the way we had discussed earlier.

The way we are implementing this, is by storing the transient data in the Redis, and before making any file updates we will check if the path is available. This will be checked as a first step before doing any fancy stuff in the git-flow. We will be implementing the retries on the backend side after each second for next 9-10 secs (we came up with this number as remote calls is taking around 3-6 sec) and if this results in the failure for all the attempts we will respond back with filepath locked exception, where user will have to trigger the flow from UI. In case of success, API will acquire a lock and release it once the operation is completed. In case of API execution end up in unforeseen results like NPE, unknown error state we will be utilising the global-exception handler to release the lock, also as a safety measure we will automatically release the lock after 20sec as a fallback to avoid locking the filesystem infinitely.

app/server/appsmith-git/src/main/java/com/appsmith/git/service/GitExecutorImpl.java
abhvsn: @AnaghHegde why do we want to handle the open and close methods for git objects, I'm unable to understand the significance of handling this on our end?

abhvsn (changes requested):
Can we please create a separate file for Redis operations which we can leverage for other processes as well to avoid code duplication?

AnaghHegde:
@abhvsn That is a good point. I will move the redis operations to a different class.

abhvsn:
@AnaghHegde I ca see you have created a separate method for redis ops, can we move this to separate class where we can have CRUD implemented based on key-value pair.

app/server/appsmith-git/src/main/java/com/appsmith/git/service/GitExecutorImpl.java
```diff
@@ -542,6 +543,7 @@ public Mono<String> mergeBranch(Path repoSuffix, String sourceBranch, String des
                     git.getRepository().writeMergeCommitMsg(null);
                     git.getRepository().writeMergeHeads(null);
                     processStopwatch.stopAndLogTimeInMillis();
+                    git.close();
```
abhvsn, line 546: Please remove it as not required

app/server/appsmith-git/src/main/java/com/appsmith/git/service/GitExecutorImpl.java
```diff
@@ -672,7 +673,8 @@ public Mono<String> checkoutRemoteBranch(Path repoSuffix, String branchName) {
                 config.setString("branch", branchName, "remote", "origin");
                 config.setString("branch", branchName, "merge", "refs/heads/" + branchName);
                 config.save();
-                return git.getRepository().getBranch();
+                String branch = git.getRepository().getBranch();
+                return branch;
```
abhvsn, line 677: Any specific reason for this, just wanted to avoid un-necessary warnings 🙂

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/RedisUtils.java
```diff
@@ -0,0 +1,38 @@
...
+public class RedisUtils {
+    private static ReactiveRedisOperations<String, String> redisOperations;
+
+    private final static String REDIS_FILE_LOCK_VALUE= "inUse";
+
+    private final static String REDIS_FILE_RELEASE_VALUE = "isFree";
+
+    private final static Duration FILE_LOCK_TIME_LIMIT = Duration.ofSeconds(20);
+
+    public static Mono<Boolean> addFileLock(Path baseRepoSuffix) {
```
abhvsn, line 23: Can we keep the key as `defaultApplicationID` instead of complete path? Asking this because in future if the path altered we will have to invalidate all the entries and it's not bringing in much value ATM

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/RedisUtils.java
```diff
@@ -0,0 +1,38 @@
...
+                        return Mono.error(new AppsmithException(AppsmithError.GIT_FILE_IN_USE));
+                    }
+                })
+                .switchIfEmpty(redisOperations.opsForValue().set(baseRepoSuffix.toString(), REDIS_FILE_LOCK_VALUE, FILE_LOCK_TIME_LIMIT));
+    }
+
+    public static Mono<Boolean> releaseFileLock(Path baseRepoSuffix) {
+        return redisOperations.opsForValue().set(baseRepoSuffix.toString(), REDIS_FILE_RELEASE_VALUE, FILE_LOCK_TIME_LIMIT);
+    }
+}
```
abhvsn, line 38: Instead of including file in the method names can we keep these methods generic like CRUD? That way it will be clear that we can use this class for all the CRUD ops on redis

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -394,6 +398,13 @@ public Mono<String> commitApplication(GitCommitDTO commitDTO, String defaultAppl
 
         boolean isSystemGenerated = isSystemGeneratedTemp;
         Mono<String> commitMono = this.getApplicationById(defaultApplicationId)
+                .flatMap(application -> {
+                    GitApplicationMetadata gitData = application.getGitApplicationMetadata();
+                    Path baseRepoSuffix =
+                            Paths.get(application.getOrganizationId(), gitData.getDefaultApplicationId(), gitData.getRepoName());
+                    return RedisUtils.addFileLock(baseRepoSuffix)
```
abhvsn, line 405: Can you help me understand how the retries will work in case the file is already locked?

app/server/appsmith-server/src/main/java/com/appsmith/server/exceptions/AppsmithError.java
```diff
@@ -139,6 +139,7 @@ public enum AppsmithError {
     GENERIC_JSON_IMPORT_ERROR(400, 4049, "Unable to import application in organization {0} with error {1}", AppsmithErrorAction.DEFAULT, null, ErrorType.BAD_REQUEST, null),
     FILE_PART_DATA_BUFFER_ERROR(500, 5017, "Failed to upload file with error: {0}", AppsmithErrorAction.DEFAULT, null, ErrorType.BAD_REQUEST, null),
     MIGRATION_ERROR(500, 5018, "This action is already migrated", AppsmithErrorAction.DEFAULT, null, ErrorType.INTERNAL_ERROR, null),
+    GIT_FILE_IN_USE(500, 5019, "File lock in use. Please wait for few seconds and try again", AppsmithErrorAction.DEFAULT, null, ErrorType.GIT_ACTION_EXECUTION_ERROR, null),
```
abhvsn, line 142:
```suggestion
    GIT_FILE_IN_USE(500, 5020, "File lock in use. Please wait for few seconds and try again", AppsmithErrorAction.DEFAULT, null, ErrorType.GIT_ACTION_EXECUTION_ERROR, null),
```

abhvsn: We are already using 5019 in `ENV_FILE_NOT_FOUND`

app/server/appsmith-server/src/test/java/com/appsmith/server/services/GitServiceTest.java
```diff
@@ -2909,4 +2907,40 @@ public void deleteBranch_cancelledMidway_success() throws GitAPIException, IOExc
...
+                .thenReturn(Mono.just("pushed successfully"));
+
+        // First request for commit operation
+        Mono<String> commitMonoReq1 = gitService.commitApplication(commitDTO, gitConnectedApplication.getId(), DEFAULT_BRANCH);
+        // Second request for commit operation
+        Mono<String> commitMonoReq2 = gitService.commitApplication(commitDTO, gitConnectedApplication.getId(), DEFAULT_BRANCH);
+
+        // Both the request to execute completely without the file lock error from jgit.
+        StepVerifier
+                .create(Mono.zip(commitMonoReq1, commitMonoReq2))
```
abhvsn, line 2939: I'm not sure if the testcase really test the flow. I'm not completely sure but even if the file locks are removed this testcase should pass, as we are mocking time consuming methods like `pushApp` and `saveAppToLocalRepo`.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -136,6 +141,15 @@ public class GitServiceCEImpl implements GitServiceCE {
     private final static String GIT_PROFILE_ERROR = "Unable to find git author configuration for logged-in user. You can" +
             " set up a git profile from the user profile section.";
 
+    private final static String REDIS_FILE_LOCK_VALUE= "inUse";
+
+    private final static String REDIS_FILE_RELEASE_VALUE = "isFree";
+
+    private final static Duration FILE_LOCK_TIME_LIMIT = Duration.ofSeconds(20);
```
abhvsn, line 148: Is it still required? If not can we please remove this.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -581,14 +622,9 @@ public Mono<String> commitApplication(GitCommitDTO commitDTO, String defaultAppl
...
-                .subscribe(sink::success, sink::error, null, sink.currentContext())
-        );
-    }
-
-    @Override
-    public Mono<String> commitApplication(GitCommitDTO commitDTO, String defaultApplicationId, String branchName) {
-        return this.commitApplication(commitDTO, defaultApplicationId, branchName, false);
+        return Mono.create(sink -> {
+            commitMono.subscribe(sink::success, sink::error, null, sink.currentContext());
+        });
```
abhvsn, line 627: Please add a central logic to remove the locks, as we can easily miss this check-in future? Even in the current implementation, we have missed this for  
`throw new AppsmithException(AppsmithError.GIT_APPLICATION_LIMIT_ERROR);`

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -2481,4 +2579,20 @@ FieldName.ORGANIZATION_ID, defaultIfNull(application.getWorkspaceId(), ""),
...
+                    } else {
+                        return Mono.error(new AppsmithException(AppsmithError.GIT_FILE_IN_USE));
+                    }
+                })
+                .switchIfEmpty(redisOperations.opsForValue().set(key, REDIS_FILE_LOCK_VALUE, FILE_LOCK_TIME_LIMIT));
+    }
+
+    public Mono<Boolean> releaseFileLock(String key) {
+        return redisOperations.opsForValue().set(key, REDIS_FILE_RELEASE_VALUE, FILE_LOCK_TIME_LIMIT);
+    }
```
abhvsn, line 2597: Let's remove this as we have already created another class for the same.

abhvsn (changes requested):
Added minor enhancements.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: I tried this approach and I could not get it working due to the initialisation problem.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/GitServiceTest.java
AnaghHegde: I will try to add a delay to the mocked methods(if possible), this way the response would be delayed.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: Yes this is required

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -136,6 +143,15 @@ public class GitServiceCEImpl implements GitServiceCE {
     private final static String GIT_PROFILE_ERROR = "Unable to find git author configuration for logged-in user. You can" +
             " set up a git profile from the user profile section.";
 
+    private final static String REDIS_FILE_LOCK_VALUE= "inUse";
```
mohanarpit, line 146: Can we avoid duplicating these constant values in different classes? If we need to re-use a configuration value, please create a config class & auto wire it to any other classes that need to use the value. This will be much less error-prone.

app/server/appsmith-git/src/main/java/com/appsmith/git/service/GitExecutorImpl.java
```diff
@@ -331,6 +330,8 @@ public Mono<Boolean> checkoutToBranch(Path repoSuffix, String branchName) {
                         .getName();
                 processStopwatch.stopAndLogTimeInMillis();
                 return StringUtils.equalsIgnoreCase(checkedOutBranch, "refs/heads/"+branchName);
+            } catch (Exception e) {
+                throw new Exception(e);
```
mohanarpit, line 334: Any reason we are throwing a naked Exception here instead of AppsmithException?

app/server/appsmith-git/src/main/java/com/appsmith/git/service/GitExecutorImpl.java
```diff
@@ -318,7 +317,7 @@ public Mono<Boolean> checkoutToBranch(Path repoSuffix, String branchName) {
             // We can safely assume that repo has been already initialised either in commit or clone flow and can directly
             // open the repo
             Path baseRepoPath = createRepoPath(repoSuffix);
-            try (Git git = Git.open(baseRepoPath.toFile())) {
+            try(Git git = Git.open(baseRepoPath.toFile())) {
```
mohanarpit, line 320: Nit: Is this line formatted as per the standard code formatting rules?

app/server/appsmith-server/src/main/java/com/appsmith/server/exceptions/GlobalExceptionHandler.java
```diff
@@ -187,7 +222,22 @@ public Mono<ResponseDTO<ErrorDTO>> catchException(Exception e, ServerWebExchange
...
         doLog(e);
-        return Mono.just(new ResponseDTO<>(appsmithError.getHttpErrorCode(), new ErrorDTO(appsmithError.getAppErrorCode(),
+        String urlPath = exchange.getRequest().getPath().toString();
+        Mono<ResponseDTO<ErrorDTO>> response = Mono.just(new ResponseDTO<>(appsmithError.getHttpErrorCode(), new ErrorDTO(appsmithError.getAppErrorCode(),
                 appsmithError.getMessage())));
+
+        return getResponseDTOMono(urlPath, response);
+    }
+
+    private Mono<ResponseDTO<ErrorDTO>> getResponseDTOMono(String urlPath, Mono<ResponseDTO<ErrorDTO>> response) {
```
mohanarpit, line 232: As a practice, we should not pass `Mono` variables into functions. The signature of this function would be a lot cleaner if we did the following:

```suggestion
    private Mono<ResponseDTO<ErrorDTO>> getResponseDTOMono(String urlPath, ResponseDTO<ErrorDTO> response) {
```

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -538,8 +562,7 @@ public Mono<String> commitApplication(GitCommitDTO commitDTO, String defaultAppl
                                         error.getClass().getName(),
                                         error.getMessage(),
                                         childApplication.getGitApplicationMetadata().getIsRepoPrivate()
-                                )
-                                        .then(Mono.error(new AppsmithException(AppsmithError.GIT_ACTION_FAILED, "commit", error.getMessage())));
+                                ).then(Mono.error(new AppsmithException(AppsmithError.GIT_ACTION_FAILED, "commit", error.getMessage())));
```
mohanarpit, line 565: Nit: Suggestion to keep this in a separate line. This improves code readability by allowing developers to skim and find the next step.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1506,7 +1546,7 @@ public Mono<List<GitBranchDTO>> listBranchForApplication(String defaultApplicati
      * @param branchName           for which the status is required
      * @return Map of json file names which are added, modified, conflicting, removed and the working tree if this is clean
      */
-    public Mono<GitStatusDTO> getStatus(String defaultApplicationId, String branchName) {
+    private Mono<GitStatusDTO> getStatus(String defaultApplicationId, String branchName, boolean isFileLock) {
```
mohanarpit, line 1549: Please correct the functional doc comment when changing the signature of the function

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1517,15 +1557,29 @@ public Mono<GitStatusDTO> getStatus(String defaultApplicationId, String branchNa
             2. Fetch the current status from local repo
          */
 
+        Mono<GitStatusDTO> statusMono = getGitApplicationMetadata(defaultApplicationId)
+                .flatMap(gitApplicationMetadata -> {
+                    Mono<Tuple2<GitApplicationMetadata, Tuple2<Application, ApplicationJson>>> tupleMono = Mono.zip(
```
mohanarpit, line 1562: Can you help me understand this code better? To the best of my understanding, we are executing `statusMono` only to zip it immediately with another Mono function. Can we not simply zip them earlier?

app/server/appsmith-git/src/main/java/com/appsmith/git/service/GitExecutorImpl.java
```diff
@@ -331,6 +330,8 @@ public Mono<Boolean> checkoutToBranch(Path repoSuffix, String branchName) {
                         .getName();
                 processStopwatch.stopAndLogTimeInMillis();
                 return StringUtils.equalsIgnoreCase(checkedOutBranch, "refs/heads/"+branchName);
+            } catch (Exception e) {
```
mohanarpit, line 333: Is it possible to catch specific exceptions here instead of a generic Exception? Will help us handle the error better and return more readable error messages back to the user.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -347,7 +355,14 @@ private Mono<GitProfile> updateGitProfileWithAppsmithProfile(String key) {
      */
     @Override
     public Mono<String> commitApplication(GitCommitDTO commitDTO, String defaultApplicationId, String branchName, boolean doAmend) {
+        return this.commitApplication(commitDTO, defaultApplicationId, branchName, doAmend, true);
+    }
+
+    public Mono<String> commitApplication(GitCommitDTO commitDTO, String defaultApplicationId, String branchName) {
```
mohanarpit, line 361: Nit: Please add functional doc comments to these functions to explain what this function does. Since these functions are overriding each other, it's useful for future devs to understand the difference in use cases.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1517,15 +1557,29 @@ public Mono<GitStatusDTO> getStatus(String defaultApplicationId, String branchNa
...
-                applicationService.findByBranchNameAndDefaultApplicationId(finalBranchName, defaultApplicationId, MANAGE_APPLICATIONS)
-                        .onErrorResume(error -> {
-                            //if the branch does not exist in local, checkout remote branch
-                            return checkoutBranch(defaultApplicationId, finalBranchName);
-                        })
-                        .zipWhen(application -> importExportApplicationService.exportApplicationById(application.getId(), SerialiseApplicationObjective.VERSION_CONTROL)))
+                    if(Boolean.TRUE.equals(isFileLock)) {
+                        // Add file lock for the status API call to avoid sending wrong info on the status
+                        return redisUtils.addFileLock(gitApplicationMetadata.getDefaultApplicationId())
+                                .retryWhen(Retry.fixedDelay(MAX_RETRIES, RETRY_DELAY).jitter(0.75)
```
mohanarpit, line 1575: Can you help me understand why we are adding `jitter` in some of the retry calls but not in all?

app/server/appsmith-server/src/main/java/com/appsmith/server/exceptions/GlobalExceptionHandler.java
AnaghHegde: Yes makes sense. But I do have seen similar functions in the code base.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: IntelliJ formatter added this to prev line. But i will update this.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: Are you referring to the `tupleMono` variable? Yes, that's a good idea.

app/server/appsmith-git/src/main/java/com/appsmith/git/service/GitExecutorImpl.java
AnaghHegde: This is handled in the `GitServiceCEImpl` class. checkoutToRemoteBranch is used in many places and error messages are thrown as per the flow.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: As of now, jitter is applied to the very frequently used API class in git which is `status` API and `mergeability`
The rest of the APIs are not hit very often and at the same time. So just a retry with a fixed delay is enough. 
But the above two, say two users open the same app in edit mode - two status calls are made on the load. Then even when we switch the branch, again status call is made from client.

app/server/appsmith-server/src/main/java/com/appsmith/server/exceptions/AppsmithError.java
```diff
@@ -139,6 +139,7 @@ public enum AppsmithError {
     GENERIC_JSON_IMPORT_ERROR(400, 4049, "Unable to import application in organization {0} with error {1}", AppsmithErrorAction.DEFAULT, null, ErrorType.BAD_REQUEST, null),
     FILE_PART_DATA_BUFFER_ERROR(500, 5017, "Failed to upload file with error: {0}", AppsmithErrorAction.DEFAULT, null, ErrorType.BAD_REQUEST, null),
     MIGRATION_ERROR(500, 5018, "This action is already migrated", AppsmithErrorAction.DEFAULT, null, ErrorType.INTERNAL_ERROR, null),
+    GIT_FILE_IN_USE(500, 5020, "File lock in use. Please wait for few seconds and try again", AppsmithErrorAction.DEFAULT, null, ErrorType.GIT_ACTION_EXECUTION_ERROR, null),
```
mohanarpit, line 142:
```suggestion
    GIT_FILE_IN_USE(500, 5020, "File lock in use. Please wait for a few seconds and try again", AppsmithErrorAction.DEFAULT, null, ErrorType.GIT_ACTION_EXECUTION_ERROR, null),
```

app/server/appsmith-git/src/main/java/com/appsmith/git/service/GitExecutorImpl.java
mohanarpit: Can you please add a comment explaining this here then? I'm sure the next developer will ask the same question again.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1549,7 +1623,7 @@ public Mono<GitStatusDTO> getStatus(String defaultApplicationId, String branchNa
                                 Mono.just(repoSuffix)
                         );
                     } catch (IOException | GitAPIException e) {
-                        return Mono.error(new AppsmithException(AppsmithError.GIT_ACTION_FAILED, "status", e.getMessage()));
+                        return redisUtils.releaseFileLock(gitData.getDefaultApplicationId()).then(Mono.error(new AppsmithException(AppsmithError.GIT_ACTION_FAILED, "status", e.getMessage())));
```
mohanarpit, line 1626: Since we are catching & releasing file locks in the `GlobalExceptionHandler`, is this still needed?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1559,16 +1633,26 @@ public Mono<GitStatusDTO> getStatus(String defaultApplicationId, String branchNa
                                 .then(branchedStatusMono)
                                 // Remove any files which are copied by hard resetting the repo
                                 .then(gitExecutor.resetToLastCommit(tuple.getT3(), branchName))
-                                .then(branchedStatusMono)
+                                .flatMap(status -> {
+                                    if(Boolean.TRUE.equals(isFileLock)) {
```
mohanarpit, line 1637: There seems to be a logical error here. The if & else clause are both returning the exact same value. Can you please check this once?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1603,36 +1696,33 @@ public Mono<MergeStatusDTO> mergeBranch(String defaultApplicationId, GitMergeDTO
...
-                    .onErrorResume(error -> {
-                        log.error("Error in repo status check application " + defaultApplicationId, error);
-                        if (error instanceof AppsmithException) {
-                            return Mono.error(error);
-                        }
-                        return Mono.error(new AppsmithException(AppsmithError.GIT_ACTION_FAILED, "status", error));
-                    });
+                    return Mono.zip(Mono.just(defaultApplication), pathToFile)
+                            .onErrorResume(error -> {
+                                log.error("Error in repo status check application " + defaultApplicationId, error);
```
mohanarpit, line 1720:
```suggestion
                                log.error("Error in repo status check for application " + defaultApplicationId, error);
```

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1702,8 +1792,7 @@ public Mono<MergeStatusDTO> mergeBranch(String defaultApplicationId, GitMergeDTO
                             AnalyticsEvents.GIT_MERGE.getEventName(),
                             application,
                             application.getGitApplicationMetadata().getIsRepoPrivate()
-                    )
-                    .thenReturn(mergeStatusDTO);
+                    ).thenReturn(mergeStatusDTO);
```
mohanarpit, line 1795: I understand this change has happened due to auto formatting. Just asking if it's possible for us to keep `.thenReturn()` in a separate line?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1736,19 +1825,24 @@ public Mono<MergeStatusDTO> isBranchMergeable(String defaultApplicationId, GitMe
                             gitApplicationMetadata.getRepoName());
 
                     //1. Hydrate from db to file system for both branch Applications
-                    return this.getStatus(defaultApplicationId, sourceBranch)
+                    return redisUtils.addFileLock(gitApplicationMetadata.getDefaultApplicationId())
```
mohanarpit, line 1828: I can see that this function `redisUtils.addFileLock().retryWhen()` is being invoked 6 times in this class. Do you think it'll be cleaner to create a private function for this instead of copy pasting the code?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -2365,7 +2459,8 @@ private Mono<GitPullDTO> pullAndRehydrateApplication(Application defaultApplicat
         */
         GitApplicationMetadata gitData = defaultApplication.getGitApplicationMetadata();
         if (isInvalidDefaultApplicationGitMetadata(gitData)) {
-            return Mono.error(new AppsmithException(AppsmithError.INVALID_GIT_CONFIGURATION, GIT_CONFIG_ERROR));
+            return redisUtils.releaseFileLock(gitData.getDefaultApplicationId())
```
mohanarpit, line 2462: If we are releasing the Redis lock in `GlobalExceptionHandler`, is this required here?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -2387,15 +2482,15 @@ private Mono<GitPullDTO> pullAndRehydrateApplication(Application defaultApplicat
                             ))
                             .onErrorResume(error -> {
                                 if(error.getMessage().contains("conflict")) {
-                                    return Mono.error(new AppsmithException(AppsmithError.GIT_PULL_CONFLICTS, error.getMessage()));
+                                    return fileLock.then(Mono.error(new AppsmithException(AppsmithError.GIT_PULL_CONFLICTS, error.getMessage())));
                                 }
                                 else if (error.getMessage().contains("Nothing to fetch")) {
```
mohanarpit, line 2487: Nit: The `else-if` clause should be on the same line as the close of the previous if-clause. This reduces chances of a future developer making a mistake and thinking there is no else clause here.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -2411,7 +2506,7 @@ else if (error.getMessage().contains("Nothing to fetch")) {
 
                     return Mono.zip(pullStatusMono, Mono.just(branchedApplication), applicationJsonMono);
                 } catch (IOException e) {
-                    return Mono.error(new AppsmithException(AppsmithError.GIT_FILE_SYSTEM_ERROR, e.getMessage()));
+                    return fileLock.then(Mono.error(new AppsmithException(AppsmithError.GIT_FILE_SYSTEM_ERROR, e.getMessage())));
```
mohanarpit, line 2509: Same here. Should we release file locks here if we are doing it centrally at `GlobalExceptionHandler`?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1151,7 +1195,15 @@ public Mono<Application> createBranch(String defaultApplicationId, GitBranchDTO
                     }
                     Path repoSuffix = Paths.get(srcApplication.getWorkspaceId(), srcBranchGitData.getDefaultApplicationId(), srcBranchGitData.getRepoName());
                     // Create a new branch from the parent checked out branch
-                    return gitExecutor.checkoutToBranch(repoSuffix, srcBranch)
+
+                    Mono<Boolean> fileLockMono = redisUtils.releaseFileLock(srcBranchGitData.getDefaultApplicationId());
```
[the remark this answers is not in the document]
AnaghHegde: I missed removing this. It was added as part of prev implementation.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1559,16 +1633,26 @@ public Mono<GitStatusDTO> getStatus(String defaultApplicationId, String branchNa
...
-        return Mono.create(sink -> statusMono
-                .subscribe(sink::success, sink::error, null, sink.currentContext())
-        );
+        return Mono.create(sink -> {
+            statusMono.subscribe(sink::success, sink::error, null, sink.currentContext());
+        });
+    }
+
+    @Override
+    public Mono<GitStatusDTO> getStatus(String defaultApplicationId, String branchName) {
```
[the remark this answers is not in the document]
AnaghHegde: Yes. The difference is that status API is used internally and we do not need to acquire lock in such scenarios, hence overriding the method. If not we would have to hardcode the flag while calling this method from the controller. That sounds okay, but what if we need to add one more parameter which makes sense only for some of the flows like merge etc and not for status API? We would end up hardcoding these values in git controller.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: It does not make sense now as we are releasing the lock in a central place.

AnaghHegde:
@mohanarpit Can you please review this PR again?

trishaanand:
@AnaghHegde : Could you please resolve the merge conflicts? And re-run the ok-to-test suite?

mohanarpit (dismissed):

mohanarpit:
@AnaghHegde This PR is approved. Please work with the team to merge this in when appropriate.

mohanarpit (approved):

Parthvi12:
Following are the issues that were observed on EC2 instance for this PR:

- [ ] Discard is struck when 2 people discard the changes at same time, ideally it should show file lock error toast

- [ ] When multiple people pull changes together , the merge conflict error is shown instead of file lock error

- [ ] When two people try to merge at the same time, git sync modal goes blank and doesn't show any file lock error

- [ ] File lock error toast is randomly coming even when any operations are not performed

app/client/cypress/support/gitSync.js
```diff
@@ -156,8 +156,8 @@ Cypress.Commands.add("createGitBranch", (branch) => {
   cy.wait(3000);
   cy.get(gitSyncLocators.branchSearchInput).type(`{selectall}${branch}{enter}`);
   // increasing timeout to reduce flakyness
-  cy.get(".bp3-spinner", { timeout: 30000 }).should("exist");
-  cy.get(".bp3-spinner", { timeout: 30000 }).should("not.exist");
+  cy.get(".bp3-spinner", { timeout: 45000 }).should("exist");
```
Aishwarya-U-R, line 159: Why are we increasing wait time to 45 seconds? Is'nt it too much wait.

Parthvi12: Git operations will take more time after this PR, this is expected behaviour due to the file locks.

Aishwarya-U-R: Introducing file locks - why does it take more time to complete?

AnaghHegde: This is also because of introducing the Queue in client side for git api requests.

Parthvi12 (approved):

Issue #11946, opened 2022-03-17
[Bug]: APPSMITH_ENCRYPTION_PASSWORD and APPSMITH_ENCRYPTION_SALT not properly set on fresh installation, when provided via ENV-Variable


WernerDreierB:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

When I specify APPSMITH_ENCRYPTION_SALT  and APPSMITH_ENCRYPTION_PASSWORD as "applicationConfig" within helm-deployment and run a fresh installation:

- Installation does not find docker.env and will generate it (correct)
- an empty file with random salt and password is generated (correct)
- the genrated-password-variables are named equally as the corresponding env-variables and therefore hide the actual values from env-variables when https://github.com/appsmithorg/appsmith/blob/release/deploy/docker/entrypoint.sh#L31 is executed (wrong)
![2022-03-17 17_05_25-Clipboard](https://user-images.githubusercontent.com/15634299/158845484-ee89050c-15be-4cce-8b87-dc0af347268d.jpg)

### Steps To Reproduce

1. Set APPSMITH_ENCRYPTION_SALT  and APPSMITH_ENCRYPTION_PASSWORD 
2. run the deployment (make sure no PVC exists beforehand, so that a fresh installation happens)
3. open a shell into the running appsmith-container
4. Execute `rm /appsmith-stacks/configuration/docker.env`
5. Execute `cp entrypoint.sh entrypoint_debug.sh`
`sed -z -i "s^init_env_file() {\n^init_env_file() {\n  printenv | grep ENCRYPTION\n^g" entrypoint_debug.sh`
`sed -z -i "s^# Build an env^  printenv | grep ENCRYPTION\n# Build an env^g" entrypoint_debug.sh`
`chmod +x  entrypoint_debug.sh`
6. Execute the file to check output `./entrypoint_debug.sh` You will find that even though ENV-Vars are correctly set, the output during runtime is wrong
7. Execute `cat templates/pre-define.env | grep ENC` to see, that wrong (randomly generated values) are used

Suggested solution:
Rename temporary variables to be something that is not passed from the outside as ENV-Variable

### Version

Self-Hosted 1.6.15 nightly

PR #12187, opened 2022-03-23
fix: renamed temporary encryption variables in docker entry-point script

sum35h:
## Description
Renamed temporary variables `APPSMITH_ENCRYPTION_SALT` and `APPSMITH_ENCRYPTION_PASSWORD`  in appsmith/docker/deploy/entrypoint.sh to  `TEMP_APPSMITH_ENCRYPTION_SALT` and `TEMP_APPSMITH_ENCRYPTION_PASSWORD` to prevent overwriting the predefined ENV variables.

Fixes # [11946](https://github.com/appsmithorg/appsmith/issues/11946)

## Type of change

## Steps taken to test
 Ran the fixed image with defined ENV variables and verified it in `templates/pre-define.env` of the container.


sharat87 (changes requested):
Two points:

1. From the code, this issue would apply for the `APPSMITH_MONGODB_PASSWORD` env variable as well right? Can we fix it for that also?
2. Can we use local variables in this function for this? Like `local generated_encryption_password=...`, which will ensure that it doesn't overwrite any global variables in the future. Also, let's prefer using lowercase names like this for such local variables.

Let me know if I'm not making much sense and I'll try to rephrase. 😄

sum35h:
> Two points:
> 
> 1. From the code, this issue would apply for the `APPSMITH_MONGODB_PASSWORD` env variable as well right? Can we fix it for that also?
> 2. Can we use local variables in this function for this? Like `local generated_encryption_password=...`, which will ensure that it doesn't overwrite any global variables in the future. Also, let's prefer using lowercase names like this for such local variables.
> 
> Let me know if I'm not making much sense and I'll try to rephrase. 😄
I've made the suggested changes, I had left out the mongo variables because in case of custom mongo I thought we only specify the mongo uri directly, as mentioned in the docs. 
Also there is a new variable `BASIC_AUTH_PASSWORD` for supervisor, though it doesn't overwrite any value should I also do the same for it?

sharat87 (approved):

PR #12241, opened 2022-03-25
fix: Mongodb not accepting valid array update query

### LINKED ISSUE #11419: [Bug]-[2960]:MongoDB update query is not accepting valid value (an array) in update field

### Current Behavior

This query is valid for MongoDB 4.2+ 
```
db.getCollection('user').updateOne(
    {"_id":"1"    },
    [{"$set":{"newField":"$name"}}]
)
```
And in order to work correctly the `$set` part needs to be in an array.
But when we try to do something similar via Appsmith, It throws an error:
Screenshot 2022-02-24 145332

### Hint
https://github.com/appsmithorg/appsmith/pull/12241#discussion_r844997566

### Steps To Reproduce

Like the Image above:
1. Create an update query in a MongoDB datasource
2. Enter an update statement in array form like this one `[{"$set":{"testField":"$status"}}]`
3. It fails when you run the query with this error: ```Execution failed with status 5005
No valid mongo command found. Please select a command from the "Command" dropdown and try again ```

### Environment

Production

### Version

Cloud

devnamrits:
## Description

Initially, when we were passing the correct MongoDB update query written in array form eg. `[{"$set":{name: "Ram singh"} },{"$set": {age: 30}}]`, it was raising an error. After this change, the valid update query is working fine. Since `Document.parse` doesn't parse array. Hence I added a default key with a JSON structure pointing towards that array. As `Document. parse` will parse this JSON hence it works fine further. 

Fixes #11419 

## Type of change

## How Has This Been Tested?

Testing is being done by passing the update command of the type array from the frontend.
Screenshot:

Screenshot 2022-03-25 at 8 47 21 AM

## Checklist:


devnamrits:
@mojtab23 Any update?

sumitsum:
@devnamrits thanks for raising the PR. I will be reviewing it.

devnamrits:
Thank you @sumitsum for reviewing it.

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/commands/UpdateMany.java
```diff
@@ -45,6 +45,7 @@ public UpdateMany(ActionConfiguration actionConfiguration) {
 
         if (validConfigurationPresentInFormData(formData, UPDATE_OPERATION)) {
             this.update = (String) getValueSafelyFromFormData(formData, UPDATE_OPERATION);
+            this.update = "{\"updateKey\":" + this.update + "}";
```
sumitsum, line 48: Hi @devnamrits , this approach has a drawback. If we do this, then we can never validate any input as valid or invalid. e.g. If `this.update` is `"abc"` then this will be parsed as a valid input which is not intended. Essentially this change converts any input into a valid input. However, what we want is to be able to detect if the user input is valid or not. Up until now we have only tried to check if the input is a valid `Document` or not, which in this case is not sufficient since MongoDB 4.2 onwards it also accepts an array of Documents (pipleine) as well. Hence, IMO, the correct approach here should be to parse for two things - (1) valid Document (2) valid array of Documents. To achieve this one good way would be create a new method `parseSafely(fieldName, input., boolean)` along the lines of `parseSafely` method that already exists and use the last parameter to check if we need to validate for array of Documents as well.

sumitsum (changes requested):

sumitsum:
Can you please add a JUnit test case for this change in `MongoPluginTest.java` ?

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/commands/UpdateMany.java
devnamrits: Okay then I'll try to validate and parse as you mentioned. 

Also, I need to add those JUnit test cases after making these changes, Right??

sumitsum: @devnamrits thanks for following up on this. Yes, once you have made the necessary changes and manually tested them, you should then add one JUnit test case to test it as well.

sumitsum:
@devnamrits , can you please confirm if you are still working on this ? If you need help regarding the implementation, we can meet and discuss as well.

sumitsum:
@devnamrits please comment here if you are still working on it, otherwise, this issue may get re-assigned.

PR #12251, opened 2022-03-25
feat: Exclude configs from actions while committing to git repo

### LINKED ISSUE #11997: [Feature]: Exclude config values from actions while committing to git repo

### SubTasks

- Exclude the app slug from git repo as the imported app from git repo will have different name 
- For actions created from datasources of type SAAS, API we need to exclude the config values from committing to git repo
- For git import provide an option for the user to populate these values during the import. Needs discussion. 

Reference - https://theappsmith.slack.com/archives/CPQNLFHTN/p1647497578863639

AnaghHegde:
## Description
Exclude the app slug from git repo as the imported app from git repo will have different name.
For actions created from datasources of type SAAS, API we need to exclude the config values from committing to git repo

Fixes #11997 

## Type of change

## How Has This Been Tested?

- Locally

## Checklist:


abhvsn:
I'll let @nidhi-nair to review this PR as I don't have the complete context here 🙂

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
```diff
@@ -644,7 +644,25 @@ public void updateIsConfiguredFlagForAllTheExistingDatasources(MongockTemplate m
         }
     }
 
-    @ChangeSet(order = "005", id = "set-application-version", author = "")
+    @ChangeSet(order = "005", id = "add-isConfigStoredAtDataSource-for-plugins", author = "")
```
nidhi-nair, line 647: Let's maintain ordering here?

app/server/appsmith-server/src/test/java/com/appsmith/server/services/GitServiceTest.java
```diff
@@ -1415,7 +1416,7 @@ public void pullChanges_upstreamChangesAvailable_pullSuccess() throws IOExceptio
         gitStatusDTO.setBehindCount(0);
         gitStatusDTO.setIsClean(true);
 
-        Mockito.when(gitFileUtils.saveApplicationToLocalRepo(Mockito.any(Path.class), Mockito.any(ApplicationJson.class), Mockito.anyString()))
+        Mockito.when(gitFileUtils.saveApplicationToLocalRepo(Mockito.any(Path.class), Mockito.any(ApplicationJson.class), Mockito.anyString(), Mockito.anyList()))
```
nidhi-nair, line 1419: Could we please add a test for `true`, `false` and `null` values here?

nidhi-nair:
@AnaghHegde, for the remote plugin migration, we'll also have Twilio and a bunch of Hubspot plugins on prod. Could you please limit that to Airtable. Also let's talk to @hiteshjoshi and get a clear set of guidelines about when we're supposed to mark this value as true or false from a product perspective? We can use Twilio and Hubspot as examples to figure this out. This will help our SaaS (remote) team set this up independently.

Also, looks like our default behaviour for this is that we do copy over the config. Is that intended? If one of the plugins skips on this config because of an oversight, what behaviour do we want to default to?

AnaghHegde:
@nidhi-nair For the default behaviour yes we do copy over the config as it does not have any sensitive data and does not affect the import flow.For Twilio and Hubspot we can have a call to discuss this in detail.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/GitServiceTest.java
AnaghHegde: It won't make sense to include this here, as we are mocking the implementation here. But i am writing test for this in `GitFileUtilsTest` class, i will push this.

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Plugin.java
```diff
@@ -82,4 +82,6 @@
     @Transient
     Map<String, String> templates;
 
+    Boolean isConfigStoredAtDataSource;
```
nidhi-nair, line 85: Nit: Can we use lowercase for datasource? This is to maintain consistency with how we refer to datasources across Appsmith.

nidhi-nair (changes requested):
Just a tiny change in the name that'll help with maintenance

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Plugin.java
AnaghHegde: @nidhi-nair If you look at the [L.72 ](https://github.com/appsmithorg/appsmith/blob/release/app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Plugin.java#L72) we have used it in capital words.

nidhi-nair: Thanks for catching this, we should fix this reference as well!

mohanarpit:
@AnaghHegde Since the linked issue has been closed because we don't wish to follow the proposed implemenetation plan, is this PR still relevant? Can we close this PR?

AnaghHegde:
Closing this PR as we have decided to not to implement this feature.

Issue #12225, opened 2022-03-24
[Bug]-[300]:Default value is displayed on widget even when it is not available in drop-down options


vivekverma2312:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

![Uploading Screenshot 2022-03-24 at 2.28.19 PM.png…]()

https://watch.screencastify.com/v/rTmvAKEDYT6lAqffXIXq

### Steps To Reproduce

1. Drag and drop a Select/multi-select/TreeSelect/Multi tree select
2. Add any random value in Default Value
3. Observe that it is shown in widget even when the default value is not present in drop-down options

### Version

Cloud

Tooluloope:
@dilippitchika Should we close this issue.
Since this was intended

dilippitchika:
@Tooluloope i think we should be smarter about this because it's causing frequent confusion.

Here's my suggestion on how to fix this -
1. When user provides only value and not {label, value} in default, we check if the option exists. If it doesn't then we don't show this in default. 
2. When user provides {label, value} for default, we check if the option exists. If it doesn't then we NOW show this in default. This is the only scenario for a user to add default when they have configured the server side filtering in the widget. 
This is applicable both for select and multi-select widgets.

Let me know your thoughts

Tooluloope:
So Conclusion here was to enable this defaultValue caching only when Serverside Filtering is enabled. But once Serverside filtering is false, we do a strict check to know if the default value is in the list of options, and Only the default values in options would be displayed.

CC @dilippitchika @somangshu

dilippitchika:
## Stats
| Stat | Values |
| ------------- | ------------- |
| Reach     | 150 |
| Effort (months)  | 1 |

techbhavin:
this issue is covered in PR[#13079](https://github.com/appsmithorg/appsmith/pull/13079)

Tooluloope:
# QA Note

The following should be tested in the PR

- For all the Select Family, The value must be strictly checked against the options except if serverside filtering is enabled, i.e all selectedValues/DefaultValues must exist in the options for it to be reflected in `selectOptionValue`
- Connecting the defaultValue from a Table or external source, should reflect the expected value in `selectOptionValue`
- Getting the expected value in a Form.
- For Treeselect widgets, The Display Mode should reflect the right value

cc @ashit-rath @wmdev0808 @somangshu @dilippitchika

dilippitchika:
[...]

PR #12302, opened 2022-03-28
fix: 12225 default value behaviour in select and multi select widgets

prsidhu:
## Description

Issue - the widgets render the provided default value even if it is unavailable in the options list.
Fix - Filter valid selections in the widget.

Fixes # (issue)

https://github.com/appsmithorg/appsmith/issues/12225

## Type of change

## How Has This Been Tested?

1. Create a new app. 
2. Drag in a select widget. The widget will render GREEN text as provided by the default value.
3. Change the options list to not include the option { value: "GREEN", label: "GREEN" }. Keep the default value as the same.
4. Refresh the widget. Observe that the widget renders with a default message : "Select option".
5. Repeat the same for multi select widget.

## Checklist:

    // Code coverage diff between base branch:release and head branch: fix-12225 
 :red_circle: | total | 56.16 **(-0.02)** | 37.65 **(-0.03)** | 35.65 **(-0.01)** | 56.4 **(-0.02)**
 :red_circle: | app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx | 67.31 **(-13.15)** | 43.59 **(-13.08)** | 60 **(-10.59)** | 68.75 **(-11.74)**
 :red_circle: | app/client/src/widgets/SelectWidget/widget/index.tsx | 53.42 **(-1.51)** | 20.34 **(-1.88)** | 50 **(0)** | 54.29 **(-1.59)**</details>


Tooluloope:
Hi @prsidhu 
This was a known fix.
Using these widgets in Serverside filtering, 
You'll realize that a selected value might not be in the list of options when the widget/options are filtered

prsidhu:
@Tooluloope What do you suggest then? The logic could be ported to the component rather than the widget.

Tooluloope:
@prsidhu a conclusion was made here https://github.com/appsmithorg/appsmith/issues/12225#issuecomment-1082910936

prsidhu:
@Tooluloope thanks for intimation. I've made the changes. Could you review it again?

Tooluloope (changes requested):
Hi @prsidhu, You'll need to act on their derived properties `SelectedOptionValues` and `SelectedOptionlabels`.

Screenshot 2022-03-31 at 09 58 48

Tooluloope (changes requested):
https://www.loom.com/share/b37e89b1579b401fae23addcd5a69048

Could you look into this crash?

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -443,6 +450,7 @@ class MultiSelectWidget extends BaseWidget<
           isString(o) || isNumber(o) ? { value: o } : { value: o.value },
         )
       : [];
+    this.validateSelections(values);
```
Tooluloope, line 453: So this shouldn't be here.
To update the SelectedOptionValue, You'll do the filtering in the derived property, and also you filter the value getting into the component.

Your implementation would cause multiple rerenders, that's why it crashes

app/client/src/widgets/SelectWidget/widget/index.tsx
```diff
@@ -356,10 +356,20 @@ class SelectWidget extends BaseWidget<SelectWidgetProps, WidgetState> {
...
+      toString(this.props.selectedOptionValue) &&
+      selectedIndex === -1 &&
+      !this.props.serverSideFiltering
+    ) {
+      /**
+       * If the provided value (default value) is not available in the options,
+       * then clear the selection; unless options are being filtered on the server.
+       */
+      this.onOptionSelected({});
+    }
```
Tooluloope, line 372: this would also always run, I think you need to validate the value, if the value is empty you don't need to run this.

Tooluloope (changes requested):
Please add comprehensive tests for what you're doing.

Lastly, let's update the title and description to reflect the changes we are making

PR #12310, opened 2022-03-28
feat: Export App with credentials for sample apps and templates

### LINKED ISSUE #12296: Allow import and export application with credentials for some cases
Currently the exported application does not contain datasource credentials. User need to manually set them when they import an application. 

This adds a manual step which breaks fork template feature. For templates, we want users to quickly see it in action. We decided to share DB credentials so that users can try them out without doing anything. For this to work, we need import with credentials.

AnaghHegde:
## Description

Currently the exported application does not contain datasource credentials. User need to manually set them when they import an application. This adds a manual step which breaks fork template feature. For templates, we want users to quickly see it in action. We decided to share DB credentials so that users can try them out without doing anything. For this to work, we need import with credentials. 
This PR lets the server API's to export application with credentials. 

Fixes #12296 

## Type of change

## How Has This Been Tested?

- Locally
- Junit Tests

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationControllerCE.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationControllerCE.java
@@ -166,10 +166,11 @@ public class ApplicationControllerCE extends BaseController<ApplicationService, 
 
     @GetMapping("/export/{id}")
     public Mono<ResponseEntity<ApplicationJson>> getApplicationFile(@PathVariable String id,
-                                                                    @RequestHeader(name = FieldName.BRANCH_NAME, required = false) String branchName) {
+                                                                    @RequestHeader(name = FieldName.BRANCH_NAME, required = false) String branchName,
+                                                                    @RequestParam(required = false) String isSampleApps) {
         log.debug("Going to export application with id: {}, branch: {}", id, branchName);
 
-        return importExportApplicationService.exportApplicationById(id, branchName)
+        return importExportApplicationService.exportApplicationById(id, branchName, isSampleApps)
                 .map(fetchedResource -> {
                     String applicationName = fetchedResource.getExportedApplication().getName();
                     HttpHeaders responseHeaders = new HttpHeaders();
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
@@ -383,7 +383,7 @@ public class ImportExportApplicationServiceCEImpl implements ImportExportApplica
                                         .removeIf(datasource -> !dbNamesUsedInActions.contains(datasource.getName()));
 
                                 // Save decrypted fields for datasources for internally used sample apps and templates only
-                                if(true) {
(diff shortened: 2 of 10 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationControllerCE.java
```diff
@@ -166,10 +166,11 @@ public ApplicationControllerCE(
 
     @GetMapping("/export/{id}")
     public Mono<ResponseEntity<ApplicationJson>> getApplicationFile(@PathVariable String id,
-                                                                    @RequestHeader(name = FieldName.BRANCH_NAME, required = false) String branchName) {
+                                                                    @RequestHeader(name = FieldName.BRANCH_NAME, required = false) String branchName,
+                                                                    @RequestParam(required = false) String isSampleApp) {
```
abhvsn, line 170: Can we please make this a boolean and have the default value false:
`@RequestParam(required = false, defaultValue = "false") Boolean isSampleApp`

abhvsn (commented):
LGTM just a minor nit

abhvsn:
@AnaghHegde how this is going to work out for the sample apps after these have been forked? As per the current implementation we are asking the client to decide if we want the creds or not but IMO this should be from server side

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationControllerCE.java
AnaghHegde: The values passed as request param will be casted to string, i will cast this before using.

AnaghHegde:
@abhvsn I agree with you. But i am not aware of the templates internals. @nayan-rafiq Will handle this.

abhvsn:
Are we sure we don't want to handle the sample apps usecase here along with the templates?

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -111,10 +117,6 @@
         UNPUBLISHED, PUBLISHED
     }
 
-    private enum IdType {
-        RESOURCE_ID, DEFAULT_RESOURCE_ID
-    }
```
abhvsn, line 116: Cleanup as not required

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/HelperMethods.java
```diff
@@ -23,17 +24,19 @@ public static void updateArchivedAtByDeletedATForActions(List<NewAction> actionL
...
-                    final String pluginName = newAction.getPluginId();
-                    if ("mongo-plugin".equals(pluginName)) {
-                        DatabaseChangelog2.migrateMongoActionsFormData(newAction);
-                    } else if ("amazons3-plugin".equals(pluginName)) {
-                        DatabaseChangelog2.migrateAmazonS3ActionsFormData(newAction);
-                    } else if ("firestore-plugin".equals(pluginName)) {
-                        DatabaseChangelog2.migrateFirestoreActionsFormData(newAction);
-                    }
-                });
+        if (!CollectionUtils.isNullOrEmpty(actionList)) {
```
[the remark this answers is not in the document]
abhvsn: This is just in case if the imported json don't have actionList, this was throwing the npe earlier which resulted in failing testcase.

abhvsn (approved):

PR #12340, opened 2022-03-29
feat: Add UsagePulse API, create-only

sharat87:
The usage pulse API at `/usage-pulse` only accepts `POST` requests, for only creating records. No payload body is expected either. The object saved contains the currently logged-in user, and the timestamp.


sharat87:
cc @riodeuno

mohanarpit (approved):

PR #12351, opened 2022-03-29
chore: Should not persist DSL for mainContainer layout changes

### LINKED ISSUE #12290: [Task]: Do not update DSL for the rightColumn property

### SubTasks

- Do not update the DSL with rightColumn property of the main container as this is coming up in the git repo as un committed changed

rahulramesha:
## Description

What is happening?

MainConatiner's DSL(width and height information) is updated whenever the canvas is first loaded or whenever Window is resized. DSL is a persisted state and should not be updated for maintaining Temporary Information.

What are the Changes done?

The MainContainer's temporary information is being stored in a separate non persisted State. And When DSL is selected for rendering the canvas, the Info from non-persisted state, The MainContainer's values are modified and sent.

Why are these changes required?

Since the DSL is updated at every page load and browser resize, that causes and unnecessary backend call and whenever user commits the DSL to git, these changes are reflected in the commit, Which should not be the case.

Fixes #12290

## Type of change

Task

## How Has This Been Tested?

Manual verification

## Checklist:

    // Code coverage diff between base branch:release and head branch: gitSyncMainCanvasFix 
 :red_circle: | total | 56.32 **(-0.03)** | 37.63 **(-0.12)** | 35.91 **(-0.03)** | 56.54 **(-0.03)**
 :red_circle: | app/client/src/components/designSystems/appsmith/SearchComponent.tsx | 75.86 **(-3.45)** | 75 **(-8.33)** | 50 **(-16.67)** | 75 **(-3.57)**
 :red_circle: | app/client/src/reducers/entityReducers/canvasWidgetsReducer.tsx | 85.71 **(-5.2)** | 100 **(0)** | 50 **(-16.67)** | 85.71 **(-5.2)**
 :sparkles: :new: | **app/client/src/reducers/uiReducers/mainCanvasReducer.ts** | **72.73** | **22.22** | **50** | **72.73**
 :green_circle: | app/client/src/selectors/editorSelectors.tsx | 75.74 **(0.31)** | 61.74 **(0)** | 60.49 **(0.49)** | 74.33 **(0.53)**
 :green_circle: | app/client/src/widgets/BaseWidget.tsx | 79.53 **(0)** | 46.67 **(2.23)** | 75.76 **(0)** | 78.69 **(0)**
 :red_circle: | app/client/src/widgets/ChartWidget/component/index.tsx | 49.41 **(-1.18)** | 22.68 **(-1.03)** | 50 **(-3.13)** | 47.47 **(-0.63)**
 :red_circle: | app/client/src/widgets/ModalWidget/widget/index.tsx | 38.2 **(-0.69)** | 0 **(0)** | 16.67 **(0)** | 37.65 **(-0.72)**
 :red_circle: | app/client/src/widgets/TableWidget/component/TableStyledWrappers.tsx | 89.66 **(0)** | 55.84 **(-0.36)** | 83.64 **(0)** | 89.66 **(0)**
 :red_circle: | app/client/src/widgets/TableWidget/component/index.tsx | 27.27 **(-1.3)** | 2.6 **(-24.67)** | 15.79 **(-5.26)** | 25.68 **(-1.35)**
 :red_circle: | app/client/src/widgets/TableWidget/widget/index.tsx | 26.62 **(-3.12)** | 6.03 **(-3.45)** | 26.92 **(-1.93)** | 26.51 **(-2.89)**
 :red_circle: | app/client/src/widgets/TextWidget/component/index.tsx | 71.79 **(-1.29)** | 45.03 **(-0.67)** | 53.85 **(-3.84)** | 71.05 **(-1.32)**</details>


rahulramesha:
cc: @yaldram

jsartisan (dismissed):

marks0351:
@rahulramesha @riodeuno This fixes the changes that happen to the MainContainer to not be included but wud need to fix the updates that happen to the canvas widgets inside individual container widgets as well.

The problem with this issue is a minor flaw of our arch. after the layout options feature, when we set a flexible/fluid layout we calculate rightColumn of main canvas as well as all the other canvas widgets(inside container widgets) in realtime which change for resize of browser so no point in saving them in the dsl. so we need to either treat them as meta properties on the app and maintain them separately or deprecate these dimensions being saved on the dsl coz they are not of much use. going with the meta properties way will have minimum disruption in my opinion.

riodeuno:
@marks0351 Shouldn't the nested CANVAS_WIDGETS automatically, fill their parent widgets? I am not sure I understand the issue completely. Could you elaborate on which properties do we update?

rahulramesha:
Had a Discussion with @marks0351, Looks like the CANVAS_WIDGETs dont update the DSL when the main Container changes. These calculations are happening on the fly based on main Container's Dimensions. So this approach seems to work.

@riodeuno but there is an issues of outdated fields in the Widgets. For canvas' dimension properties, each widget's parentColSpace and parentRowSpace are redundant. Since the logic to calculate them on the fly was implemented some time back. So DSL cleanup has to be discussed.

riodeuno:
@rahulramesha Got it. Yes, if these properties are calculated on the fly for every scenario. We can migrate and cleanup the DSLs. However, we can pick this up as a separate effort. Would you mind creating a task for this?
Also, are there any pieces of code, which specifically update these properties in our codebase today?

rahulramesha:
@riodeuno, Will create a task for this. 
Should we take the risk of cleaning up and migrating old DSLs or just streamline DSLs for new canvases?

Need to do some digging to find all the places that are updating the properties.

marks0351 (dismissed):

app/client/src/reducers/uiReducers/mainCanvasReducer.ts
```diff
@@ -0,0 +1,45 @@
...
+} from "constants/ReduxActionConstants";
+import { MAIN_CONTAINER_WIDGET_ID } from "constants/WidgetConstants";
+import { UpdateCanvasLayoutPayload } from "actions/controlActions";
+
+const initialState: MainCanvasReduxState = {
+  width: 0,
+  height: 0,
+};
+
+const mainCanvasReducer = createReducer(initialState, {
```
riodeuno, line 15: I know that `width` and `height` are primitives. However, shall we use `createImmerReducer` here? It seems like even if we need to keep the state the same, we'll be creating a new object which will change the reference and result in updates to all selectors, that use the full state.

app/client/src/selectors/editorSelectors.tsx
```diff
@@ -229,8 +231,13 @@ export const getWidgetCards = createSelector(
...
   evaluatedDataTree: DataTree,
+  mainCanvasProps: MainCanvasReduxState,
 ) => {
-  const canvasWidget = canvasWidgets[MAIN_CONTAINER_WIDGET_ID];
+  const canvasWidget = {
+    ...canvasWidgets[MAIN_CONTAINER_WIDGET_ID],
+    rightColumn: mainCanvasProps.width,
+    minHeight: mainCanvasProps.height,
+  };
   const evaluatedWidget = find(evaluatedDataTree, {
```
riodeuno, line 241: Do we need the `MainContainer` from the `evaluatedDataTree`? I was wondering if anything is evaluated in the MainContainer properties, that would justify us using the evaluated version. In the interest of reducing the amount of code run to generate the main container properties, can we simply send the main container back after updating the rightColumn and minHeight? Possibly memoizing this in a manner that doesn't change the reference of `canvasWidget` if it doesn't need to?

app/client/src/utils/hooks/useDynamicAppLayout.tsx
```diff
@@ -36,7 +36,7 @@ export const useDynamicAppLayout = () => {
   const domEntityExplorer = document.querySelector(".js-entity-explorer");
   const domPropertyPane = document.querySelector(".js-property-pane-sidebar");
   const { height: screenHeight, width: screenWidth } = useWindowSizeHooks();
-  const mainContainer = useSelector(getWidgetByID(MAIN_CONTAINER_WIDGET_ID));
+  const mainCanvasProps = useSelector(getMainCanvasProps);
```
riodeuno, line 39: As mentioned in the comment in the reducer. We will have new references for mainCanvasProps, resulting in re-renders even if we don't really change the width and height. Shall we make this reducer an `immerReducer`?
cc @SatishGandham

app/client/src/reducers/uiReducers/mainCanvasReducer.ts
riodeuno: @SatishGandham Could you comment on whether my suggestion makes sense here?

vivekverma2312:
Tested and working fine.

app/client/src/selectors/editorSelectors.tsx
rahulramesha: @riodeuno,  Not really sure why `MainContainer` from the `evaluatedDataTree` is needed here. It requires some more digging as to where evaluated values are used before removing.
The problem I see in memoizing is that `canvasWidget` needs to change either when `canvasWidgets[MAIN_CONTAINER_WIDGET_ID]`, changes or when `mainCanvasProps`. That might defeat the idea of memoizing. 
Can you please explain what you have in mind for memoizing the `canvasWidget`.

app/client/src/utils/hooks/useDynamicAppLayout.tsx
rahulramesha: @riodeuno, converted to `immerReducer`

app/client/src/reducers/uiReducers/mainCanvasReducer.ts
```diff
@@ -0,0 +1,45 @@
...
+const mainCanvasReducer = createImmerReducer(initialState, {
+  [ReduxActionTypes.INIT_CANVAS_LAYOUT]: (
+    state: MainCanvasReduxState,
+    action: ReduxAction<UpdateCanvasPayload>,
+  ) => {
+    const mainCanvas =
+      action.payload.widgets &&
+      action.payload.widgets[MAIN_CONTAINER_WIDGET_ID];
+
+    return {
```
riodeuno, line 24: Since we're using immer, we can do this as so:
```suggestion
    state.width = mainCanvas?.rightColumn || state.width;
    state.height = mainCanvas?.minHeight || state.height;
```

riodeuno:
@vivekverma2312 There have been changes now, this could effect how the feature works. Could we get this verified again?
@rahulramesha Please comment here when all changes are done so that we can do one final QA before merging.

riodeuno (approved):

rahulramesha:
@vivekverma2312, All Changes have been made. It should still work the same as before. A quick sanity on this should be enough.

vivekverma2312:
@rahulramesha LGTM

riodeuno (approved):

PR #12372, opened 2022-03-30
fix: Disabled button widget tooltips stay open

### LINKED ISSUE #7554: [Bug] Disabled button widget tooltips stay open
## Description

Tooltips on disabled buttons are not closing after the user has moved their cursor away from a button widget.

1. Create new page
2. Add button widget
3. Add tooltip
4. Disable button
5. Hover over button
6. Move cursor and tool tip remains

### Important Details

- Version: Cloud
- OS: Windows 10
- Browser: Chromium Edge
- Environment: N\A

wmdev0808:
## Description

On deploy, though the tooltip appears when button is disabled, when the cursor is not focusing on the button, the tooltip still stays. It does not go away.

Fixes #7554 

## Type of change

## How Has This Been Tested?

- Test A
- Test B

## Checklist:

    // Code coverage diff between base branch:release and head branch: fix/7554-disabled-button-enable-tooltip 
 :green_circle: | total | 56.24 **(0)** | 37.69 **(0.01)** | 35.92 **(0)** | 56.47 **(0)**
 :green_circle: | app/client/src/widgets/ButtonWidget/component/DragContainer.tsx | 66.67 **(0)** | 50 **(10)** | 66.67 **(0)** | 69.23 **(0)**
 :green_circle: | app/client/src/widgets/ButtonWidget/component/index.tsx | 50 **(0.51)** | 30.21 **(1.49)** | 45.45 **(0)** | 50.57 **(0.57)**</details>


app/client/src/widgets/ButtonWidget/component/index.tsx
```diff
@@ -237,24 +259,45 @@ export function BaseButton(props: IButtonProps & ButtonStyleProps) {
...
-        fill
-        icon={isRightAlign ? icon : iconName || icon}
-        loading={loading}
-        onClick={onClick}
-        placement={placement}
-        rightIcon={isRightAlign ? iconName || rightIcon : rightIcon}
-        text={text}
-      />
+      {disabled && tooltip ? (
+        <StyledAnchorButton
```
ashit-rath, line 263: Can you help me understand why we need to switch between AnchorButton and StyledButton?

wmdev0808: Buttons make great popover targets, but the disabled attribute on a <button> blocks all events, which interferes with the popover functioning. If you need to disable a button which triggers a popover, you should use [AnchorButton](https://blueprintjs.com/docs/#core/components/button.anchor-button) instead.

ashit-rath: > disabled attribute on blocks all events

If it blocks all events then how is the tooltip appearing on entering? That should ideally also not work. I think it could be an issue with the library itself, can you verify this?

I'm not in a favor of this solution as that would lead to unnecessary re-mounting of the component and also we would technically be changing a button to an anchor here.

ashit-rath:
@yaldram Can we fix this issue by extending your implementation of the Button `DragContainer` wrapper that you did a few days back to the `Page`  render mode?

@wmdev0808 Please check the `src/widgets/ButtonWidget/component/DragContainer.tsx` this wrapper is helping the tooltip to work correctly in the edit mode.

yaldram:
Hey @ashit-rath unfortunately this is how the Library works - https://github.com/palantir/blueprint/issues/2779#issuecomment-411446961. @wmdev0808 did it correct.

yaldram:
In render mode we have the `DragContainer` so it is working as expected but then in the deploy version we just remove it altogether.

yaldram:
One way to do this is to not remove the dragcontainer at all in the deploy mode, then we have our button wrapped by a div. Other solution is what Paul just did. Hey @ashit-rath you can pick any of these solutions, I will go with Paul's solution. Though we mount another component we do it only in render mode and in the deployed version there is no mounting - unmounting and extra wrappers.

ashit-rath:
> One way to do this is to not remove the dragcontainer at all in the deploy mode, then we have our button wrapped by a div. Other solution is what Paul just did. Hey @ashit-rath you can pick any of these solutions, I will go with Paul's solution. Though we mount another component we do it only in render mode and in the deployed version there is no mounting - unmounting and extra wrappers.

@yaldram Since we already have a DragContainer which solves this issue, let's reuse that rather than switching between 2 components. The version of blueprint we have is not known for it's performance so let's try to reduce such re-mountings even in the View mode.

PS: remounting would occur as disabled states are always accompanied by some js logic to toggle between the states.

yaldram:
Hey @ashit-rath sure. @wmdev0808 In the `src/widgets/ButtonWidget/component/DragContainer.tsx` just remove the `if` condition that only renders this in canvas mode. Also please check if the same issue is present for other `Button` Widgets like `Menu Button`, `Button Group Widget`. `IconButton` has its own `Container` in the widget/index.tsx itself. Please put this to Q/A after your changes.

yaldram:
@ashit-rath @wmdev0808 Some observations : - 
1. The `DragContainer` is used for `ButtonGroupWidget, MenuWidget and ButtonWidget`, but only the `ButtonWidget` accepts tooltip.
2. So we can only wrap the `ButtonWidget` with the `DragContainer` and it solves the tooltip issue, I just passed a simple prop `showInAllModes`.

ashit-rath (approved):

app/client/src/widgets/ButtonWidget/component/index.tsx
```diff
@@ -258,46 +241,26 @@ export function BaseButton(props: IButtonProps & ButtonStyleProps) {
       loading={loading}
       onClick={onClick}
       renderMode={renderMode}
```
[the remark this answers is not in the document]
yaldram: Yes, thanks we can remove this prop.

ashit-rath (dismissed):

YogeshJayaseelan:
@wmdev0808 Tested this PR and it is working as expected

ashit-rath (approved):

PR #12398, opened 2022-03-30
ci: Adding ingress.class parameter to nginx setting in helm chart docs

sum35h:
Specified ingress.class to nginx setting in helm chart https docs.

> While updating Appsmith with ssl using Helm, I came across an error due to ingress class of the ingress controller not matching ingress class in the ingress resource manifest(nginx), the proposed change solved the issue.

### DIFF
--- a/deploy/helm/Setup-https.md
+++ b/deploy/helm/Setup-https.md
@@ -86,6 +86,7 @@ The steps below explain how to use Ingress routes and cert-manager to configure 
   --set ingress.hosts[0].host=DOMAIN \
   --set ingress.certManagerTls[0].hosts[0]=DOMAIN \
   --set ingress.certManagerTls[0].secretName=letsencrypt-prod
+  --set ingress.class=nginx
   ```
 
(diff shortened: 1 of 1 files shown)


deploy/helm/Setup-https.md
```diff
@@ -86,6 +86,7 @@ The steps below explain how to use Ingress routes and cert-manager to configure
   --set ingress.hosts[0].host=DOMAIN \
   --set ingress.certManagerTls[0].hosts[0]=DOMAIN \
   --set ingress.certManagerTls[0].secretName=letsencrypt-prod
+  --set ingress.class=nginx
```
sharat87, line 89: I don't see the `ingress.class` configuration in our `values.yaml` at https://github.com/appsmithorg/appsmith/blob/68065f87898fd5c580f4b826e8870f8551fd51e9/deploy/helm/values.yaml#L99. Is this a supported configuration? If yes, can se add it to that file and to our docs on docs.appsmith.com as well?

Although, I see a `className` configuration. Did you mean to set that here?

sum35h: My bad, its a typo. it should be className

sharat87 (approved):

PR #12507, opened 2022-04-01
feat: Import an template inside an existing Application

### LINKED ISSUE #12336: Import an template inside an existing Application

nayan-rafiq:
## Description
This PR adds the feature to import a template inside an application. 

Fixes #12336

## Type of change

## How Has This Been Tested?

- JUnit test
- Manually

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationTemplateControllerCE.java
```diff
@@ -46,4 +49,13 @@ public ApplicationTemplateControllerCE(ApplicationTemplateService applicationTem
         return applicationTemplateService.importApplicationFromTemplate(templateId, organizationId)
                 .map(importedApp -> new ResponseDTO<>(HttpStatus.OK.value(), importedApp, null));
     }
+
+    @PostMapping("{templateId}/merge/{applicationId}")
+    public Mono<ResponseDTO<Application>> mergeTemplateWithApplication(@PathVariable String templateId,
+                                                                       @PathVariable String applicationId,
+                                                                       @RequestHeader(name = FieldName.BRANCH_NAME, required = false) String branchName,
+                                                                       @RequestBody(required = false) List<String> pagesToImport) {
+        return applicationTemplateService.mergeTemplateWithApplication(templateId, applicationId, branchName, pagesToImport)
```
AnaghHegde, line 58: I am not able to follow on why we are expecting the list of pages to in request param? Is it because we allow the users to import selected pages from templates as well?

AnaghHegde (changes requested):
Can we not repeat the same logic again here? I would suggest to modify the appJson file to do the changes `updateNewPagesbeforeMerge`. The current importApplication handles importing resources to the existing app as well as the new app. 
It would be difficult to maintain this in the coming days when we fix a bug or change something in the import flow, we would have to maintain in both the places.

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationTemplateControllerCE.java
nayan-rafiq: Yeah, that's right. Users will be able to either import the whole application or specific pages of the application.

nayan-rafiq:
@AnaghHegde Definitely it would be better if import application method is used for this purpose. I tried using that earlier but faced some issues - `gitSyncId` was just one of it. Another issue was - existing actions were being delete from pages. I'll try again.

AnaghHegde:
@nayan-rafiq Here is what i am thinking, let's not touch the ImportApplication method. It is already designed to handle merging two applications(for git sync) where it can create/delete/update resources. For templates what we can do is set the the git sync id as null after exporting the json file - this way one can use the same template in the same application multiple times. Second, for application name getting updated, we can update the appJson so that the application name is same as where it is being imported. Let me know your thoughts on this.

AnaghHegde (commented):
@nayan-rafiq Here is what i am thinking, let's not touch the ImportApplication method. It is already designed to handle merging two applications(for git sync) where it can create/delete/update resources. For templates what we can do is set the the git sync id as null after exporting the json file - this way one can use the same template in the same application multiple times. Second, for application name getting updated, we can update the appJson so that the application name is same as where it is being imported. Let me know your thoughts on this.

AnaghHegde (dismissed):

abhvsn (dismissed):

AnaghHegde (dismissed):

AnaghHegde (approved):

AnaghHegde (approved):

AnaghHegde (approved):

AnaghHegde (approved):

abhvsn (dismissed):

AnaghHegde (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -2030,21 +2030,26 @@ private Theme exportTheme(Theme srcTheme) {
...
+                application.setPublishedModeThemeId(publishedModeThemeId);
+                // this will update the theme id in DB
+                // also returning the updated application object so that theme id are available to the next pipeline
+                return applicationService.setAppTheme(
+                        application.getId(), editModeThemeId, publishedModeThemeId, MANAGE_APPLICATIONS
+                ).thenReturn(application);
+            });
+        } else {
+            return Mono.just(application);
+        }
```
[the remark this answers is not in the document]
nayan-rafiq: I've updated the PR.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -2029,27 +2029,11 @@ private Theme exportTheme(Theme srcTheme) {
...
-                        application.getId(), editModeThemeId, publishedModeThemeId, MANAGE_APPLICATIONS
-                ).thenReturn(application);
-            });
-        } else {
+    private Mono<Application> importThemes(Application application, ApplicationJson importedApplicationJson, boolean appendToApp) {
+        if(appendToApp) { // appending to existing app, theme should not change
             return Mono.just(application);
         }
-
+        return themeService.importThemesToApplication(application, importedApplicationJson);
```
abhvsn, line 2016: Thanks, @nayan-rafiq for moving this to the service class. We will be doing this clean up for other methods as well.   
We should also move the export as well to service class.

abhvsn (approved):

Issue #11538, opened 2022-03-02
plugins/forms api calls should not block the editor


SatishGandham:
After we get the list of plugins, we make calls to `plugin/_id_/form` to get the plugin form details in sequence. That adds 2-2.5 seconds to the first load time.

1. These form details are not needed on the widgets editor.
2. They should be loaded in parallel.
3. They don't need to wait for the dsl.
4. They aren't even required on the editor.
5. Editor need not wait for these apis, but it does.

If we are prefetching this, so that we already have the info when going to the respective plugins screen,
please find why the editor is made to wait for these, and see if fixing this no longer blocks the editor.

If not, remove these call from the widgets editor screen.

In the screenshot below

- Move 4 to the start
- Eliminate 5, or make them parallel and move to start/non blocking

<img width="2168" alt="Screenshot 2022-03-02 at 2 45 59 PM" src="https://user-images.githubusercontent.com/441914/156336358-d0794be5-e56b-41ef-a3e9-0b200acbe97d.png">

somangshu:
## actionable 

Decide what can be done further, Below are some strategies 
- [ ] we need to delay the form calls
- [ ] Make it non blocking
- [ ] since they are static, it makes sense to cache them

somangshu:
@SatishGandham what is the verdict here? Can you please drop a note? AFAIR, we said that this is not the way forward.

ayushpahwa:
Closing this issue since the APIs are now already non blocking.

PR #12537, opened 2022-04-04
chore: Parallellise fetch page call with actions, jsObject and  plugins

### LINKED ISSUE #11553: Ensure that the DSL,actions, js objects  calls are made as soon as possible
We wait till the features api call is back to load the editor. Then, we wait for the DSL to load to make the calls to Actions and **JSObjects**.

```
// In AppRouter.tsx
// Render the app once the feature flags have been fetched
    if (!featureFlagsFetched) return null;
```
We should be making the DSL and the two calls that happen along with it, plus JS Object and Actions as soon as possible. 
Screenshot below marks where the DSL calls and **actions**+**JSObjects** should start

Screenshot 2022-03-02 at 2 45 59 PM

arunvjn:
## Description
Enhancements to improve first page load by making API calls as soon as possible. This PR holds the change to make the first fetch page call along side actions, plugins and datasources.

Fixes #11553 

## Type of change
- Perf fix

## How Has This Been Tested?
- Manual

## Checklist:

    // Code coverage diff between base branch:release and head branch: chore/11553-rush-api-calls 
 :green_circle: | total | 56.35 **(0.02)** | 37.75 **(0.02)** | 35.94 **(0)** | 56.57 **(0.01)**
 :green_circle: | app/client/src/sagas/InitSagas.ts | 51.43 **(3.3)** | 36.67 **(3.83)** | 58.33 **(0)** | 55.1 **(3.51)**
 :green_circle: | app/client/src/selectors/commentsSelectors.ts | 85.25 **(1.64)** | 64.71 **(2.95)** | 73.33 **(0)** | 90.59 **(2.35)**</details>


app/client/src/sagas/InitSagas.ts
```diff
@@ -313,8 +300,11 @@ function* initializeEditorSaga(
     yield all([
       call(initiateEditorActions, applicationId),
       call(initiatePluginsAndDatasources),
+      call(populatePageDSLsSaga),
     ]);
 
+    yield put(fetchPluginFormConfigs());
```
riodeuno, line 306: We've moved away from using `failFastApiCalls` to a `put`. Seems like the only difference is the call of `SAFE_CRASH_APPSMITH_REQUEST` when the call fails. A few queries around this:
1. Will not calling the `SAFE_CRASH_APPSMITH_REQUEST` cause any experience difference to Appsmith developers? 
2. If so, how will it be different? Will we be able to retry or fail safely to allow Appsmith developers to continue without being blocked?

arunvjn: This is a mistake. Fixing this.

arunvjn: @riodeuno I have reverted the changes to make plugin form calls non-blocking since it has a lot more side effects than just SAFE_CRASH. The query editor pages on refresh doesn't display the form and fixing this needs quite a bit of refactoring. We will take this up as part of #11538.
This PR holds the following changes:
- parallelize GET publishedPage with actions and JS Objects in view mode.
- parallelize GET page dsl calls with actions, datasources and plugins

arunvjn: Added change to call fetchPage and fetchPublishedPage API calls parallel to actions

riodeuno (approved):

app/client/src/sagas/InitSagas.ts
```diff
@@ -329,6 +329,7 @@ function* initializeEditorSaga(
     yield all([
       call(initiateEditorActions, applicationId),
       call(initiatePluginsAndDatasources),
+      call(populatePageDSLsSaga),
```
Rishabh-Rathod, line 332: Why did we parallelize the call for all page DSLs here? @arunvjn 

fetchPage inside `initiateEditorApplicationAndPages` is already calling for currentPageDSL. 

`populatePageDSLsSaga` can be called as late as possible here, as it was added mainly for the entity explorer. 

Let me know if I misunderstood something here.

Rishabh-Rathod: I believe, we could remove `populatePageDSLsSaga` and instead add it only for individual pages fetch according to the new entity explorer design.

arunvjn: Yes, you are right. @Rishabh-Rathod 
This code was present in the previous iterations as well. 
We need the entities from other pages in the omni-bar as well. Opening omni-bar could be the trigger point.

Rishabh-Rathod: > Opening Omni-bar could be the trigger point.

We should have a separate API to get all entities of all pages, fetching complete page information is not the best approach here, if we consider scalability. 

For example:- 
Loading 10-15 pages which may each have 5-10 JsObjects, 20-30 widgets, and 10 queries/APIs just to show in the Omnibar doesn’t scale well IMO.

arunvjn: JsObjects & Actions comes from their respective APIs(queried against the applicationId). 
PopulatePageDSL call only gets us the widgets from other pages. 
Are you talking about a slim API which only gets basic widget info, for the omni-bar ? Or are you talking about slim API for actions and JS Objects as well ?

Rishabh-Rathod: > Are you talking about a slim API that only gets basic widget info, for the Omni-bar? Or are you talking about slim API for actions and JS Objects as well ?

Yes, for omnibar slim data should do that job is what I believe.

Issue #11752, opened 2022-03-10
[Feature]: Support getting values from secrets in Helm chart configuration


sharat87:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

Currently, when deploying Appsmith with Helm charts, all configuration values should be given in plain-text in the `values.yaml` file. This includes sensitive values like MonogDB connection URI, which would include the password.

Instead, the Helm charts should support getting these values from secrets and support this in a backward-compatible way.

For example, today, we can set the MongoDB URI, like the following:

```yaml
applicationConfig:
  APPSMITH_MONGODB_URI: "mongodb-uri-here"
```

In addition to this, we should support something like this:

```yaml
applicationConfig:
  APPSMITH_MONGODB_URI:
    secretName: "appsmith-secrets"
    secretKey: "mongodb-uri"
```

When configured this way, the Helm chart should just pick the MongoDB URI from this secret. This secret can be created separately by the users within the same namespace as Appsmith, by creating a resource like:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: appsmith-secrets
type: Opaque
data:
  mongodb-uri: bW9uZ291c2Vy
```

### Why should this be worked on?

This will allow the `values.yaml` file to not contain sensitive values directly, and be able to safely share in a git repo.

vagrantism:
Hey @sharat87, this is exactly what we were talking about on [Discord](https://discord.com/channels/725602949748752515/950310358004957185), thanks for opening the issue. The only potential problem I see here is the user having to specify the secret-key multiple times if the number of parameters is too large.

Given that the [list of configurable parameters for Appsmith](https://docs.appsmith.com/setup/kubernetes#parameters) is huge, you might want to just take the secret name from the user, and automatically use any matching secret key found in there. One of the ways to do this would be to mount the secret as a JSON file in a volume and reference it in the configmap.

However, the approach you have defined is simpler and it would make sense to have it implemented first and then look into my approach.

PR #12633, opened 2022-04-06
feat: Support getting values from secrets in helm chart configuration

sum35h:
## Description
- Added a new variable `secretName` in `HELM/values.yaml` to specify the K8s Secret object that has been created by the user, outside the Appsmith helm flow.
- The secret data will be added to the deployment instance as ENV variables (overwriting the values in configMap if they have also been specified there).
- The secret resource will be of the format below, with the keys corresponding to the actual Appsmith ENV variable and the values as base64-encoded string:

```
apiVersion: v1
data:
  APPSMITH_GOOGLE_MAPS_API_KEY: bW9uZ291c2Vy
kind: Secret
metadata:
  name: appsmith-secret
type: Opaque
```
Fixes # [11752](https://github.com/appsmithorg/appsmith/issues/11752)

## Type of change

## How Has This Been Tested?
Manually on AWS EKS

## Checklist:

### DIFF
--- a/deploy/helm/templates/statefulset.yaml
+++ b/deploy/helm/templates/statefulset.yaml
@@ -72,6 +72,10 @@ spec:
           envFrom:
             - configMapRef:
                 name: {{ include "appsmith.fullname" . }}
+            {{- if .Values.secretName }}
+            - secretRef:
+                name: {{ .Values.secretName }}
+            {{- end }}
       volumes:
   {{- if not .Values.persistence.enabled }}
       - name: data
--- a/deploy/helm/values.yaml
+++ b/deploy/helm/values.yaml
@@ -239,6 +239,8 @@ autoupdate:
   ##
   scheduler: "0 * * * *"
 
+secretName: ""
+
 applicationConfig:
   APPSMITH_OAUTH2_GOOGLE_CLIENT_ID: ""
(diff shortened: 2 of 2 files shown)


deploy/helm/templates/statefulset.yaml
```diff
@@ -72,6 +72,10 @@ spec:
           envFrom:
             - configMapRef:
                 name: {{ include "appsmith.fullname" . }}
+            {{- if .Values.secretName }}
+            - secretRef:
+                name: {{ .Values.secretName }}
+            {{- end }}
```
sharat87, line 78: This is a clever solution. But it feels like a double-edged sword. Can you elaborate on the below points?

1. If we define the `seretName`, then env variables will be loaded from _both_ the `configMap` as well as the `secret`. So, is it clearly defined which takes precedence if something is defined in both the places?
2. In the Helm templates, we have a `ConfigMap` resource, that we are providing by default, should we include a `Secret` resource as well, that can be empty when `secretName` is not set perhaps? Or should we leave the secret resource to the user and provide help via docs?

sharat87: As discussed, value from secrets take precedence. Let's add this note to the docs when this goes to the master branch. @sum35h

sharat87 (approved):

PR #12699, opened 2022-04-08
fix: Current branch delete exception in delete branch API

### LINKED ISSUE #12674: [Bug]: Fix NPE for delete branch API when user tries to delete the checked out branch

### Description

The server should throw the relevant error message in case of an unsuccessful transaction.

### Steps To Reproduce

1. Connect application to git
2. Create a new branch
3. Try to delete the current checked out branch
4. Internal server error is thrown

### Public Sample App

_No response_

### Version

Cloud

AnaghHegde:
## Description

> When the user tried to delete the current checked out branch, we should throw the relevant error message and handle it properly without affecting the other resources and state.

Fixes #12674 

## Type of change

## How Has This Been Tested?

- Locally

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
@@ -52,6 +52,7 @@ import com.appsmith.server.services.UserService;
 import com.appsmith.server.solutions.ImportExportApplicationService;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
+import org.eclipse.jgit.api.errors.CannotDeleteCurrentBranchException;
 import org.eclipse.jgit.api.errors.CheckoutConflictException;
 import org.eclipse.jgit.api.errors.EmptyCommitException;
 import org.eclipse.jgit.api.errors.GitAPIException;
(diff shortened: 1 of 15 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -2070,6 +2071,13 @@ private String getReason() {
                         return Mono.error(new AppsmithException(AppsmithError.GIT_ACTION_FAILED, " delete branch", "Cannot delete default branch"));
                     }
                     return gitExecutor.deleteBranch(repoPath, branchName)
+                            .onErrorResume(throwable -> {
+                                log.error(" Delete branch failed ", throwable.getMessage());
+                                if(throwable instanceof CannotDeleteCurrentBranchException) {
+                                    return Mono.error(new AppsmithException(AppsmithError.GIT_ACTION_FAILED, " delete branch", throwable.getMessage()));
```
abhvsn, line 2077: We don't need the space prefix "` `delete branch" as this has been already handled in AppsmithError

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -2070,6 +2071,13 @@ private String getReason() {
...
+                                }
+                                return Mono.error(new AppsmithException(AppsmithError.GIT_ACTION_FAILED, " delete branch"));
```
abhvsn, line 2079: Can we please use second parameter to throw the error given by JGIT here?
```
return Mono.error(new AppsmithException(AppsmithError.GIT_ACTION_FAILED, "delete branch", throwable.getMessage()));
```

abhvsn (changes requested):
Let's put up a constraint that user can't delete the branch where she has checked out in UI

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: If the exception is not because of current branch delete exception, we do not want to show it to user. Hence i am not sending the exception message here.

abhvsn: Any specific reason for this?

AnaghHegde: IIUC, we are following this across all the git errors.

abhvsn: Nope we are sending the user friendly error message if we are already aware of the root cause just like you are checking the `CannotDeleteCurrentBranchException` in the above implementation, if not we are throwing the error from JGit as is

abhvsn (approved):
Please make sure client side code handles the scenario where delete icon will not be shown for the current checked out branch on UI

PR #12726, opened 2022-04-08
fix: handle editor value update

### LINKED ISSUE #11624: [Bug]-[600]:RichTextEditor not loading at start

### Current Behavior

Currently i have a RichTextEditor.
The "default text" is filled by a database.
In widget edit mode (where you can place the appsmith widgets) it works fine.
Now I deploy the app and open it it does not work. The editor remains empty.
If I change the page in the same app via the navigation bar of Appsmith and change back again it works.
It can be a disaster if the editor is empty and when changing the data then the current data is completely deleted from the database (onChange).

### Steps To Reproduce

1. Place a RichTextEditor Widget
2. Use a database query (postgres) `{{RichTextData.data[0].data}}` (`SELECT data FROM RichTextData;`)
3. Deploy
4. Open the deployed app

### Environment

Release

### Version

Self-Hosted

techbhavin:
## Description

RichTextEditor isn’t loading at the start to fix that issue `textChange` event prevents while initializations of RichTextEditor.

Fixes #11624 

## Type of change

## Checklist:

    // Code coverage diff between base branch:release and head branch: fix/issue-11624 
 :red_circle: | total | 56.46 **(0)** | 37.91 **(-0.01)** | 36.17 **(0)** | 56.68 **(0)**


lublak:
@techbhavin there is something like this:

```js
editor.on('init', function (e) {
  editor.setContent(content);
});
```

SatishGandham (commented):
Why is it not changing?
We pass the value prop, why is it not changing when the value prop changes?

techbhavin:
As per my observation, in the initial render when props are changing, `editorRef.current` is not there, but when `editorRef.current` changes, `props.value` is stable and that will not effect on `useEffect`

image

SatishGandham (approved):

shwetha-ramesh:
@techbhavin tested this on Windows machine on Chrome browser. This issue is still partially reproducible. 
https://loom.com/share/cbacf59b9c204c00abab01cfbad0277f

app/client/src/widgets/RichTextEditorWidget/component/index.tsx
```diff
@@ -77,33 +82,38 @@ export function RichtextEditorComponent(props: RichtextEditorComponentProps) {
...
     "insertfile undo redo | formatselect | bold italic backcolor forecolor | alignleft aligncenter alignright alignjustify | bullist numlist outdent indent | link image | removeformat | table | print preview media | forecolor backcolor emoticons' | help";
 
   useEffect(() => {
-    if (!value && !props.value) return;
+    if (!value.text && !props.value) return;
     // This Prevents calling onTextChange when initialized
-    if (!isInit.current) return;
-    const timeOutId = setTimeout(() => props.onValueChange(value), 1000);
+    if (!value.isUserEdit) return;
+    const timeOutId = setTimeout(() => props.onValueChange(value.text), 1000);
```
SatishGandham, line 99: Why are you waiting a second to trigger the update?

SatishGandham (changes requested):

app/client/src/widgets/RichTextEditorWidget/component/index.tsx
SatishGandham: @Tooluloope 
Why did we add a one second debounce. Update meta property is already debounced.

techbhavin: to prevent triggering onValueChange method when the user typing. it will call when the user takes a pause while typing

SatishGandham (approved):

shwetha-ramesh:
@techbhavin re-tested this on Windows machine on Chrome browser. Working as expected now.
https://loom.com/share/ee17b306517b436888d307aaada35273

Moving to Done.

app/client/src/widgets/RichTextEditorWidget/component/index.tsx
Rishabh-Rathod: Why do we want to prevent calling `onValueChange` when the user is typing? 
I believe that will cause the local state to update unexpectedly as per current logic.

@techbhavin I would like to understand why we need the local state for the RTE component and how is it different from the input widget?

Tooluloope: - If we leave it solely to the meta property to provide the value, once a new value is provided, the cursor would shift to the beginning, which is bad UX.

Issue #12000, opened 2022-03-21
[Bug]: JS objects are being duplicated during git sync


nidhi-nair:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

A user has reported that when they branch off of their app, the JS objects are duplicated in the resulting app. This causes the app to be in erroneous state, since the collection can also not be deleted because of multiple branched collections referring to the same master collection.

### Steps To Reproduce

https://discord.com/channels/725602949748752515/953747596625707089/953750765258833970

### Version

Self-Hosted

Nikhil-Nandagopal:
@hiteshjoshi @abhvsn are we working on this? This can potentially break apps for users

abhvsn:
Other thread where this issue was getting discussed: https://discord.com/channels/725602949748752515/948548193296195605 
Issue might be bigger and not specifically related to the git, I'll try to triage the root cause as with git branching I'm not able to reproduce this.

ereio:
+1 can confirm this is happening to us on self-hosted

PR #12743, opened 2022-04-10
fix: After cloning the page with JSObject creating a new branch copies all the JSObject in a single page

### LINKED ISSUE #12724: [Bug]: Js objects are merged to single page when user creates a new branch

### Description

Js objects are merged to single page when user creates a new branch 

### Steps To Reproduce

1. Connect an application to the git 
2. Create an app with Page1
3.Add JSObject names JSObject1
4..Create new page Page2
5.Add JSObject names JSObject1
6.Clone the page 1
6.Now create a sub-branch from the master
7.All the js object will be displayed under the single page 

### Public Sample App

_No response_

### Version

release

abhvsn:
## Description

As per our current implementation we use `gitSyncId` to connect different resources in multiple Appsmith instances like pages, actions, JSObjects etc. This fiels should be unique across the instances for a resource in DB. 

Consider we have branch1 with JSObject1 and now user creates the branch2 from earlier branch in the same isnatnce. In this scenario we create the JSObject1 in branch2 but will have same `gitSyncId`. Considering we also have instance 2 where branch1 is already present but we are creating a branch2 from remote we can easily connect the JSObject1 from branch2 pulled from remote to already available JSObject1 of branch1 as we have same `gitSyncId`.

In the import application flow we use these gitSyncIds to find out the existing resources which can be connected to the imported resource. Also as we want to connect the resources across the instances we commit this field to repo as well.

The problem we were facing earlier was related to cloning the JSObject when we try to clone the page, this action resulted in duplicating the `gitSyncId` for 2 different resources in a single branch as we have not handled it in the actionCollection create method (Added a comment in the review to make this clear). Now that we have the JSObjects with duplicate gitSyncIds we end up accumulating all the JSObjects in a single page if the user tries to create a new branch, this happened because we fetch the resource with gitSyncId and then assign the `unpublishedCollection.defaultResources.pageId` to the imported resource.

Fixes #12724 #12000

## Type of change

## How Has This Been Tested?

> Manual test
> JUnit testcase

## Checklist:


AnaghHegde (approved):
LGTM.
But does it make sense to add a test case where we do the clone operation and then do create branch ?

abhvsn:
> LGTM. But does it make sense to add a test case where we do the clone operation and then do create branch ?

That will not be required as the root cause was cloning, resulted in incorrect assignment for `gitSyncId` in cloned JSObjects. Also we will be asking @Parthvi12 for cypress test for the exact scenario you have described here

Parthvi12:
@abhvsn I have added tests for this bug, will add the scenario @AnaghHegde mentioned in PR https://github.com/appsmithorg/appsmith/pull/12751 and merge it, after this PR is merged.

PR #12837, opened 2022-04-12
fix: Fixes Mongo connection pool minimum size to 0

### LINKED ISSUE #4142: Add retry limits for Mongo connection in the Mongo plugin

nidhi-nair:
The mongo plugin retries creating connections passively in the pool all the time right now because the pool is initialized with a setting to always have at least one connection open. This fix forces all datasources to allow 0 connections as well. This will allow incorrect datasource configurations to not keep attempting the connection. 

Also contains refactoring since the logic for URIs was crowding the `MongoPluginExecutor` class.

Do we also need to be checking for the duration of time that each connection is held on to even when idle?

Fixes #4142

## Type of change

## How Has This Been Tested?
- Unit tests for parameter additions
- Manually tested for connection recreation attempts

## Checklist:

### DIFF
--- a/app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
+++ b/app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
@@ -21,7 +21,6 @@ import com.appsmith.external.models.Param;
 import com.appsmith.external.models.ParsedDataType;
 import com.appsmith.external.models.Property;
 import com.appsmith.external.models.RequestParamDTO;
-import com.appsmith.external.models.SSLDetails;
 import com.appsmith.external.plugins.BasePlugin;
 import com.appsmith.external.plugins.PluginExecutor;
 import com.appsmith.external.plugins.SmartSubstitutionInterface;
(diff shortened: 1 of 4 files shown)


app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
```diff
@@ -140,7 +145,7 @@ public class MongoPlugin extends BasePlugin {
      *   - mongodb://user:pass@some-url:port,some-url:port,.../some-db...
      * - It has been grouped like this: (mongodb+srv://)(user):(pass)@(some-url)/(some-db...)?(params...)
      */
-    private static final String MONGO_URI_REGEX = "^(mongodb(?:\\+srv)?:\\/\\/)(?:(.+):(.+)@)?([^\\/\\?]+)\\/?([^\\?]+)?\\??(.+)?$";
+    private static final String MONGO_URI_REGEX = "^(mongodb(?:\\+srv)?://)(?:(.+):(.+)@)?([^/?]+)/?([^?]+)?\\??(.+)?$";
```
sumitsum, line 148: Just asking for confirmation - have we checked if it is ok to remove the back slashes.

sumitsum (approved):

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
nidhi-nair: Yes, the forward slash character does not need to be escaped at all, I've tested this with mongo uri string inputs as well. :)

Issue #9979, opened 2021-12-24
[Task]: Do not save js object variable data in the backend


ApekshaBhosale:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SubTasks

- js object variables need to be able to reset values
- create dependency with variables

Issue #9645, opened 2021-12-08
[Bug]: JS object does not read values from select widget when options are derived from a binding 


ramsaptami:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

JS object does not pick up the value from multi-option widgets such as select, checkbox, radio etc when values to those widgets are derived from a binding.

![image](https://user-images.githubusercontent.com/79509062/145172629-7dad2504-f142-4c0f-9612-d4f4bb91f1f9.png)

[![LOOM DEMO](http://cdn.loom.com/sessions/thumbnails/644e1aef17a94460ab1cfd2aca35ec47-00001.gif)](https://www.loom.com/share/644e1aef17a94460ab1cfd2aca35ec47)

### Steps To Reproduce

1. Create a JS object with an array of elements with value and label for select widget
2. Bind this to a select widget on the canvas
3. Use selected `value` or `label `property in another JS object to execute a `console.log` line and observe the value returned

### Environment

Release

### Version

Cloud

ApekshaBhosale:
we had discussion not to make js object variables reactive and partially it will be solved as part of #9979 by @ApekshaBhosale

Rishabh-Rathod:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/FE-Coders-Pod-6167d1b64ee5f20014242ab7/issues/appsmithorg/appsmith/9645?planning-poker) with ZenHub @ApekshaBhosale

PR #12999, opened 2022-04-18
fix: Saving variables as string rather than values in backend

ApekshaBhosale:
We were saving the js object variable's evaluated value in our backend which is a security breach. To solve this problem, we will be parsing using AST which stops variables from evaluating. 
Example - export default {
myVar1: Text1.text, 
}
Text1.text -> Label 

the value which was getting saved in the backend was `Label`  but now as part of this change, it will be `Text1.text`

Fixes https://github.com/appsmithorg/appsmith/issues/9979 
Fixes https://github.com/appsmithorg/appsmith/issues/9645

## Type of change

### Test Plan
- [ ] [[Test Case]: verify JS object read values from select widget when options are derived from binding](https://github.com/appsmithorg/TestSmith/issues/1899)
- [ ] [[Test Case]: verify JS object read values from Multi-Select widget when options are derived from binding](https://github.com/appsmithorg/TestSmith/issues/1900)
- Test A
- Test B

## Checklist:

    // Code coverage diff between base branch:release and head branch: new-parsing-change-js-var 
 :red_circle: | total | 56.59 **(0)** | 38.36 **(-0.01)** | 35.87 **(0.01)** | 56.83 **(-0.01)**
 :red_circle: | app/client/src/utils/JSPaneUtils.tsx | 88.66 **(-2.04)** | 81.25 **(-1.68)** | 75 **(-5)** | 87.18 **(-2.68)**
 :green_circle: | app/client/src/workers/ast.ts | 93.4 **(0.85)** | 77.5 **(0)** | 92 **(1.09)** | 94.79 **(0.67)**
 :red_circle: | app/client/src/workers/evaluationUtils.ts | 60.44 **(-1.8)** | 60.2 **(0)** | 66.15 **(-1.04)** | 58.35 **(-2.02)**
 :red_circle: | app/client/src/workers/DataTreeEvaluator/DataTreeEvaluator.ts | 75.3 **(-1.01)** | 55.72 **(-2.64)** | 74.16 **(0)** | 75.08 **(-1.07)**</details>


ramsaptami:
- [ ] JS object does not parse correctly and display correct function names in an object on copy pasted and deleted objects
https://www.loom.com/share/af994f9f305b426c8040cf16894f796b

- [ ] Async tag does not appear on using internal function

app/client/src/workers/ast.ts
```diff
@@ -299,3 +300,26 @@ const getPropertyAccessor = (propertyNode: IdentifierNode | LiteralNode) => {
...
 };
+
+export const parseJSObjectWithAST = (jsObjectBody: string) => {
+  const createString = `var jsObj = ${jsObjectBody}`;
+  const ast = parse(createString, { ecmaVersion: 11 });
+  const rawKeyValues: any = [];
+  const parsedObject: any = [];
+  simple(ast, {
+    Property(node: any) {
+      rawKeyValues.push({
```
eco-monk, line 311: We could do this directly?
```
      parsedObject.push({
        key: generate(node.key),
        value: generate(node.value),
        type: rawKeyValue.value.type,
      });
```

app/client/src/workers/ast.ts
```diff
@@ -299,3 +300,26 @@ const getPropertyAccessor = (propertyNode: IdentifierNode | LiteralNode) => {
     return `[${propertyNode.value}]`;
   }
 };
+
+export const parseJSObjectWithAST = (jsObjectBody: string) => {
+  const createString = `var jsObj = ${jsObjectBody}`;
+  const ast = parse(createString, { ecmaVersion: 11 });
```
eco-monk, line 306:
```suggestion
  const ast = parse(createString, { ecmaVersion: ECMA_VERSION });
```

eco-monk (commented):
Posting comments on just the ast part.
While I try to understand the remaining changes.

app/client/src/workers/DataTreeEvaluator/DataTreeEvaluator.ts
```diff
@@ -1031,72 +1055,78 @@ export default class DataTreeEvaluator {
...
-          });
+        if (!!result) {
+          if (!!parsedObject) {
+            parsedObject.forEach((parsedElement: any) => {
+              if (
+                parsedElement.type === "ArrowFunctionExpression" ||
+                parsedElement.type === "FunctionExpression"
+              ) {
+                try {
+                  const { result } = evaluateSync(
```
eco-monk, line 1070: Can we use `result[parsedElement.key]` instead of evaluating `parsedElement.value`?

app/client/src/workers/DataTreeEvaluator/DataTreeEvaluator.ts
```diff
@@ -723,6 +723,30 @@ export default class DataTreeEvaluator {
             _.set(currentTree, fullPropertyPath, evalPropertyValue);
             return currentTree;
           } else if (isJSAction(entity)) {
+            const variableList: Array<string> =
+              _.get(entity, "variables") || [];
+            if (variableList.indexOf(propertyPath) > -1) {
+              const getVarEvalValue = _.get(
```
eco-monk, line 729: nitpicks for more readability:
- renaming `getVarEvalValue` to `currentEvaluatedValue`
- creating `const evalValuePath = getEvalValuePath(fullPropertyPath, {
                  isPopulated: true,
                  fullPath: true,
                })` and re-use

app/client/src/workers/DataTreeEvaluator/DataTreeEvaluator.ts
```diff
@@ -1031,72 +1055,78 @@ export default class DataTreeEvaluator {
...
-        const errors = {
-          type: EvalErrorTypes.PARSE_JS_ERROR,
-          context: {
-            entity: entity,
-            propertyPath: entity.name + ".body",
-          },
-          message: e.message,
-        };
-        this.errors.push(errors);
+        //if we need to push error as popup in case
```
eco-monk, line 1129: I assume we're removing this because parse errors are already logged in our console.
Is this correct?
image

eco-monk: So what this would mean is, when a function with parse error is bound on click.
It won't show the pop-up error.

ApekshaBhosale: yes they were never used so removing it

app/client/src/workers/DataTreeEvaluator/DataTreeEvaluator.ts
ApekshaBhosale: we need to evaluate function in function form. parsedElement.value is string which will be useless which executing function

eco-monk (commented):
Please do add types for properties where we have used `any`.

ramsaptami:
- [ ] A variable in an initial error state, example when binding is incorrect {{CheckboxGroup.defaultSelectedValue}} then changed to  {{CheckboxGroup.defaultSelectedValues}} does not immediately reflect changes in evaluated value. When correct binding is changed to incorrect binding then we see that the changes are again not reflected
https://www.loom.com/share/c395022116b94c259b06ab7874086de1

Another way to repro this is by changing the binding, lets say it's Table1.pageSize and then changed to Table1.isVisible does not change the value returned by assigned variable

ramsaptami:
- [ ] Function does not return anything during the first function execution and only returns on the second run
https://www.loom.com/share/71f4c1726bfa4d22993b2aca2b6210ca

- [ ] Function that was previous async and then made sync is still executed onPageLoad
image

- [ ] When an API call is added to a function and when the function is parsed, we see an error on `response` tab but nothing appears on the `errors` tab of the debugger. Re-running the function clears this error and we can proceed
https://www.loom.com/share/0ceff5ca48544a83b1bbb1ab4935fc6f

ApekshaBhosale:
- [ ] https://github.com/appsmithorg/appsmith/issues/13197 this exists on the prod as well right? so le's not stop this from going to prod

eco-monk (approved):
Merged release and tests updated.

PR #13079, opened 2022-04-19
fix: Select widget's default selectedOptionLabel shows data from the value field and not the label field

### LINKED ISSUE #12225: [Bug]-[300]:Default value is displayed on widget even when it is not available in drop-down options

### LINKED ISSUE #12531: [Bug]-[1200]:Unable to select previously selected option from dropdown if value outside the options is set - unless other option is selected at-least once

### Description

https://user-images.githubusercontent.com/91450662/161390275-d6d4bed2-a643-4fdd-820f-d124a9668642.mov

### Steps To Reproduce

1. Drag drop a select widget - set some options for the dropdown
2. Map the Default value of the select to take values outside of existing options, example used - Table.selectedRow.step 
3. Now select step#1 from table - observe Select also reflects the value
4. **Now select step#3 from table -  observe Select also reflects the value (not one of the available select option)**
5. Now from select, try to change dropdown & select # 1 - one of the option value present
6. Observe User is unable to select # 1 unless # 2 selected
7. **Expected - All of Select option values should be available for selection anytime**

### Public Sample App

https://release.app.appsmith.com/applications/62486d45ab307a026918639e/pages/62486d45ab307a02691863a7

### Version

Prod/Release v1.6.17

### LINKED ISSUE #12554: [Bug]-[400]:Select widget's default selectedOptionLabel shows data from the value field and not the label field

### Description

When i drop a select widget and bind a text widget to use selectedOptionLabel, the value shown in text widget is the value and not the label of the item selected by default. 

However, this is fixed when I select any other value by interacting with the widget.

Also happening in the multi-select widget

### Steps To Reproduce

1. Add a new select widget
2. Bind the selectedOptionLabel to a text widget
3. See that it shows the incorrect data

[LOOM DEMO](https://www.loom.com/share/8950222cc6ee4824acad50fe41d96f6b)

### Public Sample App

_No response_

### Version

Production

wmdev0808:
Select related widgets are not exposing correct `selectedOptionLabel(s)` and `selectedOptionValue(s)` properties when:
- mounting for the first time
- changing `defaultOptionValue`
- changing `options`

> What is done:
- Always gets `selectedOptionLabel(s)` and `selectedOptionValue(s)` in cases of the above scenarios
- Even calculate nested labels and values with MultiSelectTreeWidget

Fixes #12554 
Fixes #12225 
Fixes #12531
## Type of change

## How Has This Been Tested?

- Test A
- Test B

## Checklist:

    // Code coverage diff between base branch:release and head branch: fix/12554-select-selectedOptionLabel 
 :green_circle: | total | 56.59 **(0.03)** | 38.65 **(0.11)** | 35.94 **(0.07)** | 56.84 **(0.04)**
 :red_circle: | app/client/src/utils/WorkerUtil.ts | 89.76 **(0)** | 70.59 **(-1.96)** | 100 **(0)** | 93.33 **(0)**
 :green_circle: | app/client/src/widgets/WidgetUtils.ts | 83.5 **(0.69)** | 63.25 **(0.64)** | 68.75 **(2.08)** | 80 **(0.89)**
 :green_circle: | app/client/src/widgets/MultiSelectTreeWidget/widget/index.tsx | 47.17 **(6.63)** | 16.13 **(16.13)** | 35 **(-0.29)** | 48.48 **(7.05)**
 :sparkles: :new: | **app/client/src/widgets/MultiSelectWidgetV2/widget/derived.js** | **100** | **90.63** | **100** | **100**
 :green_circle: | app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx | 72.07 **(-1.67)** | 52.94 **(4.37)** | 60 **(2.86)** | 71.43 **(-1.97)**
 :sparkles: :new: | **app/client/src/widgets/MultiSelectWidgetV2/widget/parsedDerivedProperties.ts** | **38.46** | **0** | **0** | **38.46**
 :green_circle: | app/client/src/widgets/SelectWidget/widget/index.tsx | 62.26 **(9.82)** | 46.84 **(25.87)** | 50 **(7.89)** | 63.11 **(9.95)**
 :green_circle: | app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx | 52.63 **(5.66)** | 0 **(0)** | 42.86 **(2.86)** | 53.7 **(5.31)**</details>


app/client/src/widgets/SelectWidget/widget/index.tsx
```diff
@@ -412,17 +409,29 @@ class SelectWidget extends BaseWidget<SelectWidgetProps, WidgetState> {
...
-  componentDidMount() {
-    super.componentDidMount();
-    this.changeSelectedOption();
+  componentDidMount(): void {
+    this.sanitizeSelectedOption();
   }
 
   componentDidUpdate(prevProps: SelectWidgetProps): void {
+    // Sanitizes selectedOption if options changes
+    if (xorWith(this.props.options, prevProps.options, isEqual).length > 0) {
```
ashit-rath, line 423: When serverside filtering is enabled, the options would change. Now if the new options does not have the selectedOption then we are going to remove it and post that if the options do come back we won't be showing it.

cc: @dilippitchika

ashit-rath (commented):
I am submitting a partial review here as I thing we need to re-think and re-work on the last comment.

app/client/src/widgets/SelectWidget/widget/index.tsx
```diff
@@ -393,17 +393,14 @@ class SelectWidget extends BaseWidget<SelectWidgetProps, WidgetState> {
...
+      selectedOption: "defaultOptionValue",
       filterText: "",
     };
   }
 
   static getMetaPropertiesMap(): Record<string, any> {
     return {
-      value: undefined,
-      label: undefined,
+      selectedOption: undefined,
```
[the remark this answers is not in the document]
wmdev0808: Because we are keeping the state for them in one object meta property, "selectedOption"

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -485,29 +486,48 @@ class MultiSelectWidget extends BaseWidget<
 
   static getDerivedPropertiesMap() {
     return {
-      selectedOptionLabels: `{{ this.selectedOptions ? this.selectedOptions.map((o) => _.isNil(o.label) ? o : o.label ) : [] }}`,
-      selectedOptionValues: `{{ this.selectedOptions ? this.selectedOptions.map((o) =>  _.isNil(o.value) ? o : o.value  ) : [] }}`,
+      selectedOptions: `{{
+        this.options.filter(option =>
+          this.selectedOptionsArray.map(
+            selectedOption => selectedOption.value || selectedOption
+          ).includes(option.value)
```
Tooluloope, line 493: We need to check if serverSide filtering Is enabled.

Please look through this comment and let me know if it's clear
https://github.com/appsmithorg/appsmith/issues/12225#issuecomment-1082910936

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -485,29 +486,48 @@ class MultiSelectWidget extends BaseWidget<
...
+        (selectedOption) =>
+          !!find(this.props.options, {
+            value: selectedOption.value || selectedOption,
+          }),
+      );
+      this.props.updateWidgetMetaProperty(
+        "selectedOptionsArray",
+        updatedSelectedOptionsArray,
+      );
+    }
```
Tooluloope, line 530: The same applies to these changes, in Serverside Filtering, previous Values would be misplaced.
Do check the comment if that solution is best, else you could propose a new solution

app/client/src/widgets/SelectWidget/widget/index.tsx
```diff
@@ -476,50 +483,31 @@ class SelectWidget extends BaseWidget<SelectWidgetProps, WidgetState> {
...
+    this.props.updateWidgetMetaProperty("selectedOption", selectedOption, {
+      triggerPropertyName: "onOptionChange",
+      dynamicString: this.props.onOptionChange,
+      event: {
+        type: EventType.ON_OPTION_CHANGE,
+      },
+    });
+    if (!this.props.isDirty) {
+      this.props.updateWidgetMetaProperty("isDirty", true);
+    }
```
Tooluloope, line 510: is there a reason why the `isChanged` logic was removed?

At the moment, selecting the same option would always fire OptionChanged and isDirty

app/client/src/widgets/SelectWidget/widget/index.tsx
```diff
@@ -476,50 +483,31 @@ class SelectWidget extends BaseWidget<SelectWidgetProps, WidgetState> {
...
-          type: EventType.ON_OPTION_CHANGE,
-        },
-      });
-      if (!this.props.isDirty) {
-        this.props.updateWidgetMetaProperty("isDirty", true);
-      }
-    }
+  setSelectedOption = () => {
+    const matchingOption = find(this.props.options, {
+      value: this.props.selectedOption?.value || this.props.selectedOption,
```
Tooluloope, line 495: if the value is `0` this would fail, Let's be careful with this.

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -485,29 +486,48 @@ class MultiSelectWidget extends BaseWidget<
...
       isDirty: false,
     };
   }
 
   componentDidUpdate(prevProps: MultiSelectWidgetProps): void {
+    if (xorWith(this.props.options, prevProps.options).length > 0) {
+      const updatedSelectedOptionsArray = this.props.selectedOptionsArray.filter(
+        (selectedOption) =>
+          !!find(this.props.options, {
+            value: selectedOption.value || selectedOption,
```
Tooluloope, line 523: if the value is `0` this would fail, Let's be careful with this.

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -485,29 +486,48 @@ class MultiSelectWidget extends BaseWidget<
 
   static getDerivedPropertiesMap() {
     return {
-      selectedOptionLabels: `{{ this.selectedOptions ? this.selectedOptions.map((o) => _.isNil(o.label) ? o : o.label ) : [] }}`,
-      selectedOptionValues: `{{ this.selectedOptions ? this.selectedOptions.map((o) =>  _.isNil(o.value) ? o : o.value  ) : [] }}`,
+      selectedOptions: `{{
+        this.options.filter(option =>
+          this.selectedOptionsArray.map(
+            selectedOption => selectedOption.value || selectedOption
```
Tooluloope, line 492: if the value is `0` this would fail, Let's be careful with this.

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
```diff
@@ -477,15 +489,39 @@ class SingleSelectTreeWidget extends BaseWidget<
...
+  flatOptions(options: DropdownOption[]) {
+    let result: { label: string; value: string | number }[] = [];
+    options.forEach((option) => {
+      result.push({ label: option.label, value: option.value });
+      if (Array.isArray(option.children)) {
+        result = result.concat(this.flatOptions(option.children));
+      }
+    });
+    return result;
+  }
```
Tooluloope, line 501: `Flat()` and `FlatOptions()` are similar, Could we re-use 1 of the two functions across the Widget

Or better still, add the function to widgetUtils and use in both TreeSelect widgets

Tooluloope (changes requested):
Hi @wmdev0808 could you use the nullish check rather than the falsy Check in your PR

app/client/src/widgets/MultiSelectTreeWidget/widget/index.tsx
```diff
@@ -513,6 +550,72 @@ class MultiSelectTreeWidget extends BaseWidget<
...
+      selectedOptions.map((option) => option.value),
+    );
+    this.props.updateWidgetMetaProperty("selectedLabel", selectedOptionLabels);
+    this.props.updateWidgetMetaProperty(
+      "selectedOptionValueArr",
+      selectedOptionValues,
+    );
+  }
+
+  getOptionSubTree(options: DropdownOption[], value: string | number) {
```
Tooluloope, line 588: Could you write a Unit Test for this method

Tooluloope (changes requested):
Could you write a test for serverSide filtering for Multiselect and Select, So we're sure nothing would break

app/client/src/widgets/SelectWidget/widget/index.tsx
ashit-rath: @Tooluloope Any thoughts on this?

Tooluloope: Yeah, You're right @ashit-rath 
I also raise a comment during my review.

@wmdev0808 do check this comment https://github.com/appsmithorg/appsmith/issues/12225#issuecomment-1082910936 for the proposed solution.

If you feel this can be implemented better do let us know before implementation.

@ashit-rath do you think this https://github.com/appsmithorg/appsmith/issues/12225#issuecomment-1082910936 is okay?

app/client/src/widgets/MultiSelectTreeWidget/widget/index.tsx
```diff
@@ -513,6 +550,72 @@ class MultiSelectTreeWidget extends BaseWidget<
...
+    if (targetOption) {
+      result = result.concat({
+        label: targetOption.label,
+        value: targetOption.value,
+      });
+      if (Array.isArray(targetOption.children)) {
+        // If found, Finds all decendants for the found target
+        result = result.concat(this.flatOptions(targetOption.children));
+      }
+      return result;
```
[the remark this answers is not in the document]
wmdev0808: I can't understand why you consider `Mode` for selectedOptionsValue.
That should always includes subtrees regardless of `Mode`, I think

app/client/src/widgets/MultiSelectTreeWidget/widget/index.tsx
```diff
@@ -513,6 +550,72 @@ class MultiSelectTreeWidget extends BaseWidget<
...
+    }
+    // Cannot find the target on the 1st level of the options tree
+    // Until finding the target or reaching out the leaves, loops
+    options.every((option) => {
```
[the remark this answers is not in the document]
wmdev0808: Because we don't have to traverse whole option tree, should break early if found without further proceeding.
Foreach cannot be used for that purpose

wmdev0808:
> Hi @wmdev0808 could you use the nullish check rather than the falsy Check in your PR

Could you please specify exactly which one you mean?

app/client/src/widgets/MultiSelectTreeWidget/widget/index.tsx
```diff
@@ -422,13 +421,51 @@ class MultiSelectTreeWidget extends BaseWidget<
...
+      xor(this.props.selectedOptionValueArr, prevProps.selectedOptionValueArr)
+        .length > 0 &&
+      this.props.isDirty === false
     ) {
-      this.props.updateWidgetMetaProperty("isDirty", false);
+      // Sets selectedLabel
+      this.setSelectedOptions(
+        this.props.options,
+        this.props.selectedOptionValueArr,
+      );
```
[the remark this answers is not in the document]
wmdev0808: When user select from the UI, we don't have to calculate subtree for the selected options.
However, if user changes defaultOptionValue, it leads to changing `selectedOptionValueArr`, we need to calculate subtrees.
After all this is just to identify the above two cases

app/client/src/widgets/SelectWidget/widget/index.tsx
wmdev0808: Please provide me with a demonstrating example

app/client/src/widgets/MultiSelectTreeWidget/widget/index.tsx
```diff
@@ -513,6 +550,72 @@ class MultiSelectTreeWidget extends BaseWidget<
...
+    let selectedOptions: { label: string; value: string | number }[] = [];
+    selectedValues.forEach((selectedValue) => {
+      selectedOptions = selectedOptions.concat(
+        this.getOptionSubTree(options, selectedValue),
+      );
+    });
+
+    const selectedOptionLabels = uniq(
+      selectedOptions.map((option) => option.label),
+    );
```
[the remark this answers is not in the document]
wmdev0808: Here, _.uniq is used to eliminate duplications produced by calculation

app/client/src/widgets/SelectWidget/widget/index.tsx
ashit-rath: Yeah, I was also thinking about caching but we have to have to make sure we cache it in the right way to avoid unnecessary re-rendering.

app/client/src/widgets/SelectWidget/widget/index.tsx
```diff
@@ -412,17 +409,32 @@ class SelectWidget extends BaseWidget<SelectWidgetProps, WidgetState> {
...
+        value: this.props.selectedOption.value,
+      });
+      if (!found) {
+        this.props.updateWidgetMetaProperty("selectedOption", {});
+      }
+    }
+    // Sets selectedOption
+    if (!isEqual(this.props.selectedOption, prevProps.selectedOption)) {
+      this.setSelectedOption();
+    }
```
Tooluloope, line 437: Same here, Moving this to Derived Property Calculation

app/client/src/widgets/SelectWidget/widget/index.tsx
```diff
@@ -476,13 +486,20 @@ class SelectWidget extends BaseWidget<SelectWidgetProps, WidgetState> {
...
     );
   }
 
+  setSelectedOption = () => {
+    const matchingOption = find(this.props.options, {
+      value: this.props.selectedOption.value ?? this.props.selectedOption,
+    });
+    this.props.updateWidgetMetaProperty("selectedOption", matchingOption || {});
+  };
+
```
Tooluloope, line 502: This also does strict checking which should be

Tooluloope (changes requested):

app/client/src/widgets/MultiSelectTreeWidget/widget/getOptionSubTree.test.ts
```diff
@@ -0,0 +1,124 @@
...
+      {
+        label: "Light Blue",
+        value: "LIGHT BLUE",
+      },
+    ];
+
+    const output = getOptionSubTree(inputOptions, inputValue, inputMode);
+    expect(output).toEqual(expected);
+  });
+});
```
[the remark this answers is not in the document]
wmdev0808: When I use this input, all the test cases even pass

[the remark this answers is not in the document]
Tooluloope: Try running this on the application and see the result

https://replit.com/@TOLULOPEADETULA/InsistentCookedIdentifier#index.ts

app/client/src/widgets/SelectWidget/widget/index.tsx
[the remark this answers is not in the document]
ashit-rath: Wouldn't removing `value` cause friction in the Form widget as it depends on the `value` property?

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -507,7 +515,26 @@ class MultiSelectWidget extends BaseWidget<
...
+        (selectedOption) =>
+          !!find(this.props.options, {
+            value: selectedOption.value,
+          }),
+      );
+      this.props.updateWidgetMetaProperty(
+        "selectedOptions",
+        updatedSelectedOptions,
+      );
+    }
```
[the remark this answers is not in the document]
ashit-rath: Yes, I agree. Is it also possible to move the logic in componentDidMount to derived properties?

app/client/src/widgets/MultiSelectTreeWidget/widget/getOptionSubTree.test.ts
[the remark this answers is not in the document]
wmdev0808: What is wrong with the above?

app/client/src/widgets/SelectWidget/widget/index.tsx
```diff
@@ -393,17 +393,14 @@ class SelectWidget extends BaseWidget<SelectWidgetProps, WidgetState> {
 
   static getDefaultPropertiesMap(): Record<string, string> {
     return {
-      defaultValue: "defaultOptionValue",
-      value: "defaultOptionValue",
```
Tooluloope, line 397: @wmdev0808 Please Add Value back, this was kept for form changes

app/client/src/widgets/SelectWidget/widget/index.tsx
```diff
@@ -412,37 +409,51 @@ class SelectWidget extends BaseWidget<SelectWidgetProps, WidgetState> {
...
     ) {
       this.props.updateWidgetMetaProperty("isDirty", false);
     }
-  }
+    if (
+      !this.props.serverSideFiltering &&
+      xorWith(this.props.options, prevProps.options, isEqual).length > 0
+    ) {
+      const found = find(this.props.options, {
+        value: this.props.selectedOption.value,
```
Tooluloope, line 434: this should be `value: this.props.selectedOption.value ?? this.props.selectedOption

Tooluloope (commented):
I think after these it should be Okay.

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/FormWidgets/SelectWidgets_selectedOptionLabels_spec.js
```diff
@@ -0,0 +1,353 @@
...
+    );
+    cy.get(`.t--widget-textwidget`)
+      .eq(2)
+      .should("not.contain.text", "Blue")
+      .and("contain.text", "Red");
+    cy.get(`.t--widget-textwidget`)
+      .eq(3)
+      .should("not.contain.text", "BLUE")
+      .and("contain.text", "RED");
+  });
```
Tooluloope, line 233: Same thing here, Why shouldn't it contain Red?

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/FormWidgets/SelectWidgets_selectedOptionLabels_spec.js
```diff
@@ -0,0 +1,353 @@
...
+    );
+    cy.get(`.t--widget-textwidget`)
+      .eq(6)
+      .should("contain.text", "Light Blue")
+      .and("contain.text", "Red");
+    cy.get(`.t--widget-textwidget`)
+      .eq(7)
+      .should("contain.text", "LIGHT BLUE")
+      .and("contain.text", "RED");
+  });
```
Tooluloope, line 352: Could you add tests for the Mode changes

app/client/src/widgets/SelectWidget/widget/index.tsx
Tooluloope: @wmdev0808 Should value be here or in derivedProperties,
Here, value would always be equal to the initial default Value

app/client/src/widgets/SelectWidget/widget/index.tsx
```diff
@@ -492,13 +523,7 @@ class SelectWidget extends BaseWidget<SelectWidgetProps, WidgetState> {
       isChanged = !(this.props.selectedOptionValue === selectedOption.value);
     }
     if (isChanged) {
-      if (!this.props.isDirty) {
-        this.props.updateWidgetMetaProperty("isDirty", true);
-      }
-
-      this.props.updateWidgetMetaProperty("label", selectedOption.label ?? "");
-
-      this.props.updateWidgetMetaProperty("value", selectedOption.value ?? "", {
```
Tooluloope, line 501: Value was previously been updated, Since we are not updating it anymore, please move it to derived properties
``` value: `{{this.selectedOption.value}}` ```

**Review by Tooluloope — CHANGES_REQUESTED** [review]

File: app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/FormWidgets/SelectWidgets_selectedOptionLabels_spec.js [hunk 24]
```diff
@@ -0,0 +1,353 @@
...
+          "value": "RED"
+        }
+      ]`,
+    );
+    cy.get(`.t--widget-textwidget`)
+      .eq(0)
+      .should("not.contain.text", "Blue");
+    cy.get(`.t--widget-textwidget`)
+      .eq(1)
+      .should("not.contain.text", "BLUE");
```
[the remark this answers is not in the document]
wmdev0808: > At the prior step, BLUE is removed from selectedOption by excluding it from options. So after restoring options, it should not be restored as far as severSideFiltering is off

After all, the assertion is to check the removed option in the earlier step should not restored after its options are restored

Tooluloope:
@wmdev0808 Here's the recording
https://user-images.githubusercontent.com/31691737/165291629-2675fe59-4988-4e0d-9de0-b0dbf700f88a.mov

app/client/src/widgets/SelectWidget/widget/index.tsx
```diff
@@ -476,13 +509,34 @@ class SelectWidget extends BaseWidget<SelectWidgetProps, WidgetState> {
...
+        this.props.updateWidgetMetaProperty(
+          "selectedOption",
+          this.props.defaultOptionValue,
+        );
+      } else {
+        this.props.updateWidgetMetaProperty("selectedOption", {});
+      }
+    }
+  };
+
```
ashit-rath, line 539: Please break this whole thing down to 2 functions, similar to the multi-select widget and add test to the `getSelectedOption`.
Please do write test to cover https://github.com/appsmithorg/appsmith/issues/12225#issuecomment-1110526570

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
```diff
@@ -374,20 +373,39 @@ class SingleSelectTreeWidget extends BaseWidget<
...
+    ) {
+      const found = find(this.props.options, {
+        value: this.props.selectedOption.value,
+      });
+      if (!found) {
+        this.props.updateWidgetMetaProperty("selectedOption", {});
+      }
+    }
+    // Sanitizes selectedOption
+    if (!isEqual(this.props.selectedOption, prevProps.selectedOption)) {
```
ashit-rath, line 406: Why are we checking this?

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
```diff
@@ -399,17 +417,13 @@ class SingleSelectTreeWidget extends BaseWidget<
   getPageView() {
     const options =
       isArray(this.props.options) &&
-      !this.props.__evaluation__?.errors.options.length
+      !this.props.__evaluation__?.errors.options?.length
```
ashit-rath, line 420: We shouldn't be looking into the `__evaluation__` object. Isn't there proper validation for `options`?

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
```diff
@@ -441,74 +455,66 @@ class SingleSelectTreeWidget extends BaseWidget<
...
   selectedIndex?: number;
-  options?: DropdownOption[];
+  options: DropdownOption[];
   onOptionChange: string;
   defaultOptionValue: string;
   isRequired: boolean;
   isLoading: boolean;
   allowClear: boolean;
   selectedLabel: string[];
-  selectedOption: string | number;
```
ashit-rath, line 511: Why did we remove `selectedOption`?

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
```diff
@@ -441,74 +455,66 @@ class SingleSelectTreeWidget extends BaseWidget<
...
-  value: string | number;
-  disabled?: boolean;
-  children?: DropdownOption[];
-}
-
 export interface SingleSelectTreeWidgetProps extends WidgetProps {
   placeholderText?: string;
   selectedIndex?: number;
-  options?: DropdownOption[];
+  options: DropdownOption[];
```
ashit-rath, line 511: Are we sure options would always be there?
If so then why do we have such checks [here](https://github.com/appsmithorg/appsmith/pull/13079/files#diff-477fa90c63460418797f2ccb1a946ff6789106e8272d63523808afc5dfa6d7a0R420)?

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
```diff
@@ -441,74 +455,66 @@ class SingleSelectTreeWidget extends BaseWidget<
...
-      if (Array.isArray(a.children)) {
-        result = result.concat(this.flat(a.children));
-      }
-    });
-    return result;
-  }
-
-  filterValue(value: string | number | undefined) {
-    const options = this.props.options ? this.flat(this.props.options) : [];
+  setSelectedOptionLabel(
```
ashit-rath, line 482: This isn't getting called from anywhere.

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
```diff
@@ -374,20 +373,39 @@ class SingleSelectTreeWidget extends BaseWidget<
...
       selectedOption: "defaultOptionValue",
-      selectedLabel: "defaultOptionValue",
     };
   }
 
   static getMetaPropertiesMap(): Record<string, any> {
     return {
       selectedOption: undefined,
-      selectedOptionValueArr: undefined,
-      selectedLabel: [],
```
ashit-rath, line 385: Why are we removing from here? Is this not required?
cc: @Tooluloope

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
```diff
@@ -441,74 +455,66 @@ class SingleSelectTreeWidget extends BaseWidget<
...
-      const index = findIndex(options, { value: value as string });
-      return index > -1 ? value : undefined;
-    }
+    this.props.updateWidgetMetaProperty(
+      "selectedLabel",
+      selectedLabel ? [selectedLabel] : [],
+    );
   }
 
+  setSelectedOption = () => {
```
ashit-rath, line 496: This function is not about setting selected option. 
`selectedOption` gets updated with a value only when a user selects something in the dropdown with a real value.

Here we are looping through the options to find an option that matches with the `selectedOption` and then update that value as the selectedOption? If there is a match wouldn't that value and `selectedOption` (that we use to find) would always be the same?

This function's use case is to unset `selectedOption`. If the `selectedOption` is not present in the option then update with `{}` to unset the value.

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
```diff
@@ -374,20 +373,39 @@ class SingleSelectTreeWidget extends BaseWidget<
...
+    this.setSelectedOption();
+  }
+
+  componentDidUpdate(prevProps: SingleSelectTreeWidgetProps) {
+    if (
+      xorWith(
+        flattenOptions(this.props.options),
+        flattenOptions(prevProps.options),
+        isEqual,
+      ).length > 0
```
ashit-rath, line 396: This is a very expensive operation and what makes this expensive is
1. Flattening of options
2. Xor with deep-equal
3. It would happen twice (prev option and current option)
4. the number of times `componentDidUpdate` runs.

Alternative approach: When options are changed, in the property config of option, in the `updateHook` we can set a flag to true denoting options changed. And here we can check if that flag is set; update accordingly and reset the flat
There could be an even better solution than this.
cc: @Tooluloope

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
```diff
@@ -374,20 +373,39 @@ class SingleSelectTreeWidget extends BaseWidget<
...
+    ) {
+      const found = find(this.props.options, {
+        value: this.props.selectedOption.value,
+      });
+      if (!found) {
+        this.props.updateWidgetMetaProperty("selectedOption", {});
+      }
```
ashit-rath, line 403: Isn't `setSelectedOption` for doing this?

app/client/src/widgets/MultiSelectTreeWidget/widget/index.tsx
```diff
@@ -422,13 +473,49 @@ class MultiSelectTreeWidget extends BaseWidget<
...
-      this.props.updateWidgetMetaProperty("isDirty", false);
+      this.setSelectedOptions(
+        this.props.options,
+        this.props.selectedOptionValueArr,
+      );
+    }
+    if (
+      xor(this.props.selectedOptionValueArr, prevProps.selectedOptionValueArr)
+        .length > 0 &&
+      this.props.isDirty === false
```
ashit-rath, line 507: `!this.props.isDirty`

app/client/src/widgets/MultiSelectTreeWidget/widget/index.tsx
```diff
@@ -422,13 +473,49 @@ class MultiSelectTreeWidget extends BaseWidget<
...
+  }
+
+  componentDidUpdate(prevProps: MultiSelectTreeWidgetProps) {
     if (
       xor(this.props.defaultOptionValue, prevProps.defaultOptionValue).length >
-        0 &&
-      this.props.isDirty
+      0
+    ) {
+      if (this.props.isDirty) {
```
ashit-rath, line 488: This can be clubbed with the previous condition

```js
if (xor(...) && this.props.isDirty) 
```

ashit-rath (commented):
@Tooluloope  I have skimmed through the MultiSelectTreeWidget here. Please do have a deeper look if required.

app/client/src/icons/WidgetIcons.tsx
```diff
@@ -0,0 +1,190 @@
+import React, { JSXElementConstructor } from "react";
```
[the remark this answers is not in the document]
wmdev0808: It should be removed

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
wmdev0808: No, this function is to always set/update a valid/normalized `selectedOption` object in the following use cases:
- When mounting
- When `defaultOptionValue` changes
- When `options` changes

NB: I will refactor this part for your better understanding

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
wmdev0808: @ashit-rath ,
First, thank you for your solution.
However, how can we check if `options` property has actually changed in `updateHook`?
Without the above computation, we cannot decide if the value has actually changed

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
ashit-rath: @wmdev0808  Can you give an example where the `matchingOption` and the `this.props.selectedOption` be different and the `matchingOption` is not `{}`?

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
ashit-rath: `updateHook` only runs when the property value changes.

app/client/src/widgets/SelectWidget/widget/index.tsx
```diff
@@ -492,13 +546,7 @@ class SelectWidget extends BaseWidget<SelectWidgetProps, WidgetState> {
...
     }
     if (isChanged) {
-      if (!this.props.isDirty) {
-        this.props.updateWidgetMetaProperty("isDirty", true);
-      }
-
-      this.props.updateWidgetMetaProperty("label", selectedOption.label ?? "");
-
-      this.props.updateWidgetMetaProperty("value", selectedOption.value ?? "", {
+      this.props.updateWidgetMetaProperty("selectedOption", selectedOption, {
```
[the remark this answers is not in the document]
wmdev0808: There is still a code snippet for isDirty update

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
wmdev0808: The purpose is to catch `defaultOptionValue` changes because it leads to changing selectedOption.
In addition, I've refactored that part again

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
wmdev0808: This code is from @Tooluloope ,
you should confirm with him

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
wmdev0808: I mean for the following case, for example:
- User just add/or remove a space, resulting in the same evaluation

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
wmdev0808: right

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
wmdev0808: For example, 
- If defaultOptionValue can be simple string, "BLUE" for the first time, but matchingOption is always object

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
wmdev0808: We are using selectedOption as a source of true for selectedOptionLabel and selectedOptionValue, so they are not required

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
wmdev0808: I think so.
How is that check  related with options?

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
wmdev0808: I will add it

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
ashit-rath: Still, it's better than doing deep checking in `componentDidUpdate` which would still run if one add/remove a space. It would also run for every tiny update which is not related to the options change

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
ashit-rath: Isn't `defailtOptionValue`  always an object? Don't we have similar validation like in the Select widget where we compare with the options?

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -485,8 +485,19 @@ class MultiSelectWidget extends BaseWidget<
...
+          : this.selectedOptions.map(o => o.label ?? o)
+      }}`,
+      selectedOptionValues: `{{
+        !this.serverSideFiltering
+          ? this.selectedOptions?.filter(
+            selectedOption => this.options?.some(
+              option => option.value === (selectedOption.value ?? selectedOption)
+            )).map(o => o.value ?? o)
+          : this.selectedOptions.map(o => o.value ?? o)
+      }}`,
```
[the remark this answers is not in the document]
ashit-rath: @wmdev0808 This comment isn't addressed

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -584,6 +593,35 @@ class MultiSelectWidget extends BaseWidget<
     );
   }
 
+  setSelectedOptions = () => {
+    let value: DropdownOption[] = [];
+    if (!this.props.serverSideFiltering) {
+      const matchingOptions = this.props.selectedOptions.reduce(
```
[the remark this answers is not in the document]
ashit-rath: This isn't addressed.

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -584,6 +593,35 @@ class MultiSelectWidget extends BaseWidget<
...
+        (prev: DropdownOption[], current) => {
+          const matchingOption = find(this.props.options, {
+            value: current.value ?? current,
+          }) as DropdownOption;
```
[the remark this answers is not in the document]
ashit-rath: @wmdev0808 Can you let us know about this?

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -584,6 +593,35 @@ class MultiSelectWidget extends BaseWidget<
...
+          ({
+            label: option.label ?? option,
+            value: option.value ?? option,
+          } as DropdownOption),
+      );
+      value = optionValue;
+    }
+    this.props.updateWidgetMetaProperty("selectedOptions", value);
+  };
+
```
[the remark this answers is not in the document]
ashit-rath: @wmdev0808 Can you address this comment?

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
ashit-rath: We need to verify if `selectedLabel` is used by any user. If so then we cannot remove it as it may break their app.
@dilippitchika Can we verify this in the mixpanel?

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
ashit-rath: Please do make the necessary changes

Tooluloope:
Kinda hard to get where my attention is needed here.
@wmdev0808 Could you please tag me again.

@ashit-rath thanks for helping with the reviews

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
ashit-rath: If that's the case then this is not the way of checking. 
The `selectedOption` also changes when the user clicks an option. In that case this block would unnecessary run.

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
Tooluloope: @wmdev0808 You could make it a derived property just like selectedOptionLabel, but an array.

Although, I really think we should remove it, cos it shouldn't be accessible to the users

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
Tooluloope: This was quite a long time, we could refactor this if there's something better

ashit-rath (commented):
@wmdev0808  Why is this PR up for review again if most of the major comments are not addressed?
Please push all changes, do a self-review, and then request for a review otherwise we would unnecessary go back and forth.

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
ashit-rath: We should fix the validation fn of the `options` here similar to SingleSelect. We are checking for errors here because we aren't parsing it properly.

app/client/src/widgets/SingleSelectTreeWidget/widget/index.tsx
ashit-rath: Agreed @Tooluloope, we should work towards removing it. But the problem is that all the properties of widgets are exposed by default. If we suddenly remove it, it may cause friction; which we should avoid and deprecate properties gracefully. 
Further, we only have statistics for cloud users not the self hosted.

app/client/src/widgets/WidgetUtils.test.ts
```diff
@@ -209,3 +209,60 @@ describe(".sanitizeKey", () => {
     });
   });
 });
+
+describe(".flattenOptions", () => {
+  it("Flattens the nested options array", () => {
+    const input = [
```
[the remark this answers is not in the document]
wmdev0808: The followings are out of this function features, it should be taken from validation logic. 
- null from 1
- 3, 4, 5

app/client/src/widgets/SelectWidget/widget/index.tsx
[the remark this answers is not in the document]
ashit-rath: Can you share where this is hapenning?

PR #13084, opened 2022-04-19
fix: trigger consuming previous evaluated values

### LINKED ISSUE #12318: [Bug]: Delay in capturing input value

### Description

Input value is not sent properly to an action.
e.g. if we type "abc" only "a" was getting sent to the function.

User is using `appsmith.store` to store the value and display it somewhere. Problem seems to be from debounce we have and we can not remove the debounce.

Slack thread: https://theappsmith.slack.com/archives/C02K0SZQ7V3/p1648215497106399

[LOOM DEMO](https://www.loom.com/share/e84cc1b54c6646bda47572ff61cdce30)

### Steps To Reproduce

**Setup**
- Create a moderately complex app, you can clone our Aforce app.
- Add an input field, button, and a text widget
- On button click save the input value to appsmith store
- Show the stored stored value in the text widget

**Action**
- Type into the input, and click the submit button very fast.
- Try this multiple times.

**Observation**
- You will see that in some cases text widget is not updating correctly. It doesn't have the complete typed text

Note: Though the recording is on editor, it happens in view mode as well.

### Public Sample App

https://dev.appsmith.com/applications/624151206c86dd5be6fdd5a9/pages/624151216c86dd5be6fdd5c4

### Version

cloud & self hosted

### LINKED ISSUE #12365: [Bug]: `resetWidget` fails to reset the current page number of the Table Widget

### Description

A user wants to reset the `pageNo` of the table. This should be possible with the `resetWidget` function.

#### Use Case -
The user has external filters that operate on the table. Every time they change the filter, the pageNo should start from 1. But it stays to what it was before changing the filter (even after using resetWidget). 

### Steps To Reproduce

1. Create a table & add pagination
2. Go to next page (2)
3. Try to reset it on the external button click.
4. You'll see `pageNo` does not get reset back to 1

### Public Sample App

_No response_

### Version

Cloud

### LINKED ISSUE #12903: [Bug]: input values getting truncated

### Description

User reported issue, and prima facie this seems to be unrelated to #12318 

1. application has a form with some input fields and a submit button
2. when user enters some values in these widgets and clicks submit, an insert query is called
3. sporadically, the values from these inputs are truncated. For example, "Hotel" is stored as "Ho", and "200" is stored as "2"

please note: The user is directly using the format `input1.text` to access these values in the insert query, and they don't use any store values.

Some things we investigated along with the user:
1. the problems occurs randomly, on random fields
2. this is not related to delay time, because even if the user waits 1 minute before submitting the form, sometimes the truncated values are being persisted to the db

PS: I have a copy of the app json if needed.

### Steps To Reproduce

explained above

### Public Sample App

_No response_

### Version

Cloud

Rishabh-Rathod:
## Problem

We noticed on trigger events like for example `onTextChange` ,`onSubmit` latest evaluated values were not consumed, and instead previous evaluated values were passed. This led to incorrect values getting saved and passed to the queries / function / `storeValue`, etc.

## Solution

We are changing how we maintain meta values in our architecture, 
- Earlier `metaHOC` component state was used to store meta values that were debounced and pushed in batch to `metaReducer` and trigger evaluation.
- We remove the `metaHOC` state and directly update values to `metaReducer` and trigger evaluation separately via a debouched function.

Screenshot 2022-05-10 at 2 07 38 PM
Screenshot 2022-05-10 at 2 07 45 PM

Fixes #12318 
Fixes #12903 
Fixes #12365

## Type of change

### Test Plan

1. Cypress 

- [x] https://github.com/appsmithorg/TestSmith/issues/1700
- [x] https://github.com/appsmithorg/TestSmith/issues/1701
- [x] https://github.com/appsmithorg/TestSmith/issues/1702
- [x] https://github.com/appsmithorg/TestSmith/issues/1704
- [x] https://github.com/appsmithorg/TestSmith/issues/1705
- [x] https://github.com/appsmithorg/TestSmith/issues/1706
- [ ] https://github.com/appsmithorg/TestSmith/issues/1707
- [ ] https://github.com/appsmithorg/TestSmith/issues/1708
- [ ] https://github.com/appsmithorg/TestSmith/issues/1709

2. Jest
- [x] metaReducer
- [x] evalMetaUpdates
- [x] overrideWidgetProperties

3. Verification for user's application which faced this issue.

- [x] Aforce : verified by @ramsaptami 
- [x] Prisk Admin : verified by @Rishabh-Rathod 
- [ ] Datar ai OEM ( Issue was not reproducible here )
- [x] Datar ai check in/check out.

4. For performance degradation test

We will use A force Application community issues page events

| Test Case | Edit mode   ||  View Mode   ||
|---|---|---|---|---|
|    |  PR's DP  |  Release  | PR's DP | Release |
| Table, Row switch 5 times | Screenshot 2022-05-23 at 2 54 44 PM | Screenshot 2022-05-23 at 2 54 52 PM | Screenshot 2022-05-23 at 2 58 32 PM | Screenshot 2022-05-23 at 2 58 39 PM |

## Q&A

1. Why `metaHOC` has `leading: true`?
https://github.com/appsmithorg/appsmith/pull/13084/files#r874388200 

## Checklist:

    // Code coverage diff between base branch:release and head branch: fix/input-delay 
 :green_circle: | total | 56.65 **(0.03)** | 38.65 **(0.03)** | 36.04 **(0.1)** | 56.91 **(0.04)**
 :green_circle: | app/client/src/actions/metaActions.ts | 90 **(11.43)** | 100 **(0)** | 66.67 **(41.67)** | 85.71 **(15.71)**
 :green_circle: | app/client/src/components/editorComponents/EditorContextProvider.tsx | 93.33 **(0.47)** | 100 **(0)** | 75 **(8.33)** | 93.33 **(0.47)**
 :green_circle: | app/client/src/entities/DataTree/dataTreeWidget.ts | 98.57 **(0.02)** | 83.33 **(0)** | 100 **(0)** | 98.36 **(0.03)**
 :green_circle: | app/client/src/entities/Widget/utils.ts | 91.96 **(0)** | 91.3 **(2.17)** | 100 **(0)** | 91.89 **(0)**
 :sparkles: :new: | **app/client/src/reducers/entityReducers/metaReducer/index.ts** | **75** | **25** | **75** | **72.97**
 :green_circle: | app/client/src/reducers/evaluationReducers/treeReducer.ts | 40 **(3.64)** | 0 **(0)** | 50 **(0)** | 47.06 **(0)**
 :red_circle: | app/client/src/sagas/EvaluationsSaga.ts | 19.23 **(-0.17)** | 1.9 **(-0.04)** | 7.41 **(0)** | 21.95 **(-0.22)**
 :green_circle: | app/client/src/sagas/selectors.tsx | 79 **(1)** | 58.33 **(0)** | 66.67 **(4.17)** | 75.32 **(1.64)**
 :green_circle: | app/client/src/utils/WidgetFactory.tsx | 79.1 **(13.43)** | 64.29 **(21.43)** | 72.73 **(27.28)** | 79.1 **(13.43)**
 :red_circle: | app/client/src/widgets/BaseWidget.tsx | 79.23 **(-0.61)** | 46.94 **(-2)** | 76.47 **(0)** | 79.03 **(0)**
 :green_circle: | app/client/src/widgets/MetaHOC.tsx | 86.15 **(1.77)** | 57.89 **(4.04)** | 86.67 **(-5.64)** | 90 **(6.95)**
 :red_circle: | app/client/src/widgets/WidgetUtils.ts | 82.81 **(0)** | 61.74 **(-0.87)** | 66.67 **(0)** | 79.11 **(0)**
 :red_circle: | app/client/src/widgets/BaseInputWidget/component/index.tsx | 65.68 **(0)** | 40 **(-0.67)** | 60.87 **(0)** | 62.99 **(0)**
 :green_circle: | app/client/src/widgets/JSONFormWidget/widget/index.tsx | 27.12 **(0.23)** | 0 **(0)** | 20 **(0)** | 27.43 **(0.24)**
 :red_circle: | app/client/src/widgets/TabsWidget/component/PageTabs.tsx | 95.74 **(-2.13)** | 70 **(-15)** | 100 **(0)** | 95.45 **(-2.28)**
 :red_circle: | app/client/src/widgets/TabsWidget/widget/index.tsx | 64.29 **(-14.28)** | 43.06 **(-4.08)** | 78.95 **(0)** | 64.44 **(-13.34)**
 :red_circle: | app/client/src/workers/evaluation.worker.ts | 7.95 **(-0.16)** | 0 **(0)** | 33.33 **(0)** | 8.45 **(-0.18)**
 :green_circle: | app/client/src/workers/evaluationUtils.ts | 64.17 **(0.33)** | 62.93 **(0.74)** | 68.75 **(0)** | 62.37 **(0.4)**
 :green_circle: | app/client/src/workers/DataTreeEvaluator/DataTreeEvaluator.ts | 80.79 **(0.17)** | 59.94 **(0)** | 78.65 **(0)** | 80.66 **(0.07)**
 :x: | ~~app/client/src/reducers/entityReducers/metaReducer.ts~~ | ~~33.33~~ | ~~0~~ | ~~20~~ | ~~35~~</details>


Rishabh-Rathod:
Screenshot 2022-04-24 at 1 45 58 AM

Even after current changes truncated values seem to be still present which means for a few cases widget fails to pass the latest value to the worker via unEvalTree.

Rishabh-Rathod:
Screenshot 2022-04-25 at 11 47 46 AM

Root cause of truncated issue, seems to be related to `inputText`

Rishabh-Rathod:
NOTE - https://github.com/appsmithorg/appsmith/pull/13415#discussion_r861802377

Rishabh-Rathod:
We added a wait time to check if the test passes. Need to add wait time does give a clear signal of performance degradation, we will need to analyze performance in AForceApplication.

Rishabh-Rathod:
Add jest tests for
- [ ] metaReducer
- [ ] evalMetaUpdates
- [ ] metaHOC

app/client/src/widgets/TabsWidget/widget/index.tsx
```diff
@@ -257,10 +256,13 @@ class TabsWidget extends BaseWidget<
   }
 
   componentDidUpdate(prevProps: TabsWidgetProps<TabContainerWidgetProps>) {
-    const visibleTabs = this.getVisibleTabs();
-    if (this.props.defaultTab) {
-      if (this.props.defaultTab !== prevProps.defaultTab) {
-        const selectedTab = _.find(visibleTabs, {
+    if (!isEqual(prevProps, this.props)) {
```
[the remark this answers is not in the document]
Rishabh-Rathod: This is fixed, there was an issue with the new mapStateToProps in `metaHOC` which is fixed.

Rishabh-Rathod (commented):
Noticed copying, pasting, delete shortcuts on canvas for widgets not working.

app/client/src/widgets/MetaHOC.tsx
```diff
@@ -121,99 +113,65 @@ const withMeta = (WrappedWidget: typeof BaseWidget) => {
...
+      this.debouncedTriggerEvalOnMetaUpdate();
+    };
 
     updatedProps = () => {
       return {
-        ...this.props,
-        ...this.state,
-        updateWidgetMetaProperty: this.updateWidgetMetaProperty,
-        syncUpdateWidgetMetaProperty: this.syncUpdateWidgetMetaProperty,
+        ...this.initialMetaState, // this contains stale default values and are used when widget is reset. Ideally, widget should reset to its default values instead of stale default values.
```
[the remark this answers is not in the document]
Rishabh-Rathod: https://github.com/appsmithorg/appsmith/issues/13540

Rishabh-Rathod:
@SatishGandham @sbalaji1192 @hetunandu 

Jest tests for the changes are added.
It is ready for review.

app/client/cypress/integration/Smoke_TestSuite/Application/CommunityIssues_Spec.ts
```diff
@@ -264,11 +264,11 @@ describe("AForce - Community Issues page validations", function () {
 
   })
 
-  it("9. Validate Updating issue from Details tab", () => {
-
+  it("9. Validate Updating issue from Details tab & #13588", () => {
```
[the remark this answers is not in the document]
Rishabh-Rathod: I will add a comment explaining information about issue #13588, 
I was asked to add the issue number in the test description by @Aishwarya-U-R.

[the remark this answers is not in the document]
hetunandu: I feel that issue numbers are hard to make sense directly and always has to look them up. A better description is standalone and does not need any context of an older bug. Ideally, issue number/link should be in the comments to add extra context over the already present description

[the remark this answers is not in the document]
Aishwarya-U-R: I suggested to add the issue number, since we will know in future that a script exists for such issue when we pick it up for bugs automation & to avoid any duplication of scripts. We can describe it all the way we want - but please do not remove the issue #. We need it for reference.

app/client/src/reducers/evaluationReducers/treeReducer.ts
```diff
@@ -20,9 +20,9 @@ const evaluatedTreeReducer = createImmerReducer(initialState, {
       removedPaths: [string];
     }>,
   ) => {
-    const { dataTree, updates } = action.payload;
+    const { updates } = action.payload;
     if (updates.length === 0) {
-      return dataTree;
```
Rishabh-Rathod, line 25: NOTE: It was noticed that when we returned `dataTree` from the worker as state, the reference of objects inside the widget got changed and shallow comparison failed. 

Why does it happen only for this PR?
1. Need to analyze what happens on release, when we have `updates` as empty 
2. If updates are not empty on release then that might explain this behavior but then the question arises why do we see empty `updates` only for this PR?

cc: @SatishGandham

Rishabh-Rathod: `updates` i.e., `const updates = diff(oldDataTree, dataTree) || [];` are not empty on release branch which means.
We never set treeReducer `state = dataTree` and hence this behavior only was observed on this PR.

Next Steps:-
Why do we run 2 evaluations in this PR for a single meta change? 
Observation - The second evaluation results in empty `updates`.

Rishabh-Rathod: Observation - 
`triggerEvalOnMetaUpdate` was called twice for updating inside Input widget for updateWidgetMetaProperty `dragDisabled` and `isFocused`

Rishabh-Rathod: `metaHOC` debounce was running for both leading and trailing ends which was incorrect, it should be running only for trailing end.

Solved with latest commit.

Rishabh-Rathod:
ok-to-test sha=34f203d

SatishGandham (dismissed):
Looks good to me except the treeReducer change. Please double check.
Because of the nature of the  change, please get a second review and have it tested by QA.

ashit-rath (dismissed):
JSONForm changes LGTM!

sbalaji1192:
> @SatishGandham @sbalaji1192
> 
> Mentioning a use-case where the `leading: true` is useful in the MetaHOC debounce function.
> 
> 1. Add an input widget and make it required.
> 2. Type random characters very fast, on release you would notice that the Input widget becomes valid but in this PR until the user doesn’t stop typing it stays invalid.
> 
> Ideally, In such a situation where the widget knows that it needs to trigger an evaluation for a change. I believe we should keep an option in `updateWidgetMetaProperty` to trigger evaluation without debouncing instead of evaluating before and after evaluation.
> 
> Please, let me know what do you think?

@Rishabh-Rathod Do you think this will break some applications? 
What are the widgets that will be triggering the evaluations without debouncing?

app/client/src/widgets/MetaHOC.tsx
```diff
@@ -20,94 +23,87 @@ export interface WithMeta {
...
+  class MetaHOC extends React.PureComponent<metaHOCProps> {
     static contextType = EditorContext;
-    updatedProperties = new Map<string, true>();
-    propertyTriggers = new Map<string, DebouncedExecuteActionPayload>();
-
-    debouncedHandleUpdateWidgetMetaProperty = debounce(
-      this.handleUpdateWidgetMetaProperty.bind(this),
-      200,
-      {
-        leading: true,
```
Rishabh-Rathod, line 41: Create a thread to discuss the need for `leading: true` in some cases and performance impact.

Rishabh-Rathod: > @Rishabh-Rathod Do you think this will break some applications?
What are the widgets that will be triggering the evaluations without debouncing?

@sbalaji1192
The intention is to not have any breaking changes. Currently, on release evaluation runs for both the leading and trailing end of debouncing. 
- This makes sure to run validation for the first change made. 
for example, 

| Event | Release | DP |
|---|---|---|
| if we typed "test" in the Input widget | For the first character `t` evaluation will run and the widget will get isValid as true | After the user stops typing i.e., input.text = `test` then we run the evaluation so while the user was typing `test` they will see input widget value as invalid in UI until they stop typing |

The difference is mainly in the timing of running evaluation here and for the end-users difference is they will see value validation running only after they stop typing. ( A very small but noticeable change when typing very fast )

Now, we have 2 approaches to solve this IMO.
1. Run evaluation on both leading and trailing end and let it work how it used to work. ( I will be trying this 1st )
2. As the widget is aware that when the value is changed from empty to any value or from any value to empty, validation should run sooner to give users faster feedback of validation on their values. 
NOTE: If the user has configured the `Valid` field then the first evaluation run is not helpful. ( source of truth is the last evaluation run in both the case, first evaluation run just runs the evaluation of the first value change and that might make change the `isValid` property of the widget, and accordingly, the widget shows UI )

cc: @SatishGandham

Rishabh-Rathod: @hetunandu  Can you share knowledge on why we had `leading: true` in debounce function of MetaHOC?

sbalaji1192: @Rishabh-Rathod 
1. The original issue happened because we were running the evaluations on leading and trailing, right? will Introducing the same behavior again would lead to a similar issue?
2. So the only widget that gets changed in quick succession is the Input widget (Correct me if I'm wrong). I guess we can go with `leading : true` only for the input widget (as you mentioned through the `updateWidgetMetaProperty `), because a debounced update to the isValid will get perceived by the user as a performance issue.

app/client/src/reducers/evaluationReducers/treeReducer.ts
```diff
@@ -20,9 +20,9 @@ const evaluatedTreeReducer = createImmerReducer(initialState, {
...
+      return state;
```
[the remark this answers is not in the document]
Rishabh-Rathod: If we return the `dataTree` here all the object references will change and it will lead to all the widget's re-rendering.

app/client/src/widgets/MetaHOC.tsx
Rishabh-Rathod: I missed a part from earlier implementation that setState's callback was helping in batching `updateWidgetMetaProperty`. 

When widget called 
```
this.props.updateWidgetMetaProperty("dragDisabled", focusState);
this.props.updateWidgetMetaProperty("isFocused", focusState);
``` 

`setState` was batching those updates and then it ran the callback function only once. 

Now, I added 
1. `leading: true` in denounced function to solve the validation problem
2.  `setState callback` for batch updates. ( Avoiding to run `debouncedTriggerEvalOnMetaUpdate` for each updateWidgetMetaProperty )
3. `updatedProperties` map to always check if there are meta updates that need to trigger evaluation to push to the worker.

Rishabh-Rathod: > The original issue happened because we were running the evaluations on leading and trailing, right? will Introducing the same behavior again would lead to a similar issue?

No, the problem was not that we were running evaluation on `leading` and `trailing` but instead the values in metaReducer were not updated while we were running the evaluation by a trigger like onSubmit ( no debounce ).

This was about the problem of the delayed value.

sbalaji1192: @Rishabh-Rathod I was referring to the issue where the last characters of the input widgets `text` get truncated.

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/Binding/TextTable_spec.js
```diff
@@ -35,6 +35,7 @@ describe("Text-Table Binding Functionality", function() {
       cy.EvaluateDataType("string");
       cy.validateEvaluatedValue(tabValue);
       cy.PublishtheApp();
+      cy.wait(500); // table selection shouldn't get clear.
```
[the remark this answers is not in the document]
Rishabh-Rathod: Hey @SatishGandham, 
These changes were undone. For the initial implementation, there were failures in the cypress test and this change was done as a temporary patch fix.

app/client/src/widgets/MetaHOC.tsx
Rishabh-Rathod: @SatishGandham No it shouldn't bring the performance regression back. 
I have updated the screenshots in the performance degradation test. Those numbers seem fine to me.

app/client/src/workers/evaluation.worker.ts
```diff
@@ -103,6 +104,7 @@ ctx.addEventListener(
         let evaluationOrder: string[] = [];
         let unEvalUpdates: DataTreeDiff[] = [];
         let jsUpdates: Record<string, any> = {};
+        let evalMetaUpdates: EvalMetaUpdates = [];
 
         try {
           if (!dataTreeEvaluator) {
```
eco-monk, line 110: Is my assumption correct?
We don't bother about meta updates here, because this is the absolute firstTree where initialMeta values are going to takeover.

eco-monk (commented):
Worker changes look good to me.
We're sending a list of `overriddenMetaProperties` to `metaReducer`.
Perviously, the `diffs` were sent and the `metaReducer` had to figure out the actual meta changes from that.

app/client/src/workers/evaluation.worker.ts
Rishabh-Rathod: > Previously, the diffs were sent and the metaReducer had to figure out the actual meta changes from that.

The previous diff was a comparison of old datatree with new dataTree, when the user changes the meta values of a widget, those updates are not supposed to be present inside evalMetaUpdates.
1. We only want to change `metaReducer` when we change the meta value inside the evaluation because of the default value change. (re-evaluation)

Rishabh-Rathod: > Is my assumption correct? We don't bother about meta updates here, because this is the absolute firstTree where initialMeta values are going to takeover.

Actually, we could have added `evalMetaUpdates` for the createFirstTree also but the reasons why I didn’t add were
1. It could lead to unnecessary re-render.  
let's take an example of an input widget,
In createFirstTree, all the properties will get evaluated which means `defaultText` will also get evaluated, let's say `defaultText = "ABC"`. When we evaluate `defaultText` it will also make `text = "ABC"`.
 
Widgets will read this `text` from the dataTree received as props to metaHOC. 

https://github.com/appsmithorg/appsmith/blob/6f8afb250a0d7a3120f833e6052ba82f81d2ca29/app/client/src/widgets/MetaHOC.tsx#L175-L181

If we send `evalMetaUpdates` for createFirstTree then widgets will re-render, due to update in `metaState` which in this case is not needed. 

`metaState` is useful when we want to override the evaluated values on user change.

2. We only define meta value when the widget calls `updateWidgetMetaProperty`

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/Binding/TextTable_spec.js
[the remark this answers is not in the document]
Rishabh-Rathod: This was removed in later commits.

SatishGandham (dismissed):

app/client/src/entities/DataTree/dataTreeWidget.ts
```diff
@@ -204,8 +208,14 @@ export const generateDataTreeWidget = (
   });
 
   const meta = _.merge(overridingMetaProps, widgetMetaProps);
-
-  Object.entries(widgetMetaProps).forEach(([key, value]) => {
+  const mergedProperties = _.merge(
+    defaultMetaProps,
+    widgetMetaProps,
+    derivedProps,
```
Rishabh-Rathod, line 214: @SatishGandham 
We should retain the sequence of merges to avoid the unexpected value of the derived property if the same property name is also a meta property.

- Why these changes are done in the same PR?
jest test failed due to this as in one of the tests, a derived property `selectedRow` was directly assigned value. To unblock this PR and retain the earlier logic made the changes here.

Rishabh-Rathod:
### RELEASE 
Screenshot 2022-05-23 at 9 18 17 AM
### PR
Screenshot 2022-05-23 at 9 26 08 AM

app/client/src/entities/DataTree/dataTreeWidget.ts
```diff
@@ -191,21 +195,30 @@ export const generateDataTreeWidget = (
   const {
     dataTreeWidgetWithoutMetaProps: dataTreeWidget,
     defaultMetaProps,
+    derivedProps,
     overridingMetaPropsMap,
   } = generateDataTreeWidgetWithoutMetaMemoized(widget);
   const overridingMetaProps: Record<string, unknown> = {};
-
+  // cloning default and derived props to avoid mutation
+  const clonedDefaultMetaProps = klona(defaultMetaProps);
```
SatishGandham, line 203: @Rishabh-Rathod , Check if merging is faster compared to clone and spread.

`merge({},defaultMetaProps,widgetMetaProps,derivedProps)`

Rishabh-Rathod: Though the performance was very close, when tried adding a table widget in a few cases merge was faster.
There were cases when clone and spread were faster but the occurrence of such cases was less.

Hence, going forward with `_.merge`

SatishGandham (approved):

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/FormWidgets/AllInputWidgets_spec.js
```diff
@@ -0,0 +1,216 @@
...
+
+    it("2. StoreValue should have complete input value", () => {
+      if (widgetSelector === WIDGET.INPUT_WIDGET_V2) {
+        cy.openPropertyPane(widgetSelector);
+        cy.selectDropdownValue(".t--property-control-datatype", "Text");
+        cy.get(".t--property-control-required label")
+          .last()
+          .click({ force: true });
+        cy.closePropertyPane();
+      }
```
Aishwarya-U-R, line 130: Since Input widget is already defaulted to 'Text' - lets comment this code for now!

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/FormWidgets/AllInputWidgets_spec.js
```diff
@@ -0,0 +1,216 @@
...
+        `{{storeValue('textPayloadOnSubmit',${testConfig.widgetPrefixName}1.text)}}`,
+      );
+      // Bind to stored value above
+      cy.openPropertyPane(WIDGET.TEXT);
+      cy.updateCodeInput(
+        PROPERTY_SELECTOR.text,
+        "{{appsmith.store.textPayloadOnSubmit}}",
+      );
+      cy.closePropertyPane();
+      cy.get(getWidgetInputSelector(widgetSelector)).clear();
```
Aishwarya-U-R, line 148: Cypress .clear errors out when no value is present to clear. To check this with a run!

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/FormWidgets/AllInputWidgets_spec.js
```diff
@@ -0,0 +1,216 @@
...
+        cy.get(getWidgetSelector(WIDGET.TEXT)).should("have.text", expected);
+      });
+    });
+
+    it("3. Api params getting correct input values", () => {
+      // Set onClick action, storing value
+      cy.openPropertyPane(WIDGET.BUTTON_WIDGET);
+      cy.get(PROPERTY_SELECTOR.onClick)
+        .find(".t--js-toggle")
+        .click();
```
Aishwarya-U-R, line 175: Button Widget is already toggled on! So this line should ideally toggle it off, to check with a run!

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/FormWidgets/AllInputWidgets_spec.js
```diff
@@ -0,0 +1,216 @@
...
+          expect(
+            interception.response.body.data.request.headers.value,
+          ).to.deep.equal([expected]);
+        });
+      });
+    });
+
+    it("4. Delete all the widgets on canvas", () => {
+      cy.get(getWidgetSelector(WIDGET.BUTTON_WIDGET)).click();
+      cy.get("body").type(`{del}`, { force: true });
```
Aishwarya-U-R, line 207: To change to widget delete - for increasing test scope!

Aishwarya-U-R (commented):
Initial review done!

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/FormWidgets/AllInputWidgets_spec.js
Rishabh-Rathod: done

SatishGandham (approved):

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/FormWidgets/AllInputWidgets_spec.js
Rishabh-Rathod: It was decided that @Aishwarya-U-R will pick it up separately. 
As this is not related to current PR and is focused on adding a different way of deletion covered in the automation test.

app/client/src/widgets/TabsWidget/widget/index.tsx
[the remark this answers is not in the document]
ashit-rath: @Rishabh-Rathod Why are we doing deep equal on the whole props? Isn't that going to be expensive? 
I think we only need to compare few props i.e `selectedTabWidgetId`, `defaultTab` and probably the `tabsObj` as this block seems to only deal with the correction of the `selectedTab` for a few cases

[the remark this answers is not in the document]
Rishabh-Rathod: I should have removed it back then seems like a miss by me. ( this comment was meant for reminding me 😞 )
> I think we only need to compare few props i.e selectedTabWidgetId, defaultTab and probably the tabsObj as this block seems to only deal with the correction of the selectedTab for a few cases

I wasn't aware of all the props which we need to track. As it seems we are aware of it in that case we can fix this here https://github.com/appsmithorg/appsmith/pull/15202 

I will be more vigilant from next time. Thanks for pointing it out. :)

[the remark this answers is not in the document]
ashit-rath: Got it, thanks for the explanation @Rishabh-Rathod.

PR #13126, opened 2022-04-20
fix: Page order sequence for git import

### LINKED ISSUE #12944: [Bug]: When we do git import the application does not have the same page sequence as in original git application.

### Description

When we do git import the application does not have the same page sequence as in original git application.
image

image

### Steps To Reproduce

1. Create new application
2. Add pages
3. Connect app to git
4. Enow import git connect app from git import
5. check the page index it will be different from original app

### Public Sample App

_No response_

### Version

release

AnaghHegde:
## Description

> Page order sequence gets messed up when we do git import. The reason is because of the parallel execution of task by stream API, which is used to process page from the application json file. We don't have control over this. So we have decided to add the page order to the metadata of json file which is exported. This list is then referred while importing application into the org. 

Fixes #12944 

## Type of change

## How Has This Been Tested?

- Locally 
- Junit Tests

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/domains/ApplicationJson.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/domains/ApplicationJson.java
@@ -35,6 +35,8 @@ public class ApplicationJson {
 
     List<NewPage> pageList;
 
+    List<String> pageOrder;
+
     String publishedDefaultPageName;
     
     String unpublishedDefaultPageName;
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/domains/ApplicationJson.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/domains/ApplicationJson.java
@@ -37,6 +37,8 @@ public class ApplicationJson {
 
     List<String> pageOrder;
 
+    List<String> publishedPageOrder;
+
     String publishedDefaultPageName;
     
     String unpublishedDefaultPageName;
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
@@ -77,6 +77,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -861,7 +862,7 @@ public class ImportExportApplicationServiceCEImpl implements ImportExportApplica
                             .map(newPageList -> {
                                 // Check if the pages order match with json file
                                 List<String> pageOrderList;
-                                if(applicationJson.getPageOrder().isEmpty()) {
+                                if(Optional.ofNullable(applicationJson.getPageOrder()).isEmpty()) {
                                     pageOrderList = importedNewPageList.stream().map(newPage -> newPage.getUnpublishedPage().getName()).collect(Collectors.toList());
                                 } else {
                                     pageOrderList = applicationJson.getPageOrder();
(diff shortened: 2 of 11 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/domains/ApplicationJson.java
```diff
@@ -35,6 +35,8 @@ public class ApplicationJson {
 
     List<NewPage> pageList;
 
+    List<String> pageOrder;
```
abhvsn, line 38: We should have a seperate list for unpublished and published pages as the sequence might be different

AnaghHegde: @abhvsn IIUC, it is not possible right? Say you have Page1 at top in edit mode and in view mode it is at bottom. Now while you are creating a new App, the same page will have two different order and we have to ignore one of these two. And IMO, edit mode takes precedence over view mode.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -244,9 +244,14 @@ public Mono<ApplicationJson> exportApplicationById(String applicationId, Seriali
 
                     return pageFlux
```
abhvsn, line 245: This again result in mismatched order can we instead take the reference from `pages` and `publishedPages` from the application object that will be easier to deal with. Also we are referencing the pages from application object only

AnaghHegde: It does not matter now as we are storing the page order as a metadata, which is used in import flow. So no matter how the pages are stored in json-file/git-repo, the order is maintained as we keep update the order in metadata during the export flow here - https://github.com/appsmithorg/appsmith/blob/4a704d415f7abfe484114ddf2fb2c696f6d13345/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java#L254.

abhvsn (commented):
Please take a look at the case mentioned in the comment also can we add the testcase with the mentioned scenario

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -253,6 +254,13 @@ public Mono<ApplicationJson> exportApplicationById(String applicationId, Seriali
                                     pageOrder.add(page.getUnpublishedPage().getName());
                                 }
                                 applicationJson.setPageOrder(pageOrder);
+
+                                Collections.sort(newPages, Comparator.comparing(newPage -> publishedPageOrderList.indexOf(newPage.getId())));
```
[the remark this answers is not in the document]
AnaghHegde: @abhvsn This is handled correctly during the import. But when we add the page order to json file during export, i have missed out on this case. Let me fix this.

app/server/appsmith-server/src/test/java/com/appsmith/server/solutions/ImportExportApplicationServiceTests.java
```diff
@@ -2359,22 +2359,30 @@ public void exportAndImportApplication_withMultiplePages_PagesOrderIsMaintained(
...
-                    List<NewPage> pageList = applicationJson.getPageList();
-                    assertThat(pageList.get(0).getUnpublishedPage().getName()).isEqualTo("123");
-                    assertThat(pageList.get(1).getUnpublishedPage().getName()).isEqualTo("abc");
-                    assertThat(pageList.get(2).getUnpublishedPage().getName()).isEqualTo("Page1");
+                    List<String> pageList = applicationJson.getPageOrder();
+                    assertThat(pageList.get(0)).isEqualTo("123");
+                    assertThat(pageList.get(1)).isEqualTo("abc");
+                    assertThat(pageList.get(2)).isEqualTo("Page1");
+
+                    List<String> publishedPageList = applicationJson.getPageOrder();
```
abhvsn, line 2375:
```suggestion
                    List<String> publishedPageList = applicationJson.publishedPageOrder();
```

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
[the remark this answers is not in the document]
abhvsn: Now getting the NPE during the export

abhvsn (changes requested):
Now I'm getting the NPE for the for the scenario described https://github.com/appsmithorg/appsmith/pull/13126/files#r857259455 at 
`pageOrder.add(page.getPublishedPage().getName());`

abhvsn (approved):

PR #13154, opened 2022-04-20
feat: added healthcheck to docker

sum35h:
## Description
Added **healthcheck** to the fat dockerfile. The `healthcheck.sh` script used `supervisorctl` to check for the status of all the process to determine the health.
Added healthcheck to thin server and editor docker file.

## Type of change

## How Has This Been Tested?
Manually

    // Code coverage diff between base branch:release and head branch: feat/docker-healthcheck 
 :green_circle: | total | 56.47 **(0.01)** | 37.92 **(0.01)** | 36.15 **(0)** | 56.69 **(0.01)**
 :green_circle: | app/client/src/utils/WorkerUtil.ts | 89.76 **(0.78)** | 70.59 **(0)** | 100 **(0)** | 93.33 **(0.95)**</details>

### DIFF
--- a/app/client/Dockerfile
+++ b/app/client/Dockerfile
@@ -16,4 +16,5 @@ COPY ./docker/templates/nginx-app-https.conf.template /nginx-app-https.conf.temp
 
 # This is the script that is used to start Nginx when the Docker container starts
 COPY ./docker/start-nginx.sh /start-nginx.sh
+HEALTHCHECK --interval=15s --timeout=15s --start-period=15s --retries=3 CMD curl -f http://localhost:80/ || exit 1
 CMD ["/start-nginx.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -86,7 +86,7 @@ COPY ./deploy/docker/templates/supervisord/ templates/supervisord/
 COPY ./deploy/docker/templates/cron.d /etc/cron.d/
 RUN chmod 0644 /etc/cron.d/*
 
-RUN chmod +x entrypoint.sh renew-certificate.sh
+RUN chmod +x entrypoint.sh renew-certificate.sh healthcheck.sh
 
 # Update path to load appsmith utils tool as default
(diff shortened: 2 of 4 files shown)


app/client/Dockerfile
```diff
@@ -16,4 +16,5 @@ COPY ./docker/templates/nginx-app-https.conf.template /nginx-app-https.conf.temp
 
 # This is the script that is used to start Nginx when the Docker container starts
 COPY ./docker/start-nginx.sh /start-nginx.sh
+HEALTHCHECK --interval=15s --timeout=15s --start-period=15s --retries=3 CMD curl -f http://localhost:80/ || exit 1
```
sharat87, line 19: What's the reason for `|| exit 1`? If the curl command failed, it would anyway exit with a non-zero status code right?

sharat87 (approved):

renatocron:
Need to add some way to check if the HTTPS was already configured, or tho test the editor directly
After setting up the letscencrypt/https, I changed the healthcheck script to use port 443:

``` 
if [[ $(curl -k -s -w "%{http_code}\n" https://localhost:443/ -o /dev/null) -ne 200 ]]; then
   echo 'ERROR: Editor is down';
   healthy=false
fi
```

sharat87:
@renatocron, the latest version of that check actually does take https into account, by adding the `-L` argument to `curl`. With this, `curl` will just accept the redirect response from the http endpoint and test the https endpoint as well. See https://github.com/appsmithorg/appsmith/pull/13947/files#diff-359045826464e68364521558edab24d3a9b3de87bdbeeec9cedb1869e0560586.

Also, in general, the `:443` is not needed when using `https://` in the URL. This is because 443 is the default port of https URLs.

renatocron:
@sharat87 thank you, I should have check the file was modified 5 days ago before commenting here!

The https://github.com/appsmithorg/appsmith/blob/release/deploy/docker/scripts/healthcheck.sh indeed does work!

PR #13155, opened 2022-04-20
chore: Added jcmd to Appsmith server image

### LINKED ISSUE #12959: [Task] : Instrument server heap monitoring and create baseline for tracking memory consumption

nidhi-nair:
Fixes #12959

### DIFF
--- a/app/server/Dockerfile
+++ b/app/server/Dockerfile
@@ -1,6 +1,12 @@
 #When you are building, name it appsmith-server which is how it is referenced in docker-compose.yml
 
+FROM adoptopenjdk/openjdk11:jdk-11.0.10_9-alpine
+
+RUN ${JAVA_HOME}/bin/jlink --module-path jmods --add-modules jdk.jcmd --output /jcmd
+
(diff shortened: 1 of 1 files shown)


app/server/Dockerfile
```diff
@@ -1,7 +1,13 @@
 #When you are building, name it appsmith-server which is how it is referenced in docker-compose.yml
 
+FROM adoptopenjdk/openjdk11:jdk-11.0.10_9-alpine
```
sharat87, line 3: Do we need two `FROM` lines? I think that creates a multi-stage build, but since both the `FROM` lines are based on the same image, we may not need it?

nidhi-nair: They're different images no? I'm picking the jcmd binary out of the jdk and using it with the jre image

sharat87: Got it, my bad. Thanks.

sharat87 (approved):

nidhi-nair:
Merging without tests since it is a docker related change.

PR #13159, opened 2022-04-20
feat: Track recently forked templates

### LINKED ISSUE #12339: Recently used templates | BE
The templates a user have recently imported. “Recently” is described as last five templates.

nayan-rafiq:
## Description

This PR adds feature to track recently forked templates by the user. 

Fixes #12339

## Type of change

## How Has This Been Tested?

- JUnit tests

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationTemplateControllerCE.java
```diff
@@ -24,7 +24,7 @@ public ApplicationTemplateControllerCE(ApplicationTemplateService applicationTem
 
     @GetMapping
     public Mono<ResponseDTO<List<ApplicationTemplate>>> getAll() {
-        return applicationTemplateService.getActiveTemplates().collectList()
+        return applicationTemplateService.getActiveTemplates(null).collectList()
                 .map(templates -> new ResponseDTO<>(HttpStatus.OK.value(), templates, null));
```
AnaghHegde, line 28: Why are we setting the value to null here ?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserDataServiceCEImpl.java
```diff
@@ -264,6 +264,17 @@ public Mono<UserData> updateLastUsedAppAndOrgList(Application application) {
         });
     }
 
+    @Override
+    public Mono<UserData> addTemplateIdToLastUsedList(String templateId) {
+        return this.getForCurrentUser().flatMap(userData -> {
+            // set recently used template ids
+            userData.setRecentlyUsedTemplateIds(
+                    addIdToRecentList(userData.getRecentlyUsedTemplateIds(), templateId, 5)
```
AnaghHegde, line 272: What does the number 5 represents here?

AnaghHegde (changes requested):
One last thing, i noticed that we are adding the template ids to the userData object. Suppose if i use a same template multiples times, will the same id is added to the userData object multiple time? If yes, then can we keep the count or reorder the list of ids so that last used is always present at the top ?

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationTemplateControllerCE.java
nayan-rafiq: `getActiveTemplates()` takes an optional list of id. If present, they'll be used to fetch only the provided templates. This method is called in two places - `get all templates` and `get recent templates`. We're providing a list of id for the later one. When getting all templates, this parameter is not required and hence sending null.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserDataServiceCEImpl.java
nayan-rafiq: We want to keep track of the last 5 recently used templates.

nayan-rafiq:
> One last thing, i noticed that we are adding the template ids to the userData object. Suppose if i use a same template multiples times, will the same id is added to the userData object multiple time? If yes, then can we keep the count or reorder the list of ids so that last used is always present at the top ?

Good point. It actually does this. The algorithm to add a new template id to the list is: 

- Prepend template id to the list
- Remove duplicates from the list
- Truncate the list to 5 if number of records are more than 5

The function that does this already exists and we use the same method to track recently used applications and organizations.

PR #13200, opened 2022-04-21
fix: Prevent sync functions from running on pageload

### LINKED ISSUE #13197: [Bug]: Async functions turned into sync code is still executed on page load even though the option to do so is unavailable

### Description

When a function containing async code is made into sync function - then any on page action that was previously set is retained and the sync is run on page load

### Steps To Reproduce

1. Add a JS object and mark an async function to run on page load
2. Remove async code in the function and refresh the page - observe that the previously marked async function is run again

[LOOM DEMO](https://www.loom.com/share/a74a18aa5c5940b3a51cc9f5eb4eb05b)

### Public Sample App

_No response_

### Version

Cloud

ohansFavour:
## Description

When the async keyword is removed from an async function originally set to run on page load , the corresponding sync function runs onPageLoad. 

Solution

`executeOnLoad` should always be false for a sync function.

Fixes #13197 

## Type of change

## How Has This Been Tested?

Jest

## Checklist:

    // Code coverage diff between base branch:release and head branch: fix/sync-function-running-on-pageload 
 :green_circle: | total | 56.56 **(0.11)** | 38.01 **(0.1)** | 36.23 **(0.06)** | 56.78 **(0.1)**
 :green_circle: | app/client/src/utils/JSPaneUtils.tsx | 84.88 **(74.41)** | 79.07 **(79.07)** | 80 **(80)** | 84.06 **(75.36)**
 :green_circle: | app/client/src/utils/WorkerUtil.ts | 89.76 **(0.78)** | 70.59 **(1.96)** | 100 **(0)** | 93.33 **(0.95)**</details>

### DIFF
--- a/app/client/src/utils/JSPaneUtils.tsx
+++ b/app/client/src/utils/JSPaneUtils.tsx
@@ -156,15 +156,6 @@ export const getDifferenceInJSCollection = (
   } else {
     changedVariables = jsAction.variables;
   }
-  console.log(parsedBody, {
-    yes: {
-      newActions: toBeAddedActions,
-      updateActions: toBeUpdatedActions,
-      deletedActions: toBearchivedActions,
-      nameChangedActions: nameChangedActions,
-      changedVariables: changedVariables,
-    },
-  });
   return {
     newActions: toBeAddedActions,
     updateActions: toBeUpdatedActions,
--- a/app/client/src/utils/JSPaneUtils.tsx
+++ b/app/client/src/utils/JSPaneUtils.tsx
@@ -156,6 +156,15 @@ export const getDifferenceInJSCollection = (
   } else {
     changedVariables = jsAction.variables;
   }
+  console.log(parsedBody, {
+    yes: {
+      newActions: toBeAddedActions,
+      updateActions: toBeUpdatedActions,
+      deletedActions: toBearchivedActions,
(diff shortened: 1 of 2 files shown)


app/client/src/utils/JSPaneUtils.tsx
```diff
@@ -51,6 +51,7 @@ export const getDifferenceInJSCollection = (
               jsArguments: action.arguments,
               isAsync: action.isAsync,
             },
+            executeOnLoad: action.isAsync ? preExisted.executeOnLoad : false,
```
ApekshaBhosale, line 54: if it is not async it shouldn't be executed on page load. it is still coming in the layout call? because we take functions from there. if you set this, I think this will manually set the value and behavior becomes different. can you talk to @nidhi-nair and confirm this?

ApekshaBhosale (changes requested):

app/client/src/utils/JSPaneUtils.tsx
ohansFavour: @ApekshaBhosale, with this change, in layout calls;  sync functions don't come as part of pageActions (which run on page load). After the modification here, `updateJSCollection` causes update of the executeOnLoad property on the server side.

PR #13208, opened 2022-04-22
fix: Added contains and not equals options to where clause in Google Sheets

### LINKED ISSUE #11920: [Feature] Add operator similar to SQL like for text filtering (Google Sheets)
Currently the GSheets integration has a condition to check if a value contains a column, `in`, but not a condition to check if the column contains the value.

The new conditions could be `like` and `not like` or `contains` and `doesn't contain`.

nidhi-nair:
## Description
This PR introduce the contains and in operator in the old format of Google Sheets. The change also handles a few null scenarios for the values. This change will need to be propagated to the UQI format as well. The operators and handling have been introduced in both places, but the null handling will need to be checked since that piece of code was already missing for the newer filter format.

Fixes #11920 

## Type of change

## How Has This Been Tested?
- Enhanced JUnit test case
- Manually tested

## Checklist:

### DIFF
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/constants/ConditionalOperator.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/constants/ConditionalOperator.java
@@ -73,4 +73,10 @@ public enum ConditionalOperator {
             return "or";
         }
     },
+    CONTAINS {
+        @Override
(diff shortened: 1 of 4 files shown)


leotom2000 (approved):

nidhi-nair:
Merging since all tests have passed but test result failed.

PR #13245, opened 2022-04-24
feat: Apply serverSide Filtering Condition in Select Widget

Tooluloope:
> Pull Request Template
>

## Description

Fixes # (issue)

## Type of change

## How Has This Been Tested?

- Test A
- Test B

## Checklist:


app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -619,19 +622,32 @@ class MultiSelectWidget extends BaseWidget<
...
+      const optionValue = this.props.selectedOptions.map(
+        (option) =>
+          ({
+            label: option.label ?? option,
+            value: option.value ?? option,
+          } as DropdownOption),
+      );
+      value = optionValue;
+    }
+    this.props.updateWidgetMetaProperty("selectedOptions", value);
```
Tooluloope, line 650: @ashit-rath Could you help confirm if this would suffice for the serverside filtering Implementation

ashit-rath:
@Tooluloope Please update the description of this PR with appropriate title

Tooluloope:
> @Tooluloope Please update the description of this PR with appropriate title

Done.

Although this PR meant to be for #13079

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
ashit-rath: So for the server side, we are not bothering about the selected option being present in the options list, we always show whatever is selected but we do bother when server side is disabled i.e if the select option is not present options we completely remove it.
This is what is a bit confusing to me as the options can be populated dynamically irrespective of server side enabled or not. It's just that when server side is enabled the onFilterUpdate event is enabled.
Is this the right way forward from a product point of view?

Tooluloope: IMO, as long as the option changes and it's not serverside filtered, it should be strictly checked

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
```diff
@@ -486,8 +486,8 @@ class MultiSelectWidget extends BaseWidget<
 
   static getDerivedPropertiesMap() {
     return {
-      selectedOptionLabels: `{{this.selectedOptions.map(o => o.label)}}`,
-      selectedOptionValues: `{{this.selectedOptions.map(o => o.value)}}`,
+      selectedOptionLabels: `{{this.selectedOptionValues.map(value => _.find(this.options, {value})?.label)}}`,
+      selectedOptionValues: `{{!this.serverSideFiltering ? this.selectedOptions?.filter(selectedOption => this.options?.some(option => option.value === (selectedOption.value ?? selectedOption))).map(o => o.value ?? o)  : this.selectedOptions.map(o => o.value ?? o)}}`,
```
Tooluloope, line 490: @ashit-rath @wmdev0808 Here was the solution I proposed earlier, Handling the Serverside check in the derived Properties. Here, nothing is changed in place but just consumed.

It feels hacky but then @wmdev0808 You could make it look better.

wmdev0808 (changes requested):
@Tooluloope ,
First, thank you for your trouble.
BTW, isDirty does not work as expected
> Steps to reproduce: 
- Drop a text widget and bind to `isDirty`
- Select an option from the widget, resulting in `isDirty` to `true`
- Change `Default Value`, then the bound property `isDirty` does not change to `false`

app/client/src/widgets/MultiSelectWidgetV2/widget/index.tsx
Tooluloope: @ashit-rath

PR #13252, opened 2022-04-25
feat: Delete theme when application is deleted

### LINKED ISSUE #13251: Archive theme when application is archived
When application is archived, we should archive the related themes as well. Otherwise we'll end up having lots of orphan themes. It'll also make it hard to archive records.

nayan-rafiq:
## Description
This PR archives the related themes when an application is deleted aka archived.

Fixes #13251

## Type of change

## How Has This Been Tested?
- JUnit tests
- Manually

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ThemeServiceCEImpl.java
```diff
@@ -394,4 +394,17 @@ public Mono<Theme> getOrSaveTheme(Theme theme, Application destApplication) {
...
+    /**
+     * This will archive themes related to the provided Application.
+     * It'll delete any theme that was saved for this application. It'll also delete the draft themes for this Application.
+     * @param application Application object
+     * @return Provided Application publisher
+     */
+    @Override
+    public Mono<Application> archiveApplicationThemes(Application application) {
+        return repository.archiveByApplicationId(application.getId())
+                .then(repository.archiveDraftThemesById(application.getEditModeThemeId(), application.getPublishedModeThemeId()))
```
abhvsn, line 407: Little confused, are we saying that there will be more than 1 theme present in the DB for the application though only 1 will be active at a time for which the id will be present in the application object. Is this because we give an available list of themes that can be applied to the application?

abhvsn (approved):
LGTM

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ThemeServiceCEImpl.java
nayan-rafiq: Yes, you got it right. Application themes are of two types - draft theme and saved theme. Draft themes are created when user customizes a theme but haven't saved yet. Draft themes are lost when user chooses another theme. There are two draft themes for each mode. Saved themes are created from draft themes - only difference is they are persisted when a different theme is set to the application. There can be multiple saved themes - but they correspond to an application.

PR #13255, opened 2022-04-25
 fix: Git connected Application forking should use default branch application

### LINKED ISSUE #13185: [Bug]: The application doesn't fork based on the new default branch

### Description

The application doesn't fork based on the new default branch still it forks the master branch 

### Steps To Reproduce

1. Connect the application to the git
2. now create a new branch
3. From git hub change the default branch to the newly created branch b1
4. Now come back to the application sync the branch
5. now master branch will be removed as a default branch
6. Now go to the home page
7. fork the application
8. The forked application has the details of master branch instead of forked branch

### Public Sample App

_No response_

### Version

release

AnaghHegde:
## Description

> For git connected application user can update the default branch. When the user updated the default branch forking service should use the default branched application not the original(appsmith default) application. 

Fixes #13185 

## Type of change

## How Has This Been Tested?

- JUnit Tests
- Locally

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ApplicationForkingServiceCEImpl.java
```diff
@@ -38,7 +39,20 @@ public class ApplicationForkingServiceCEImpl implements ApplicationForkingServic
 
     public Mono<Application> forkApplicationToOrganization(String srcApplicationId, String targetOrganizationId) {
         final Mono<Application> sourceApplicationMono = applicationService.findById(srcApplicationId, AclPermission.READ_APPLICATIONS)
-                .switchIfEmpty(Mono.error(new AppsmithException(AppsmithError.NO_RESOURCE_FOUND, FieldName.APPLICATION, srcApplicationId)));
+                .switchIfEmpty(Mono.error(new AppsmithException(AppsmithError.NO_RESOURCE_FOUND, FieldName.APPLICATION, srcApplicationId)))
+                .flatMap(application -> {
```
abhvsn, line 43: Let's make changes to upstream method as it has branch reference already:
```
public Mono<Application> forkApplicationToOrganization(String srcApplicationId,
                                                           String targetOrganizationId,
                                                           String branchName)
```

abhvsn (commented):
Can you please fix this for other APIs as well:
1. Duplicate
2. Export

AnaghHegde:
@abhvsn We do not fetch the application there. It would be good to add it here right?

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ApplicationForkingServiceCEImpl.java
abhvsn: Can we please move this to requested location just like we had done for cloneApplication?

abhvsn:
> @abhvsn We do not fetch the application there. It would be good to add it here right?

But here we expect we have the actual `srcApplicationId`, which is a abstraction layer we have between the actual business logic and the controller

abhvsn (changes requested):
Please follow the same implementation that we have for duplicate application API. Also can you please add the same for export application

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ApplicationForkingServiceCEImpl.java
```diff
@@ -92,6 +94,23 @@ public Mono<Application> forkApplicationToOrganization(String srcApplicationId,
...
+                    .switchIfEmpty(Mono.error(new AppsmithException(AppsmithError.NO_RESOURCE_FOUND, FieldName.APPLICATION, srcApplicationId)))
+                    .flatMap(application -> {
+                        // For git connected application user can update the default branch
+                        // In such cases we should fork the application from the new default branch
+                        if (!Optional.ofNullable(application.getGitApplicationMetadata()).isEmpty()
+                                && !application.getGitApplicationMetadata().getBranchName().equals(application.getGitApplicationMetadata().getDefaultBranchName())) {
+                            return applicationService.findBranchedApplicationId(
+                                    application.getGitApplicationMetadata().getDefaultBranchName(),
+                                    srcApplicationId,
+                                    AclPermission.MANAGE_APPLICATIONS
```
abhvsn, line 108: Please update this with read_permission as forking is enabled for view access as well for sample apps

abhvsn (changes requested):

PR #13257, opened 2022-04-25
fix: remove cell background from empty rows

### LINKED ISSUE #12774: [Bug]-[500]:Table column border is sensitive and triggers both move and resize actions on mouse click leading to bad UX

### Description

Resizing table columns causes columns to get rearranged since the border is sensitive to both resize and rearrange action. This causes UX issues where user has to be slow while resizing as seen in the video and control is still on the selected column so either the column is resized or rearranged when user lets go of control by release click action.

### Steps To Reproduce

1. Go to a table widget and add a few columns
2. Try to re-size columns by hovering over the column boundary to see the resize icon.
3. Click to select the resize option and observe cursor change to move icon
4. Let go of cursor click at various points on the column header and observe that the column size readjusts to the point where the resize action is registered (see video attached at the 4 second mark to see this behaviour)

[LOOM DEMO](https://www.loom.com/share/70dd8d17372b4fd392b6b67d7948147f)

### Public Sample App

_No response_

### Version

Cloud

### LINKED ISSUE #12823: [Bug]-[748]:Table widget - Cell background for Step column is applied incorrectly

### Description

When cell background colour is applied to the Step column, the cell background colour is applied for the entire column as well as the other cells which do not contain data.
It should only be applied to the cells of the Step column, which has data (the first three cells of the step column)

-- UPDATE 10 May 22 @dilippitchika 
- Cell background will be applied to the full column/columns, not just the rows which have data. This is to prevent inconsistency in the UI.

### Steps To Reproduce

1. Drag and drop a table widget onto the canvas.
2. In the Property Pane, go to the Step column settings
3. Choose a Cell Background colour as say orange or yellow
The cell background colour is applied for the entire Step column as well as the other cells which do not contain data. 
It should only be applied to the cells of the Step column, which has data (the first three cells of the step column), as per behaviour of other columns
Tablewidget_Step_Cellbackground

### Public Sample App

_No response_

### Version

Release

techbhavin:
## Description

> Fix Cell background for Step column is applied incorrectly

Fixes #12823

## Type of change

## Checklist:

    // Code coverage diff between base branch:release and head branch: fix/issue-12823 
 :green_circle: | total | 56.56 **(0)** | 38.56 **(0.01)** | 35.87 **(0)** | 56.8 **(0)**
 :green_circle: | app/client/src/widgets/TableWidget/component/TableUtilities.tsx | 35.68 **(0.2)** | 24.32 **(2.01)** | 15.79 **(0)** | 33.33 **(0.19)**</details>


Tooluloope:
@techbhavin is this the right connected issue?

https://github.com/appsmithorg/appsmith/issues/12774

techbhavin:
@Tooluloope, thanks for pointing out here that PR was linked with the wrong issue.  

Now , Linked with correct issue

Tooluloope (changes requested):
LGTM
Could you add a Cypress test for this fix

chandannkumar:
@techbhavin 
Observation: While adding **Cell Background Color** for entire Table widget,
In prod, it fills all the cells including empty cells with selected color whereas in DP we can see that it only fills the data populated cells and not empty cells.
What could be the expected behavior here?

App:


DP:

Tooluloope:
@dilippitchika What should be the expected behavior?

dilippitchika:
Based on this issue's description and the PR name it is suggested that we remove the background from empty rows. So the approach taken here is correct.

However, the challenge with this is it's sometimes inconsistent visually, when a user navigates to last page or when a user searches/filters. @momcilo-appsmith could you review this once on priority? Should we show the background for all rows or only rows which have data. I think we should show for all rows, irrespective of whether they have data or not.

uxdsatyajit:
@dilippitchika 
looking into it

UXFalcon:
looking into it @dilippitchika

Tooluloope:
@UXFalcon any update?

UXFalcon:
> @UXFalcon any update?

Yes, I have reviewed the scenarios, It's wise to say that the behaviour is correct on all rows and not just data specific, if we remove background  then the basic design flow would be broken 
CC @dilippitchika

Tooluloope:
@techbhavin Any update on this PR

techbhavin:
@Tooluloope  i am working on it

Tooluloope (dismissed):

Tooluloope (approved):

chandannkumar:

PR #13263, opened 2022-04-25
fix: fix action objects missing plugin id and plugin type info in database

### LINKED ISSUE #11927: [Bug]: Erratic behaviour with missing pluginId for the actions created by JSObjects

### Description

Some users have reported that their app is crashing, after investigating this resulted in the missing `pluginType` and `pluginId` for the actions connected to the JSObejct. This happens with all the actions attached to the JSObject which is at fault but if we look into the DB, JSObject looks fine with `pluginType` and `pluginId` is also available.

### Steps To Reproduce

-

### Public Sample App

_No response_

### Version

Cloud

sumitsum:
## Description
- We have encountered some cases where `pluginId` and `pluginType` were found to be missing in the action objects related to `JSObject`. Since the exact cause is not yet know, this PR introduces a change to set correct `pluginId` and `pluginType` for any action object when the action object is read. 
- To find the correct `pluginId` and `pluginType` the following logic is used:
  - If JS collection id is not null, then assume that the `pluginType` is `JS` and fetch the corresponding `pluginId` from the database.
  - Otherwise, if `pluginId` is present in the nested datasource object, then use it to fetch and set `pluginType` and `pluginId`.
- This change modifies all the `read action` and `update action` flows to check for missing `pluginId` and `pluginType`. The reason for adding it to `read action` is that in case a user's application has already been corrupted, it will allow the application to be usable. Once the application has been opened, any further update to the action will fix the action. 
- Another related problem would be `pluginType` and `pluginId` being malformed. However, this issue would be picked up separately, reason being that:
  - we have not yet encountered or seen any such cases being reported 
  - this issue much more difficult to deal with and would require some discussion with other engineers around a suitable fix
  - tracking it here: https://github.com/appsmithorg/appsmith/issues/13281
- This change does not address the root cause of the problem - i.e why the plugin id and type go missing. Instead it attempts to repair the damage by supplying the correct values. Another issue has been raised to track the root cause analysis for this problem: https://github.com/appsmithorg/appsmith/issues/13280 

Fixes #11927

## Type of change

## How Has This Been Tested?
- JUnit TC
- Manual

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/NewActionServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/NewActionServiceCEImpl.java
@@ -373,10 +373,10 @@ public class NewActionServiceCEImpl extends BaseService<NewActionRepository, New
                     action.setInvalids(invalids);
                     action.setPluginName(plugin.getName());
                     newAction.setUnpublishedAction(action);
-                    newAction.setPluginType(plugin.getType());
-                    newAction.setPluginId(plugin.getId());
                     return newAction;
-                }).map(this::extractAndSetJsonPathKeys)
+                })
+                .flatMap(this::sanitizeAction)
+                .map(this::extractAndSetJsonPathKeys)
                 .map(updatedAction -> {
                     // In case of external datasource (not embedded) instead of storing the entire datasource
                     // again inside the action, instead replace it with just the datasource ID. This is so that
@@ -1123,17 +1123,13 @@ public class NewActionServiceCEImpl extends BaseService<NewActionRepository, New
     @Override
     public Mono<NewAction> findById(String id) {
         return repository.findById(id)
-                .flux()
-                .flatMap(this::sanitizeAction)
-                .last();
+                .flatMap(this::sanitizeAction);
     }
 
     @Override
     public Mono<NewAction> findById(String id, AclPermission aclPermission) {
         return repository.findById(id, aclPermission)
-                .flux()
-                .flatMap(this::sanitizeAction)
-                .last();
+                .flatMap(this::sanitizeAction);
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/NewActionServiceCEImpl.java
```diff
@@ -1113,27 +1116,36 @@ public Flux<NewAction> findUnpublishedOnLoadActionsExplicitSetByUserInPage(Strin
...
-                .findUnpublishedActionsByNameInAndPageId(names, pageId, MANAGE_ACTIONS);
+                .findUnpublishedActionsByNameInAndPageId(names, pageId, MANAGE_ACTIONS)
+                .flatMap(this::sanitizeAction);
     }
 
     @Override
     public Mono<NewAction> findById(String id) {
-        return repository.findById(id);
+        return repository.findById(id)
+                .flux()
```
nidhi-nair, line 1126: Hey why are we creating a flux here?

nidhi-nair:
Hey @sumitsum  so I understand that this change you're making is so that JS objects can be recovered, but by changing the value during read we're introducing inconsistency between the db and the UI. Do you think there's a chance that this might complicate matters further down the line? Instead, can we do the exact same thing during writes?

sumitsum:
> * in case a user's application has already been corrupted, it will allow the application to be usable. Once the application has been opened, any further update to the action will fix the action.

@nidhi-nair thanks for pointing this out. I have updated the change to do the following: 
- add the `sanitizeAction` flow to `update action` methods 
- remove the `save` step from the `sanitizeAction`
So now even though `sanitizeAction` is a part of the `get action` flow, it will not persist any change to the database. Also, we would want to keep the `sanitizeAction` flow with the `get action` flow as well because in case a user's application has already been corrupted, it will allow the application to load without crashing and any subsequent update action api call will fix the database entry. However, if we remove it from the `get action` flow then the application may become inaccessible (may crash) which will not allow it to be recovered.

nidhi-nair (approved):

PR #13339, opened 2022-04-27
feat: Analytics to log source data limit exceed

### LINKED ISSUE #13220: [Task] : Add/Reuse an event to measure the JSON form 50 field limit
We need a way to measure the number of users who are hitting the 50 field limit in the JSON form.

Suggested event changes
1. Widget property update to contain an extra field key called meta_info which can contain this detail
2. Or We can use the debugger error message, but here we also have to start calling the debugger in the field.

ashit-rath:
## Description
Analytics to log source data limit exceed.

Fixes #13220

## Type of change

## Checklist:

### DIFF
--- a/app/client/src/widgets/JSONFormWidget/widget/helper.ts
+++ b/app/client/src/widgets/JSONFormWidget/widget/helper.ts
@@ -272,6 +272,7 @@ export const computeSchema = ({
       updatedValue: currSourceData,
       metaInfo: {
         limitExceeded: true,
+        currentLimit: MAX_ALLOWED_FIELDS,
       },
     });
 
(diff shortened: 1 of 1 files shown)


dilippitchika (commented):
@ashit-rath one doubt, if the data is bound to a query and the query only sends more than 50 fields in view mode, will this event be fired?

ashit-rath:
@dilippitchika Yes, it should. 
I think we can verify that in the DP

dilippitchika:
Tested, working as expected.

dilippitchika (approved):
LGTM

ashit-rath:
@dilippitchika We don't need to track the cases `limitExceeded: false` right?

dilippitchika:
No we don't need to track those @ashit-rath

Issue #12606, opened 2022-04-05
Task: Create reducers for storing auto-height layout for computations


ankurrsinghal:
Auto-height computations work well when we have a data structure based on the current DSL of the page generated and ready.

This task is to create reducers and relevant actions to store this data structure.

Issue #11038, opened 2022-02-09
[Epic] Dynamically Change height for widgets (New Size Options for Container and other widgets)


momcilo-appsmith:
## Objective

Enable widgets to mimic standard HTML block elements by automatically extending the height to fit contents.

## Success Metrics *

| Goal | Metric |
| --- | --- |
| Mimic HTML block component behaviour | Feature enabled numbers in deployed apps?  (TBD) |

Requirements

| Requirement | User story |
| --- | --- |
| Block or "Container" like widgets extend height to fit all contents | Web developers are used to seeing block elements extend until configured with a fixed height (Standard HTML block element behaviour). Appsmith developers can now find a parallel behaviour in Appsmith |
| Elements have a concept of max-height, a configurable property  | One of the common CSS properties used by web developers to restrict how much an element can "grow" in size, is using the max-height property. Appsmith developers can now mimic this in Appsmith |
| Hide this feature from certain widgets, that do not allow this behaviour | Not all HTML elements increase in height based on contents, especially the ones which don't have contents like `br` and `hr`, similarly, widget developers can configure if a widget can extend height dynamically in widget configurations |

## DX updates
- Widgets will have a new property called Height with two options in a dropdown `FIXED`, `HUG_CONTENTS`
- When the property above is `HUG_CONTENTS`, Appsmith developers will see more properties called `Max Height in Rows` : `number` and `Min Height in Rows`: `number`. These define the maximum/minimum height to which the widget can grow/shrink automatically. 
- Default Min Height is 2 rows (this is also the minimum value which can be configured), while Default Max Height is 1000 rows or some arbitrarily large value.

- `FIXED`: This is the current mode for widgets, where the heights once defined, shall not be changed
- `HUG_CONTENTS`: 
   - In the edit mode, the height will change dynamically, and Appsmith developers will not be allowed to manually resize the height. These changes will be persisted in the DSL.
   - In the view mode, the height will change dynamically but the values will not be persisted in the DSL.

Note: These properties do not conflict with "Scroll contents" property as once the height is dynamic, the widget will not have a scroll

## Documentation

- A detailed documentation with screen recordings demonstrating the logic of reflow
- A detailed documentation with screen recordings demonstrating the two new properties

## Reflow Logic (UX)

[...]

PR #13360, opened 2022-04-27
Redux action to handle multiple widget properties updates

### LINKED ISSUE #12494: Task: Store temporary positions and heights for dynamic height in widgets
As widgets will be reflowed when a widget increases height based on its contents (Dynamic height feature), we need to make sure we store this transient data.
This task identifies the data structure and the storage mechanism as well as its implementation.

riodeuno:
## Description

- Create UPDATE_MULTIPLE_WIDGET_PROPERTIES action
- Update specific paths in the reducer, maintaining reference for other paths
- Add unit tests for reducer

Fixes #12494 #12606

## Type of change

## How Has This Been Tested?
- Unit tests for reducers

## Checklist:

    // Code coverage diff between base branch:release and head branch: feat/update-multiple-widgets-reducer 
 :green_circle: | total | 56.5 **(0.01)** | 37.97 **(0.01)** | 35.84 **(0.01)** | 56.73 **(0.01)**
 :sparkles: :new: | **app/client/src/reducers/entityReducers/canvasWidgetsReducer.ts** | **95.45** | **100** | **75** | **93.33**
 :sparkles: :new: | **app/client/src/reducers/entityReducers/widgetConfigReducer.ts** | **100** | **100** | **100** | **100**
 :x: | ~~app/client/src/reducers/entityReducers/canvasWidgetsReducer.tsx~~ | ~~85.71~~ | ~~100~~ | ~~50~~ | ~~85.71~~
 :x: | ~~app/client/src/reducers/entityReducers/widgetConfigReducer.tsx~~ | ~~100~~ | ~~100~~ | ~~100~~ | ~~100~~</details>


app/client/src/reducers/entityReducers/canvasWidgetsReducer.ts
```diff
@@ -0,0 +1,70 @@
...
+  ) => {
+    return action.payload.widgets;
+  },
+  [ReduxActionTypes.UPDATE_LAYOUT]: (
+    state: CanvasWidgetsReduxState,
+    action: ReduxAction<UpdateCanvasPayload>,
+  ) => {
+    return action.payload.widgets;
+  },
+  [ReduxActionTypes.UPDATE_MULTIPLE_WIDGET_PROPERTIES]: (
```
rahulramesha, line 44: This seems to work very similar to `BATCH_UPDATE_MULTIPLE_WIDGETS_PROPERTY`. Difference is, it is being handled in a saga instead of the reducer, could that be reused instead of creating a new Action?

riodeuno: Good point. 
Upon further thought, there is a particular difference here. I don't wish for these changes to go through evaluations. The BATCH_UPDATE_MULTIPLE_WIDGETS_PROPERTY will cause evaluations, and hence it is being persisted into the redux store using `UPDATE_WIDGETS`. `UPDATE_WIDGETS` is an action which our evaluation sagas listen to.

1. We could make sure that we don't all `UPDATE_WIDGETS` and always call `UPDATE_MULTIPLE_WIDGET_PROPERTIES`, this will help us avoid updating all widgets and only update particular paths. Needs a relatively large refactor and might be necessary in the upcoming connected widgets update.

2. Let `UPDATE_MULTIPLE_WIDGET_PROPERTIES ` be evaluated as well. We might want to do so in the future, however, at the moment i makes more sense not to do so, as it will cause unnecessary computations, which are moot.

3. When `BATCH_UPDATE_MULTIPLE_WIDGETS_PROPERTY` action is dispatched, if a reducer is also handling this action, then the reducer will run first and then will the saga. This is something which might not be desirable. 


Thoughts?

rahulramesha: Since this is a specific scenario, that does not require evaluations and escapes the Undo - Redo capture. We should probably rename this to describe it that way or developers will end up using this in their changes, which might break the undo-redo.
 
 Also, if we are using this to update the dynamic height of the widgets, wont this persist and Save the DSL, which might cause problems to the Git work flow?

riodeuno: Yep, changing the name sounds appropriate.

Could you help me think through this:
Let's say we update the height of a few widgets dynamically -- we don't save the DSL and we don't evaluate and we escape undo-redo stack --; later, a new property update occurs, this time, it causes evaluations and dispatches the `BATCH_UPDATE_MULTIPLE_WIDGETS_PROPERTY` action. The DSL will be saved along with the dynamic height changes from earlier. I am wondering how all this will effect git and undo-redo, Will this cause issues?

rahulramesha: We can try out the flow,
1. We update heights and position of few widgets dynamically, it won't be saved to DSL, it will escape Undo capture
2. User performs another action, it will update new property and the changes that was made while updating dynamic height. This will be captured in the undo stack and save the DSL to the backend
From here,
- If the user performs undo action, it will not only undo the property updated but also the heights and positions changes in the dynamic height. The code might correct itself again by calling dynamic height, but in the end it will be a annoying flicker.
- if the user commits the change to the git, It will involve all the dynamic height related position changes. The user will end up seeing changes to the DSL that were not directly the his actions. Imagine changing a single word and seeing 100s of lines changed in the git repository

riodeuno: Yep, my thoughts exactly. Seems like we need to look into the details of the edit mode. The view mode is fine, because, we can simply use a temp store separate from the canvasWidgets and have it override during render.

riodeuno: @rahulramesha Based on yesterday's discussion, I'm reopening this. The current idea is to create the POC, investigate the flicker duration, and try to optimise the rendering such that the flicker isn't annoying to users.

rahulramesha: @riodeuno Got it, If it is just for the sake of POC maybe we can change the base branch to the other branch and test it out there instead of release.

riodeuno: @rahulramesha Sounds good. This will be the base branch. I won't remove this from draft at this moment.

riodeuno:
Closing this for now. We'll re-create a new PR for this change as a part of the #11038 EPIC

PR #13377, opened 2022-04-27
task: 12949 Trim unnecessary re-renders in baseWidget layers 

### LINKED ISSUE #12949: [Task]: Prevent the layers on top of widgets from rendering in edit mode
In edit mode each widget has these 9 layers on top of it.

- PositionedContainer
-  SnipeableComponent
- DraggableComponent
- WidgetNameComponent
- ResizableComponent
- OverlayCommentsWrapper
- PreventInteractionsOverlay
- PreviewModeComponent
- ErrorBoundary

Whenever you change a property in the property pane, or edit a meta property  all these layers render for all widgets on the canvas and none of these have to re render.

Though this PR handles most of the issues with wasted widget renders #12727, we will still have a problem when there are lot connected widgets. For example a large form widget depending on the table selected row or list widget active item.

**Proposed solution**

Refactor the getWidgetView in BaseWidget class and 

- Fix the selectors used by all the layers to prevent wasted renders.
- Memoize the positioned containers style calculation.
- Inline the components instead of call them as functions [Not sure this step is required]

prsidhu:
## Description

Target:
1. Remove unnecessary re-renders in onion layers of the base widget.
2. Remove unnecessary renders in sibling widgets upon selection of a widget.

Changes:
1. Create selectors for checking if a widget is selected or focused.
2. Memoize WidgetName component.

Fixes #12949

## How Has This Been Tested?

1. Clone the A force app.
2. Click on an issue. Wait for the description to load. Click on the title input.
3. Turn on the React Profiler.
4.  Click on Title input and type a character.
5. Turn off the React profiler
6. Observe that there are no re-renders in the onion layers of other components.

Repeat steps 2-6 on the aforce app on release to observe the difference in the number of re-renders.

Repeat steps 2-6 on the release aforce and the cloned app on this branch. Profile using the performance tab. Observe the difference in the scripting times in the two scenarios.

## Checklist:

    // Code coverage diff between base branch:release and head branch: task/12949 
 :green_circle: | total | 56.62 **(0.01)** | 38.54 **(0)** | 35.89 **(0)** | 56.86 **(0.01)**
 :red_circle: | app/client/src/components/editorComponents/ResizableComponent.tsx | 63.1 **(-1.27)** | 21.88 **(2.21)** | 44.44 **(-13.89)** | 63.1 **(-1.27)**
 :red_circle: | app/client/src/components/editorComponents/SnipeableComponent.tsx | 77.78 **(0)** | 27.27 **(0)** | 20 **(-13.33)** | 77.78 **(0)**
 :red_circle: | app/client/src/components/editorComponents/WidgetNameComponent/index.tsx | 96.83 **(-0.14)** | 82.69 **(-0.95)** | 100 **(0)** | 96.72 **(-0.16)**
 :red_circle: | app/client/src/selectors/propertyPaneSelectors.tsx | 84.72 **(-1.39)** | 64.81 **(0)** | 83.33 **(-8.34)** | 87.1 **(-1.61)**
 :green_circle: | app/client/src/selectors/ui.tsx | 80.95 **(3.17)** | 100 **(0)** | 42.86 **(9.53)** | 71.43 **(4.76)**
 :green_circle: | app/client/src/selectors/widgetSelectors.ts | 73.21 **(10.71)** | 33.33 **(-2.38)** | 82.35 **(25.21)** | 71.43 **(9.53)**
 :green_circle: | app/client/src/utils/hooks/useClickToSelectWidget.tsx | 64.81 **(1.04)** | 64 **(15.11)** | 40 **(-10)** | 62.75 **(0.63)**
 :green_circle: | app/client/src/widgets/BaseWidget.tsx | 80.3 **(0.46)** | 48.94 **(0)** | 77.14 **(0.67)** | 79.53 **(0.5)**</details>


app/client/src/widgets/BaseWidget.tsx
```diff
@@ -193,7 +221,7 @@ abstract class BaseWidget<
     return (Number(this.props.labelWidth) || 0) * this.props.parentColumnSpace;
   };
 
-  getErrorCount = memoize((evalErrors: Record<string, EvaluationError[]>) => {
+  getErrorCount = moize((evalErrors: Record<string, EvaluationError[]>) => {
```
SatishGandham, line 224: Use only one memoization library.

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -382,23 +410,18 @@ abstract class BaseWidget<
   /**
    * generates styles that positions the widget
    */
-  private getPositionStyle(): BaseStyle {
+  private getPositionStyle = () => {
     const { componentHeight, componentWidth } = this.getComponentDimensions();
-
-    return {
-      positionType: PositionTypes.ABSOLUTE,
+    return getPositionStyle_(
```
SatishGandham, line 415: A better name please :)

app/client/src/components/editorComponents/DraggableComponent.tsx
```diff
@@ -70,24 +74,17 @@ export const canDrag = (
 function DraggableComponent(props: DraggableComponentProps) {
   // Dispatch hook handy to set a widget as focused/selected
   const { focusWidget, selectWidget } = useWidgetSelection();
-
   const isCommentMode = useSelector(commentModeSelector);
   const isSnipingMode = useSelector(snipingModeSelector);
   const isPreviewMode = useSelector(previewModeSelector);
+  const isSelected = useSelector(isCurrentWidgetSelected(props.widgetId));
```
SatishGandham, line 80: nit pick, but can we change the name to isCurrentWidgetSelected isWidgetSelected since it works for any widget.

app/client/src/components/editorComponents/ResizableComponent.tsx
```diff
@@ -64,15 +70,15 @@ export const ResizableComponent = memo(function ResizableComponent(
   const showTableFilterPane = useShowTableFilterPane();
   const { selectWidget } = useWidgetSelection();
   const { setIsResizing } = useWidgetDragResize();
-  const selectedWidget = useSelector(
```
SatishGandham, line 67: We don't need to use dynamic selector when the selector is already returning a primitive value.

app/client/src/selectors/widgetSelectors.ts
```diff
@@ -49,3 +50,33 @@ export const getParentWidget = createSelector(
     return;
   },
 );
+
+// Check if current widget is in the list of selected widgets
+export const isCurrentWidgetSelected = (widgetId: string) => {
```
SatishGandham, line 55: We don't need dynamic selectors when the selector returns a primitive value. Please make necessary changes to all such selectors.

app/client/src/components/editorComponents/ResizableComponent.tsx
```diff
@@ -242,7 +240,7 @@ export const ResizableComponent = memo(function ResizableComponent(
     };
 
     return omit(allHandles, get(props, "disabledResizeHandles", []));
-  }, [props]);
+  }, [get(props, "disabledResizeHandles", [])]);
```
riodeuno, line 243: Can we move the `get(props, "disabledResizeHandles", [])` out into a variable and use it in both the dependencies and inside the `useMemo`? Seems more readable that way.

app/client/src/components/editorComponents/WidgetNameComponent/index.tsx
```diff
@@ -122,25 +124,18 @@ export function WidgetNameComponent(props: WidgetNameComponentProps) {
...
-    selectedWidget === props.widgetId ||
-    selectedWidgets.includes(props.widgetId);
+  const showAsSelected = isLastSelected || isSelected;
 
-  const isMultiSelectedWidget =
-    selectedWidgets &&
-    selectedWidgets.length > 1 &&
-    selectedWidgets.includes(props.widgetId);
+  const isMultiSelected = useSelector(isMultiSelectedWidget(props.widgetId));
   const shouldShowWidgetName = () => {
```
riodeuno, line 130: Should this be a `useCallback` to avoid defining this function every render?

riodeuno: Alternatively, shall we make it a value using `useMemo`?

app/client/src/selectors/widgetSelectors.ts
```diff
@@ -49,3 +53,66 @@ export const getParentWidget = createSelector(
...
+// Check if widget is in the list of selected widgets
+export const isWidgetSelected = (widgetId: string) => {
+  return createSelector(getSelectedWidgets, (widgets): boolean =>
+    widgets.includes(widgetId),
+  );
+};
+
+export const isCurrentWidgetFocused = (widgetId: string) => {
+  return createSelector(
+    (state: AppState) => state.ui.widgetDragResize.focusedWidget,
```
riodeuno, line 66: Should we create a function called `getFocusedWidget` for consistency?

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -40,6 +41,33 @@ import AppsmithConsole from "utils/AppsmithConsole";
 import { ENTITY_TYPE } from "entities/AppsmithConsole";
 import PreviewModeComponent from "components/editorComponents/PreviewModeComponent";
 
+const memoizedPositionStyle = memoize(
```
riodeuno, line 44: This function is used only in one place. 
`xPositionUnit` and `yPositionUnit` are practically useless today, as everything is in pixels.
`componentHeight` and `componentWidth` come from a different function.
The only computations happening here are for `yPosition` and `xPosition`, and the values are primitives.

Based on the above, shall we simply remove this function and compute the `yPosition` and `xPosition` just in time for usage?
Will that help?

cc @SatishGandham @prsidhu

riodeuno: I am concerned about the `maxSize:1000`, and wondering if this is necessary. No matter what, the memoize function will still compute to confirm if it actually has to run the function. So, instead we could use that compute time in computing the `yPosition` and `xPosition` directly. There won't be any objects which will cause reference inequalities.

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -193,15 +221,15 @@ abstract class BaseWidget<
     return (Number(this.props.labelWidth) || 0) * this.props.parentColumnSpace;
   };
 
-  getErrorCount = memoize((evalErrors: Record<string, EvaluationError[]>) => {
+  getErrorCount = (evalErrors: Record<string, EvaluationError[]>): number => {
```
riodeuno, line 224: This returns a primitive. Are we sure that the memoization making an impact?

riodeuno (commented):
This is awesome! We've identified and removed un-needed computations in components using the new selectors. 

I have a few queries.

app/client/src/widgets/BaseWidget.tsx
SatishGandham: @riodeuno , We are memoizing so that we don't return a new style object.  We are not saving computing time, we are saving the render time for positioned container.

We need a maxSize because, the same function will be called for all the widgets. If we maintain the cache only for the last time, we will never hit the cache.

@prsidhu , Can you confirm the above.

app/client/src/widgets/BaseWidget.tsx
prsidhu: Memoization wasn't making an impact here. Hence, we've chosen to remove it.

app/client/src/widgets/BaseWidget.tsx
prsidhu: @SatishGandham @riodeuno This cache is shared across all widgets. Removing maxSize makes the use pointless as the cache is never hit.

SatishGandham (dismissed):

SatishGandham:
@vivekverma2312 , @btsgh This needs QA. 
We are optimizing how the widget is rendered in edit mode. 
See if everything works as expected, test drag and drop, resize, move.

@prsidhu , Please add if anything else needs to be tested.

riodeuno:
@prsidhu Could you update the PR description with how we're measuring the improved performance, and what the metrics are between production/release and this PR?

app/client/src/widgets/BaseWidget.tsx
riodeuno: @prsidhu @SatishGandham Yep, the maxSize caching should work. Did we give any thought to what I suggested regarding not using the function at all, and moving to simpler primitive computations of `yPosition` and `xPosition`?

Do let me know  your thoughts on this approach.

prsidhu:
@riodeuno Updated the description. Would this suffice?

riodeuno:
@prsidhu Do you have the screenshots of the scripting times handy? It will help to show the improvement as well.

PR #13387, opened 2022-04-27
fix: Unable to see apps in home page after git connect fails

### LINKED ISSUE #13385: [Bug]: Unable to see application in home page after the git connect flow is aborted in middle

### Description

User is Unable to see application in home page after the git connect flow is aborted in middle 

### Steps To Reproduce

1. Create an app and click on connect to git
2. Enter the url and generate keys
3. Save the edit mode url of the app in a different place
4. Now go back to home page and refresh 
5. Observe the app is missing from the home page

### Public Sample App

_No response_

### Version

Cloud v1.6.22

AnaghHegde:
## Description

> If the user aborts git connect flow or if there is any error then the app is missing from the homepage. 

Fixes #13385 

## Type of change

## How Has This Been Tested?

- Locally
- JUnit tests

## Checklist:


abhvsn:
We should fix this in connect flow as that's the root cause for failure cases. Also what happens if we end up in having empty default branch for multiple branches of course considering the error scenario that user has reported?

AnaghHegde:
@abhvsn This fix is for the empty default branch, when only the ssh key is present and no other information is present in db. 
Not sure what is the issue in connect flow. AM i missing something here?

abhvsn:
@AnaghHegde this is required but I'm suggesting, what happens when connect flow is not successful because of some failure in that case also we end up in stale application should we remove the gitApplicationMetadata to keep the DB clean

AnaghHegde:
@abhvsn it is good to remove the git metadata from db in case of failures to keep things clean.But this won't affects us. 
I have added test cases for the current scenario. But we don't have to specifically check for both branchName and defaultBranchName as we don't have case where only one of them is present.

abhvsn:
Still I think adding a testcase for connect and create branch flow is must though removing a gitMetadata can be optional, I'm asking this because we may end up in branched applications in DB without defaultBranchName but have all the other fields which will result in duplicates in homepage

AnaghHegde:
@abhvsn In create branch flow, we set both defaultBranch and branchName fields. So if the resource is created in DD then both of them will be present. It won't be a case where only branchName is present and not defaultBranchName. We copy the gitApplicationMetadata from the parent app and then we update fields before saving it to DB. So i don't see how we end up in a state you are mentioning. The same goes for connect flow. If we save the branch name in db then the default branchName also gets updated as they are all updated in the same map.

abhvsn (dismissed):
LGTM
Q: Are we going to add the testcases in separate PR for above mentioned scenario?

AnaghHegde:
@abhvsn I have added test case for the above case in this PR. If you look at this L.304 , i am testing the same scenario.
https://github.com/appsmithorg/appsmith/blob/4dfd800d9b2f53b7f4338615767a1e675cf8d7bc/app/server/appsmith-server/src/test/java/com/appsmith/server/solutions/ApplicationFetcherUnitTest.java#L304

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ApplicationFetcherCEImpl.java
```diff
@@ -113,8 +113,9 @@ public Mono<UserHomepageDTO> getAllApplications() {
                     Flux<Application> applicationFlux = applicationRepository
                             .findByMultipleOrganizationIds(orgIds, READ_APPLICATIONS)
                             .filter(application -> application.getGitApplicationMetadata() == null
-                                    || (!StringUtils.isEmpty(application.getGitApplicationMetadata().getDefaultBranchName())
-                                    && application.getGitApplicationMetadata().getBranchName().equals(application.getGitApplicationMetadata().getDefaultBranchName()))
+                                            ||  (StringUtils.isEmpty(application.getGitApplicationMetadata().getDefaultBranchName())
+                                                            && StringUtils.isEmpty(application.getGitApplicationMetadata().getBranchName()))
+                                            || application.getGitApplicationMetadata().getBranchName().equals(application.getGitApplicationMetadata().getDefaultBranchName())
```
sharat87, line 118: I think there's a potential NPE here. if `.getDefaultBranchName()` is non-empty, and `.getBranchName()` is `null`, then we'll get an NPE here, right? Can we instead use something like `StringUtils.equals` or something, that knows how to handle `null` values?

AnaghHegde: @sharat87 `StringUtils.isEmpty` checks for both null and empty. Is that not sufficient? Do i need to check for null again with equals? Reference - https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/StringUtils.html#isEmpty(java.lang.String)

sharat87: Yes, but that call won't happen at all here.

In the case I described above, the condition on 116 will be `false`, and due to short-circuiting, the condition on 117 won't be called at all. But since the result of the full condition on 116-117 is `false`, the condition in line 118 will be called. Here, if `.getBranchName()` returns `null`, we'll get an NPE. The `isEmpty` was never called with `.getBranchName()`.

AnaghHegde: Ahh yes, Good catch. I did not think of the short circuit eval in the `&&` operator.

sharat87 (approved):

PR #13441, opened 2022-04-29
remember email when invalid credentials are given

### LINKED ISSUE #7803: [Enhancement] Login on incorrect credentials, email field should retain the entered email.
## Summary

On the login screen, if an incorrect email or password is entered, both the email and password fields are reset.
Ideally, the entered email should be retained and the password can be cleared off.

ganganimaulik:
## Description
stores username in localstorage & displays it when ?error=true is present in URL. otherwise it will use "" empty string.

let me know if I should proceed adding cypress test for this

Fixes #7803

## Type of change

## How Has This Been Tested?

- try to login with invalid credentials and make sure username(email) input is persisted 

## Checklist:

### DIFF
--- a/app/client/src/ce/pages/UserAuth/Login.tsx
+++ b/app/client/src/ce/pages/UserAuth/Login.tsx
@@ -30,6 +30,7 @@ import FormTextField from "components/ads/formFields/TextField";
 import Button, { Size } from "components/ads/Button";
 import ThirdPartyAuth from "@appsmith/pages/UserAuth/ThirdPartyAuth";
 import { ThirdPartyLoginRegistry } from "pages/UserAuth/ThirdPartyLoginRegistry";
+import localStorage from "utils/localStorage";
 import { isEmail, isEmptyString } from "utils/formhelpers";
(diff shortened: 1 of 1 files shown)


arunvjn (changes requested):
@ganganimaulik Please use [sessionStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) here instead of localStorage since they are short lived and unique to every tab (unless you duplicate tabs).

ganganimaulik:
> @ganganimaulik Please use [sessionStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage) here instead of localStorage since they are short lived and unique to every tab (unless you duplicate tabs).

thanks for suggestion. updated

arunvjn:
@ganganimaulik There seems to be a merge conflict in mapStateToProps function. `isPasswordFieldDirty` looks like a newly added field. Could you resolve this ? Also, extract mapStateToProps function from the connect statement.

arunvjn (approved):

Issue #13434, opened 2022-04-29
[Bug]: Don't import the datasource when destination org doesn't have specific plugin installed 


abhvsn:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

We have started introducing custom plugins at org level for simpler integrations with third party like twilio etc. This may result in null values getting stored for the datasource object during the import-application flow as destination org may not have the custom plugin installed. This results in making the org useless as we can't open any of the application in the edit mode.    

### Steps To Reproduce

1. Install custom plugin
2. Create action based on this plugin
3. Export application
4. Import application in other org which don't have this custom plugin
5. Try to open the application in the edit mode, 

### Version

Cloud: v1.6.22

abhvsn:
Closing as a part of https://github.com/appsmithorg/appsmith/pull/13487

PR #13487, opened 2022-05-02
fix: Fix missing pluginId issue for datasource during the import flow if destination org don't have sepcific plugin installed

### LINKED ISSUE #13630: [Bug]:The discard changes progress bar will be stopped in the middle and the action is not completed

### Description

The discard changes progress bar will be stopped in the middle and the action is not completed
image

### Steps To Reproduce

1. Connect application to git
2. Create a new Branch from master
3. Add a new page in the branch 
4. now do the discard change the modal will be the 

### Public Sample App

_No response_

### Version

Deploy Preview

abhvsn:
## Description

> In import-export flow we were expecting the same plugins should be installed in both source and destination org. but with the custom plugins coming in this assumption will not hold true and imported org might end up in datasources without pluginIds. With this PR we are checking if the destination org contains all the required plugins and throwing the exception if any of the plugin is not available.

Fixes #13630 [#13434](https://github.com/appsmithorg/appsmith/issues/13434)

## Type of change

## How Has This Been Tested?

> JUnit TC
> Manual testing

## Checklist:


nidhi-nair:
Hey @abhvsn, what is the expected behaviour for when import happens between incompatible orgs? Do we simply disallow it?

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -716,6 +716,13 @@ public Mono<Application> importApplicationInOrganization(String organizationId,
                             .filter(datasource -> datasource.getGitSyncId() != null)
                             .forEach(datasource -> savedDatasourcesGitIdToDatasourceMap.put(datasource.getGitSyncId(), datasource));
 
+                    // Check if the destination org have all the required plugins installed
+                    assert importedDatasourceList != null;
```
mohanarpit, line 720: Any particular reason we are using assert here instead of doing an if-check & returning an error message? 

Also, if we do write assert statements, can we please modify them to use the form: 
```
assert importedDatasourceList != null : "Got an invalid list for datasources"; 
```
The string expression returns a specific error message in the AssertionException instead of a generic one.

abhvsn:
> Hey @abhvsn, what is the expected behaviour for when import happens between incompatible orgs? Do we simply disallow it?

@nidhi-nair IMO if the actions are going to fail in the imported org we end up importing the application in broken state from user's perspective. So I thought limiting the import in non-compatible orgs makes more sense, by providing the specific error.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
abhvsn: We never expects the datasourceList as null value during the import, as during export if datasources are not attached to the application we end up with empty array. This check was added as IntelliJ was giving the warning, but as you suggested I'll add the `if` check to be consistent with other implementations.

nidhi-nair:
Gotcha. My question was more from a UX perspective. We're returning with a generic `plugin not installed` error from here, which might not help the developer user figure out what they are supposed to do or which organization their app can be imported into. Do you want to make the error message more helpful in that case?

abhvsn:
> Gotcha. My question was more from a UX perspective. We're returning with a generic `plugin not installed` error from here, which might not help the developer user figure out what they are supposed to do or which organization their app can be imported into. Do you want to make the error message more helpful in that case?

@nidhi-nair do we have a doc on how to install the custom plugin in an org? Do you think we should provide a doc URL on how to install the plugin in their org? Something like providing learn more links with the static error 
`Unable to find the {plugin-name} plugin, please follow the document to install required plugins!` 

Screenshot 2022-05-05 at 3 43 36 PM

nidhi-nair (approved):

PR #13541, opened 2022-05-04
fix: NPE when `connectionCookie` is `undefined`

sharat87:
Fix NPE in RTS when the `connectionCookie` is `undefined`.

### DIFF
--- a/app/rts/src/server.ts
+++ b/app/rts/src/server.ts
@@ -167,51 +167,54 @@ async function tryAuth(socket: Socket) {
     /* ********************************************************* */
 
     // const host = socket.handshake.headers.host;
-    const connectionCookie = socket.handshake.headers.cookie;
+    const connectionCookie = socket?.handshake?.headers?.cookie;
     if (connectionCookie == null || connectionCookie === "") {
         return false;
     }
 
     const matchedCookie = connectionCookie.match(/\bSESSION=\S+/);
-    if (matchedCookie) {
-        const sessionCookie = matchedCookie[0];
-        let response;
-        try {
-            response = await axios.request({
-                method: "GET",
-                url: API_BASE_URL + "/users/me",
-                headers: {
-                    Cookie: sessionCookie,
-                },
-            });
-        } catch (error) {
-            if (error.response?.status === 401) {
-                console.info("401 received when authenticating user with cookie: " + sessionCookie);
-            } else if (error.response) {
-                log.error("Error response received while authentication: ", error.response);
-            } else {
-                log.error("Error authenticating", error);
-            }
-            return false;
-        }
-
-        const email = response.data.data.email;
-        const name = response.data.data.name ? response.data.data.name : email;
-
-        // If the session check API succeeds & the email/name is anonymousUser, then the user is not authenticated
-        // and we should not allow them to join any rooms
-        if (email === "anonymousUser" || name === "anonymousUser") {
-            return false;
-        }
-
-        socket.data.email = email;
-        socket.data.name = name;
-
-        if (socket.data.pendingRoomId) {  // an appId or pageId is pending for this socket, join now
-            joinEditRoom(socket, socket.data.pendingRoomId, socket.data.pendingRoomPrefix);
-        }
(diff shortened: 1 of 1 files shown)


app/rts/src/server.ts
```diff
@@ -168,49 +168,50 @@ async function tryAuth(socket: Socket) {
...
-            } catch (error) {
-                if (error.response?.status === 401) {
-                    console.info("401 received when authenticating user with cookie: " + sessionCookie);
-                } else if (error.response) {
-                    log.error("Error response received while authentication: ", error.response);
-                } else {
-                    log.error("Error authenticating", error);
-                }
-                return false;
+    if (connectionCookie == null || connectionCookie === "") {
```
IAmAnubhavSaini, line 171: `===` for the first check.

app/rts/src/server.ts
```diff
@@ -168,49 +168,50 @@ async function tryAuth(socket: Socket) {
...
+        return false;
+    }
+
+    const matchedCookie = connectionCookie.match(/\bSESSION=\S+/);
```
IAmAnubhavSaini, line 175: Following check is missing:

```
if (!matchedCookie) {
    return false;
}
```

Then the code block in `matchedCookie` can be unindented too.

app/rts/src/server.ts
```diff
@@ -168,49 +168,50 @@ async function tryAuth(socket: Socket) {
...
+                log.error("Error response received while authentication: ", error.response);
+            } else {
+                log.error("Error authenticating", error);
             }
+            return false;
+        }
 
-            const email = response.data.data.email;
-            const name = response.data.data.name ? response.data.data.name : email;
+        const email = response.data.data.email;
```
IAmAnubhavSaini, line 198: `response?.data?.data?.email`

app/rts/src/server.ts
```diff
@@ -168,49 +168,50 @@ async function tryAuth(socket: Socket) {
 
     // const host = socket.handshake.headers.host;
     const connectionCookie = socket.handshake.headers.cookie;
```
IAmAnubhavSaini, line 170: `socket?.handshake?.headers?.cookie`

app/rts/src/server.ts
```diff
@@ -168,49 +168,50 @@ async function tryAuth(socket: Socket) {
...
+            } else {
+                log.error("Error authenticating", error);
             }
+            return false;
+        }
 
-            const email = response.data.data.email;
-            const name = response.data.data.name ? response.data.data.name : email;
+        const email = response.data.data.email;
+        const name = response.data.data.name ? response.data.data.name : email;
```
IAmAnubhavSaini, line 199: same as above

IAmAnubhavSaini (changes requested):

app/rts/src/server.ts
sharat87: That was the problem actually. If we use `===` for `null`, then `undefined` will go through. I want to check for `null` and `undefined`. It is recommended to use `==` for null-checks in Javascript, and `===` everywhere else. Like the `smart` option in eslint: <https://eslint.org/docs/rules/eqeqeq>.

app/rts/src/server.ts
sharat87: Done.

app/rts/src/server.ts
sharat87: Done.

app/rts/src/server.ts
sharat87: Done.

app/rts/src/server.ts
sharat87: Good call, done. :)

app/rts/src/server.ts
IAmAnubhavSaini: @sharat87 This change is pending. For maintainability it is very important for code to be unsurprising and boring.

sharat87: Done.

IAmAnubhavSaini (approved):

PR #13552, opened 2022-05-04
feat: dynamic height text widget standardisation

### LINKED ISSUE #13043: [Task]:  Standardise structure of Text widget to enable Dynamic Height.

### SubTasks

_No response_

ankurrsinghal:
## Description

1. Passed the `contentRef` to `TextComponent` from `InputWidgetV2`.
2. Wrapped the `TextComponent` class component in the React.forwardRef, and added a new optional prop `innerRef` to 
pass it to its child.
3. `StyledText` of `TextComponent` does not scroll as it's `overflow-y` is hidden. So I have to remove that so that it has a `scrollHeight`.
4. Wrapped the `StyledText` empty tag in a `div` and set the forwarded ref to it's ref.
5. Now this `div` has a scroll height to be used by the `BaseWidget`.

Fixes #13043

## Type of change

## How Has This Been Tested?

- Manually

## Checklist:


app/client/src/widgets/TextWidget/component/index.tsx
```diff
@@ -108,11 +108,6 @@ export const StyledText = styled(Text)<{
       ? `calc(100% - ${ELLIPSIS_HEIGHT}px)`
       : "100%"};
   overflow-x: hidden;
-  overflow-y: ${(props) =>
```
riodeuno, line 111: @ankurrsinghal I think we need this, because, this will take effect when we reach the max-height.

riodeuno (changes requested):
We need to look into how we can include dynamic height independent of the scroll/truncate feature.

PR #13619, opened 2022-05-05
feat: Add default url for REST API actions

### LINKED ISSUE #11697: [Feature]: Set the mock API by default when adding a new API

### Summary

When user adds a new data source and tries to add a API, we show the mock api path as placeholder. But when user starts editing the path, it's gone. It would be better if mock API path is there by default so that user can see it in action.

It'll make user more interested if they can try an API without thinking much. From my experience, it's not likely that someone has an API ready at hand when they've arrived here. After coming to this page when they find that an API URL is required, they need to find an JSON API first before testing this feature. 

If the mock-api URL was by default there, user's will be able to try that out and get familiar with this feature quickly.

ganganimaulik:
## Description
Set mock api as default value and display value

Fixes #11697
 
## Type of change
 
## How Has This Been Tested?

 
- create new datasource -> create new api
- if it's first api action for the project then it will have mock url as value otherwise only placeholder will be there with mock url.

## Checklist:


NilanshBansal:
Closing this as the original issue https://github.com/appsmithorg/appsmith/issues/11697 is closed

PR #13703, opened 2022-05-10
chore: Adds analytics for Property Pane keyboard navigation

### LINKED ISSUE #13217: [Task]: Send Analytics events for Property pane keyboard navigation
Event name is`PROPERTY_PANE_KEYPRESS`.
Meta data:
- key - The name of the key pressed
- widgetType
- propertyName

Also send additional parameter `isUpdatedViaKeyboard` to the existing `WIDGET_PROPERTY_UPDATE` event

aswathkk:
## Description

Adds `PROPERTY_PANE_KEYPRESS` event
Adds `isUpdatedViaKeyboard` parameter to `WIDGET_PROPERTY_UPDATE` event
Analytics events added for all the Property controls.

Fixes #13217

## Type of change

## How Has This Been Tested?

- Manual

## Checklist:

    // Code coverage diff between base branch:release and head branch: pp-kbd-analytics 
 :green_circle: | total | 56.76 **(0.05)** | 38.68 **(0.01)** | 36.27 **(0.05)** | 57.02 **(0.06)**
 :green_circle: | app/client/src/components/ads/ButtonTabComponent.tsx | 94.64 **(0.76)** | 95.83 **(0)** | 77.78 **(0)** | 93.48 **(1.17)**
 :green_circle: | app/client/src/components/ads/ColorPickerComponentV2.tsx | 79.64 **(0.79)** | 72.14 **(0)** | 75 **(0)** | 79.62 **(0.85)**
 :green_circle: | app/client/src/components/ads/Dropdown.tsx | 90.33 **(0.23)** | 71.97 **(0)** | 81.48 **(0)** | 89.96 **(0.29)**
 :red_circle: | app/client/src/components/ads/StepComponent.tsx | 89.74 **(-3.36)** | 83.33 **(-6.67)** | 80 **(-20)** | 89.74 **(-3.36)**
 :red_circle: | app/client/src/components/ads/Switch.tsx | 81.25 **(-18.75)** | 50 **(-25)** | 50 **(-50)** | 81.25 **(-18.75)**
 :green_circle: | app/client/src/components/ads/TreeDropdown.tsx | 95.75 **(0.18)** | 77.27 **(-0.25)** | 97.37 **(0)** | 96.92 **(0.13)**
 :red_circle: | app/client/src/components/editorComponents/ActionCreator/Fields.tsx | 35 **(-0.31)** | 24.46 **(-0.82)** | 10.39 **(0)** | 34.34 **(-0.31)**
 :green_circle: | app/client/src/components/editorComponents/CodeEditor/index.tsx | 77.16 **(0.19)** | 58.97 **(0.56)** | 65.22 **(0)** | 78.72 **(0.26)**
 :red_circle: | app/client/src/components/propertyControls/ActionSelectorControl.tsx | 82.35 **(-5.15)** | 40 **(-26.67)** | 75 **(0)** | 85.71 **(0)**
 :red_circle: | app/client/src/components/propertyControls/BaseControl.tsx | 72.22 **(-2.78)** | 44.44 **(0)** | 40 **(-10)** | 68.75 **(-2.68)**
 :red_circle: | app/client/src/components/propertyControls/BorderRadiusOptionsControl.tsx | 88.89 **(-5.23)** | 25 **(-25)** | 80 **(0)** | 86.67 **(-6.19)**
 :red_circle: | app/client/src/components/propertyControls/BoxShadowOptionsControl.tsx | 89.47 **(-4.97)** | 50 **(-25)** | 80 **(0)** | 87.5 **(-5.83)**
 :red_circle: | app/client/src/components/propertyControls/ButtonTabControl.tsx | 60.71 **(-2.25)** | 23.08 **(-4.19)** | 60 **(0)** | 61.9 **(0)**
 :red_circle: | app/client/src/components/propertyControls/IconSelectControl.tsx | 91.45 **(-0.04)** | 75.65 **(-0.46)** | 92 **(-3.83)** | 91.37 **(-0.1)**
 :red_circle: | app/client/src/components/propertyControls/IconTabControl.tsx | 69.57 **(-3.16)** | 28.57 **(-11.43)** | 75 **(0)** | 75 **(0)**
 :red_circle: | app/client/src/components/propertyControls/KeyValueComponent.tsx | 15.38 **(-0.2)** | 22.22 **(-1.78)** | 0 **(0)** | 16.67 **(-0.23)**
 :red_circle: | app/client/src/components/propertyControls/LabelAlignmentOptionsControl.tsx | 63.64 **(-3.03)** | 25 **(-8.33)** | 40 **(0)** | 72.22 **(0)**
 :red_circle: | app/client/src/components/propertyControls/NumericInputControl.tsx | 46.15 **(-3.85)** | 40 **(0)** | 40 **(-10)** | 47.83 **(-2.17)**
 :red_circle: | app/client/src/components/propertyControls/OptionControl.tsx | 66.67 **(-4.76)** | 0 **(0)** | 50 **(0)** | 75 **(0)**
 :red_circle: | app/client/src/components/propertyControls/StyledControls.tsx | 54.08 **(-1.24)** | 50.51 **(0)** | 41.3 **(0)** | 54.64 **(-1.27)**
 :red_circle: | app/client/src/components/propertyControls/SwitchControl.tsx | 87.5 **(-5.36)** | 76.92 **(-3.08)** | 71.43 **(-3.57)** | 86.36 **(-5.31)**
 :red_circle: | app/client/src/pages/Editor/PropertyPane/PropertyPaneView.tsx | 87.76 **(-1.98)** | 70.97 **(1.4)** | 62.5 **(2.5)** | 91.49 **(-3.1)**
 :green_circle: | app/client/src/utils/AppsmithUtils.tsx | 57.96 **(0.88)** | 40.98 **(0.15)** | 40.82 **(1.24)** | 53.33 **(0.9)**
 :red_circle: | app/client/src/utils/PropertyControlRegistry.tsx | 87.1 **(-12.9)** | 40 **(-60)** | 91.67 **(-8.33)** | 86.67 **(-13.33)**
 :red_circle: | app/client/src/utils/WorkerUtil.ts | 88.98 **(-0.78)** | 70.59 **(-1.96)** | 100 **(0)** | 92.38 **(-0.95)**
 :sparkles: :new: | **app/client/src/utils/hooks/useInteractionAnalyticsEvent.ts** | **94.12** | **80** | **100** | **100**</details>


app/client/src/components/ads/Dropdown.tsx
```diff
@@ -949,6 +954,9 @@ export default function Dropdown(props: DropdownProps) {
     (e: React.KeyboardEvent) => {
       switch (e.key) {
         case "Escape":
+          dropdownWrapperRef.current?.dispatchEvent(
```
riodeuno, line 957: Seems like the same code across all implementations.
I was wondering, if we can create a hook, this hook would return a `ref`, and a `trigger` function. Which can be called as and when needed, and the `ref` can be re-used if there already is a ref attached to the element.

What do you think?

app/client/src/components/propertyControls/BaseControl.tsx
```diff
@@ -10,9 +10,17 @@ import { CodeEditorExpected } from "components/editorComponents/CodeEditor";
 
 // eslint-disable-next-line @typescript-eslint/ban-types
 class BaseControl<P extends ControlProps, S = {}> extends Component<P, S> {
-  updateProperty(propertyName: string, propertyValue: any) {
+  updateProperty(
+    propertyName: string,
+    propertyValue: any,
+    isUpdatedViaKeyboard?: boolean,
```
riodeuno, line 16: What if we make this `isUpdatedViaKeyboard` required, instead of optional? 
Will it make the update property workflow more robust? Will it need a lot of changes?

riodeuno (commented):
I can see that the approach works. I am concerned about the modularity of this approach. There are a few points I would like for us to consider:

1. Separation of concerns: Can we separate the property control from the design system components, with the design system components, unaware if they're being used in property pane controls?

2. Reusable hooks: Some of these code blocks can be re-used in multiple controls, if we can create a hook, or create an HOC around the BaseControl. It will then be easier to maintain, as all a control developer would have to do is wrap the control in an HOC. I'm thinking similar to what we have with `withMeta` and `BaseWidget` implementations.

Let's think about this from the original goals' perspective. The goal is not to add the specific events we discussed, the goal is to add whatever event we can to get the required metrics. Basically, if you think the analytic events which we had discussed with @vuiets isn't going to go well with the feature design, we can consider adding other events or upgrading existing events towards the goal.

app/client/src/pages/Editor/PropertyPane/PropertyPaneView.tsx
```diff
@@ -47,6 +59,26 @@ function PropertyPaneView(
     return true;
   }, [widgetProperties?.type, excludeList]);
 
+  const handleKbdEvent = (e: Event) => {
```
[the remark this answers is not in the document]
aswathkk: Hooks make sense if this is a generic logic that could be used anywhere else. But in this case, it is only used for the PropertyPaneView component to capture keyboard analytics event.

app/client/src/components/propertyControls/BaseControl.tsx
aswathkk: Let me try doing that.

app/client/src/components/ads/Dropdown.tsx
aswathkk: I agree to abstract the logic to a hook. That makes sense.

Even I wanted to make the component agnostic of whether they are a part of the property pane. Unfortunately, there is no other way to get analytics data I guess.

How about changing the name from `propertyPaneKbdEvent` to `keyboardAnalyticsEvent`?

aswathkk:
@riodeuno 
Regarding separation of concerns:
Have a custom generic event `keyboardAnalyticsEvent` that is emitted from the DS component.
Capture this event in Property control and do whatever we want. What do you think about this?  

Regarding 2nd point, let me try to implement the logic into the BaseControl. And add a hook that could be used by the components to emit events.

app/client/src/components/propertyControls/BaseControl.tsx
aswathkk: This is done temporarily only for DropdownControl.
Once every other control is ready, we can make it required. Otherwise we will get errors.

riodeuno:
@aswathkk 

> Have a custom generic event keyboardAnalyticsEvent that is emitted from the DS component.

Yes, this makes sense to me. All our design system components can use a hook to do trigger events, which can be handled when it bubbles up. The reason for hooks, is once again separation of concerns, and making it DRY and re-usable.

app/client/src/components/ads/Switch.tsx
```diff
@@ -15,15 +16,31 @@ const StyledSwitch = styled(Switch)`
...
-    <StyledSwitch
-      {...props}
-      className={
-        props.className
-          ? props.className + " " + replayHighlightClass
-          : replayHighlightClass
-      }
-      tabIndex={0}
-    />
+    <div ref={wrapperRef}>
```
riodeuno, line 33: We can actually add a ref to the `StyledSwitch` in this case I believe. `[inputRef](https://blueprintjs.com/docs/versions/3/#core/components/switch)` is available in the Blueprint switch component. Please check it out, and see if you can re-use it. 

The reason for this is to reduce the number of DOM nodes we create. Every reduction helps.

app/client/src/utils/hooks/useInteractionAnalyticsEvent.tsx
```diff
@@ -0,0 +1,9 @@
+import { interactionAnalyticsEvent } from "utils/AppsmithUtils";
```
riodeuno, line 1: I think we should rename the file to be `ts` instead of `tsx`. Does typescript or React complain if we do so?

app/client/src/utils/PropertyControlRegistry.tsx
```diff
@@ -4,17 +4,77 @@ import {
...
+    };
+
+    static getControlType() {
+      return WrappedControl.getControlType();
+    }
+
+    render() {
+      return (
+        <div ref={this.containerRef}>
+          <WrappedControl {...this.props} />
```
riodeuno, line 60: In this case as well, we can pass the ref without create a new wrapping `div`. You can check with @ankurrsinghal about these refs for class components, he has been working extensively with refs for the dynamic height project.

app/client/src/components/ads/Switch.tsx
aswathkk: I tried that first. But, it wasn't working since it was expecting a callback ref instead.
Even the callback ref approach failed and that's why added a parent div here.

app/client/src/components/propertyControls/SwitchControl.tsx
```diff
@@ -1,22 +1,51 @@
...
+      </div>
     );
   }
 
   onToggle = () => {
-    this.updateProperty(this.props.propertyName, !this.props.propertyValue);
+    this.updateProperty(
+      this.props.propertyName,
+      !this.props.propertyValue,
+      this.isUpdatedViaKeyboard,
```
riodeuno, line 46: @aswathkk I did not clearly understand what you meant by attempting an HOC. Could you elaborate on the following:
1. What was your goal?
2. What is the ideal code flow you would like to have here, for the goal?

app/client/src/components/ads/Switch.tsx
aswathkk: Resolved this by adding a callback ref

app/client/src/utils/hooks/useInteractionAnalyticsEvent.tsx
aswathkk: Renamed to `ts` and started re-using the ref as you suggested

app/client/src/utils/PropertyControlRegistry.tsx
ankurrsinghal: Hey Abhinav, I believe `WrappedControl` here is an abstraction for all the different types of property control components, it cannot be a ForwardRef component but we have to add it to all these different property controls component, which might be a little complex.`

app/client/src/components/propertyControls/SwitchControl.tsx
aswathkk: Sorry. HOC is no longer needed since this is the only place we could use this logic.

aswathkk:
@riodeuno The DP doesn't have the logs starting with "Event fired locally".
How could @vuiets take a look at this DP in that case?

riodeuno:
@aswathkk I believe the loglevel localstorage key when set to "DEBUG" would show you the console logs in these scenarios. This is because, we use the `log.debug` function to log these entries on the console for debugging. (`AnalyticsUtil.tsx`).

The default value in our deploy previews for this is "ERROR" instead of "DEBUG".

We could do one of the following:
- Set `REACT_APP_CLIENT_LOG_LEVEL` env variable for this particular PR in the vercel configuration. Not sure if this can be easily achieved.
- Create a separate instance with the logging enabled so that @vuiets can review
- Change one line in this PR so that logging is enabled. This can either be done by using `log.error` or hard-coding `debug` as the value for the config in `configs/index.ts`. However, we will need to tread carefully here, as it is important to revert this change once it has been reviewed, and the change has to be that one specific line which disables the logging.

aswathkk:
@riodeuno Thanks for showing how this can be done.

Another hack that I found after going through the codebase is,
Just execute the following code snippet in the browser console to log the event:
```
window.analytics = {
    track: (name, data) => console.log("Event fired", name, data)
}
```
This is based on [AnalyticsUtil.tsx:349](https://github.com/appsmithorg/appsmith/blob/95e6b4fb9814af549a929e90d0ae43a077aa7de4/app/client/src/utils/AnalyticsUtil.tsx#L349)

riodeuno:
@aswathkk Sounds good, this is the third type of change, I recommended, where we need to revert it after review, before merging into release.

aswathkk:
@riodeuno No. We don't have to make this change in the codebase. Rather we just need to execute it on the browser console of the DP.

riodeuno:
Oh right, of course! Nice approach!

app/client/src/components/propertyControls/NumericInputControl.tsx
```diff
@@ -48,6 +49,9 @@ class NumericInputControl extends BaseControl<NumericInputControlProps> {
     return (
       <StyledNumericInput
         fill
+        inputRef={(elm) => {
```
rahulramesha, line 52: can we use `React.createRef ` instead?

app/client/src/components/propertyControls/SwitchControl.tsx
```diff
@@ -1,22 +1,51 @@
...
   render() {
     return (
-      <Switch
-        checked={this.props.propertyValue}
-        className={this.props.propertyValue ? "checked" : "unchecked"}
-        defaultChecked={this.props.propertyValue}
-        large
-        onChange={this.onToggle}
-      />
+      <div ref={this.containerRef}>
```
rahulramesha, line 30: NIT: Maybe we can `forwardRef` to `Switch` and then to Blueprint's `Switch` with `inputRef`

app/client/src/components/ads/ColorPickerComponentV2.tsx
```diff
@@ -178,10 +180,10 @@ function ColorPickerPopup(props: ColorPickerPopupProps) {
                     color === colorCode ? "ring-1" : ""
                   }`}
                   key={colorCode}
-                  onClick={() => {
+                  onClick={(e) => {
                     setColor(colorCode);
                     setIsOpen(false);
-                    changeColor(colorCode);
+                    changeColor(colorCode, !e.isTrusted);
```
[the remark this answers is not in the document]
aswathkk: From the [docs](https://dom.spec.whatwg.org/#dom-event-istrusted), it says that the value of `isTrusted` will be false when the click event is triggered using the `click()` method of the DOM.

In our case, when the user press `Enter` key, we are manually calling the `click()` method of this DOM element which makes the `e.isTrusted` to `false`.

I think this is a reliable behaviour and I couldn't find any other ways to make this is updated by keyboard.

app/client/src/components/propertyControls/NumericInputControl.tsx
aswathkk: @rahulramesha 
Do you mean, using `createRef` to create a ref
``` jsx
inputRef = React.createRef(null)
```
and pass it to `inputRef` prop?
``` jsx
<StyledNumericInput
  fill
  inputRef={inputRef}
```

If that's the case, the `StyledNumericInput`'s inputRef was only accepting a callbackRef

app/client/src/components/propertyControls/SwitchControl.tsx
aswathkk: blueprint's `Switch` component's `inputRef` accepts only callback refs.

app/client/src/components/propertyControls/NumericInputControl.tsx
rahulramesha: Got it. I only checked the docs, It said that the `inputRef` also accepted `RefObject` as well, But doesn't seem to accept it.

rahulramesha (dismissed):

vuiets:
Looks good to me. Thanks @aswathkk!

rahulramesha (approved):
Code looks good to me. But i'm just curious, why are we recording Tab events whenever a user tabs away from a propertyControl, that is whenever a user is navigating using tab, it records a tab event. Is this intentional?

aswathkk:
@rahulramesha We are tracking only the tab keypress happening within the property pane. Nothing outside the property pane.

rahulramesha:
> @rahulramesha We are tracking only the tab keypress happening within the property pane. Nothing outside the property pane.

yes, i understand that, just curious as to why, since it logs all navigation using tabs for propertyControls.

vuiets:
@rahulramesha we are trying to measure if a user is able to achieve their objective using keyboard navigation alone, for example if we want to we can now set up a funnel to see how many people used keyboard navigation to set the color of a button. If we don't log `Tab` events we would be able to say.

That being said do you see any performance issues in doing this?

rahulramesha:
> @rahulramesha we are trying to measure if a user is able to achieve their objective using keyboard navigation alone, for example if we want to we can now set up a funnel to see how many people used keyboard navigation to set the color of a button. If we don't log `Tab` events we would be able to say.
> 
> That being said do you see any performance issues in doing this?

Got it. No issues, was just curious.

riodeuno:
@aswathkk Could we rename the "ADS" into something else? Like "DesignSystem" or simply "DS".

app/client/src/components/editorComponents/ActionCreator/Fields.tsx
```diff
@@ -277,8 +277,12 @@ const views = {
             defaultText={props.defaultText}
             displayValue={props.displayValue}
             getDefaults={props.getDefaults}
-            onSelect={(value, defaultValue?: string) => {
-              props.set(value, defaultValue);
+            onSelect={(
```
riodeuno, line 280: I was wondering if we can simply say: `onSelect = {props.set}`?

riodeuno (commented):
@aswathkk A few comments. The rest looks fine to me.

app/client/src/components/ads/Dropdown.tsx
```diff
@@ -954,6 +973,7 @@ export default function Dropdown(props: DropdownProps) {
     (e: React.KeyboardEvent) => {
       switch (e.key) {
         case "Escape":
+          emitKeyboardAnalyticsEvent(dropdownWrapperRef.current, e.key);
```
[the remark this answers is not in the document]
riodeuno: I was considering the aspect that the parent should handle any generic event. 

For example, let's say the DS component exposes `handleDSEvent` prop. The implementing parent could check for the `type` or `meta` property, if it sees that the `meta.type` is a type of event which needs to log analytics, it can call the AnalyticUtils to log the event. In other cases, the events simply get ignored.

Does the above approach makes sense? Let me know if I'm missing context.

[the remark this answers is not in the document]
aswathkk: @riodeuno Yes. I understood.
But, that is what I'm doing here.

emitKeyboardAnalyticsEvent is just a local function in this file that abstracts the logic of adding event type as KEYBOARD_ANALYTICS.
Are you concerned about the name of this function?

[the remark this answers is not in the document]
riodeuno: Yes, this is a part of the issue. Another issue is that there is an event type `DSEventTypes.KEYBOARD_ANALYTICS` , the break in abstraction here is that the design system is not agnostic of the keyboard event analytics. 

Ideally, I wouldn't want the analytics related code to be in the DS component itself. I'd like for these to be defined and passed as a prop, where the prop is a function handler which takes any event of the `DSEventTypes` (provided by the DS component) as arguments. 

These events should be generic and something to be expected of any component in any design system. 
So, `DSEventTypes.KEYPRESS` could be valid, however, `DSEventTypes.KEYBOARD_ANALYTICS` may not be.

All of the above is with the understanding that `KEYBOARD_ANALYTICS` will not be useful to other users of this design system. If we anticipate a larger use-case, we can go ahead with the current implementation.

riodeuno (approved):

rohitagarwal88 (approved):

PR #13705, opened 2022-05-10
chore: Add isSystemGenerated field to analytics for git operations

### LINKED ISSUE #13477: [Task]: Add `isSystemGenerated` field in git-sync analytics events
## Description
We want to distinguish the events generated from Appsmith vs the user-generated events to optimize the mixpanel funnels.

AnaghHegde:
## Description

> We want to distinguish the events generated from Appsmith vs the user-generated events to optimize the mixpanel funnels.

Fixes #13477 

## Type of change

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -503,12 +509,14 @@ public Mono<String> commitApplication(GitCommitDTO commitDTO, String defaultAppl
...
                                 "up a git profile from the user profile section.";
+
                         return addAnalyticsForGitOperation(
                                 AnalyticsEvents.GIT_COMMIT.getEventName(),
                                 childApplication,
                                 AppsmithError.INVALID_GIT_CONFIGURATION.getErrorType(),
                                 AppsmithError.INVALID_GIT_CONFIGURATION.getMessage(errorMessage),
-                                childApplication.getGitApplicationMetadata().getIsRepoPrivate()
+                                childApplication.getGitApplicationMetadata().getIsRepoPrivate(),
+                                isSystemGenerated
```
abhvsn, line 519: Why are we adding the `isSystemGenerated` for error events?

abhvsn (changes requested):
As we are using the `isSystemGenerated` flag only in commit flow, does it make sense to keep 2 methods:
1. addAnalyticsForGitOperation(String eventName, Application application, Boolean isRepoPrivate, Boolean isSystemGenerated)
2. addAnalyticsForGitOperation(String eventName, Application application, Boolean isRepoPrivate)

This will prevent us from making larger change if this is only specific to certain flow

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: It is set false. But yeah i get what you are saying.

AnaghHegde:
@abhvsn Only commit is system generated action. All other cases are user generated. I would add a new method just for commit operation.

abhvsn:
@AnaghHegde can you please resolve the conflicts, I can start the review once this is done

AnaghHegde:
@abhvsn I have not made any changes. I want you to reply for the comments.

abhvsn:
> @abhvsn Only commit is system generated action. All other cases are user generated. I could add a new method just for commit operation. But if i do that then `isSystemGenerated` will be present for few cases and not at all present in other cases. What if i add one method where we set this only for commit operation. This way we would not have to deal with the problem of having to check if the value is present or not while building the charts in mixpanel.

Is this different from what I was expecting in above comment? Can you please elaborate how your approach will be different than what I was expecting here. The other part where you were asking for having `isSystemGenerated` boolean for all the events still can be achieved by specifically setting that variable in 
`addAnalyticsForGitOperation(String eventName, Application application, Boolean isRepoPrivate)`
https://github.com/appsmithorg/appsmith/pull/13705#pullrequestreview-970319295

AnaghHegde:
@abhvsn per your suggestion we would still be passing the default values to the method where we actually call the analytics service method. I am suggesting we would have a separate implementation just for the commit flow.

abhvsn (approved):

PR #13722, opened 2022-05-10
Added HA and PV

sub1983:
## Description
The main changes bought here are to implement a Horizontal POD Auto Scaler and the Persistent Volume.

-  Added hpa.yaml 
-   To enable autoscaling we have changed stateful set to deployment 
-  The Persistent volume has been added in the /appsmith-data path so now will be available on every pods
- the hpa uses metrics Server which needs to be setup
-  an efs-csi driver is used to setup the persistent  volume

### Setup metrics server run below 
`kubectl apply -f https://github.com/kubernetes-sigs/metrics-server/releases/latest/download/components.yaml '

###  To setup aws-efs-csi-driver

`helm repo add aws-efs-csi-driver https://kubernetes-sigs.github.io/aws-efs-csi-driver/
`

` helm upgrade --install aws-efs-csi-driver --namespace kube-system aws-efs-csi-driver/aws-efs-csi-driver -f efs/values.yaml `

Ref: [
](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/charts/aws-efs-csi-driver/values.yaml)
 Add the IAM role under the annotations of values.yaml 

Detailed Steps for aws-efs-csi-driver can be found in the [link](https://docs.aws.amazon.com/eks/latest/userguide/efs-csi.html)

## Checklist:

### DIFF
--- a/deploy/helm/templates/ingress.yaml
+++ b/deploy/helm/templates/ingress.yaml
@@ -1,14 +1,7 @@
 {{- if .Values.ingress.enabled -}}
 {{- $fullName := include "appsmith.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
-{{- if and .Values.ingress.className (not (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion)) }}
-  {{- if not (hasKey .Values.ingress.annotations "kubernetes.io/ingress.class") }}
-  {{- $_ := set .Values.ingress.annotations "kubernetes.io/ingress.class" .Values.ingress.className}}
-  {{- end }}
-{{- end }}
-{{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion -}}
-apiVersion: networking.k8s.io/v1
-{{- else if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
 apiVersion: networking.k8s.io/v1beta1
 {{- else -}}
 apiVersion: extensions/v1beta1
@@ -16,27 +9,17 @@ apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
   name: {{ $fullName }}
-  namespace: {{ include "appsmith.namespace" . }}
   labels:
     {{- include "appsmith.labels" . | nindent 4 }}
   {{- with .Values.ingress.annotations }}
   annotations:
     {{- toYaml . | nindent 4 }}
   {{- end }}
-    {{- if .Values.ingress.certManager }}
-    kubernetes.io/tls-acme: "true"
-    {{- end }}
 spec:
+  ingressClassName: {{ .Values.ingress.ingressClassName }}
   {{- if .Values.ingress.tls }}
   tls:
-    {{- range .Values.ingress.secrets }}
-    - hosts:
-        {{- range .hosts }}
-        - {{ . | quote }}
-        {{- end }}
-      secretName: {{ .secretName }}
-    {{- end }}
-    {{- range .Values.ingress.certManagerTls }}
+    {{- range .Values.ingress.tls }}
     - hosts:
         {{- range .hosts }}
         - {{ . | quote }}
@@ -45,38 +28,15 @@ spec:
     {{- end }}
   {{- end }}
   rules:
-    {{- if not .Values.ingress.hosts }}
-    - host:
-      http:
-        paths:
-          - path: /
-            pathType: Prefix
-            backend:
-              {{- if semverCompare ">=1.19-0" $.Capabilities.KubeVersion.GitVersion }}
-              service:
-                name: {{ $fullName }}
-                port:
-                  number: {{ $svcPort }}
-              {{- else }}
-              serviceName: {{ $fullName }}
-              servicePort: {{ $svcPort }}
-              {{- end }}
-    {{- end }}
     {{- range .Values.ingress.hosts }}
     - host: {{ .host | quote }}
       http:
         paths:
-          - path: /
-            pathType: Prefix
+          {{- range .paths }}
+          - path: {{ .path }}
             backend:
-              {{- if semverCompare ">=1.19-0" $.Capabilities.KubeVersion.GitVersion }}
-              service:
-                name: {{ $fullName }}
-                port:
-                  number: {{ $svcPort }}
-              {{- else }}
               serviceName: {{ $fullName }}
               servicePort: {{ $svcPort }}
-              {{- end }}
+          {{- end }}
     {{- end }}
-{{- end }}
+  {{- end }}
--- a/deploy/helm/templates/persistentVolume.yaml
+++ b/deploy/helm/templates/persistentVolume.yaml
@@ -12,8 +12,11 @@ spec:
   {{- range .Values.persistence.accessModes }}
     - {{ . | quote }}
   {{- end }}
-  persistentVolumeReclaimPolicy: Delete
-  storageClassName: {{ .Values.persistence.storageClass | quote }}
+  persistentVolumeReclaimPolicy: Retain
+  storageClassName: efs-pv
+  csi:
+    driver: efs.csi.aws.com
+    volumeHandle: fs-090c491a0740aa9e3
   local:
     path: {{ .Values.persistence.storagePath }} # Path to the directory this PV refers to.
   nodeAffinity: # nodeAffinity is required when using local volumes.
--- a/deploy/helm/values.yaml
+++ b/deploy/helm/values.yaml
@@ -1,43 +1,20 @@
-## @section Global parameters
-## Global Docker image parameters
-## Please, note that this will override the image parameters, including dependencies, configured to use the global value
-## Current available global Docker image parameters: imageRegistry, imagePullSecrets and storageClass
-##
 
-## @param global.storageClass Global StorageClass for Persistent Volume(s)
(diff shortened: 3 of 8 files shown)


sub1983:
@sharat87  Can you please review the PR and add your comments

deploy/helm/values.yaml
```diff
@@ -263,8 +265,10 @@ applicationConfig:
   APPSMITH_RECAPTCHA_SITE_KEY: ""
   APPSMITH_RECAPTCHA_SECRET_KEY: ""
   APPSMITH_RECAPTCHA_ENABLED: ""
-  APPSMITH_MONGODB_URI: ""
-  APPSMITH_REDIS_URL: ""
+  APPSMITH_MONGODB_URI: "mongodb+srv://appsmith:appsmith@appsmith.yj62f.mongodb.net/newappsmith12"
```
mohanarpit, line 268: Please don't commit credentials. This is now a security breach for us. :( 
I'd highly recommend that we destroy this MongoDB & Redis cluster **_immediately_**.

deploy/helm/templates/persistentVolume.yaml
```diff
@@ -12,8 +12,11 @@ spec:
   {{- range .Values.persistence.accessModes }}
     - {{ . | quote }}
   {{- end }}
-  persistentVolumeReclaimPolicy: Delete
-  storageClassName: {{ .Values.persistence.storageClass | quote }}
+  persistentVolumeReclaimPolicy: Retain
+  storageClassName: efs-pv
+  csi:
+    driver: efs.csi.aws.com
```
mohanarpit, line 18: Based on the value of the driver here, will this Helm chart work on another cloud provider like Azure or on minikube?

deploy/helm/values.yaml
sub1983: hey @mohanarpit  these are dummy values i created and is already destroyed after the testing and also these are not in our official account as well

deploy/helm/templates/persistentVolume.yaml
sub1983: Nope, @mohanarpit  the Persistent Volume confs we have to create separately depending on the storage type  as this was tested in AWS and we use EFS as storage for AWS this conf will  only work with AWS. 
The generic solution will be difficult as store type of each cloud provider is different. We can create different conf for widely used cloud providers like azure, gcp ..etc..

deploy/helm/values.yaml
sharat87: In such cases, please use dummy URIs, or non-URIs like `APPSMITH_MONGODB_URI: "valid-mongodb-uri-here"` or something of that sort.

deploy/helm/values.yaml
```diff
@@ -1,43 +1,20 @@
...
 
-## @param global.storageClass Global StorageClass for Persistent Volume(s)
-## @param global.namespaceOverride Override the namespace for resource deployed by the chart, but can itself be overridden by the local namespaceOverride
-##
-global:
-  storageClass: ""
-  namespaceOverride: ""
-## @param fullnameOverride String to fully override appsmith.fullname template
-##
-fullnameOverride: ""
```
sharat87, line 15: We seem to be removing a lot of variables in the default `values.yaml` file. Can you confirm if there really are not used anywhere and are safe to remove? For example, I see we are removing `global.namespaceOverride`, but the `_helpers.tpl` is still using it.

Secondly, what happens on existing installations? Ones that have already defined these values in their `values.yaml` and are running with that? Will thier installations break? If yes, can we avoid that? If no, can we minimize the damage and document how they can resolve please?

deploy/helm/values.yaml
```diff
@@ -1,43 +1,20 @@
...
-commonAnnotations: {}
-## @param schedulerName Name of the scheduler (other than default) to dispatch pods
-## ref: https://kubernetes.io/docs/tasks/administer-cluster/configure-multiple-schedulers/
-##
-schedulerName: ""
-## @param strategyType StrategyType for Appsmith&reg; statefulset
-## It can be set to RollingUpdate or Recreate by default.
-##
+
+useStatefulSet: false
```
sharat87, line 4: If statefulset is disabled, then won't the git-storage data be lost on pod recreation, or when there's an update to Appsmith?

deploy/helm/values.yaml
```diff
@@ -1,43 +1,20 @@
...
-## Image
-##
+
 image:
   registry: index.docker.io
   repository: appsmith/appsmith-ce
   pullPolicy: Always
   # Overrides the image tag whose default is the chart appVersion.
-  tag: "latest"
+  tag: "release"
```
sharat87, line 13: Please change this to `latest`, since this is the public Him charts. We want our self-hosted users to be on `latest`, unless they choose otherwise.

deploy/helm/templates/ingress.yaml
```diff
@@ -1,42 +1,25 @@
...
-{{- else if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
 apiVersion: networking.k8s.io/v1beta1
 {{- else -}}
 apiVersion: extensions/v1beta1
 {{- end }}
 kind: Ingress
 metadata:
   name: {{ $fullName }}
-  namespace: {{ include "appsmith.namespace" . }}
```
sharat87, line 19: I see that we are _adding_ this in other resource files and _removing_ it here? Shouldn't we stay consistent? Or is this file an exception?

deploy/helm/values.yaml
```diff
@@ -1,43 +1,20 @@
...
+## Namespace
+##
+namespace:
+  create: true
```
sharat87, line 17: I don't see anywhere that we are using this value. Should we remove it, if it's unused?

deploy/helm/values.yaml
```diff
@@ -99,17 +76,19 @@ service:
...
+  #kubernetes.io/ingress.class: "nginx"
     # cert-manager.io/cluster-issuer: "letsencrypt-prod"
     # nginx.ingress.kubernetes.io/ssl-redirect: "true"
     # nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
-  hosts: []
-    # - host: appsmith-domain.me
+  hosts:
+    - host: "a9175efee4fac4960bd75ea99b3d000b-08abe2a45de00e2c.elb.ap-south-1.amazonaws.com"
+      paths:
+      - path: "/"
```
sharat87, line 91: Can we remove this hard-coded hostname please?

deploy/helm/values.yaml
```diff
@@ -99,17 +76,19 @@ service:
 ingress:
   ## @param ingress.enabled Enable ingress record generation for Ghost
   ##
-  enabled: false
+  enabled: true
```
sharat87, line 79: We seem to be changing the default here? Would this be a breaking change for our users? If yes, can we do something to avoid breaking changes please?

deploy/helm/values.yaml
```diff
@@ -99,17 +76,19 @@ service:
...
   ## @param ingress.annotations Additional custom annotations for the ingress record
   ## NOTE: If `ingress.certManager=true`, annotation `kubernetes.io/tls-acme: "true"` will automatically be added
   ##
-  annotations: {}
-    # kubernetes.io/ingress.class: nginx
```
sharat87, line 107: This config seems to be still used in the `ingress.yaml`, can we not remove it please?

deploy/helm/values.yaml
```diff
@@ -161,11 +140,18 @@ resources:
...
-  #   memory: 128Mi
+
   limits: {}
-  requests: {}
+ # requests:
+  #   cpu: 100m
+   #  memory: 202Mi
+
+autoscaling:
+  enabled: true
```
sharat87, line 150: Please disable autoscaling by default.

deploy/helm/values.yaml
```diff
@@ -263,8 +265,10 @@ applicationConfig:
   APPSMITH_RECAPTCHA_SITE_KEY: ""
   APPSMITH_RECAPTCHA_SECRET_KEY: ""
   APPSMITH_RECAPTCHA_ENABLED: ""
-  APPSMITH_MONGODB_URI: ""
-  APPSMITH_REDIS_URL: ""
+  APPSMITH_MONGODB_URI: "mongodb+srv://appsmith:appsmith@appsmith.yj62f.mongodb.net/newappsmith12"
+  APPSMITH_REDIS_URL: "redis://subin-test-001.ihh5sj.0001.aps1.cache.amazonaws.com:6379"
```
sharat87, line 269: Please use a dummy URL here as well.

deploy/helm/templates/ingress.yaml
```diff
@@ -45,38 +28,15 @@ spec:
     {{- end }}
   {{- end }}
   rules:
-    {{- if not .Values.ingress.hosts }}
```
sharat87, line 48: Can you elaborate why we are removing this by default for our new users?

deploy/helm/templates/ingress.yaml
```diff
@@ -45,38 +28,15 @@ spec:
...
-    - host:
-      http:
-        paths:
-          - path: /
-            pathType: Prefix
-            backend:
-              {{- if semverCompare ">=1.19-0" $.Capabilities.KubeVersion.GitVersion }}
```
sharat87, line 55: I see these checks removed in a few places. Does this mean we won't be supporting some older versions of Kubernetes?

sharat87 (changes requested):
Lots of things here. Can you please take a look at the whole PR once and let me know for the next review please? Thanks

PR #13735, opened 2022-05-11
chore: Upgrade dependencies reported by Dependabot

sharat87:
Upgrades vulnerable dependencies in all plugins except for MySQL. That one is still failing and I'll fix it in a separate PR. Issue #14475


app/server/appsmith-plugins/mysqlPlugin/src/test/java/com/external/plugins/MySqlPluginTest.java
```diff
@@ -174,21 +162,6 @@ public void testConnectMySQLContainer() {
                 .verifyComplete();
     }
 
-    @Test
```
nidhi-nair, line 177: @sharat87 Are we removing these to get across the MySQL failures?

sharat87: Thanks for catching that. This file was not supposed to be committed.

sharat87: This is the error: https://github.com/appsmithorg/appsmith/runs/6420125481?check_suite_focus=true#step:7:2984

Do we know why this is happening? Or should we upgrade dependencies for the MySQL plugin in a separate PR and just move ahead with the others in this PR?

nidhi-nair: Yes let's split the MySQL upgrade. Last I remember this was happening because of a duplicate dependency. Using the same files fixed it on my local but not on the deployed version though.

sharat87: Already reverted changes to MySQL plugin's dependencies, but it still fails. I'll see what else I need to revert here. Marking this as draft until then.

nidhi-nair (approved):

PR #13739, opened 2022-05-11
fix: JSONForm - hidden field value reflects in formData

### LINKED ISSUE #13704: [Bug]: JSON form when a field is hidden do not show the field in formData

### Description

When I set a field to be hidden in the JSON form, the key and value still show up in formData and the value it shows is the older value when the field is visible. This creates a problem for the user because they are sending incorrect data.

### Steps To Reproduce

1. Add a new json form
2. Toggle visibility of a field with JS
3. Observe the key and value show up when the field is hidden

### Public Sample App

_No response_

### Version

Production - 1.6.24

ashit-rath:
## Description

When a field is hidden then the field's key and value reflect in `formData` which it should not as the field is hidden.

Fixes #13704 

## Type of change

## How Has This Been Tested?

Cypress test:
1. The following are done for all the field types present in JSONForm
  a. Field is first tested if visible and it exists in the `formData`
  b. It is hidden and tested if the field is hidden in the form and not present in the `formData`
  c. The visibility is turned back on and tested if the field is visible and the default value is present in the `formData` for that 
        field.
2. On load hidden field visiblity
  a. A normal field and an array field is hidden and published
  b. It is tested if the field is hidden in the form and not present in the `formData`

Jest test:
Extended the existing test case for the function that converts React Hook Form's internal form data to JSONForm's `formData` so that all the variations are tested when a field is set as hidden.

## Checklist:

### DIFF
--- a/app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/FormWidgets/JSONFormWidget/JSONForm_HiddenFields_spec.js
+++ b/app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/FormWidgets/JSONFormWidget/JSONForm_HiddenFields_spec.js
@@ -1,0 +1,223 @@
+const commonlocators = require("../../../../../locators/commonlocators.json");
+const dslWithSchema = require("../../../../../fixtures/jsonFormDslWithSchema.json");
+const widgetsPage = require("../../../../../locators/Widgets.json");
+
+const fieldPrefix = ".t--jsonformfield";
+const backBtn = ".t--property-pane-back-btn";
+
+function hideAndVerifyProperties(fieldName, fieldValue, resolveFieldValue) {
+  // Check if visible
+  cy.get(`${fieldPrefix}-${fieldName}`).should("exist");
+
+  // Hide field
+  cy.togglebarDisable(".t--property-control-visible input");
+
+  cy.wait(2000);
+
+  // Check if hidden
+  cy.get(`${fieldPrefix}-${fieldName}`).should("not.exist");
+
+  // Check if key in formData is also hidden
+  cy.get(`${widgetsPage.textWidget} .bp3-ui-text`).then(($el) => {
+    const formData = JSON.parse($el.text());
+    const formDataValue = resolveFieldValue
+      ? resolveFieldValue(formData)
+      : formData[fieldName];
+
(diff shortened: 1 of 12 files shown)


app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/FormWidgets/JSONFormWidget/JSONForm_HiddenFields_spec.js
```diff
@@ -0,0 +1,223 @@
...
+const backBtn = ".t--property-pane-back-btn";
+
+function hideAndVerifyProperties(fieldName, fieldValue, resolveFieldValue) {
+  // Check if visible
+  cy.get(`${fieldPrefix}-${fieldName}`).should("exist");
+
+  // Hide field
+  cy.togglebarDisable(".t--property-control-visible input");
+
+  cy.wait(2000);
```
sbalaji1192, line 15: Could you add a comment as to why we need 2 sec wait time

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/FormWidgets/JSONFormWidget/JSONForm_HiddenFields_spec.js
```diff
@@ -0,0 +1,223 @@
...
+    cy.wrap(formDataValue).should("be.undefined");
+  });
+
+  // Show field
+  cy.togglebar(".t--property-control-visible input");
+
+  // Check if visible
+  cy.get(`${fieldPrefix}-${fieldName}`).should("exist");
+
+  cy.wait(2000);
```
sbalaji1192, line 36: here too

app/client/cypress/support/widgetCommands.js
```diff
@@ -533,6 +533,16 @@ Cypress.Commands.add("openFieldConfiguration", (fieldIdentifier) => {
   cy.wait(1000);
 });
 
+Cypress.Commands.add("deleteField", (fieldIdentifier) => {
+  cy.get(
```
sbalaji1192, line 537: Seems like this is only being used by json form. could you update the command name to be more precise?

sbalaji1192 (approved):

PR #13935, opened 2022-05-18
feat: Add conceptual Tenant

### LINKED ISSUE #13528: [Task] Introduce Tenant Entity
- [ ] #13827
- [ ] #13828
- [ ] Create a default tenant for all instances and migrate the existing users and organizations to the default tenant. (Note : This change would also mean addition of tenant_id in the user object requiring the existing redis sessions to be cleared programmatically when this change goes through.)

trishaanand:
Fixes #13528


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
```diff
@@ -822,4 +832,84 @@ public void updateSequenceNamesFromOrganizationToWorkspace(MongockTemplate mongo
...
+        tenantQuery.addCriteria(where(fieldName(QTenant.tenant.slug)).is("default"));
+        Tenant tenant = mongockTemplate.findOne(tenantQuery, Tenant.class);
+
+        // if tenant already exists, don't create a new one.
+        if (tenant != null) {
+            return;
+        }
+
+        Tenant defaultTenant = new Tenant();
+        defaultTenant.setDisplayName("Appsmith");
```
mohanarpit, line 848: Can we rename this to "Default"? Makes it consistent with the slug name.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
```diff
@@ -822,4 +832,84 @@ public void updateSequenceNamesFromOrganizationToWorkspace(MongockTemplate mongo
...
+
+        Query workspaceQueryAll = new Query();
+        workspaceQueryAll.addCriteria(where(fieldName(QWorkspace.workspace.deleted)).is(FALSE));
+        workspaceQueryAll.addCriteria(where(fieldName(QWorkspace.workspace.deletedAt)).is(null));
+
+        //Memory optimization note:
+        //Call stream instead of findAll to avoid out of memory if the collection is big
+        //stream implementation lazy loads the data using underlying cursor open on the collection
+        //the data is loaded as and when needed by the pipeline
+        try(Stream<Workspace> stream = mongockTemplate.stream(workspaceQueryAll, Workspace.class)
```
mohanarpit, line 872: Instead of iterating over each workspace, can we write a single MongoDB query to update all workspace entries (deleted or not) with the tenantId?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
```diff
@@ -822,4 +832,84 @@ public void updateSequenceNamesFromOrganizationToWorkspace(MongockTemplate mongo
...
+
+        Query userQueryAll = new Query();
+        userQueryAll.addCriteria(where(fieldName(QUser.user.deleted)).is(FALSE));
+        userQueryAll.addCriteria(where(fieldName(QUser.user.deletedAt)).is(null));
+
+        //Memory optimization note:
+        //Call stream instead of findAll to avoid out of memory if the collection is big
+        //stream implementation lazy loads the data using underlying cursor open on the collection
+        //the data is loaded as and when needed by the pipeline
+        try(Stream<User> stream = mongockTemplate.stream(userQueryAll, User.class)
```
mohanarpit, line 897: Instead of iterating over each user, can we write a single MongoDB query to update all user entries (deleted or not) with the tenantId?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/TenantServiceCEImpl.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Override
+    public Mono<String> getDefaultTenantId() {
+        // If the default tenant id does not exist in cache, find from repository and set the cache.
+        if (defaultTenantIdCacheMono == null ) {
+            defaultTenantIdCacheMono = getDefaultTenantIdFromRepository();
+        }
+
+        return defaultTenantIdCacheMono
+                .onErrorResume(throwable -> {
```
mohanarpit, line 26: This error statement can be removed because we seem to be calling the same function again. Unless we know what error is thrown and can handle those errors in particular fashion, we shouldn't call the same function twice hoping for different results.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/TenantServiceCEImpl.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public TenantServiceCEImpl(TenantRepository tenantRepository) {
+        this.tenantRepository = tenantRepository;
+    }
+
+    @Override
+    public Mono<String> getDefaultTenantId() {
+        // If the default tenant id does not exist in cache, find from repository and set the cache.
+        if (defaultTenantIdCacheMono == null ) {
+            defaultTenantIdCacheMono = getDefaultTenantIdFromRepository();
```
mohanarpit, line 22: This mono variable doesn't seem to be cached. Are we confident that each invocation to this function will not trigger a DB call?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/TenantServiceCEImpl.java
```diff
@@ -0,0 +1,36 @@
...
+import com.appsmith.server.constants.FieldName;
+import com.appsmith.server.domains.Tenant;
+import com.appsmith.server.repositories.TenantRepository;
+import reactor.core.publisher.Mono;
+
+public class TenantServiceCEImpl implements TenantServiceCE {
+
+    private final TenantRepository tenantRepository;
+
+    private Mono<String> defaultTenantIdCacheMono = null;
```
sidhantgoel, line 12: Mono can't be cached, it is a single value publisher once published a Mono can not be reused, consider caching String instead or consider using CacheMono helper functions.

sidhantgoel (changes requested):
Caching of default tenant id needs little work, everything else looks good to me.

mohanarpit (approved):

PR #13947, opened 2022-05-18
fix: added healthcheck for editor with https

sum35h:
## Description
Added a fix for the docker health check on editor with ssl enabled for `appsmith-ce` fat container and the slim client `appsmith-editor`

    // Code coverage diff between base branch:release and head branch: FIX/health-check-https 

### DIFF
--- a/app/client/Dockerfile
+++ b/app/client/Dockerfile
@@ -17,5 +17,5 @@ COPY ./docker/templates/nginx-app-https.conf.template /nginx-app-https.conf.temp
 
 # This is the script that is used to start Nginx when the Docker container starts
 COPY ./docker/start-nginx.sh /start-nginx.sh
-HEALTHCHECK --interval=15s --timeout=15s --start-period=15s --retries=3 CMD curl -f http://localhost/ || curl -fk https://localhost/ || exit 1
+HEALTHCHECK --interval=15s --timeout=15s --start-period=15s --retries=3 CMD curl -Lfk http://localhost/ || exit 1
 CMD ["/start-nginx.sh"]
--- a/app/client/Dockerfile
+++ b/app/client/Dockerfile
@@ -1,3 +1,4 @@
+
 FROM nginx:1.20-alpine
 
 COPY ./build /var/www/appsmith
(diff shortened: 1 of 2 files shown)


pratapaprasanna (dismissed):

app/client/Dockerfile
```diff
@@ -16,5 +17,5 @@ COPY ./docker/templates/nginx-app-https.conf.template /nginx-app-https.conf.temp
 
 # This is the script that is used to start Nginx when the Docker container starts
 COPY ./docker/start-nginx.sh /start-nginx.sh
-HEALTHCHECK --interval=15s --timeout=15s --start-period=15s --retries=3 CMD curl -f http://localhost:80/ || exit 1
+HEALTHCHECK --interval=15s --timeout=15s --start-period=15s --retries=3 CMD curl -f http://localhost/ || curl -fk https://localhost/ || exit 1
```
sharat87, line 20: Do we need to check for both http and https? The http URL will redirect to HTTPS when available right? So, just adding `-L` to the http curl command will address this I think?

sum35h: Yes, that looks good. Wasn't aware of that option.

deploy/docker/scripts/healthcheck.sh
```diff
@@ -11,9 +11,11 @@ while read -r line
     else
       echo "PROCESS: $process - STATUS: $status"
       if [[ "$process" == 'editor' ]]; then
-        if [[ $(curl -s -w "%{http_code}\n" http://localhost:80/ -o /dev/null) -ne 200 ]]; then
+        if [[ $(curl -s -w "%{http_code}\n" http://localhost/ -o /dev/null) -ne 200 ]]; then
```
sharat87, line 14: Could we do the same thing here as well please? Using `-L`?

sharat87 (changes requested):

deploy/docker/scripts/healthcheck.sh
sum35h: @sharat87 I checked it with -L here, but the status code still comes to be 301.

sharat87 (approved):

pratapaprasanna (approved):

PR #13957, opened 2022-05-19
Feat: Add workflow to publish packer image

### LINKED ISSUE #14144: [Feature]: Automate the ami build creation

### Summary

For every appsmith release we release an ami image.

It would be great if we can have an automated pipeline for engineers for this tedious task.

It reduces the effort engineers put in for every appsmith-release.

pratapaprasanna:
## Description

For every appsmith-release  we release an ami image into aws and Digital Ocean to make life of users easy

Fixes #14144 

## Type of change

## How Has This Been Tested?
- Tested the pipeline.

### DIFF
--- a/.github/workflows/packer-ami-build.yml
+++ b/.github/workflows/packer-ami-build.yml
@@ -1,0 +1,67 @@
+
+---
+
+name: Ami Packer Build
+
+on:
+  pull_request:
+  push:
+    branches: ["release", "ci/packer-ansible"]
+
+jobs:
+  packer:
+    runs-on: ubuntu-latest
+    name: ami-packer-build
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ap-south-1
+
+      - name: Install python
+        uses: actions/setup-python@v3
+        with:
+          python-version: '3.8'
+
+      - name: Install ansible
+        run: pip install ansible
+      
+      - name: Install Packer
+        run: |
+          sudo apt-get update; sudo apt-get install -y curl gnupg software-properties-common ;
+          curl -fsSL https://apt.releases.hashicorp.com/gpg | sudo apt-key add - ;
+          sudo apt-add-repository -y "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main" ;
+          sudo apt-get update && sudo apt-get install -y packer;
+
+      - name: Build artifacts for Dev
+        if : |
+           github.event_name == 'pull_request'
+        run: |
+          cd deploy/packer;
+          packer build -var 'vpc_id=vpc-41b8bd29' -var 'subnet_id=subnet-26ac3f6a' template.json.pkr.hcl
+
+      - name: Build artifacts for Prod
+        # Run this job irrespective of tests failing, if this is the release branch;
+        # or only if the tests pass, if this is the master branch.
+        if: (success() && github.ref == 'refs/heads/master') ||
+          ( always() &&
+          (
+          github.event_name == 'workflow_dispatch' ||
+          github.event_name == 'push' ||
+          (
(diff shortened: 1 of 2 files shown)


.github/workflows/packer-ami-build.yml
```diff
@@ -0,0 +1,52 @@
...
+        run: |
+          sudo apt-get update; sudo apt-get install -y curl gnupg software-properties-common ;
+          curl -fsSL https://apt.releases.hashicorp.com/gpg | sudo apt-key add - ;
+          sudo apt-add-repository -y "deb [arch=amd64] https://apt.releases.hashicorp.com $(lsb_release -cs) main" ;
+          sudo apt-get update && sudo apt-get install -y packer;
+
+      - name: Build artifacts
+        run : |
+          cd deploy/packer;
+          packer build -var vpc_id=${{ secrets.VPC_DEV }} -var subnet_id=${{ secrets.DEV_SUBNET }} -var token=${{ secrets.DO_TOKEN }} template.json.pkr.hcl
```
mohanarpit, line 51: For my understanding, what do the `vpc_id` and `subnet_id` parameters do?

Also, do we need to do this for both dev & prod environments?

.github/workflows/packer-ami-build.yml
```diff
@@ -0,0 +1,52 @@
...
+
+on:
+  # This line enables manual triggering of this workflow.
+  workflow_dispatch:
+
+jobs:
+  packer:
+    if: |
+      github.event_name == 'workflow_dispatch' ||
+      github.event_name == 'push' ||
```
mohanarpit, line 14: Since this CI workflow is being triggered only on `workflow_dispatch` do we need these if-checks?

.github/workflows/packer-ami-build.yml
```diff
@@ -0,0 +1,52 @@
...
+
+    runs-on: ubuntu-latest
+    name: ami-packer-build
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v2
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
```
mohanarpit, line 27: We are configuring AWS credentials in the workflow, but I don't see the AWS CLI being used anywhere. Am I missing something?

pratapaprasanna: When we trigger packer pipeline. It creates aws resources(programmatically), for that purpose we need to configure our aws credentials.

.github/workflows/packer-ami-build.yml
pratapaprasanna: I am off the opinion that it would be useful if people can access ami in both dev as well as prod.
As it can help to create instances with ease.

Though, the ami can be copied over manually it feels like an UN-necessary work.

.github/workflows/packer-ami-build.yml
pratapaprasanna: Frankly, i felt we land in some situation as described in [here](https://github.com/appsmithorg/appsmith/blob/release/.github/workflows/test-build-docker-image.yml#L18) so added those lines

mohanarpit: Those if conditions were written because: 
1. There are more ways to trigger the job than just a manual run. This is specified [here](https://github.com/appsmithorg/appsmith/blob/release/.github/workflows/test-build-docker-image.yml#L8).
2. Earlier, this job used to run whenever a PR was approved. Those if conditions have remained even though we changed our process to use `/ok-to-test`. We should remove the PR approval events.

.github/workflows/packer-ami-build.yml
mohanarpit: Is this for publishing the AMI to multiple regions? If yes, then we publish our AMI in a lot of regions other than `ap-south-1` or `us-east-2`. 

Also, what is the variable `DO_TOKEN`? Is this the token for DigitalOcean? For my understanding, does this single Packer command build the machine images for both cloud providers?

pratapaprasanna: OK, then will create the ami in only region. whoever needs it will have to copy ami

sharat87 (approved):

.github/workflows/packer-ami-build.yml
sharat87: Same line, can we change `DO_TOKEN` to `DIGITALOCEAN_TOKEN` please? Took me a little while to realize that. :)

PR #13985, opened 2022-05-20
Fixes encryption handler

sidhantgoel:
> Pull Request Template
>

## Description

Fixes bug in EncryptionHandler that does not work with collection any other than List. This makes EncryptionHandler work with any collection.

Fixes # (issue)

## Type of change

## How Has This Been Tested?

- Test A
- Test B

## Checklist:


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/annotations/encryption/EncryptionHandler.java
```diff
@@ -207,7 +208,8 @@ List<CandidateField> findCandidateFieldsForType(@NonNull Object source) {
         
     }
 
-    synchronized boolean convertEncryption(Object source, Function<String, String> transformer) {
+    // Return value of convertEncryption does not accurately tell you whether there are encrypted properties or not but can be treated as good estimate
```
nidhi-nair, line 211: Hey could you please help me understand in what cases the estimation here would be off @sidhantgoel ?

sidhantgoel: When there is collection_unknown and map_unknown in fields it will return true while it may not be true.
1 question, what is need of making this function synchronized?
@nidhi-nair

nidhi-nair: It's so that the cache we're using does not keep recalculating the values during load time. It basically waits for when that source class type has a type defined and then uses that. We should may be look at improving the conditions for when the recalculation happens.

nidhi-nair (approved):

nidhi-nair:
Recording notes from convo @sidhantgoel 

Revert includes:
1. Introducing a safe path for unencrypted data to be ignored during transformation.
Reason: We should not introduce a code path that allows unencrypted data to flow and be stored in the DB. We should try to figure out how those unencrypted data points are getting stored and fix those paths instead.
2. Return data as null instead
Reason: This might lead to unexpected behaviour in code.
3. Unidentified behaviour where cache is returning incorrect type for empty set
Reason: This is not yet fixed, needs to be investigated.

PR #13996, opened 2022-05-20
feat: Return recently used templates first

### LINKED ISSUE #13920: Sort the list of templates based on recently used
The list of templates should be sorted based on the most recently forked templates. 

If user has no templates forked yet, it'll be sorted alphabetically
If user has recently forked two templates, those two will be in the beginning and then rest of the templates alphabetically.

nayan-rafiq:
## Description

If user has recently used templates set, those should come first in the response of get templates API

Fixes #13920

## Type of change

## How Has This Been Tested?

- JUnit test
- Manually

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
```diff
@@ -85,28 +85,40 @@ public Flux<ApplicationTemplate> getActiveTemplates(List<String> templateIds) {
...
-                    if (clientResponse.statusCode().equals(HttpStatus.OK)) {
-                        return clientResponse.bodyToFlux(ApplicationTemplate.class);
-                    } else if (clientResponse.statusCode().isError()) {
-                        return Flux.error(new AppsmithException(AppsmithError.CLOUD_SERVICES_ERROR, clientResponse.statusCode()));
-                    } else {
-                        return clientResponse.createException().flatMapMany(Flux::error);
-                    }
-                });
+        // To avoid compile error as applicationTemplateFlux will be updated inside another anonymous function later
+        var lambdaContext = new Object() {
```
abhvsn, line 89: Can we use CollectList on `applicationTemplateFlux` and return the Mono<List<ApplicationTemplate>>. This way we can change the index of the templates using the userData object and can control the sequence even in downstream chain.

abhvsn (commented):
LGTM just a minor comment on the way sorting is handled in the current implementation

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
nayan-rafiq: @abhvsn Thanks a lot for the suggestion. I've updated the code and it looks much more simpler now.

abhvsn (approved):
LGTM

PR #14120, opened 2022-05-26
chore: Enable serialisation for `JsonIgnore` fields during export flow to avoid adding intermediate fields in DTO

### LINKED ISSUE #11701: [Feature]: Implement intermediate DTOs or JsonView for serializing and deserializing DB objects to JSON files

### Summary

The way import-export is implemented has a single intermediate DTO `ApplicationJson` between the DB and the JSON file. This results in handling the edge cases for JsonIgnored fields from multiple collections confined to that single collection. 

More context can be found at: https://theappsmith.slack.com/archives/CPQNLFHTN/p1646717508835169

It becomes difficult to maintain the code, as there is no clear separation between the fields present in a single intermediate DTO. A possible solve for this is implementing the JsonView or separate DTOs for each collection which is a part of import-export.

abhvsn:
## Description

> As we have a single DTO to handle the conversion between DB objects and the serialised objects which will be saved in the exported JSON, it becomes difficult to maintain the conversion custom logic in special cases where we want to export the `JsonIgnored` fields. In such cases today we are writing a custom logic which can be found for fields like pageOrder, invisibleActionFields, mongoEscapedWidgetNames etc. In this PR we are trying to tackle this issue by serializing the complete object which also includes JsonIgnored fields and then sanitising the object needed for export.  

Fixes #11701 https://github.com/appsmithorg/appsmith/issues/14410

## Type of change

## How Has This Been Tested?

> JUnit testcase
> Manual testing

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationControllerCE.java
```diff
@@ -160,22 +157,15 @@ public Mono<ResponseDTO<Application>> forkApplication(
     }
 
     @GetMapping("/export/{id}")
-    public Mono<ResponseEntity<ApplicationJson>> getApplicationFile(@PathVariable String id,
-                                                                    @RequestHeader(name = FieldName.BRANCH_NAME, required = false) String branchName) {
```
AnaghHegde, line 164: I am not able to understand why we had to change return type to `Object`?

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/ApplicationJson.java
```diff
@@ -36,20 +40,26 @@ public class ApplicationJson {
...
-    List<String> publishedPageOrder = new ArrayList<>();
+    @Deprecated
+    List<String> publishedPageOrder;
 
+    @Deprecated
     String publishedDefaultPageName;
-    
+
+    @Deprecated
     String unpublishedDefaultPageName;
```
AnaghHegde, line 54: Why are we deprecating the defaultPageNames?

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/ApplicationJson.java
```diff
@@ -36,20 +40,26 @@ public class ApplicationJson {
...
     
     List<NewAction> actionList;
 
     List<ActionCollection> actionCollectionList;
 
+    // TODO remove the plain text fields during the export once we have a way to address sample apps DB authentication
     Map<String, DecryptedSensitiveFields> decryptedFields;
```
AnaghHegde, line 60: Thanks for adding this comment ❤️

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CompareDslActionDTO.java
```diff
@@ -10,9 +10,9 @@ public class CompareDslActionDTO implements Comparator<DslActionDTO>, Serializab
     // Method to compare DslActionDTO based on id
     @Override
     public int compare(DslActionDTO action1, DslActionDTO action2) {
-        if (action1 != null && !StringUtils.isEmpty(action1.getId())
-                && action2 != null && !StringUtils.isEmpty(action2.getId())) {
-            return action1.getId().compareTo(action2.getId());
```
AnaghHegde, line 15: Are we sure that just comparing by the names would not break the flow in other places? I am not sure if we just compare the actions by names is correct except the Import-Export flow.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/MigrationHelperMethods.java
```diff
@@ -0,0 +1,153 @@
+package com.appsmith.server.migrations;
+
```
AnaghHegde, line 2: Maybe i am wrong but looking at change shown in git diff, I have noticed that `HelperMethods.java` renamed to `MigrationHelperMethods.java`. It would be good if we just rename instead of deleting and creating a new file. This would help in maintaining the history.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -212,34 +214,9 @@ public Mono<ApplicationJson> exportApplicationById(String applicationId, Seriali
...
-                                .ifPresent(
-                                        publishedDefaultPage -> applicationJson.setPublishedDefaultPageName(publishedDefaultPage.getId())
-                                );
-                    }
-
                     // Refactor application to remove the ids
                     final String workspaceId = application.getOrganizationId();
-                    List<String> pageOrderList = application.getPages().stream().map(applicationPage -> applicationPage.getId()).collect(Collectors.toList());
-                    List<String> publishedPageOrderList = application.getPublishedPages().stream().map(applicationPage -> applicationPage.getId()).collect(Collectors.toList());
-                    removeUnwantedFieldsFromApplicationDuringExport(application);
```
AnaghHegde, line 242: Can you help me understand how are we handling the pageOrder ?

AnaghHegde (commented):
Minor nit, Rest Looks good to me.

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationControllerCE.java
abhvsn: This is to avoid removing JsonIgnored fields which will be the default behavior for Spring. We need this during client-side API calls but not in the exported files. Another option for conditional serializing is to use the JsonView but that was a hassle for maintenance, I have shared the experience in the issue comments attached to this PR.

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/ApplicationJson.java
abhvsn: As the application object will contain all this information this will not be required henceforth

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/MigrationHelperMethods.java
abhvsn: We have renamed the file and added few more methods to handle the migrations here.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
abhvsn: Application object is going to have all the details about pages like page order in published vs unpublished pages along with which page is default one in both views as domain objects will be serialised as is including the JsonIgnored fields henceforth.

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CompareDslActionDTO.java
abhvsn: We are using this compare only for import-export as in other places it doesn't matter in what sequence DSL actions will be fetched, which is why we are using `Set` instead of `List`. We need the DSL actions for git in specific sequence as it may result in uncommitted changes because of the git diff.

AnaghHegde (approved):

AnaghHegde (approved):

AnaghHegde (approved):

AnaghHegde (approved):

Issue #14123, opened 2022-05-26
[Task]: Add branch name param for git analytics events


abhvsn:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SubTasks

Must to have a branch field for commit application event as we want to track the activity at branch level.

PR #14133, opened 2022-05-27
chore: Add branch name param for git analytics events

abhvsn:
Fixes [#14123](https://github.com/appsmithorg/appsmith/issues/14123)

## Type of change

## How Has This Been Tested?

> Manual testing on segment


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -2424,19 +2424,20 @@ private Mono<Application> addAnalyticsForGitOperation(String eventName,
...
+        }
+        Map<String, Object> analyticsProps = new HashMap<>(Map.of(
+                FieldName.APPLICATION_ID, defaultApplicationId,
+                FieldName.ORGANIZATION_ID, defaultIfNull(application.getOrganizationId(), ""),
                 "branchApplicationId", defaultIfNull(application.getId(), ""),
                 "isRepoPrivate", defaultIfNull(isRepoPrivate, ""),
                 "gitHostingProvider", defaultIfNull(gitHostingProvider, ""),
-                "isSystemGenerated", defaultIfNull(isSystemGenerated, "")
+                "isSystemGenerated", defaultIfNull(isSystemGenerated, ""),
+                FieldName.BRANCH_NAME, branchName
```
AnaghHegde, line 2440: Can we have a null check here just like other fields?

AnaghHegde (commented):
Minor nit, rest LGTM.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
abhvsn: We are initialising `branchName` as empty string, so we don't need that here

AnaghHegde: I saw that, but what if the branchName is somehow empty in the db object itself? We have seen that happening in other cases, ex createdAt field being empty for few datasources.

abhvsn: In that case, this field will be null which will then be handled in `AnalyticsService.sendEvent`
```

// Segment throws an NPE if any value in `properties` is null.
        for (final Map.Entry<String, Object> entry : analyticsProperties.entrySet()) {
            if (entry.getValue() == null) {
                analyticsProperties.put(entry.getKey(), "");
            }
        }
```

Actually I was going to remove that check for other fields as well if that's OK

AnaghHegde: No still we would get NPE in the sentry as the value we are trying to access in GitServiceCeImpl class would be null. That's why i was keen on adding a defaultIfNull check. This way the sentry would not be bombarded with NPEs.

abhvsn: Can you explain I'm unable to understand how will we get NPE? We are doing the null check for `gitData` 🤔

AnaghHegde: Say the gitData is not null but the branchName inside the gitData is null. Now during the execution gitData.getBranchName is null and throws NPE. So i am suggesting to wrap the entire gitData.getBranchName inside the defaultIfNull.   https://github.com/appsmithorg/appsmith/pull/14133/files/642979aba6ff9d2300b5ae3792c4f8aa7ac03ff1#diff-0f8118156ce59d5121a19e466c70b2bccbe5f6c05f720c4f1eee056f519658b5R2430

abhvsn: Can you explain this bit `Now during the execution gitData.getBranchName is null and throws NPE.`?

AnaghHegde (dismissed):

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: My bad. I was looking at it wrong way.

abhvsn: @AnaghHegde above null check was causing issues for `Map.of` though it was handled in `sendEvent` method. I have made the necessary changes, please review the changes again

PR #14184, opened 2022-05-30
Feat: Perform file ops only when the resource is updated

### LINKED ISSUE #13825: [Task]: Instead of deleting all the files only delete which are removed and update the remaining

### LINKED ISSUE #13893: [Feature]: Perform file ops only when the resource is updated
Currently, we are deleting all the files in the repo and again recreating them from DB resources. This results in unwanted file operations even if the resource is not altered since the last commit. One way to solve this is by checking the updated at field for both corresponding file and the DB object and doing the selective file operations.

New Scope(02/06/22): Add test cases for helper class FileUtils inside appsmith-git.

vishnu-gp:
## Description

This PR optimises git-related file operations. With this, the resources like Pages, JsonObjects and Queries are only updated in the file system only if there is a change in the same available on the Database.

Git Status FS application save Execution times:

Before optimisation:
- First Load: 1650ms
- After first load: 500ms (Avg)

After Optimisation:
- First Load: 1115ms
- After first load: 200ms (Avg)

Fixes #13893 
Fixes #13825

## Type of change

## How Has This Been Tested?

- Manually via the local app

## Checklist:


app/server/appsmith-git/src/main/java/com/appsmith/git/helpers/FileUtilsImpl.java
```diff
@@ -265,6 +305,36 @@ private void scanAndDeleteFileForDeletedResources(Set<String> validResources, Pa
...
+     * deletion of respective resource from DB
+     * @param validResources resources those are still available in DB
+     * @param resourceDirectory directory which needs to be scanned for possible file deletion operations
+     */
+    private void scanAndDeleteDirectoryForDeletedResources(Set<String> validResources, Path resourceDirectory) {
+        // Scan resource directory and delete any unwanted directory if present
+        // unwanted directory : corresponding resource from DB has been deleted
+        try (Stream<Path> paths = Files.walk(resourceDirectory, 1)) {
+            paths
+                    .filter(path -> Files.isDirectory(path) && !path.equals(resourceDirectory) && !validResources.contains(path.getFileName().toString()))
```
abhvsn, line 319: Can we please get rid of contains as this may create issues when 1 of the resources is the substring of other?
e.g. page1 and page11

app/server/appsmith-git/src/main/java/com/appsmith/git/helpers/FileUtilsImpl.java
```diff
@@ -265,6 +305,36 @@ private void scanAndDeleteFileForDeletedResources(Set<String> validResources, Pa
...
+                    .forEach(this::deleteDirectory);
```
abhvsn, line 320: Not sure if I got this correctly but are we clearing the complete directory instead of single file. Consider we delete the `action1` from `page1`, so `resourceDirectory = {basePath}/page1/actions` what will happen if we have few more actions action2, action3 still attached to the page1?

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/GitFileUtils.java
```diff
@@ -158,6 +159,9 @@ public ApplicationGitReference createApplicationReference(ApplicationJson applic
                     String pageName = newPage.getUnpublishedPage() != null
                             ? newPage.getUnpublishedPage().getName()
                             : newPage.getPublishedPage().getName();
+                    if(newPage.getUpdatedAt() != null){
+                        updatedResources.get(PAGE_LIST).put(pageName, true);
+                    }
```
abhvsn, line 164: Not able to understand the implementation here, can you please elaborate?

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -237,6 +237,7 @@ public Mono<ApplicationJson> exportApplicationById(String applicationId, Seriali
 
                     // Refactor application to remove the ids
                     final String workspaceId = application.getOrganizationId();
+                    Instant applicationLastCommittedAt = application.getGitApplicationMetadata().getLastCommittedAt();
```
abhvsn, line 240: This may end up in NPE when application is not connected to git

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -299,6 +300,11 @@ public Mono<ApplicationJson> exportApplicationById(String applicationId, Seriali
                                         }
                                     }
                                     newPage.setApplicationId(null);
+                                    Instant newPageUpdatedAt = newPage.getUpdatedAt();
+                                    boolean isNewPageNotUpdated = applicationLastCommittedAt != null && newPageUpdatedAt != null && applicationLastCommittedAt.isAfter(newPageUpdatedAt);
+                                    if(isNewPageNotUpdated){
+                                        newPage.setUpdatedAt(null);
+                                    }
```
abhvsn, line 307: Should we create another field `Set<String> updatedResource` to list all the updated resources since last commit and pass this to `saveFile` method to avoid the dependency on updatedAt field. Asking this because we are not committing the updatedAt field and devs can make the mistake of removing this field during the export flow

abhvsn (changes requested):

app/server/appsmith-git/src/main/java/com/appsmith/git/helpers/FileUtilsImpl.java
vishnu-gp: @abhvsn The idea of this method is to delete non-existing directory objects. Basically deletes directories inside the resourceDirectory which are not in the validResource list. The only use case we have now is deleting the page directory if the page is deleted. In that case, resourceDirectory would be {basePath}/pages

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/GitFileUtils.java
vishnu-gp: @abhvsn At this point, if the page resource has a non-null updatedAt field means it has updates after the last commit. The logic for this can be found in ImportExportApplicationService.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
vishnu-gp: @abhvsn We can have a different boolean field to avoid dependency on updatedAt but then removing that field from being committed can be the same problem. A list also works, the current `Map<String, Map<String, Boolean>> updatedResources;` is a map of all the updated resources only which would take care of only invoking saveFile() for updated resources only.

app/server/appsmith-git/src/main/java/com/appsmith/git/helpers/FileUtilsImpl.java
abhvsn: Does this mean we are not deleting the files selectively? If yes then can we please modify the implementation accordingly?

app/server/appsmith-git/src/main/java/com/appsmith/git/helpers/FileUtilsImpl.java
vishnu-gp: @abhvsn This checks and matches the exact directory name in the Set and I don't think the substring issue will happen here.

app/server/appsmith-git/src/main/java/com/appsmith/git/helpers/FileUtilsImpl.java
abhvsn: This looks sorted as we are using this to remove the page directory when the page is removed from the application.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
vishnu-gp: @abhvsn The latest commit would solve the dependency on updatedAt and addresses the other issues mentioned here.

app/server/appsmith-git/src/main/java/com/appsmith/git/helpers/FileUtilsImpl.java
```diff
@@ -184,18 +190,33 @@ public Mono<Path> saveApplicationToGitRepo(Path baseRepoSuffix,
                         if (names.length > 1 && StringUtils.hasLength(names[1])) {
                             // For actions, we are referring to validNames to maintain unique file names as just name
                             // field don't guarantee unique constraint for actions within JSObject
+                            Boolean isResourceUpdated = updatedResources.get(ACTION_LIST).contains(names[0])  ;
```
abhvsn, line 193: What happens when we have multiple actions with the same name on different pages?

app/server/appsmith-git/src/main/java/com/appsmith/git/helpers/FileUtilsImpl.java
```diff
@@ -205,14 +226,27 @@ public Mono<Path> saveApplicationToGitRepo(Path baseRepoSuffix,
...
-                                    resource.getValue(),
-                                    pageSpecificDirectory.resolve(ACTION_COLLECTION_DIRECTORY).resolve(actionCollectionName + CommonConstants.JSON_EXTENSION),
-                                    gson
-                            );
+
+                            if(!validActionCollectionsMap.containsKey(pageName)) {
+                                validActionCollectionsMap.put(pageName, new HashSet<>());
+                            }
+                            validActionCollectionsMap.get(pageName).add(actionCollectionName + CommonConstants.JSON_EXTENSION);
+                            Boolean isResourceUpdated = updatedResources.get(ACTION_COLLECTION_LIST).contains(actionCollectionName);
```
abhvsn, line 234: Same as above.

app/server/appsmith-git/src/test/java/com/appsmith/git/helpers/FileUtilsImplTest.java
```diff
@@ -0,0 +1,76 @@
...
+        fileUtils = new FileUtilsImpl(gitServiceConfig, gitExecutor);
+    }
+
+    private static final String localDirectory = "localTestDirectory";
+    private static final Path localDirectoryPath = Path.of(localDirectory);
+
+    @Test
+    public void testScanAndDeleteDirectoryForDeletedResources() {
+        String pageDirectory = localDirectory + File.separator +"pages";
+        Path pageDirectoryPath = Path.of(pageDirectory);
```
abhvsn, line 36: Can we please use the Path to create the directory path to be consistent?

app/server/appsmith-git/src/test/java/com/appsmith/git/helpers/FileUtilsImplTest.java
```diff
@@ -0,0 +1,76 @@
...
+            Set<String> validFSDirectoryList = paths
+                    .filter(path -> Files.isDirectory(path) && !path.equals(pageDirectoryPath))
+                    .map(Path::getFileName)
+                    .map(Path::toString)
+                    .collect(Collectors.toSet());
+            Assertions.assertEquals(validDirectoryList, validFSDirectoryList);
+        } catch (IOException e) {
+            Assertions.fail("Error while scanning directory");
+        }
+    }
```
abhvsn, line 75: Let's include the testcase for `scanAndDeleteFileForDeletedResources` as well

vishnu-gp: @abhvsn Done

app/server/appsmith-git/src/main/java/com/appsmith/git/helpers/FileUtilsImpl.java
vishnu-gp: @abhvsn corrected

app/server/appsmith-server/src/main/java/com/appsmith/server/constants/FieldName.java
```diff
@@ -111,4 +111,5 @@ public class FieldName {
     public static final String THEME = "theme";
     public static final String EDIT_MODE_THEME = "editModeTheme";
     public static final String FLOW_NAME = "flowName";
+    public static final String NAME_SEPARATOR = "##ENTITY_SEPARATOR##";
```
abhvsn, line 114: Please use the `com/appsmith/external/constants/GitConstants.java` so as to avoid keeping the same copy in multiple places.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -395,7 +396,7 @@ public Mono<ApplicationJson> exportApplicationById(String applicationId, Seriali
                                 actionCollections.forEach(actionCollection -> {
                                     ActionCollectionDTO publishedActionCollectionDTO = actionCollection.getPublishedCollection();
                                     ActionCollectionDTO unpublishedActionCollectionDTO = actionCollection.getUnpublishedCollection();
-                                    String actionCollectionName = publishedActionCollectionDTO != null ? publishedActionCollectionDTO.getName() : unpublishedActionCollectionDTO != null ? unpublishedActionCollectionDTO.getName() : null;
+                                    String actionCollectionName = unpublishedActionCollectionDTO != null ? unpublishedActionCollectionDTO.getName() + NAME_SEPARATOR + unpublishedActionCollectionDTO.getPageId() : publishedActionCollectionDTO != null ? publishedActionCollectionDTO.getName() + NAME_SEPARATOR + publishedActionCollectionDTO.getPageId() : null;
```
abhvsn, line 399: It will be great if we can make this more readable 🙂

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -477,7 +478,7 @@ public Mono<ApplicationJson> exportApplicationById(String applicationId, Seriali
                                         invisibleActionFieldsMap.put(newAction.getId(), invisibleActionFields);
                                     }
 
-                                    String newActionName = unpublishedActionDTO != null ? unpublishedActionDTO.getValidName() : publishedActionDTO != null ? unpublishedActionDTO.getValidName(): null;
+                                    String newActionName = unpublishedActionDTO != null ? unpublishedActionDTO.getValidName() + NAME_SEPARATOR + unpublishedActionDTO.getPageId() : publishedActionDTO != null ? publishedActionDTO.getValidName() + NAME_SEPARATOR + publishedActionDTO.getPageId(): null;
```
abhvsn, line 481: Same here

abhvsn:
LGTM some minor nits.
@vishnu-gp can you also resolve the merge conflicts.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
vishnu-gp: @abhvsn Done

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
vishnu-gp: @abhvsn Done

abhvsn (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/ApplicationJson.java
```diff
@@ -48,6 +48,14 @@ public class ApplicationJson {
 
     List<ActionCollection> actionCollectionList;
 
+    /**
+     * This field will be used to store map of files to be updated in local file system by comparing the recent
+     * changes in database and the last local git commit.
+     * This field can be used while saving resources to local file system and only update the resource files which
+     * are updated in the database.
+     */
+    Map<String, Set<String>> updatedResources;
```
abhvsn, line 57: @vishnu-gp as this should not be included in the exported file let's JsonIgnore this field.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -242,6 +242,8 @@ public Mono<ApplicationJson> exportApplicationById(String applicationId, Seriali
                     final String workspaceId = application.getOrganizationId();
                     GitApplicationMetadata gitApplicationMetadata = application.getGitApplicationMetadata();
                     Instant applicationLastCommittedAt = gitApplicationMetadata != null ? gitApplicationMetadata.getLastCommittedAt() : null;
+                    boolean isClientSchemaMigrated = application.getClientSchemaVersion() != JsonSchemaVersions.clientVersion;
+                    boolean isServerSchemaMigrated = application.getServerSchemaVersion() != JsonSchemaVersions.serverVersion;
```
abhvsn, line 246: Just to be consistent with rest of the implementation can we change this to 
`!JsonSchemaVersions.serverVersion.equals(application.getServerSchemaVersion());`

vishnu-gp: @abhvsn Done

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/ApplicationJson.java
vishnu-gp: @abhvsn Done

abhvsn (approved):

PR #14340, opened 2022-06-07
chore: Update get ssh key api response

### LINKED ISSUE #14339: [Task]: Update the response for get ssh key response

### SubTasks

- When the ssh key is not empty for an app instead of throwing the error state we should send other datapoints like supported key types, reference doc url as there are still need by FE
- Update the response code from 201 to 200 as this is a GET request.

AnaghHegde:
## Description

> When the ssh key is not empty for an app instead of throwing the error state we should send other datapoints like supported key types, reference doc url as there are still need by FE

Fixes #14339 

## Type of change

## How Has This Been Tested?

- Locally
- Junit tests

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1123,12 +1123,18 @@ public Mono<Application> createBranch(String defaultApplicationId, GitBranchDTO
...
                     }
                     return applicationService.getSshKey(gitData.getDefaultApplicationId())
-                            .map(gitAuthDTO -> {
+                            .flatMap(gitAuthDTO -> {
+                                if(StringUtils.isEmptyOrNull(gitAuthDTO.getPublicKey())) {
+                                    return Mono.error(new AppsmithException(
+                                            AppsmithError.INVALID_GIT_CONFIGURATION,
+                                            "Can't find valid SSH key. Please configure the application with git"
+                                    ));
+                                }
```
abhvsn, line 1132: 1. How can we end up in this scenario?
2. Also this seems like a SSH key config error can we please replace this with INVALID_GIT_SSH_CONFIGURATION in that case?

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ApplicationServiceTest.java
```diff
@@ -2886,4 +2886,22 @@ public void getApplicationConnectedToGit_defaultBranchUpdated_returnBranchSpecif
...
+        app.setOrganizationId(orgId);
+        app.setName("getSshKey_notConnectedToGit_SuccessResponseWithoutSshKey");
+        Application application = applicationPageService.createApplication(app).block();
+
+        StepVerifier
+                .create(applicationService.getSshKey(application.getId()))
+                .assertNext(gitAuthDTO -> {
+                    assertThat(gitAuthDTO.getDocUrl()).isEqualTo(Assets.GIT_DEPLOY_KEY_DOC_URL);
+                    assertThat(gitAuthDTO.getGitSupportedSSHKeyType()).isNotEmpty();
+                    assertThat(gitAuthDTO.getPublicKey()).isNull();
```
abhvsn, line 2903: If I understood correctly, this should be empty and not null.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/GitServiceTest.java
```diff
@@ -2622,7 +2622,7 @@ public void deleteBranch_defaultBranch_ThrowError() throws IOException, GitAPIEx
     @Test
     @WithUserDetails(value = "api_user")
     public void deleteBranch_defaultBranchUpdated_Success() throws IOException, GitAPIException {
-        Application application = createApplicationConnectedToGit("deleteBranch_defaultBranchUpdated_Success", "master");
+        Application application = createApplicationConnectedToGit("deleteBranch_defaultBranchUpdated_Successw", "master");
```
abhvsn, line 2625: Please revert this change.

abhvsn (changes requested):
Minor comments rest LGTM

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationControllerCE.java
```diff
@@ -198,7 +198,7 @@ public Mono<ResponseDTO<GitAuth>> generateSSHKeyPair(@PathVariable String applic
     @GetMapping("/ssh-keypair/{applicationId}")
     public Mono<ResponseDTO<GitAuthDTO>> getSSHKey(@PathVariable String applicationId) {
         return service.getSshKey(applicationId)
-                .map(created -> new ResponseDTO<>(HttpStatus.CREATED.value(), created, null));
+                .map(created -> new ResponseDTO<>(HttpStatus.OK.value(), created, null));
```
IAmAnubhavSaini, line 201: 👍

IAmAnubhavSaini (commented):
There's a typo as Abhijeet mentioned, rest looks ok to me.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/GitServiceTest.java
AnaghHegde: This is on purpose because of the duplicate key error in test class.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: I have added this check because the same was present in getSShKey of `ApplicationServiceClass` and now we do not throw the error for the same case

abhvsn: I am still of the opinion we will not end up in this error state because we are making changes in `CreateBranch` flow, but there is no harm I suppose. Can we use the standard SSH key error response as this is specifically related to SSH keys?

abhvsn (approved):

abhvsn (dismissed):

AnaghHegde:
@IAmAnubhavSaini Can you try this in local env? Because the cypress tests are failing for the same reason i have mentioned above. Key gen step is failing in cypress tests.

abhvsn (approved):

abhvsn (approved):

abhvsn (approved):

IAmAnubhavSaini:
This is causing a frontend issue after implementation. API is returning an empty string, front end is dependent upon error when ssh-keypair is not generated; so for the time being front-end is going to save the protocols and the learn more link above ssh-key input.

Closing for now.

PR #14374, opened 2022-06-08
fix: Garbled characters in API response (#13515)

### LINKED ISSUE #13515: [Bug]-[1600]:REST API response with charset=windows-1253 - blank in query pane, garbled in evaluated value

### Description

Creating this issue on behalf of a user. The user is calling a REST API endpoint on Appsmith and Postman:

1. **On Appsmith**
    a. on clicking the run button, the response status is 200 (sucess), but the response tab is otherwise empty
    b. when the result of the query is bound to a widget, the widget's content and the evaluated value of the result both show garbled characters, like this:
Screenshot 2022-04-29 at 4 08 39 PM

2. **On postman**
    a. the user is able to see a response
    b. the header includes content type as `application/json;charset = windows-1253`
    c. the response also includes both english and greek characters

a snippet from the response:
"objs": [
        {
            "COMPANY": "sensitive(was a number)",
            "COMPANYNAME": "retracted, english characters",
            "BRANCH": "sensitive(was a number)",
            "BRANCHNAME": "Εδρα",
            "MODULE": "0",
            "MODULENAME": "Χρήστης",
            "REFID": "sensitive(was number)",
            "REFIDNAME": "WEB",
            "USERID": "sensitive(was a number)",
            "FINALDATE": "",
            "ROLES": "",
            "XSECURITY": "",
            "EXPTIME": ""
        }
    ],
    "ver": "5.00.521.11454",
    "sn": "sensitive(was a number)",
    "off": false,
    "pin": false,
    "appid": "sensitive(was a number)"

### Steps To Reproduce

NA - we have yet to set up an example service that returns such a response

### Public Sample App

_No response_

### Version

self hosted

subrata71:
## Description
The issue has two variants.

- Currently `content-type` matching for `application/json` type is done by the following code snippet

```
  if (MediaType.APPLICATION_JSON.equals(contentType) ||
      MediaType.APPLICATION_JSON_UTF8.equals(contentType)) {
	  //Do something awesome
  }
```

When the `contentType` variable holds a value like `application/json;charset=windows-1253` then it doesn’t satisfy any of the above clauses. Thus it falls into the default case which is `TEXT`. As a result, the response being a `JSON` type the client will treat the data as `TEXT`.

- If an API has a `Content-Encoding` header then Appsmith doesn’t decode the response before rendering it to the client. As a result, the client sees garbled data.

Read [RCA](https://www.notion.so/appsmith/RCA-Garbled-Characters-in-API-Response-4f00ddb75cd74fb5b5e8987d552eac81)

Fixes #13515 

## Type of change

## How Has This Been Tested?

- Manual
- JUnit

## Checklist:


app/server/appsmith-plugins/restApiPlugin/src/main/java/com/external/plugins/RestApiPlugin.java
```diff
@@ -460,8 +460,7 @@ public Mono<ActionExecutionResult> executeCommon(APIConnection apiConnection,
                              * Handle XML response. Currently we only handle JSON & Image responses. The other kind of responses
                              * are kept as is and returned as a string.
                              */
-                            if (MediaType.APPLICATION_JSON.equals(contentType) ||
-                                    MediaType.APPLICATION_JSON_UTF8.equals(contentType)) {
+                            if (contentType.includes(MediaType.APPLICATION_JSON)) {
```
sumitsum, line 463: can you please add a check `contentType != null` or check manually if such a condition gets handled elsewhere ?

sumitsum:
Can you please add a JUnit test case for this ?

app/server/appsmith-plugins/restApiPlugin/src/main/java/com/external/plugins/RestApiPlugin.java
subrata71: >can you please add a check contentType != null or check manually if such a condition gets handled elsewhere ?

This `contentType` variable is set from the API response header. There's no `null` checking on this header. But isn't the `content-type` header supposed to exist always in the response header? 

Anyway, if I have to add `null` checking on this variable then what course of action should be taken in case it's `null`?

sumitsum: @subrata71 in case it is null, then we should proceed assuming a default for contentType.

subrata71: I have made the necessary changes. Can you please check @sumitsum ?

subrata71:
> Can you please add a JUnit test case for this ?

I have added a unit test. Can you please check @sumitsum ?

app/server/appsmith-plugins/restApiPlugin/src/test/java/com/external/plugins/RestApiPluginTest.java
```diff
@@ -1031,5 +1031,31 @@ public void testGetApiWithBody() {
                 })
                 .verifyComplete();
     }
+
```
subrata71, line 1034: ## Test case API Details
The test case written for this fix consumes the following API.

API Endpoint: http://postman-echo.com/gzip
Request Method: `GET`
Response Headers: `Content-Encoding: gzip`
Response Body:
```
{
    "gzipped": true,
    "headers": {
        "x-forwarded-proto": "http",
        "x-forwarded-port": "80",
        "host": "postman-echo.com",
        "x-amzn-trace-id": "Root=1-62a6cfd3-7d5423a350e1717264d17c8d",
        "user-agent": "PostmanRuntime/7.29.0",
        "accept": "*/*",
        "postman-token": "db82e216-d8db-4cdf-a53a-5ecb61c2cf7f",
        "accept-encoding": "gzip, deflate, br"
    },
    "method": "GET"
}
```

## Validating TC 
Before this fix, this API response would appear to be garbled on the client-side due to the fact the response is encoded in `gzip` format. 

After this fix, the response appears after being decoded by the server-side. Since the response is decoded so the client will be able to read the response properly. In this test case, I tried to read just one property from the response i.e. `"gzipped" : true`

sumitsum (approved):

sumitsum (approved):

btsgh:
Tested this fix on a local environment (dev.appsmith.com) 
For the test apis provided by @subrata71  - there are no garbled characters in the error message nor in the response. 
Also verified with a few other apis from the site - https://apipheny.io/free-api/. Compared the APIs behavior with release and Postman, and found that all are working fine.

PR #14427, opened 2022-06-09
feat: All outgoing requests should go through the configured proxy

### LINKED ISSUE #11749: [Feature]: Add support for Appsmith to connect through a http or https proxy server

### Summary

When the user self hosts Appsmith in their network, there is a very high chance that all network calls to the outside world will be blocked. Usually the outside/internet calls are done via proxy server which is used only for the purpose of interacting with the outside world. 

This feature lets the users to use features which are dependent on Cloud Service such as OAuth, GitSync, Licencing etc without them having to whitelist ip/domain of Cloud Service. 

We use cloud service for a lot of activities now a day. When the user self hosts appsmith in their network, there is a very high chance that all network calls to the outside world will be blocked. Usually the outside/internet calls are done via proxy server which is used only for the purpose of interacting with the outside world. With this support we can support such requirements from users.

sharat87:
Fixes #11749

## To test

1. Install `mitmproxy` with `brew install mitmproxy`.
2. Start the proxy with `mitmweb --listen-port 9020 --web-port 9021`.
4. Start Appmith with:
   ```
   docker run --name ace -p 8001:80 -v ~/appsmith-stacks:/appsmith-stacks -d --pull always -e HTTP_PROXY=http://host.docker.internal:9020 -e HTTPS_PROXY=http://host.docker.internal:9020 appsmith/appsmith-ce:release
   ```
5. Now launch Appsmith at http://localhost:8001, load the templates page, and check that a request to load templates from cloud-services shows up in the proxy's web UI.
6. Run any API request in Appsmith and see that that also appears in the proxy's web UI.
7. Try connect a git repo and check that git related queries are also made via the proxy.


app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/EnvManagerCEImpl.java
```diff
@@ -159,12 +159,54 @@ public List<String> transformEnvContent(String envContent, Map<String, String> c
...
 
         for (final String name : remainingChangedNames) {
-            outLines.add(name + "=" + changes.get(name));
+            outLines.add(name + "='" + changes.get(name).replace("'", "'\\''") + "'");
         }
 
         return outLines;
     }
 
+    /*
```
trishaanand, line 168: @sharat87 : Left this change by mistake?

sharat87: Of course I did. I really need to stop doing multiple things at once. 😖

sharat87: Resolved, thanks for the find.

app/server/appsmith-interfaces/src/main/java/com/appsmith/util/WebClientUtils.java
```diff
@@ -0,0 +1,39 @@
...
+                .baseUrl(baseUrl)
+                .build();
+    }
+
+    private static boolean shouldUseSystemProxy() {
+        return "true".equals(System.getProperty("java.net.useSystemProxies"))
+                && (!System.getProperty("http.proxyHost", "").isEmpty() || !System.getProperty("https.proxyHost", "").isEmpty());
+    }
+
+    public static WebClient.Builder builder() {
```
nidhi-nair, line 28: Hey why have we left REST API out of this? @sharat87

sharat87: Rest API plugin also uses the same piece of code, but is duplicated. I couldn't combine the two APIs to work togather, and have put it aside. Let me see if I can do it without making the implementation too messy.

sharat87: Done, thanks. :)

app/server/appsmith-interfaces/src/main/java/com/appsmith/util/WebClientUtils.java
```diff
@@ -0,0 +1,51 @@
...
+        HttpClient httpClient = applyProxyIfConfigured(HttpClient.create());
+
+        if (shouldUseSystemProxy()) {
+            httpClient = httpClient.proxyWithSystemProperties();
+        }
+
+        return builder(httpClient);
+    }
+
+    public static WebClient.Builder builder(HttpClient httpClient) {
```
nidhi-nair, line 38: I might be missing something here but this code path does not seem to go through the `applyProxyIfConfigured` method right? @sharat87

sharat87: Yep, good catch! Fixed now. :)

nidhi-nair (approved):
LGTM

sharat87:
@nidhi-nair, one last review/approval please? :D

nidhi-nair (approved):
Woohoo!! 🎉🎉🎉

PR #14485, opened 2022-06-13
perf: Widget re-rendering refactor

### LINKED ISSUE #12938: [Task]: Fix re-rendering of all Canvas Widgets unnecessarily due to small change in one widget

### Benchmark:
measured initial performance (this for one user action).
- In edit an improvement from ~250ms -> ~150ms
- In view an improvement from ~150ms -> ~80ms

### SubTasks

_No response_

ashit-rath:
## Description

This PR has core logic changes on how widgets are rendered. Instead of the entire DSL flowing through the flow of creating canvases and widgets, Only a structure similar to DSL is used to do the Same. This Structure only contains a bare bones structure, and the actual widget properties are selected directly from the store for each individual widgets.

Changes also includes,
- Updating state of changed widgets instead of all the widgets in the Redux state (using diff to find the updated widgets and also added to framework to skip diffing when updated widgets are already know, which can be used in the future on state updates).
- Selecting occupied spaces in components only when needed
- Replaced `commentModeSelector` which has expensive logic inside the selector itself that ran on every state change with a `useCommentMode` that only performs the calculation on when dependencies are changed.
- update selected widget state only when a needed instead of updating redux state on every action.
 
Fixes #12938

## Type of change

- Performance improvement

## How Has This Been Tested?

Manual UI and all existing tests should pass.
 
## Checklist:


app/client/src/utils/WidgetFactory.tsx
```diff
@@ -96,10 +98,10 @@ class WidgetFactory {
   }
 
   static createWidget(
-    widgetData: WidgetDataProps,
+    widgetData: CanvasWidgetStructure,
     renderMode: RenderMode,
```
ashit-rath, line 102: Remove?

app/client/src/pages/Editor/WidgetsEditor/CanvasContainer.tsx
```diff
@@ -48,7 +49,9 @@ function CanvasContainer() {
   const dispatch = useDispatch();
   const currentPageId = useSelector(getCurrentPageId);
   const isFetchingPage = useSelector(getIsFetchingPage);
-  const widgets = useSelector(getCanvasWidgetDsl);
+  // const widgets = useSelector(getCanvasWidgetDsl);
+  const canvasWidth = useSelector(getCanvasWidth);
+  const widgetsStructure = useSelector(getCanvasWidgetsStructure);
```
ashit-rath, line 54: Add deep equal check.

app/client/src/reducers/entityReducers/canvasWidgetsStructureReducer.ts
```diff
@@ -0,0 +1,63 @@
...
+    state: CanvasWidgetsStructureReduxState,
+    action: ReduxAction<UpdateCanvasPayload>,
+  ) => {
+    return denormalize("0", action.payload.widgets);
+  },
+  [ReduxActionTypes.UPDATE_LAYOUT]: (
+    state: CanvasWidgetsStructureReduxState,
+    action: ReduxAction<UpdateCanvasPayload>,
+  ) => {
+    return denormalize("0", action.payload.widgets);
```
ashit-rath, line 59: Check for re-rendering

app/client/src/utils/hooks/useCommentMode.ts
```diff
@@ -0,0 +1,23 @@
+import { matchBuilderPath, matchViewerPath } from "constants/routes";
```
Rishabh-Rathod, line 1: Can we separate `useCommentMode()` changes and widget re-rendering changes, if they are not dependent? 
- This way we could measure performance separately for both the changes.
- Maybe `useCommentMode()` can be shipped sooner and we could continue refactor for widget re-rendering.

rahulramesha: @Rishabh-Rathod I agree that changing it to a different PR helps in reviews, but I am clubbing in as much performance changes where ever visible in this PR as we are planning to have a complete regression on this PR.

Rishabh-Rathod: > I am clubbing in as much performance changes where ever visible in this PR as we are planning to have a complete regression on this PR.


This PR might take more time IMO to be ready. 
One of my concerns is regarding https://github.com/appsmithorg/appsmith/pull/14485#discussion_r905786853
- We will need to determine if denormalization here is causing re-rendering of widgets. If yes then we will need to figure out solution to it.
( please let me know if this is no more a concern  )

 
Few of the optimisations could be shipped without a need of separate regression testing. 
( normal release regression testing might serve the purpose ) 
_we could merge those code to release and take pull to this PR._

- Performance improvements per optimisation will be hard to determine as we club multiple optimisations into single PR. 

If we keep shipping improvements separately we would have clarity on improvements per optimisation.

I could help if in case separating code is bit complicated here.

riodeuno:
@rahulramesha The title says, WIP. Is there anything in particular you'd like to review and discuss?

app/client/src/widgets/FormWidget/widget/index.tsx
```diff
@@ -60,16 +57,21 @@ class FormWidget extends ContainerWidget {
...
+          child.children?.length &&
           this.checkFormValueChanges(get(child, "children[0]")),
       );
     }
 
     return hasChanges;
   }
 
+  getChildContainer = () => {
+    const { childWidgets = [] } = this.props;
```
riodeuno, line 69: Now I see the use of `childWidgets` 😄 
I was wondering if we can create a new Widget API called `getDerivedPropertiesFromChildren`, and have the evaluation phase determine these?
cc @SatishGandham

app/client/src/widgets/FormWidget/widget/index.tsx
```diff
@@ -89,16 +91,26 @@ class FormWidget extends ContainerWidget {
...
+          grandChild.onReset = this.handleResetInputs;
+          return grandChild;
+        },
+      );
     }
-    return super.renderChildWidget(childWidgetData);
+
+    // TODO(Ashit): Fix this
+    // eslint-disable-next-line
+    // @ts-ignore
```
riodeuno, line 112: As @ashit-rath mentioned, let us try to avoid disabling eslint.

app/client/src/widgets/ListWidget/widget/index.tsx
```diff
@@ -650,11 +650,26 @@ class ListWidget extends BaseWidget<ListWidgetProps<WidgetProps>, WidgetState> {
     return updatedChildren;
   };
 
+  /**
+   * We add a flag here to not fetch the widgets from the canvasWidgets
+   * in the metaHOC base on the widget id. Rather use the props as is.
+   */
+  addFlags = (children: DSLWidget[]) => {
+    return (children || []).map((childWidget) => {
+      childWidget.skipWidgetPropsHydration = true;
```
riodeuno, line 659: This can be done via enhancements. `widgets/ListWidget/index.ts`

app/client/src/widgets/withWidgetProps.tsx
```diff
@@ -0,0 +1,114 @@
...
+  getChildWidgets,
+  createLoadingWidget,
+  getRenderMode,
+} from "selectors/editorSelectors";
+import { AppState } from "reducers";
+import { CanvasWidgetStructure } from "./constants";
+import { getCanvasWidget } from "selectors/entitiesSelector";
+import { useSelector } from "react-redux";
+
+const WIDGETS_WITH_CHILD_WIDGETS = ["LIST_WIDGET", "FORM_WIDGET"];
```
riodeuno, line 23: I was wondering if we can move this to a widget config? In the `index.ts` file within the individual widget folders?

riodeuno: If we can have the `getDerviedPropertiesFromChildren` API, this won't be necessary.

app/client/src/widgets/withWidgetProps.tsx
```diff
@@ -0,0 +1,114 @@
...
+    );
+
+    const childWidgets = useSelector((state: AppState) => {
+      if (!WIDGETS_WITH_CHILD_WIDGETS.includes(type)) return undefined;
+      return getChildWidgets(state, widgetId);
+    }, equal);
+
+    let widgetProps: WidgetProps = {} as WidgetProps;
+
+    if (!skipWidgetPropsHydration) {
```
riodeuno, line 52: Since we have an opportunity to hydrate like this, we could perform the property updates which we do in the ContainerWidget for the child CanvasWidget, here.

app/client/src/reducers/entityReducers/canvasWidgetsStructureReducer.ts
```diff
@@ -0,0 +1,76 @@
...
+      children?: string[];
+    })
+  | orType;
+
+export type CanvasWidgetsStructureReduxState = {
+  children?: CanvasWidgetsStructureReduxState[];
+  type: WidgetType;
+  widgetId: string;
+  parentId?: string;
+  bottomRow: number;
```
riodeuno, line 27: `topRow` is not mentioned here. Is that intentional?

app/client/src/reducers/entityReducers/canvasWidgetsStructureReducer.ts
```diff
@@ -0,0 +1,76 @@
...
+};
+
+const initialState: CanvasWidgetsStructureReduxState = {
+  type: "CANVAS_WIDGET",
+  widgetId: MAIN_CONTAINER_WIDGET_ID,
+  bottomRow: 10,
```
riodeuno, line 33: @rahulramesha I believe we've added a default min bottomRow here. Shall we use the same constant?

app/client/src/constants/WidgetConstants.tsx
```diff
@@ -138,6 +138,15 @@ export const WIDGET_STATIC_PROPS = {
   noContainerOffset: false,
 };
 
+export const WIDGET_DSL_STRUCTURE_PROPS = {
+  children: true,
+  type: true,
+  widgetId: true,
+  parentId: true,
+  topRow: true,
+  bottomRow: true,
```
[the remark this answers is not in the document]
riodeuno: Also, how about making this an `enum` instead?

app/client/src/reducers/entityReducers/canvasWidgetsStructureReducer.ts
riodeuno: If we're using the `UPDATE_LAYOUT` action, it is possible that even for property updates, we're re-computing the canvas widget structure. Resulting in the Canvas re-render. 
However, I believe, we've measured that re-renders no longer happen unless  the structure actually changes. 
So, I was wondering why it is not the case as I imagine. Please help me understand.

app/client/src/selectors/editorSelectors.tsx
```diff
@@ -294,6 +310,73 @@ export const getCanvasWidgetDsl = createSelector(
...
+);
+
+function getWidgetSpecificChildProps(type: string) {
+  if (type === "FORM_WIDGET") {
+    return ["value", "isDirty", "isValid", "isLoading", "children"];
+  }
+}
+
+/**
+ * Method to build a child widget tree
```
riodeuno, line 330: This seems like a regular function. Could we move this out of the selectors and add unit tests for this? Also, elaboration on what is happening here would help.

riodeuno (changes requested):

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/Widgets/Button/ButtonGroup_spec.js
```diff
@@ -5,6 +5,7 @@ const menuButton =
   ".t--buttongroup-widget .bp3-popover2-target > div > button > div";
 
 describe("Button Group Widget Functionality", function() {
+  const mod = Cypress.platform === "darwin" ? "meta" : "ctrl";
```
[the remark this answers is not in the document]
rahulramesha: Couldn't see anything, other files seem to be using similiar code as well. Will try to move this into a separate File, to be accessed for others.

app/client/src/actions/pageActions.tsx
```diff
@@ -132,10 +132,11 @@ export const updateAndSaveLayout = (
   widgets: CanvasWidgetsReduxState,
   isRetry?: boolean,
   shouldReplay?: boolean,
+  updatedWidgetIds?: string[],
```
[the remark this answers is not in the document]
rahulramesha: For updating the canvasWidgets Redux State, in this PR we are currently diffing to get the value, but this variable tells us which widgets were updated only those get updated in the redux state. For now it is only implemented for only one saga, this can be used for other sagas in the future and get rid of the diffing.

app/client/src/pages/common/CanvasArenas/hooks/useBlocksToBeDraggedOnCanvas.ts
```diff
@@ -117,7 +117,8 @@ export const useBlocksToBeDraggedOnCanvas = ({
     (state: AppState) => state.ui.widgetDragResize.isResizing,
   );
   const selectedWidgets = useSelector(getSelectedWidgets);
-  const occupiedSpaces = useSelector(getOccupiedSpaces, isEqual) || {};
+  const occupiedSpaces =
```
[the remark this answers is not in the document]
rahulramesha: This was included in the PR as this caused re-render of widget components even when that particular widget did not update. As the main goal of the PR is to make sure only those widget that changed gets rendered, included this as well.

app/client/src/reducers/entityReducers/canvasWidgetsReducer.tsx
```diff
@@ -25,10 +47,29 @@ const canvasWidgetsReducer = createImmerReducer(initialState, {
...
     action: ReduxAction<UpdateCanvasPayload>,
   ) => {
-    return action.payload.widgets;
+    let listOfUpdatedWidgets;
+    // if payload has knowledge of which widgets were changed, use that
+    if (action.payload.updatedWidgetIds) {
+      listOfUpdatedWidgets = action.payload.updatedWidgetIds;
+    } // else diff out the widgets that need to be updated
+    else {
+      const updateLayoutDiff = diff(state, action.payload.widgets);
```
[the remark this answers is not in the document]
rahulramesha: `updatedWidgetIds` in the payload is already provided in the top as an alternative to this, only one saga uses it now. Later on, once all sagas use this payload, we can even remove the diffing.

For benchmarking, will try to do that and update you, but i'm confident that whatever the time it took to diff, it would have taken even more time in rendering if there was no diffing.

app/client/src/selectors/entitiesSelector.ts
```diff
@@ -466,6 +466,15 @@ export const getAppStoreData = (state: AppState): AppStoreState =>
 export const getCanvasWidgets = (state: AppState): CanvasWidgetsReduxState =>
   state.entities.canvasWidgets;
 
+export const getCanvasWidget = createSelector(
```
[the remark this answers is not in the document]
rahulramesha: yeah, now that i see it, both are similar, with check this out.

app/client/src/widgets/CanvasWidget.tsx
```diff
@@ -43,29 +45,19 @@ class CanvasWidget extends ContainerWidget {
...
-    }
 
-    // We don't render invisible widgets in view mode
-    if (
-      this.props.renderMode === RenderModes.PAGE &&
-      !childWidgetData.isVisible
-    ) {
-      return null;
-    }
+    const childWidget = klona(childWidgetData);
```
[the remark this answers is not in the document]
rahulramesha: @ashit-rath can explain better on this one.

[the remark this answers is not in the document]
ashit-rath: @riodeuno In this method we were modify the `childWidgetData` object by doing 
```js
childWidgetData.parentColumnSpace = snapSpaces.snapColumnSpace;
childWidgetData.parentRowSpace = snapSpaces.snapRowSpace;
```
As the `childWidgetData` was directly coming from the `props` of the widget component, they were not extensible and we were getting `Cannot add property parentColumnSpace, object is not extensible` error.

Not sure why we were not getting this error before.

app/client/src/utils/WidgetFactory.tsx
```diff
@@ -108,6 +104,8 @@ class WidgetFactory {
     const widgetBuilder = this.widgetMap.get(widgetData.type);
     if (widgetBuilder) {
       // TODO validate props here
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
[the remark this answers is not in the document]
ashit-rath: We should remove the disable eslint and fix the ts error by modifying the `widgetMap` typing to use the `CanvasWidgetStructure`

```ts
static widgetMap: Map<
    WidgetType,
    WidgetBuilder<CanvasWidgetStructure, WidgetState>
  > = new Map();
  ```
cc: @rahulramesha

File: app/client/src/constants/WidgetConstants.tsx (hunk 12 above)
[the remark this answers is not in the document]
Comment (SatishGandham) [reply app/client/src/constants/WidgetConstants.tsx:147]: @rahulramesha , How are widgets getting these props? Are they subscribing to them individually?

File: app/client/src/pages/AppViewer/AppViewerPageContainer.tsx [hunk 21]
```diff
@@ -35,7 +33,8 @@ type AppViewerPageContainerProps = RouteComponentProps<AppViewerRouteParams>;
 
 function AppViewerPageContainer(props: AppViewerPageContainerProps) {
   const currentPageName = useSelector(getCurrentPageName);
-  const widgets = useSelector(getCanvasWidgetDsl);
+  const widgetsStructure = useSelector(getCanvasWidgetsStructure);
```
SatishGandham, line 36: What does this do? Does it return the widgets as nested json?

app/client/src/widgets/CanvasWidget.tsx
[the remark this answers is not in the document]
riodeuno: The error is because the props are now immer objects. So, we can't change the value without using `produce` or extracting it. If cloning is necessary, I would recommend using `produce` from `immerjs`. If we can avoid this, it would be for the best. In fact, we could use `produce` once, in the `withWidgetProps.tsx` and update all properties we need to update there. This way, we will have one single point of mutation.

app/client/src/widgets/withWidgetProps.tsx
```diff
@@ -0,0 +1,114 @@
...
+       * is set to false
+       * If the isVisible property is undefined and the widget is not MODAL_WIDGET then isVisible
+       * is set to true
+       */
+      widgetProps.isVisible =
+        canvasWidgetProps.isVisible ??
+        canvasWidgetProps.type !== "MODAL_WIDGET";
+
+      if (widgetId === MAIN_CONTAINER_WIDGET_ID) {
+        widgetProps.rightColumn = canvasWidgetProps.rightColumn;
```
Rishabh-Rathod, line 76: As above we have widgetProps generated using canvasWidgetProps `widgetProps = { ...canvasWidgetProps };` 
we might not need this code, correct?

app/client/src/widgets/withWidgetProps.tsx
```diff
@@ -0,0 +1,114 @@
...
+        widgetProps.bottomRow = canvasWidgetProps.bottomRow;
+        widgetProps.minHeight = canvasWidgetProps.minHeight;
+        widgetProps.parentColumnSpace = canvasWidgetProps.parentColumnSpace;
+        widgetProps.parentRowSpace = canvasWidgetProps.parentRowSpace;
```
Rishabh-Rathod, line 80: As above we have widgetProps generated using canvasWidgetProps `widgetProps = { ...canvasWidgetProps };` 
we might not need this code, correct?

app/client/src/widgets/withWidgetProps.tsx
```diff
@@ -0,0 +1,114 @@
...
+        widgetProps.canExtend = props.canExtend;
+        widgetProps.parentId = props.parentId;
+      } else {
+        widgetProps.parentColumnSpace = props.parentColumnSpace;
+        widgetProps.parentRowSpace = props.parentRowSpace;
+        widgetProps.parentId = props.parentId;
+
+        // Form Widget Props
+        widgetProps.onReset = props.onReset;
+        if ("isFormValid" in props) widgetProps.isFormValid = props.isFormValid;
```
Rishabh-Rathod, line 95: Shouldn't we check if widget type is form widget then only run above logic?

app/client/src/widgets/withWidgetProps.tsx
```diff
@@ -0,0 +1,114 @@
...
+      }
```
Rishabh-Rathod, line 96: Instead of if-else ladder can we use switch case here according to widget type taking particular action?

app/client/src/widgets/withWidgetProps.tsx
```diff
@@ -0,0 +1,114 @@
...
+    );
+    const renderMode = useSelector(getRenderMode);
+    const evaluatedWidget = useSelector((state: AppState) =>
+      getWidgetEvalValues(state, canvasWidget?.widgetName),
+    );
+    const isLoading = useSelector((state: AppState) =>
+      getIsWidgetLoading(state, canvasWidget?.widgetName),
+    );
+
+    const childWidgets = useSelector((state: AppState) => {
```
Rishabh-Rathod, line 45: Just thinking out loud, instead of using `useSelector` to subscribe the store what if inside `getChildWidgets` we get store value from store.getState(). 
- I am assuming that whenever a child widget value will change, it will also change a value in parent making parent to re-render. 
- That re-render would be enough to trigger `getChildWidgets` and do the needed calculation.
- Like FormWidget has `data` and ListWidget has `childrenMetaPropertiesMap`, `childrenDefaultPropertiesMap`, `childrenEntityDefinitions` , `privateWidgets`

rahulramesha: hmm, It sounds possible but don't think we can be sure that form widgets gets updated values of its children.

app/client/src/widgets/withWidgetProps.tsx
rahulramesha: Thanks for pointing it out, must have missed this

app/client/src/widgets/withWidgetProps.tsx
rahulramesha: `isFormValid` will only exist in Form widget props, so don't think there is a need to check for form widget.

app/client/src/widgets/withWidgetProps.tsx
rahulramesha: Will be removing the first if condition, so it will just be a single if.. else statement

app/client/src/widgets/withWidgetProps.tsx
rahulramesha: This sounds possible but don't think, We can be always sure the parent gets updated child widgets.

app/client/src/sagas/WidgetOperationSagas.tsx
```diff
@@ -620,14 +620,23 @@ function* batchUpdateMultipleWidgetsPropertiesSaga(
...
     performance.now() - start,
     "ms",
   );
 
   // Save the layout
-  yield put(updateAndSaveLayout(updatedStateWidgets));
+  yield put(
+    updateAndSaveLayout(
+      updatedStateWidgets,
+      undefined,
```
ashit-rath, line 635: Instead of having 4 parameters for the `updateAndSaveLayout`, can we have just 2 and it could look something like
```
updateAndSaveLayout(updatedStateWidgets, options)
```
the `options` would have the reset of the parameters as properties.
There are some benefits of doing this
1. We won't have to pass `undefined`
2. The options object make the values more verbose.
3. Order does not matter anymore and less chance of mistakes

app/client/src/utils/hooks/useCommentMode.ts
```diff
@@ -0,0 +1,29 @@
...
+  const [isCommentMode, setCommentMode] = useState(commentMode);
+  useEffect(() => {
+    const pathName = window.location.pathname;
+    const onEditorOrViewerPage =
+      matchBuilderPath(pathName) || matchViewerPath(pathName);
+
+    if ((window as any).isCommentModeForced) setCommentMode(true);
+    else setCommentMode(commentMode && !!onEditorOrViewerPage);
+  }, [
+    window.location.pathname,
```
ashit-rath, line 23: Can we also add `setCommentMode` to the deps just to be safe?

app/client/src/widgets/withWidgetProps.tsx
Rishabh-Rathod: Yes, but as this else for is for all widgets except `MAIN_CONTAINER_WIDGET_ID` and `CANVAS_WIDGET` shouldn’t we first check if type is form widget then only do 
```
widgetProps.onReset = props.onReset;
 if ("isFormValid" in props) widgetProps.isFormValid = props.isFormValid;
```
I understand that `"isFormValid" in props` is only in FORM_WIDGET but won't it be better to have a separate case for FORM_WIDGET ? where we do
```
case "FORM_WIDGET": {
widgetProps.onReset = props.onReset;
widgetProps.isFormValid = props.isFormValid;
}
```

app/client/src/widgets/withWidgetProps.tsx
Rishabh-Rathod: https://github.com/appsmithorg/appsmith/pull/14485#discussion_r926314305
As above comments mention to add case for `FORM_WIDGET` if that seems valid maybe switch case could still be better for readability but just a suggestion.

app/client/src/widgets/withWidgetProps.tsx
Rishabh-Rathod: > don't think, We can be always sure the parent gets updated child widgets.

Makes sense to implement current way as we are planning to solve FormWidget and ListWidget issue separately soon and there is no performance degradation for those widgets.

app/client/src/pages/Editor/WidgetsEditor/CanvasContainer.tsx
```diff
@@ -48,7 +50,8 @@ function CanvasContainer() {
   const dispatch = useDispatch();
   const currentPageId = useSelector(getCurrentPageId);
   const isFetchingPage = useSelector(getIsFetchingPage);
-  const widgets = useSelector(getCanvasWidgetDsl);
+  const canvasWidth = useSelector(getCanvasWidth);
+  const widgetsStructure = useSelector(getCanvasWidgetsStructure, isEqual);
```
SatishGandham, line 54: Check how canvasWidgetsStructure is updated and see if we need isEqual here.

app/client/src/pages/AppViewer/AppViewerPageContainer.tsx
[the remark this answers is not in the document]
SatishGandham: We may need deep equal, check how getCanvasWidgetsStructure is updated.
Make all uses of this selector consistent.

app/client/src/resizable/resizenreflow/index.tsx
```diff
@@ -17,7 +17,7 @@ import {
   ReflowedSpace,
 } from "reflow/reflowTypes";
 import { getNearestParentCanvas } from "utils/generators";
-import { getOccupiedSpaces } from "selectors/editorSelectors";
+import { getContainerOccupiedSpacesSelectorWhileResizing } from "selectors/editorSelectors";
```
SatishGandham, line 20: Do we use selector in the name of the selectors?

app/client/src/reducers/entityReducers/canvasWidgetsReducer.tsx
SatishGandham, line 56: `updateLayoutDiff` name feels like an action or function name instead of a variable name. Can we change it to
updateLayoutDiffs

SatishGandham: Action item #1, create an issue to pass down updated widget list from all sagas.

app/client/src/selectors/propertyPaneSelectors.tsx
```diff
@@ -153,22 +153,31 @@ const populateEvaluatedWidgetProperties = (
   return evaluatedProperties;
 };
 
+const getCurrentEvaluatedWidget = createSelector(
+  getCurrentWidgetProperties,
```
SatishGandham, line 157: What is current widget?
Since we are only using widget name, can't we use the selected widgets?

And, why are we using ID's to maintain selected widgets?

app/client/src/widgets/StatboxWidget/widget/index.tsx
```diff
@@ -103,17 +102,6 @@ class StatboxWidget extends ContainerWidget {
     ];
   }
 
-  renderChildWidget(childWidgetData: WidgetProps): React.ReactNode {
```
SatishGandham, line 106: Are we removing this because this is an override from the container widget?

ashit-rath: Yes this is an override.
The logic used here was not functional and from a product's perspective, the functionality seemed irrelevant. So we decided to remove it.
Removing this gives us a brownie point to not persist the modified properties in the `withWidgetProps`.

app/client/src/pages/Editor/WidgetsEditor/CanvasContainer.tsx
rahulramesha: This is required because the `widgetsStructure` changes reference during denormalization

SatishGandham: Ok, there are places where we are using this selector without isEqual. Please check if we need deep equal there.

app/client/src/resizable/resizenreflow/index.tsx
rahulramesha: `getContainerOccupiedSpacesSelectorWhileResizing` is a method that returns a selector and not a selector itself. selector was added in the name to indicate that.

app/client/src/widgets/FormWidget/widget/index.tsx
```diff
@@ -60,16 +57,21 @@ class FormWidget extends ContainerWidget {
...
           this.checkFormValueChanges(get(child, "children[0]")),
       );
     }
 
     return hasChanges;
   }
 
+  getChildContainer = () => {
+    const { childWidgets = [] } = this.props;
+    return { ...childWidgets[0] };
```
SatishGandham, line 70: Why are we spreading it and creating a new top level object?

app/client/src/widgets/CanvasWidget.tsx
SatishGandham, line 51: @rahulramesha , @ashit-rath 
Since our CanvasWidgetStructure is very basic without nesting except for children array, can we use simple spread?

app/client/src/selectors/propertyPaneSelectors.tsx
rahulramesha: Current widget is the widget that was selected. selected widgets are stored in IDs, so we need to access the widgetName by getting the properties.

I think we store selected Widget in IDs because we store normalized values by IDs as well.

SatishGandham: Action item 3, use only one unique identifier for widgets.

app/client/src/widgets/FormWidget/widget/index.tsx
rahulramesha: This was to make sure it doesnt mutate the childWidget.

app/client/src/reducers/uiReducers/dragResizeReducer.ts
```diff
@@ -123,7 +125,7 @@ export const widgetDraggingReducer = createImmerReducer(initialState, {
     action: ReduxAction<{ widgetIds?: string[] }>,
   ) => {
     const { widgetIds } = action.payload;
-    if (widgetIds) {
+    if (widgetIds && !areArraysEqual(widgetIds, state.selectedWidgets)) {
```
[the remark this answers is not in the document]
rahulramesha: should check arrays regardless of order hence deepEqual might not work here

app/client/src/reducers/entityReducers/canvasWidgetsStructureReducer.ts
```diff
@@ -0,0 +1,76 @@
...
+  rootWidgetId: string,
+  widgets: Record<string, FlattenedWidgetProps>,
+): CanvasWidgetStructure {
+  const rootWidget = widgets[rootWidgetId];
+
+  const children = (rootWidget.children || []).map((childId) =>
+    denormalize(childId, widgets),
+  );
+
+  const staticProps = [...Object.keys(WIDGET_DSL_STRUCTURE_PROPS)];
```
SatishGandham, line 52: Is this spread required?

app/client/src/selectors/editorSelectors.tsx
```diff
@@ -294,6 +310,73 @@ export const getCanvasWidgetDsl = createSelector(
...
+  [
+    getCanvasWidgets,
+    getDataTree,
+    getLoadingEntities,
+    (_state: AppState, widgetId: string) => widgetId,
+  ],
+  buildChildWidgetTree,
+);
+
+function getWidgetSpecificChildProps(type: string) {
```
SatishGandham, line 323: It's not a good sign if we have to handle per widget here.
We should find a better way to do it. 
Action item 4

app/client/src/widgets/CanvasWidget.tsx
ashit-rath: Yes a spread should work here

app/client/src/selectors/editorSelectors.tsx
```diff
@@ -35,6 +35,13 @@ import { Page } from "@appsmith/constants/ReduxActionConstants";
 import { PLACEHOLDER_APP_SLUG, PLACEHOLDER_PAGE_SLUG } from "constants/routes";
 import { ApplicationVersion } from "actions/applicationActions";
 import { MainCanvasReduxState } from "reducers/uiReducers/mainCanvasReducer";
+import { LoadingEntitiesState } from "reducers/evaluationReducers/loadingEntitiesReducer";
+import _ from "lodash";
```
SatishGandham, line 39: Please import only the functions we are using.
I see that we are only using bind so `import {bind} from "lodash";`

app/client/src/selectors/editorSelectors.tsx
```diff
@@ -329,101 +412,171 @@ const getWidgetSpacesForContainer = (
...
 ) {
-  return createSelector(getWidgets, (widgets: CanvasWidgetsReduxState):
-    | OccupiedSpace[]
-    | undefined => {
-    if (containerId === null || containerId === undefined) return undefined;
-
-    const containerWidget: FlattenedWidgetProps = widgets[containerId];
+  return createSelector(
+    getWidgets,
+    _.bind(generateOccupiedSpacesForContainer, null, _, true, containerId),
```
SatishGandham, line 511: What does this do?

app/client/src/reducers/entityReducers/canvasWidgetsStructureReducer.ts
rahulramesha: Will remove this

app/client/src/selectors/editorSelectors.tsx
SatishGandham: Won't passing null or undefined work? Is bind required at all? We are  not passing `this` context anyway.
My concern is with that import statement, webpack will include the entire lodash library.

Can you please add some comments to the bind line. I haven't used bind a lot and I find it confusing!!

app/client/src/widgets/withWidgetProps.tsx
rahulramesha: The problem is, even with the switch case, we still need to check the condition `"isFormValid" in props` because of how it is written here, https://github.com/appsmithorg/appsmith/blob/9dea16317216d7ffb9796cdb59329bcb84e34e93/app/client/src/widgets/FormButtonWidget/widget/index.tsx#L158.
The code will look like, 
```
case "FORM_WIDGET": {
  if ("isFormValid" in props) {
    widgetProps.onReset = props.onReset;
    widgetProps.isFormValid = props.isFormValid;
  }
}
```

app/client/src/widgets/withWidgetProps.tsx
rahulramesha: Yes, This change is temporary. there is an issue to move this out by using derivedProperties. https://github.com/appsmithorg/appsmith/issues/15097

app/client/src/widgets/ContainerWidget/widget/index.tsx
```diff
@@ -161,26 +162,23 @@ class ContainerWidget extends BaseWidget<
...
-    }
+  renderChildWidget(childWidgetData: CanvasWidgetStructure): React.ReactNode {
+    const childWidget = { ...childWidgetData };
 
     const { componentHeight, componentWidth } = this.getComponentDimensions();
 
-    childWidgetData.rightColumn = componentWidth;
-    childWidgetData.bottomRow = this.props.shouldScrollContents
-      ? childWidgetData.bottomRow
+    childWidget.rightColumn = componentWidth;
```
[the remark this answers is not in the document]
rahulramesha: @riodeuno, Sounds like a viable solution to have this logic elsewere. Maybe we can explore the solution in a separate PR

app/client/src/constants/AppConstants.ts
```diff
@@ -1,9 +1,13 @@
 import localStorage from "utils/localStorage";
+import { GridDefaults } from "./WidgetConstants";
 
 export const CANVAS_DEFAULT_HEIGHT_PX = 1292;
 export const CANVAS_DEFAULT_MIN_HEIGHT_PX = 380;
 export const CANVAS_DEFAULT_GRID_HEIGHT_PX = 1;
 export const CANVAS_DEFAULT_GRID_WIDTH_PX = 1;
+export const CANVAS_DEFAULT_MIN_HEIGHT_ROWS = Math.ceil(
```
SatishGandham, line 8: Shouldn't this be `CANVAS_DEFAULT_MIN_ROWS`

SatishGandham:
/ok-to-test sha=5ea91c8

Because it's nice to see the dip in scripting numbers 😀

app/client/src/utils/widgetRenderUtils.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { WidgetProps } from "widgets/BaseWidget";
+import { LoadingEntitiesState } from "reducers/evaluationReducers/loadingEntitiesReducer";
+
+/**
+ * checks if array of strings are equal regardless of order
+ * @param arr1
+ * @param arr2
+ * @returns
+ */
+export function areArraysEqual(arr1: string[], arr2: string[]) {
```
SatishGandham, line 22: This should be moved to global utils.

app/client/src/reducers/entityReducers/canvasWidgetsStructureReducer.ts
SatishGandham: cc: @rahulramesha

SatishGandham (approved):

app/client/src/widgets/ContainerWidget/widget/index.tsx
```diff
@@ -162,25 +162,22 @@ class ContainerWidget extends BaseWidget<
...
+    childWidget.bottomRow = this.props.shouldScrollContents
+      ? childWidget.bottomRow
       : componentHeight;
-    childWidgetData.minHeight = componentHeight;
-    childWidgetData.isVisible = this.props.isVisible;
-    childWidgetData.shouldScrollContents = false;
-    childWidgetData.canExtend = this.props.shouldScrollContents;
+    childWidget.minHeight = componentHeight;
+    // childWidget.isVisible = this.props.isVisible;
```
ashit-rath, line 174: Can we remove this commented out code?

app/client/src/widgets/ModalWidget/widget/index.tsx
```diff
@@ -111,18 +111,20 @@ export class ModalWidget extends BaseWidget<ModalWidgetProps, WidgetState> {
...
+    childData.shouldScrollContents = false;
+    childData.canExtend = this.props.shouldScrollContents;
+    childData.bottomRow = this.props.shouldScrollContents
+      ? Math.max(childData.bottomRow, this.props.height)
       : this.props.height;
-    childWidgetData.isVisible = this.props.isVisible;
-    childWidgetData.containerStyle = "none";
-    childWidgetData.minHeight = this.props.height;
-    childWidgetData.rightColumn =
+    childData.isVisible = this.props.isVisible;
```
ashit-rath, line 121: Do we need to set this? as if the widget's `isVisible` is set to false
View Mode - the modal widget will not render so the child won't render
Edit Mode - the modal widget will get an overlay and I think doesn't matter if the child widget `isVisible` is true as the parent overlay should be suffice.

I think we removed it in the container widget and it's also not being used in the `withWidgetProps`

ashit-rath (commented):
Couple of tiny comments, otherwise LGTM!

rahulramesha:
@SatishGandham @ashit-rath @Rishabh-Rathod @riodeuno This is ready for QA. Please let me know if this requires anymore changes.

cc: @somangshu

somangshu:
cc @vivekverma2312 @laveena-en please plan how we are going to do a full round of testing all widgets here.

riodeuno:
@rahulramesha I don't have any other comments other than the ones I've already shared in the review. I have noticed that a few of them don't have responses. Would you and other contributors to this PR go through them and respond?

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -442,6 +477,7 @@ export interface WidgetBaseProps {
   parentId?: string;
   renderMode: RenderMode;
   version: number;
+  childWidgets?: DataTreeWidget[];
```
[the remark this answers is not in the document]
rahulramesha: The `childWidgets` are specifically used by List widget and Form widget, these are different from `children`, because we childWidgets contain evaluatedValues as well, which are required to determine form or list widget's state. However this is a temporary arrangement and there is already an issue to move that logic to derived properties.

app/client/src/widgets/CanvasWidget.tsx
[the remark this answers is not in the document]
rahulramesha: This is resolved, we are not cloning this anymore.

app/client/src/widgets/FormWidget/widget/index.tsx
rahulramesha: yes, there is already an issue to do exactly this. Once that is done, we can get rid of `childWidgets`.

app/client/src/widgets/ListWidget/widget/index.tsx
rahulramesha: cc: @ashit-rath

app/client/src/widgets/withWidgetProps.tsx
rahulramesha: As mentioned in above comment, we can explore moving that logic in a separate PR

rahulramesha:
@riodeuno apologies, must have lost track of few of them in all the review comments, replied to the queries, 
for, https://github.com/appsmithorg/appsmith/pull/14485#discussion_r924420866
As you guessed since the structure is recalculated, the reference changes and should re-render, but we select the state at only one place and we have a deepEqual there, which prevents them from re-renders. Since the structure sized down, the deepEqual will also not take much time to calculate.

app/client/src/widgets/ListWidget/widget/index.tsx
ashit-rath: Yes we can do it in enhancements but the intention was to keep this property transient and not part of the DSL.

riodeuno:
@rahulramesha Sounds good for now. In general, using `deepEquals` should be the last resort, as it is more of a patch fix in my opinion. Ideally, to prevent re-renders, the component data itself should have referential integrity.  This applies to data used by react components in general.

Here, this means that instead of calling `UPDATE_LAYOUT` where we replace all values, we should ideally update individual paths which have changed. This would make the `deepEquals` moot.

app/client/src/reducers/entityReducers/canvasWidgetsReducer.tsx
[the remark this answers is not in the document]
riodeuno: Sounds good. I've created a new issue for this. We'll tackle it separately.

rahulramesha:
@riodeuno yes, but in this particular scenario, where we consume the state in only one component per page for the CanvasStructure that contains all the values, diffing logic to update specific paths would take up more time or be equal to the deepEqual.

But Ideally, in the future, when we decide to further replace children with just their IDs instead of whole tree structure of children, That would be a much better solution without deepEqual. But currently to do that will require more cleanup and there are more unknowns with respect to how few widgets work.

app/client/src/widgets/ListWidget/widget/index.tsx
riodeuno: @ashit-rath Could you elaborate on the rationale behind this? A little background on what `skipWidgetPropsHydration` does would help.

ashit-rath: In the `withWidgetProps` HOC, a widget's properties are computed by looking into the evaluated value, canvasWidget value and certain positional properties of the parent.

For the list widget's children, this is not applicable as they are generated by the List widget and are not present in any reducer. Further any positional manipulation like the `parentColumnSpace` or `bottomRow` can throw off the calculations made by the list widget. So whatever the `withWidgetProps` receives, we would render it untouched.
https://github.com/appsmithorg/appsmith/pull/14485/files#diff-c28fdfd0ee823c81a1a1f4908d76304f7d647da672da05e8d98492426ae631b5R56


We are planning to remove this in the list v2.

somangshu:
@rahulramesha there is a pending change request, Can you request the author for an approval or make the change if required. 

@vivekverma2312 @laveena-en can you please give us an update on what is happening here with the QA regression.

rahulramesha:
@riodeuno, looks like there is change request still on the PR, since the review comments were all answered, could you please give a final review on the PR.

riodeuno (approved):

vivekverma2312:
@somangshu Testing is in progress. 7 widgets are pending, will be done by today EOD.

somangshu:
Awesome. 

@rahulramesha we can merge this as soon as we have the signoff on the rest

vivekverma2312:
@rahulramesha Completed regression on widgets and working fine.

SatishGandham (approved):
Looks good. 
Congratulations @rahulramesha 🥳🥳🥳

PR #14274, opened 2022-06-03
fix: Handle string datatype in api response

ChandanBalajiBP:
## Description

> When a response to an API is string for TABLE datatype, app crashes. 

Fixes #14242

## Type of change
- Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #14502, opened 2022-06-13
fix: skipping parse of empty array response

### LINKED ISSUE #14493: [Bug]: App Crashes when running a Select query that results in Empty table/No records [MySQL, Postgres]

### Description

When I run a `MySQL` query with `IS NULL` syntax keeping the prepared statement turned off then the following unexpected behaviors are noticed in the local environment (release branch) and in the release.app.appsmith.com as well.

## Shows Oops both in local and release.app.appsmith.com

Screenshot 2022-06-13 at 6 21 37 PM

## Broken Front-end only in local (release branch)

Screenshot 2022-06-12 at 12 56 18 AM

## Expected Behaviour found in app.appsmith.com

Screenshot 2022-06-13 at 6 23 22 PM

### Steps To Reproduce

1. Create a `MySQL` datasource
2. Write a query containing IS `NULL` syntax where the query is not supposed to return any row because of no matching rows found in the table
3. You should see a page saying **Oops! Something went wrong** and later you will see some front-end source code that basically breaks the entire front-end (Happens to exist in the local environment on the release branch but **Oops** page appears in `release.app.appsmith.com` too)'

## Demonstration
[Loom Video](https://www.loom.com/share/d4fac848b5194d2b906d8d921bdd05cc)

### Public Sample App

_No response_

### Version

Cloud / Self Hosted - v1.7.1

ayushpahwa:
## Description

- When we get a response from API, the data can be either null, undefined, not an array and an array but with no data
- We had handled everything else except the case of an empty array
- This commit fixes that, basically skipping parsing when the array is empty
Related PR: [14274](https://github.com/appsmithorg/appsmith/pull/14274)

Fixes #14493 

## Type of change

## How Has This Been Tested?

- Run MySQL and PgSQL queries that return an empty array as output. This used to cause crashes in the release since in array we were trying to parse its children as objects, which were null.

## Checklist:


AmanAgarwal041 (dismissed):

Aishwarya-U-R (approved):
Approving cypress test

ChandanBalajiBP (approved):

Issue #10659, opened 2022-01-26
[Task] Blueprint Saga workflows


riodeuno:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SubTasks

- [ ] Run blueprint sagas on widget move to another widget
- [ ] Run blueprint sagas when undo-delete a widget
- [ ] Run blueprint sagas when pasting a widget

riodeuno:
@sbalaji1192 I believe we should go with approach 1. instead of the original approach 2. I think that these `onAddChild` and `onRemoveChild` operations don't belong in the Blueprint, as the Blueprint is meant to deal with the construction of the widget and are not supposed to be run during the lifecycle of the widget on the canvas.

riodeuno:
@sbalaji1192 @somangshu @rohitagarwal88 @SatishGandham Is this update needed for the List Widget V2 efforts?

sbalaji1192:
@Tooluloope Could you please confirm if this is needed for List widget v2 implementation?

techbhavin:
Need to Improve 
- In the form widget current approach is the Button widget with two properties Hidden base on the approach  `requireParentProps` so it ideal way as per discussion [here](https://github.com/appsmithorg/appsmith/pull/14510#discussion_r913438735)

PR #14510, opened 2022-06-14
fix: depercate form button widget

### LINKED ISSUE #3613: [Feature]-[800]: Deprecate form button widget
## Description

Since we have resetWidget as a trigger function, we can deprecate the form button widget added with every form widget. This needs to be a backwards compatible change.

techbhavin:
## Description

> Deprecate the form button widget 

Fixes #3613 

## Type of change

## Checklist:


app/client/src/sagas/ActionExecution/ActionExecutionSagas.ts
```diff
@@ -149,7 +150,7 @@ export function* executeActionTriggers(
   return response;
 }
 
-export function* executeAppAction(payload: ExecuteTriggerPayload) {
+export function* executeAppAction(payload: ExecuteTriggerPayload): any {
```
jsartisan, line 153: @Rishabh-Rathod can you take a look at this change in this action saga file?

app/client/src/widgets/ButtonWidget/widget/index.tsx
```diff
@@ -321,6 +351,8 @@ class ButtonWidget extends BaseWidget<ButtonWidgetProps, ButtonWidgetState> {
           callback: this.handleActionComplete,
         },
       });
+    } else if (this.props.resetFormOnClick && this.props.onReset) {
```
jsartisan, line 354: What happens when onClick is also set and reset form on clicked is also toggled? Which of the actions will take priority? I think it should be resetFormOnClick. But from the code, it seems onClick will take priority. @dilippitchika can confirm on this.

jsartisan (approved):

jsartisan:
@Rishabh-Rathod there is one change in the action saga file. Can you take a look at that if that change is okay?

dilippitchika:
@aswathkk this impacts the property pane for button widget. Can you also check this once?

app/client/src/sagas/ActionExecution/ActionExecutionSagas.ts
Rishabh-Rathod: @techbhavin  i will need some context here about what are we trying to do in this PR and mainly `executeAppAction` and `initiateActionTriggerExecution`

aswathkk:
Thanks @dilippitchika
I will keep an eye on this PR.

aswathkk:
Thanks, @dilippitchika. I will take care of this once we merge this PR to release.

app/client/src/sagas/ActionExecution/ActionExecutionSagas.ts
techbhavin: @Rishabh-Rathod  answering to your questions 

1. Why do we need changes made in executeActionTriggers?

here to handle API execution callback, a response is handled. in almost trigger type, previously nothing was returned and not required to return but as per changes related to API execution, here I change for other types
`let response: unknown[] = [{ success: true }];`
in the location trigger type
```
ActionTriggerType.GET_CURRENT_LOCATION
ActionTriggerType.WATCH_CURRENT_LOCATION
ActionTriggerType.STOP_WATCHING_CURRENT_LOCATION
```
if the action executes successfully, we get current location data as a response for that case setting 
`set(response, "0.success", true);`
if there is an error to get the current location, initiateActionTriggerExecution actions catch block will execute

if not set the success flag into executeActionTriggers, I was not found any other way to decide that executed action
`const res: unknown[] = yield call(executeAppAction, action.payload); `
( at line 191 ) for which trigger type and how to check API success error callback.


2. Would like to understand how do we plan to handle the reset of form in the below cases.
as per line 198, currently support 1st level trigger
if Api1 was successful but Api2 failed. the form will reset as per the current implementation

thanks for giving other possible scenarios. have to create and check that.

dilippitchika: @Rishabh-Rathod i don't think this will ever be foolproof, the user can choose to switch this off and use resetWidget to do it themselves when the 2nd api call succeeds.

app/client/src/widgets/ButtonWidget/widget/index.tsx
techbhavin: @jsartisan , Here

this is handled by callback if there is onClick added, callback always takes care of post actions of onClick which is handled into handleActionComplete function if there is no onClick, and reset on submit enable. it just reset the form.

here onClick should be on priority coz we are dealing with API success error callbacks. so reset should be executed after onClick executed

app/client/src/pages/Editor/PropertyPane/PropertySection.tsx
```diff
@@ -49,7 +50,11 @@ type PropertySectionProps = {
   id: string;
   name: string;
   children?: ReactNode;
-  hidden?: (props: any, propertyPath: string) => boolean;
+  hidden?: (
+    props: any,
+    propertyPath: string,
+    shouldShowFormControl?: boolean,
```
riodeuno, line 56: Have we considered using the [childEnhancements](https://github.com/appsmithorg/appsmith/blob/703b0efda6a7ccc461458a3906ed1b93f1fb1c43/app/client/src/sagas/WidgetEnhancementHelpers.ts#L48)?

riodeuno: @jsartisan Could you help understand if we can use the widget enhancements instead of having this custom property only for this type of widget?

app/client/src/pages/Editor/PropertyPane/PropertySection.tsx
```diff
@@ -65,8 +70,25 @@ export const PropertySection = memo((props: PropertySectionProps) => {
...
+   * used to show hide property based on parent widget if required #3631
+   */
+  const shouldShowFormControl = useSelector((state: AppState) => {
+    const canvasWidgets = state.entities.canvasWidgets;
+    const widgetParentId = widgetProps.parentId as string;
+    if (canvasWidgets.hasOwnProperty(widgetParentId)) {
+      const widget = canvasWidgets[widgetParentId];
+      if (widget.parentId && canvasWidgets.hasOwnProperty(widget.parentId)) {
+        const parent = canvasWidgets[widget.parentId];
+        return parent.type === "FORM_WIDGET" && parent.parentId === "0";
```
riodeuno, line 83: This is an abstraction leak. We should be able to avoid this by specifying in the widget config, regarding any such enhancements.

chandannkumar:
@techbhavin Observations:
1. When "reset form on success" field is switched on it also resets if there's an error in the API
https://www.loom.com/share/d706265a9d5b4359b8d63d67254269ac
refer issue: https://github.com/appsmithorg/appsmith/issues/3613#issuecomment-1138453150

2. Form options can be seen twice on FormButton widget (on Migration)
image

app/client/src/constants/PropertyControlConstants.tsx
```diff
@@ -15,7 +15,11 @@ export type PropertyPaneSectionConfig = {
   sectionName: string;
   id?: string;
   children: PropertyPaneConfig[];
-  hidden?: (props: any, propertyPath: string) => boolean;
+  hidden?: (
+    props: any,
+    propertyPath: string,
+    parentWidgetId?: string,
```
riodeuno, line 21: The use of `parentWidgetId` here doesn't seem appropriate to me. 

One reason, is that this argument name, doesn't seem to reflect which widget's id it actually is. For example, if I have a button within a container within a form widget. Will this give me Form Widget's id or the Container Widget's Id?

Second reason, is that other widgets are effecting change in this widget. Ideally, I would like this to be a property of the widget itself. This could be made possible by enhancing the widget when it is created/moved in/pasted in, etc.

Please [see this](https://github.com/appsmithorg/appsmith/issues/10659) for a similar scenario.

app/client/src/sagas/selectors.tsx
```diff
@@ -175,3 +178,30 @@ export const getWidgetImmediateChildren = createSelector(
     return childrenIds;
   },
 );
+
+/**
+ * get actual parent of widget based on widgetId
+ * for button inside form, button's parent is form
+ * for button on canvas, parent is main container
+ */
+export const getWidgetParent = (widgetId: string) => {
```
riodeuno, line 187: I am not sure I understand this implementation. The way I see it, if `widgetId` is _not_ a widget of type "CANVAS_WIDGET" and `canvasWidgets[widgetId].parentId` is not `MAIN_CONTAINER_WIDGET_ID`, the parent we're looking for, is `canvasWidgets[canvasWidgets[widgetId].parentId]`. I don't see the need for the while loop. 

Are we looking for a specific type of parent? For example, FORM_WIDGET? if so, this implementation seems incorrect.

app/client/src/widgets/ButtonWidget/widget/index.tsx
```diff
@@ -121,6 +127,38 @@ class ButtonWidget extends BaseWidget<ButtonWidgetProps, ButtonWidgetState> {
           },
         ],
       },
+      {
+        sectionName: "Form options",
+        hidden: (
+          props: ButtonWidgetProps,
+          propertyPath: string,
+          parentWidgetId?: string,
+        ) => parentWidgetId === MAIN_CONTAINER_WIDGET_ID,
```
riodeuno, line 136: The widget seems aware of `MAIN_CONTAINER_WIDGET_ID`, this is an abstraction leak. Can we rethink how to implement this while maintaining abstractions?

app/client/src/sagas/ActionExecution/ActionExecutionSagas.ts
```diff
@@ -67,7 +68,7 @@ export function* executeActionTriggers(
   triggerMeta: TriggerMeta,
 ): any {
   // when called via a promise, a trigger can return some value to be used in .then
-  let response: unknown[] = [];
+  let response: unknown[] = [{ success: true }];
```
Rishabh-Rathod, line 71: Quoting @ApekshaBhosale query
`if there is an error while doing any calls, why is success is true by default?`

app/client/src/constants/PropertyControlConstants.tsx
sbalaji1192: @riodeuno We want to show two properties on the button property pane only when it's placed inside the form widget. there is a problem with enhancing the button widget on `BlueprintOperationTypes.CHILD_OPERATIONS` since it won't get called when the button is dragged outside the form or when the button is dragged into the canvas and then dragged into the form widget. So the solution I proposed to bhavin was to add a property to property configuration `requireParentProps`. if enabled, we will inject the closest widget parent prop into the hidden function. This solution would be scalable as well. We can do this injection into updateHook functions as well, if needed.

app/client/src/sagas/selectors.tsx
sbalaji1192: +1

chandannkumar:
@techbhavin DP link is broken

chandannkumar:
@techbhavin 
Buttons can be clicked when **Disabled Invalid Forms** is enabled and also button seems disabled in form
https://www.loom.com/share/03c1da03b57844889cebfbaffeef728c

chandannkumar:
@techbhavin 
DP link seems broken

chandannkumar:
Tested this PR and working as expected.
- Tested form button deprecation warning message
- Tested migration of form button which works fine
- Tested Button widget functionality in place of Form button
- Tested Button behavior when added inside form. Form options should be visible
- Tested Button behavior when added on canvas and not on form. Form options must gets hidden
- Tested for Button functionality must be same as Form button

app/client/src/constants/PropertyControlConstants.tsx
riodeuno: @sbalaji1192 I agree that the `CHILD_OPERATIONS` is not properly implemented. We have a [task to fix this here](https://github.com/appsmithorg/appsmith/issues/10659). 

Here is my concern with the approach  you've suggested.
With the `requireParentProps`, we cannot be certain which parent we're talking about, for example, if there is a Tabs Widget within a Form Widget, within which we have a button widget. Will we be showing the two new properties?
If we will only be showing these properties, if the nearest parent is a FormWidget, then it should work with the approach you've suggested. 
However, the abstraction breaks here, because, now the widgets make decisions based on the parent props (FormWidget), instead of the FormWidget deciding for the child widget. In the end, the two new properties, are properties related to the FormWidget and not the ButtonWidget.

With that said, if it is possible for us to revert to using the `onAddChild` mechanism in the future as described in #10659, I'm ok with moving ahead with this for now. Let's make sure to refer to #10659 in the comments above this implementation, and move forward.

We will have to get to #10659 before we move ahead with ListWidgetV2 in any case.

sbalaji1192: @riodeuno Yeah. We are only concerned about the immediate widget parent. If its deeply nested inside we are not gonna show the two properties.
I agree with the abstraction leak here. Like you mentioned we can move forward, with a condition that we will circle back once we add `onAddChild` functionality.
@techbhavin Could you please add a todo to refactor this when we address #10659. And also please leave a comment about this implementation on the #10659 issue as well.

riodeuno (approved):

PR #14532, opened 2022-06-14
feat: Deprecation warning message update to reference the widget that is replacing it

### LINKED ISSUE #14380: [Feature]: Support a way to reference the widget user needs to drag in the deprecation message

### Summary

Within the deprecation message, we wish to inform the user that they need to drag a specific widget to upgrade this. 
App viewers pod needs a way to provide the name of the widget the user needs to update to.

The message can be the following - 
The widget you are currently using is now deprecated. You can drag a new button widget and configure it to work in place of this widget.

But also check with @jnikhila to finalise.

Qualities of a good message
1. Explains what the problem is with the current widget
2. Gives a clear next step on what they immediately need to do
3. Provides a way to learn more about the change

This will help us communicate to the user that they must drag the specific widget which maybe sometimes different than the current widget they are working on. 

Example
1. Form button widget doesn't exist anymore a user will have to drag a button widget
2. Input widget is broken down into multiple widgets like phone and currency, a user who is using the phone from an older input widget must see the deprecation message show that they need to drag a phone widget and not an input widget.

rahulramesha:
## Description

Added a config field `replacement` for deprecated widgets to reference the widget that is replacing the deprecated widget.
Also, added a call to action link at the bottom of the banner, which can be updated by passing the `ctaText` and `ctaURL`.

Fixes #14380 

## Type of change

## How Has This Been Tested?
Manual UI

## Checklist:


app/client/src/pages/Editor/utils.ts
```diff
@@ -237,6 +242,43 @@ export const useQuery = () => {
...
+ * @param WidgetType
+ * @returns
+ */
 export function isWidgetDeprecated(WidgetType: WidgetType) {
-  return !!WidgetFactory.widgetConfigMap.get(WidgetType)?.isDeprecated;
+  const currentWidgetConfig = WidgetFactory.widgetConfigMap.get(WidgetType);
+  const isDeprecated = !!currentWidgetConfig?.isDeprecated;
+  let widgetReplacedWith;
+  if (isDeprecated && currentWidgetConfig?.replacement) {
+    widgetReplacedWith = WidgetFactory.widgetConfigMap.get(
```
aswathkk, line 258: @rahulramesha 
One scenario I can think about is that this particular widget that you get here might also be deprecated.
Instead of directly showing the displayName of this widget, we can recursively check whether the widget is deprecated and only show the displayName of the non-deprecated widget.

aswathkk: That is the best solution. 💯

app/client/src/components/ads/BannerMessage.tsx
```diff
@@ -70,6 +90,15 @@ export function BannerMessage(props: Props) {
           <MessageHeader>{props.messageHeader}</MessageHeader>
         )}
         <MessageText>{props.message}</MessageText>
+        {props.ctaText && props.ctaURL && (
+          <CTALink
+            href={props.ctaURL}
+            onClick={handleCTAOnClick}
+            textColor={props.textColor}
```
riodeuno, line 97: Does `textColor` show a warning in dev mode? If it does, shall we use the `$`[ prefix](https://styled-components.com/docs/api#transient-props)?

riodeuno (approved):

app/client/src/components/ads/BannerMessage.tsx
rahulramesha: checked that, it doesn't show a warning in dev mode. Learnt about this $[ prefix](https://styled-components.com/docs/api#transient-props) though :).

PR #14558, opened 2022-06-15
feature: GraphQL pagination server side support

### LINKED ISSUE #13830: [Feature]: [GraphQL] Implement pagination tab - server side changes only

### Summary

- Add server side changes to implement cursor pagination for GraphQL plugin.

.

sumitsum:
## Description
- This PR provides server-side support for pagination feature of GraphQL plugin. Two types of pagination is supported: limit offset based and cursor based. 
- In case a user has configured a query variable in both the query variables editor box and the pagination tab then a warning / hint message is returned to the client. Hint message is computed and returned at the time of query edit and at the time of query execution.
- Minor refactor: Some singleton pattern helper classes have been modified to allow for inheritance and extension. This was also suggested as a feedback on one of the earlier PRs. 

Fixes #13830 

## Type of change

## How Has This Been Tested?
- Manual
- TBD: JUnit TC will be added as part of a different PR.

## Checklist:

### DIFF
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ActionConfiguration.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ActionConfiguration.java
@@ -54,6 +54,14 @@ public class ActionConfiguration implements AppsmithDomain {
     String next;
     String prev;
 
+    /**
+     * This field is supposed to contain all action config data that may contain valid references to itself. This
+     * field has been excluded from dynamic binding path map computation in PageLoadActionUtilCEImpl.java so that any
+     * self referencing binding in this data path will not cause a cyclic dependency error. This field should ideally
+     * replace the usage of `next` and `prev` fields defined above. 
+     */
+    Object selfReferencingData;
+
     // DB action fields
 
     // JS action fields
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ActionConfiguration.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ActionConfiguration.java
@@ -11,6 +11,7 @@ import org.springframework.http.HttpMethod;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import static com.appsmith.external.constants.ActionConstants.DEFAULT_ACTION_EXECUTION_TIMEOUT_MS;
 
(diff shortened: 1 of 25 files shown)


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ActionConfiguration.java
```diff
@@ -54,6 +54,14 @@ public class ActionConfiguration implements AppsmithDomain {
     String next;
     String prev;
 
+    /**
+     * This field is supposed to contain all action config data that may contain valid references to itself. This
+     * field has been excluded from dynamic binding path map computation in PageLoadActionUtilCEImpl.java so that any
+     * self referencing binding in this data path will not cause a cyclic dependency error. This field should ideally
+     * replace the usage of `next` and `prev` fields defined above. 
+     */
+    Object selfReferencingData;
```
nidhi-nair, line 63: Instead of introducing another object to store data in, do you think we can store this information in `pluginSpecifiedTemplates` and have a list of paths in this location instead? That way we have a cleaner configuration for accessing data, and we'll have separated the use case of storing configuration related information and cyclical reference configuration.

This will come in useful especially wrt future migrations.

nidhi-nair (changes requested):

app/server/appsmith-plugins/graphqlPlugin/src/main/java/com/external/utils/GraphQLPaginationUtils.java
```diff
@@ -0,0 +1,215 @@
...
+            paginationData = getValueSafelyFromFormData((Map) actionConfiguration.getSelfReferencingData(),
+                    "paginationData.cursorBased", Map.class, null);
+        }
+
+        if (isEmpty(paginationData)) {
+            return paginationData;
+        }
+
+        HashMap<String, String> transformedPaginationData = new HashMap<String, String>();
+        if (PaginationType.PAGE_NO.equals(actionConfiguration.getPaginationType())) {
```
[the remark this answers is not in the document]
sumitsum: As discussed, we can take this up later based on future requirements. I have created an issue to track it: https://github.com/appsmithorg/appsmith/issues/15034

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ActionConfiguration.java
sumitsum: Done. I have replaced this identifier with `selfReferencingDataPaths` which contains a set of paths instead of the data itself.

nidhi-nair (approved):
LGTM

PR #14581, opened 2022-06-15
fix: Property pane components re-render

### LINKED ISSUE #6417: [Perf] Further Optimisations of PropertyPane Connections
PropertyPane Connections were optimised to an extent as part of the editing optimisations. But this can be further optimised with optimising other selectors and moving code to fetch link on click instead of updating every time.

### LINKED ISSUE #14580: [Task]: Some Property pane components re-renders when they are not supposed to be

### SubTasks

The following components are re-rendering when we make a small change to any property in the property pane
- PropertyPaneSidebar
- PropertyPaneView
- ProeprtyPaneConnections
- CodeEditor

aswathkk:
## Description

The following components were re-rendering when any property changes
- PropertyPaneSidebar
- PropertyPaneView
- ProeprtyPaneConnections

After inspection, I found that it was due to the selectors they were using. Modified the selector code to fix the re-rendering issue.

Fixes #14580 
Fixes #6417

## Type of change

## How Has This Been Tested?

- Manual

## Checklist:


rahulramesha (commented):
@aswathkk As discussed please verify changes with respect to `() => true` for selectors.

app/client/src/components/editorComponents/Debugger/hooks/debuggerHooks.ts
```diff
@@ -102,7 +104,19 @@ export const useSelectedEntity = () => {
 };
 
 export const useEntityLink = () => {
-  const dataTree = useSelector(getDataTree);
+  const dataTree = useSelector(getDataTree, (left, right) => {
```
SatishGandham, line 107: This is not the right solve for this.
We should not use use selector at all. We don't need the dataTree until the user click the link and its not used to render any element, so use can access the store directly in the returned function.

app/client/src/pages/Editor/Explorer/Widgets/useNavigateToWidget.ts
```diff
@@ -10,12 +10,20 @@ import { navigateToCanvas } from "./utils";
...
+import equal from "fast-deep-equal/es6";
+import { mapValues } from "lodash";
 
 const WidgetTypes = WidgetFactory.widgetTypes;
 
 export const useNavigateToWidget = () => {
   const params = useParams<ExplorerURLParams>();
-  const allWidgets = useSelector(getCurrentPageWidgets);
+
+  const allWidgets = useSelector(getCurrentPageWidgets, (left, right) => {
```
SatishGandham, line 21: Same as previous comment.

app/client/src/pages/Editor/PropertyPane/PropertyPaneConnections.tsx
```diff
@@ -191,8 +191,11 @@ const doConnectionsHaveErrors = (
 };
 
 const useDependencyList = (name: string) => {
-  const dataTree = useSelector(getDataTree);
-  const deps = useSelector((state: AppState) => state.evaluations.dependencies);
+  const dataTree = useSelector(getDataTree, () => true);
```
SatishGandham, line 194: Will this work when the dependencies change? 

For example, on text widget I change the value to {{Input1.text}}, will the dependency list be updated?

app/client/src/pages/Editor/PropertyPane/PropertyPaneView.tsx
```diff
@@ -46,7 +48,13 @@ function PropertyPaneView(
 ) {
   const dispatch = useDispatch();
   const { ...panel } = props;
-  const widgetProperties: any = useSelector(getWidgetPropsForPropertyPane);
+  const widgetProperties = useSelector(
+    getWidgetPropsForPropertyPane,
+    (left, right) => {
```
SatishGandham, line 53: 1. Use prev and next instead of left right. This is the convention followed in react and in Appsmith.
2. Can you explain what you are doing here.

SatishGandham (changes requested):

app/client/src/pages/Editor/PropertyPane/PropertyPaneConnections.tsx
aswathkk: @SatishGandham @rahulramesha This doesn't work when the dependency changes.

I will explore the right solution and update this.

app/client/src/pages/Editor/PropertyPane/PropertyPaneView.tsx
aswathkk: 1. okay. I will follow the convention.
2. getWidgetPropsForPropertyPane is returning some data that is not being used in this component. So, this one renders whenever some unwanted data changes. To resolve this, I pass a comparator function to `useSelector` which checks the equality between only the properties that are being used in the component.

SatishGandham: In that case, please write a selector that give it only the information it needs.

app/client/src/components/editorComponents/Debugger/hooks/debuggerHooks.ts
aswathkk: Correct.
This was also pointed out by @rahulramesha 
I will use a different approach.

app/client/src/components/editorComponents/Debugger/hooks/debuggerHooks.ts
```diff
@@ -102,46 +103,43 @@ export const useSelectedEntity = () => {
...
-      } else if (isJSAction(entity)) {
-        history.push(
-          jsCollectionIdURL({
-            applicationSlug,
-            pageSlug,
-            pageId,
-            collectionId: entity.actionId,
-          }),
+  const navigateToEntity = useCallback((name) => {
+    const dataTree = getDataTree(store.getState());
```
SatishGandham, line 113: @aswathkk , To maintain existing logic, you need the dataTree at the time the link was clicked. 

For this to work,  subsequent updates to datatree shouldn't change the existing links.

aswathkk: @SatishGandham I didn't get you.
According to previous logic, an update to dataTree will also update navigateToEntity function. Check the dependency list of useCallback in the previous case.
Is that something else you were talking about?

app/client/src/pages/Editor/PropertyPane/PropertyPaneConnections.tsx
```diff
@@ -190,9 +193,16 @@ const doConnectionsHaveErrors = (
...
+);
+
 const useDependencyList = (name: string) => {
-  const dataTree = useSelector(getDataTree);
-  const deps = useSelector((state: AppState) => state.evaluations.dependencies);
+  const dataTree = useSelector(getDataTreeWithOnlyId, equal);
+  const inverseDependencyMap = useSelector(
+    (state: AppState) => state.evaluations.dependencies.inverseDependencyMap,
+    equal,
+  );
```
SatishGandham, line 205: We don't need deep equal at L204. It changes only when it actually changes.
Please verify.

app/client/src/pages/Editor/PropertyPane/PropertyPaneConnections.tsx
```diff
@@ -190,9 +193,16 @@ const doConnectionsHaveErrors = (
...
 };
 
+const getDataTreeWithOnlyId = createSelector(getDataTree, (tree) =>
+  mapValues(pick(tree, ["ENTITY_TYPE", "widgetId", "actionId"])),
+);
+
 const useDependencyList = (name: string) => {
-  const dataTree = useSelector(getDataTree);
-  const deps = useSelector((state: AppState) => state.evaluations.dependencies);
+  const dataTree = useSelector(getDataTreeWithOnlyId, equal);
```
SatishGandham, line 201: Shouldn't this be getDataTreeWithOnlyIds

app/client/src/pages/Editor/PropertyPane/PropertyPaneView.tsx
```diff
@@ -39,14 +42,35 @@ export const excludeList: WidgetType[] = [
...
+  type: string;
+  widgetId: string;
+  widgetName: string;
+  displayName: string;
+};
+
+const getWidgetPropsForPropertyPaneView = createSelector(
+  getWidgetPropsForPropertyPane,
+  (props) =>
+    pick(props, [
```
SatishGandham, line 55: Check the time taken by pick in the chrome profiler.

app/client/src/pages/Editor/PropertyPane/PropertyPaneView.tsx
```diff
@@ -39,14 +42,35 @@ export const excludeList: WidgetType[] = [
...
 ];
 
+type WidgetPropertiesForPropertyPaneView = {
+  type: string;
+  widgetId: string;
+  widgetName: string;
+  displayName: string;
+};
+
+const getWidgetPropsForPropertyPaneView = createSelector(
```
SatishGandham, line 52: See if moving the selector to selectors folder, or a separate file in the same folder makes sense.

SatishGandham (changes requested):

app/client/src/pages/Editor/PropertyPane/PropertyPaneConnections.tsx
aswathkk: Yeah. 💯  I will make the change.

app/client/src/components/editorComponents/PropertyPaneSidebar.tsx
```diff
@@ -61,13 +59,6 @@ export const PropertyPaneSidebar = memo((props: Props) => {
   const keepThemeWhileDragging =
     prevSelectedWidgetId.current === undefined && shouldNotRenderPane;
 
-  const selectedWidgets = useMemo(
```
rahulramesha, line 64: Was thinking about this and remembered why `selectedWidgets` was accessed from `canvasWidgets`.
This was to make sure that the `selected widgetIds` still exists in `canvasWidgets`. This was a mechanism to verify if the selected widgets had not been cut or deleted or the user had selected widgets from previous pages.

To fix this you could create a `selector` that had access to `canvasWidgets`, from there you could filter out the `widgetIds` that did not exist in `canvasWidgets`. You could use such selector here with an `equal` check.

rahulramesha (changes requested):

app/client/src/selectors/propertyPaneSelectors.tsx
```diff
@@ -56,6 +57,35 @@ export const getWidgetPropsForPropertyPane = createSelector(
...
+      "widgetName",
+      "displayName",
+    ]) as WidgetPropertiesForPropertyPaneView,
+);
+
+export const selectedWidgetsPresentInCanvas = createSelector(
+  getWidgets,
+  getSelectedWidgets,
+  // Make sure the widgets in selectedWidgets still exists in the canvasWidgets
+  (canvasWidgets, selectedWidgets) => {
```
SatishGandham, line 82: How about iterating over selected widgets and pop the array if its not on the canavs?
We won't need compact.

SatishGandham (changes requested):
Looks good except for few minor clarification and may be changes.

app/client/src/pages/Editor/PropertyPane/PropertyPaneConnections.tsx
aswathkk: Yes. Reference changes.

app/client/src/pages/Editor/PropertyPane/PropertyPaneView.tsx
```diff
@@ -46,7 +47,10 @@ function PropertyPaneView(
 ) {
   const dispatch = useDispatch();
   const { ...panel } = props;
-  const widgetProperties: any = useSelector(getWidgetPropsForPropertyPane);
+  const widgetProperties = useSelector(
+    getWidgetPropsForPropertyPaneView,
```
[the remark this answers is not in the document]
aswathkk: It is hardcoded to 4 keys ["type", "widgetId", "widgetName", "displayName"]

app/client/src/pages/Editor/PropertyPane/PropertyPaneConnections.tsx
aswathkk: I tested it by replacing `equal` function with a custom function and found that reference changes (using `===` operator)

SatishGandham (dismissed):

PR #14636, opened 2022-06-17
ci: Use buildjet runners for Cypress tests, and fewer of them

sharat87:
Use buildjet runners for Cypress tests, and fewer of them.

### DIFF
--- a/.github/workflows/test-build-docker-image.yml
+++ b/.github/workflows/test-build-docker-image.yml
@@ -761,7 +761,7 @@ jobs:
       (github.event_name == 'pull_request_review' &&
       github.event.review.state == 'approved' &&
       github.event.pull_request.head.repo.full_name == github.repository))
-    runs-on: buildjet-8vcpu-ubuntu-2004
+    runs-on: buildjet-4vcpu-ubuntu-2004
     defaults:
       run:
         working-directory: app/client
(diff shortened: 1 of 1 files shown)


SatishGandham (approved):
Is cypress capable of using 8 vcpus?
I think 4vpus should be enough.

If the goal is to make the test faster, setting chrome process made perf tests faster by about 20%.

Changes look good code wise.

sharat87:
@SatishGandham, thanks. It's mostly to free up GitHub's own runners so that they are available for other workflows, but, hopefully it'll speed them up too.

Changed to 4vcpu, let's see. :)

mohanarpit (approved):

mohanarpit:
@SatishGandham How did you set the Chrome process priority to max in Cypress? Can you link to the snippet of workflow/script?

SatishGandham:
@mohanarpit 
Find all chrome process and use renice to update their priority. Sudo works on these runners.
https://github.com/appsmithorg/appsmith/blob/49980ab2cfe799d8d1d892c230046508c61439ac/app/client/perf/src/perf.js#L67

sharat87:
Doesn't Cypress constantly start new Chrome processes? Should we do that renice command in a background loop? Something like:

```bash
(
  while true; do
    echo-chrome-pids | xargs -I% renice -20 %
    sleep 5
  done
) &
```

May be?

SatishGandham:
@sharat87 
It would start chrome using some command. So we can call the setPriority function after that. Someone from QA team will know better.

Else that while loop also works, i tried it before.

cc: @YatinChaubal @Aishwarya-U-R

sharat87:
@SatishGandham, looks like a longer discussion, so I'm going to go ahead and merge this. We'll have more data once we see this run a few times.

Issue #14503, opened 2022-06-13
[Bug]: JS object is in an error state but no error is displayed on debugger or browser console


ramsaptami:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

JS object goes into error during execution but no error is displayed either on errors tab on the debugger or on browser console.

### Steps To Reproduce

1.  Go to JS object at [this](https://release.app.appsmith.com/app/scratch-pad/js-error-issue-62a7690dd901b06688ea14fb/edit/jsObjects/62a7690dd901b06688ea1513) location and run function `myFun1`
2. Observe the execution error 

### Public Sample App

https://release.app.appsmith.com/app/scratch-pad/js-error-issue-62a7690dd901b06688ea14fb/edit/jsObjects/62a7690dd901b06688ea1513

### Version

Cloud

ohansFavour:
Two issues here

1. Errors found while executing a js function don't get logged in the debugger.
2. `myVar2: Input1.text.toString()` causes parsing of the js object to be wrong.

PR #14468, opened 2022-06-10
fix: reset meta to default value

Rishabh-Rathod:
## Description

Earlier when we reset a widget, we used to delete all the meta properties stored in the meta reducer. 

- Expectation for reset was always to bring the widget to its initial stage. 
That is meta values should get reset to the current default value. This PR resets widget meta values to default values.

Fixes #14198

## Type of change

- Bug fix (non-breaking change which fixes an issue)

#### Cypress

- [x] [[Test Case] : Verify “Multiselect widget” reset to its default value on submit success](https://github.com/appsmithorg/TestSmith/issues/1878)
- [x] [[Test Case] : Verify “MultiSelect widget” behavior on manually deselecting the items inside widget](https://github.com/appsmithorg/TestSmith/issues/1879)
- [x] [[Test Case] : verify “Text widget” fetch data from “multiselect widget” on setting “text widget’s” property as {{ {{MultiSelect[n].selectedOptionValues}} }}](https://github.com/appsmithorg/TestSmith/issues/1880)
- [ ] https://github.com/appsmithorg/TestSmith/issues/1898

#### Jest 

- [x] metaReducer reset widget add more test case

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented on my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #14658, opened 2022-06-18
fix: multiple storeValue action completion track

### LINKED ISSUE #14653: [Bug]: Consecutive storeValue calls don't work and only the first call is executed

### Description

Consecutive `storeValue()` does not work anymore - regressed issue. Only the first call goes through

### Steps To Reproduce

1. Use consecutive `storeValue()` bound either in a widget JS or in a `promise.all()` and observe that the code fails and saves only the first `storeValue()` call

https://release.app.appsmith.com/app/scratch-pad/iframe-629d9ca4b1677a11502ab3e2/edit

### Public Sample App

https://release.app.appsmith.com/app/scratch-pad/iframe-629d9ca4b1677a11502ab3e2/edit

### Version

Cloud

Rishabh-Rathod:
### Problem 

`yield take(ReduxActionTypes.UPDATE_APP_STORE_EVALUATED)` inside storeValueLocally saga can catch previous storeValue action and complete the latest call too.
This was not expected behaviour with promisified storeValue as we need to track completion of each storeValue action separately.

Fixes #14653

## Type of change

## How Has This Been Tested?
Manual & Cypress

- [x] https://github.com/appsmithorg/TestSmith/issues/1893

## Checklist:

### DIFF
--- a/app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/ActionExecution/StoreValue_spec.ts
+++ b/app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/ActionExecution/StoreValue_spec.ts
@@ -1,0 +1,61 @@
+import { ObjectsRegistry } from "../../../../support/Objects/Registry";
+
+const { AggregateHelper: agHelper, JSEditor: jsEditor } = ObjectsRegistry;
+
+describe("storeValue Action test", () => {
+  before(() => {
+    //
+  });
+
+  it("1. Running consecutive storeValue actions and await", function() {
+    const jsObjectBody = `export default {
+      myFun1: () => {
+        let values =
+          [
+            storeValue('val1', 'number 1'),
+            storeValue('val2', 'number 2'),
+            storeValue('val3', 'number 3'),
+            storeValue('val4', 'number 4')
+          ]
+        return Promise.all(values)
+          .then(() => {	
+            showAlert(JSON.stringify(appsmith.store))
+        })
+          .catch((err) => { 
+            return showAlert('Could not store values in store ' + err.toString());
+          })
+      }
+    }`;
+
+    jsEditor.CreateJSObject(jsObjectBody, {
+      paste: true,
+      completeReplace: true,
+      toRun: false,
+      shouldCreateNewJSObj: true,
+    });
+
+    agHelper.AssertAutoSave();
+
+    jsEditor.EditJSObj(jsObjectBody);
+
+    agHelper.AssertAutoSave();
+
+    // running twice due to bug
+    Cypress._.times(2, () => {
+      cy.get(jsEditor._runButton)
+        .first()
+        .click()
+        .wait(3000);
+    });
+
+    agHelper.ValidateToastMessage(
+      JSON.stringify({
+        val1: "number 1",
+        val2: "number 2",
+        val3: "number 3",
+        val4: "number 4",
+      }),
+      2,
+    );
+  });
+});
--- a/app/client/src/sagas/ActionExecution/StoreActionSaga.ts
+++ b/app/client/src/sagas/ActionExecution/StoreActionSaga.ts
@@ -10,7 +10,7 @@ import { getAppStoreData } from "selectors/entitiesSelector";
 import { StoreValueActionDescription } from "entities/DataTree/actionTriggers";
 import { getCurrentGitBranch } from "selectors/gitSyncSelectors";
 import { getCurrentApplicationId } from "selectors/editorSelectors";
-import { ReduxActionTypes } from "@appsmith/constants/ReduxActionConstants";
+import { updateAppStoreEvaluated } from "actions/pageActions";
 
 export default function* storeValueLocally(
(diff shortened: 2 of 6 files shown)


app/client/src/sagas/ActionExecution/StoreActionSaga.ts
```diff
@@ -34,10 +34,22 @@ export default function* storeValueLocally(
...
       text: `store('${action.key}', '${action.value}', false)`,
     });
   }
-  yield take(ReduxActionTypes.UPDATE_APP_STORE_EVALUATED);
+  /* It is possible that user calls multiple storeValue function together, in such case we need to track completion of each action separately
+  We use uniqueRequestId to differentiate each storeValueAction here.
+  */
+  while (true) {
+    const returnedAction: StoreValueActionDescription = yield take(
+      updateAppStoreEvaluated().type,
```
ohansFavour, line 47: Can we use `ReduxActionTypes.UPDATE_APP_STORE_EVALUATED` instead of updateAppStoreEvaluated().type?.

Reasons: 

1. Most times during code inspection, we do a string search of the `redux action type` to see how it is used within the codebase.

2. For consistency across the codebase.

ohansFavour: This looks good. Few more points

1. Can we add a cypress test too?

2. Can we also in this PR use the same pattern for other actions? e.g `resetWidget`

ohansFavour (commented):
This looks good. Few more points

1. Can we add a cypress test too?

2. Can we also in this PR use the same pattern for other actions? e.g `resetWidget`

app/client/src/sagas/ActionExecution/StoreActionSaga.ts
Rishabh-Rathod: sure, i will revert this. 

Intention was to slowly adopt usage of action creators everywhere but i realise that for `take` it is not at all intuitive. 

Problem, i have seen with directly using `ReduxActionTypes.UPDATE_APP_STORE_EVALUATED` are mainly when we do `put` or `dispatch` directly without actionCreator and avoid the type check of payload. ( we had bugs in past due to this practice. )

Hopefully, we will come up with better solution for this in future.

Rishabh-Rathod:
> 2\. Can we also in this PR use the same pattern for other actions? e.g `resetWidget`

I will also fix this for `resetWidget`. That was the only action i found which has similar implementation and issue but that will go in as separate PR so that we could ship `storeValue` fix ASAP. 

- `resetWidget` will again need more testing 
( there is a core change in resetWidget functionality in this PR https://github.com/appsmithorg/appsmith/pull/14468 )

According to new test plan of it, i would decide how to add resetWidget fix. ( merging the testing effort for reset )

Also, problem is less severe for `resetWidget`. hence, we could separately ship them.
@ohansFavour

app/client/src/sagas/ActionExecution/StoreActionSaga.ts
Rishabh-Rathod: done

ramsaptami (approved):

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/ActionExecution/StoreValue_spec.ts
```diff
@@ -0,0 +1,61 @@
...
+      toRun: false,
+      shouldCreateNewJSObj: true,
+    });
+
+    agHelper.AssertAutoSave();
+
+    jsEditor.EditJSObj(jsObjectBody);
+
+    agHelper.AssertAutoSave();
+
```
Aishwarya-U-R, line 42: Firstly - please move this case into ClientSidetests/Binding/Promises - which already has promise related tests. (by appending right tc number)
And - CreateJSObj has AssertAuotSave - so this can be removed from line 37.
And why are we editing JSOBj here?

ramsaptami (approved):

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/ActionExecution/StoreValue_spec.ts
Rishabh-Rathod: > Firstly - please move this case into ClientSidetests/Binding/Promises - which already has promise related tests. (by appending right tc number)

@Aishwarya-U-R  we will need to test each action like `storeValue`, `resetWidget` separately.

This issue was related to storeValue particularly not handling promise well. ( not a general promise issue )

We will be add more test cases here in future specific to storeValue ( which maybe not be related to promise ) hence a separate file.

> And - CreateJSObj has AssertAuotSave - so this can be removed from line 37.
And why are we editing JSOBj here?

Editing JSObject is not required here, there is a bug in JSObject https://github.com/appsmithorg/appsmith/issues/14503 which i was trying to bypass. 

Current test code, is able to reliably bypass it. I will add modify this test case once we fix above issue.

PR #14668, opened 2022-06-19
fix: prevent cyclic dependency when switching to js mode

### LINKED ISSUE #14669: [Bug]: Cyclic Dependency Errors with Google sheets Form Controls

### Description

A cyclic dependency error is shown when switching from Component mode to JSON mode for any Form control that contains dynamic bindings, This is due to the dynamic binding pathlist not being updated when we make a switch. This incorrect binding path list results in an issue with the evaluations flow.

### Steps To Reproduce

1. Go to new Google sheets.
2. Switch to Fetch many, populate the sheet url and sheet name
3. Go to the sorting component and add dynamic bindings within it e.g. {{""}}
4. Then switch to JSON mode
5. You'll observe a cyclic dependency error message being displayed.

### Public Sample App

_No response_

### Version

UQI branch version.

ohansFavour:
## Description

When translating diffs, if an array value changes to a string, all `MemberExpressions` of the array should be removed from the dependency map.

For example.

- Previous

```
ENTITY.options = [
{
label: "name",
value: {{name}}
},
{
label: "name",
value: {{name2}}
}
]
```

- Current (after switching to js mode)

```
ENTITY.options = "[
{
label: "name",
value: {{name}}
},
{
label: "name",
value: {{name2}}
}
]"
```

ENTITY.options[0] and ENTITY.options[1] (and their child nodes eg, ENTITY.options[0].value) should be removed from the dependencyMap

Fixes #14669 

## Type of change

## How Has This Been Tested?

Jest

## Checklist:

### DIFF
--- a/app/client/src/workers/evaluationUtils.ts
+++ b/app/client/src/workers/evaluationUtils.ts
@@ -182,11 +182,41 @@ export const translateDiffEventToDataTreeDiffEvent = (
           },
         ];
       } else if (rhsChange || lhsChange) {
-        result.event = DataTreeDiffEvent.EDIT;
-        result.payload = {
-          propertyPath,
-          value: difference.rhs,
-        };
+        result = [
+          {
+            event: DataTreeDiffEvent.EDIT,
+            payload: {
+              propertyPath,
+              value: difference.rhs,
+            },
+          },
+        ];
+        /**
+         * If lhs is an array/object
+         * Add delete events for all memberExpressions
+         */
+        if (Array.isArray(difference.lhs)) {
+          difference.lhs.forEach((diff, idx) => {
+            (result as DataTreeDiff[]).push({
+              event: DataTreeDiffEvent.DELETE,
+              payload: {
+                propertyPath: `${propertyPath}[${idx}]`,
+              },
+            });
+          });
+        }
+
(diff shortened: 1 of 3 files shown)


ApekshaBhosale (commented):
won't this create a problem in the case where we are trying to add a dependency for Api1.data[0]? just wondering @ohansFavour

app/client/src/workers/evaluationUtils.ts
```diff
@@ -182,11 +182,41 @@ export const translateDiffEventToDataTreeDiffEvent = (
...
+              propertyPath,
+              value: difference.rhs,
+            },
+          },
+        ];
+        /**
+         * If lhs is an array/object
+         * Add delete events for all memberExpressions
+         */
+        if (Array.isArray(difference.lhs)) {
```
ohansFavour, line 198: @ApekshaBhosale, here if the left-hand side is an array/object definitely, the rhs is a string (because of the check in line 184)

ApekshaBhosale (approved):

Rishabh-Rathod (approved):
LGTM, Let's create a task to ideally detect such changes using `dynamicBindingPathList` as discussed.

ApekshaBhosale (approved):

PR #14676, opened 2022-06-20
feat: Adding policies on workspace creation + BaseRepository user -> permission mapping

trishaanand:


app/server/appsmith-server/src/main/java/com/appsmith/server/domains/UserInGroup.java
```diff
@@ -2,16 +2,22 @@
 
 import lombok.Data;
 import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
 
 @EqualsAndHashCode
+@Getter
+@Setter
```
sidhantgoel, line 11: Already @Data is there @Getter and @Setter not required

sidhantgoel (approved):

PR #14736, opened 2022-06-22
feat: Update test cases with the new RBAC model

trishaanand:
- ActionServiceCE_Test.java done except `checkNewActionAndNewDatasourceAnonymousPermissionInPublicApp` which tests for public apps.

TODO :
- Public application functionality.


nayan-rafiq (approved):
LGTM

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/BaseAppsmithRepositoryImpl.java
```diff
@@ -109,11 +109,20 @@ public static final Criteria notDeleted() {
...
     public static final Criteria userAcl(Set<String> permissionGroups, AclPermission permission) {
+
+        // Check if the permission is being provided by any of the permission groups
         Criteria permissionGroupCriteria = Criteria.where(fieldName(QBaseDomain.baseDomain.policies))
-                .elemMatch(Criteria.where("permissionGroups").all(permissionGroups)
+                .elemMatch(Criteria.where("permissionGroups").in(permissionGroups)
                         .and("permission").is(permission.getValue()));
 
-        return permissionGroupCriteria;
+        // Also check if the permission is being provided by being assigned to ANONYMOUS_USER, aka
```
sidhantgoel, line 118: I was thinking, get current permission groups function will include anonymous permission group as well that we can create per tenant? so it will avoid special condition for anonymous user and keeping the users field.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserServiceCEImpl.java
```diff
@@ -747,10 +743,11 @@ public Mono<List<User>> inviteUsers(InviteUsersDTO inviteUsersDTO, String origin
                     return Mono.empty();
                 });
 
-        return bulkAddUserResultMono.then(inviteUsersFlux.collectList());
+        return Mono.zip(bulkAddUserResultMono, sendAnalyticsEventMono).then(inviteUsersMono);
```
sidhantgoel, line 746: we don't have to wait for sendAnalyticsEventMono? should we?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserWorkspaceServiceCEImpl.java
```diff
@@ -524,8 +524,9 @@ public Mono<List<UserAndGroupDTO>> getWorkspaceMembers(String workspaceId) {
                 .flatMapMany(Flux::fromIterable)
                 .map(UserAndGroupDTO::getUsername)
                 .collect(Collectors.toSet())
-                .flatMapMany(usernames -> userRepository.findAllByEmails(usernames, AclPermission.READ_USERS))
-                .collectMap(User::getUsername).cache();
+                .flatMapMany(usernames -> userRepository.findAllByEmails(usernames))
```
sidhantgoel, line 527: should we not have read permission on users to to list them?

trishaanand: Yeah. We most probably have to add a permission on Tenant to list all users to achieve this.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserServiceCEImpl.java
trishaanand: This zip would wait for both of then to finish and then trigger the inviteUsersMono

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/BaseAppsmithRepositoryImpl.java
trishaanand: Makes sense.

PR #14738, opened 2022-06-22
feat: filter error messages from hidden widgets

### LINKED ISSUE #12309: [Feature]-[2500]:Misreporting errors for bindings

### Summary

When a user creates a binding that is dependant on an async query, or user input based event, it reports an error if the query has not been executed, or input hasn't been captured.

A user will open their editor, and see that there are a large number of errors, however these errors are all in non-reachable states. For instance, A table containing orders of a particular user is hidden till a user is selected from a list. This is expected behaviour, and this should not be reported as an error till we are sure it is an error (when a user is selected from the list)

This exact behaviour inflates the error count, and when the user sees their complete app, functioning as intended, reporting errors, it makes Appsmith feel unstable.

To address this

- Errors for hidden elements and Modals should not show up as errors till they are made visible or interactable. Users often hide these kinds of values behind tabs, hidden elements, Modals, etc.
 

-  This leads to user confusion
- Users are put off by/concerned with a large error list 
- This makes appsmith seem unstable

eco-monk:
## Description
Hide error messages when:
- Widget is hidden via container
- Widget is hidden via tabs
- Widget is hidden via modal
- Widget is hidden via `Visible` property

Note: Exception when `isVisible` binding has an error. All errors in the widget will be shown in that case.

https://www.loom.com/share/d48a19706e634e9883a2e79aa5ba7fd1

Fixes #12309

## Type of change

## Checklist:


arunvjn:
@eco-monk Are we tackling API errors here ? For example, if i am in the cavas and the API isn't visible, are we planning these suppress the errors from it ?

app/client/src/selectors/debuggerSelectors.tsx
```diff
@@ -1,17 +1,73 @@
...
     if (hideErrors) return {};
 
-    return errors;
+    const filteredErrors = Object.fromEntries(
+      Object.entries(errors).filter(([, error]) => {
+        const entity = error?.source?.name && dataTree[error.source.name];
+        // filter error - when widget or parent widget is hidden
+        // parent widgets e.g. modal, tab, container
+        if (entity && isWidget(entity)) {
+          return entity.isVisible
```
arunvjn, line 25: Just for my understanding,  we need this because parent widget being invisible doesn't mean that we mark child widget as not visible.

akash-codemonk:
There is one scenario with the chart widget where if we have an error in the custom fusion chart field it remains even when we switch to say a line chart.


https://user-images.githubusercontent.com/67054171/175488844-7f7ce162-7372-49c8-8539-c7c4c20bf33f.mov

app/client/src/selectors/debuggerSelectors.tsx
eco-monk: Yes, parent widget having `isVisible = false` doesn't reflect in the children's visibility state.

e.g. in dataTree
```
{
   container1: {
      isVisible: false
   },
   input1: {
     parent: "container1"
     isVisible: true, // will remain true
   }
}
```

eco-monk: I was exploring to see, if this can be changed.
But trying to keep the child updated based on the parent's visibility state seems inappropriate.

Especially when isVisiblity can have a user written binding.

eco-monk:
> @eco-monk Are we tackling API errors here ? For example, if i am in the cavas and the API isn't visible, are we planning these suppress the errors from it ?

@arunvjn 
That is not handled here. Only errors from hidden widgets are handled.
Although, it's an interesting case to consider.

The upside being, cleaning up logs. 
And the downside being, not having a way to know if any API has failed, unless the user visits the API page.

Perhaps we can consider having log categories or something along those lines?

Will bring this up with @kocharrahul7.

eco-monk:
> There is one scenario with the chart widget where if we have an error in the custom fusion chart field it remains even when we switch to say a line chart.
> 
>  chart-widget.mov

@akash-codemonk 
That is not handled here, this seems like a bug in the chart widget.
Here, we're focusing on filtering errors from hidden widgets in canvas.

app/client/src/selectors/debuggerSelectors.tsx
eco-monk: This needs some design and product thought. (overwriting isVisiblity)
Because, the user needs to be informed of the change in behaviour.
If not, they'd find their evaluated valued being replaced internally by the system, without much explanation.

Will create a separate issue.

arunvjn (dismissed):

eco-monk:
Working on a newly identified case.
Show errors for widget, when `isVisible` binding itself has a error.

ramsaptami:
Tested ok on DP

eco-monk:
Discussed with @ramsaptami. A new case to handle.
On hiding the widget, trigger field errors are not filtered.

ramsaptami:
Edge case identified on this PR but is an existing issue: https://www.loom.com/share/8ecd19e4dad045e3b2f271fceffe07d9
Issue: When a triggered field returns an error and this widget is inside a hidden container and then deleted, then the error of this deleted widget persists on the debugger
@eco-monk will raise this issue - DP testing is completed

eco-monk:
Created #15213

akash-codemonk (approved):

PR #14762, opened 2022-06-23
feat: Add keyType to git import key generation flow

### LINKED ISSUE #14761: [Task]: Add keyType support for git import key generation flow

### SubTasks

The api used in case of import is “v1/git/import/keys”; because application id is not present while importing an application, this api was created

AnaghHegde:
## Description

> Import flow is blocked: The api used in case of import is “v1/git/import/keys”; because application id is not present while importing an application, this api was created.
The problem is that this API doesn’t support keyType to be passed from FE. For Azure, and any other service, except GitHub, RSA is a valid key type.

Fixes #14761 

## Type of change

## How Has This Been Tested?

- Junit
- Locally

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/GitControllerCE.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/GitControllerCE.java
@@ -196,8 +196,8 @@ public class GitControllerCE {
     }
 
     @GetMapping("/import/keys")
-    public Mono<ResponseDTO<GitAuth>> generateKeyForGitImport() {
-        return service.generateSSHKey()
+    public Mono<ResponseDTO<GitAuth>> generateKeyForGitImport(@RequestParam(required = false) String keyType) {
+        return service.generateSSHKey(keyType)
                 .map(result -> new ResponseDTO<>(HttpStatus.OK.value(), result, null));
     }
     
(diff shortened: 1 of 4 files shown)


abhvsn (dismissed):

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/GitControllerCE.java
```diff
@@ -196,8 +196,8 @@ public Mono<ResponseDTO<String>> createConflictedBranch(@PathVariable String def
     }
 
     @GetMapping("/import/keys")
-    public Mono<ResponseDTO<GitAuth>> generateKeyForGitImport() {
-        return service.generateSSHKey()
+    public Mono<ResponseDTO<GitAuth>> generateKeyForGitImport(@RequestParam(required = false) String keyType) {
```
IAmAnubhavSaini, line 199: You can use "ECDSA" as default value.

abhvsn (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/GitControllerCE.java
IAmAnubhavSaini: @AnaghHegde 

Here: https://github.com/appsmithorg/appsmith/blob/1c07db8e8077da5fb9c8aff6cb7be6cf5a3f77a8/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java#L2020 what happens when the value is null, will it be able to handle that.

IAmAnubhavSaini (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/GitControllerCE.java
AnaghHegde: This is handled in the DeployKeyGenerator. When the value is null the default(ECDSA) key is generated.

abhvsn (approved):

Issue #14791, opened 2022-06-24
Refactor Git API's to remove appId from the path


AnaghHegde:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SubTasks

- Move the appId to request param

PR #14800, opened 2022-06-24
chore: Refactor git api to include application id as request param instead of the path variable

AnaghHegde:
## Description

> This is related #10856. In order to release the lock used in git for error scenarios, we are going to use the `GlobalExceptionHandler`. We need the appId to release the lock from redis. It would be hard and error prone to extract the appId from the path as it is not consistent. Hence moving the applicationId to request param. 

Fixes https://github.com/appsmithorg/appsmith/issues/14791

## Type of change

- Refactor 
- 
## How Has This Been Tested?

- Locally

## Checklist:


abhvsn:
@AnaghHegde does it make sense to keep the defaultApplicationId as a path param in a structured manner instead of moving this to query param? Today, as the defaultApplicationId don't have any structure which is causing a problem for us to extract the same from the URL. I'm proposing this because query params are not fixed part of URL path, though we can make this compulsory but still if this is possible with path param I'll go with that. 
We can include the structure by adding a prefix like: 
`{base_url}/{path_prefix}/application/{defaultApplicationId}/{remainingPathIfAny}`

From this path now you can extract the defaultApplicationId from `ServerWebExchange.getRequest().getPath()`.

Also this needs changes from client side as well can you please update the client side URLs in the same PR?

AnaghHegde:
@abhvsn We discussed the approach you mentioned in the last catchup. I thought we agreed to go with the query param approach. Client side changes are already present in this param with the query param approach.

abhvsn:
Yes I agree, but I was trying to still achieve this with path param with the structured path approach.

app/client/src/api/GitSyncAPI.tsx
```diff
@@ -97,39 +106,49 @@ class GitSyncAPI extends Api {
     const queryParams = {} as { pruneBranches?: boolean };
     if (pruneBranches) queryParams.pruneBranches = true;
     return Api.get(
-      `${GitSyncAPI.baseURL}/branch/${applicationId}`,
+      `${GitSyncAPI.baseURL}/list/branch?applicationId=${applicationId}&`,
       queryParams,
```
abhvsn, line 110: @AnaghHegde can we add all the query params in `queryParams` object. Also, can we replicate the same strategy in other places to maintain consistency?

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/GitControllerCE.java
```diff
@@ -59,11 +65,11 @@ public Mono<ResponseDTO<Map<String, GitProfile>>> saveGitProfile(@RequestBody Gi
                 .map(response -> new ResponseDTO<>(HttpStatus.OK.value(), response, null));
     }
 
-    @PutMapping("/profile/{defaultApplicationId}")
-    public Mono<ResponseDTO<Map<String, GitProfile>>> saveGitProfile(@PathVariable String defaultApplicationId,
+    @PutMapping("/profile")
+    public Mono<ResponseDTO<Map<String, GitProfile>>> saveGitProfile(@RequestParam String applicationId,
```
abhvsn, line 69:
```suggestion
    public Mono<ResponseDTO<Map<String, GitProfile>>> saveGitProfile(@RequestParam(name = FieldName.APPLICATION_ID) String defaultApplicationId,
```

abhvsn: This can be implemented in other places as well to keep the nomenclature consistent.

abhvsn (commented):
Minor comments added to improve the code readability.

mohanarpit:
For my context, can someone help me understand this change? Why is it hard to extract the `applicationId` from the URL in the `GlobalExceptionHandler` ? Moving the `applicationId` to a query parameter doesn't seem like an elegant solution. Hence, we should be very careful about this change. We will be going against the grain on REST API design best practices with this.

mohanarpit:
> As stated earlier we can use {base_path}/app/{appId}/{other_path_param} instead of moving the appId to query param.

I agree with you here. We should try to re-use the path parameters itself instead of moving to a query parameter.

abhvsn (commented):
Minor comment rest LGTM

app/client/src/api/GitSyncAPI.tsx
```diff
@@ -97,43 +100,52 @@ class GitSyncAPI extends Api {
...
   }
 
   static disconnectGit({ applicationId }: { applicationId: string }) {
-    return Api.post(`${GitSyncAPI.baseURL}/disconnect/${applicationId}`);
+    return Api.post(`${GitSyncAPI.baseURL}/disconnect/app/${applicationId}`);
   }
 
   static importApp(payload: ConnectToGitPayload, workspaceId: string) {
-    return Api.post(`${GitSyncAPI.baseURL}/import/${workspaceId}`, payload);
+    return Api.post(`${GitSyncAPI.baseURL}/import/app/${workspaceId}`, payload);
```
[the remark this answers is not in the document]
AnaghHegde: Good catch. Let me fix this.

abhvsn (dismissed):

IAmAnubhavSaini (approved):

Parthvi12 (approved):
Cypress tests changes LGTM.

Parthvi12 (approved):

Issue #14836, opened 2022-06-27
[Bug]: App crashing on navigating to error pane within JS object


rimildeyjsr:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Opening the JS object and then clicking on the error tab crashes the whole app.

On investigating, it was found that the EntityLink is expecting the name as a string, but receiving it as {label: string, array: <string>}

Seen on only one app instance - I can provide the exported json

### Steps To Reproduce

1. Go to JS object
2. click on error tab
3. app crashes

### Version

1.6.6

ApekshaBhosale:
JS Object to add to get this error 

`export default {
	check: () => {
		let newConfig = Fetch_Label_Config.data;
		if (!this.labelExists(Table1.selectedRow.name)) {
			newConfig = this.addLabelToConfig(Table1.selectedRow.name, edit_desc.text, edit_color.text, edit_pod.selectedOptionValue, edit_is_pod.isSwitchedOn)
		}
		console.log(newConfig);
		newConfig = this.moveLabelPod(newConfig);
		newConfig = this.renameConfigLabel(newConfig);
		newConfig.runners[0].issue.labels = Utils.sortPods(newConfig.runners[0].issue.labels);
		if (edit_is_pod.isSwitchedOn && !this.getPodLabels().includes(Table1.selectedRow.name)) {
			newConfig.runners[0].issue.labels[edit_name.text] = {
				conditions: [],
				requires: 1
			}
		}
		return newConfig;
	},
	getPodLabels: () => {
		console.log(Fetch_Label_Config.data.runners[0])
		return Object.keys(Fetch_Label_Config.data.runners[0].issue.labels);

	},
	getPodForLabel: (label) => {
		const config = Fetch_Label_Config.data;
		const runners = config.runners;
		const pods = this.getPodLabels();
		let foundPod;
		pods.map((podLabel) => {
			const foundLabel = runners[0].issue.labels[podLabel].conditions.find((condition) => {
				return condition.label === label && podLabel !== label
			})
			if (foundLabel) {
				foundPod = podLabel
			}
		});
		return foundPod;
	},
	removeLabelFromPods: (label, config = Fetch_Label_Config.data) => {
		const pods = this.getPodLabels();
		pods.map((podLabel) => {
			config.runners[0].issue.labels[podLabel].conditions = 
				config.runners[0].issue.labels[podLabel].conditions.filter((condition) => {
				return label === podLabel || condition.label !== label
[...]

PR #14842, opened 2022-06-27
fix: JsObject parsing

Rishabh-Rathod:
## Description

When JSObject function has code for variable declaration like below. In such cases, those properties also gets added as JSObject property which is not expected.  

```
newConfig.runners[0].issue.labels[edit_name.text] = {
				conditions: [],
				requires: 1,
                                 testFunc: () => { 
					console.log()
				}
			}
```

We will need to make sure to only parse JSobject's properties and not internal object property or function, this PR changes logic to do so.

Screenshot 2022-06-27 at 6 45 37 PM

### Solution

- Change the parsing logic to track variableDeclarator and check for JSObject being declared and find its properties using the AST Node.
- As we are exactly checking JSobject Node we won't have any incorrect properties defined or any property missed.

Fixes https://github.com/appsmithorg/appsmith/issues/14836

## Type of change

#### Jest test
- [x] Add variable declaration inside JSObject's function and verify AST parsing

#### Cypress test
- [ ] Add variable declaration inside JSObject's function and verify async function list
- [ ] [Nested async code in an array should not appear on the settings tab of a JS object](https://github.com/appsmithorg/TestSmith/issues/1919)
- [ ] [Nested async code in an object should not appear on the settings tab of a JS object](https://github.com/appsmithorg/TestSmith/issues/1918)

## Checklist:

### DIFF
--- a/app/client/src/workers/ast.test.ts
+++ b/app/client/src/workers/ast.test.ts
@@ -308,7 +308,6 @@ describe("parseJSObjectWithAST", () => {
       },
     ];
     const resultParsedObject = parseJSObjectWithAST(body);
-    console.log(resultParsedObject, parsedObject);
     expect(resultParsedObject).toStrictEqual(parsedObject);
   });
 });
--- a/app/client/src/workers/ast.test.ts
+++ b/app/client/src/workers/ast.test.ts
@@ -310,4 +310,40 @@ describe("parseJSObjectWithAST", () => {
     const resultParsedObject = parseJSObjectWithAST(body);
     expect(resultParsedObject).toStrictEqual(parsedObject);
   });
+  it("parse js object with variable declaration inside function", () => {
+    const body = `{
+      myFun1: () => {
+        const a = {
+          conditions: [],
+          requires: 1,
+          testFunc: () => {},
+          testFunc2: function(){}
+        };
+      },
+      myFun2: async () => {
+        //use async-await or promises
+      }
+    }`;
+    const parsedObject = [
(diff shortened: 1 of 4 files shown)


app/client/src/workers/ast.test.ts
```diff
@@ -309,4 +310,40 @@ describe("parseJSObjectWithAST", () => {
     const resultParsedObject = parseJSObjectWithAST(body);
     expect(resultParsedObject).toStrictEqual(parsedObject);
   });
+  it("parse js object with variable declaration inside function", () => {
```
Rishabh-Rathod, line 313: Are we handling below case? ( it doesn't seem like this is handled, please check )
1. `oneJSFunc: this.alreadyPresentJSFunc` 
this gets treated as variable instead of function and hence not sure 
   - running `oneJSFunc` will be handled and if `alreadyPresentJSFunc` is async when calling `oneJSFunc` will run async actions or not?

Rishabh-Rathod (commented):
Initial review

Rishabh-Rathod (commented):
When we refactor JSObject old variables still are present in JSObject dataTree. 

Please fix this.

ohansFavour (dismissed):

ramsaptami:
Tested for nested functions in an object and array and they do not appear on the settings tab or available to run via dropdown

Rishabh-Rathod:
We will be automating the test cases mentioned in test plan once we unblock the release by merging this PR.

PR #14873, opened 2022-06-28
fix: Saving values with special chars in env file

### LINKED ISSUE #13335: [Bug]: If there's a space in a value in `docker.env`, the server fails to start

### Description

If any of the values in the `docker.env` file contain a space, like this:

```sh
APPSMITH_ADMIN_EMAILS=one@example.com, two@example.com
```

Then the server will fail to start with an error like the following:

```
line 10: two@example.com: command not found
```

There's no way to recover without going and editing the `docker.env` file to fix this.

### Steps To Reproduce

Start a new self-hosted instance, go to admin settings, in admin emails, enter `one@ex.com, two@ex.com`, and click on save and restart. The server won't start.

### Public Sample App

_No response_

### Version

Self Hosted

sharat87:
Also included some refactoring to fix several compile time warnings and lint errors, including proper formatting.

Fixes #13335.


nayan-rafiq (approved):
LGTM

PR #14892, opened 2022-06-29
fix: JS toggle issue with Numeric Input & multi select dropdown controls

### LINKED ISSUE #14810: [Bug]: File Picker - JS Toggle on Allowed File Types - JS toggle un-clickable till field is cleared

### Description

With File Picker widget, if we do a JS toggle on Allowed File Types property, with any of the properties selected, the JS button remains non clickable until  the field is cleared. This behavior is seen even if the Allowed file types are of valid file types.

Expected Behavior - If the allowed file types are of valid file types, the JS button should not be disabled. 

### Steps To Reproduce

1. Drop a file picker widget onto the canvas.
2. From the Allowed File Types property, select Images. 
3. Now click on the JS toggle button
4. The button gets disabled immediately and unless we clear the field, it does not get enabled again
5. This behavior is seen even if the Allowed file types are of valid file types, which is incorrect.

If the allowed file types entered are of valid file types, the toggle should not get disabled. Only if we enter some unaccepted value, it should get disabled.
Screenshot 2022-06-24 at 10 14 46 PM

### Public Sample App

_No response_

### Version

Release

### LINKED ISSUE #14867: [Bug]: JS toggle for Label width property is not working properly

### Description

Right after clicking the JS button, it becomes disabled.
https://www.loom.com/share/05f6e6c4d8b145398dfe72af6f1c2fdf

### Steps To Reproduce

1. Add any widget with a label width property (example: Input widget)
2. click on JS toggle button for the label width

### Public Sample App

_No response_

### Version

cloud

aswathkk:
## Description

NumericInput control which is used for the Label width property was not having a JS toggle validation. Also, the validation logic written for Dropdown control was not handling multi-select. These two validations were added to fix the issues with these controls.

Fixes #14867 
Fixes #14810 

## Type of change

## How Has This Been Tested?

- Manual
- Jest tests are added

## Checklist:


app/client/src/components/propertyControls/DropDownControl.tsx
```diff
@@ -129,14 +129,28 @@ class DropDownControl extends BaseControl<DropDownControlProps> {
...
+  ): boolean {
+    const allowedValues = new Set(
+      config?.options?.map((x: { value: string | number }) =>
+        x.value.toString(),
+      ),
+    );
+    if (config.isMultiSelect) {
+      try {
+        const values = JSON.parse(value);
+        for (const x of values.map((x: string | number) => x.toString())) {
```
rahulramesha, line 144: Looks like you are just iterating over the list twice first with `.map` and then with `for` loop. Can you just use `x.toString()` below in the if loop?

Also don't feel like a good idea to use `.map` to get the array you are iterating over inside the `for` condition. You could just create the array before and then use that in the `for` condition.

rahulramesha (changes requested):

app/client/src/components/propertyControls/DropDownControl.tsx
aswathkk: My bad 🤦 

@rahulramesha 
Totally agree with your first point. I will make the change and commit.

But, I didn't get the second point. Did you mention that in terms of readability?

rahulramesha (approved):

app/client/src/components/propertyControls/DropDownControl.tsx
rahulramesha: @aswathkk yes, it is better to not have logic inside conditional statements

PR #14930, opened 2022-06-30
fix: removing throwing error because as of now it's only used while determing async function

### LINKED ISSUE #12639: [Bug]: Unexpected error in console due to current `isFunctionAsync` check

### Description

Screenshot 2022-04-06 at 11 14 55 PM

### Steps To Reproduce

1. Create an app with a Filepicker widget and an API name `Api1`
2. Create a JS Object with the code mentioned below
```
export default {
	myFun2:() => {
		console.log(FilePicker1.files);
		Api1.run().then(()=>{
      showAlert("hello");
    })
	}
}
```
3. Refresh the page and notice the dev tools console log

### Public Sample App

https://release.app.appsmith.com/console-errors-isasync/page1-624daa8fdbe66e13e8d84ad4

### Version

1.6.17

ApekshaBhosale:
## Description

Users keep seeing Async function called in a sync field in the console. There are produced as part of checking each js object function's behavior i.e sync or async. This part of the code is only being used by isFunctionAsync hence commenting it out

Fixes #12639 

## Type of change

## Checklist:


app/client/src/workers/PromisifyAction.ts
```diff
@@ -27,7 +27,7 @@ export const promisifyAction = (
      * @link isFunctionAsync
      * */
     self.IS_ASYNC = true;
-    throw new Error("Async function called in a sync field");
+    //throw new Error("Async function called in a sync field");
```
Rishabh-Rathod, line 30: We will start seeing jest test failures due to this as we have jest test for verifying throwing of error.

While i was trying to fix this, i wasn't sure able to understand 
- why do we throw error and where do we use it?
- why we are not able to catch this thrown error? 

If we are aware that removing this line is safe then we could also remove the jest test for it.

ApekshaBhosale: as it's not valid anymore. this gets used in promisfication. removed test cases

app/client/src/workers/PromisifyAction.test.ts
```diff
@@ -21,14 +21,14 @@ describe("promise execution", () => {
...
-    expect(postMessageMock).not.toHaveBeenCalled();
-  });
+  //we no longer add async functions to sync fields during evaluation hence this is not valid anymore
+  // it("throws when allow async is not enabled", () => {
+  //   self.ALLOW_ASYNC = false;
+  //   self.IS_ASYNC = false;
+  //   expect(dataTreeWithFunctions.showAlert).toThrowError();
+  //   expect(self.IS_ASYNC).toBe(true);
+  //   expect(postMessageMock).not.toHaveBeenCalled();
+  // });
```
ohansFavour, line 31: Why are we commenting instead of total removal or modification of test case?

Issue #14886, opened 2022-06-29
[Bug]: No error is thrown if the input widget contains characters greater than the max limit


shadabbuchh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

No error is thrown if one enters some text in the Input widget and then sets a `Max Chars` property such that it is less than the number of characters entered.
https://www.loom.com/share/1b94a690ce674008ac909f9dc6eb1ba1

Please note the following:

1. No error is triggered once you make a small change focus & blur the input.
2. On hitting `Backspace` the characters get erased. (No error is triggered)
3. A developer is not allowed to further enter any characters.
4. There is no error visible for the appsmith dev on the property pane

Please refer to the below comment to get clarity on the solution for this issue:
https://github.com/appsmithorg/appsmith/issues/14830#issuecomment-1168572017

### Steps To Reproduce

1. D&D a Input widget.
2. Enter 20 characters.
3. Not set a limit on the `Max Chars` property.

### Version

Cloud

somangshu:
@techbhavin build an understand of what can be achieved from the above. Get in touch with @dilippitchika if you need clarity on the Requirement

cc @sbalaji1192 please review any changes w.r.t the same

Issue #14830, opened 2022-06-27
[Bug]: Input widget issues


shadabbuchh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

1. Error message when `Default Text` has more characters than the `Max Chars` is incorrect. It should be: `Default Text length must be less than or equal to Max Chars allowed.`
<img width="531" alt="Screenshot 2022-06-27 at 3 03 02 PM" src="https://user-images.githubusercontent.com/39921438/175911116-c5ff9607-cebe-4b4a-9c3d-81837f818cb9.png">

2. ~If a user enters some text in the widget and then sets a `Max Chars` such that it is less than the initial text entered, no error is thrown. (This is reproducible in release. Works fine in prod)
https://www.loom.com/share/1b94a690ce674008ac909f9dc6eb1ba1~

### Steps To Reproduce

1. D&D an `Input` widget.

### Version

Cloud

somangshu:
@jnikhila can you provide copy for the given content
> Default Text length must be less than or equal to Max Chars allowed

@shadabbuchh for the 2nd point
- The error is not visible to the app viewers since they will not have access to the property pane
- The field is not dirty, Does it trigger the error once you make a small change focus & blur the input? How is the behaviour different in Prod and Release?


cc @souma-ghosh

jnikhila:
@somangshu - I think it will be helpful to users if we tell them what the current max chars limit is and how many characters they have exceeded. In which case, the users will be able to reword the data.

Below are a few messages that we can use:

`Message 1 - The entered text exceeds the MaxChars limit of <AllowedNumberOfChars> characters by <TheExceededChars> characters.`

For example, message 1: "The entered text exceeds the MaxChars limit of 150 characters by 20 characters."

`Message 2 - You have exceeded the <MaxChars> characters limit for the <Field_Name> field by <TheExceededChars> characters.`

For example, message 2: "You have exceeded the 150 characters limit for the input1 field by 20 characters."

`Message 3 - The <Field_Name> field has a max length of <MaxChars> characters, and you have exceeded it by <TheExceededChars> characters.`

For example, message 3: The input1 field has a max length of 150 characters, and you have exceeded it by 20 characters.

Note - I've assumed the data in the angular brackets can be populated at runtime. Please let me know if that's not the case.

shadabbuchh:
> * The field is not dirty, Does it trigger the error once you make a small change focus & blur the input? How is the behaviour different in Prod and Release?

[...]

PR #14934, opened 2022-06-30
fix: show the error message for text input widget

techbhavin:
## Description

> No error is thrown if one enters some text in the Input widget and then sets a Max Chars property such that it is less than the number of characters entered.

Fixes [#14886](https://github.com/appsmithorg/appsmith/issues/14886)

## Type of change

## How Has This Been Tested?

- D&D a Input widget.
- Enter 20 characters.
- Not set a limit on the Max Chars property.

## Checklist:


app/client/src/ce/constants/messages.ts
```diff
@@ -22,6 +22,8 @@ export const API_PATH_START_WITH_SLASH_ERROR = () => `Path cannot start with /`;
 export const FIELD_REQUIRED_ERROR = () => `This field is required`;
 export const INPUT_DEFAULT_TEXT_MAX_CHAR_ERROR = () =>
   `Default Text length must be less than Max Chars allowed`;
+export const INPUT_TEXT_MAX_CHAR_ERROR = () =>
+  `Input Text length must be less than Max Chars allowed`;
```
mohanarpit, line 26:
```suggestion
  `Input text length must be less than max chars allowed`;
```

Nit: Correcting the sentence casing here.

Also, is it possible to modify this error message to - Eg: "Input text length must be less than 140 characters"? In this example, 140 is the max chars property. This way, a user is able to accurately determine & fix the error in their input.

app/client/src/ce/constants/messages.ts
```diff
@@ -22,6 +22,8 @@ export const API_PATH_START_WITH_SLASH_ERROR = () => `Path cannot start with /`;
...
 export const INPUT_DEFAULT_TEXT_MAX_NUM_ERROR = () =>
   `Default Text value must be less than Max number allowed`;
```
mohanarpit, line 28: I know this is not part of the scope, but can we apply the same logic to this error message as well? Correct the sentence casing and also add the max number allowed in the input.

app/client/src/ce/constants/messages.ts
somangshu: Getting picked up in #14830

mohanarpit (approved):

Tooluloope:
@techbhavin Failed DP

Issue #14543, opened 2022-06-14
[Bug]: onPageLoad action failure due to incorrect layoutOnLoadActions in pageResponse


Rishabh-Rathod:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

| Expected | Observed |
|---|---|
| ![SCR-20220614-sp0](https://user-images.githubusercontent.com/23132741/173618887-ad12c5de-0b74-440a-bb72-3a764f6a6370.png) | ![SCR-20220614-src](https://user-images.githubusercontent.com/23132741/173618920-397067de-a46d-4f22-8801-595c21713284.png)

We noticed incorrect actionId for users application. 

### Steps To Reproduce

Not sure yet

This was a discord user query and we were able to reproduce issue in user's application.

## Workaround 

1. If there are uncommitted changes then commit & push then 
2. discard & pull. 

That should fix the issue as the application went into bad state while importing ( pull or git import or normal import) 

https://docs.appsmith.com/core-concepts/version-control-with-git/pull-and-sync#discard-and-pull-changes

### Version

1.7.2

abhvsn:
We are implementing the transactions to address this issue. The POC is will be completed by 28th June.

cc: @hiteshjoshi

Nikhil-Nandagopal:
@hiteshjoshi @AnaghHegde can someone fix this issue? It's been open for so long

AnaghHegde:
@Nikhil-Nandagopal PR is already raised for this. It's waiting for a review from @mohanarpit and @sharat87
Link to PR - https://github.com/appsmithorg/appsmith/pull/14939

PR #14939, opened 2022-07-01
feat: Support mongodb transactions for import application flow

abhvsn:
## Description

Import application flow is a multi-stage process where we touch all the mongo DB collections required to store the application in DB, which today includes pages, actions, JSObjects, etc. When the flow is complete then only we can say the application is successfully imported and DB won't have any stale objects stored. But in a negative scenario where the flow might break because of some unknown exceptions, objects stored earlier become stale and reside in DB forever. With this PR we are establishing the infra for transactions to avoid saving stale DB objects. This is achieved by a rollback mechanism in case an exception is thrown in the middle of execution.  

Note: Since transactions are built on concepts of logical sessions they require mecahnics (like oplog) which are only available in replica set environment.

You can always convert a standalone to a single noded replica set and transactions will work with this one node.

[https://docs.mongodb.com/manual/tutorial/convert-standalone-to-replica-set/](https://docs.mongodb.com/manual/tutorial/convert-standalone-to-replica-set/?_ga=2.53712296.1453502529.1669195955-2065030687.1664009965)

```
For local dev earlier we used to run the single node without replica set. Henceforth if we want to take the advantage of transactions please follow the steps to run mongod single node in local with replica set:

- Upgrade the MongoDB version to 4.4 or higher if it’s 4.2 or lower (https://www.mongodb.com/docs/manual/release-notes/4.4/#std-label-4.4-upgrade)
- Close the mongoDB instance running in your local
- Start the mongoDB in replica set mode and initiate the replica set
    - mongod --port 27017 --dbpath <path/to/db> --replSet <replica-set-name> && mongo --eval “rs.initiate()”
- One can use following commands to check replica set status: 
    - mongo appsmith
    - rs.status()
- By this time you should have the mongo running with replica set
```
Screenshot 2022-07-01 at 10 31 27 PM

This also means mongodb connection string used in env file will now include the replica-set name if one wants to leverage transactions:
`mongodb://localhost:27017/appsmith?replicaSet={replica-set-name}`

Fixes https://github.com/appsmithorg/appsmith/issues/14543

## Type of change

## How Has This Been Tested?

> JUnit 
> Manual test

## Checklist:


app/server/appsmith-server/src/test/java/com/appsmith/server/configurations/EmbeddedMongoConfig.java
```diff
@@ -1,21 +1,93 @@
...
+     */
     @Bean
-    public ImmutableMongodConfig prepareMongodConfig() throws IOException {
-        ImmutableMongodConfig mongoConfigConfig = MongodConfig.builder()
-                .version(Version.Main.V4_2)
-                .build();
-        return mongoConfigConfig;
+    @ConditionalOnMissingBean
+    public MongodConfig embeddedMongoConfiguration(EmbeddedMongoProperties embeddedProperties) throws IOException {
+        ImmutableMongodConfig.Builder builder = MongodConfig.builder().version(Version.Main.V4_4);
```
abhvsn, line 63: We need to upgrade to v4.4 on the prod instance as well to support the transaction.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -1148,7 +1150,8 @@ private Mono<Application> importApplicationInWorkspace(String workspaceId,
                                 analyticsService.sendEvent(AnalyticsEvents.UNIT_EXECUTION_TIME.getEventName(), tuple.getT2().getUsername(), data);
                                 return application;
                             });
-                });
+                })
+                .as(transactionalOperator::transactional);
```
abhvsn, line 1154: This line will enable the transactions in the import application flow.

sharat87:
@abhvsn, prod DB is at v4.4 now. :)

AnaghHegde:
@sharat87 Did we not upgrade it to 4.6x in the scheduled downtime of v1.8.0?

AnaghHegde:
@sharat87 When are we planning to upgrade the mongo version to 5x?

AnaghHegde:
@sharat87 Since Abhijeet is off, I will be picking this up. Can you please review this PR when you have time?

sidhantgoel (commented):
LGTM, jut 1 thing, should we a positive test case also, or it will be taken care in other testcases?

AnaghHegde:
It will be handled by existing scenarios.

sidhantgoel (approved):

PR #14942, opened 2022-07-01
feat: Add API to get sample apps related to a template

### LINKED ISSUE #14786: [Task] GET Apps connected to a template. (BE)
The Get API for Backend to get connected apps of a template.

nayan-rafiq:
## Description
This PR adds query params to the get similar templates API.

Fixes #14786

## Type of change

## How Has This Been Tested?
- Manually

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
```diff
@@ -55,10 +55,16 @@ public ApplicationTemplateServiceCEImpl(CloudServicesConfig cloudServicesConfig,
...
     @Override
-    public Flux<ApplicationTemplate> getSimilarTemplates(String templateId) {
-        final String apiUrl = String.format("%s/api/v1/app-templates/%s/similar?version=%s",
-                cloudServicesConfig.getBaseUrl(), templateId, releaseNotesService.getReleasedVersion()
-        );
+    public Flux<ApplicationTemplate> getSimilarTemplates(String templateId, MultiValueMap<String, String> params) {
+        UriComponents uriComponents = UriComponentsBuilder
+                .fromUriString(cloudServicesConfig.getBaseUrl())
+                .pathSegment("api/v1/app-templates", templateId, "similar")
+                .queryParams(params)
```
sharat87, line 62: Looks like the params is passed on as received without looking inside? Can we validate this map? Or, even better, just take the params we want/expect from this map, and pass _only_ those that we know.

sharat87 (changes requested):
Just one change I'd like to see, otherwise good to go.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
nayan-rafiq: @sharat87 This API is actually a proxy to the underlying cloud services API, so I didn't validate the map attributes here as it's validated in the cloud services API. If in future we need an additional parameter, we only need to change in the cloud services. Do you think validating the parameters here as well is necessary?

sharat87: Yeah, figured, but it feels a little implicit. It's probably just me being paranoid again, 🤷

sharat87 (approved):

PR #14957, opened 2022-07-01
fix: Add migration to fix deleted themes when branch was deleted

### LINKED ISSUE #15247: [Bug]-[4932]:Cannot read properties of undefined (reading 'properties')

### Description

After the last update of appsmith, every time there's an issue happening using "GIT", I'm not actually sure how is happening, because after a day or 2 starts happening

Screen Shot 2022-07-17 at 11 22 50 AM

It works again, after re-importing the project from git

### Steps To Reproduce

1. Import project from git
2. Change any page 
3. Commit changes
4 In this step is when we wait after a day or 2, start happening the issue

### Public Sample App

_No response_

### Version

v1.7.8

nayan-rafiq:
## Description

Due to a bug, customized themes set to the parent branch where deleted when user deletes git branch. The bug has been fixed already. This PR adds a DB migration that fixes the data if there is any.

Fixes #15247 

## Type of change

## How Has This Been Tested?

- Manually

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
```diff
@@ -1038,4 +1040,49 @@ public void migrateOrganizationIdToWorkspaceIdInNewActionDatasource(MongockTempl
...
+        Set<String> set = new HashSet<>();
+        set.addAll(customizedEditModeThemeIds);
+        set.addAll(customizedPublishedModeThemeIds);
+
+        Update update = new Update().set(fieldName(QTheme.theme.deleted), false)
+                .unset(fieldName(QTheme.theme.deletedAt));
+        Criteria deletedCustomThemes = Criteria.where(fieldName(QTheme.theme.id)).in(set)
+                .and(fieldName(QTheme.theme.deleted)).is(true);
+
+        mongockTemplate.updateMulti(new Query(deletedCustomThemes), update, Theme.class);
```
abhvsn, line 1086: @nayan-rafiq isue we are seeing is happening when users have custom theme with multiple branches and if any of the branch is deleted all the branches will end up in a broken state. This is happening because of the same `themeIds` are referenced in all the branches, where we expect separate objects in `Theme collection` with separate `themeIds` for different branches. With the current migration I'm not getting where are we inserting new documents in the `Theme collection` and updating the corresponding ids in the branched application object in DB.

nayan-rafiq: When user has customized the theme and created a new branch, the newly created branch have the same theme ids as parent branch (it should've been copied). Now, if user deletes the new branch, it deletes the themes as well. So, if we set `deleted=false` to those themes, would it not be enough? It'll be a problem if we've option to restore a deleted branch. Is that happening today?

abhvsn: @nayan-rafiq consider user has already created 3 branches before we fixed the issue for the custom theme app. As per our expectation we should have 3 seperate theme ids attached to these apps which will not be the case. Now consider after we run the migration there won't be any change considering all the branches are still active.
In this scenario (after we run the migration) what will happen if user tries to delete the earlier branch (out of 3 branches created earlier)? As we are not checking if the theme id is being used in some other branch, it will get deleted and we will end up in same situation again.

mohanarpit (approved):

PR #14962, opened 2022-07-02
fix: JSObject AST parsing logic

### LINKED ISSUE #15009: [Task]: Add check to match `jsObject` node for AST parsing logic

### SubTasks

As we set the `jsObjectVariableName` in the jsCode which we parse for AST for JSObject.

we could only focus on particular variable name instead of all variableDeclarator.
Hence adding `node.id.name === jsObjectVariableName`.

It worked without it as jsObject was the last node to be called in variableDeclarator when parsing AST.

Rishabh-Rathod:
## Description

As we set the jsObjectVariableName in the jsCode which we parse for AST, 
- we could only focus on particular variable name instead of all variableDeclarator.
 
Hence adding `node.id.name === jsObjectVariableName`. 

It worked without it as jsObject was the last node to be called in variableDeclarator when parsing AST. 

Fixes #15009

## Type of change

## Checklist:


ohansFavour (changes requested):
- Attach issue to this PR
- Add jest test prevent this from happening in the future

ohansFavour (approved):

Rishabh-Rathod:
Testing for this PR is skipped as this changes doesn’t change any functionality and just add a logical check in code.

PR #14992, opened 2022-07-04
fix: Update maincanvas height on widget resize, drag or delete operation

### LINKED ISSUE #14991: [Bug]: Canvas empty Space does not go away even after modifying the bottom most row

### Description

In few applications, The canvas has extra empty space at the bottom of the canvas, even though there are now widgets present in that area.
Resizing, Dragging or even deleting a widget should recalculate the canvas length and adjust it accordingly. But even these operations does not change the bottom most row of the canvas, resulting in extra space at the bottom of canvas

### Steps To Reproduce

1. This is only reproducible in few applications, on only few pages
2. Export appsmith stand-up app
3. Import the app
4. Navigate to Standup Form, and Scroll down
5. Observe that the empty space at the bottom of the canvas
6. try a widget operation like resizing a widget, dragging a widget or deleting a widget.
7. Observe that these operations does not trigger any change in Main canvas height and the empty canvas space still exists
8. Expected behaviour, When a widget is resized, dragged or deleted on the Main canvas, The main canvas should recalculate its height and get rid of empty space.

### Public Sample App

https://app.appsmith.com/app/appsmith-standup/standup-form-5fb67116322ddc249377f5eb?branch=master

### Version

Cloud

rahulramesha:
## Description

Changes include using updated mainCanvas minHeights during widget operations, instead of the one stored in DSL.

minHeight was calculated by using the minHeight from the DSL, If the last widget's bottom row was smaller than the screen height it would set minHeight as screenHeight. But this did not cater for canvases whose last widget's bottom row was longer than screenheight, then the minHeight would be set from DSL, even if that value was too high. Now the minHeight in edit mode will always be equal to the screenHeight of the editor.

That minHeight for mainCanvas is used in calculation and updating mainCanvas BottomRow whenever a widget is dragged or resized or deleted.

Fixes #14991 

## Type of change

## How Has This Been Tested?

Manual, Cypress

## Checklist:


app/client/src/sagas/CanvasSagas/DraggingCanvasSagas.ts
```diff
@@ -44,11 +51,21 @@ export function* getCanvasSizeAfterWidgetMove(
...
+    );
+    mainCanvasMinHeight = mainCanvasProps?.height;
+  }
+
   if (canvasWidget) {
     const occupiedSpacesByChildren: OccupiedSpace[] | undefined = yield select(
       getOccupiedSpacesSelectorForContainer(canvasWidgetId),
     );
-    const canvasMinHeight = canvasWidget.minHeight || 0;
+    const canvasMinHeight = mainCanvasMinHeight || canvasWidget.minHeight || 0;
```
riodeuno, line 68: `canvasWidget.minHeight` is too large a value today. We'll get a scroll in most cases. Is it possible to have a constant of say 400px which we can use as a fallback? Also, is it possible that we'll always have `mainCanvasMinHeight` defined? If it is not, in what scenarios will it be `undefined` or 0?

app/client/src/sagas/WidgetDeletionSagas.ts
```diff
@@ -358,7 +383,7 @@ function resizeCanvasToLowestWidget(
   }
 
   let lowestBottomRow = Math.ceil(
-    (finalWidgets[parentId].minHeight || 0) /
+    (mainCanvasMinHeight || finalWidgets[parentId].minHeight || 0) /
```
riodeuno, line 386: Same comment as above for this.

app/client/src/widgets/SelectWidget/widget/index.tsx
```diff
@@ -65,7 +65,8 @@ export function defaultOptionValueValidation(
   } else {
     isValid = false;
     parsed = {};
-    message = `value does not evaluate to type: string | number | { "label": "label1", "value": "value1" }`;
+    message =
+      'value does not evaluate to type: string | number | { "label": "label1", "value": "value1" }';
```
riodeuno, line 69: 😓 Was this the issue you were facing with the dev mode validations? I am wondering how to avoid this in the future. Do you think it makes sense to automatically convert ``` to `'` when parsing the `fnString`?

app/client/src/sagas/WidgetDeletionSagas.ts
```diff
@@ -274,7 +288,17 @@ function* deleteAllSelectedWidgetsSaga(
...
+    //Main canvas's minheight keeps varying, hence retrieving updated value
+    let mainCanvasMinHeight;
+    if (parentId === MAIN_CONTAINER_WIDGET_ID) {
+      const mainCanvasProps: MainCanvasReduxState = yield select(
+        getMainCanvasProps,
+      );
+      mainCanvasMinHeight = mainCanvasProps?.height;
+    }
+
+    resizeCanvasToLowestWidget(finalWidgets, parentId, mainCanvasMinHeight);
```
riodeuno, line 301: Has the `resizeCanvasToLowestWidget` function now changed because there is a third argument?

riodeuno:
@rahulramesha Could you elaborate on the logic and the details of the what was the case before and how it is different now, in the description of this PR? It will help me get some context.

app/client/src/widgets/SelectWidget/widget/index.tsx
rahulramesha: @riodeuno yes, this was the one that caused me a lot of headache in finding it. Today we just convert function to string by `toString()`. Automatically converting it might cause a lot of issues with Template literals. I was thinking if we could you a library to translate it like how Compression Plugin does, but using such things during runtime will increase load times. Do you have any recommendations to avoid this in the future?

app/client/src/sagas/WidgetDeletionSagas.ts
rahulramesha: yes

app/client/src/widgets/SelectWidget/widget/index.tsx
riodeuno: @rahulramesha Ideally these conversions from function to string, should happen once in the lifecycle of a session. This means, we shouldn't see a performance overhead when doing this at runtime. Would you mind creating an issue where you could document the issue you were facing, so that we can tackle it in a separate PR?

riodeuno (approved):

Sripriya93:
Have verified the fix with five different applications, so after dragging, deleting or resizing, we are able to see the extra space is eliminated while we deploy the application.

PR #15000, opened 2022-07-05
fix: initial selected value not getting displayed

### LINKED ISSUE #14999: [Bug]: Default selected value of border radius or box shadow is not visible in Table column property

### Description

The default selected value of border-radius or box-shadow for Table widget's column property is not visible.
It should be shown which item is selected out of the available options.

Screenshot 2022-07-05 at 10 40 42 AM
 

### Steps To Reproduce

1. Add Table widget
2. Goto action column's property and take a look at border-radius & box-shadow

### Public Sample App

_No response_

### Version

1.7.6

aswathkk:
## Description

Since we are using COMPUTE_VALUE customJS control for Table columns, the evaluated value sent to the property controls becomes an array and the controls fail to render the initial selected item.
The fix is to send only one value out of the array. This logic might fail if users are using some bindings that can generate different values inside an array. But, we are anyway disabling the JS toggle for those scenarios.

Fixes #14999

## Type of change

## How Has This Been Tested?

- Manual

## Checklist:


app/client/src/utils/PropertyControlFactory.tsx
```diff
@@ -37,19 +38,24 @@ class PropertyControlFactory {
...
     let controlBuilder;
+    let evaluatedValue = controlData.evaluatedValue;
 
     if (preferEditor) {
       controlBuilder = customEditor
         ? this.controlMap.get(customEditor)
         : this.controlMap.get("CODE_EDITOR");
     } else {
+      if (customEditor === "COMPUTE_VALUE" && isArray(evaluatedValue)) {
+        evaluatedValue = evaluatedValue[0];
```
riodeuno, line 49: Would it make more sense to use `if(uniq(evaluatedValue).length === 1)` condition before assigning?

riodeuno: I am assuming the case in which all values are the same, and in which all values are not the same. If I understand the solution correctly, we allow the JS toggle if all the values in the array are the same. Is this correct?

aswathkk: We disable the JS toggle if there are any bindings except the binding of the default theme. So, what I was saying is that on line number 49, the value of `evaluatedValue` will always be an array where all the values are the same.

riodeuno (approved):

Sripriya93:
Have verified Table widget "Action" column's Button property and style, moving to done

PR #15005, opened 2022-07-05
feat: add hostname to log file

### LINKED ISSUE #15004: [Bug]: log file overwritten in shared deployments

### Description

We deploy appsmith on kubernetes and share the same efs across multiple pods.

Problem:
When the logs are written to` /appsmith-stacks` and we have multiple pods writing to the same directory.
logs of appsmith process are getting over-written. In a nut-shell it is more like a race-condition where the last process to generate logs get over-written on the efs drive. 

This causes in-consistency when we debug issues in any setup.

### Steps To Reproduce

go to release cluster verify any service logs, They only contain partial confirmation
eg: 
logs in log-dna are in-consistent with logs in efs volume.

### Public Sample App

release.app.appsmith.com

### Version

eks-instance/cloud

pratapaprasanna:
Fixes #15004

## Type of change

## How Has This Been Tested?

Added host-name there by enabling/making sure the log files have different name and there-by, keeping the data consistent.

Screenshot 2022-07-05 at 12 31 28 PM


deploy/docker/templates/supervisord/application_process/editor.conf
```diff
@@ -5,8 +5,8 @@ autostart=true
 autorestart=true
 startsecs=0
 startretries=3
-stdout_logfile=/appsmith-stacks/logs/%(program_name)s/access.log
-stderr_logfile=/appsmith-stacks/logs/%(program_name)s/error.log
+stdout_logfile=/appsmith-stacks/logs/%(program_name)s/access%(ENV_HOSTNAME)s.log
```
mohanarpit, line 8:
```suggestion
stdout_logfile=/appsmith-stacks/logs/%(program_name)s/access-%(ENV_HOSTNAME)s.log
```

Minor: Adding a hyphen to differentiate between the service & hostname in the file. Please make this change in all the log configurations.

Also, I'm curious. What does `s` at the end of the string specify? Is it to convert the hostname to a string?

pratapaprasanna: The problem with adding `-` is that when we try deploying appsmith on instances where env variable HOSTNAME is not present the file_name would be 
`backend-.log` which doesnt look good i feel. So, removed the -. if you suggest to keep - we shall 

`s` i am guessing its like format-specifiers only

sharat87: It's Python's standard extension to `printf`-style formatting. Like `"hello %(name)s" % {"name": "Shri"}` is the same as `"hello %s" % "Shri"`.

sharat87: Is it possible add a `-` to the `HOSTNAME` env variable itself, if it isn't empty? Sorry I thought that was how this was implemented. The log file names look quite amateurish without a dash. :|

sharat87: Just realized, I _think_ `HOSTNAME` is a standard env variable. Let's not change it then.

Perhaps, something like:

```sh
APPSMITH_LOG_SUFFIX=""
if [[ -n "${HOSTNAME-}" ]]; then
  APPSMITH_LOG_SUFFIX="-$HOSTNAME"
fi
```

And then use `APPSMITH_LOG_SUFFIX` instead?

Also, if `HOSTNAME` is a standard env variable, it will _always_ be present, right? Like, the Docker container is always going to be an Ubuntu system, and whether it runs on Kubernetes or not, if Ubuntu sets `HOSTNAME`, it'll always be available no? In that case, should we just go with Arpit's suggested change in the first comment in this thread?

sharat87 (approved):

PR #15023, opened 2022-07-05
fix: reset selectedFiles meta prop only on specific reason event

### LINKED ISSUE #15002: [Bug]: Filepicker not retaining the data when i move to a query page

### Description

When i upload a file to filepicker and bind a data to query, i am unable to see the uploaded file values

### Steps To Reproduce

1. Add a new filepicker widget
2. Upload a file
3. Change page to a query page
4. Reference filepicker.files[0]
5. Observe that the reference shows undefined

### Public Sample App

_No response_

### Version

Production

keyurparalkar:
## Description

It was observed that `selectedFiles` meta property was getting set to blank when component was unmounted.

This happens so because `uppy.close` function emits `cancel-all`, and `file-removed` events. Due to this, handler attached on `file-removed` event also gets executed that in turns updates the `selectedFiles` meta property. 

To avoid this scenario, uppy provides us another argument called `reason` that helps us to identify the caller event.

Refer to this Uppy's PR - `https://github.com/transloadit/uppy/pull/2323`

Fixes #15002 

## Type of change

## How Has This Been Tested?
Updated the `FilePickerV2_spec.js` file. This files tests the following scenario - 

1. Create an API with the following binding `{{FilePicker1.files[0]}}` as dummy.
2. Then go to the widgets page to see if the uploaded file still exists.

## Checklist:


techbhavin (approved):

SatishGandham (commented):
@keyurparalkar , while this fixes the problem, please find the root cause so that we don't make the same mistake again and to ensure there is no regression some where else.

keyurparalkar:
> @keyurparalkar , while this fixes the problem, please find the root cause so that we don't make the same mistake again and to ensure there is no regression some where else.

@SatishGandham here is the [RCA](https://theappsmith.slack.com/archives/C02K0SZQ7V3/p1657117091537179?thread_ts=1657105300.947129&cid=C02K0SZQ7V3) for this issue.

PR #15027, opened 2022-07-05
fix: added handling for array of objects containing blob urls for multiple file uploads

### LINKED ISSUE #15025: [Bug]: Array of files not working with multipart/form-data in execute action

### Description

When you try to upload `multiple` files in a single key using `multipart/form-data`, then the upload doesn't work.
Here are the screenshots that represent the application
Screenshot 2022-07-05 at 5 24 42 PM

Screenshot 2022-07-05 at 5 21 19 PM
.

### Steps To Reproduce

1. Create an application that has file picker with support of multiple files and increase the size limit
2. Add an API that sends the files `{{Filepicker1.files}}` as array.
3. Observe that the files uploaded result in 53 bytes size only

### Public Sample App

_No response_

### Version

Both

AmanAgarwal041:
## Description

Created a function `resolvingBlobUrls` that handles : 
- objects having blob urls.
- blob urls directly
- else return the value unchanged

Added a handling for array of objects that might have blob urls for multiple file uploads.

Fixes #15025

## Type of change

## How Has This Been Tested?

Manually by uploading multiple files each > 8 mb.

## Checklist:

### DIFF
--- a/app/client/src/sagas/ActionExecution/PluginActionSaga.ts
+++ b/app/client/src/sagas/ActionExecution/PluginActionSaga.ts
@@ -30,7 +30,7 @@ import {
   getAppMode,
   getCurrentApplication,
 } from "selectors/applicationSelectors";
-import _, { get, isString, set } from "lodash";
+import _, { get, isArray, isString, set } from "lodash";
 import AppsmithConsole from "utils/AppsmithConsole";
 import { ENTITY_TYPE, PLATFORM_ERROR } from "entities/AppsmithConsole";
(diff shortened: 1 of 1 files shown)


Irongade (approved):

Parthvi12:
@prapullac Cypress tests for this flow is not possible today afaik, because we are switching domains and with current cypress version its not feasible.

prapullac:
As per the above comment by Parthvi Since this is an urgent fix we can progress this.

PR #15177, opened 2022-07-13
fix: Checkbox reskinning

### LINKED ISSUE #10975: [Enhancement] : Checkbox widget reskinning
https://www.notion.so/appsmith/Checkbox-b8a235faf3b74fa2aff59f68cafc0c05

jsartisan:
Checkbox reskinning

Fixes #10975

## Type of change
- Style updates

## How Has This Been Tested?
- Manually tested

## Checklist:

### DIFF
--- a/app/client/src/components/wds/Checkbox/index.tsx
+++ b/app/client/src/components/wds/Checkbox/index.tsx
@@ -1,114 +1,116 @@
 import styled from "styled-components";
-import { Checkbox as BlueprintCheckbox } from "@blueprintjs/core";
+import { Alignment, Checkbox as BlueprintCheckbox } from "@blueprintjs/core";
 
 import { Colors } from "constants/Colors";
-import { lightenColor, darkenColor } from "widgets/WidgetUtils";
+import { darkenColor } from "widgets/WidgetUtils";
 
 type StyledCheckboxProps = {
   checked?: boolean;
   disabled?: boolean;
-  backgroundColor?: string;
   borderRadius?: string;
   indeterminate?: boolean;
   hasError?: boolean;
+  accentColor?: string;
   inputRef?: (el: HTMLInputElement | null) => any;
 };
 
-const DISABLED_ICON_SVG =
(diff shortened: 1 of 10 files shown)


app/client/src/components/wds/Checkbox/index.tsx
```diff
@@ -1,114 +1,115 @@
...
+    // checked icon
+    &.bp3-control.bp3-checkbox input:checked ~ .bp3-control-indicator::before {
+      background-image: url(/widgets/check-icon.svg);
+    }
   `}
 `;
 
 Checkbox.defaultProps = {
-  backgroundColor: "#553DE9",
+  accentColor: "#553DE9",
```
keyurparalkar, line 112: @jsartisan is this value needs to hard coded or does it rely on the global theme's primary color? 
If it relies on the global theme value then can we use css variables in this case?

keyurparalkar (changes requested):

app/client/src/widgets/CheckboxGroupWidget/component/index.tsx
```diff
@@ -227,12 +231,20 @@ function CheckboxGroupComponent(props: CheckboxGroupComponentProps) {
...
+                <CheckboxLabel
+                  className="t--checkbox-widget-label"
+                  disabled={isDisabled}
+                  labelPosition={LabelPosition.Left}
+                >
+                  {option.label}
+                </CheckboxLabel>
+              }
               onChange={onChange(option.value)}
-              rowSpace={rowSpace}
```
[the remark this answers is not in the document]
jsartisan: The reason is we don't need this property. Previously It was adding height equal to the space for the chbkox. But just setting height equal to the height available is the way to go here.

app/client/src/components/wds/Checkbox/index.tsx
jsartisan: Removed. I actually added these just for testing.

keyurparalkar: @jsartisan it seems this is not removed yet. Can you remove it?

keyurparalkar (changes requested):

keyurparalkar (approved):
LGTM!

Issue #14694, opened 2022-06-20
[Bug]: Tab widget always defaults to the first tab when reset on both internal property and UI


ramsaptami:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Resetting tab widget 

### Steps To Reproduce

1. Drop a tabs widget on canvas and add a few extra tabs on the widget
2. Set default tab to `Tab 2`
3. Drop a button and a text widget on the canvas and bind tab widget's selectedTab property on to the text widget and set a reset trigger for the tab widget on the button
4. Change selected tab to **Tab 1** and observe the change on the text widget to **Tab 1**
5. Click on the button widget to trigger reset and observe the text widget change from **Tab 1** to **Tab 2** and back to **Tab 1**
6. Change the selected tab to **Tab 3** and observe the change on the text widget to **Tab 3**
7. Click on the button widget to trigger reset and observe the text widget change from **Tab 3** to **Tab 2** and then to **Tab 1**
8. UI in all cases of widget reset always goes back to Tab 1

https://www.loom.com/share/770c8a43686a4b96b814d405a0ac41a0

### Version

Cloud

dilippitchika:
Reset widget fuction in tab widget doesn't work as expected. If the default tab is tab2 resetting it should go to tab 2 and not tab 1. This is broken.

dilippitchika:
@techbhavin please triage this issue, no need to fix right now. Just need to know why this is happening.

techbhavin:
After resetting the Tab widget, widget getting `selectedTabWidgetId`(meta property) is undefined. so It takes the first tab as the default tab in the Tabs widget.

![image](https://user-images.githubusercontent.com/58818598/178743460-22db4159-13fa-4cb2-8cbe-008fca3fa783.png)

@dilippitchika

techbhavin:
Hello @dilippitchika, Yes we can fix this issue from our side.

PR #15202, opened 2022-07-14
fix: set default on reset of The Tab widget

techbhavin:
## Description

> Fixed to set a default value on reset action for the Tabs widget.

Fixes [#14694](https://github.com/appsmithorg/appsmith/issues/14694)

## Type of change

## How Has This Been Tested?

1. Reset the Tabs widget With the default value.
2. Reset the Tabs widget Without the default value.

## Checklist:

- [x] I have performed a self-review of my code


Rishabh-Rathod (commented):
Could we please fix failing cypress test before the code review and also add cypress test case for verification of the fix.

app/client/src/widgets/TabsWidget/widget/index.tsx
```diff
@@ -350,31 +357,36 @@ class TabsWidget extends BaseWidget<
     return [];
   };
 
+  setDefaultSelectedTabWidgetId = () => {
+    const visibleTabs = this.getVisibleTabs();
+    // Find the default Tab object
+    const selectedTab = find(visibleTabs, {
```
ashit-rath, line 363: Let's rename `selectedTab` to `defaultTab`?

app/client/src/widgets/TabsWidget/widget/index.tsx
```diff
@@ -328,11 +328,18 @@ class TabsWidget extends BaseWidget<
...
+        if (Object.keys(this.props.tabsObj || {}).length) {
+          // If we have a defaultTab
+          if (this.props.defaultTab) {
+            this.setDefaultSelectedTabWidgetId();
+          } else {
+            // If no tab is selected
+            // Select the first tab in the tabs list.
+            this.props.updateWidgetMetaProperty(
+              "selectedTabWidgetId",
+              visibleTabs.length ? visibleTabs[0].widgetId : undefined,
```
ashit-rath, line 340: We could replace this with
```suggestion
              visibleTabs?.[0]?.widgetId,
```

app/client/src/widgets/TabsWidget/widget/index.tsx
```diff
@@ -350,31 +357,36 @@ class TabsWidget extends BaseWidget<
...
+    // Find the default Tab object
+    const selectedTab = find(visibleTabs, {
+      label: this.props.defaultTab,
+    });
+    // Find the default Tab id
+    const selectedTabWidgetId = selectedTab
+      ? selectedTab.widgetId
+      : visibleTabs.length
+      ? visibleTabs[0].widgetId
+      : undefined; // in case the default tab is deleted
```
ashit-rath, line 371: We could also replace this with
```suggestion
    const selectedTabWidgetId = selectedTab?.widgetId ??  visibleTabs?.[0]?.widgetId;
```

Also please do modify `selectedTabWidgetId` to `defaultTabWidgetId`

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/Widgets/Tab_reset_spec.js
```diff
@@ -0,0 +1,40 @@
+const LayoutPage = require("../../../../locators/Layout.json");
```
ashit-rath, line 1: Observation:
1. The `it` block always represents one test case for a test spec. Right now it seems like each `it ` block here is a step leading to the last `it` block which is the actual test case to validate the reset functionality. So please add the whole test case in one it statement.
2. I think we should add another test case to check if the default tab is deleted and then the tab is reset, it should default to the first tab.

ashit-rath (commented):
@techbhavin The `How Has This Been Tested?` section is for the devs to write down a summary of the tests that they have written; not "how to reproduce" the issue. It gives the QA/Reviewers/PMs/EMs a quick way to verify if all the cases are covered. 
Please do update this section.

app/client/src/widgets/TabsWidget/widget/index.tsx
```diff
@@ -328,11 +328,18 @@ class TabsWidget extends BaseWidget<
...
-            visibleTabs[0].widgetId,
-          );
+        if (Object.keys(this.props.tabsObj || {}).length) {
+          // If we have a defaultTab
+          if (this.props.defaultTab) {
+            this.setDefaultSelectedTabWidgetId();
+          } else {
+            // If no tab is selected
+            // Select the first tab in the tabs list.
+            this.props.updateWidgetMetaProperty(
```
ashit-rath, line 338: Do we need the `else` case? Wouldn't the `setDefaultSelectedTabWidgetId` method deal with this case as well where the defaultTab is not present and assigns the `visibleTabs?.[0]?.widgetId`?

If yes, then we could remove the whole if-else and directly call the `setDefaultSelectedTabWidgetId`

app/client/src/widgets/TabsWidget/widget/index.tsx
```diff
@@ -350,31 +357,34 @@ class TabsWidget extends BaseWidget<
...
+        "selectedTabWidgetId",
+        defaultTabWidgetId,
+      );
+    }
+  };
+
   componentDidMount() {
     const visibleTabs = this.getVisibleTabs();
     // If we have a defaultTab
     if (this.props.defaultTab && Object.keys(this.props.tabsObj || {}).length) {
```
ashit-rath, line 386: If we modify the whole `if-else` condition here to something like 
```js
if (Object.keys(this.props.tabsObj || {}).length) {
  this.setDefaultSelectedTabWidgetId();
}
```

Would it do the same that that we are doing here?

Rishabh-Rathod (commented):
Please also remove deep equality check in `TabsWidget/widget/index.tsx`

`if (!isEqual(prevProps, this.props)) {`

As mentioned here - https://github.com/appsmithorg/appsmith/pull/13084#discussion_r926271194

cc: @ashit-rath

ashit-rath:
@techbhavin I think we should relook at what we are doing in the `componentDidUpdate`. It seems like the logic is very similar to what we are doing in `setDefaultSelectedTabWidgetId`. 
Can you check and see if we can refactor and remove redundant bits?

app/client/src/widgets/TabsWidget/widget/index.tsx
```diff
@@ -291,50 +291,33 @@ class TabsWidget extends BaseWidget<
...
-        const selectedTabWidgetId = selectedTab
-          ? selectedTab.widgetId
-          : undefined;
-        this.props.updateWidgetMetaProperty(
-          "selectedTabWidgetId",
-          selectedTabWidgetId,
-        );
-      }
-      // if selected tab is deleted
+    const visibleTabs = this.getVisibleTabs();
```
ashit-rath, line 294: The `componentDidUpdate` seems to check majorly 3 things
1. default tab change
2. a. selected tab changed and selectedId is not present in tabs
    b. selected tab changed and tabs not present
    c. selected tab changed and selectedTabId is null/undefined
3. selected tab but not present in visibleTabs.

They seems to be related to tab not being present in the tabs object expect point 1.

I feel like we can condense the whole componentDidUpdate with the following

```
const selectedTab = find(visibleTabs, {
  widgetId: this.props.selectedTabWidgetId,
});

if (this.props.defaultTab !== prevProps.defaultTab || !selectedTab) {
  this.setDefaultSelectedTabWidgetId();
}
```
Let me know if I'm missing something here or this doesn't cover the cases.

If this works fine then please do check if we have appropriate cypress test for all these scenarios.

ashit-rath (approved):
Great job @techbhavin

chandannkumar:
Tested this PR and working as expected. Tested for
- Reset the Tabs widget With the default value.
- Reset the Tabs widget Without the default value.

ashit-rath (approved):

PR #15203, opened 2022-07-14
feat: appsmithctl auto-backup and cleanup

sum35h:
## Description
- Auto-cleanup of backup files using env variable `APPSMITH_BACKUP_ARCHIVE_LIMIT` (default value is 4) after every backup.
- Updated docker file to include watchtower hook scripts.
- Error mail interval to 6hrs.

## Checklist:


deploy/docker/utils/bin/backup.js
```diff
@@ -51,14 +51,19 @@ async function run() {
     await fsPromises.rm(backupRootPath, { recursive: true, force: true });
 
     console.log('Finished taking a baceup at', archivePath);
```
mohanarpit, line 53: Nit: Typo fix.
```suggestion
    console.log('Finished taking a backup at', archivePath);
```

deploy/docker/utils/bin/backup.js
```diff
@@ -121,6 +126,20 @@ async function createFinalArchive(destFolder, timestamp) {
   return archive;
 }
 
+async function postBackupCleanup(){
+  console.log('Post-backup cleanup')
```
mohanarpit, line 130: Suggestion: Please improve the log messages to be user friendly or remove them. 

Eg: "Starting the cleanup task after taking a backup"

This is because these log messages will be read by & inspected by users. Having user friendly log messages goes a long way in helping users understand what is happening in the script.

deploy/docker/utils/bin/utils.js
```diff
@@ -73,10 +73,32 @@ async function listLocalBackupFiles(){ // Ascending order
   return backupFiles;
 }
 
+
+async function updateLastBackupErrorMailSentInMilliSec(ts) {
+  try {
+    await fsPromises.access(Constants.BACKUP_PATH);
+  } catch (error) {
+    await fsPromises.mkdir(Constants.BACKUP_PATH);
+  }
```
sharat87, line 82: This can actually be simplified with the `recursive` option. From the docs:

> Calling fsPromises.mkdir() when path is a directory that exists results in a rejection only when recursive is false.
> -- https://nodejs.org/api/fs.html#fspromisesmkdirpath-options

```suggestion
  await fsPromises.mkdir(Constants.BACKUP_PATH, { recursive: true });
```

deploy/docker/utils/bin/utils.js
```diff
@@ -73,10 +73,32 @@ async function listLocalBackupFiles(){ // Ascending order
...
+}
+
+async function getLastBackupErrorMailSentInMilliSec() {
+  try {
+    await fsPromises.access(Constants.BACKUP_PATH + '/last-error-mail-ts');
+  } catch (error) {
+    return 0;
+  }
+  const ts = await fsPromises.readFile(BACKUP_PATH + '/last-error-mail-ts');
+  return parseInt(ts, 10);
```
sharat87, line 93: The call to `readFile` itself would do the access right? So can we just use that in the `try` block?

```suggestion
  try {
    const ts = await fsPromises.readFile(BACKUP_PATH + '/last-error-mail-ts');
    return parseInt(ts, 10);
  } catch (error) {
    return 0;
  }
```

deploy/docker/watchtower-hooks/pre-check.sh
```diff
@@ -0,0 +1 @@
+exit 0
```
sharat87, line 1: We are making these files _executable_, with the `chmod +x` in the `Dockerfile`. For scripts that'll be executable, it's best to always have a shebang line. It _may_ be okay for us even without it, since these scripts only ever run on Ubuntu, but I'd still advise including the shebang line.


```suggestion
#!/usr/bin/env bash
exit 0
```

Same in the `pre-update.sh` script as well.

sharat87 (changes requested):

deploy/docker/watchtower-hooks/pre-check.sh
sum35h: chore: review change

deploy/docker/watchtower-hooks/pre-update.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+appsmithctl backup --error_mail || exit 1
```
sharat87, line 2: This is `--error-mail`, right?

```suggestion
appsmithctl backup --error-mail || exit 1
```

deploy/docker/utils/bin/utils.js
```diff
@@ -73,10 +73,27 @@ async function listLocalBackupFiles(){ // Ascending order
...
 
+
+async function updateLastBackupErrorMailSentInMilliSec(ts) {
+  await fsPromises.mkdir(Constants.BACKUP_PATH, { recursive: true });
+  await fsPromises.writeFile(Constants.BACKUP_PATH  + '/last-error-mail-ts', ts.toString());
+}
+
+async function getLastBackupErrorMailSentInMilliSec() {
+  try {
+    const ts = await fsPromises.readFile(BACKUP_PATH + '/last-error-mail-ts');
```
sharat87, line 84: Can we move this file name or full path to a constant as well please?

sharat87 (changes requested):

sharat87 (approved):

sharat87 (approved):

PR #15205, opened 2022-07-14
fix: JSONForm default array item not deletable when default value initialized with undefined

### LINKED ISSUE #15136: [Bug]: JSON form doesn't remove first item on clicking delete

### Description

When i add array of data as default and then add a few more items and delete the default data, the data is not getting deleted. This only happens when the default data is coming from a store value.

https://discord.com/channels/725602949748752515/725609493974614076/994795375758164069
https://theappsmith.slack.com/archives/C0341RERY4R/p1657260698310879

### Steps To Reproduce

1. Add a JSON form
2. Bind source data to a store value using array as default data
3. Now update the store value to show the source data
4. Add a few more items in the array
5. Remove the default item
6. Observe it doesn't get removed.

### Public Sample App

https://app.appsmith.com/app/show-delete-error/page1-62cd0abc01213a676a35fd53

### Version

Production

ashit-rath:
## Description
When the default value of an array field is initialized with undefined and then the default value is introduced then the default items in the forms are not removable.

Eg
If the source data during the form load is
```json
{}
```

and after some operation it is modified to 

```json
{
  "arr": [
    { "name": "test" }
  ]
}
```

The array item with the value "test" persists on delete.

Reason - In the `ArrayField` component, the on delete callback  `remove` had a missing dependency in the `useCallback` hook, due to which the stale value was being used during remove and caused inconsistency in the delete operation.

Fixes #15136

## Type of change

## How Has This Been Tested?

Cypress:
1 cypress test is added to validate the use case.
- A JSONForm loaded with appropriate fields but with empty `Source Data` value (to mimic the empty initial state of the form and the array field). 
- Then the source data is added 
- `Add New`  button is clicked to add a new array item
- Some dummy data is inserted to the new array item
- The first array item is delete (which has the field value from the default source data)
- The array item is observed to be deleted and the one that was created using the `Add New` is persisted as the first item in the array

## Checklist:


sbalaji1192 (approved):

chandannkumar:

PR #15228, opened 2022-07-15
Fixes issue where multiselect height decreases on selecting more options

### LINKED ISSUE #13970: [Bug]: MultiSelect widget decreases in height when more options are selected

### Description

The different types of select widget have different box heights if the selected options width exceeds the box width.

image

### Steps To Reproduce

1. Open a new app.
2. Drag one of each select-style widget and place them side by side.
3. Select multiple things in the multi select widgets.
4. Resize the multi-select widget to a smaller width than the options selected.
5. Observe that the height of the box decreases.

### Public Sample App

_No response_

### Version

Cloud 1.7

souma-ghosh:
## Description

Height of multi select widget decreases when we select label position top and then add enough options such that the widget shows "+1 more".
Compared the multi select widget to multi tree select widget (same issue not present in multi tree select) and found that there were 2 CSS properties missing in multi select which were applied in multi tree select. Applying these CSS properties solved the inconsistent height issue with multi select widget

Fixes #13970 

## Type of change

## How Has This Been Tested?

- Manually tested

## Checklist:

### DIFF
--- a/app/client/src/widgets/MultiSelectWidgetV2/component/index.styled.tsx
+++ b/app/client/src/widgets/MultiSelectWidgetV2/component/index.styled.tsx
@@ -415,6 +415,10 @@ export const MultiSelectContainer = styled.div<{
         appearance: none;
       }
     }
+    .rc-select-selection-overflow-item-suffix {
+      position: relative !important;
+      left: 0px !important;
+    }
   }
   && .rc-select-disabled {
     cursor: not-allowed;


app/client/src/widgets/MultiSelectWidgetV2/component/index.styled.tsx
```diff
@@ -415,6 +415,10 @@ export const MultiSelectContainer = styled.div<{
         appearance: none;
       }
     }
+    .rc-select-selection-overflow-item-suffix {
+      position: relative !important;
+      left: 0px !important;
```
Tooluloope, line 420: Avoid using important, use the styled component specificity

Tooluloope (changes requested):

app/client/src/widgets/MultiSelectWidgetV2/component/index.styled.tsx
souma-ghosh: Since some inline styles are getting applied dynamically (by the rc-select library) when there are more options in the multiselect, the only way to override that is by using `!improtant` in this case. Those inline styles itself are causing the height issue.

Tooluloope (approved):

PR #15239, opened 2022-07-15
fix: Unwanted width calculation in Dropdown

### LINKED ISSUE #15018: [Bug]: Dropdown used for property pane connections causes expensive style recalculations

### Description

When there is a evaluation cycle, dropdown used in property pane connections triggers an expensive style recalculation, costing about 40-80ms per cycle. 

Since it's dimensions don't change, there is no need for calling this property which triggers these calculations.
Screenshot 2022-07-05 at 12 26 56 PM

Note: The above profile is with 4x throttling.

### Steps To Reproduce

Create a simple app with couple of widgets with one of them having some incoming and outgoing connections.

### Public Sample App

_No response_

### Version

Both

aswathkk:
## Description

Makes the options width calculation to be done only when required

Fixes #15018

## Type of change

## How Has This Been Tested?

- Manual

## Checklist:


app/client/src/components/ads/Dropdown.tsx
```diff
@@ -1149,12 +1149,31 @@ export default function Dropdown(props: DropdownProps) {
...
-  }
+  const onParentResize = useCallback(
+    debounce(() => {
+      requestAnimationFrame(() => {
+        if (dropdownWrapperRef.current) {
+          const { width } = dropdownWrapperRef.current.getBoundingClientRect();
+          setDropdownWrapperWidth(`${width}px`);
+        }
+      });
+    }, 1000),
```
SatishGandham, line 1164: 1000 is too long, set it under 300.

app/client/src/components/ads/Dropdown.tsx
```diff
@@ -1149,12 +1149,31 @@ export default function Dropdown(props: DropdownProps) {
...
+    [dropdownWrapperRef.current],
+  );
+
+  useEffect(() => {
+    const resizeObserver = new ResizeObserver(onParentResize);
+    if (dropdownWrapperRef.current && props.fillOptions)
+      resizeObserver.observe(dropdownWrapperRef.current);
+
+    return () => {
```
SatishGandham, line 1173: 🍕 for cleaning up :)

SatishGandham (changes requested):

app/client/src/components/ads/Dropdown.tsx
```diff
@@ -1149,12 +1149,31 @@ export default function Dropdown(props: DropdownProps) {
...
+          const { width } = dropdownWrapperRef.current.getBoundingClientRect();
+          setDropdownWrapperWidth(`${width}px`);
+        }
+      });
+    }, 1000),
+    [dropdownWrapperRef.current],
+  );
+
+  useEffect(() => {
+    const resizeObserver = new ResizeObserver(onParentResize);
```
SatishGandham, line 1169: Where are we specifying the element which needs to be observed?
Does it check only width or both? we need to check only width.

app/client/src/components/ads/Dropdown.tsx
aswathkk: @SatishGandham 
We need to recalculate the width only before the user opens the options dropdown.
IMO, 1000 is acceptable.
Because after resizing, it will take at least a second for the user to click and open the dropdown.

SatishGandham: 1second is a long time and there is no downside to reducing the duration. It will ensure we do not degrade experience of users who are fast.

aswathkk: The downside is that we will be doing this geBoundingClientRect calculation. Yeah I agree that we should not degrade experience of users who are fast

app/client/src/components/ads/Dropdown.tsx
aswathkk: Element is specified in line number 1176 `dropdownWrapperRef.current`

app/client/src/components/ads/Dropdown.tsx
aswathkk: Updated to 300ms since I also faced issue being a fast user 😄

SatishGandham (approved):
LGTM.
@aswathkk , keep an eye out for other forced reflows where we can use resizeObserver.

SatishGandham (approved):

PR #15259, opened 2022-07-18
feat: prevent secrets to be pushed to repo

### LINKED ISSUE #15042: [Feature]: Prevent secrets and keys to be pushed into github

### Summary

Recently, we pushed a secret token and key from AWS to our GitHub repository which in turn raised an alarm with AWS and caused a quarantine of the key. We should add a pre-commit hook to check if someone is trying to push a secret or key accidentally and stop them.

To prevent accidental push of org-wide secrets and keys to public domain via our github repo.

ayushpahwa:
## Description

Currently, our repository is prone to any dev pushing secrets to the repository by mistake. This can become a major problem very quickly and in the case of services like AWS, an alarm is raised and the key is deactivated causing apps using the key to break.

After this PR is merged, people will need to have [git secrets](https://github.com/awslabs/git-secrets) installed in their systems and have to register the AWS secrets to the repo.

The setup instructions have been mentioned [here](https://www.notion.so/appsmith/Getting-Started-e81ccc35463343b28e09c200cdbe5f42#c3183e8e6e404701920dd288b8506360)

Notes: 
1. right now, since we are using the pre-commit hooks using husky, the hooks work when we have done yarn install. 
2. Also, the scan only checks the directory we are in while making the change or raising the PR. This is because it uses `git ls-files` to get the list of files to scan.
3. If someone doesn't run `git-secrets --register-aws`, the tests will pass automatically since there is no regex registered with the system to scan against.

Fixes #15042

## Type of change

## Checklist:


mohanarpit:
@ayushpahwa Can you also add the instructions to Contributing.md ? This way, external contributors will also have enough information on how to setup the repo and contribute to our code base.

nidhi-nair:
Tested on my local with the dependency installed and it works for the 20 character alphanumeric rule. We'll need to add more relevant rules as we go though. 
Haven't tested the experience without the installation, but Ayush mentioned that that is meant to fail as well. LGTM.

mohanarpit (approved):

PR #15314, opened 2022-07-20
fix: Log users out to fix serialization error from upgrading Spring

sharat87:


mohanarpit (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
```diff
@@ -1367,4 +1368,10 @@ private static Map<String, Object> updateWhereClauseFormat(Object oldWhereClause
 
         return newWhereClause;
     }
+
+    @ChangeSet(order = "021", id = "flush-spring-redis-keys-2", author = "")
+    public void migrateGoogleSheetsToUqi(ReactiveRedisOperations<String, String> reactiveRedisOperations) {
```
trishaanand, line 1373: @sharat87 : Would it be possible to change the name of the function?

PR #15323, opened 2022-07-20
chore:  fix cannot read properties of null (reading 'id') sentry error

### LINKED ISSUE #15220: TypeError: Cannot read properties of null (reading 'id')
Sentry Issue: [APPSMITH-33E](https://sentry.io/organizations/appsmith/issues/3423630864/?referrer=github_integration)

```
TypeError: Cannot read properties of null (reading 'id')
  at call (entities/Engine/AppEditorEngine.ts:173:33)
...
(45 additional frame(s) were not displayed)
```

akash-codemonk:
## Description

Fixes #15220 

When one of the api calls fails in [loadAppEntities](https://github.com/appsmithorg/appsmith/pull/15323/files#diff-c5adb6ac7260cbbb59606009da3fc3b2ef51bd56a03dc5fab9e34399fd48290aR165)(fetching js collections, actions etc)due to an invalid session we continue to call the proceeding methods `loadGit` and `completeChore` even when the user is logged out. So the store values are reset or undefined which is when a call to `completeChore` throws an error at https://github.com/appsmithorg/appsmith/pull/15323/files#diff-c5adb6ac7260cbbb59606009da3fc3b2ef51bd56a03dc5fab9e34399fd48290aR179 which is reported by sentry.

Release

https://user-images.githubusercontent.com/67054171/179992982-cfa6c992-12f6-4bfb-a723-a94b713e990b.mov

Post fix

https://user-images.githubusercontent.com/67054171/179993414-8d3be91f-f33d-4474-9c3e-463c88452f80.mov

## Type of change

## How Has This Been Tested?

Manually

## Checklist:


arunvjn:
@akash-codemonk Would it be possible to throw specific errors and handle it in the catch block of the ```startAppEngine``` method ?

app/client/src/sagas/InitSagas.ts
```diff
@@ -89,7 +89,13 @@ export function* startAppEngine(action: ReduxAction<AppEnginePayload>) {
...
     );
     yield call(engine.loadAppURL, toLoadPageId, action.payload.pageId);
-    yield call(engine.loadAppEntities, toLoadPageId, applicationId);
+    const loadAppEntities: boolean = yield call(
+      engine.loadAppEntities,
+      toLoadPageId,
+      applicationId,
+    );
+    // If loading application entities fails we do not proceed further
+    if (!loadAppEntities) return;
```
arunvjn, line 98: If possible avoid such conditionals here. Can we create error classes like ActionNotFound, ThemeNotFound etc, and skip sentry logging for these errors ?

app/client/src/sagas/InitSagas.ts
```diff
@@ -95,7 +100,15 @@ export function* startAppEngine(action: ReduxAction<AppEnginePayload>) {
...
+    if (
+      !(
+        e instanceof PluginFormConfigsNotFoundError ||
+        e instanceof PluginsNotFoundError ||
+        e instanceof ActionsNotFoundError
+      )
+    ) {
+      Sentry.captureException(e);
+    }
     if (e instanceof PageNotFoundError) return;
```
arunvjn, line 112: I think we can skip sentry error logging for PageNotFound error as well, since it usually happens around the time you switch branch. 
Also can you create a base error class, something like ```EngineAPIError``` that extends Error. 

```
class EngineAPIError extends Error {}
class PageNotFound extends EngineAPIError {}
class PluginsNotFoundError extends EngineAPIError {}
```
And in the catch block we can skip sentry error logging for all instances of EngineAPIError

akash-codemonk: Also returning early for these errors. `SAFE_CRASH_APPSMITH_REQUEST` does get called when these apis have errors https://github.com/appsmithorg/appsmith/blob/release/app/client/src/sagas/InitSagas.ts#L65

arunvjn: Yes makes sense 👍🏽. Please make sure to check this for published apps as well.

akash-codemonk: Throwing a error there as well when api calls fail. Works fine on view page.

arunvjn (approved):
Looks good

PR #15348, opened 2022-07-21
feat: Control where embedding of Appsmith is allowed

sharat87:
Currently, Appsmith apps can be embedded in frames on other websites, with no restrictions/controls. This PR adds fine controls on this.

By default, embeddind in frames is _only_ allowed on the same domain. That is, if your Appsmith is available at `http://mydomain.com`, then only pages on `http://mydomain.com` will be able to embed apps.

To change/customize this, we've introduced `APPSMITH_ALLOWED_FRAME_ANCESTORS` environment variable. To allow another domain like `http://trusted-other.com` to also embed apps from your Appsmith, use:

```sh
APPSMITH_ALLOWED_FRAME_ANCESTORS="'self' http://trusted-other.com"
```

Or, to allow all subdomains on `mycompany.com`, use:

```sh
APPSMITH_ALLOWED_FRAME_ANCESTORS="'self' http://*.mycompany.com"
```

Under the covers, this feature uses a `Content-Security-Policy` header with `frame-ancestors` directive. Learn more at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors.


mohanarpit (approved):

mohanarpit:
@sharat87 For my understanding, once this PR is merged, existing embedded Appsmith apps will stop working right? If yes, is there a way that we can communicate this clearly to them?

sharat87:
@mohanarpit, yes. We should add this to our release notes, but other than emailing everybody and announcing on Discord, I don't think there's much we can do. Once this is merged, I'll add this to our docs as well so we have a link to share along with these announcements. Does that sound okay?

mohanarpit:
@sharat87 I'd suggesting adding this to docs earlier (read: right now). This way, we'll have the resolution recorded for users and we can link to that page in our Discord, Release notes, and any other communication. I'm suggesting this because the documentation team will also need to allocate bandwidth to add this change.

sharat87:
@mohanarpit Sure. I was going to do it after it goes to release, and before it goes to master, but this PR being approved is enough reason to believe that it'll be merged anyway. I'll start drafting up a page for this right away. Thanks!

mohanarpit (approved):

sharat87:
Docs PR at https://github.com/appsmithorg/appsmith-docs/pull/430

lodhi2017294:
Hi Everyone,
Kindly Please help 
Refused to frame 'our self deployed domain' because an ancestor violates the following Content Security Policy directive: "frame-ancestors 'self'". 
I am getting the above error 
on our self deployed domain
As per your solution, we have to add --> APPSMITH_ALLOWED_FRAME_ANCESTORS="'self' http://trusted-other.com/"
But I got confused where to add this one --> APPSMITH_ALLOWED_FRAME_ANCESTORS="'self' http://trusted-other.com/"
Appsmith apps doesn't have any option to add this one

sharat87:
Hey @lodhi2017294, sorry this wasn't clear. This is an environment variable to be added to the file `stacks/configuration/docker.env` on the server where your Appsmith instance is running. This `stacks` folder should be located in the same folder as your `docker-compose.yml` file.

rodrigo-mahamud:
I'm having this same issue: Refused to frame 'our self deployed domain' because an ancestor violates the following Content Security Policy directive: "frame-ancestors 'self'". 

But i have set up my app in appsmith cloud and i want to embed the app in my custom domain. But i don't know where i have to indicate that i want to connect to appsmith from my html iframe

PR #15368, opened 2022-07-22
feat: JSON Session serialization on Redis

### LINKED ISSUE #15315: [Enhancement]: Store sessions as JSON data instead of using JDK serialization of Spring objects

### Description

Spring classes change at every Spring release, and since we use JDK serialized versions of these objects, they fail to deserialize when we upgrade Spring.

To avoid this, we need to own the serialization format for sessions. Perhaps JSON, and not have it dependent on Spring version in use.

### Steps To Reproduce

Start an Appsmith instance at version `v1.7.8`, signup/login to create a session, then switch to a version with updated Spring, like the current `release` branch at the time of writing, or `v1.7.9` when it's out, and try to load your applications page again. You should see an error page.

### Public Sample App

_No response_

### Version

Both

sharat87:
Fixes #15315


app/server/appsmith-server/src/main/java/com/appsmith/server/domains/UserSession.java
```diff
@@ -0,0 +1,112 @@
...
+
+        return session;
+    }
+
+    /**
+     * Performs the reverse of fromToken method. Given a UserSession object, create a Spring Security authentication
+     * token. This authentication token can then be wrapped in a SecurityContext and used as the user's session.
+     * @return A Spring Security authentication token representing the user's session. Usually an instance of UsernamePasswordAuthenticationToken or Oauth2AuthenticationToken.
+     */
+    public Authentication makeToken() {
```
nidhi-nair, line 89: For my understanding, if we are to extend the props inside of `UserSession` in the future, would that invalidate the existing token? Or do we limit the identifiers to the current set and only invalidate in case there is a change in how we identify a unique user?

sharat87: No problem in extending. It's JSON serialization, so it works just like adding a new field to one of our domain classes and then loading JSON data from MongoDB. The new fields just take `null` or their default values, if set.

nidhi-nair: Oh so what that means is that for the duration that that session is valid, the new fields would not be present in that user's token? When would the new session token be created in that case? Sorry I feel a little muddled about this!

sharat87: Yes. The session is recreated only when they logout and login again. But, that shouldn't be a problem We don't expect to be adding fields to this class regularly. The primary purpose is to give the email of the logged-in user. Additional details can be fetched from the DB accordingly. We can include some extra information in the session object to save those DB calls, if needed, but then we need to have a way to update the session object as well, when something changes. Take name for example. If we were to save it in the Session object as well, then, when the user changes their name in the Profile page, we'll have to update the Session object as well.

If we add new fields to the session, they will be present only in new sessions. I'd rather we handle the case where that new field is `null`, instead of setting up some form of migrations for this.

nidhi-nair: Gotcha! Thanks for the explanation! :)

nidhi-nair (approved):

PR #15377, opened 2022-07-22
feat: db migrations of objects to new rbac model

sidhantgoel:
> Pull Request Template
>

## Description

Fixes # (issue)

## Type of change

## How Has This Been Tested?

- Test A
- Test B

## Checklist:


trishaanand (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
```diff
@@ -1075,7 +1092,357 @@ public void addAnonymousUser(MongockTemplate mongockTemplate) {
...
+        viewerPermissionGroup = mongockTemplate.save(viewerPermissionGroup);
+
+        return Set.of(adminPermissionGroup, developerPermissionGroup, viewerPermissionGroup);
+    }
+
+    @ChangeSet(order = "021", id = "add-default-permission-groups", author = "")
+    public void addDefaultPermissionGroups(MongockTemplate mongockTemplate, WorkspaceService workspaceService, @NonLockGuarded PolicyUtils policyUtils) {
+        // Drop PermissionGroup collection
+        // This ensures that migration can run again if aborted in between
+        mongockTemplate.dropCollection(PermissionGroup.class);
```
trishaanand, line 1263: Instead of dropping the collection, could we instead only run do this for workspaces who do not have a permission group associated yet? This would be a long running migration. Instead of re-running the same piece, it might be helpful to run for workspaces which haven't been migrated yet.

sidhantgoel: We could do that but there will be stale permission groups remaining if it drops before the point workspace has not been saved but permission groups was, so we need to delete those also otherwise they will be present in queries with default permission group, this is a simplest solution, I know it is a long running migration but it should run in one go, this I just added as a precaution.

PR #15405, opened 2022-07-24
chore: Optimise code editor renders

### LINKED ISSUE #15157: [Bug]: Optimize codeeditor data subscriptions

### Description

CodeEditor subscribes to dynamic data but it is not used for display purposes. Dynamic data changes on every eval cycle causing all code editors on the screen to rerender.

DynamicData is used for the hinters and its updated on component mount and update, since this is only required on focus.

Remove the subscription to dynamic data at component level and get the data directly from store on focus or hover.

While you are here see if we can remove any further subscriptions to the redux store. `recentEntities` in particualr.

### Steps To Reproduce

- Create an app with single button
- Make all possible properties into JS
- Store `new Date()` on button click to trigger an eval cycle

1. Take a react profile of the button click, you will see all the codeditors render.
2. Take a chrome profile of button click, you will see purple color in the profile marking render of codeeditor.

**GOAL**: CodeEditor should not re-render

- Ensure that the UX is not changed in anyway due to doing this work on demand.
- Run the cypress tests ASAP after initial work to estimate the time required to fix failing cypress tests if any.

### Public Sample App

_No response_

### Version

Both

arunvjn:
## Description
Modifying one code editor instance in API pane would re-render all the other code editor components in the page as well, since it subscribes to dataTree. This code change stops unfocused code editors from re-rendering. Also, removed unused code.

## Type of change
- Perf fix

## How Has This Been Tested?
- Manual

## Checklist:


app/client/src/components/editorComponents/CodeEditor/index.tsx
```diff
@@ -351,36 +351,28 @@ class CodeEditor extends Component<Props, State> {
...
-          this.editor.setValue("");
-        }
-        CodeEditor.updateMarkings(this.editor, this.props.marking);
-      } else {
-        // Update the dynamic bindings for autocomplete
-        if (prevProps.dynamicData !== this.props.dynamicData) {
-          this.hinters.forEach(
-            (hinter) => hinter.update && hinter.update(this.props.dynamicData),
-          );
+      // const currentMode = this.editor.getOption("mode");
```
SatishGandham, line 356: This looks like a callback. I think `if (this.state.isFocused) return;` should be inside.

Rishabh-Rathod: Also, as we are registering a callback here won't the `prevProps and this.props` will have stale closure values that gets updated on `props` change?

arunvjn: Both the values should be the same since editor.operation is a sync call. It was only a syntactical change and is not relevant to this PR. Removing it.

app/client/src/components/editorComponents/CodeEditor/index.tsx
```diff
@@ -794,6 +786,12 @@ class CodeEditor extends Component<Props, State> {
     };
   };
 
+  shouldComponentUpdate(nextProps: Props, nextState: State) {
+    if (this.props.dynamicData !== nextProps.dynamicData)
+      return nextState.isFocused;
+    return true;
```
SatishGandham, line 792: Shouldn't the default return value be false?

arunvjn: Nope. `shouldComponentUpdate` is called when props or state change and the default behaviour for such changes is to re-render.

SatishGandham: Does dynamic data change independently?

arunvjn: I don't understand the question. Can you give an example here ? Anything that will trigger eval will modify dynamic data

SatishGandham: I was thinking about what happens when widget data changes, but line 791 returns false.
But for focused widget it will return true, so no problem.

SatishGandham (dismissed):

SatishGandham (commented):
Check with @rahulramesha and see if we can fix the cyclic dependeceny and go with the original solution we thought.

Can you take a profile with subscriptions removed and see the improvement we are going to get.

arunvjn:
No of re-renders for keypress event in property pane with 10 code mirror instances has come down from 15 to 5. @SatishGandham

arunvjn:
Since we were getting stale values of evaluated values for the code editor instances, I have reverted the code and it now uses components lifecycle methods to avoid re-renders. 
We need to recompute the new dataTree on every render cycle, since linting & validation (which gets called on every render) relies on having the latest dataTree. Having the dataTree be passed as a prop here is the right way to go here.
@SatishGandham

app/client/src/components/editorComponents/CodeEditor/index.tsx
```diff
@@ -785,7 +782,10 @@ class CodeEditor extends Component<Props, State> {
       this.state.hasLintError && this.setState({ hasLintError: false });
     }
 
-    const pathEvaluatedValue = _.get(dataTree, getEvalValuePath(dataTreePath));
+    const pathEvaluatedValue = _.get(
+      this.props.dynamicData,
```
SatishGandham, line 786: Why are we changing dataTree to dynamicdata?

SatishGandham (commented):
Where are we handling it differently for JS Objects?

app/client/src/components/editorComponents/CodeEditor/index.tsx
arunvjn: dataTree is dynamicData. Only the function signature has changed here

SatishGandham (approved):

app/client/src/components/editorComponents/CodeEditor/index.tsx
```diff
@@ -353,37 +353,35 @@ class CodeEditor extends Component<Props, State> {
     window.addEventListener("keydown", this.handleKeydown);
   }
 
+  shouldComponentUpdate(nextProps: Props, nextState: State) {
+    if (this.props.dynamicData !== nextProps.dynamicData)
+      return nextState.isFocused || !!nextProps.isJSObject;
```
ohansFavour, line 358: This change seems to have affected linting. Check out this recording (https://www.loom.com/share/a0022142ddc9419ea4be504af6a99bb1)

@arunvjn @SatishGandham

arunvjn: @ohansFavour Right fix for this is to move the lintCode logic outside the code editor component? Can we have a place in redux store that contains all the lint errors ?

ohansFavour: @arunvjn I think the lintCode logic will still be needed in the code editor component, this logic only helps to add the squiggle line to the code editor (actual determination of lint errors in code happens in the worker during the eval flow). Having a central place in the redux store for lint errors makes sense, I will work on this.

arunvjn: Yes, thats right. We need to avoid trigger to compute lint errors from codeEditor if we are doing it. 
Ideal behaviour is for us to remove the dataTree subscription on the codeEditor entirely. 
dataTree is being used to compute/look validation errors, lint errors and slash commands. We want to be able to get to a state where we would simply look up the exact lint error using the dataTreePath and add squiggle lines on the editor. Same goes with validation errors as well.

PR #15440, opened 2022-07-25
chore: Better server logging infra

### LINKED ISSUE #9457: [Bug] NullPointerException: iterator returned a null value
Sentry Issue: [APPSMITH-1C6](https://sentry.io/organizations/appsmith/issues/2798744294/?referrer=github_integration)

```
NullPointerException: iterator returned a null value
    at reactor.core.publisher.FluxIterable$IterableSubscription.poll(FluxIterable.java:424)
```

nidhi-nair:
## Description
The primary intent of this PR was to include request metadata into Sentry for the error logging that we have. This was not possible to do so far because the `GlobalExceptionHandler` loses track of the context of the user by the time flow of execution reaches there. There are a bunch of fixes that this PR introduces to address this, and a few other concerns with our logging.

1. We now explicitly pass the context map via Appsmith exceptions so that the exception handler can use this information while logging errors.
2. Introduced a hook on to Webflux signals so that the MDC context from the original subscription flows into all the threads that are spawned during the course of an API request execution. This makes sure that the error signals also retain the original MDC context data so that it may flow through to Sentry
3. A side effect of this is that every thread within our server now logs with context (yay!)
4. Also worth mentioning is that we used to record the original thread of the request in the MDC context. Since these threads are meant to be changing across the course of execution, the context has been made to update with the new thread names where applicable. However, I wonder about the usefulness of this information at the present moment. Leaving the information in till we qualify it as unnecessary, if at all.
5. The plugin modules were not printing slf4j logs. The issue was minor here, we hadn't enabled debug level logs for that directory.
6. The print statements have also been converted into the appropriate slf4j log statements. This has been done in an almost 1-to-1 manner to avoid complicating the PR. We can revisit cleaning up logs in a subsequent change.
7. And of course, the addition of this information into Sentry.
8. Have also now included an ISO8601 formatted timestamp to the start of every log statement.

Fixes #9457

## Type of change
- Chore

## How Has This Been Tested?
- Manually tested by comparing logs and object built for Sentry

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/MDCConfig.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/MDCConfig.java
@@ -15,9 +15,9 @@ import javax.annotation.PreDestroy;
 import java.util.Map;
 
 @Configuration
-public class MdcConfig {
+public class MDCConfig {
 
-    private static final String MDC_CONTEXT_REACTOR_KEY = "MdcConfig";
+    private static final String MDC_CONTEXT_REACTOR_KEY = "MDCConfig";
 
     @PostConstruct
     void contextOperatorHook() {
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/exceptions/BaseException.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/exceptions/BaseException.java
@@ -1,9 +1,18 @@
 package com.appsmith.external.exceptions;
 
+import lombok.Getter;
+import org.slf4j.MDC;
+
+import java.util.Map;
+
+@Getter
 public abstract class BaseException extends RuntimeException {
+
+    private Map<String, String> contextMap;
 
     public BaseException(String message) {
         super(message);
+        contextMap = MDC.getCopyOfContextMap();
     }
 
     public abstract Integer getHttpStatus();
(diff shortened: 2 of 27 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/MDCConfig.java
```diff
@@ -0,0 +1,87 @@
...
+
+        @Override
+        public void onNext(T t) {
+            this.copyToMdc(coreSubscriber.currentContext());
+            coreSubscriber.onNext(t);
+        }
+
+        @Override
+        public void onError(Throwable throwable) {
+            coreSubscriber.onError(throwable);
```
subrata71, line 58: Hey @nidhi-nair ! 
In the case of `onError` event, the current context is not copied only the error is propagated. Is this done on purpose?

nidhi-nair: Hey! So onError signals don't switch threads, because the follow on the subscriber's thread itself. onError and onComplete are terminal signals, so they always go back to the subscriber after these calls are made.

subrata71 (approved):
Looks good to me

nidhi-nair:
Added test plan approved since the change only requires regression testing.

PR #15470, opened 2022-07-26
fix: Added type static hierarchy to allow compatible types

### LINKED ISSUE #15444: [Bug]: Google Sheets - Issue with a column that has different data types

### Description

If a column in Google sheets has different data types, for eg; Numbers and Text, only one of them is fetched in the data and the other returns empty values.

Screenshot 2022-07-26 at 10 45 20
Screenshot 2022-07-26 at 10 46 36
Screenshot 2022-07-26 at 10 47 28

### Steps To Reproduce

1. In a column on Google Sheets, add both numbers and text
2. Fetch the data on Appsmith and view the response
3. You will see values in certain objects missing

### Public Sample App

_No response_

### Version

v1.7.9

nidhi-nair:
## Description

The issue that this PR is attached to suffers from two problems. 
1. When the data type in the first row of the data is a string, and a subsequent row sends an integer type, we were not displaying the integer type data because of type mismatch
2. When the data type in the first row of the data is an integer, and a subsequent row sends a string type, we were not displaying the string type data because of type mismatch

The first issue is a matter of raising the type in its hierarchy, which is a compatible change. This PR introduces this hierarchy for our types to fix this issue. However, this is a very hacky fix to get the user out of this state of unusability. Ideally, we would want to rely on a single known type for that column.

The second issue is not handled by this PR at all. This is because the expectation is to widen the understood schema which we do not have the capability for currently. We will be able to solve this holistically as a result of one of the data type handling project phases.

Initial problem:
Screenshot 2022-07-26 at 11 20 04 PM
Screenshot 2022-07-26 at 11 20 41 PM

Current fix:
Screenshot 2022-07-26 at 11 22 25 PM

In order to fix the second issue, a workaround that users can use for now is to use a dummy row in their data that specifies data in the expected data type. For example, with the following data: 
Screenshot 2022-07-26 at 11 23 59 PM

We can get the following result:
Screenshot 2022-07-26 at 11 24 21 PM

Be wary that in order to show clean results within Appsmith, we would need to offset pagination by 1 (to get rid of the row that specifies data types)

Fixes #15444

## Type of change

## How Has This Been Tested?
- JUnit test
- Manually tested

## Checklist:

### DIFF
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/ce/FilterDataServiceCE.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/ce/FilterDataServiceCE.java
@@ -78,6 +78,17 @@ public class FilterDataServiceCE implements IFilterDataServiceCE {
             ConditionalOperator.NOT_IN, "NOT IN"
     );
 
+    private static final Map<DataType, Set<DataType>> datatypeCompatibilityMap = Map.of(
+            DataType.INTEGER, Set.of(),
+            DataType.LONG, Set.of(DataType.INTEGER),
+            DataType.FLOAT, Set.of(DataType.INTEGER, DataType.LONG),
+            DataType.DOUBLE, Set.of(DataType.INTEGER, DataType.LONG, DataType.FLOAT),
+            DataType.BOOLEAN, Set.of(),
+            DataType.STRING, Set.of(DataType.INTEGER, DataType.LONG, DataType.FLOAT, DataType.DOUBLE, DataType.BOOLEAN, DataType.DATE, DataType.TIME, DataType.TIMESTAMP),
+            DataType.DATE, Set.of(),
+            DataType.TIMESTAMP, Set.of()
+    );
+
     public FilterDataServiceCE() {
 
         objectMapper = new ObjectMapper();
@@ -119,7 +130,7 @@ public class FilterDataServiceCE implements IFilterDataServiceCE {
         insertAllData(tableName, items, schema, dataTypeConversionMap);
 
         // Filter the data
-        List<Map<String, Object>> finalResults = executeFilterQueryNew(tableName, schema, uqiDataFilterParams);
+        List<Map<String, Object>> finalResults = executeFilterQueryNew(tableName, schema, uqiDataFilterParams, dataTypeConversionMap);
 
         // Now that the data has been filtered. Clean Up. Drop the table
         dropTable(tableName);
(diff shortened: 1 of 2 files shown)


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/ce/FilterDataServiceCE.java
```diff
@@ -78,6 +78,17 @@ public class FilterDataServiceCE implements IFilterDataServiceCE {
             ConditionalOperator.NOT_IN, "NOT IN"
     );
 
+    private static final Map<DataType, Set<DataType>> datatypeCompatibilityMap = Map.of(
+            DataType.INTEGER, Set.of(),
```
subrata71, line 82: As you mentioned in the description of the PR it's not fixing the data type issue for the following scenario. 
So for a dataset like the below, it will return null/empty in the corresponding value of the column

| SL  | Name |
| ------------- | ------------- |
| 1  | Front end  |
| Two  | Backend  |

```
[
  {
    "SL": 1,
    "Name": "Front end"
  },
  {
    "SL": "",
    "Name": "Backend"
  }
]
```

I am trying to verify my understanding on the same.

subrata71 (approved):
The change looks good according to the description of the PR.

o2lkd:
Hi, 
Thanks for looking into this issue #15444 , but this work around doesn't seem ideal in my use-case. Will there be a fix deployed to this instead of adding a dummy row?
Thanks.

nidhi-nair:
Hey @o2lkd ! First off, I'm sorry that you had to face the issue at all. This work around is due a current limitation in our implementation. We do have another project running that will fix the underlying problem as well, but please expect some delay in seeing that fix since there are a few core changes required to get to that solution. 

If you don't mind could you please tell us a little more about why this workaround does not work for you? Let's try to figure out if there's another way we can fix this for you?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/ce/FilterDataServiceCE.java
nidhi-nair: Yes, that's correct!

o2lkd:
Hey @nidhi-nair,

Thank you for getting in touch, really appreciate the support I've received on this! Hope the fix deploys soon. 

The issue is with adding another dummy row in the database, my team including myself are running some complex automation on the data-set through external APIs and services, and adding row will trigger them. Its not ideal to pause them all for this plus the affected data at the moment is low, but I'm expecting more dataflow of same type in the future.  Hope this clarifies your question.

nidhi-nair:
Thanks for you inputs on this @o2lkd. I'm about to merge this change in, let's continue our conversation on [this new issue](https://github.com/appsmithorg/appsmith/issues/15519) I've created to track possible solutions to this predicament.

nidhi-nair:
Manual QA approval acquired, automation not possible since this involves a Google Sheets query execution.

o2lkd:
Hi @nidhi-nair , came here to say that data is now displaying correctly in the app! Really appreciate your help and support.

PR #15563, opened 2022-08-01
fix: undo redo property pane border focus

### LINKED ISSUE #14735: [Bug]: Undo Redo flashing is not visible anymore

### Description

When we do `Ctrl + z`, we used to get a yellow flash on the property control where the undo is applied.
Looks like that is no longer available.

### Steps To Reproduce

Make some changes to any widget's property and press `Ctrl + z`

### Public Sample App

_No response_

### Version

release

yaldram:
## Description
- When we do undo after changing the Properties from the Property Pane we should scroll to the element and show a yellow border (basically focus on the element).

Fixes #14735

## Type of change

## How Has This Been Tested?

- Tested Manually on the canvas.

## Checklist:


app/client/src/sagas/ReplaySaga.ts
```diff
@@ -119,9 +119,13 @@ export function* openPropertyPaneSaga(replay: any) {
...
     }
 
+    /**
+     * Property Updates - ['widgets', 'hluiy8se0g', 'menuColor']
+     * We take the 1st and 2nd element to generate the element id.
+     */
     flashElementsById(
       btoa(
-        replay.widgets[replayWidgetId].propertyUpdates.slice(0, 2).join("."),
+        replay.widgets[replayWidgetId].propertyUpdates.slice(1, 3).join("."),
```
jsartisan, line 128: How was this working before? Double-check this, please.

rahulramesha: @yaldram, as @jsartisan mentioned, could you check what changed

yaldram: Yeah I did triarge this issue, I think the ordering might have changed. The above changes are only affecting the flashing while undo, so there won't be any other side-effects for sure. Also our goal was to create the same ids as the property pane controls so we get it from the 1 and 2nd elements.

rahulramesha (approved):
LGTM

albinAppsmith (approved):
Approving ADS changes

PR #15591, opened 2022-08-01
fix: fix Mongo plugin query execution failure 

### LINKED ISSUE #15548: [Bug]: Mongo plugin: queries are erroring out with message "state should be: server session pool is open"

### Description

- Mongo plugin: queries are erroring out with message "state should be: server session pool is open"
- Ref: https://theappsmith.slack.com/archives/C0341RERY4R/p1659076053299979

### Steps To Reproduce

https://theappsmith.slack.com/archives/C0341RERY4R/p1659076053299979

### Public Sample App

_No response_

### Version

self hosted, cloud

# Temporary fix
- It may be temporarily fixed by either re-saving the datasource or creating a new datasource and linking the queries against it.

sumitsum:
## Description
- Currently some cases have been seen where a Mongo plugin query fails with error message: `state should be: server session pool is open`. 
  - As per my investigation, this occurs because the connection pool at this point has already been closed. Hence running any query against the closed connection pool results in a `IllegalStateException`. 
  - At this point it is not known what is causing the connection pool to close. 
- This PR introduces a change to handle the `IllegalConnectionException` as a `StaleConnectionException` which would result in the connection pool being re-created. 
  - Note: It was not possible to check for closed connection before executing the query because the `MongoClient` object does not seem to expose any `isClosed` kind of method.  

Fixes #15548

## Type of change

## How Has This Been Tested?
- Manual
- JUnit TC

## Checklist:

### DIFF
--- a/app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
+++ b/app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
@@ -617,7 +617,7 @@ public class MongoPlugin extends BasePlugin {
                     .map(uriString -> MongoClients.create(
                             MongoClientSettings
                                     .builder()
-                                    .applyToConnectionPoolSettings(pool -> pool.applySettings(ConnectionPoolSettings.builder().minSize(0).build()))
+                                    .applyToConnectionPoolSettings(t -> t.applySettings(ConnectionPoolSettings.builder().minSize(0).build()))
                                     .applyConnectionString(new ConnectionString(uriString))
                                     .build()
                     ))
--- a/app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
+++ b/app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
@@ -336,6 +336,15 @@ public class MongoPlugin extends BasePlugin {
                                     error.getErrorMessage()
                             )
                     )
+                    /**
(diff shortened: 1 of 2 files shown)


app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
```diff
@@ -810,6 +819,21 @@ public Mono<DatasourceStructure> getStructure(MongoClient mongoClient, Datasourc
...
+                     * This is to catch the cases when Mongo connection pool closes for some reason and hence throws
+                     * IllegalStateException when query is run.
+                     * Ref: https://github.com/appsmithorg/appsmith/issues/15548
+                     */
+                    .onErrorMap(
+                            IllegalStateException.class,
+                            error -> new StaleConnectionException()
+                    )
+                    // This is an experimental fix to handle the scenario where after a period of inactivity, the mongo
+                    // database drops the connection which makes the client throw the following exception.
```
nidhi-nair, line 832: QQ, where/when are we seeing this error?

sumitsum: @nidhi-nair this was added earlier (not sure when, would have to check the commit history) to `execute` method. I have justed ported it to `getStructure` as well.

nidhi-nair (approved):

sumitsum:
@prapullac skipping cypress tests for this : https://theappsmith.slack.com/archives/C02KTH6EHUG/p1659367917217739

PR #15650, opened 2022-08-03
fix: JSONForm field event trigger binding get updated formData values

### LINKED ISSUE #14302: [Bug]: Previously selected option is displayed on alert message in Select type field of JSONForm

### Description

When we select a option from the list of select type field in JSONForm, the alert message displayed the previous selected option instead of current selected option
https://www.loom.com/share/0da050c030e041878a3c6b4aec22402d

P.S: Checked with Select Widget, the behavior is correct in Select widget

### Steps To Reproduce

1. Add JSONForm onto the canvas
2. Add select type field
3. Configure onOptionChange as {{showAlert(JSONForm1.formData.employee_id)}}
4. Observe that the alert message appears are always of the previously selected options

### Public Sample App

_No response_

### Version

Cloud/ Production

ashit-rath:
## Description
If a field in JSONForm has an onChange event enabled, and the formData is used in the onChange event code, then the event binding always gets the old `formData` values.

Fixes #14302 

Reason :
Internally formData values were updated after the event got triggered.

Solution:
When any formData change dependent event is fired, the event is queued and only triggered after the formData has the updated values.

## Type of change

## How Has This Been Tested?
Cypress tests were added for all the widgets that have an onChange like event.

Steps
1. Source data (with only one field) is added to JSONForm and field type is modified to the field considered for testing.
2. The JS mode is enabled and a binding is added. Eg. `{{showAlert(formData.color)}}`
3. The field value is changed
4. An alert is observed with the latest value change.

## Checklist:


sbalaji1192 (approved):

laveena-en:
Tested and verified.

PR #15656, opened 2022-08-03
fix: navigation issue in the Iframe widget due to src change

### LINKED ISSUE #15585: [Bug]-[64]:iFrame - On setting onSrcDocChanged event to navigate to a different page, user is unable to return to the original page where the iFrame is present.

### Description

https://www.loom.com/share/c99ecf443f2944e1aae6082ce067a355

### Steps To Reproduce

1. Drop an iFrame widget and add in the srcDoc details
2. Set the onClick event to navigate to a different page (same window) onSrcDocChanged
3. Change the content of the SrcDoc
4.  User gets navigated to the set page (page 1 in this case) but is unable to come back to page 4 (original page)

### Public Sample App

_No response_

### Version

Cloud

techbhavin:
## Description

> Fixed the navigation issue in the Iframe widget due to the src change.

Fixes #15585 

## Type of change

## Checklist:


app/client/src/widgets/IframeWidget/component/index.tsx
```diff
@@ -115,8 +116,8 @@ function IframeComponent(props: IframeComponentProps) {
   }, [source]);
 
   useEffect(() => {
-    if (isFirstRender.current) {
-      isFirstRender.current = false;
+    if (isFirstSrcDocRender.current) {
```
keyurparalkar, line 119: Is this the condition that prevents the onSrcDocChange to execute on first render? @techbhavin

app/client/src/widgets/IframeWidget/component/index.tsx
```diff
@@ -102,8 +103,8 @@ function IframeComponent(props: IframeComponentProps) {
   }, []);
 
   useEffect(() => {
-    if (isFirstRender.current) {
-      isFirstRender.current = false;
+    if (isFirstURLRender.current) {
```
keyurparalkar, line 106: @techbhavin are we renaming the variable here?

keyurparalkar (changes requested):
@techbhavin can we add cypress test cases here

app/client/src/widgets/IframeWidget/component/index.tsx
techbhavin: Yes @keyurparalkar

laveena-en:
Tested and verified.

keyurparalkar (approved):
LGTM!

PR #15684, opened 2022-08-03
feat: Added more data to existing analytics events

vishnu-gp:
## Description
This PR adds more data to existing analytics events for audit purposes

## Type of change

## How Has This Been Tested?

- Local tests

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCEImpl.java
@@ -165,7 +165,7 @@ public class AnalyticsServiceCEImpl implements AnalyticsServiceCE {
         }
 
         // In case of action execution, event.getEventName() only is used to support backward compatibility of event name
-        final String eventTag = event.equals(AnalyticsEvents.EXECUTE_ACTION) ? event.getEventName() : event.getEventName() + "_" + object.getClass().getSimpleName().toUpperCase();
+        final String eventTag = AnalyticsEvents.EXECUTE_ACTION.equals(event) ? event.getEventName() : event.getEventName() + "_" + object.getClass().getSimpleName().toUpperCase();
 
         // We will create an anonymous user object for event tracking if no user is present
         // Without this, a lot of flows meant for anonymous users will error out
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCEImpl.java
@@ -198,8 +198,8 @@ public class AnalyticsServiceCEImpl implements AnalyticsServiceCE {
                     analyticsProperties.put("oid", object.getId());
                     if (extraProperties != null) {
                         analyticsProperties.putAll(extraProperties);
-                        // To avoid sending extra audit data to analytics
-                        analyticsProperties.remove(FieldName.AUDIT_DATA);
+                        // To avoid sending extra event data to analytics
+                        analyticsProperties.remove(FieldName.EVENT_DATA);
                     }
 
                     sendEvent(eventTag, username, analyticsProperties);
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCEImpl.java
@@ -164,7 +164,8 @@ public class AnalyticsServiceCEImpl implements AnalyticsServiceCE {
             return Mono.just(object);
         }
 
-        final String eventTag = event.getEventName() + "_" + object.getClass().getSimpleName().toUpperCase();
+        // In case of action execution, event.getEventName() only is used to support backward compatibility of event name
+        final String eventTag = event.equals(AnalyticsEvents.EXECUTE_ACTION) ? event.getEventName() : event.getEventName() + "_" + object.getClass().getSimpleName().toUpperCase();
 
         // We will create an anonymous user object for event tracking if no user is present
         // Without this, a lot of flows meant for anonymous users will error out
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/NewActionServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/NewActionServiceCEImpl.java
@@ -979,7 +979,15 @@ public class NewActionServiceCEImpl extends BaseService<NewActionRepository, New
             ActionExecutionResult actionExecutionResult,
             Long timeElapsed
     ) {
-
+        // Since we're loading the application from DB *only* for analytics, we check if analytics is
+        // active before making the call to DB.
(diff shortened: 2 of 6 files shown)


AnaghHegde:
@vishnu-gp Can you please add the view mode as well to all the analytics events?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCEImpl.java
```diff
@@ -164,7 +164,8 @@ public <T extends BaseDomain> Mono<T> sendObjectEvent(AnalyticsEvents event, T o
             return Mono.just(object);
         }
 
-        final String eventTag = event.getEventName() + "_" + object.getClass().getSimpleName().toUpperCase();
+        // In case of action execution, event.getEventName() only is used to support backward compatibility of event name
+        final String eventTag = event.equals(AnalyticsEvents.EXECUTE_ACTION) ? event.getEventName() : event.getEventName() + "_" + object.getClass().getSimpleName().toUpperCase();
```
mohanarpit, line 168:
```suggestion
        final String eventTag = AnalyticsEvents.EXECUTE_ACTION.equals(event) ? event.getEventName() : event.getEventName() + "_" + object.getClass().getSimpleName().toUpperCase();
```
Nit: In this particular case, it doesn't matter. But as a practice, always invoke the `equals` function on the static variable. Avoids NPEs.

vishnu-gp: Done

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/NewActionServiceCEImpl.java
```diff
@@ -980,12 +980,6 @@ private Mono<ActionExecutionRequest> sendExecuteAnalyticsEvent(
             Long timeElapsed
     ) {
 
-        // Since we're loading the application from DB *only* for analytics, we check if analytics is
-        // active before making the call to DB.
-        if (!analyticsService.isActive()) {
```
mohanarpit, line 985: Can you help me understand why we are removing this check? In CE code, we shouldn't be querying the DB if analytics is disabled.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/NewActionServiceCEImpl.java
```diff
@@ -1112,8 +1106,18 @@ private Mono<ActionExecutionRequest> sendExecuteAnalyticsEvent(
                                 "statusCode", actionExecutionResult.getStatusCode()
                         ));
                     }
-
-                    analyticsService.sendEvent(AnalyticsEvents.EXECUTE_ACTION.getEventName(), user.getUsername(), data);
+                    final Map<String, Object> auditData = Map.of(
```
mohanarpit, line 1109: Suggestion to rename this variable to anything but `auditData`. How about `eventData` ? :)

vishnu-gp: `eventData` might be taken in some places or create confusion. I will check the possible names other than `auditData`

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/NewActionServiceCEImpl.java
vishnu-gp: @mohanarpit Audit logging is done from the base of Analytics Events, If we don't remove this check, audit logs will not be logged for instances with telemetry disabled. Initially thought of sending an `analyticsEvent` from inside this if block to make it compatible but then decided this way to avoid future confusions.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/NewActionServiceCEImpl.java
vishnu-gp: How about `eventMetaData`? @mohanarpit

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/NewActionServiceCEImpl.java
mohanarpit: But in CE, we don't want to log audit information at all right?

vishnu-gp: No, we don't. But the data for those are being emitted from current analytics events which are present in the CE classes and giving a future scope if in case we want Audit Logs in CE. If removing this if() is not possible then I would have to override this method from EE just for this particular event.

Also, I think this DB querying part should have been handled by AnalyticsService where we already check `analyticsService.isActive()`.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/NewActionServiceCEImpl.java
vishnu-gp: Done

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/NewActionServiceCEImpl.java
mohanarpit: The DB querying portion cannot be handled by the `AnalyticsService` because that class is agnostic to the data that is being logged. Hence, we need to send complete data to that class.

vishnu-gp: @mohanarpit Changed as per the discussion and the changes were tested with telemetry on and off.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCEImpl.java
```diff
@@ -197,8 +198,8 @@ public <T extends BaseDomain> Mono<T> sendObjectEvent(AnalyticsEvents event, T o
                     analyticsProperties.put("oid", object.getId());
                     if (extraProperties != null) {
                         analyticsProperties.putAll(extraProperties);
-                        // To avoid sending extra audit data to analytics
-                        analyticsProperties.remove(FieldName.AUDIT_DATA);
+                        // To avoid sending extra event data to analytics
+                        analyticsProperties.remove(FieldName.EVENT_DATA);
```
mohanarpit, line 202: Sorry, I don't understand this line. Why are we adding the entire hashmap `extraProperties` to `analyticsProperties` and then removing `eventData`?

mohanarpit (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCEImpl.java
vishnu-gp: @mohanarpit This is to avoid audit-related extra data being sent to Analytics while still being able to access the data in AuditLogs by overriding this method.

vishnu-gp:
@allcontributors add @vishnu-gp as a contributor for code

PR #15783, opened 2022-08-06
feat: Enable caching functionality on appsmith-server module

trishaanand:


trishaanand:
Server workflow running at : https://github.com/appsmithorg/appsmith/actions/runs/2809732977

trishaanand:
Server workflow after fixing compilation issue : https://github.com/appsmithorg/appsmith/actions/runs/2809917503

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/BaseAppsmithRepositoryImpl.java
```diff
@@ -377,4 +337,38 @@ public Mono<T> findByGitSyncIdAndDefaultApplicationId(String defaultApplicationI
...
+                });
+    }
+
+    protected Mono<Set<String>> getAnonymousUserPermissionGroups() {
+        Criteria anonymousUserCriteria = Criteria.where(fieldName(QUser.user.email)).is(FieldName.ANONYMOUS_USER);
+
+        Query query = new Query();
+        query.addCriteria(anonymousUserCriteria);
+
+        return mongoOperations.findOne(query, User.class)
```
sidhantgoel, line 370: anonymous user can be cached too

app/server/reactive-caching/src/main/java/com/appsmith/caching/aspects/CacheAspect.java
```diff
@@ -139,7 +137,7 @@ private String deriveKey(String expression, String[] parameterNames, Object[] ar
      * @return Result of the method call, either cached or after calling the original method
      * @throws Throwable
      */
-    @Around("execution(public * *(..)) && @annotation(com.appsmith.caching.annotations.Cache)")
+    @Around("@annotation(com.appsmith.caching.annotations.Cache)")
```
sidhantgoel, line 140: Is this required?

app/server/reactive-caching/src/main/java/com/appsmith/caching/aspects/CacheAspect.java
```diff
@@ -175,7 +173,7 @@ public Object cacheable(ProceedingJoinPoint joinPoint) throws Throwable {
      * @return Mono<Void> that will complete after evicting the key from the cache
      * @throws Throwable
      */
-    @Around("execution(public * *(..)) && @annotation(com.appsmith.caching.annotations.CacheEvict)")
+    @Around("@annotation(com.appsmith.caching.annotations.CacheEvict)")
```
sidhantgoel, line 176: same

app/server/appsmith-server/src/main/java/com/appsmith/server/ServerApplication.java
```diff
@@ -3,10 +3,12 @@
 import org.springframework.boot.Banner;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.boot.builder.SpringApplicationBuilder;
+import org.springframework.context.annotation.ComponentScan;
 import org.springframework.scheduling.annotation.EnableScheduling;
 
 @SpringBootApplication
 @EnableScheduling
+@ComponentScan(basePackages = {"com.appsmith"})
```
sidhantgoel, line 11: not required now

app/server/appsmith-interfaces/pom.xml
```diff
@@ -134,6 +134,15 @@
             <artifactId>reactor-netty-core</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
```
sidhantgoel, line 137: This change is not required

app/server/reactive-caching/src/main/java/com/appsmith/caching/aspects/CacheAspect.java
trishaanand: No. Not required. This was before I realized that the caching method only works when called via the bean and not otherwise. So that means that the function (to be cached) would be public. I can remove this.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/BaseAppsmithRepositoryImpl.java
trishaanand: It automatically gets cached because the underlying function uses `this.getPermissionGroupsOfUser(anonymousUser)` as well.

sidhantgoel: Yeah so permission groups of anonymousUser gets cached, but we are reading anonymousUser everytime from DB. Also anonymousUser should also include tenantId criteria.

trishaanand: Makes sense. I will save this in memory so that we don't go fetching the anonymous user every time.

trishaanand: Done.

trishaanand:
Appsmith Server workflow : https://github.com/appsmithorg/appsmith/actions/runs/2819760673

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationServiceCEImpl.java
```diff
@@ -414,13 +415,19 @@ private Mono<PermissionGroup> createPublicPermissionGroup(Application applicatio
                 .findFirst()
                 .get();
 
+
         // Let this newly created permission group be assignable by everyone who has permission for make public application
         Policy assignPermissionGroup = Policy.builder()
                 .permission(ASSIGN_PERMISSION_GROUPS.getValue())
                 .permissionGroups(makePublicPolicy.getPermissionGroups())
                 .build();
+        // Let this newly created permission group also be manageable by everyone who has permission for make public application
```
sidhantgoel, line 424: are we creating auto generated groups MANAGEABLE by user?

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CacheableRepositoryHelperCEImpl.java
```diff
@@ -0,0 +1,102 @@
...
+
+    private String defaultTenantId;
+
+    public CacheableRepositoryHelperCEImpl(ReactiveMongoOperations mongoOperations) {
+        this.mongoOperations = mongoOperations;
+        this.defaultTenantId = null;
+        this.tenantAnonymousUserMap = new HashMap<>();
+    }
+
+    @Cache(cacheName = "permissionGroupsForUser", key="{#user.email + #user.tenantId}")
```
sidhantgoel, line 36: Can we user user.id as a key, it covers both email and tenantId and possibly remove tenantId argument from eviction function

trishaanand: This is easier to debug. Since tenant id gets cached in memory, its not a performance degradation on using the tenant as a parameter.

trishaanand:
Server workflow running at : https://github.com/appsmithorg/appsmith/actions/runs/2831766377

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationServiceCEImpl.java
trishaanand: Lets remove this permission assigning as part of your unassign permission on permission group PR

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CacheableRepositoryHelperCEImpl.java
sidhantgoel: problem is not tenantId, If we use email, more queries will be required to fetch emails of users to evict them, because at most of the places userId information is already available for eviction. and you can easily find userId by querying database for debugging.

trishaanand: Actually email is part of the session user object that we store in redis. That's why the previous ACL implementation was based on emails. We dont fetch the email. Sometimes the id is actually absent in the user session and where we must ahve the id, we do a repo fetch with email. This is a problem that we have faced. Happens rarely, but happens. But email is ALWAYS present.

trishaanand:
Got a go ahead to merge from @sidhantgoel. Merging.

PR #15815, opened 2022-08-08
fix: Remove unwanted logs which may confuse the users

### LINKED ISSUE #15811: [Bug]: Triggering action fields causes console errors for undefined style and content configs of the widget

### Description

When a widget's trigger field contains an action and this field is triggered then we see browser console errors for undefined style and content configs for that widget

### Steps To Reproduce

1. Add any widget, like Switch on the canvas
2. Add a showAlert action on its trigger field
3. Uncheck/check the switch widget and observe the browser console for errors as seen in the screenshot below

image

### Public Sample App

_No response_

### Version

Cloud

aswathkk:
## Description

Error messages related to widget configuration which are helpful only for Widget developers might cause confusion to our users.
So, removing them for a better UX.

Fixes #15811

## Type of change

## How Has This Been Tested?

- Manual

## Checklist:

### DIFF
--- a/app/client/src/utils/WidgetFactory.tsx
+++ b/app/client/src/utils/WidgetFactory.tsx
@@ -223,7 +223,6 @@ class WidgetFactory {
   ): readonly PropertyPaneConfig[] {
     const map = this.propertyPaneContentConfigsMap.get(type);
     if (!map) {
-      log.error("Widget property pane content configs not defined", type);
       return [];
     }
     return map;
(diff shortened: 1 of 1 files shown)


yaldram (approved):
Can we do this - 
```js
const map = this.propertyPaneContentConfigsMap.get(type) || [];
return map
```
Or we can directly return whichever is more readable.

ramsaptami:
Tested and verified on DP

aswathkk:
@yaldram I will consider improving this in a better manner when we remove the feature flag of new property pane updates.

PR #914, opened 2020-10-04
Fix: Do not allow for spaces inside create applications form.

jyash97:
### Current Behavior: 

We can submit an empty application in the Create application form, though BE gives an error it is better to do a client validation instead of waiting for API response to give an error.

### New Behavior:

https://www.loom.com/share/11db2d803aa94158a39e3ae23d3eb559

PR #15837, opened 2022-08-08
fix: Allow "spaces" in theme names while saving and display a more relevant error message 

### LINKED ISSUE #14840: [Bug]-[98]:Save Theme does not support 'space' to save as a theme name

### Description

Saving Theme does not support 'space'. The name with the space is not getting saved as a theme name

image

### Steps To Reproduce

1. Click on canvas to see theming properties
2. Click on 3 dots and click on 'save theme'
3. Try to save theme name with names which have spaces. for eg: My new theme
5. Observe that it does not support spaces as a name and throws error

### Public Sample App

_No response_

### Version

Cloud/Production

dhruvikn:
## Description
This PR intends to allow users to include "space(s)" in their theme name while creating/saving a new theme. Moreover, a new and more relevant error message is displayed which lets the user know what all is accepted in the theme name input.

> No spaces allowed previously and now
> No spaces allowed previously and now

> Old and new error messages
> Old and new error messages

Previously, while saving a new theme, we got an error if the name had any "space". However, Appsmith already has an in-house theme with a space in its name. 😄 

> "Water Lily" Theme
> CleanShot 2022-08-08 at 13 38 29@2x

Fixes #14840

## Type of change

## How Has This Been Tested?
- Manually Tested

> Few test cases
> Few test cases

## Checklist:


app/client/src/ce/constants/messages.ts
```diff
@@ -30,6 +30,8 @@ export const VALID_FUNCTION_NAME_ERROR = () =>
 export const UNIQUE_NAME_ERROR = () => `Name must be unique`;
 export const NAME_SPACE_ERROR = () => `Name must not have spaces`;
 export const SPECIAL_CHARACTER_ERROR = () => `Name must be alphanumeric`;
+export const APLHANUMERIC_HYPHEN_SLASH_SPACE_ERROR = () =>
+  `Name must only contain alphanumeric characters, hyphen, slash, and space`;
```
somangshu, line 34: @dilippitchika does the copy look ok to you?

cc @jnikhila

app/client/src/pages/Editor/ThemePropertyPane/SaveThemeModal.tsx
```diff
@@ -11,7 +11,7 @@ import { getAppThemes } from "selectors/appThemingSelectors";
 import {
   createMessage,
   ERROR_MESSAGE_NAME_EMPTY,
-  SPECIAL_CHARACTER_ERROR,
```
somangshu, line 14: @dhruvikn is this being used elsewhere? Why are we not changing the same?

somangshu (approved):
Apart from the comments, LGTM

app/client/src/pages/Editor/ThemePropertyPane/SaveThemeModal.tsx
dhruvikn: No, the SPECIAL_CHARACTER_ERROR is not being used anywhere else. We can remove it entirely. 
However, APLHANUMERIC_HYPHEN_SLASH_SPACE_ERROR seems a more suitable name for the message.

somangshu: Lets go for that

app/client/src/ce/constants/messages.ts
jnikhila: @somangshu - looks good to me.

jnikhila: @somangshu - On another note, why are there different error messages for different validations for the same field? Can we club them all to show a single error message? Adding detail and getting one after the other error message will frustrate users.

somangshu: @jnikhila the different validation messages are for different kinds of error, clubbing this mean the user will not have a pin pointed idea of what is doing wrong and also might have to jump into the debugger every time, AFAIK This change was made because the user did not have the a good idea about the error. I am not sure if in terms of UX this is a problem, @momcilo-appsmith @dilippitchika can comment on that better. 

@dhruvikn for the other suggested change, we should create a new issue with the relevant details and take it up later. Can you do this?
> Just one minor thing, if we can use something like "A name must begin with a letter and cannot start with a space" instead of "Name must not have spaces"- as it means that the name field cannot have spaces at all.

somangshu (dismissed):
LGTM

app/client/src/ce/constants/messages.ts
dhruvikn: Hi @jnikhila @somangshu. I think there is some confusion. Let me try to paint a clear picture. We have a file called messages.ts that has a lot of messages which are shown on particular conditions throughout Appsmith. 

However, for this Theme Name field, there are only three messages used. These three messages are shown on different occasions. The messages for the Theme Name field are - 
```
ERROR_MESSAGE_NAME_EMPTY = Please select a name
APLHANUMERIC_HYPHEN_SLASH_SPACE_ERROR = Name must only contain alphanumeric characters, hyphen, slash, and space
UNIQUE_NAME_ERROR = Name must be unique
```

Now, as for the `NAME_SPACE_ERROR = Name must not have spaces`, the author intended to say that "Name must not have spaces". Reference - https://github.com/appsmithorg/appsmith/pull/914

chandannkumar:

PR #15873, opened 2022-08-09
Changes for adding unassign permission on permission group

sidhantgoel:
I has given thou access to various assets, on mine own permission it can be taken away, but I give thou the power to abandon them


sidhantgoel:
Workflow
https://github.com/appsmithorg/appsmith/actions/runs/2826087509

app/server/appsmith-server/src/main/java/com/appsmith/server/acl/ce/PolicyGeneratorCE.java
```diff
@@ -166,6 +166,7 @@ private void createThemePolicyGraph() {
 
     private void createPermissionGroupPolicyGraph() {
         lateralGraph.addEdge(AclPermission.MANAGE_PERMISSION_GROUPS, AclPermission.ASSIGN_PERMISSION_GROUPS);
+        lateralGraph.addEdge(AclPermission.MANAGE_PERMISSION_GROUPS, AclPermission.UNASSIGN_PERMISSION_GROUPS);
```
trishaanand, line 169: Hey, no one should have manage permission on CE. We should add a lateral edge from assign to unassign. And remove the usage of manage permission groups (or the permission itself) from our code base.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
```diff
@@ -1517,19 +1517,17 @@ private String getDefaultNameForGroupInWorkspace(String prefix, String workspace
...
         adminPermissionGroup.setDefaultWorkspaceId(workspaceId);
         adminPermissionGroup.setTenantId(workspace.getTenantId());
         adminPermissionGroup.setDescription(FieldName.WORKSPACE_ADMINISTRATOR_DESCRIPTION);
-        Set<Permission> workspacePermissions = AppsmithRole.ORGANIZATION_ADMIN
-                .getPermissions()
-                .stream()
-                .filter(aclPermission -> aclPermission.getEntity().equals(Workspace.class))
-                .map(aclPermission -> new Permission(workspace.getId(), aclPermission))
-                .collect(Collectors.toSet());
-        adminPermissionGroup.setPermissions(workspacePermissions);
```
trishaanand, line 1532: You have overridden storing of permissions in the permission group. Could we bring it back at least till we do a PoC on doing things without these permissions?

app/server/appsmith-server/src/main/java/com/appsmith/server/acl/ce/PolicyGeneratorCE.java
sidhantgoel: no one has manage permission, I've just added the link because it is logical, but we are not giving manage permission.
About unassign from assign I didn't do it because not everyone who can invite can remove, for example developer can invite developers but they can't remove them only admins can remove, so I'm giving unassign to admin permission group and assign as before. manage we are not giving.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
sidhantgoel: I've removed this here because it is already happening in next function, we are keeping both Permissions and sing them to build Policies for now.

sidhantgoel: I just removed this because it was happening twice, once here and then again in next function. We are still storing permissions and using them to build policies.

PR #15881, opened 2022-08-10
fix: Make default DB compulsory in URI string format

### LINKED ISSUE #12713: [Bug]-[210]:Mongo plugin URI parsing issue

### Description

Ref: https://theappsmith.slack.com/archives/C0341RERY4R/p1649405309648129
Cannot share the URI string publicly here for security reasons. 

### Steps To Reproduce

https://theappsmith.slack.com/archives/C0341RERY4R/p1649405309648129 
Cannot share the URI string publicly here for security reasons. 

Workaround: Use the form based datasource config page.

### Public Sample App

_No response_

### Version

cloud

nidhi-nair:
## Description

In the form mode of configuring a Mongo datasource, we have made the default DB a compulsory field so we know which DB to run queries against. This PR adds the same validation for the URI string format.

On running queries:
Screenshot 2022-08-10 at 11 19 22 AM

On testing datasource:
Screenshot 2022-08-10 at 11 16 33 AM

Fixes #12713 

## Type of change

## How Has This Been Tested?
- JUnit Test
- Manual test

## Checklist:


app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/utils/MongoPluginUtils.java
```diff
@@ -107,12 +107,16 @@ private static MongoCommand getMongoCommand(ActionConfiguration actionConfigurat
...
-        String databaseName = datasourceConfiguration.getConnection().getDefaultDatabaseName();
+        if (datasourceConfiguration.getConnection() != null) {
+            databaseName = datasourceConfiguration.getConnection().getDefaultDatabaseName();
+        }
 
         // If that's not available, pick the authentication database.
         final DBAuth authentication = (DBAuth) datasourceConfiguration.getAuthentication();
-        if (StringUtils.isEmpty(databaseName) && authentication != null) {
+        if (!StringUtils.hasLength(databaseName) && authentication != null) {
             databaseName = authentication.getDatabaseName();
```
subrata71, line 120: Hey @nidhi-nair 
What if the `authentication.getDatabaseName()` also returns `null`? Won't the users, who fall into this case, see the same error like "name can not be null"?

nidhi-nair: The validations before this step should have taken care of this case @subrata71 . This PR ensures that we will always have a default db name. In fact, we could even remove this condition without expecting trouble, but I'll leave that for the future, just so we're able to discover any other loopholes in the validations.

subrata71: But doesn't it have an effect on the existing datasource configuration with missing db name? @nidhi-nair

nidhi-nair: Hmm .. Good point. We only do validations during save and test, so executions would not discover this for existing datasources. I'll throw an error redirecting them to reconfigure the datasource here as well. Nice catch!

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/utils/MongoPluginUtils.java
```diff
@@ -107,15 +107,23 @@ private static MongoCommand getMongoCommand(ActionConfiguration actionConfigurat
...
         }
 
+        if (databaseName == null) {
+            throw new AppsmithPluginException(AppsmithPluginError.PLUGIN_DATASOURCE_ARGUMENT_ERROR, "Missing default database name.");
+        }
```
subrata71, line 125: I wonder if someday we refactor these error messages and forget to do the same in the test cases. I understand this should not be part of this PR but I can forsee that it might happen. @nidhi-nair

subrata71 (approved):
The change looks good to me. I hope this will help this issue from ever happening again. 🙏

subrata71 (approved):

PR #15883, opened 2022-08-10
Cleanup on PolicyUtils

sidhantgoel:
1. Removed outdated functions that uses users and commented the code using them, that was mostly from Comment feature.
2. Made functions to return and use Set of Policy as it is more appropriate return type, that also removes unnecessary creation of Map and Map if required can be created by the calling function.
3. Removed deepcopy and optimized the logic for adding and removing Policies to the object.


sidhantgoel:
Server Workflow
https://github.com/appsmithorg/appsmith/actions/runs/2830398862

PR #15954, opened 2022-08-12
feat: Table widget V2 Inline editing validation for plain text & number column

### LINKED ISSUE #13670: [Feature]: Validations for inline editing in input widget

### Summary

Add validations to the input widget within the table

Completes the input inline editing flow

sbalaji1192:
## Description

Now editable column has the option to validate the input text before saving.

Fixes #13670

## Type of change

## How Has This Been Tested?
Cypress test cases were added to verify the feature

Table widget inline editing validation functionality
  1. should check that validation only appears when editable enabled
  2. should check that validation only appears for plain text and number
  3. should check that regex, valid & required appear for the plain text column
  4. should check that min, max, regex, valid & required appear for the number column
  5. should check the validation property for the plain text column
     a. Regex
     b. Valid
     c. Required
  6. should check validation property for number column
     a. Min
     b. Max
  7. should check the error message property
  8. should check the editable cell actions when there is a validation error
     a. save should only work when there is no error
     b. discard should only work when there is no error
  9. should check that the save/discard button is disabled when there is a validation error

## Checklist:


chandannkumar:
Observation:
1. Regex validation is not working for Number column type


2. Valid field is not working for Number type column
https://www.loom.com/share/aa6a2da4ae004ab1923b645b4caa2d9f

3. Error on the Table cell can seen outside the canvas as well
https://www.loom.com/share/cebd81a7029d46579472d4137fd285c7

4. There is no auto-suggestion for configuring valid field on Number or Plain text column type.
Like using **{{editedValue.length < 10}}**  no auto-suggestion was suggesting these

app/client/src/widgets/TableWidgetV2/widget/index.tsx
```diff
@@ -1093,6 +1098,20 @@ class TableWidgetV2 extends BaseWidget<TableWidgetProps, WidgetState> {
     return "TABLE_WIDGET_V2";
   }
 
+  getColumnByAlias(alias: string) {
+    const { primaryColumns } = this.props;
+
+    if (primaryColumns) {
+      const column = Object.values(primaryColumns).find(
+        (column) => column.alias === alias,
```
ashit-rath, line 1106: `alias` can have non-ascii values right?
If so then can we do `column[alias]`?

ashit-rath: Also is the method `getColumnByAlias` being used anywhere? Can't seem to find it

app/client/src/widgets/TableWidgetV2/widget/index.tsx
```diff
@@ -1519,10 +1542,11 @@ class TableWidgetV2 extends BaseWidget<TableWidgetProps, WidgetState> {
             columnType={column.columnType}
             compactMode={compactMode}
             disabledEditIcon={
-              this.props.inlineEditingSaveOption ===
+              (this.props.inlineEditingSaveOption ===
```
ashit-rath, line 1545: Can we put this condition in a variable? it feels quite long to be part of component render.

app/client/src/widgets/TableWidgetV2/widget/index.tsx
```diff
@@ -543,7 +546,8 @@ class TableWidgetV2 extends BaseWidget<TableWidgetProps, WidgetState> {
      */
     if (isTableDataModified) {
       this.props.updateWidgetMetaProperty("transientTableData", {});
-      this.props.updateWidgetMetaProperty("editableCell", {});
+      this.props.updateWidgetMetaProperty("editableCell", defaultEditableCell);
```
ashit-rath, line 549: Can we call `clearEditableCell` here?

app/client/src/widgets/TableWidgetV2/widget/propertyConfig/PanelConfig/Validations/Common.ts
```diff
@@ -0,0 +1,52 @@
+import { ValidationTypes } from "constants/WidgetValidation";
+
+export default [
+  {
+    helpText:
+      "Adds a validation to the cell value which displays an error on failure",
+    propertyName: "validation.regex",
```
ashit-rath, line 7: Can we have `propertyName` as the first key of the property objects? It helps finding the code faster.

ashit-rath (commented):
Couple of things I noticed while trying out the DP

- The border here looks a bit weird to me as if the right and top border is thinner than the left and bottom ones.  Can you verify once?
Screenshot 2022-09-07 at 10 06 25 AM

- The validation sort of didn't work for me, is my binding incorrect here?
Screenshot 2022-09-07 at 10 12 48 AM

app/client/src/widgets/TableWidgetV2/widget/index.tsx
sbalaji1192: Stale code removed it.

app/client/src/widgets/TableWidgetV2/widget/derived.js
```diff
@@ -587,4 +587,88 @@ export default {
...
+
+    if (editedColumn && editedColumn.validation) {
+      const validation = editedColumn.validation;
+
+      /* General validations */
+      if (!validation.isColumnEditableCellRequired && value === "") {
+        return true;
+      } else if (
+        (!_.isNil(validation.isColumnEditableCellValid) &&
+          !validation.isColumnEditableCellValid) ||
```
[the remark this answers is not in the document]
sbalaji1192: We are making sure that isColumnEditableCellValid is not null or undefined before using that in negation.

app/client/src/widgets/TableWidgetV2/constants.ts
```diff
@@ -78,14 +79,15 @@ export interface TableWidgetProps extends WidgetProps, WithMeta, TableStyles {
...
+  isEditableCellValid: boolean;
 }
 
 export const getCurrentRowBinding = (
   entityName: string,
   userInput: string,
   withBinding = true,
 ) => {
-  let rowBinding = `${entityName}.sanatizedTableData.map((currentRow) => ( ${userInput}))`;
+  let rowBinding = `${entityName}.processedTableData.map((currentRow, currentIndex) => ( ${userInput}))`;
```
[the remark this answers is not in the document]
sbalaji1192: it's only used in the test. Let me move this to the test file.

app/client/src/widgets/TableWidgetV2/component/cellComponents/InlineCellEditor.tsx
```diff
@@ -122,9 +131,26 @@ export function InlineCellEditor({
   );
 
   const onTextChange = useCallback(
-    (data: string) => {
+    (inputValue: string) => {
```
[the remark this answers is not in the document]
sbalaji1192: Here the value will be coming from the input component which will always be a string.

app/client/src/components/editorComponents/ErrorTooltip.tsx
```diff
@@ -38,9 +38,31 @@ interface Props {
   message: string;
   children: JSX.Element;
   customClass?: string;
+  preventOverflow?: string;
```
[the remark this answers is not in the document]
sbalaji1192: This value is gettings passed to a component within a compoenet. So I guess we stick with the selector. If in the future we conditionally render the tooltip. we can avoid the unnecessary querySelector call if we let the errortooltip call it. Thoughs?

sbalaji1192:
@ashit-rath Regarding your two points
1. I just zoomed in and had a closer look. the border seems to be fine. Can you try once as well?
2. Binding is incorrect you should use `editedValue` instead `currentRow.step`. while entering value currentRow.step value points to older value.

app/client/src/components/editorComponents/ErrorTooltip.tsx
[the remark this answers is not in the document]
ashit-rath: Okay fair enough.

Can we modify the prop name to something that represent we need a selector here? Right now `preventOverflow` doesn't feel very intuitive.

laveena-en:
A couple of observations here:

Can we have the error toast to read "This field is required" instead of invalid text when a field is set to required. That would be consistent with the other widgets and differentiate from an actual invalid input.



Invalid values are being allowed even when the valid condition is set https://www.loom.com/share/26d45749c5c4467f90be537d70634191

The error text is seen on the canvas on re-sizing https://www.loom.com/share/4b607ba78f6f424e9b73552cec075e4c

ashit-rath (approved):

sbalaji1192:
@laveena-en 
#1 - fixed
#2 - not reproducible
#3 - not an issue. checked with @dilippitchika . we can ignore this.

laveena-en:
LGTM!

ashit-rath (approved):

PR #15965, opened 2022-08-12
chore: integrate msw to browser to mock api requests

### LINKED ISSUE #14395: [Task]: mock API calls/response for RBAC

### SubTasks

> As of now we are just consuming the mock data directly by keeping it in a variable and just setting it to the state on load, we are not getting it async in an effect. So if we could mock the process using a mock response data in a json file, once the API is done we can just replace the path to json file with path to url endpoint.

- mock API calls
- mock response data for RBAC 
- create sagas and actions

berzerkeer:
## Description

As of now, when the backend api's are not ready we mock api data by returning a JSON in promise or just directly using it. There is no way to  mock the `serve` part unless you put the json files in a public folder and fetch using axios or fetch. So `msw` is made to solve that, we are already using it for our test env and now I am adding it to browser env.  We only use it in `development` and the `serviceworker` also only gets registered in the dev environment. 

Fixes #14395

## Type of change

## How Has This Been Tested?

- Manual

## Checklist:


app/client/public/mockServiceWorker.js
```diff
@@ -0,0 +1,322 @@
...
+    }
+  }
+})
+
+// Resolve the "master" client for the given event.
+// Client that issues a request doesn't necessarily equal the client
+// that registered the worker. It's with the latter the worker should
+// communicate with during the response resolving phase.
+async function resolveMasterClient(event) {
+  const client = await self.clients.get(event.clientId)
```
IAmAnubhavSaini, line 90: If you are awaiting, then you need to handle for error via try/catch.

app/client/public/mockServiceWorker.js
```diff
@@ -0,0 +1,322 @@
...
+  const client = await resolveMasterClient(event)
+  const response = await getResponse(event, client, requestId)
+
+  // Send back the response clone for the "response:*" life-cycle events.
+  // Ensure MSW is active and ready to handle the message, otherwise
+  // this message will pend indefinitely.
+  if (activeClientIds.has(client.id)) {
+    ;(async function () {
+      const clonedResponse = response.clone()
+      sendToClient(client, {
```
IAmAnubhavSaini, line 120: 1. You are returning `Promise` from that function. Need `await` here.

app/client/public/mockServiceWorker.js
```diff
@@ -0,0 +1,322 @@
...
+
+async function handleRequest(event, requestId) {
+  const client = await resolveMasterClient(event)
+  const response = await getResponse(event, client, requestId)
+
+  // Send back the response clone for the "response:*" life-cycle events.
+  // Ensure MSW is active and ready to handle the message, otherwise
+  // this message will pend indefinitely.
+  if (activeClientIds.has(client.id)) {
+    ;(async function () {
```
IAmAnubhavSaini, line 118: 2. This doesn't need `;`.
3. Need await here. An async function, even IIAFE, should be awaited.

app/client/public/mockServiceWorker.js
```diff
@@ -0,0 +1,322 @@
...
+function respondWithMock(clientMessage) {
+  return new Response(clientMessage.payload.body, {
+    ...clientMessage.payload,
+    headers: clientMessage.payload.headers,
+  })
+}
+
+function uuidv4() {
+  return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, function (c) {
+    const r = (Math.random() * 16) | 0
```
IAmAnubhavSaini, line 318: Use crypto API.

https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API

IAmAnubhavSaini: https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID

IAmAnubhavSaini (changes requested):

berzerkeer:
@IAmAnubhavSaini the file `mockServiceWorker.js` is auto generated by msw. I didn't create it.

IAmAnubhavSaini (dismissed):

ankitakinger (approved):

PR #16020, opened 2022-08-16
feat: use Depot for Docker builds in CI

kylegalbraith:
(context for this is an email exchange with @mohanarpit)

### Problem/background

Faster Docker builds!

When building Docker images in GitHub Actions, the lack of a native Docker build cache means builds can be quite slow, and even options like `cache-to` / `cache-from` can eat into build time since they must serialize / deserialize the cache to tarballs which comes at a performance cost. And Actions currently only offers hosted Intel environments, if you need to build Arm images, you need to use emulation/QEMU, which can be slow.

[Depot](https://depot.dev/) is a hosted Docker build service that we've created to make this faster - we spawn managed VMs running BuildKit on-demand to handle container builds. Those VMs have 4 CPUs (2 more than Actions today), 8GB of memory, and a persistent 50GB local SSD disk. This allows all of BuildKit's native cache features to just work and persist across builds. Our builders start in just a few seconds with the SSD disk already ready, no need to load the cache before the build can start.

We have a CLI, `depot build`, that embeds a copy of `docker buildx` build that routes builds to our infrastructure. Since it's using Buildx as a library, it's typically just a matter of swapping docker build for depot build to send builds to the remote builders.

@mohanarpit signed up for Depot and we chatted over email about how we could drastically speed things up over here. Arpit suggested we open a PR so we can see the speed up with our own eyes.

### Changes

To use Depot here, I've made the following changes:

1. I have added a `permissions` block to the jobs building images. This block allows the `depot/build-push-action` to authenticate to an Appsmith Depot project via [GitHub's OIDC token exchange](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect). There is a trust relationship on the Appsmith project in Depot that allows OIDC tokens from this org + repository to be exchanged for temporary tokens to run the image build on our builders. This allows you to avoid static tokens and get a temporary token that is only good for the one build it was assigned 🎉 

2. A new action, `depot/setup-action`, configures the runner with the `depot` CLI that will be used to execute an image build in Depot. This is a prerequisite for the `depot/build-push-action`. 

3. The action, `depot/build-push-action`, works exactly like the `docker/build-push-action` but authenticates the image builds to Depot and sends the build context to our remote builders for faster builds with persistent caching.

I'll add a few comments below of minor details.

The TL;DR of how this all works:

- Previously, in places you were making use of `docker build` and `docker push` to build an image with a specific tag and then pushing it to a registry.
- Now, `depot/setup-action` installs the [Depot CLI](https://github.com/depot/cli), which embeds the `docker buildx build` client to tunnel to a remote builder VM that's launched on demand, with a persistent SSD cache. The `depot/build-push-action` uses this action to launch a builder and build an image via the parameters passed into the action (`tags`, `build_args`, `platforms`, etc). It builds the image based on these parameters and can then push the image to registries or load it locally.
- Finally, with Depot, native Arm and Intel images are supported out of the box, so there is no need to configure Buildx or QEMU emulation where you need multi-platform images. Simply pass the `platforms` in the `depot/build-push-action` and we build the image for both architectures in parallel. 

If you have any questions, feel free to ping me (@kylegalbraith) or Jacob (@jacobwgillespie), we'd be happy to help in any way!


depot.json
```diff
@@ -0,0 +1 @@
+{"id":"ts81441f91"}
```
kylegalbraith, line 1: This is the `appsmith` project ID in Depot. This file is used by the `depot` CLI in CI and locally to know which project to route the Docker build too. Currently, all Dockerfiles in this project are using the same project.

.github/workflows/test-build-docker-image.yml
```diff
@@ -533,9 +540,13 @@ jobs:
...
       - name: Build  docker image
         if: steps.run_result.outputs.run_result != 'success'
-        working-directory: "."
-        run: |
-          docker build -t fatcontainer .
+        uses: depot/build-push-action@v1
+        with:
+          context: .
+          push: false
+          load: true
```
kylegalbraith, line 547: By default `depot build` doesn't return the built image to the caller. We do this because typically in CI you want to build and push the image, so returning the built image to the caller is a waste of network bandwidth. This can be changed by passing the `--load` flag which is what this is doing here as the image is ran below.

.github/workflows/github-release.yml
```diff
@@ -218,66 +224,94 @@ jobs:
...
+        uses: depot/build-push-action@v1
+        with:
+          context: app/rts
+          push: true
+          tags: |
+            ${{ secrets.DOCKER_HUB_ORGANIZATION }}/appsmith-rts:${{needs.prelude.outputs.tag}}
+
+      # Only build & tag with latest if the tag doesn't contain beta
+      - name: Build and push RTS image latest
+        if: '!contains(needs.prelude.outputs.tag, "beta")'
```
kylegalbraith, line 308: This is GitHub Actions equivalent of what the `if` condition was checking via the generic action step. Only build and push the RTS latest image if this is not a beta tag.

mohanarpit:
@kylegalbraith Thank you so much for opening a PR to improve the build speed for the Appsmith project.

kylegalbraith:
> @kylegalbraith Thank you so much for opening a PR to improve the build speed for the Appsmith project.

Awesome, let us know if we can do anything to help get this merged in. It looks like the UI tests are what is failing which may be unrelated to this change.

kylegalbraith:
This was handled in https://github.com/appsmithorg/appsmith/pull/16658.

PR #16021, opened 2022-08-16
feat:setup pseudo canvas widget reducer for evaluation

### LINKED ISSUE #15779: List V2 POC: Setup pseudoCanvasWidgetReducer for evaluation and render pseudo widgets
## Setup pseudoWidgetReducer to evaluation

Derived properties are evaluated in the evaluation engine during an evaluation cycle. For normal widgets, canvas widget properties and meta properties are injected into the `uneval tree` before sending it to the evaluation engine. The `this` present in the derived properties is replaced with the widget’s name and that reference is used during the evaluation.

The pseudo widget needs to be present in the `uneval tree` and for it to be present in the    `uneval tree`, it has to be present in the canvasWidget reducer.

We need to introduce a new reducer called `pseudoCanvasWidgetReducer` reducer which would hold the pseudo widget’s props. We shouldn't touch the existing canvasWidget reducer for this as the operations related with the new reducer would be different from canvasWidgetReducer so it should be kept isolated.

During the creation of the `uneval tree` ; along with canvasWidgets, the pseudoCanvasWidgets has to be picked up as a part of the `uneval tree` along with its derived props (which evaluation engine would do by default) and has to be processed like normal evaluation.

## Render pseudo widgets
The pseudo widgets would be rendered using the same logic of the new widget rendering flow [#14485](https://github.com/appsmithorg/appsmith/pull/14485) i.e the List widget would receive pseudo widget child structure and the `withWidgetProps` would subscribe to it's props in the pseudoWidgetReducer

Tooluloope:
- [x]  introduce a new reducer called `pseudoCanvasWidgetReducer`
- [x] `pseudoCanvaswidget` needs to be present in the `uneval tree`
- [x] Render pseudo widgets

## Description

Fixes #15779 

## Type of change

## How Has This Been Tested?

- Test A
- Test B

## Checklist:


app/client/src/actions/metaWidgetActions.ts
```diff
@@ -0,0 +1,27 @@
+import { ReduxActionTypes } from "@appsmith/constants/ReduxActionConstants";
+import { ModifyMetaWidgetPayload } from "reducers/entityReducers/metaCanvasWidgetsReducer";
+import { FlattenedWidgetProps } from "widgets/constants";
+
+export const addMetaWidget = (
```
SatishGandham, line 5: I think we only need to have updateMetaWidget. Can we clean up the code after checking with @ashit-rath .
Shouldn't this be plural?`updateMetaWidgets`

app/client/src/entities/DataTree/dataTreeFactory.ts
```diff
@@ -148,6 +149,7 @@ type DataTreeSeed = {
   appData: AppDataState;
   jsActions: JSCollectionDataState;
   theme: AppTheme["properties"];
+  metaCanvasWidgets: MetaCanvasWidgetsReduxState;
```
SatishGandham, line 152: How do we treat them differently compared to regular widgets?
Can we combine these with them

app/client/src/entities/DataTree/dataTreeFactory.ts
```diff
@@ -203,13 +206,25 @@ export class DataTreeFactory {
...
+    const endMetaCanvasWidgets = performance.now();
+
     const end = performance.now();
 
     const out = {
       total: end - start,
       widgets: endWidgets - startWidgets,
       actions: endActions - startActions,
       jsActions: endJsActions - startJsActions,
+      metaCanvasWidgets: endMetaCanvasWidgets - startMetaCanvasWidgets,
```
SatishGandham, line 227: Can you share the results of this?

app/client/src/sagas/selectors.tsx
```diff
@@ -14,11 +14,18 @@ import { ActionData } from "reducers/entityReducers/actionsReducer";
...
+import { MetaCanvasWidgetsReduxState } from "reducers/entityReducers/metaCanvasWidgetsReducer";
 
 export const getWidgets = (state: AppState): CanvasWidgetsReduxState => {
   return state.entities.canvasWidgets;
 };
 
+export const getMetaCanvasWidgets = (
+  state: AppState,
+): MetaCanvasWidgetsReduxState => {
+  return state.entities.metaCanvasWidgets;
```
SatishGandham, line 26: @ashit-rath Have we considered treating them as regular widgets? Is it possible?

SatishGandham (changes requested):

app/client/src/sagas/selectors.tsx
ashit-rath: @SatishGandham Do you mean not to introduce the concept of `metaWidgets` and just add to the canvasWidgetsReducer like regular widgets?

SatishGandham: We may call them that, and might be needed to exclude from update layout,  but does the worker/evaluations need to know about it?

ashit-rath: The main reason to introduce this new reducer was to avoid the mixup in the canvas reducer. 
There are a lot of components that subscribe to this reducer and list widget can add 100s of widgets to the reducer so every subscription has to be carefully done in order to not include the metaWidgets.

Mixing also sort of gives an access to add/remove the canvasWidgets using the apis dynamically, though we can restrict this but all of these extra logic may create problems.

Also the first row item widgets are also meta widgets with the actual widget's widgetName and widgetId. The reason why we do it meta, is to make sure the properties where currentItem is used gets evaluated according to the index. We cannot alter the property values of the actual canvasWidget for the first row item the way we do it for the metaWidgets.

ashit-rath:
@Tooluloope Can you modify the base branch of this PR to the one that you merged? It would be easier to review if we can see only the changes you made.

app/client/src/widgets/withWidgetProps.tsx
```diff
@@ -91,10 +98,13 @@ function withWidgetProps(WrappedWidget: typeof BaseWidget) {
           return computeMainContainerWidget(canvasWidget, mainCanvasProps);
         }
 
-        if (isMetaCanvasWidget) return metaCanvasWidget;
+        //if (isMetaCanvasWidget) return metaCanvasWidget;
```
ashit-rath, line 101: Let's remove this line

ashit-rath (approved):
LGTM!

ashit-rath:
@Tooluloope Let's not merge this PR right now.
There are some performance profiling we are doing on the other POCs and until we do that we want to hold off on merging this.

app/client/src/pages/Editor/PropertyPane/PropertyControl.tsx
```diff
@@ -305,7 +305,7 @@ const PropertyControl = memo((props: Props) => {
     // is changed on propertypane. For e.g - set/update parent property
     if (childWidgetPropertyUpdateEnhancementFn) {
       const hookPropertiesUpdates = childWidgetPropertyUpdateEnhancementFn(
-        widgetProperties.widgetName,
+        widgetProperties,
```
SatishGandham, line 308: We are changing the value type here. Is this intentional?

ashit-rath: @SatishGandham Yes this is intentional, `childWidgetPropertyUpdateEnhancementFn` was always called with the `widgetName` only. This worked for List v1 but in v2 we needed the `widgetId` instead.

So instead of add another argument to the `childWidgetPropertyUpdateEnhancementFn` function, we decided to make the whole props available.

@Tooluloope Have we made sure this works in the List v1?

SatishGandham (approved):

app/client/src/pages/Editor/PropertyPane/PropertyControl.tsx
Tooluloope: Yes, this works in V1.
Using `widgetProperties. widgetName` in  List widget's `propertyUpdateHook`

PR #16073, opened 2022-08-16
fix: search functionality in tree select

### LINKED ISSUE #15119: [Bug]-[440]:Dropdown search behaviour is incorrect when search field does not match listed options

### Description

When a search text does not match dropdown options on a tree-select/multi-tree select widget then the options tray should not open up and display only matched options when there is a match. Behvaiour is expected to be in line with select/multi-select widget. 

### Steps To Reproduce

1. Drop a tree & mutli-tree select widgets onto the canvas.
 2. Use the filter property to search for values present in the options tray - all options are displayed with the matched option appearing in bold.
 3. Search for values not present in the options tray - all options are displayed even though there aren't any matches

https://www.loom.com/share/da799fb76265438db4f9651e080d6d88

### Public Sample App

_No response_

### Version

Cloud

techbhavin:
## Description

> Show not found text when the search field does not match listed options

Fixes #15119 

## Type of change

## Checklist:

### DIFF
--- a/app/client/package.json
+++ b/app/client/package.json
@@ -92,8 +92,8 @@
     "prismjs": "^1.27.0",
     "punycode": "^2.1.1",
     "rc-pagination": "^3.1.3",
-    "rc-select": "13.2.1",
-    "rc-tree-select": "^4.4.0-alpha.2",
+    "rc-select": "^14.1.9",
+    "rc-tree-select": "^5.4.0",
     "re-reselect": "^3.4.0",
     "react": "^16.12.0",
     "react-base-table": "^1.9.1",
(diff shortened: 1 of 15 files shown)


app/client/package.json
```diff
@@ -92,8 +92,8 @@
     "prismjs": "^1.27.0",
     "punycode": "^2.1.1",
     "rc-pagination": "^3.1.3",
-    "rc-select": "13.2.1",
-    "rc-tree-select": "^4.4.0-alpha.2",
+    "rc-select": "^14.1.9",
```
keyurparalkar, line 95: @techbhavin Is this an expected change?

keyurparalkar (commented):
@techbhavin Can you please write tests for this bug?

app/client/package.json
techbhavin: @keyurparalkar issue is happening with the older version in RC-select so we update the package version. 

Here All changes are related to type change so do you think here test case is needed?

keyurparalkar: @techbhavin Yes, since it seems this is going to be the default behaviour during searching inside tree select widget

keyurparalkar (approved):
LGTM!

laveena-en:
Tested and verified.

PR #16076, opened 2022-08-16
fix: old column names exists even after renamed columns

### LINKED ISSUE #16011: [Bug]: Old column names continue to exist even after renaming the column in Google Sheet

### Description
GS-Bug-Image1

I renamed a column in an existing data source (google sheets) and the API correctly picks up the name of the column, but when I try to run it, it shows it can not locate the old Column and ends up in an error.

Old Name : Machine List
New Name : MachineList

Edit: Things got a lot weirder, When I try to create a completely different API to get data from a different sheet, I’m getting the same “Machine List” not found error - it was never there, to begin with. I don't know what's going on anymore.

GS-Bug-Image2

### Source
This issue has been reported first in the community.appsmith.com forum.
https://community.appsmith.com/t/api-referring-to-old-column-names-after-changing-them-and-giving-error-when-loading-data/1295

### Steps To Reproduce

1. Create a Google Sheet (GS) datasource
2. Save and authorize GS datasource
3. Create an API on this GS datasource with operation `Fetch Many`
4. Select one or more columns from the `Columns` drop-down list
5. Now go to your Google Sheet and rename one column
6. Go back to the API that you created in step 3
7. You will see the new column names appearing under the `Columns` drop-down list
8. Now select one or more columns from the `Columns` drop-down list
9. You will see an error like
    Filtering failure seen:  Column "< Old column name >" not found

### Public Sample App

_No response_

### Version

Cloud / Self Hosted

AmanAgarwal041:
## Description

Added the condition to remove older column names and update the projection values to the existing columns only.
eg : 
 - *projection value* : key - `entities.actions[actionIndex].config.actionConfiguration.formData.projection.data`  , value - `['a', 'b', 'c']`
 - *trigger value* : key - `evaluations.triggers[actionId].rows_get_projection.fetchDynamicValues.data`, value - `[{ label: 'A', value: 'a' }, { label: 'B', value: 'b' }, { label: 'D', value: 'd' }]`

`selectedOptions` are created by filtering all options with only option value present in selectedValue.

- *options*: `[{ label: 'A', value: 'a' }, { label: 'B', value: 'b' }, { label: 'D', value: 'd' }]`
- *selectedOptions* : `[{ label: 'A', value: 'a' }, { label: 'B', value: 'b' }]`
- *selectedValue*: `['a', 'b', 'c']`
- *tempSelectedValue*: `['a', 'b']`  - is updated in redux form field's state

Fix adds a condition to check if the selected values for a dropdown (i.e. *projection value*) has values that are not present in the selected options then filter out those extra values and set filtered values to selectedValue if the previous `selectedValue` length is not equal to `tempSelectedValue` length.

Fixes #16011

## Type of change

## How Has This Been Tested?

- Test A
- Test B

## Checklist:

### DIFF
--- a/app/client/src/components/formControls/DropDownControl.tsx
+++ b/app/client/src/components/formControls/DropDownControl.tsx
@@ -181,6 +181,33 @@ function renderDropdown(
     }
   };
 
+  if (props.options.length > 0) {
+    if (props.isMultiSelect) {
+      const tempSelectedValues: string[] = [];
+      selectedOptions.forEach((option: DropdownOption) => {
+        if (selectedValue.includes(option.value as string)) {
+          tempSelectedValues.push(option.value as string);
+        }
+      });
+      if (tempSelectedValues.length !== selectedValue.length) {
+        selectedValue = [...tempSelectedValues];
+        props.input?.onChange(tempSelectedValues);
+      }
+    } else {
+      let tempSelectedValues = "";
+      selectedOptions.forEach((option: DropdownOption) => {
+        if (selectedValue === (option.value as string)) {
+          tempSelectedValues = option.value as string;
+        }
+      });
+
+      if (selectedValue !== tempSelectedValues) {
+        selectedValue = tempSelectedValues;
+        props.input?.onChange(tempSelectedValues);
+      }
+    }
+  }
+
   return (
     <Dropdown
       allowDeselection={props?.isMultiSelect}


app/client/src/components/formControls/DropDownControl.tsx
```diff
@@ -181,6 +181,33 @@ function renderDropdown(
...
+        }
+      });
+
+      if (selectedValue !== tempSelectedValues) {
+        selectedValue = tempSelectedValues;
+        props.input?.onChange(tempSelectedValues);
+      }
+    }
+  }
+
```
ChandanBalajiBP, line 210: Can we move this code to a new function and call it? Both if and else block contains similar code..

ChandanBalajiBP (commented):
Can we move this code to a new function and call it? Both if and else block contains similar code..

AmanAgarwal041:
> Can we move this code to a new function and call it? Both if and else block contains similar code..

If block is for the multi select which represents selected items as array and else block is for single value which represent selected item as a string. So, moving them in a different function would also consist of a similar code.

ChandanBalajiBP (approved):

PR #16079, opened 2022-08-17
fix: welcome tour ending on page refresh

### LINKED ISSUE #10923: [Bug]-[700]:Welcome tutorial ends on page refresh

### Current Behavior

When the user is in the middle of the tutorial and the page gets refreshed the tutorial ends abruptly.

https://loom.com/share/1dd574b0442e4630a94b8747aafe1cfd

### Steps To Reproduce

1. Start the Welcome tutorial > Click on refresh on any step of the tutorial and observe 

### Environment

Release

### Version

Cloud

akash-codemonk:
## Description

We want to still have the guided tour running when the current page is refreshed. Note that we don't restore at the initial screen where we show the start building cta. Also we don't restore all states like for e.g focused inputs, if the widget sidebar was open we don't reopen it on refresh etc. but the tour continues to be usable. We also go back a few steps in cases where we cannot resume with the latest step on refresh.

Code changes
- To do this have moved the tour related store values to a separate reducer, which is then synced with session storage at every step.
- Minor refactor to make reuse code for both localstorage and sessionstorage. 
- Refresh the page in cypress while running the tour test.

Release 

https://user-images.githubusercontent.com/67054171/185177909-3b19262e-d2d2-4567-ada1-cadcaea08def.mov

Post fix

https://user-images.githubusercontent.com/67054171/185178036-ee5bbef1-f403-4801-9261-c3aa4321be1b.mov

Fixes #10923

## Type of change

## How Has This Been Tested?

Cypress

## Checklist:


app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/Onboarding/GuidedTour_spec.js
```diff
@@ -19,6 +19,8 @@ describe("Guided Tour", function() {
     cy.get(guidedTourLocators.welcomeTour).click();
     cy.get(guidedTourLocators.startBuilding).click();
     cy.get(explorerLocators.entityExplorer).should("not.be.visible");
+    // Refresh the page to validate if the tour resumes
+    cy.reload();
```
hetunandu, line 23: After reload, assert that the guide banner is still showing

app/client/src/pages/Editor/GuidedTour/utils.ts
```diff
@@ -245,3 +251,8 @@ export function showIndicator(
 export function hideIndicator() {
   indicatorHelperInstance.destroy();
 }
+
+export function closeSidebar(dispatch: Dispatch<ReduxAction<any>>) {
```
hetunandu, line 255: We should not pass dispatch as an argument here. Instead you can batch these actions, via redux batch. Returning an array of actions in this function, and dispatching it from outside should work

app/client/src/utils/localStorage.tsx
```diff
@@ -16,23 +16,41 @@ class LocalStorageNotSupportedError extends Error {
   }
 }
 
-export const getLocalStorage = () => {
-  const storage = window.localStorage;
+type WebStorageType = "local" | "session";
```
hetunandu, line 19: Can we create a new WebStorage or Session storage class instead? Lets not combine these.

hetunandu (changes requested):

app/client/src/utils/localStorage.tsx
akash-codemonk: Created two classes LocalStorage and SessionStorage which extends WebStorage as both local and session storage have the same api.

hetunandu (approved):

PR #16088, opened 2022-08-17
chore: remove mockServiceWorker.js from build since its not needed in prod

### LINKED ISSUE #14395: [Task]: mock API calls/response for RBAC

berzerkeer:
## Description
Though we only use msw in `development` environment, when we run `yarn build` the `public/mockServiceWorker.js` still gets bundled and will be an unused file in build. We don't want that, so added a `postbuild` script which removes the file from the build folder post build.

Fixes #14395

## Type of change

## How Has This Been Tested?

- Manual

## Checklist:

### DIFF
--- a/app/client/package.json
+++ b/app/client/package.json
@@ -176,6 +176,7 @@
     "test:unit": "$(npm bin)/jest -b --colors --no-cache --silent --coverage --collectCoverage=true --coverageDirectory='../../' --coverageReporters='json-summary'",
     "test:jest": "$(npm bin)/jest --watch",
     "generate:widget": "plop --plopfile generators/index.js",
+    "postbuild": "rm build/mockServiceWorker.js",
     "postinstall": "patch-package && CURRENT_SCOPE=client node ../shared/install-dependencies.js",
     "preinstall": "CURRENT_SCOPE=client node ../shared/build-shared-dep.js"
   },


yaldram (approved):

berzerkeer:
Ref: https://github.com/appsmithorg/appsmith/pull/15965

app/client/package.json
```diff
@@ -176,6 +176,7 @@
     "test:unit": "$(npm bin)/jest -b --colors --no-cache --silent --coverage --collectCoverage=true --coverageDirectory='../../' --coverageReporters='json-summary'",
     "test:jest": "$(npm bin)/jest --watch",
     "generate:widget": "plop --plopfile generators/index.js",
+    "postbuild": "rm build/mockServiceWorker.js",
```
IAmAnubhavSaini, line 179: Is there no `prebuild` step or `skip` file option for build? 

Ideally, we would want to not have it in `build/`.

IAmAnubhavSaini (approved):

app/client/package.json
berzerkeer: I am not sure , https://github.com/mswjs/msw/issues/291 this is something from a github thread and I initially tried to exclude the file using webpack but didn't succeed.

yaldram (approved):

PR #16096, opened 2022-08-17
fix: Add mongo/redis to helm chart dependencies

pratapaprasanna:
## Description

Add redis and mongo as individual pods to our helm charts.

Fixes [15929](https://github.com/appsmithorg/appsmith/issues/15929)

PFA

**Redis:**
Screenshot 2022-08-17 at 1 42 43 PM

Screenshot 2022-08-17 at 1 44 54 PM

Screenshot 2022-08-17 at 3 28 43 PM
Screenshot 2022-08-17 at 3 28 49 PM

**Mongo deployment pics:**
Screenshot 2022-08-30 at 5 38 23 PM
Screenshot 2022-08-30 at 5 39 17 PM
Screenshot 2022-08-30 at 5 39 31 PM

**Mongo:**
Demo of the deployment of how it looks.

https://user-images.githubusercontent.com/15846947/190350964-2ec81f73-9c46-4cb6-b476-a0b1908c8f93.mov

## Without redis and mongo
```
(base) ➜  helm git:(ops/inhouse/db) ✗ head values.yaml
## Redis parameters
redis:
  enabled: false
  auth:
    enabled: false
  replica:
    replicaCount: 1

mongodb:
  enabled: false
(base) ➜  helm git:(ops/inhouse/db) ✗ helm install appsmith . -n goutham --create-namespace
Error: INSTALLATION FAILED: An error occurred while checking for chart dependencies. You may need to run `helm dependency build` to fetch missing dependencies: found in Chart.yaml, but missing in charts/ directory: redis, mongodb
(base) ➜  helm git:(ops/inhouse/db) ✗ helm dep build
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "appsmith-ce" chart repository
...Successfully got an update from the "ingress-nginx" chart repository
...Successfully got an update from the "jetstack" chart repository
...Successfully got an update from the "aws-efs-csi-driver" chart repository
...Successfully got an update from the "maildev" chart repository
...Successfully got an update from the "mongodb" chart repository
...Successfully got an update from the "grafana" chart repository
...Successfully got an update from the "appsmith" chart repository
...Successfully got an update from the "prometheus-community" chart repository
...Successfully got an update from the "logdna" chart repository
...Successfully got an update from the "bitnami" chart repository
...Successfully got an update from the "stable" chart repository
Update Complete. ⎈Happy Helming!⎈
Saving 2 charts
Downloading redis from repo https://charts.bitnami.com/bitnami
Downloading mongodb from repo https://charts.bitnami.com/bitnami
Deleting outdated charts
(base) ➜  helm git:(ops/inhouse/db) ✗ helm install appsmith . -n goutham --create-namespace
NAME: appsmith
LAST DEPLOYED: Fri Oct 14 15:07:10 2022
NAMESPACE: goutham
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
1. Get the application URL by running these commands:
  export POD_NAME=$(kubectl get pods --namespace goutham -l "app.kubernetes.io/name=appsmith,app.kubernetes.io/instance=appsmith" -o jsonpath="{.items[0].metadata.name}")
  export CONTAINER_PORT=$(kubectl get pod --namespace goutham $POD_NAME -o jsonpath="{.spec.containers[0].ports[0].containerPort}")
  echo "Visit http://127.0.0.1:8080 to use your application"
  kubectl --namespace goutham port-forward $POD_NAME 8080:$CONTAINER_PORT

To expose your Appsmith service to be accessible from the Internet, please refer to our docs here https://docs.appsmith.com/setup/kubernetes#expose-appsmith.
```
### ConfigMap
```
(base) ➜  helm git:(ops/inhouse/db) ✗ kubectl get configmap appsmith -o yaml -n goutham
apiVersion: v1
data:
  APPSMITH_DISABLE_IFRAME_WIDGET_SANDBOX: "false"
kind: ConfigMap
metadata:
  annotations:
    meta.helm.sh/release-name: appsmith
    meta.helm.sh/release-namespace: goutham
  creationTimestamp: "2022-10-14T09:37:11Z"
  labels:
    app.kubernetes.io/instance: appsmith
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: appsmith
    app.kubernetes.io/version: 1.16.0
    appsmith.sh/chart: appsmith-1.5.3
  name: appsmith
  namespace: goutham
  resourceVersion: "35030762"
  uid: d1e3db85-be50-4944-a2e8-066bd03522da

```
### Init containers 
```
# Source: appsmith/templates/statefulset.yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: appsmith
  namespace: goutham
  labels:
    appsmith.sh/chart: appsmith-1.5.3
    app.kubernetes.io/name: appsmith
    app.kubernetes.io/instance: appsmith
    app.kubernetes.io/version: "1.16.0"
    app.kubernetes.io/managed-by: Helm
spec:
  replicas: 1
  serviceName: appsmith
  updateStrategy:
    type: RollingUpdate
  selector:
    matchLabels:
      app.kubernetes.io/name: appsmith
      app.kubernetes.io/instance: appsmith
  template:
    metadata:
      labels:
        app.kubernetes.io/name: appsmith
        app.kubernetes.io/instance: appsmith
    spec:
      serviceAccountName: appsmith
      securityContext:
        {}
      initContainers:
      containers:
        - name: appsmith
          securityContext:
            {}
```
### configMap
```
---
# Source: appsmith/templates/configMap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: appsmith
  namespace: goutham
  labels:
    appsmith.sh/chart: appsmith-1.5.3
    app.kubernetes.io/name: appsmith
    app.kubernetes.io/instance: appsmith
    app.kubernetes.io/version: "1.16.0"
    app.kubernetes.io/managed-by: Helm
data:
  APPSMITH_DISABLE_IFRAME_WIDGET_SANDBOX: "false"
  APPSMITH_MONGODB_URI: mongodb+srv://root:password@appsmith-mongodb.goutham.svc.cluster.local/appsmith?retryWrites=true&authSource=admin&ssl=false
  APPSMITH_REDIS_URL: redis://appsmith-redis-master.goutham.svc.cluster.local:6379
```

### Init containers
```
---
# Source: appsmith/templates/statefulset.yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: appsmith
  namespace: goutham
  labels:
    appsmith.sh/chart: appsmith-1.5.3
    app.kubernetes.io/name: appsmith
    app.kubernetes.io/instance: appsmith
    app.kubernetes.io/version: "1.16.0"
    app.kubernetes.io/managed-by: Helm
spec:
  replicas: 1
  serviceName: appsmith
  updateStrategy:
    type: RollingUpdate
  selector:
    matchLabels:
      app.kubernetes.io/name: appsmith
      app.kubernetes.io/instance: appsmith
  template:
    metadata:
      labels:
        app.kubernetes.io/name: appsmith
        app.kubernetes.io/instance: appsmith
    spec:
      serviceAccountName: appsmith
      securityContext:
        {}
      initContainers:
      - name: redis-init-container
        image: alpine
        command: ['sh', '-c', "apk add redis; until redis-cli -h appsmith-redis-master.goutham.svc.cluster.local ping; do echo waiting for redis; sleep 2; done"]
      - name: mongo-init-container
        image: docker.io/bitnami/mongodb:4.4.11-debian-10-r12
        command: ['sh', '-c', "until mongo --host appsmith-mongodb.goutham.svc.cluster.local --eval 'db.runCommand({ping:1})' ; do echo waiting for mongo; sleep 2; done"]
      containers:
        - name: appsmith
          securityContext:
            {}
```


deploy/helm/values.yaml
```diff
@@ -265,8 +281,8 @@ applicationConfig:
   APPSMITH_RECAPTCHA_SITE_KEY: ""
   APPSMITH_RECAPTCHA_SECRET_KEY: ""
   APPSMITH_RECAPTCHA_ENABLED: ""
-  APPSMITH_MONGODB_URI: ""
-  APPSMITH_REDIS_URL: ""
+  APPSMITH_MONGODB_URI: "mongodb+srv://root:password@appsmith-mongodb-headless/appsmith?retryWrites=true"
+  APPSMITH_REDIS_URL: "redis://appsmith-redis-master:6379"
```
sharat87, line 285: Can we rename these to just `appsmith-mongodb` and `appsmith-redis`? 😄

deploy/helm/values.yaml
```diff
@@ -1,3 +1,19 @@
+## Redis parameters
+redis:
+  auth:
+    enabled: false
+  replica:
+    replicaCount: 2
```
sharat87, line 6: Let's set the default replicas of Redis to 1. Redis usage isn't that high today that we should default to 2 replicas. Besides, if a user wants it, they can change this value anyway.

deploy/helm/templates/statefulset.yaml
```diff
@@ -43,6 +43,13 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      initContainers:
+      - name: redis-init-container
+        image: alpine
+        command: ['sh', '-c', "apk add redis; until redis-cli -h appsmith-redis-master.{{.Release.Namespace}}.svc.cluster.local ping; do echo waiting for redis; sleep 2; done"]
+      - name: mongo-init-container
+        image: docker.io/bitnami/mongodb:4.4.11-debian-10-r12
+        command: ['sh', '-c', "until mongo --host appsmith-mongodb-headless --eval 'db.runCommand({ping:1})' ; do echo waiting for mongo; sleep 2; done"]
```
sharat87, line 52: Question, doesn't the host here need a `.{{.Release.Namespace}}.svc.cluster.local` suffix like we have one for Redis? 🤔

sharat87 (commented):
@pratapaprasanna, thanks for opening this.

This new chart will need manual upgrading, right? For example, if someone is running the current Helm Chart (`v1.5.3`), their MongoDB data would be in the Appsmith pod's PV. But with this chart, that data should be in the mongodb pods' PV. Which means moving from that Helm Chart to this one, will need some migration effort, right? Can we document the steps needed for someone already using current Helm Chart to migrate to this one please?

Also, because of this reason, let's make this Helm Chart version `v2.0.0` in `Chart.yaml`. What say?

cc @mohanarpit

mohanarpit:
> @pratapaprasanna, thanks for opening this.
> 
> This new chart will need manual upgrading, right? For example, if someone is running the current Helm Chart (`v1.5.3`), their MongoDB data would be in the Appsmith pod's PV. But with this chart, that data should be in the mongodb pods' PV. Which means moving from that Helm Chart to this one, will need some migration effort, right? Can we document the steps needed for someone already using current Helm Chart to migrate to this one please?
> 
> Also, because of this reason, let's make this Helm Chart version `v2.0.0` in `Chart.yaml`. What say?
> 
> cc @mohanarpit

Will this mean that we will deprecate v1.x for our Helm charts and urge users to follow the migration steps v2.x? If yes, will any critical fixes be made to both v1.x & v2.x Helm charts? Also, when users run the command `helm upgrade` today, will Helm pick up the v1.x chart or v2.x chart?

sharat87:
@mohanarpit, I'd rather we take a call on deprecating v1.x, after we know what it takes to upgrade to this Helm Charts. If it's a simple one or two steps, then lets deprecate. We are still hosting charts all the way from 1.3 I think, which is just becoming a baggage.

Yeah, critical fixes will need to be made in both, but, we haven't seen any such critical issues in our Helm Charts so far. 🤞

I don't think `helm upgrade` upgrades the version of the Helm charts. That command is to apply any changes in our `values.yml` configuration to our existing chart. @pratapaprasanna, please correct if this is not right.

deploy/helm/templates/statefulset.yaml
pratapaprasanna: Good catch, we should change it to that way.

deploy/helm/values.yaml
pratapaprasanna: We can change the svc name of mongo and i see that we cannot do the same for redis.

sharat87:
@pratapaprasanna, I see that you've requested a review, but I don't see a comment with your thoughts on backwards compatibility and a migration path. Can you take a look at https://github.com/appsmithorg/appsmith/pull/16096#pullrequestreview-1094413731 and share your thoughts please?

deploy/helm/values.yaml
```diff
@@ -265,8 +282,8 @@ applicationConfig:
   APPSMITH_RECAPTCHA_SITE_KEY: ""
   APPSMITH_RECAPTCHA_SECRET_KEY: ""
   APPSMITH_RECAPTCHA_ENABLED: ""
-  APPSMITH_MONGODB_URI: ""
-  APPSMITH_REDIS_URL: ""
+  APPSMITH_MONGODB_URI: "mongodb+srv://root:password@appsmith-mongodb.appsmith.svc.cluster.local/appsmith?retryWrites=true&authSource=admin&ssl=false"
+  APPSMITH_REDIS_URL: "redis://appsmith-redis-master.appsmith.svc.cluster.local:6379"
```
pratapaprasanna, line 286: @sharat87 @mohanarpit @sum35h I need your inputs in this 

```appsmith-redis-master.appsmith.svc.cluster.local``` and 
```appsmith-mongodb.appsmith.svc.cluster.local```
will be variable . I mean the k8s convention is 

```<service_name>.<release_namespace>.svc.cluster.local```
So, should we just leave it empty and in the document we explain on how the values should be given?
any inputs ?

Comment (sharat87) [reply deploy/helm/values.yaml:286]: These values are used to fill up a `ConfigMap`, right? If yes, in the `ConfigMap` template, can we resolve this? Like, we check if this value if empty, if yes, then we set it use the namespace value since we would have this information there, right? This way, users can leave it empty, or not set it at all, and our `ConfigMap` template will infer the value itself.

I'm trying to see if we can do this for the user, but if that's not possible, then of course we can document it and direct users to set it.

Comment (sharat87) [reply deploy/helm/values.yaml:286]: @pratapaprasanna, hey, can we make a call on this? Let me know if we should discuss this further on a call.

Comment (pratapaprasanna) [reply deploy/helm/values.yaml]: Done, Updated, thanks for the suggestion

**Review by sharat87 — COMMENTED** [review]
@pratapaprasanna, I recall we have developed migration steps for users to move to this new MongoDB, right? Is that part of a separate PR, on `appsmith-docs`? Will we be merging it after this goes out?

File: .gitignore [hunk 5]
```diff
@@ -22,6 +22,10 @@ app/client/perf/traces/*
 .history
 stacks
 
+# helm dependencies
+deploy/helm/charts
+deploy/helm/Chart.lock
```
sharat87, line 27: Let's keep the lock file no? If `charts` is like `node_modules`, then I'm thinking `Chart.lock` is like the `yarn.lock` file that pins the dependencies for predictable builds. The docs at <https://helm.sh/docs/helm/helm_dependency_update/#synopsis> say:

> On successful update, this will generate a lock file that can be used to rebuild the dependencies to an exact version.

I think it's a good idea to keep it committed.

sharat87: Oh wait, I also see that we _are_ committing the `Chart.lock` file. Can we remove this like then please? 😅

```suggestion
```

deploy/helm/templates/configMap.yaml
```diff
@@ -7,7 +11,17 @@ metadata:
     {{- include "appsmith.labels" . | nindent 4 }}
 data:
   {{- range $key, $value := .Values.applicationConfig }}
+  {{- if and (eq "APPSMITH_MONGODB_URI" $key) ( not $value) }}
+  {{- if $.Values.mongodb.enabled }}
+  {{ $key }}: mongodb+srv://{{ $mongoUser }}:{{ $mongoPassword }}@{{ $mongoServicename }}.{{ $nameSpace }}.svc.cluster.local/appsmith?retryWrites=true&authSource=admin&ssl=false
+  {{- end }}
+  {{- end }}
```
sharat87, line 18: This looks like we're missing a condition. If condition on line 14 is true, and on line 15 is false, that is `mongodb.enabled` is `false`, we should just revert to previous behavior, right? Like this:

```suggestion
  {{- if and (eq "APPSMITH_MONGODB_URI" $key) ( not $value) }}
  {{- if $.Values.mongodb.enabled }}
  {{ $key }}: mongodb+srv://{{ $mongoUser }}:{{ $mongoPassword }}@{{ $mongoServicename }}.{{ $nameSpace }}.svc.cluster.local/appsmith?retryWrites=true&authSource=admin&ssl=false
  {{- else }}
  {{ $key }}: {{ $value | quote }}
  {{- end }}
  {{- end }}
```

This is to keep existing behavior, more than anything else. What say? Similarly for Redis as well.

deploy/helm/templates/statefulset.yaml
```diff
@@ -43,6 +43,17 @@ spec:
...
+      {{- if .Values.redis.enabled }}
+      - name: redis-init-container
+        image: alpine
+        command: ['sh', '-c', "apk add redis; until redis-cli -h appsmith-redis-master.{{.Release.Namespace}}.svc.cluster.local ping; do echo waiting for redis; sleep 2; done"]
+      {{- end }}
+      {{- if .Values.mongodb.enabled }}
+      - name: mongo-init-container
+        image: docker.io/bitnami/mongodb:4.4.11-debian-10-r12
+        command: ['sh', '-c', "until mongo --host appsmith-mongodb.{{.Release.Namespace}}.svc.cluster.local --eval 'db.runCommand({ping:1})' ; do echo waiting for mongo; sleep 2; done"]
+      {{- end }}
```
sharat87, line 56: If both `mongodb.enabled` and `redis.enabled` are `false`, then `initContainers` would be empty, right? Wouldn't that cause a syntax error here? Since, I think empty maps should be written as `initContainers: {}`?

pratapaprasanna: It worked. I mean once they are disabled initcontainers defaulted to an emptydict. So, everything worked fine.There was no syntax error

deploy/helm/templates/configMap.yaml
pratapaprasanna: Yes, previous behavior is an empty value/whatever the user gives right. 
So,
if the `APPSMITH_MONGODB_URI` doesnt contain the value and if mongodb is enabled.
```{{ $key }}: mongodb+srv://{{ $mongoUser }}:{{ $mongoPassword }}@{{ $mongoServicename }}.{{ $nameSpace }}.svc.cluster.local/appsmith?retryWrites=true&authSource=admin&ssl=false```
if the `APPSMITH_MONGODB_URI` doesnt contain the value and if mongodb is disabled. 
 ```{{ $key }}: {{ $value | quote }}```

pratapaprasanna: That will happen in the end because of the for loop at line 25.

sharat87: Oh, I see. But, that means `APPSMITH_MONGODB_URI` and `APPSMITH_REDIS_URL` could be duplicated, right? Like, if they are printed from lines 16 and 21, they can also be printed in the for loop no? Can we rewrite this using `else` statements to avoid duplicates?

pratapaprasanna: You mean what if we have duplicate values in values.yaml. Like, having two different values for APPSMITH_MONGODB_URI in the values.yaml ?

sharat87 (approved):

ep-Alex:
Hi,

im currently facing an issue with in an kubernetes environment which doesn't have access to the internet.
The appsmith container is always in the INIT:0/2 state because the helm chart is using an alpine image and is installing the redis-cli through the command durring the initialization: 
 
 `sh
      -c
      apk add redis; until redis-cli -h appsmith-redis-master.manufacturing-appsmith-dev.svc.cluster.local ping; do echo waiting for redis; sleep 2; done`
 
that is why the init process is logging the following messages all the time:

`sh: redis-cli: not found`
`waiting for redis`

looks like the current helm chart is not usable in such an offline environment and should be improved.

Issue #14135, opened 2022-05-27
[Task]: Data type parsing logic for evaluated values


nidhi-nair:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SubTasks

- Identify valid data types(primitive)
- Implement logic to map these data types
- Add map of evaluated value data types to action execution request

ChandanBalajiBP:
Few cases considered for typeof
//True cases
// Numbers
typeof 37                === 'number';
typeof 3.14              === 'number';
typeof Math.LN2          === 'number';
typeof Infinity          === 'number';
typeof NaN               === 'number'; // Despite being "Not-A-Number"
typeof Number(1)         === 'number'; // This case is not handled by typeof. so loking for better approach.

// Strings
typeof ""                === 'string';
typeof "bla"             === 'string';
typeof (typeof 1)        === 'string'; // typeof always return a string
typeof String("abc")     === 'string'; // but never use this form!

// Booleans
typeof true              === 'boolean';
typeof false             === 'boolean';
typeof Boolean(true)     === 'boolean'; // but never use this form!

// Undefined
typeof undefined         === 'undefined';
typeof blabla            === 'undefined'; // an undefined variable

// Objects
typeof {a:1}             === 'object';
typeof [1, 2, 4]         === 'object'; // use Array.isArray or Object.prototype.toString.call to differentiate regular objects from arrays
typeof new Date()        === 'object';
typeof new Boolean(true) === 'object'; // this is confusing. Don't use!
typeof new Number(1)     === 'object'; // this is confusing. Don't use!
typeof new String("abc") === 'object';  // this is confusing. Don't use!

// Functions
typeof function(){}      === 'function';
typeof Math.sin          === 'function';

//False cases - typeof returns wrong value for wrapper objects
Example:
const message = new String('This will not work');
typeof message //Return 'object' - but expected output is 'string'
typeof value === 'string' || value instanceof String; //can be resolved using

ChandanBalajiBP:
Have chosen below function to find the datatype
const trueTypeOf = (obj) => Object.prototype.toString.call(obj).slice(8, -1).toLowerCase()

Test cases considered - includes all above and 
trueTypeOf([]); // array
trueTypeOf({}); // object
trueTypeOf(''); // string
trueTypeOf(new Date()); // date
trueTypeOf(1); // number
trueTypeOf(function () {}); // function.
trueTypeOf(/test/i); // regexp
trueTypeOf(true); // boolean
[...]

PR #16106, opened 2022-08-17
chore: Modified action execution payload to support data types

### LINKED ISSUE #15555: [Task]: Consume map of autogenerated names and actual binding parameters and ensure compatibility

### Summary

As part of the Datatype Handling project, the client-side will send the possible data type of each binding parameter along with its value. The client-side will send such payload with the autogenerated name of the actual binding parameter with a view to reducing the payload size. 

The client-side will send the possible data type (_JS Type)_ of each binding parameter along with its value. The client-side will send such payload with the autogenerated name of the actual binding parameter with a view to reducing the payload size. 
The client-side will also send the mapping between the autogenerated name and the actual binding parameter name in a different part called parameterMap. The sample request with three binding parameters would look as follows:

```
------WebKitFormBoundaryj9bhZXa3DECRp83J
Content-Disposition: form-data; name="executeActionDTO"

{"actionId":"627a09673b06f26463e16ba8","viewMode":false, "paramProperties": [{"k1": "STRING"}, {"k2": "NULL"}, {"k3": "NULL"}]}
------WebKitFormBoundaryj9bhZXa3DECRp83J
Content-Disposition: form-data; name="parameterMap"

{"k1":"Text1.text","k2":"Table1.data", "k3": "Api1.data"}
------WebKitFormBoundaryj9bhZXa3DECRp83J
Content-Disposition: form-data; name="k1" filename="blob"
Content-Type: text/plain
abc
------WebKitFormBoundaryj9bhZXa3DECRp83J
Content-Disposition: form-data; name="k2"; filename="blob"
Content-Type: text/plain

null
------WebKitFormBoundaryj9bhZXa3DECRp83J
Content-Disposition: form-data; name="k3"; filename="blob"
Content-Type: text/plain

null
------WebKitFormBoundaryj9bhZXa3DECRp83J--
```

In this task, the server will consume this new payload and will not apply any logic to the identification of the data type rather it will use the parameterMap in order to decode the original binding parameter's name. It also needs to be ensured that the change in the action execution request body payload doesn't break any existing implementation and hence ensuring compatibility.

subrata71:
### Summary

As part of the Datatype Handling project, the client-side will send the possible data type of each binding parameter along with its value. The client-side will send such payload with the autogenerated name of the actual binding parameter with a view to reducing the payload size. 

### Description

The client-side will send the possible data type (_JS Type)_ of each binding parameter along with its value. The client-side will send such payload with the autogenerated name of the actual binding parameter with a view to reducing the payload size. 
The client-side will also send the mapping between the autogenerated name and the actual binding parameter name in a different part called parameterMap. The sample request with three binding parameters would look as follows:

```
------WebKitFormBoundaryj9bhZXa3DECRp83J
Content-Disposition: form-data; name="executeActionDTO"

{"actionId":"627a09673b06f26463e16ba8","viewMode":false, "paramProperties": {"k1": "STRING", "k2": "NULL","k3": "NULL"}}
------WebKitFormBoundaryj9bhZXa3DECRp83J
Content-Disposition: form-data; name="parameterMap"

{"Text1.text": "k1","Table1.data": "k2", "Api1.data": "k3"}
------WebKitFormBoundaryj9bhZXa3DECRp83J
Content-Disposition: form-data; name="k1" filename="blob"
Content-Type: text/plain
abc
------WebKitFormBoundaryj9bhZXa3DECRp83J
Content-Disposition: form-data; name="k2"; filename="blob"
Content-Type: text/plain

null
------WebKitFormBoundaryj9bhZXa3DECRp83J
Content-Disposition: form-data; name="k3"; filename="blob"
Content-Type: text/plain

null
------WebKitFormBoundaryj9bhZXa3DECRp83J--
```

In this task, the server will consume this new payload and will not apply any logic to the identification of the data type rather it will use the parameterMap in order to decode the original binding parameter's name. It also needs to be ensured that the change in the action execution request body payload doesn't break any existing implementation and hence ensuring compatibility.

Fixes #15555 #14135

## Type of change

## How Has This Been Tested?

- Manual
- JUnit

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/NewActionServiceCEImpl.java
```diff
@@ -859,22 +858,41 @@ public Mono<ActionExecutionResult> executeAction(Flux<Part> partFlux, String bra
                                     try {
                                         return Mono.just(objectMapper.readValue(byteData, ExecuteActionDTO.class));
                                     } catch (IOException e) {
+                                        log.error("Error in deserializing ExecuteActionDTO", e);
```
nidhi-nair, line 861: Nit: Is this a duplicate log considering we will also have a log from the GlobalExceptionHandler @subrata71 ?

subrata71: Yes, `GlobalExceptionHandler` will take this up but here the entire `stacktrace` is printed. We expect that this won't happen too often but if it happens then the `stacktrace` would give us better insight. Does it make sense? @nidhi-nair

nidhi-nair (approved):

app/client/src/workers/evaluationUtils.ts
```diff
@@ -704,6 +704,19 @@ export const isTrueObject = (
...
+ * @param value
+ * @returns datatype of the received value as string
+ */
+export const findDatatype = (value: unknown) => {
+  return Object.prototype.toString
+    .call(value)
+    .slice(8, -1)
+    .toLowerCase();
+};
+
```
Rishabh-Rathod, line 719: Hey @subrata71, 
I see that we added this method in `evaluationUtils` file which is in the worker folder and used it in `PluginActionSaga` file which is the main thread part of the code.  

This will lead to the bundling of `evaluationUtils` file both for the worker bundle and the main bundle.  

Was this intentional?

subrata71: @ChandanBalajiBP Can you please answer Rishabh's query?

Rishabh-Rathod: Has a discussion with @ChandanBalajiBP. This was unintentional and I will move it to a different file in my PR.

Issue #16104, opened 2022-08-17
[Bug]:  appsmithctl check-replica-set fails to connect to an external APPSMITH_MONGODB, when it is running as a container in the same host


sum35h:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

When running the appsmithctl check-replica-set utils command with the APPSMITH_MONGODB_URL pointing to a mongo running as a container in the same host, the following error is thrown 
```
Error trying to check replicaset MongoServerSelectionError: getaddrinfo ENOTFOUND 26543e32cf22
    at Timeout._onTimeout (/opt/appsmith/utils/node_modules/mongodb/lib/sdam/topology.js:312:38)
    at listOnTimeout (internal/timers.js:557:17)
    at processTimers (internal/timers.js:500:7) {
  reason: TopologyDescription {
    type: 'ReplicaSetNoPrimary',
    servers: Map(1) { '26543e32cf22:27017' => [ServerDescription] },
    stale: false,
    compatible: true,
    heartbeatFrequencyMS: 10000,
    localThresholdMS: 15,
    setName: 'mr1',
    commonWireVersion: 6,
    logicalSessionTimeoutMinutes: undefined
  }
}
```

### Steps To Reproduce

1. Deploy a Mongodb container(mongo:latest) locally with replica set enabled
2. Deploy an Appsmith container on the same host and specify the mongo uri in the docker.env file:
`APPSMITH_MONGODB_URI=mongodb://<user>:<password>@host.docker.internal:27017/appsmith?authSource=admin
`
3. Observe the appsmith container logs to find the error thrown when the entrypoint.sh uses the  appsmithctl check-replica-set command.

### Version

v1.7.14

PR #16151, opened 2022-08-19
fix: entrypoint.sh external mongo replica set check

sum35h:
Updated `entrypoint.sh` to use a mongo shell eval command instead of` appsmithctl check-replica-set` for **external mongo db**.

Partially fixes(Workaround): [16104](https://github.com/appsmithorg/appsmith/issues/16104)

### DIFF
--- a/deploy/docker/entrypoint.sh
+++ b/deploy/docker/entrypoint.sh
@@ -168,7 +168,7 @@ init_replica_set() {
     # Check mongodb cloud Replica Set
     echo "Checking Replica Set of external MongoDB"
 
-    if appsmithctl check_replica_set; then
+    if [[ $(mongo $APPSMITH_MONGODB_URI --quiet --eval "rs.status().members.length") -gt 1 ]]; then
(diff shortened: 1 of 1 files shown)


deploy/docker/entrypoint.sh
```diff
@@ -168,7 +168,7 @@ init_replica_set() {
     # Check mongodb cloud Replica Set
     echo "Checking Replica Set of external MongoDB"
 
-    if appsmithctl check_replica_set; then
+    if [[ $(mongo $APPSMITH_MONGODB_URI --quiet --eval "rs.status().ok") -eq 1 ]]; then
```
sharat87, line 171: Let's wrap the URL in double-quotes. If it contains any special characters in its password, for example, it'll still work. :)

```suggestion
    if [[ $(mongo "$APPSMITH_MONGODB_URI" --quiet --eval "rs.status().ok") -eq 1 ]]; then
```

sharat87 (changes requested):

sharat87 (approved):

Issue #13978, opened 2022-05-20
[Bug]: Curl Import Not Detecting Right Body Type


cokoghenun:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Whenever a URL is imported via CURL, the right body type is not detected and always defaults to RAW

### Steps To Reproduce

Try importing this multipart form request

```curl
curl --location --request POST 'https://api.stripe.com/v1/transfers' \
--header 'Authorization: Basic c2tfdGVzdF81MUpqSlh0RlRiN09mV3VhUjFTcXJoNFF3cU5haDZHcDBhanRpMGhqZlczWjl4MXg4RzFRYUhxNlRZcXBvd1IwVGczcDFadWViQjAwSHB4VnBiUmg6' \
--header 'Content-Type: application/x-www-form-urlencoded' \
--data-urlencode 'amount=1' \
--data-urlencode 'currency=usd' \
--data-urlencode 'destination=acct_1L0mdx2Ym60mujGg'
```

### Version

Cloud/v1.6.26-SNAPSHOT

AmanAgarwal041:
Here's the RCA :
File `CurlImporterServiceCEImpl.java` has a function `parse` - line number `310`. This function parses the curl command being imported from the client side, but it does not handles the `formData` in the `actionConfiguration` object which determines the Body Type for the client side editor. 
If we set `apiContentType` key inside `actionConfiguration.formData` object with respect to the `contentType` being calculated, then the issue is resolved.

cokoghenun:
@sondermanish any updates on this? 
I get PTSD every time I realize I'm going to use CURL on Appmsith 😂

sondermanish:
@cokoghenun trying to figure out the issue. Let's cure cURL. #nomorePTSD

prapullac:
@sneha122 : Verified on the DP and looks good to me

PR #16159, opened 2022-08-19
fix: cURL not detecting the right body type

sondermanish:
## Description

This commit fixes the bug https://github.com/appsmithorg/appsmith/issues/13978 and changes two files:
- CurlImporterServiceTest.java
- CurlImporterServiceCEImpl.java

Additionally this commit fixes the Body type detection for REST APIs with GET method  when imported from cURL.

Fixes https://github.com/appsmithorg/appsmith/issues/13978

## Type of change

- Bug fix (Now the cURL import will be able to detect the right body type)

## How Has This Been Tested?
- This PR has been tested:
        * Manually
        * Junit
        
- As the Junit coverage is adequate, we don't require Cypress test case for this PR. 

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/CurlImporterServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/CurlImporterServiceCEImpl.java
@@ -52,7 +52,6 @@ public class CurlImporterServiceCEImpl extends BaseApiImporter implements CurlIm
     private static final String ARG_USER = "--user";
     private static final String ARG_USER_AGENT = "--user-agent";
     private static final String FORM_DATA_KEY = "apiContentType";
-    private static final String CONTENT_TYPE = "content-type";
 
     private final PluginService pluginService;
     private final LayoutActionService layoutActionService;
@@ -350,7 +349,7 @@ public class CurlImporterServiceCEImpl extends BaseApiImporter implements CurlIm
                 }
                 if ("content-type".equalsIgnoreCase(parts[0])) {
                     contentType = parts[1];
-                    parts[0] = CONTENT_TYPE;
+                    parts[0] = HttpHeaders.CONTENT_TYPE;
                     //Setting the apiContentType to the content-type detected in the header with the key word content-type.
                     // required for RestAPI calls with GET method having body.
                     actionConfiguration.setFormData(Map.of(FORM_DATA_KEY, contentType));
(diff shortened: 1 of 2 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/CurlImporterServiceCEImpl.java
```diff
@@ -50,6 +51,8 @@ public class CurlImporterServiceCEImpl extends BaseApiImporter implements CurlIm
     private static final String ARG_COOKIE = "--cookie";
     private static final String ARG_USER = "--user";
     private static final String ARG_USER_AGENT = "--user-agent";
+    private static final String FORM_DATA_KEY = "apiContentType";
+    private static final String CONTENT_TYPE = "content-type";
```
nidhi-nair, line 55: @manish535 I get that we're switching to lower case here for consistent behaviour with what the client understands, but can we make the client more lenient with how it understands the header as well? HTTP headers are meant to be [case insensitive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers) and we should not force this rigidity at any layer.

Also, in general it is recommended to retain the casing that was originally present here for readability and standardization, so if anything we should only be making client side changes for this fix.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/CurlImporterServiceTest.java
```diff
@@ -335,7 +336,7 @@ public void multilineCommand() throws AppsmithException {
 
         final ActionConfiguration actionConfiguration = action.getActionConfiguration();
         assertEmptyPath(action);
-        assertHeaders(action, new Property("Content-Type", "application/json"));
```
nidhi-nair, line 338: Also, were you able to figure out why the client is using the header and not the `formdata.apiContentType` property to render the correct body tab?

sondermanish: The front-end doesn't use this property as of now to make detect the right body type, rather it is dependent on the parsed header for right detection, the formData.apiContentType is used for detecting the Body type when there is get request, if it is null then GET request will be sent without Body, verified with manual testing. Have made changes to address that.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/CurlImporterServiceCEImpl.java
sondermanish: @nidhi-nair Have made changes accordingly, This PR is not complete as the issue will persist as long as Front-end doesn't add  leniency in header checks. Assigning the issue to @AmanAgarwal041.

nidhi-nair:
Closing in favour of #16549

PR #16240, opened 2022-08-23
feat: Add server-side base data types (#15784)

### LINKED ISSUE #15784: [Feature]: Add all the supported server-side Appsmith data types

### SubTasks

_No response_

subrata71:
## Description

This commit takes care of the following two things
- It adds the server-side base data types which are considered common data types across plugins
- It adds a few MySQL specific types

Fixes #15784 

## Type of change

## How Has This Been Tested?

- This PR only adds the server-side base Appsmith types. There's no usage of these data types so no explicit test case is required. Only CI would suffice its validity.

## Checklist:


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/datatypes/mysql/MySQLDatetimeType.java
```diff
@@ -0,0 +1,58 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import reactor.core.Exceptions;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+import java.time.format.DateTimeParseException;
+import java.util.regex.Matcher;
+
+public class MySQLDatetimeType implements AppsmithType {
```
nidhi-nair, line 17: Nit: Should we maintain camel case throughout?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/constants/DataType.java
```diff
@@ -17,5 +17,6 @@ public enum DataType {
     JSON_OBJECT,
     TIMESTAMP,
     BSON,
-    BSON_SPECIAL_DATA_TYPES
-}
+    BSON_SPECIAL_DATA_TYPES,
+    BIGDECIMAL
```
nidhi-nair, line 21: Where will we be using this new type?

nidhi-nair (commented):
Minor comments

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/constants/DataType.java
subrata71: We are barely able to use BIGDECIMAL as identified data type. I have just added this as the base data type for preferring this to any other types e.g. Float, Double while dealing with fractional numbers / monetary data (Futuristic thought). 

We don't have so many cases to go with the double type even. You know that the server first tries to parse fractional numbers as Float and Double will get a chance if and only if float overflows.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/datatypes/mysql/MySQLDatetimeType.java
nidhi-nair: Oh yes, I meant that we'd want to use `MySQLDateTimeType` instead right?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/datatypes/AppsmithType.java
```diff
@@ -0,0 +1,10 @@
+package com.appsmith.external.datatypes;
+
+import com.appsmith.external.constants.DataType;
+
+import java.util.function.Predicate;
+
+public interface AppsmithType extends Predicate<String> {
+    String performSmartSubstitution(String value);
+    DataType type();
```
nidhi-nair, line 9: What is the purpose of this method?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/constants/DataType.java
nidhi-nair: I'm guessing in terms of JDBC types, this would be a numeric?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/datatypes/mysql/MySQLDatetimeType.java
subrata71: Changed accordingly. Thanks

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/constants/DataType.java
subrata71: The corresponding `JDBC` type is `DECIMAL` but it may not be applicable for all the SQL plugins e.g. MySQL. In MySQL r2dbc library there are only two methods available to bind parameters:
- `bindNull` (To bind null object)
- `bind` (To bind not-null object)
@nidhi-nair

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/datatypes/AppsmithType.java
subrata71: FYI, `DataType` is the existing enum type that is being used in so many places. 

### Let me explain the way I thought

Let's say the `AppsmithType` for a binding parameter is `DateType`. So, we are supposed to set the binding through the following function of that SQL library. 
`preparedStatement.setDate(value)`

We can do this in either of the ways
- `if (appsmithType instance of DateType) preparedStatement.setDate(value);`
- `if (DataType.DATE.equals(appsmithType.type()) preparedStatement.setDate(value);`
I chose the latter approach because there are many occurrences where the `DataType` enum is used.

@nidhi-nair

nidhi-nair: Wouldn't the enum be redundant after we move to the new types? Are you thinking of phasing it out eventually?

subrata71: Without the Datatype enum how will we decide which function to use for binding the parameter?


### To bind timestamp in Postgres
preparedStatement.setTimestamp(Date.valueOf(value))

### To bind date in Postgres
preparedStatement.setDate(Timestamp.valueOf(value));

Can you please elaborate on your thought?
@nidhi-nair

nidhi-nair: I was thinking that we'd have a method, let's call it `prepareValue` for now, that we'd call to perform this operation. This way, the consumption of the type does not have to handle types at all anymore which means we can get rid of all the switch statements, and the `AppsmithType` would take care of the prepared statement implementation as well. I assume the downside is that we'd need to pass the prepared statement and index around to `prepareValue` but that would still be a stateless operation. What do you think?

subrata71: Sorry, I still have some grey areas. I am not sure how `prepareValue` function will take care of this without switch-like statements. 

Not every plugin comes with index-based binding. 
For example

**MySQL**
`bind(int index, Object value);`

**Postgres**
This plugin offers different methods for allowing setting different types of parameters 
```
setBoolean(int parameterIndex, boolean x) 
setBigDecimal(int parameterIndex, BigDecimal x)
setString(int parameterIndex, String x)
```

@nidhi-nair

PR #16252, opened 2022-08-23
feat: Feat/rts configuration for ast

### LINKED ISSUE #15647: [Task]: Create Endpoint to return an AST for a string.

### SubTasks

_No response_

Irongade:
Setting up rts server for Ast endpoints

Fixes #15647

## How Has This Been Tested?

- Test A
- Test B

## Checklist:


app/rts/src/controllers/Ast/AstController.ts
```diff
@@ -0,0 +1,35 @@
...
+type ScriptToIndentifiersType = {
+  script: string;
+  evalVersion?: number;
+};
+export default class AstController extends BaseController {
+  constructor() {
+    super();
+  }
+
+  getDependentIndentifiers(req: Request, res: Response) {
```
nidhi-nair, line 15: Nit: Looks like a typo

app/rts/src/controllers/Ast/AstController.ts
```diff
@@ -0,0 +1,35 @@
...
+    try {
+      const errors = validationResult(req);
+      if (!errors.isEmpty())
+        return super.sendError(res, "Validation error", errors);
+
+      // By default the application eval version is set to be 2
+      const { script, evalVersion = 2 } : ScriptToIndentifiersType = req.body;
+      const data = AstService.getIndentifiersFromScript(script, evalVersion);
```
nidhi-nair, line 23: Nit: Same typo here

app/rts/src/controllers/Ast/AstController.ts
AmanAgarwal041: Fixed

app/rts/src/controllers/Ast/AstController.ts
AmanAgarwal041: Fixed

app/rts/src/sockets/index.ts
```diff
@@ -5,6 +5,6 @@ import log from "loglevel";
 
 // Initializing Multiple Sockets
 export function initializeSockets(io: Server) {
-  watchMongoDB(io).catch((error) => log.error("Error watching MongoDB", error));
+  // watchMongoDB(io).catch((error) => log.error("Error watching MongoDB", error));
```
Irongade, line 8: Let's not forget to uncomment this after we are done, so as not to break anything

app/rts/src/controllers/BaseController.ts
```diff
@@ -0,0 +1,61 @@
...
+      message,
+      data: result,
+    });
+  }
+
+  sendError(
+    response: Response,
+    error: string,
+    errorMessage,
+    code: number = 400
```
mohanarpit, line 40: Minor: Suggest using https://www.npmjs.com/package/http-status-codes instead of hard coding HTTP status codes. A constant ensures that we are always compliant with the HTTP spec.

app/rts/src/routes/ast_routes.ts
```diff
@@ -0,0 +1,20 @@
+import express from "express";
+import AstController from "../controllers/Ast/AstController";
+import AstRules from "../middlewares/rules/ast";
+
+const router = express.Router();
+const astController = new AstController();
+
+router.post(
+  "/single-script-identifiers",
```
mohanarpit, line 9: Minor: Suggest constructing the URL as `api/v1/single-script-identifiers` . This ensures that we can create v2 version of the API if required in the future.

app/rts/src/controllers/BaseController.ts
Irongade: Very good suggestion, This has been done.

app/rts/src/controllers/Ast/AstController.ts
```diff
@@ -0,0 +1,72 @@
...
+    try {
+      const errors = validationResult(req);
+      if (!errors.isEmpty())
+        return super.sendError(res, "Validation error", errors);
+
+      // By default the application eval version is set to be 2
+      const { scripts, evalVersion = 2 }: MultipleScriptToIdentifiersType =
+        req.body;
+
+      Promise.all(
```
ohansFavour, line 55: This seems like the only difference between getDependentIdentifiers and getMultipleDependentIdentifiers. Can we reduce duplicated logic here?

app/rts/src/controllers/BaseController.ts
```diff
@@ -0,0 +1,62 @@
+import { Response } from "express";
+import { ValidationError } from "express-validator";
+import { StatusCodes } from "http-status-codes";
+
+type ErrorData = {
+  error: string | string[];
+  validation_error?: ValidationError[];
```
ohansFavour, line 7: suggested change=> validationErrors?: ValidationError[]

app/rts/src/controllers/BaseController.ts
```diff
@@ -0,0 +1,62 @@
...
+  success: boolean;
+  message?: string;
+  data: any;
+};
+
+export default class BaseController {
+  serverErrorMessaage = "Something went wrong";
+  sendResponse(
+    response: Response,
+    result: any,
```
ohansFavour, line 26: If there is no schema here, can the type be `unknown` instead?

app/rts/src/middlewares/Validator.ts
```diff
@@ -0,0 +1,13 @@
+import BaseController from "@controllers/BaseController";
+import { validationResult } from "express-validator";
+import { Response, Request, NextFunction } from "express";
+
+export class Validator extends BaseController {
```
mohanarpit, line 5: QQ: Are we also going to add a validator to validate if the request is coming from a logged in user's session?

app/rts/src/server.ts
```diff
@@ -6,7 +6,11 @@ import log, { LogLevelDesc } from "loglevel";
 import { VERSION as buildVersion } from "./version"; // release version of the api
 import { initializeSockets } from "./sockets";
 
+// routes
+import ast_routes from "./routes/ast_routes";
+
 const RTS_BASE_PATH = "/rts";
+const RTS_BASE_API_PATH = "/rts-api/v1";
```
mohanarpit, line 13: Any particular reason we are using separate endpoints like `rts-api/v1` for the API as compared to `rts/api/v1`

AmanAgarwal041: We have socket server mounted on `/rts`. If we mount https server on `/rts/api/v1` we get an error on request `{"code":0,"message":"Transport unknown"}`

app/rts/src/middlewares/Validator.ts
AmanAgarwal041: We are not sure right now. Right now rts api path is not exposed via nginx conf for fat container. But we can access it through using host and port. We will be introducing authorization in server side implementation if required cc - @nidhi-nair

mohanarpit: The `rts` endpoint is exposed on Nginx on the fat container as shown [here](https://github.com/appsmithorg/appsmith/blob/release/deploy/docker/templates/nginx/nginx-app-http.conf.template.sh#L81). I think we have so many different Nginx configurations in the code base, hence it gets a little confusing.

Also, I think (to be confirmed) there will be an issue when communicating via localhost:port in HA environments. 

@pratapaprasanna Do you think k8s deployments will support the scenario where appsmith server talks to RTS via localhost:port? Or will this break?

app/rts/src/server.ts
mohanarpit: In the Nginx file, we can setup separate endpoints for `/rts` and `/rts/api`. This way it'll redirect the api to the correct endpoints.

Dockerfile
```diff
@@ -60,7 +60,6 @@ COPY ./app/client/build editor/
 
 # Add RTS - Application Layer
 COPY ./app/rts/package.json ./app/rts/dist rts/
-COPY ./app/rts/node_modules rts/node_modules
```
mohanarpit, line 63: Why are we removing the `node_modules` from being copied into the Docker container. If I'm not mistaken, this is required to run the rts inside the fat Docker container.

AmanAgarwal041: We are already copying `node_modules` in the build.sh file https://github.com/appsmithorg/appsmith/pull/16252/files#diff-a1afc7e36108e70e01e2a3364adc449c5a764cc20f79275ac76e0aa325e3bfcdR8 , so we need not to copy it once again. The docker copy command was copying the `node_modules` and `@shared/ast` was getting copied with symlink rather than the full folder with contents. This problem is solved by copying node_modules folder into dist in the build process only.

mohanarpit: Thanks. That makes sense. 
Minor: Can you please add a one-line comment in the `build.sh` file explaining this? I'm asking because I don't think it's a general practice to copy node_modules into the `dist` folder for Nodejs builds.

app/rts/src/middlewares/Validator.ts
Irongade: @pratapaprasanna We would like you thoughts here please.

PR #16253, opened 2022-08-23
fix: add JUnit test cases for GraphQL plugin

### LINKED ISSUE #12388: [Task]: graphql Action test cases

### SubTasks

_As an engineer, I should ensure the quality and sanity of the code flows by providing automated test cases on the server side_

@Sumit Kumar placeholder to list the test cases here

1. Test for successfull graphQL query with content type json.
2. Test for successfull graphQL query with content type graphql.
3. Add test for successful graphQL query with query variables.
4. Test for successfull graphQL query with smart substitution in query variables.
5. Test for successfull graphQL query with smart substitution in query body.
6. Test for successfull graphQL query with both get and post type of requests.
7. Add test to catch syntax error in query body. 
8. Add test to catch syntax error in query variables. 
9. (Manual ?) test with Oauth

sumitsum:
## Description
- add JUnit test cases for GraphQL plugin. This PR addresses all the use cases mentioned in #12388 and more.
- re-locate JUnit test cases for common code that was re-factored from REST API plugin.
- add change to assume offset value as `0`in limit-offset based pagination when null or blank value is provided by the user.
- add support for `application/graphql` content type.

Fixes #12388

## Type of change
- Task (neither a bug fix nor a feature)

## How Has This Been Tested?
- JUnit TC
- Manual

## Checklist:


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ApiContentType.java
```diff
@@ -12,7 +10,8 @@ public enum ApiContentType {
     JSON("application/json"),
     FORM_URLENCODED("application/x-www-form-urlencoded"),
     MULTIPART_FORM_DATA("multipart/form-data"),
-    RAW("raw")
+    RAW("raw"),
+    GRAPHQL("application/graphql")
```
nidhi-nair, line 14: Does adding the type here affect REST API in any way? I believe this list is supposed to be used as a reference for supported body tabs in the REST API pane.

app/server/appsmith-plugins/graphqlPlugin/src/test/java/com/external/plugins/GraphQLPluginTest.java
```diff
@@ -0,0 +1,1043 @@
...
+                            assertEquals(token, field.getValue().get(0).asText());
+                        }
+                    });
+
+                })
+                .verifyComplete();
+    }
+
+    @Test
+    public void testInvalidSignature() {
```
nidhi-nair, line 315: Sorry what are we testing for in this method?

app/server/appsmith-plugins/graphqlPlugin/src/test/java/com/external/plugins/GraphQLPluginTest.java
```diff
@@ -0,0 +1,1043 @@
...
+                    JSONParser jsonParser = new JSONParser(JSONParser.MODE_PERMISSIVE);
+                    ObjectMapper objectMapper = new ObjectMapper();
+                    try {
+                        JSONObject resultJson = (JSONObject) jsonParser.parse(String.valueOf(result.getBody()));
+                        // Object resultData = ((JSONObject) resultJson.get("json")).get("query");
+                        String resultData = ((JSONObject) resultJson.get("json")).getAsString("query");
+                        String parsedJsonAsString = objectMapper.writeValueAsString(resultData);
+                        assertEquals(expectedQueryBody, resultData);
+                    } catch (ParseException | JsonProcessingException e) {
+                        e.printStackTrace();
```
nidhi-nair, line 411: Can we use an assertion error here to make sure the test fails if we do get to this LOC?

app/server/appsmith-plugins/graphqlPlugin/src/test/java/com/external/plugins/GraphQLPluginTest.java
```diff
@@ -0,0 +1,1043 @@
...
+        Mono<GraphQLPlugin.GraphQLPluginExecutor> pluginExecutorMono = Mono.just(pluginExecutor);
+        Mono<Set<String>> invalidsMono = pluginExecutorMono.map(executor -> executor.validateDatasource(datasourceConfiguration));
+
+        StepVerifier
+                .create(invalidsMono)
+                .assertNext(invalids -> invalids.containsAll(Set.of("Missing Client ID", "Missing Client Secret", "Missing Access Token URL")));
+    }
+
+    @Test
+    public void testSmartSubstitutionQueryBody1() {
```
nidhi-nair, line 359: Could you please specify what different scenarios we're covering in both tests? It is not easy to understand this from the method names right now.

app/server/appsmith-plugins/graphqlPlugin/src/test/java/com/external/plugins/GraphQLPluginTest.java
```diff
@@ -0,0 +1,1043 @@
...
+                "      name\n" +
+                "    }\n" +
+                "  }\n" +
+                "}");
+        Mono<ActionExecutionResult> resultMono = pluginExecutor.executeParameterized(null, new ExecuteActionDTO(), dsConfig, actionConfig);
+
+        StepVerifier.create(resultMono)
+                .verifyErrorSatisfies(error -> {
+                    assertTrue(error instanceof AppsmithPluginException);
+                    String expectedMessage = "Invalid GraphQL body: Invalid Syntax : There are more tokens in the " +
```
nidhi-nair, line 241: Are these error messages something that the endpoint used in this test throws? Or is it something Appsmith will consistently throw for such a query body? I'm a little uncomfortable introducing tests that rely on so many different services because we'll be dependent on their validity/accuracy for our tests to pass.

As a separate point, do you think we should use MockWebServer instead?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ApiContentType.java
sumitsum: @nidhi-nair as of now `ApiContentType` enum models all the types that the `Body` tab of REST API Plugin has. One way to look at this would be that `application/graphql` is a type that corresponds to the `Body` tab of GraphQL Plugin, and since both the plugins are a derivative of the same base REST API model we could keep this here. Othewise I will create an issue to re-factor this code. Currently we rely on a common flow and hence the same enum `ApiContentType` is used for both the plugins. In case we want to move separate out the types for both the plugins then some refactor would be required. This refactor could be taken up post shipping to production, since currently there should be no negative impact to REST API plugin due to this change.

app/server/appsmith-plugins/graphqlPlugin/src/test/java/com/external/plugins/GraphQLPluginTest.java
sumitsum: @nidhi-nair Sorry, I couldn't understand your comment here. Are you saying that we are relying on any external endpoint here ? If so, then we are not relying on any external endpoint here for the validation. We use a graphql package that does the validation and produces the message.

app/server/appsmith-plugins/graphqlPlugin/src/test/java/com/external/plugins/GraphQLPluginTest.java
sumitsum: I have copied over this test from the RestAPI tests but doesn't look like it is testing anything useful here, hence removed it now.

app/server/appsmith-plugins/graphqlPlugin/src/test/java/com/external/plugins/GraphQLPluginTest.java
sumitsum: I have added comments to both the methods to clarify this.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ApiContentType.java
nidhi-nair: Yes I think we should maintain separate derivative types for both plugins. That way any extension to the feature can be managed easily. I'll leave the planning for refactoring up to you. :)

app/server/appsmith-plugins/graphqlPlugin/src/test/java/com/external/plugins/GraphQLPluginTest.java
nidhi-nair: We should opt to rename the method instead of writing comments where possible so that the tests are self explanatory even from just the run history. Do you think we can rename these methods?

app/server/appsmith-plugins/graphqlPlugin/src/test/java/com/external/plugins/GraphQLPluginTest.java
nidhi-nair: It was a question for the entire test suite actually. I think I spotted some Rick and Morty url in the PR. The more we rely on external APIs the more prone to failures our test suite will be. I was wondering if we can mock those out?

nidhi-nair (commented):
Added a few more comments.

app/server/appsmith-plugins/graphqlPlugin/src/test/java/com/external/plugins/GraphQLPluginTest.java
sumitsum: Done.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ApiContentType.java
sumitsum: As discussed, I have created this issue to track it: https://github.com/appsmithorg/appsmith/issues/16457 .

app/server/appsmith-plugins/graphqlPlugin/src/test/java/com/external/plugins/GraphQLPluginTest.java
sumitsum: As discussed, some of the test cases have been migrated to use the GraphQL server within our `TED` docker container which has been included as a `testcontainer`. Rest of the test cases would be migrated later. The plan would be to migrate possible TC to use mock server and then also work with QA to enhance the `TED` GraphQL server so that the rest of the test cases could be migrated. I have created this issue to track it: https://github.com/appsmithorg/appsmith/issues/16434

nidhi-nair (approved):

PR #16263, opened 2022-08-24
fix: update def on focus

### LINKED ISSUE #14749: [Bug]: Lot of deviation in time taken to setup autocomplete

### Description

`startAutoComplete`  took 3600ms in one profile and 2400ms in other profile. That is 50% deviation, while the rest of the code has around 2.5% deviation.

It would help with performance monitoring and performance in general if this deviation can be reduced.
Profiles can be found in the slack thread.

https://theappsmith.slack.com/archives/C024GUDM0LT/p1655467501490139

### Steps To Reproduce

Check the profiles on slack thread attached.

### Public Sample App

_No response_

### Version

Both / Latest

### LINKED ISSUE #16206: [Task]: update autocomplete defs on code editor focus

### SubTasks

Currently, autocomplete customDataTree defs are updated when the code editor gets mounted on the tree. As a result, when two code editor with different customDataTree is present on the property pane, the second code editor's defs overwrites the previous one. So we need to update the defs only when code editor gets focused.

Rishabh-Rathod:
## Description

It was noticed that the TernServer's `deleteDefs` and `addDefs` method are expensive which were internally ran when `TernServer.updateDef` API was called.
- Earlier we were running `TernServer.updateDef` for each codeEditor onMount which was expensive and not required. Now, it is changed to be done only when the editor is focused.
    - Avoiding to run `TernServer.updateDef`  for each codeEditor instance.
- In case when editor with the same customDataTreeDef's requirement is opened in that case we do a deep comparison of previous `customDataTreeDef` and only if different then trigger `TernServer.updateDef`
 

Fixes #16206 
Fixes #14749

## Type of change

- Enhancement to support `editedValue` feature in autocomplete added [here](https://github.com/appsmithorg/appsmith/pull/15954)
- Performance Improvement 

## How Has This Been Tested?

- [ ] Check performance impact

Future test 
- [ ] Once Table Widget `evaluatedValue` [feature](https://github.com/appsmithorg/appsmith/pull/15954/files) is added autocompletion test for property supporting `evaluatedValue` and property which doesn’t to make sure customDataTreeDef updates

## Checklist:


Rishabh-Rathod (commented):
Self-review done.

app/client/src/utils/autocomplete/customDefUtils.ts
```diff
@@ -0,0 +1,38 @@
...
+  AdditionalDynamicDataTree,
+  customTreeTypeDefCreator,
+} from "./customTreeTypeDefCreator";
+import TernServer from "./TernServer";
+
+let lastCustomDataDef: AdditionalDynamicDataTree | undefined;
+
+export const updateCustomDef = (customData?: AdditionalDynamicDataTree) => {
+  if (customData && !isEmpty(customData)) {
+    const customDataDef = customTreeTypeDefCreator(customData);
```
arunvjn, line 14: Can we cache last customData ? We can compare customData and lastCustomData. Avoids customTreeTypeDef calculation

app/client/src/utils/autocomplete/customDefUtils.ts
```diff
@@ -0,0 +1,38 @@
...
+
+      debug(
+        "Tern: updateDef for customDataTree took",
+        (performance.now() - start).toFixed(),
+        "ms",
+      );
+
+      lastCustomDataDef = customDataDef;
+    }
+  } else if (lastCustomDataDef) {
```
arunvjn, line 28: Can you explain what happens here ? If the lastCustomData and the newCustomData are the same, will it not simply remove the definitions from Tern ?

Rishabh-Rathod: When `customData` is not sent or is empty in those cases we need to `removeDef`.

As `removeDef` also is expensive call we try to only do it once to actually remove customDataDef. 

Once we run removeDef
- lastCustomDataDef becomes `undefined` and hence we have a check that only if `lastCustomDataDef` was defined then run removeDef.

Rishabh-Rathod: > it will remove the definitions right ?

No, it will do nothing. 
it will go inside 1st if the case where customData if defined then it will generate customDataDef and deep compare with previous customDataDef as they are same it will not go inside this condition`if (!equal(lastCustomDataDef, customDataDef)) {`

arunvjn: My bad. Read it wrong!

arunvjn (approved):
Looks good!

arunvjn (approved):

arunvjn (approved):

app/client/src/utils/autocomplete/customDefUtils.ts
```diff
@@ -0,0 +1,38 @@
+import equal from "fast-deep-equal/es6";
+import { isEmpty } from "lodash";
+import { debug } from "loglevel";
+import {
+  AdditionalDynamicDataTree,
+  customTreeTypeDefCreator,
+} from "./customTreeTypeDefCreator";
+import TernServer from "./TernServer";
+
+let lastCustomDataDef: AdditionalDynamicDataTree | undefined;
```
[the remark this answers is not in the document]
Rishabh-Rathod: done

sbalaji1192 (approved):

PR #16317, opened 2022-08-26
feat: Maintain Focus and Context Phase 1

### LINKED ISSUE #15158: [Feature]-[5000]:Maintaining cursor position/context when changing pages

#### Job Story:
When I switch between different APIs, Queries, JSObjects, and properties for widgets within an application, I want to ensure the previous APIs, Queries, JSObjects, and properties for widgets is exactly how I left it, so I can maintain context, and not feel lost when I return to that context.

#### Solution:
Retaining the cursor and scroll position of APIs, Queries, JSObjects, and properties for widgets, allowing users to quickly scan back and fourth between pages

#### decision points to be taken:
depth of retention: how many levels deep do we retain information for? (at the very least 1, but we would like to do a degree of performance testing, and find the appropriate blend)

1. Watching users interact with appsmith, and seeing them navigate back and fourth from one page to another, and scan across their entire file to find where they would like to be/were
2. This is good hygine, and best practice for having a robust IDE

### LINKED ISSUE #16235: [Task] Deprecate auto refresh on chunk load fail

## Background
With the use of RTS to identify release version mismatch and prompts to refresh the page, the auto-refresh approach is moot.

- Remove references to `donotReloadOnChunkLoadFail`
- Remove auto-refresh mechanism handled [here](https://github.com/appsmithorg/appsmith/blob/5cf06cd071072959b04b8bf516cec1120a928e70/app/client/public/index.html#L122)

hetunandu:
## Description
Phase 1 of maintaining focus and context

Fixes #16235
Fixes #15158

## Type of change

## How Has This Been Tested?

TODO

## Checklist:


app/client/src/components/editorComponents/CodeEditor/EvaluatedValuePopup.tsx
```diff
@@ -369,11 +394,23 @@ export const CurrentValueViewer = memo(
 
 function PopoverContent(props: PopoverContentProps) {
   const typeTextRef = React.createRef<HTMLPreElement>();
-  const [openExpectedDataType, setOpenExpectedDataType] = useState(false);
+  const dispatch = useDispatch();
+  const popupContext = useSelector((state: AppState) =>
+    getEvaluatedPopupState(state, props.dataTreePath),
+  );
+  const [openExpectedDataType, setOpenExpectedDataType] = useState(
```
hetunandu, line 401: Why maintain a local state over here?

app/client/src/components/editorComponents/CodeEditor/index.tsx
```diff
@@ -529,11 +539,13 @@ class CodeEditor extends Component<Props, State> {
...
   };
 
-  handleEditorBlur = () => {
+  handleEditorBlur = (cm: CodeMirror.Editor) => {
     this.handleChange();
     this.setState({ isFocused: false });
     this.editor.setOption("matchBrackets", false);
     this.handleCustomGutter(null);
+    const { ch, line } = cm.getCursor();
+    this.props.setCursorPosition(this.props.dataTreePath, { line, ch });
```
hetunandu, line 548: Check if dataTreePath exists first

app/client/src/components/editorComponents/CodeEditor/index.test.tsx
```diff
@@ -1,109 +0,0 @@
-import React from "react";
```
hetunandu, line 1: Why delete this test?

app/client/src/components/editorComponents/PropertyPaneSidebar.tsx
```diff
@@ -71,6 +74,11 @@ export const PropertyPaneSidebar = memo((props: Props) => {
     PerformanceTracker.stopTracking();
   });
 
+  useEffect(() => {
```
hetunandu, line 77: I feel this should be invoked on the canvas and not the property pane

app/client/src/components/editorComponents/CodeEditor/EvaluatedValuePopup.tsx
rahulramesha: With the local state, the component can work individually even when dataTreePath is undefined. Also we only need to set default state rather than using Redux state all the time for every single codeEditor on the Page.

app/client/src/components/editorComponents/CodeEditor/index.tsx
rahulramesha: I'm handling the undefined situation directly in the Reducer, can check here as well

app/client/src/components/editorComponents/CodeEditor/index.test.tsx
rahulramesha: Deleted by mistake will restore it

app/client/src/components/editorComponents/PropertyPaneSidebar.tsx
rahulramesha: If we have to this invoked in canvas, we will have to subscribe to get selected Widgets state there, which will cause it to re-render the canvas when selected widget state change. Invoking it here since this component will always exist with canvas and already subscribes to this state

Richarex:
@hetunandu 
When the user creates two pages and on **"Page1"** from the **Property pane** under **"Theme Properties"** closes 'Font' , 'Color' & 'Border' properties and switches to "Page2" observes that the same changes on **"Page2"** & vice-versa.
1. Go to DP URL:- https://appsmith-eputc306c-get-appsmith.vercel.app/user/login
2. Click on 'New Workspace' & create a new app 
3. Create two pages on the same app "Page1" & "Page2"
4. Now from the 'Property Pane' under **"Theme Properties"** close 'Font' , 'Color' & 'Border' properties
5. Switch to **"Page2"** & observe

**Actual Result**
**"Page1"** theme properties changes are reflecting on **"Page2"**

**Expected Result**
**"Page1"** theme properties changes should not reflect on **"Page2"** & vice- versa

**NOTE**
Working fine on old apps.
Refer to the attached video.
https://www.loom.com/share/6ddb454c7dff4969ab9a715c14b6eb56

rahulramesha:
@Richarex, this seems to be existing behaviour in release, could you please check and confirm?. We haven't really thought about this use case, Will tackle this in Phase 2 of context switching.

Richarex:
@hetunandu JS object editor page when the user selects **"Logs"** tab & applies the filter **"Console Logs"** switches to other API and navigating back to JS object editor page the applied filter is reset. Please refer the video attached: https://www.loom.com/share/69dca84f51354a82b0eb9f7da9d1a38a

PR #16335, opened 2022-08-26
chore: some datasources not appearing in templates tab

### LINKED ISSUE #16334: [Bug]: Twilio not appearing in the filter list in the templates tab.

### Description

bug

Although twillio is added to the datasource list of the template, it isn't appearing in the list of filters.

### Steps To Reproduce

1. Add a new template with twilio as a datasource.
2. Go to the templates tab and see that the template is present, but the twilio datasource isn't in the filter list.

More context:  https://theappsmith.slack.com/archives/C02RCR2PURZ/p1660912771736459

### Public Sample App

_No response_

### Version

Cloud

akash-codemonk:
## Description

Fixes #16334 

Currently the template creator app(which is an appsmith app) adds the `packageName` of the plugins/datasources to each of the templates metadata. The frontend then fetches information like image, name of the plugin from the packageName to display the UI. Some plugins have the same package name for e.g airtable and twilio have `saas-plugin` as the package name. To workaround issue for now the fix adds support for plugin id along with the package name, which is unique. 

We are not shifting to pluginId completely since the metadata needs to be updated for all the templates, so for now we will support both on the frontend and remove dependency on packageName once all the template metadata has been updated.

## Type of change

## How Has This Been Tested?

Manually

## Checklist:


app/client/src/selectors/templatesSelectors.tsx
```diff
@@ -104,15 +104,34 @@ export const getSearchedTemplateList = createSelector(
...
   getDefaultPlugins,
-  (plugins) => {
-    return plugins.map((plugin) => {
-      return {
-        label: plugin.name,
-        value: plugin.packageName,
-      };
+  (templates, plugins) => {
+    const datasourceFilters: Filter[] = [];
+    templates.map((template) => {
```
eco-monk, line 113: Nitpicks:
- We could use `forEach` since we're not returning anything.
- I don't know how long the list of datasources can be. If it can be really long, we could create `datasourceFilters` as a map initially and convert it to array later in the end. 
`dataSourceMap.has` can be used instead of `datasourceFilters.find`.

eco-monk (approved):

PR #16351, opened 2022-08-27
fix: Super user updated from docker.env on every restart

### LINKED ISSUE #16324: [Bug]: Unable to Configure Super User Directly in the docker.env File

### Description

The implementation of making **super users** works only when changed via the **Admin Settings** page but not if it was directly changed in the `docker.env` file.

https://app.intercom.com/a/apps/y10e7138/inbox/inbox/conversation/164629100161174#part_id=initial-part-164629100161174-1413541376

### Steps To Reproduce

1. Configure the super user directly in the docker.env file.
2. Restart the instance.
3. The changes do not take effect.

### Public Sample App

_No response_

### Version

Self-Hosted

trishaanand:
Fixes #16324


mohanarpit (approved):

sharat87:
@trishaanand, if I remove an email from this env variable, will they stop being an admin? Asking since I couldn't figure this out from the migration code.

trishaanand:
> @trishaanand, if I remove an email from this env variable, will they stop being an admin? Asking since I couldn't figure this out from the migration code.

@sharat87 : Yes they would. Only the emails mentioned in the field would then have the super user permissions.

PR #16356, opened 2022-08-29
chore: Refactor repo limit check function

AnaghHegde:
## How Has This Been Tested?
- Locally

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -399,18 +399,12 @@ public Mono<String> commitApplication(GitCommitDTO commitDTO, String defaultAppl
...
                                     return Mono.just(defaultApplication);
                                 }
-                                return this.getApplicationCountWithPrivateRepo(workspaceId)
-                                        .map(privateRepoCount -> {
-                                            if (limit >= privateRepoCount) {
-                                                return defaultApplication;
-                                            }
-                                            throw new AppsmithException(AppsmithError.GIT_APPLICATION_LIMIT_ERROR);
-                                        });
+                                throw new AppsmithException(AppsmithError.GIT_APPLICATION_LIMIT_ERROR);
```
trishaanand, line 407: Minor Q : Any reason why we are not sending analytics events in this scenario but seem to be doing that for the rest of the instances where repo limit has been reached?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -2457,4 +2434,20 @@ FieldName.ORGANIZATION_ID, defaultIfNull(application.getWorkspaceId(), ""),
                     return application;
                 });
     }
+
+    private Mono<Boolean> isRepoLimitReached(String workspaceId, Boolean isClearCache) {
+        return gitCloudServicesUtils.getPrivateRepoLimitForOrg(workspaceId, isClearCache)
```
trishaanand, line 2439: We don't seem to be using the `isClearCache` variable. Can you help me understand the variable and how it was being used earlier? Thanks!

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: This is a commit flow. This is a hack where the user makes the repo public while connecting but changes this to private and the commit will fail. So we decided this is not required in analytics.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
AnaghHegde: This is used in the util class where we decide if we need to fetch the limit from the cloud service or use it from the cache. For connect and import flow we fetch the data from CS but for commit(scenario mentioned in the above comment) we use the value from the cache.

trishaanand (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -2457,4 +2434,20 @@ FieldName.ORGANIZATION_ID, defaultIfNull(application.getWorkspaceId(), ""),
                     return application;
                 });
     }
+
+    private Mono<Boolean> isRepoLimitReached(String workspaceId, Boolean isClearCache) {
```
trishaanand, line 2438:
```suggestion
    protected Mono<Boolean> isRepoLimitReached(String workspaceId, Boolean isClearCache) {
```

trishaanand (approved):

trishaanand (approved):

PR #16401, opened 2022-08-30
fix: TreeWidgets issues

### LINKED ISSUE #15466: [Bug]: TreeSelect | Unable to capture right Label data OnPageLoad (first load)

### Description

Screenshot 2022-07-26 at 8 46 28 PM
Screenshot 2022-08-29 at 5 07 07 PM

### Steps To Reproduce

1. Dragdrop Tree select widget/MultiTree & map for the selected options display
2. Observe the selected option value & label data
3. **While the label data is Blue - the data captured is only "B" - can be seen in screenshot - TreeSelect**
4. **While the label data is Green - the data captured is "GREEN" - which is option value - MultiTreeSelect**
5. Now change the selected valued in Tree Select & observe right data is captured
6. **Expected - The right Label (and Value) data should be available (at all times and) during first load also**

### Public Sample App

_No response_

### Version

Prod/Release v1.7.10

### LINKED ISSUE #16365: [Bug]: "selectedOptionValue" in TreeSelect widget does not work as expected

### Description

Even when we clear the selected option in a TreeSelect widget, the value of the same is still stored.

**Note:** 

1. This happens when there is a `Default Value` mentioned for the TreeSelect widget.
2. This is happening only in release and not in Prod.

https://www.loom.com/share/efbfe26332a948f5b9d63917229b8e25

### Steps To Reproduce

1. D&D TreeSelect & Text widgets.
3. Bind the Text widget to the TreeSelect widget such that the `Text` property of the Text widget is `{{TreeSelect1.selectedOptionValue}}`.
4. Now clear the selected option in the TreeSelect widget.

**Expected:** The text in the Text widget should also disappear. 

### Public Sample App

_No response_

### Version

Cloud

### LINKED ISSUE #16367: [Bug]: Issues in MultiTreeSelct widget

### Description

Two issues are happening with the MultiTreeSelect widget in release:

1. The `selectedOptionLabels` displays the values of the selected options instead of the labels.
2. On binding it with a Text widget using: `{{MultiTreeSelect1.selectedOptionLabels}}` when we clear all the selected options, the Text widget still displays the labels.

**Note:**

This happens when there is a Default Value mentioned for the TreeSelect widget.
This is happening only in release and not in Prod.

https://www.loom.com/share/8e6e11471fe2409fb786d556db0292e9

### Steps To Reproduce

1. D&D MultiTreeSelect & Text widgets.
2. Bind the Text widget to the MultiTreeSelect widget such that the Text property of the Text widget is {{MultiTreeSelect1.selectedOptionLabels}}.
4. Now clear the selected option in the TreeSelect widget.

**Expected:** The text in the Text widget should also disappear.

### Public Sample App

_No response_

### Version

Cloud

Tooluloope:
## Description

This Fixes issues with Both Single Tree Select and Multi Tree Select.

- Clearing Values
- `SelectedOptionLabels` for both widgets

Fixes #16367
Fixes #16365
Fixes #15466

## Type of change

## How Has This Been Tested?
- Unit Test

## Checklist:

### DIFF
--- a/app/client/src/widgets/MultiSelectTreeWidget/widget/index.tsx
+++ b/app/client/src/widgets/MultiSelectTreeWidget/widget/index.tsx
@@ -2,7 +2,7 @@ import React, { ReactNode } from "react";
 import BaseWidget, { WidgetProps, WidgetState } from "widgets/BaseWidget";
 import { TextSize, WidgetType } from "constants/WidgetConstants";
 import { EventType } from "constants/AppsmithActionConstants/ActionConstants";
-import { isArray, findIndex, xor } from "lodash";
+import { isArray, xor } from "lodash";
 import {
   ValidationResponse,
   ValidationTypes,
@@ -17,8 +17,6 @@ import MultiTreeSelectComponent from "../component";
 import { LabelPosition } from "components/constants";
 import { Alignment } from "@blueprintjs/core";
 import derivedProperties from "./parseDerivedProperties";
-import equal from "fast-deep-equal/es6";
-import { flat } from "widgets/WidgetUtils";
 
 function defaultOptionValueValidation(value: unknown): ValidationResponse {
   let values: string[] = [];
@@ -846,6 +844,7 @@ class MultiSelectTreeWidget extends BaseWidget<
     return {
       value: `{{this.selectedOptionValues}}`,
       isValid: `{{(()=>{${derivedProperties.getIsValid}})()}}`,
+      flattenedOptions: `{{(()=>{${derivedProperties.getFlattenedOptions}})()}}`,
       selectedOptionValues: `{{(()=>{${derivedProperties.getSelectedOptionValues}})()}}`,
       selectedOptionLabels: `{{(()=>{${derivedProperties.getSelectedOptionLabels}})()}}`,
     };
@@ -866,12 +865,6 @@ class MultiSelectTreeWidget extends BaseWidget<
     };
   }
 
-  // to avoid calling flat every time
-  componentDidMount() {
-    const flattenedOptions = flat(this.props.options ?? []);
-    this.props.updateWidgetMetaProperty("flattenedOptions", flattenedOptions);
-  }
-
   componentDidUpdate(prevProps: MultiSelectTreeWidgetProps): void {
     if (
       xor(this.props.defaultOptionValue, prevProps.defaultOptionValue).length >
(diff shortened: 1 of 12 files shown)


app/client/src/widgets/MultiSelectTreeWidget/widget/index.tsx
```diff
@@ -872,20 +880,15 @@ class MultiSelectTreeWidget extends BaseWidget<
     ) {
       this.props.updateWidgetMetaProperty("isDirty", false);
     }
+
+    if (!equal(prevProps.options, this.props.options)) {
+      const flattenedOptions = flat(this.props.options ?? []);
+      this.props.updateWidgetMetaProperty("flattenedOptions", flattenedOptions);
```
ashit-rath, line 886: Can we make `flattenedOptions` as derived property?

app/client/src/widgets/WidgetUtils.ts
```diff
@@ -589,3 +589,27 @@ export function composePropertyUpdateHook(
...
+  children?: DropdownOption[];
+}
+
+/*
+ * Helps flatten nested Array of objects
+ *  Array -> Object { value, label,  children : Array -> Object { value, label } }
+ * This would be flattened to Array -> { value, label } , { value, label }
+ */
+
+export const flat = (array: DropdownOption[]) => {
```
ashit-rath, line 606: I think there is a lodash function to flatten which you can use in the derived properties.

app/client/src/widgets/SingleSelectTreeWidget/widget/derived.js
```diff
@@ -0,0 +1,49 @@
...
+      options,
+      (option) =>
+        option.label === label && option.value === props.selectedOptionValue,
+    );
+    if (labelIndex === -1) {
+      if (
+        !_.isNil(props.selectedOptionValue) &&
+        props.selectedOptionValue !== ""
+      ) {
+        const selectedOption = _.find(
```
ashit-rath, line 34: Why are we not directly finding the option which has the `value` same as `pops.selectedOptionValue` and return the label?

app/client/src/widgets/WidgetUtils.ts
Tooluloope: Not sure it would work for this use case. What youre refering to is that `Flatten` or `flattenDeep`

app/client/src/widgets/SingleSelectTreeWidget/widget/derived.test.ts
```diff
@@ -0,0 +1,132 @@
...
+        _,
+      );
+
+      expect(isValid).toBeTruthy();
+    });
+
+    it("return true when isRequired true and selectedOptionValue is not empty", () => {
+      const isValid = derivedProperty.getIsValid(
+        {
+          selectedOptionValue: "GREEN",
```
ashit-rath, line 49: This can also be a number right? If yes, then can we add cases for 0 and 1 as they sort of represent falsey and truthy values

ashit-rath (approved):

shadabbuchh:
All the reported issues pertaining to `TreeSelect` & `MultiTreeSelect` widgets are fixed and working fine.
https://www.loom.com/share/72b543e08d984647a758e168da972c0e

ashit-rath (approved):

PR #16438, opened 2022-09-01
feat: List v2 POC - Override base widgets default API

### LINKED ISSUE #15778: List V2 POC: Override widget primary methods
As the meta widgets are not part of the main canvas widgets, the original widget api's won't work normally.
In order to make sure the methods like `updateWidgetProperty` / `updateWidgetMetaProperty` / `executeAction` work properly, we need to override them.

ashit-rath:
## Description
Meta widget's APIs such as `updateWidgetProperty`/`executeAction` would get called on the meta widgets properties. Ideally any sort of such updates should be propagated to the actual widgets so that it gets cloned to the reset of the meta widgets present in different rows of the list widget.

In order to override, the EditorContext's values are overridden and that in turn would modify the behaviour in the `BaseWidget` of every meta widget.

The `MetaWidgetContextProvider` can be used in any widget to override any child widgets API. Only those methods that are passed to `MetaWidgetContextProvider` are overridden the rest stay the same as the original API methods.

Fixes #15778 

## Type of change

## Checklist:

### DIFF
--- a/app/client/src/widgets/ListWidgetV2/widget/index.tsx
+++ b/app/client/src/widgets/ListWidgetV2/widget/index.tsx
@@ -748,7 +748,7 @@ class ListWidget extends BaseWidget<
 
   onClientPageChange = (page: number) => this.setState({ page });
 
-  meta_executeAction = (triggerPayload: ExecuteTriggerPayload) => {
+  overrideExecuteAction = (triggerPayload: ExecuteTriggerPayload) => {
     const { id: metaWidgetId } = triggerPayload?.source || {};
 
     if (metaWidgetId) {
@@ -774,7 +774,7 @@ class ListWidget extends BaseWidget<
     }
   };
 
-  meta_batchUpdateWidgetProperty = (
+  overrideBatchUpdateWidgetProperty = (
     metaWidgetId: string,
     updates: BatchPropertyUpdatePayload,
     shouldReplay: boolean,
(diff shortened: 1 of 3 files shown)


app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -714,6 +748,79 @@ class ListWidget extends BaseWidget<
...
+    const { id: metaWidgetId } = triggerPayload?.source || {};
+
+    if (metaWidgetId) {
+      const metadata = this.metaWidgetsReferenceData[metaWidgetId];
+      const { listData = [] } = this.props;
+
+      const globalContext = {
+        currentIndex: metadata.index,
+        currentItem: listData[metadata.index],
+        ...triggerPayload.globalContext,
```
Tooluloope, line 761: is there a reason why we are adding this to globalContext when it already exist in `triggerPayload`

Tooluloope (approved):

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -714,6 +748,79 @@ class ListWidget extends BaseWidget<
 
   onClientPageChange = (page: number) => this.setState({ page });
 
+  meta_executeAction = (triggerPayload: ExecuteTriggerPayload) => {
```
SatishGandham, line 751: We don't use _ in our function names anywhere.

ashit-rath: Agreed but having the name `metaExecuteAction` or `executionActionMeta` seemed very confusing.

PR #16488, opened 2022-09-02
chore: bump ds version

tanvibhakta:
update design system version to stable-20


shadabbuchh:
The colour issue for the radio buttons is fixed in this PR.

albinAppsmith (approved):

PR #16501, opened 2022-09-03
fix: remove RapidAPI plugin src code from API server

### LINKED ISSUE #16440: [Task]: Remove RapidAPI plugin code from the API server

### SubTasks

- At the moment we do not have plans to support RapidAPI integration and hence it is necessary to remove the related plugin code from the API server so that it doesn't lead to tech debt or force developers to consider its presence.

sumitsum:
## Description
- Remove RapidAPI plugin src code from the API server since we are not supporting the plugin at the moment. 
- This PR does the following: 
  - mark the plugin as `deleted` in the `plugin` collection.
  - mark the plugin as `deleted` in the `workspace` collection.
  - mark the datasources related to the plugin as `deleted`.
  - mark the related actions as `deleted`
  - remove the plugin from the `pom.xml` file of `appsmith-plugins` module
  - when reading the list of installed plugins in a workspace only select those plugins that are not marked as `deleted`
  - remove code that adds the plugin to the `plugin` collection
- Please note that everywhere only `soft` delete happens during the data migration phase. This is done for two reasons: 
  - it makes it possible to recover older state since data would still exist.
  - no change in indexing in case any index has been built upon any collection.

Fixes #16440 

## Type of change
- Tech debt

## How Has This Been Tested?
- Manual
- Automated test does not seem feasible for this change.

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
```diff
@@ -288,18 +288,6 @@ public void initialPlugins(MongockTemplate mongoTemplate) {
         } catch (DuplicateKeyException e) {
             log.warn("mongo-plugin already present in database.");
         }
-
```
mohanarpit, line 291: As a principle, we shouldn't modify older database changelogs unless necessary. Is removing this code required?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
```diff
@@ -2435,4 +2436,104 @@ public void updateBadThemeState(MongockTemplate mongockTemplate, @NonLockGuarded
...
+    @ChangeSet(order = "035", id = "delete-rapid-api-plugin-related-items", author = "")
+    public void deleteRapidApiPluginRelatedItems(MongockTemplate mongockTemplate) {
+        Plugin rapidApiPlugin = mongockTemplate.findOne(query(where("packageName").is("rapidapi-plugin")),
+                Plugin.class);
+
+        if (rapidApiPlugin == null) {
+            return;
+        }
+
+        softDeleteAllRapidApiActions(rapidApiPlugin, mongockTemplate);
```
nidhi-nair, line 2448: Nit: Can we make this into a generic method so migrations for deleting plugins are easier in the future?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
```diff
@@ -2435,4 +2436,104 @@ public void updateBadThemeState(MongockTemplate mongockTemplate, @NonLockGuarded
...
+        softDeleteAllRapidApiDatasources(rapidApiPlugin, mongockTemplate);
+        softDeleteRapidApiPluginFromAllWorkspaces(rapidApiPlugin, mongockTemplate);
+        softDeletePlugin(rapidApiPlugin, mongockTemplate);
+    }
+
+    private void softDeleteRapidApiPluginFromAllWorkspaces(Plugin rapidApiPlugin, MongockTemplate mongockTemplate) {
+        Query queryToGetNonDeletedWorkspaces = new Query(where("deleted").ne(true));
```
nidhi-nair, line 2455: If we are doing soft deletes for everything, would it makes sense to perform these migrations for soft deleted workspaces as well?

nidhi-nair (commented):
Minor comments

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
sumitsum: @mohanarpit agreed, I have removed this change. Initially I thought that maybe we should stop the installation of the plugin itself. However, removing the code from database changelog does seem like an anti-pattern.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
sumitsum: Done.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
sumitsum: Done.

nidhi-nair (dismissed):

sumitsum:
> @sumitsum just noticed that there are a couple of RapidAPI related properties in the `application.properties` file. Could you please check if we need those?

@nidhi-nair thanks for pointing this out. I have updated the PR to remove the related properties.

nidhi-nair (approved):

PR #16507, opened 2022-09-04
chore: update-bad-theme should only run once.

trishaanand:
Update bad theme migration should only run once.


trishaanand:
Cancelled ok-to-test because it wouldn't test for this code change

mohanarpit (approved):

PR #12979, opened 2022-04-18
fix: handle dropdown flickers

techbhavin:
## Description

> fixed flickers issue on SelectTree widget while clicking on the default selection area and filter has text entered

Fixes #12714

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented on my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

## Test coverage results :test_tube:
<details><summary>:white_circle: Total coverage has not changed</summary>

    // Code coverage diff between base branch:release and head branch: fix/issue-12714 
Status | File | % Stmts | % Branch | % Funcs | % Lines 
 -----|-----|---------|----------|---------|------ 
 :green_circle: | total | 56.45 **(-0.01)** | 37.92 **(0.01)** | 36.17 **(0)** | 56.68 **(0)**
 :green_circle: | app/client/src/utils/autocomplete/TernServer.ts | 52.94 **(0.23)** | 41.67 **(0.84)** | 36.21 **(0)** | 56.99 **(0.25)**
 :red_circle: | app/client/src/widgets/SingleSelectTreeWidget/component/index.tsx | 15.91 **(-0.56)** | 0 **(0)** | 0 **(0)** | 17.72 **(-0.7)**</details>

PR #16542, opened 2022-09-05
fix:allow clear button in tree select widget

### LINKED ISSUE #16522: [Bug]: The "x" icon in TreeSelect widget is visible even when no options are selected

### Description

On enabling the `Allow Clearing Value` property we get the `x` icon in the widget which ideally should be visible only when there is any option selected in the widget. But right now we see it even when there is no option selected. Also when we set an `onOptionChange` event and click on the `x` button, the event gets triggered.

**Note:** This bug is present only in `Release` and not in `Production`, hence marking it as `Regression`.

https://www.loom.com/share/7662751f8550499cbe65b86a7fa624ca

### Steps To Reproduce

1. D&D a TreeSelect widget.
2. Enable the `Allow Clearing Value` property.
3. Now click on the `x` icon to clear the selected option.

**Expected:** The `x` icon should disappear once the selected option is cleared.

### Public Sample App

_No response_

### Version

Cloud

Tooluloope:
## Description
Fix Allow Clear button in TreeSelect

Fixes #16522

## Type of change

## How Has This Been Tested?

- Test A
- Test B

## Checklist:


app/client/src/widgets/SingleSelectTreeWidget/component/index.tsx
```diff
@@ -304,7 +308,7 @@ function SingleSelectTreeComponent({
           treeDefaultExpandAll={expandAll}
           treeIcon
           treeNodeFilterProp="label"
-          value={filter ? "" : value} // value should empty when filter value exist otherwise dropdown flickers #12714
+          value={value}
```
keyurparalkar, line 311: @Tooluloope can we check if the issue: https://github.com/appsmithorg/appsmith/pull/12979 occurs with this change?

Tooluloope: Also tested in DP

keyurparalkar (dismissed):
LGTM!

chandannkumar:
@Tooluloope 
- Placeholder is not visible on the Tree select widget. 
https://www.loom.com/share/1ba8d0e1c341411fb45159d8c9809b77

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/Widgets/Select/Single_Select_Tree_spec.js
```diff
@@ -8,107 +8,129 @@ describe("Single Select Widget Functionality", function() {
...
-    cy.treeSelectDropdown("Light Blue");
-    // Check if isDirty is set to true
-    cy.get(".t--widget-textwidget").should("contain", "true");
-    // Change defaultText
-    cy.openPropertyPane("singleselecttreewidget");
-    cy.updateCodeInput(".t--property-control-defaultselectedvalue", "RED");
-    // Check if isDirty is reset to false
-    cy.get(".t--widget-textwidget").should("contain", "false");
-  });
+  // it("1. Check isDirty meta property", function() {
```
keyurparalkar, line 11: @Tooluloope Why are we commenting these test cases? Can we remove the comments?

Tooluloope: Forgot to revert after writing the tests, thanks for catching that

keyurparalkar (approved):

PR #16591, opened 2022-09-06
feat: added WIDGET_DRAG event.

### LINKED ISSUE #15345: [Task]: Misfiring WIDGET_DRAG analytics event

### SubTasks

The `WIDGET_DRAG` frontend analytics event has no data collected for the past 30 days and has hardly any for the last 6 months. Needs to be looked at.

yaldram:
## Description

- Added a `WIDGET_DRAG` event, when the user finishes moving the widget on the canvas and places it.

Fixes #15345

## Type of change

## How Has This Been Tested?

- Not tested, might need to test it either on release or production

## Checklist:


yaldram:
When we move a widget, we have the following scenarios - 
- We can move a single widget.
- We can move multiple widgets.
- We can move the widget / widgets on the canvas or inside a container.

Therefore, in the analytics event we have the following information: - 
1. widgets - array field with `widget.type` & `widget.widgetName`.
2. multiple - boolean field to indicate if we are moving a single widget or multiple widgets.
3. mainCanvas - boolean field to indicate if we are moving the widgets on the canvas or inside a container widget.

vuiets:
Hey great coverage for the event here guys 👏 Thanks for the detailed write up too.

Instead of `mainCanvas` can we add a property to look at the source and destination of the dragged widget? Look for different terms instead of `source` though since its a meta property that exists to tell which widget was dragged, let this remain as is as multiple widget-related events use this convention.

rahulramesha:
@vuiets by source and destinations do you mean the source canvas to destination canvas, eg, from mainCanvas to Container widget?

yaldram:
I was thinking instead of the `mainCanvas` as a boolean field can we have a filed called containerWidget which will have the name of the container widget. 
 Also, I don't think source and destination might help us, like during form alignment or widget alignment my co-ordinates don't change drastically.

app/client/src/sagas/CanvasSagas/DraggingCanvasSagas.ts
```diff
@@ -255,6 +257,26 @@ function* moveWidgetsSaga(
...
     }
     yield put(updateAndSaveLayout(updatedWidgetsOnMove));
+    AnalyticsUtil.logEvent("WIDGET_DRAG", {
+      widgets: draggedBlocksToUpdate.map((block) => {
+        const widget = updatedWidgetsOnMove[block.widgetId];
+
+        const parentId = block.updateWidgetParams.payload.parentId;
+        const newParentId = block.updateWidgetParams.payload.newParentId;
+
+        const parentWidget = allWidgets[parentId].widgetName;
```
rahulramesha, line 267: @yaldram Instead of widgetName, it would be better if we log widgetType, to better understand which types of widget are the parents.
Also to point out, parent Widget of all widgets will always be a CANVAS_WIDGET. The ParentWidget's parent will give you the actual Type of widget (Except for MainCanvas)

Changes required,
1. Check if Parent Widget is main Canvas, if not log the Parent's ParentWidget Type. for both source and destination.
2. If one of them is main Canvas, their type is also CANVAS_WIDGET, instead log it as MAIN_CONTAINER
3. instead of mainCanvas parameter as boolean, it would be better to log a boolean to indicate if source and destination parentId are same.

cc: @vuiets

rahulramesha (changes requested):

app/client/src/sagas/CanvasSagas/DraggingCanvasSagas.ts
vuiets: Are you looking for any input from me here?

rahulramesha: @vuiets just wanted to keep you in loop for the logs Arsalan will be adding

vuiets: Got it, thanks @rahulramesha!

rahulramesha (approved):

PR #16596, opened 2022-09-07
chore: Fix Jest test Flakiness

### LINKED ISSUE #16525: [Task]: Resolve Flaky Jest test failures

### SubTasks

We were experiencing frequent failures for the following Jest tests:
- ColorPickerComponentV2.test.tsx
- IconSelectControl.test.tsx

Example failures:
- https://github.com/appsmithorg/appsmith/runs/8145823657?check_suite_focus=true
- https://github.com/appsmithorg/appsmith/runs/8170350211?check_suite_focus=true

aswathkk:
## Description

Fixes Flakiness for the following Jest tests:

#### ColorPickerComponentV2.test.tsx
The test was taking a long time and sometimes resulted in a time-out error since we were directly using the redux store rather than a mock store. Thanks to @Rishabh-Rathod for pointing this out.

#### IconSelectControl.test.tsx
Right after opening the Icon Selector, the component focuses on the search input. This is a behaviour from Blueprintjs and sometimes it takes more than 1000ms to focus on the search input. In such cases the Jest test fails since the default timeout for `waitFor` function is 1000ms. This is fixed by increasing the timeout to 3000ms.

Fixes #16525 

## Type of change

- Jest test updates

## Checklist:


app/client/src/components/propertyControls/IconSelectControl.test.tsx
```diff
@@ -74,14 +78,14 @@ describe("<IconSelectControl /> - Keyboard navigation", () => {
     await waitForElementToBeRemoved(screen.getAllByRole("list"));
   });
 
-  it("Pressing '{ArrowDown}' while search is in focus should remove the focus", async () => {
+  it.only("Pressing '{ArrowDown}' while search is in focus should remove the focus", async () => {
```
yaldram, line 81: You can remove the `.only` from here.

yaldram (approved):

app/client/src/components/propertyControls/IconSelectControl.test.tsx
aswathkk: My bad! 🤦 
Used it to test only that test case. Fixed it.

Rishabh-Rathod:
Hey @aswathkk, 
Can you please share the difference in the time it used to take before and after this change for both the jest test to complete?

mohanarpit:
@aswathkk Is this a performance problem in these components? Also, I don't see either component in the design system. Have these components not been migrated yet? 

CC: @albinAppsmith @tanvibhakta

aswathkk:
@Rishabh-Rathod 

| Test Case | Time in release branch | Time in this branch |
|-------|--------------------------------|---------------------------|
| ColorPickerComponentV2.test.tsx | ~50s | ~25s |
| IconSelectControl.test.tsx | ~10s | ~13s |

@mohanarpit This is not a performance problem. In the case of ColorPickerComponentV2's test, it is an issue with the way we use the component for the testing. We were passing the entire store which causes the test to be slow.
In the case of IconSelectControl, it is an unpredictable delay during search input autofocus (which is internally done by Blueprintjs) that makes the test to be flaky.

albinAppsmith (approved):

Rishabh-Rathod (approved):
LGTM. 

As we noticed 50% save in time when using the mock store. I think we should do it for all such tests in the future. 
I have created a task for it https://github.com/appsmithorg/appsmith/issues/16636 this could be picked file by file by multiple folks.

PR #16619, opened 2022-09-07
feat: Add a service layer to get AppsmithType from client-type (#16508)

### LINKED ISSUE #16508: [Task]: Create a service layer to get AppsmithType from the given client-side data type and value

## Description
Identification of the correct AppsmithType from the given client-side data type and the evaluated value is depicted in this [Miro board](https://miro.com/app/board/uXjVPfSzqXg=/). 

Identification of AppsmithType

subrata71:
This PR takes care of the following things
- It creates a service layer to get the target AppsmithType from the client-side data type and the evaluated value
- This service layer is currently not consumed by any plugins
- A full JUnit test suit for MySQL specific data types

## Description

Fixes #16508 

## Type of change

## How Has This Been Tested?

- JUnit

## Checklist:


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/ce/DataTypeServiceCEImpl.java
```diff
@@ -0,0 +1,52 @@
...
+        defaultAppsmithTypes.put(ClientDataType.NULL, List.of(new NullType()));
+
+        defaultAppsmithTypes.put(ClientDataType.ARRAY, List.of(new ArrayType()));
+
+        defaultAppsmithTypes.put(ClientDataType.BOOLEAN, List.of(new BooleanType()));
+
+        defaultAppsmithTypes.put(ClientDataType.NUMBER, List.of(
+           new IntegerType(),
+           new LongType(),
+           new FloatType(),
```
nidhi-nair, line 22: Weren't we going to skip floats?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/ce/DataTypeServiceCEImpl.java
```diff
@@ -0,0 +1,52 @@
...
+import com.appsmith.external.datatypes.*;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class DataTypeServiceCEImpl implements DataTypeServiceCE {
+     final Map<ClientDataType, List<AppsmithType>> defaultAppsmithTypes = new HashMap<>();
+
+    public DataTypeServiceCEImpl() {
```
nidhi-nair, line 12: How would plugins use their types in this service?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/ce/DataTypeServiceCEImpl.java
subrata71: Yes, yes. Sorry, I didn't get any definitive guideline for that. Thought you are not 100% convinced. :-| 
Anyway, will update the PR by removing the `FloatType`.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/ce/DataTypeServiceCEImpl.java
subrata71: Plugins can pass their types by invoking the following function:

`AppsmithType appsmithType = DataTypeStringUtils.stringToKnownAppsmithTypeConverter(clientDataType, value, pluginSpecificTypes);`

Where plugin will define its supported data types in the respective plugin class this way

```
private static Map<ClientDataType, List<AppsmithType>> pluginSpecificTypes;

    static {
        pluginSpecificTypes = new HashMap<>();
        pluginSpecificTypes.put(ClientDataType.NULL, List.of(new NullType()));

        pluginSpecificTypes.put(ClientDataType.BOOLEAN, List.of(new MySQLBooleanType()));

        pluginSpecificTypes.put(ClientDataType.NUMBER, List.of(
                new IntegerType(),
                new LongType(),
                new DoubleType(),
                new BigDecimalType()
        ));

        pluginSpecificTypes.put(ClientDataType.OBJECT, List.of(
                new JsonObjectType(),
                new StringType()
        ));

        pluginSpecificTypes.put(ClientDataType.STRING, List.of(
                new TimeType(),
                new MySQLDateType(),
                new MySQLDateTimeType(),
                new StringType()
        ));
```

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/ce/DataTypeServiceCEImpl.java
subrata71: Removed FloatType and updated the PR.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/DataTypeStringUtils.java
```diff
@@ -58,6 +62,23 @@ public class DataTypeStringUtils {
...
+    /**
+     *
+     * @param clientDataType the client-side identified data type e.g. NULL, ARRAY, BOOLEAN, OBJECT
+     * @param value the evaluated value of the binding parameter
+     * @param args varargs will be used for passing optional plugin-specific types. In case args is not sent then the DataTypeService will
+     *             consider the default server-side data types for the identification of data type from the given client-side data type and
+     *             the evaulated value
+     * @return
+     */
+    public static AppsmithType stringToKnownAppsmithTypeConverter(ClientDataType clientDataType, String value, Object... args) {
```
nidhi-nair, line 76: Can we avoid creating an intermediate method like this one? Let's isolate this functionality to just the `DataTypeService` instead?

nidhi-nair, line 76: Also, could you please share your reasoning as to the choice of variable argument `args` instead of a specific one?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/ce/DataTypeServiceCEImpl.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Override
+    public AppsmithType getAppsmithType(ClientDataType clientDataType, String value, Map<ClientDataType, List<AppsmithType>> pluginSpecificTypes) {
+        for (AppsmithType currentType : pluginSpecificTypes.get(clientDataType)) {
+            if (currentType.test(value)) {
+                return currentType;
+            }
+        }
+        //TODO: Send some insights to Mixpanel as this case isn't expected at all
+        return null;
```
nidhi-nair, line 56: Shall we default to String here as the base type, just so we always have a backup expectation and avoid NPEs?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/DataTypeServiceImpl.java
```diff
@@ -0,0 +1,18 @@
...
+import com.appsmith.external.services.ce.DataTypeServiceCEImpl;
+
+public class DataTypeServiceImpl extends DataTypeServiceCEImpl implements DataTypeService {
+    private static DataTypeServiceImpl instance;
+
+    private DataTypeServiceImpl() {
+        super();
+    }
+
+    public static synchronized DataTypeServiceImpl getInstance() {
```
nidhi-nair, line 12: I'm a little iffy about having all our integration services be singletons. What do you think about extending the Spring context to our other modules?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/ce/DataTypeServiceCEImpl.java
```diff
@@ -0,0 +1,58 @@
+package com.appsmith.external.services.ce;
+
+import com.appsmith.external.datatypes.*;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class DataTypeServiceCEImpl implements DataTypeServiceCE {
+     final Map<ClientDataType, List<AppsmithType>> defaultAppsmithTypes = new HashMap<>();
```
nidhi-nair, line 10: Could you please add a comment here explaining the structure of this map to a consumer of the datatype service?

nidhi-nair, line 10: Could you please add a comment here explaining the structure of this map to a consumer of the datatype service?

nidhi-nair (changes requested):
@subrata71 sorry again for taking time on this PR review. :)

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/ce/DataTypeServiceCEImpl.java
nidhi-nair: We could create another data type to server this purpose as well. One which would always test to true, and does no transformation.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/DataTypeServiceImpl.java
subrata71: We can avoid singleton behavior for a class like DataTypeService. This fits well as a util class. @nidhi-nair

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/DataTypeStringUtils.java
subrata71: If the DataTypeService is written as a util class then we don't need to use DataTypeStringUtils class anymore. What do you think?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/ce/DataTypeServiceCEImpl.java
subrata71: I will add a comment on this method but let's decide first whether we should implement this as a util class or not.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/ce/DataTypeServiceCEImpl.java
subrata71: I kept StringType as the fallback data type for the Object type. Do you think it should also be the fallback data type for the rest of the data types? If so, then we can do like this

```
@Override
public AppsmithType getAppsmithType(ClientDataType clientDataType, String value, Map<ClientDataType, List<AppsmithType>> pluginSpecificTypes) {
        for (AppsmithType currentType : pluginSpecificTypes.get(clientDataType)) {
            if (currentType.test(value)) {
                return currentType;
            }
        }
        //Send StringType as the fallback data type
        return new StringType();
}
```

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/DataTypeStringUtils.java
nidhi-nair: Hmm ... it does look like we'll only ever have stateless utilities here. Plus there's a low chance of needing to depend on another service. I think we can do what you suggest here. :)

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/ce/DataTypeServiceCEImpl.java
nidhi-nair: Could you please give an example of when converting something identified as an object by JavaScript will be useful to convert into a string?

I suggested a separate type so that we're able to distinguish between data that has been genuinely identified as a string, and data that seems to be in error state but we want to treat as a string. That way if any plugin wants to handle the error scenario differently, they'll be able to.

subrata71: I don't have a very good example but let's consider this

The type of new Date() in javascript is of object type. And for object type the respective Appsmith type is JsonObjectType but it's not a valid JSON, right? 
In that case if the fallback datatype were StringType then it could take over.

Does it make any sense to you? @nidhi-nair 

That's a nice idea to keep a separate fallback data type for the case you mentioned. I am going to name it "FallbackDataType"

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/services/DataTypeServiceImpl.java
nidhi-nair: We seem to be talking about the same thing in two different [threads](https://github.com/appsmithorg/appsmith/pull/16619#discussion_r971465406). :P

Yes, let's go ahead with a utility class.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/DataTypeService.java
```diff
@@ -0,0 +1,93 @@
+package com.appsmith.external.helpers;
+
+import com.appsmith.external.datatypes.*;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class DataTypeService {
```
nidhi-nair, line 9: Nit: Let's call this Utils for consistent guidelines?

subrata71: Done. @nidhi-nair

nidhi-nair (approved):

PR #16632, opened 2022-09-08
fix: Use JSON to serialize sessions for OAuth

sharat87:
We are already using JSON serialization for form based sessions, but, apparently this is failing for OAuth based sessions.

The next time we upgrade Spring Boot, say, even to the current version of 2.7.3, all users who logged in with OAuth, will need to be logged out. This fix means that anyone who logged in with OAuth _after this fix is out_, don't have to be logged out.


app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/RedisConfig.java
```diff
@@ -73,18 +76,38 @@ ReactiveRedisTemplate<String, Object> reactiveRedisTemplate(ReactiveRedisConnect
...
             if (t instanceof SecurityContext) {
-                final UserSession session = UserSession.fromToken(((SecurityContext) t).getAuthentication());
+                final UserSessionDTO session = UserSessionDTO.fromToken(((SecurityContext) t).getAuthentication());
                 final byte[] bytes = jsonSerializer.serialize(session);
                 return bytes == null ? null : ByteUtils.concat(SESSION_DATA_PREFIX, bytes);
+
+            } else if ((t instanceof Map)) {
+                final Map<?, ?> data = (Map<?, ?>) t;
+                if (data.size() == 1
+                        && (data.get("google") instanceof OAuth2AuthorizedClient || data.get("github") instanceof OAuth2AuthorizedClient)) {
```
mohanarpit, line 97: Can we use enums like `LoginSource.GOOGLE` instead of raw strings?

sharat87:
@mohanarpit @trishaanand, can I get a review/approval on this please? Thanks! :)

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/RedisConfig.java
```diff
@@ -73,18 +77,38 @@ ReactiveRedisTemplate<String, Object> reactiveRedisTemplate(ReactiveRedisConnect
...
             if (t instanceof SecurityContext) {
-                final UserSession session = UserSession.fromToken(((SecurityContext) t).getAuthentication());
+                final UserSessionDTO session = UserSessionDTO.fromToken(((SecurityContext) t).getAuthentication());
                 final byte[] bytes = jsonSerializer.serialize(session);
                 return bytes == null ? null : ByteUtils.concat(SESSION_DATA_PREFIX, bytes);
+
+            } else if ((t instanceof Map)) {
+                final Map<?, ?> data = (Map<?, ?>) t;
+                if (data.size() == 1
+                        && (data.get(LoginSource.GOOGLE.name().toLowerCase()) instanceof OAuth2AuthorizedClient || data.get(LoginSource.GITHUB.name().toLowerCase()) instanceof OAuth2AuthorizedClient)) {
```
trishaanand, line 98: QQ : What would be the expectation here if the login is via OIDC/SAML (both being OAuth2 integrations on Appsmith)

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/RedisConfig.java
```diff
@@ -73,18 +77,38 @@ ReactiveRedisTemplate<String, Object> reactiveRedisTemplate(ReactiveRedisConnect
...
+                    final String firstAndOnlyKey = (String) data.keySet().iterator().next();
+                    final OAuth2AuthorizedClient client = (OAuth2AuthorizedClient) ((Map<?, ?>) t).get(firstAndOnlyKey);
+                    final OAuth2AuthorizedClientDTO dto;
+                    try {
+                        dto = OAuth2AuthorizedClientDTO.fromOAuth2AuthorizedClient(firstAndOnlyKey, client);
+                    } catch (Exception e) {
+                        e.printStackTrace();
```
trishaanand, line 105: Super minor nit : Could we use Appsmith Exceptions here?

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/RedisConfig.java
```diff
@@ -94,13 +118,24 @@ public byte[] serialize(Object t) {
...
+                final UserSessionDTO session = jsonSerializer.deserialize(data, UserSessionDTO.class);
 
                 if (session == null) {
                     throw new IllegalArgumentException("Could not deserialize user session, got null");
                 }
 
                 return new SecurityContextImpl(session.makeToken());
+
+            } else if (ByteUtils.startsWith(bytes, OAUTH_CLIENT_PREFIX)) {
+                final byte[] data = Arrays.copyOfRange(bytes, "appsmith-oauth-client:".getBytes().length, bytes.length);
```
trishaanand, line 130:
```suggestion
                final byte[] data = Arrays.copyOfRange(bytes, OAUTH_CLIENT_PREFIX.length, bytes.length);
```

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/OAuth2AuthorizedClientDTO.java
```diff
@@ -0,0 +1,164 @@
...
+                )
+        );
+    }
+
+    public OAuth2AuthorizedClient makeOAuth2AuthorizedClient() {
+        OAuth2AccessToken.TokenType tokenType;
+        if (OAuth2AccessToken.TokenType.BEARER.getValue().equals(accessToken.tokenType)) {
+            tokenType = OAuth2AccessToken.TokenType.BEARER;
+        } else {
+            throw new IllegalArgumentException("Could not deserialize OAuth2AuthorizedClient, unknown token type: " + accessToken.tokenType);
```
trishaanand, line 119: Minor nit : Possible to use Appsmith Exceptions? Also, should we do a handling for this scenario?

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/OAuth2AuthorizedClientDTO.java
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        System.out.println("!!!!! registrationId: " + clientRegistration.registrationId);
```
trishaanand, line 122: Minor minor nit : If required, could we format this? Else, if not required, we can also remove this line.
```suggestion
```

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/OAuth2AuthorizedClientDTO.java
```diff
@@ -0,0 +1,164 @@
...
+                        rt.getTokenValue(),
+                        rt.getIssuedAt().toString(),
+                        rt.getExpiresAt().toString()
+                )
+        );
+    }
+
+    public OAuth2AuthorizedClient makeOAuth2AuthorizedClient() {
+        OAuth2AccessToken.TokenType tokenType;
+        if (OAuth2AccessToken.TokenType.BEARER.getValue().equals(accessToken.tokenType)) {
```
trishaanand, line 116: QQ : Do we need to add null handling for accessToken before accessing token type?

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/OAuth2AuthorizedClientDTO.java
```diff
@@ -0,0 +1,164 @@
...
+            tokenType = OAuth2AccessToken.TokenType.BEARER;
+        } else {
+            throw new IllegalArgumentException("Could not deserialize OAuth2AuthorizedClient, unknown token type: " + accessToken.tokenType);
+        }
+
+        System.out.println("!!!!! registrationId: " + clientRegistration.registrationId);
+
+        final OAuth2AuthorizedClient client = new OAuth2AuthorizedClient(
```
trishaanand, line 124: QQ & a nit : Do we need to add a sanity check for null pointer handling for all the nested fields access below?

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/OAuth2AuthorizedClientDTO.java
```diff
@@ -0,0 +1,164 @@
...
+                        Set.copyOf(accessToken.scopes)
+                ),
+                refreshToken == null ? null : new OAuth2RefreshToken(
+                        refreshToken.tokenValue,
+                        Instant.parse(refreshToken.issuedAt),
+                        Instant.parse(refreshToken.expiresAt)
+                )
+        );
+
+        System.out.println("!!!!! client registrationId: " + client.getClientRegistration().getRegistrationId());
```
trishaanand, line 159: Nit : Could we remove it if not required, or format it otherwise?
```suggestion
```

trishaanand (commented):
Super minor nits.

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/RedisConfig.java
sharat87: To be honest, no idea yet. But after seeing your comment, I'm now looking forward to investigate that, once this fix goes to EE. 🙂

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/OAuth2AuthorizedClientDTO.java
sharat87: Ah, the old frustrated dev message! Thanks for catching this!

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/OAuth2AuthorizedClientDTO.java
sharat87: I tried to handle it, but, I couldn't come up with a good recovery strategy. Ideally, if there's an error with deserialisation of a session, I'm thinking, a good experience is to log that user out, so they can login again and create a new session for themselves. But that feels like a separate fix and a separate PR. Does that sound okay?

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/OAuth2AuthorizedClientDTO.java
sharat87: I'm a little hesitant to have Appsmith exceptions here. This is not a direct part of any CRUD API or any other API. If this fails, I expect users to see "Internal Server Error" only, and not "Session deserialisation error" or something. With such an error message, we are giving away slightly internal information, but the user can't do anything with this information other than reach out to support. So, I'm a little inclined to keeping it this way. What do you think?

sharat87:
@mohanarpit @trishaanand, can one of you take a second look when you get a chance please?

trishaanand (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/LoginSource.java
```diff
@@ -9,6 +9,8 @@
 public enum LoginSource {
     GOOGLE, FORM, GITHUB, OIDC;
 
+    public static final Set<LoginSource> oauthSources = Set.of(GOOGLE, GITHUB, OIDC);
```
trishaanand, line 12: Lets make the LoginSource file match as it is in EE (with the same LoginSource values) and add KEYCLOAK to oauthSources here itself.

trishaanand (approved):
Minor nit.

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/LoginSource.java
sharat87: You sure? I was wondering why even OIDC was listed here, but shrugged it off. If anything, I thought you'd ask me to _remove_ OIDC here instead of adding the other missing one. 😅

trishaanand: Nah. Instead of removing OIDC, lets just bring KEYCLOAK and keep this file as is across both.

sharat87: Done.

trishaanand (approved):

PR #16667, opened 2022-09-09
feat: Support running NGINX for development without Docker

sharat87:
This PR enhances the `start-https.sh` script to add the following capabilities:

1. Start NGINX without Docker, but support Docker too.
2. Start dev server without https, if needed.
3. Ability to set a custom env file.

However, this script should be backwards compatible with documented usages of the previous version.

To try before this PR is merged, get the script without having to checkout to this branch with:

```sh
curl -o start-https.sh 'https://raw.githubusercontent.com/appsmithorg/appsmith/feat/dev-nginx-without-docker/app/client/start-https.sh'
```

After that, do a `./start-https.sh -h` to learn more.


riodeuno:
@sharat87 Could you help me understand the circumstances in which we will need this?

> Start dev server without https, if needed.

app/client/start-https.sh
```diff
@@ -1,106 +1,312 @@
...
+--without-docker: Start NGINX directly. Fail if NGINX is not installed.
+
+        If neither of the above are set, we check if NGINX is installed, and use that if yes, or Docker otherwise.
+
+--https: Require start with https.
+ --http: Require start with http.
+
+        If neither of the above ar set, then we check if mkcert is available, and use https if yes, or http otherwise.
+
+--env-file: Specify an alternate env file. Defaults to `.env` at the root of the project.
```
riodeuno, line 23: This is a nice enhancement. Thanks!

app/client/start-https.sh
```diff
@@ -1,106 +1,312 @@
...
+
+A single positionl argument can be given to set the backent server proxy address. Example:
```
riodeuno, line 25: A couple of typos here `positionl` and `backent`

app/client/start-https.sh
```diff
@@ -1,106 +1,312 @@
...
-    exit
+key_file="$working_dir/$domain-key.pem"
+cert_file="$working_dir/$domain.pem"
+
+if [[ -z ${use_https-} ]]; then
+    if type mkcert; then
+        use_https=1
+    else
+        echo 'SSL cert isn'\''t there, and `mkcert` is not installed either. Starting with http.' >&2
+        echo 'Please `brew install mkcert` and ru-run this script to start with https.' >&2
```
riodeuno, line 96: typo : `re-run`

sharat87:
@riodeuno yes, thanks for asking. I personally needed it once a few weeks ago when I was testing the redirect URLs generated by the backend server. I wanted to confirm if it is correctly producing `http://` or `https://` based on what the user is actually using. Further tests were when there was an extra NGINX in front of this NGINX and testing with that combination. There's a few others, but that's the idea.

Another situation I needed this was when a user reported an issue that things weren't working without HTTPS. Unfortunately, we all run, work and test with https turned on, but that's not true for all self-hosted users. A good many of them run without https. There should be an easy way for us to test without `https`. The problem turned out to be that some critical asset (TinyMCE or something, I don't recall) wasn't loading because of a mixed content error. :|

app/client/start-https.sh
```diff
@@ -1,106 +1,312 @@
...
-# cd docker && mkcert -install && mkcert "*.appsmith.com" && cd ..
-# If this returns a hash successfully, then you can access the application locally using https://dev.appsmith.com
+set -o errexit
+set -o nounset
+set -o pipefail
+if [[ -n ${TRACE-} ]]; then
+    set -o xtrace
+fi
+
+if [[ ${1-} == -h ]]; then
```
riodeuno, line 10: Finally a `-h` in our scripts!!  Can we also use `--help`, as I default to that most of the time.

sharat87: Good catch. Done. We've now got `-h`, `--help`, `help`, `h`, and even `----------help` for extra frustrated folks. 😄

riodeuno (dismissed):

riodeuno:
QQ: Will the pem files now be created with `dev.appsmith.com` as a domain by default instead of `wildcard.appsmith.com`?

sharat87:
@riodeuno, yes. I'm hoping that's okay? Since we just use it for one domain, figured we don't need wildcard magic here.

Thanks for the review! 😄

PR #16678, opened 2022-09-09
ci: Fix `build_args` to be `build-args` for Depot builds

kylegalbraith:
It looks like this was a typo carried over from merging in https://github.com/appsmithorg/appsmith/pull/16658. This allows the arg to actually be passed into the action going forward.

cc: @trishaanand @mohanarpit


mohanarpit (approved):

mohanarpit:
Thanks @kylegalbraith for the fix! I didn't notice this. :)

PR #16689, opened 2022-09-12
feat: parse CSV data using File Picker Widget

### LINKED ISSUE #15747: [Feature]: Option to parse CSV Data as per RFC 4180

### Summary

I am able to upload my CSV file via File picker as a text, and then use javascript snippet to parse the same to return an array of objects which i can feed it to my table widget.

This works fine for a simple CSV file. But if any of the data values in the CSV contains newline characters or commas, then the snippet being used returns incorrect data.

When creating a CSV file, need to follow certain standards as per RFC 4180.
Example would be, when using commas or newline characters within the data value, it needs to be enclosed in double quotes or so.

This can get fairly complex to parse via native javascript. Hence we need an alternative, whether it is a library function which we can invoke or a widget's capability to parse the data.

Tooluloope:
## Description

This feature helps to parse csv data to JSON which is usable on the platform.

This uses [`papaparse`
](https://www.npmjs.com/package/papaparse) to parse csv.

**Why I chose Papaparse**

- Still maintained and Highest weekly downloads
- Auto detect delimiter in csv
- Stream large files
- Header row support

Fixes #15747

## Type of change

## How Has This Been Tested?

- Test A
- Test B

## Checklist:


Tooluloope:
@aswathkk I made some further improvements to the `helperText`

aswathkk (commented):
@Tooluloope helperText related changes look good to me.
It would be great if you could include a test for it. It would prevent any future issues.

app/client/src/constants/PropertyControlConstants.tsx
```diff
@@ -47,7 +47,10 @@ export type PropertyPaneControlConfig = {
   id?: string;
   label: string | ((props: WidgetProps, propertyPath: string) => string);
   propertyName: string;
+  // Serves in the tooltip
   helpText?: string;
+  //Dynamic text serves below the property pane inputs
+  helperText?: ((props: any) => string) | string;
```
ashit-rath, line 53: Always avoid the use of `any`.  
Ideally it should be something like
`helperText?: ((props: TWidgetProps) => string) | string;`
where `TWidgetProps` is a generic that represents the widget properties so that when the helperText is defined, appropriate typings appear.

```ts
export type  PropertyPaneControlConfig<TWidgetProps extends WidgetProps = WidgetProps> = {
  ... // rest of typings
  helperText?: ((props: TWidgetProps) => string) | string;
}
```

app/client/src/constants/PropertyControlConstants.tsx
```diff
@@ -47,7 +47,10 @@ export type PropertyPaneControlConfig = {
   id?: string;
   label: string | ((props: WidgetProps, propertyPath: string) => string);
   propertyName: string;
+  // Serves in the tooltip
```
ashit-rath, line 50: Ideally we should rename this to `tooltipText` or rename the `helperText` to something else.
A widget developer as to come looking for this file to know the difference between the two or test out the code to know the difference.

app/client/src/widgets/FilePickerWidgetV2/widget/index.tsx
```diff
@@ -21,7 +21,14 @@ import { createGlobalStyle } from "styled-components";
...
 import CloseIcon from "assets/icons/ads/cross.svg";
 import { Colors } from "constants/Colors";
+import Papa from "papaparse";
 
+const CSV_ARRAY_LABEL = "Array (CSVs only)";
+const CSV_FILE_TYPE_REGEX = /.+(\/csv)$/;
+
+const isCSVFileType = (str: string) => CSV_FILE_TYPE_REGEX.test(str);
+
+type Result = string | Buffer | null;
```
ashit-rath, line 31: Add a new line after this. Separation of concern

app/client/src/widgets/FilePickerWidgetV2/widget/index.tsx
```diff
@@ -831,7 +866,11 @@ class FilePickerWidget extends BaseWidget<
               const newFile = {
                 type: file.type,
                 id: file.id,
-                data: reader.result,
+                data: this.csvParser(
+                  reader.result as Result,
```
ashit-rath, line 870: By default render.result type is `string | ArrayBuffer | null` 
Why are we casting it to `Result`?

app/client/src/widgets/FilePickerWidgetV2/widget/index.tsx
```diff
@@ -831,7 +866,11 @@ class FilePickerWidget extends BaseWidget<
               const newFile = {
                 type: file.type,
                 id: file.id,
-                data: reader.result,
+                data: this.csvParser(
```
ashit-rath, line 869: I don't think we should name this method as `csvParser`
It seems like we always expect csv and we parse it, unless we check the method.

Either you can rename this function to `parseUploadResult` to imply that we handle all the cases there or do your `fileDataType` check here and then call the `csvParser` on the result.

app/client/src/widgets/FilePickerWidgetV2/widget/index.tsx
```diff
@@ -959,6 +998,50 @@ class FilePickerWidget extends BaseWidget<
     );
   }
 
+  csvParser(result: Result, fileType: string, dataFormat: FileDataTypes) {
+    if (dataFormat !== FileDataTypes.Array) {
+      return result;
+    }
+
+    if (dataFormat === FileDataTypes.Array && !isCSVFileType(fileType)) {
```
ashit-rath, line 1006: No need to check for `Array` again as we are dismissing this in the first if condition and better club this and the `!result` check here.
It's always best to club all conditions for same return values.

app/client/src/widgets/FilePickerWidgetV2/widget/index.tsx
```diff
@@ -959,6 +998,50 @@ class FilePickerWidget extends BaseWidget<
...
+      return [];
+    }
+    if (!result) {
+      return [];
+    }
+
+    const data: any[] = [];
+    const errors: any[] = [];
+
+    function chunk(results: Papa.ParseStepResult<any>) {
```
ashit-rath, line 1016: Avoid `any`

app/client/src/widgets/FilePickerWidgetV2/widget/index.tsx
```diff
@@ -959,6 +998,50 @@ class FilePickerWidget extends BaseWidget<
...
+      if (results?.errors?.length) {
+        errors.push(...results.errors);
+      }
+      data.push(...results.data);
+    }
+
+    if (typeof result === "string") {
```
ashit-rath, line 1023: If result is not string should we return `[]`?

app/client/src/widgets/FilePickerWidgetV2/widget/index.tsx
```diff
@@ -959,6 +998,50 @@ class FilePickerWidget extends BaseWidget<
...
+
+        Papa.parse(result, config);
+
+        const endParsing = performance.now();
+
+        log.debug("### PARSING timing", `${endParsing - startParsing} ms`);
+
+        return data;
+      } catch (error) {
+        log.error(errors);
```
ashit-rath, line 1040: Also here do we need to return `[]`?

app/client/src/widgets/FilePickerWidgetV2/widget/index.tsx
```diff
@@ -959,6 +998,50 @@ class FilePickerWidget extends BaseWidget<
...
+        chunk,
+      };
+      try {
+        const startParsing = performance.now();
+
+        Papa.parse(result, config);
+
+        const endParsing = performance.now();
+
+        log.debug("### PARSING timing", `${endParsing - startParsing} ms`);
```
ashit-rath, line 1036: Lets have more specifics in the message as there are a lot of performance logs. Best to mention widget type, widget name and what parsing are we doing here.

app/client/src/widgets/FilePickerWidgetV2/widget/index.tsx
```diff
@@ -959,6 +998,50 @@ class FilePickerWidget extends BaseWidget<
...
+    }
+
+    if (dataFormat === FileDataTypes.Array && !isCSVFileType(fileType)) {
+      return [];
+    }
+    if (!result) {
+      return [];
+    }
+
+    const data: any[] = [];
```
ashit-rath, line 1013: Instead of any can we use?
`Record<string, string>[]`

app/client/cypress/fixtures/Test csv.csv
```diff
@@ -0,0 +1,3 @@
+Data Id,String,Number,Boolean,Empty,Date
+hsa-miR-942-5p,Blue,23.788,TRUE,,"Wednesday, 20 January 1999"
+hsa-miR-943,Black,1000,FALSE,,2022-09-15
```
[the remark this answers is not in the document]
Tooluloope: Adding a new line here would just add an extra row to the CSV which shouldn't be

app/client/src/constants/PropertyControlConstants.tsx
Tooluloope: @aswathkk Should I take care of this here cos it might be a bit confusing to widget developers?

ashit-rath (approved):

ashit-rath:
@Tooluloope If I have the file-type as CSV and I upload an image file, do we show any error in that upload modal itself?
Also do we auto select `Allowed File Type` to `Text` or change the options to only have `Text`?
cc: @dilippitchika

app/client/src/constants/PropertyControlConstants.tsx
aswathkk: @Tooluloope If we do that, we should change the property in all the widgets. So, I would recommend creating a separate issue and PR for that task instead of doing it here.

app/client/cypress/fixtures/Test csv.csv
```diff
@@ -0,0 +1,3 @@
+Data Id,String,Number,Boolean,Empty,Date
```
mohanarpit, line 1: Minor nit: Please avoid spaces in file names. In this case, the file name should be `Test_csv.csv`. This is because on the command line, having spaces in file names is cumbersome & needs to be escaped. 

I just noticed this as I was browsing through some of the PRs. Hence, adding this minor comment here.

Tooluloope: Got it @mohanarpit

Tooluloope:
> If I have the file-type as CSV and I upload an image file, do we show any error in that upload modal itself?
Right now we don't show any error. The data is parsed as an empty array

@ashit-rath

dilippitchika:
Ok just thinking this through
When users select all/any as the types supported in the filepicker and user selects array in file type. 
1. We can append the following sentence to the helper text. "All non csv filetypes will have an empty value."
2. We can't show any error here for user because the developer for some reason is allowing this.

When user selects other filetypes which don't contain csv
1. We can hide the option for array
2. If the user then goes ahead and changes the filetype to contain csv, we now have to show the csv option again. Revalidation needs to happen here, which is an open issue on fe-coders.
3. The options can come from JS as well, so it's hard to have this dependency.
This is too much work for such a small item

I think we should just add the following in the helper text "All non csv filetypes will have an empty value." along with the current helper text for the field. @ashit-rath @Tooluloope 


> Also do we auto select Allowed File Type to Text or change the options to only have Text?
@ashit-rath did not understand this

ashit-rath:
@dilippitchika Makes sense to add additional message to helper text.
> Also do we auto select Allowed File Type to Text or change the options to only have Text?

When user selects `Data Format` as CSV, we could just auto select the correct file types for them (text, csv) or we could only show the appropriate options and hide image/video etc.

> The options can come from JS as well, so it's hard to have this dependency.

When we have a binding in the fileTypes, the value is stored as `allowedFileTypes`. Options are just for the dropdown and does not have any relevance with the JS binding.

app/client/src/widgets/FilePickerWidgetV2/widget/index.tsx
```diff
@@ -487,6 +499,11 @@ class FilePickerWidget extends BaseWidget<
             propertyName: "fileDataType",
             label: "Data Format",
             controlType: "DROP_DOWN",
+            helperText: (props: FilePickerWidgetProps) => {
+              return props.fileDataType === FileDataTypes.Array
+                ? "Larger files will slow down the app. In such cases upload files to a database and query them."
```
ashit-rath, line 504: Are we updating the `helperText` here?

ashit-rath (approved):

chandannkumar:
Tested this PR and LGTM
- Tested with uploading large csv files via File picker and display on table widget
- Tested value in csv file with comma, new line, double quotes
- Tested uploading of non-csv files should show empty value
- Tested File picker widget by uploading files of all supported data formats

PR #16690, opened 2022-09-12
fix: No-op Asserts

sharat87:
There are many occurrences of the following:

```java
assertThat(...)
```

This call itself, doesn't _assert_ anything. It's return value has to be followed up with the actual assertion we want to do. This PR fixes such calls to actually perform the assertion.

So, turns out, quite a few of these were failing, so this PR also fixes them.

### DIFF
--- a/app/server/appsmith-server/src/test/java/com/appsmith/server/services/PageServiceTest.java
+++ b/app/server/appsmith-server/src/test/java/com/appsmith/server/services/PageServiceTest.java
@@ -311,7 +311,7 @@ public class PageServiceTest {
                     PageDTO page = tuple.getT1();
                     assertThat(page).isNotNull();
                     assertThat(page.getId()).isNotNull();
-                    assertThat("PageServiceTest TestApp".equals(page.getName()));
+                    assertThat("PageServiceTest TestApp").isEqualTo(page.getName());
 
                     assertThat(page.getPolicies()).isNotEmpty();
 
@@ -668,18 +668,18 @@ public class PageServiceTest {
                     assertThat(actionWithoutCollection.getUnpublishedAction().getName()).isEqualTo("PageAction");
 
                     // Confirm that executeOnLoad is cloned as well.
-                    assertThat(Boolean.TRUE.equals(actionWithoutCollection.getUnpublishedAction().getExecuteOnLoad()));
+                    assertThat(actionWithoutCollection.getUnpublishedAction().getExecuteOnLoad()).isTrue();
 
                     // Check if collections got copied too
                     List<ActionCollection> collections = tuple.getT3();
-                    assertThat(collections.size()).isEqualTo(1);
+                    assertThat(collections).hasSize(1);
                     assertThat(collections.get(0).getPublishedCollection()).isNull();
                     assertThat(collections.get(0).getUnpublishedCollection()).isNotNull();
                     assertThat(collections.get(0).getUnpublishedCollection().getPageId()).isEqualTo(clonedPage.getId());
 
                     // Check if the parent page collections are not altered
                     List<ActionCollection> parentPageCollections = tuple.getT4();
-                    assertThat(parentPageCollections.size()).isEqualTo(1);
+                    assertThat(parentPageCollections).hasSize(1);
                     assertThat(parentPageCollections.get(0).getPublishedCollection()).isNotNull();
                     assertThat(parentPageCollections.get(0).getUnpublishedCollection()).isNotNull();
                     assertThat(parentPageCollections.get(0).getUnpublishedCollection().getPageId()).isEqualTo(page.getId());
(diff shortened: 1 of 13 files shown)


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/WidgetSuggestionDTO.java
```diff
@@ -2,9 +2,11 @@
 
 import lombok.Getter;
 import lombok.Setter;
+import lombok.ToString;
 
 @Getter
 @Setter
+@ToString
```
sharat87, line 9: Without this, test failures show up like this:

```
Expecting actual:
  [com.appsmith.external.models.WidgetSuggestionDTO@6bd7d6a,
    com.appsmith.external.models.WidgetSuggestionDTO@661a9355,
    com.appsmith.external.models.WidgetSuggestionDTO@52056ad9]
to contain exactly in any order:
  [com.appsmith.external.models.WidgetSuggestionDTO@167f0224,
    com.appsmith.external.models.WidgetSuggestionDTO@3fbe1bde,
    com.appsmith.external.models.WidgetSuggestionDTO@51be402f]
elements not found:
  [com.appsmith.external.models.WidgetSuggestionDTO@167f0224]
and elements not expected:
  [com.appsmith.external.models.WidgetSuggestionDTO@6bd7d6a]
when comparing values using recursive field/property by field/property comparator on all fields/properties using the following configuration:
```

And, with this line, it shows up as:

```
Expecting actual:
  [WidgetSuggestionDTO(type=SELECT_WIDGET, bindingQuery=data.map( (obj) =>{ return  {'label': obj.width, 'value': obj.url } })),
    WidgetSuggestionDTO(type=TABLE_WIDGET_V2, bindingQuery=data),
    WidgetSuggestionDTO(type=TEXT_WIDGET, bindingQuery=data)]
to contain exactly in any order:
  [WidgetSuggestionDTO(type=SELECT_WIDGET, bindingQuery=data.map( (obj) =>{ return  {'label': obj.url, 'value': obj.width } })),
    WidgetSuggestionDTO(type=TABLE_WIDGET_V2, bindingQuery=data),
    WidgetSuggestionDTO(type=TEXT_WIDGET, bindingQuery=data)]
elements not found:
  [WidgetSuggestionDTO(type=SELECT_WIDGET, bindingQuery=data.map( (obj) =>{ return  {'label': obj.url, 'value': obj.width } }))]
and elements not expected:
  [WidgetSuggestionDTO(type=SELECT_WIDGET, bindingQuery=data.map( (obj) =>{ return  {'label': obj.width, 'value': obj.url } }))]
when comparing values using recursive field/property by field/property comparator on all fields/properties using the following configuration:
```

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ce/ActionServiceCE_Test.java
```diff
@@ -2415,6 +2415,52 @@ public void testWidgetSuggestionAfterExecutionWithJsonArrayObjectData() throws J
...
     public void testWidgetSuggestionNestedData() throws JsonProcessingException {
 
+        Mockito.when(pluginExecutorHelper.getPluginExecutor(Mockito.any())).thenReturn(Mono.just(pluginExecutor));
+        Mockito.when(pluginExecutor.getHintMessages(Mockito.any(), Mockito.any()))
+                .thenReturn(Mono.zip(Mono.just(new HashSet<>()), Mono.just(new HashSet<>())));
+        ActionExecutionResult mockResult = new ActionExecutionResult();
+        final String data = "{\"data\": {\n" +
+                "    \"next\": \"https://mock-api.appsmith.com/users?page=2&pageSize=10\",\n" +
+                "    \"previous\": null,\n" +
+                "    \"users\": [1, 2, 3]\n" +
```
sharat87, line 2425: This is what makes the `TEXT_WIDGET` contain `users`.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/PageServiceTest.java
```diff
@@ -863,11 +863,12 @@ public void clonePage_whenPageCloned_defaultIdsRetained() {
                     assertThat(actionWithoutCollection.getUnpublishedAction().getDefaultResources().getPageId()).isEqualTo(clonedPage.getDefaultResources().getPageId());
 
                     // Confirm that executeOnLoad is cloned as well.
-                    assertThat(Boolean.TRUE.equals(actions.get(0).getUnpublishedAction().getExecuteOnLoad()));
+                    // TODO: Fix failing test.
+                    //assertThat(actions.get(0).getUnpublishedAction().getExecuteOnLoad()).isTrue();
```
sharat87, line 867: This test is failing, and I'm not sure why. I'm not even sure if the test needs correction, or the code being tested. Could use some help.

I'm commenting it, with a `TODO` comment, since this way, anyone reading it will know that this test is incomplete and not asserting, instead of potentially giving a false impression.

nidhi-nair: Should we create a high priority issue to fix this? I think we were meant to refactor the clone page path entirely, might help to have this on the radar for when that happens.

sharat87: Sure, yes. Honestly I didn't understand the impact of this test. I can create an issue referencing this TODO item once this PR is merged. That fix can come in as a separate PR.

nidhi-nair (approved):

app/server/appsmith-server/src/test/java/com/appsmith/server/services/PageServiceTest.java
sharat87: Opened issue at https://github.com/appsmithorg/appsmith/issues/16724

PR #16698, opened 2022-09-12
feat: grouping in console logs

### LINKED ISSUE #16333: [Feature]: Grouping in logs tab for repeated log statements

### Summary

The logs tab gets very cluttered due to repeated logs. We can collapse the repeated logs into a single one along with a tag to show multiple logs of the same one.

For ease of consumption of log outputs.

ayushpahwa:
## Description

This PR adds functionality to group multiple repeated logs into one single log. This helps to declutter our logs tab.

Fixes #16333 

## Type of change

## How Has This Been Tested?
- **Test Plan**-

- [ ] - https://github.com/appsmithorg/TestSmith/issues/2062

## Checklist:


app/client/src/components/editorComponents/Debugger/hooks/debuggerHooks.ts
```diff
@@ -59,7 +58,7 @@ export const usePagination = (data: Log[], itemsPerPage = 50) => {
   useEffect(() => {
     const data = currentData();
     setPaginatedData(data);
-  }, [currentPage, data.length]);
+  }, [currentPage, data.length, data[data.length - 1]?.occurrenceCount]);
```
arunvjn, line 61: Is the data here somehow different from the data inside the callback ?

app/client/src/utils/AppsmithConsole.ts
```diff
@@ -109,6 +116,35 @@ function deleteError(id: string, analytics?: Log["analytics"]) {
...
+) {
+  const outputArray = currentLogs;
+  incomingLogs.forEach((incomingLog) => {
+    if (outputArray.length === 0) {
+      outputArray.push(incomingLog);
+    } else {
+      const lastLog = outputArray[outputArray.length - 1];
+      // The equality needs to be tested without occurrenceCount, since that is a dynamic value
+      if (
+        !equal(
```
arunvjn, line 132: Can we do a log text compare here instead of deep equal? 
lastLog.text === incomingLog.text ?

app/client/src/utils/AppsmithConsole.ts
```diff
@@ -109,6 +116,35 @@ function deleteError(id: string, analytics?: Log["analytics"]) {
   dispatchAction(deleteErrorLogInit(id, analytics));
 }
 
+// check the last message from the current log and update the occurrence count
+export function removeRepeatedLogsAndMerge(
+  currentLogs: Log[],
+  incomingLogs: Log[],
+) {
+  const outputArray = currentLogs;
+  incomingLogs.forEach((incomingLog) => {
```
arunvjn, line 125: Can we use map/reduce/of/native for loops over forEach? forEach operations are slower in general.

app/client/src/utils/AppsmithConsole.ts
arunvjn: Also, if the last log in a previous execution and the first log in the new execution are the same, they would get grouped here

ramsaptami:
QA has been completed on DP and is good to go to release
CC: @ayushpahwa

app/client/src/utils/AppsmithConsole.ts
ayushpahwa: @arunvjn no, we can't do that because we need to make sure everything including the time of the log and source should be the same. Hence we are doing a deep equal. This also means that two logs with different timestamps will not be combined.

app/client/src/components/editorComponents/Debugger/hooks/debuggerHooks.ts
ayushpahwa: The data is same but this method is comparing just the length of the data here. We need to add a comparison to the last log's occurrence count also or it will not update the UI in that case.

app/client/src/utils/AppsmithConsole.ts
ayushpahwa: Done, please check

arunvjn (approved):

Aishwarya-U-R (commented):
Cypress looks good!

ayushpahwa:
Adding test plan approved as verified by QA (both manual and automated)

arunvjn (approved):

PR #16701, opened 2022-09-12
feat: Autoassigns column datatype in table widget

### LINKED ISSUE #12831: [Feature]-[2400]:Autoassign the column datatype in table widget instead of default Plain Text depending upon the column data

### Summary

Currently for any type of data in table widget `Plain Text` is auto-assigned. User will have to change it manually if one wants to. 

One of the use-case where we find this useful is in generate CRUD page feature, where we have implemented the server side pagination along with the order by operator which depends upon the `table.sortOrder.column` and we don't have any control over user's data so default assignment don't work here.

Sample query:  
```
SELECT * FROM public."users"
WHERE "field1" ilike '%{{data_table.searchText || ""}}%'
ORDER BY {{data_table.sortOrder.column || 'id'}} {{data_table.sortOrder.order || 'ASC'}}
LIMIT {{data_table.pageSize}}
OFFSET {{(data_table.pageNo - 1) * data_table.pageSize}};
```

souma-ghosh:
## Description

When creating a column from tableData we were auto-assigning columns to "Plain Text" type by default without looking at the data of the column.

In this PR (In order to auto-assign better column types) before creating each column
1. We pick up the the first non-null value for that column from tableData
2. Assign datatype to the column according to the value

Note: Currently limited the check for non-null values to `maxRowsToCheck` rows, where `maxRowsToCheck = 5`

Datatypes being checked for and the corresponding column type being assigned:
- `number`: Number
- `boolean`: Checkbox
- `string`: Plain Text
- popular date formats: Date
- fallback/default: Plain Text

Fixes #12831

## Type of change

## How Has This Been Tested?

- Jest tests

## Checklist:


app/client/src/widgets/TableWidgetV2/widget/utilities.ts
```diff
@@ -610,3 +611,53 @@ export const createEditActionColumn = (props: TableWidgetProps) => {
...
+    row = 0;
+  const maxRowsToCheck = 5;
+  /*
+    In below while loop we are trying to get a non-null value from
+    subsequent rows in case first few rows are null
+    Limited to checking upto maxRowsToCheck
+  */
+  while (!columnValue && row < maxRowsToCheck) {
+    if (tableData?.[row]?.[columnKey]) {
+      columnValue = tableData[row][columnKey];
```
sbalaji1192, line 633: we need to break here once we found the non null value.

sbalaji1192: And also it could be an empty string and false. so you need to update the condition to check specifically for the null value.

app/client/src/widgets/TableWidgetV2/widget/utilities.ts
```diff
@@ -610,3 +611,53 @@ export const createEditActionColumn = (props: TableWidgetProps) => {
...
+    }
+    row++;
+  }
+
+  if (!columnValue) {
+    return ColumnTypes.TEXT;
+  }
+
+  try {
+    parsedColumnValue = _.isString(columnValue)
```
sbalaji1192, line 643: We don't need to parse. we should distinguish `"false"`  from `false`. Because checkbox treats `"false"` as true value. So if we set the type checkbox. it will show up as checked.

app/client/src/widgets/TableWidgetV2/widget/utilities.ts
```diff
@@ -610,3 +611,53 @@ export const createEditActionColumn = (props: TableWidgetProps) => {
...
+      ? JSON.parse(columnValue)
+      : columnValue;
+  } catch (e) {
+    // eslint-disable-next-line no-console
+    console.error(
```
sbalaji1192, line 648: need to return the default type here.

app/client/src/widgets/TableWidgetV2/widget/utilities.ts
souma-ghosh: Should we remove the whole try catch block here then? Since its only for the parsing and nothing else @sbalaji1192

sbalaji1192: yeah. remove the block

souma-ghosh: Without the parsing `typeof "23"` will give us string, hence the column will get TEXT type instead of NUMBER type. Is this going to be an issue? @sbalaji1192

sbalaji1192: that's alright. We should only consider `23` as a number type.

app/client/src/widgets/TableWidgetV2/widget/utilities.ts
```diff
@@ -611,3 +612,42 @@ export const createEditActionColumn = (props: TableWidgetProps) => {
...
+  let columnValue = null,
+    row = 0;
+  const maxRowsToCheck = 5;
+  /*
+    In below while loop we are trying to get a non-null value from
+    subsequent rows in case first few rows are null
+    Limited to checking upto maxRowsToCheck
+  */
+  while (
+    (columnValue === null || columnValue === undefined) &&
```
sbalaji1192, line 632: use _.isNill

app/client/src/widgets/TableWidgetV2/widget/utilities.ts
```diff
@@ -611,3 +612,42 @@ export const createEditActionColumn = (props: TableWidgetProps) => {
...
+    row < maxRowsToCheck
+  ) {
+    if (tableData?.[row]?.[columnKey] !== null) {
```
sbalaji1192, line 635: use _.isNil

sbalaji1192: you need to add break here.

laveena-en:
@souma-ghosh On changing the type of a value, it isn’t reflecting on the table (In this case, I changed a boolean value to string). Not sure if this pertains to this PR. Could you have a look and confirm please? 

https://www.loom.com/share/a5afac6d7b3e4a419dcbc57d77e94b62

souma-ghosh:
> @souma-ghosh On changing the type of a value, it isn’t reflecting on the table (In this case, I changed a boolean value to string). Not sure if this pertains to this PR. Could you have a look and confirm please?
> 
> https://www.loom.com/share/a5afac6d7b3e4a419dcbc57d77e94b62

@sbalaji1192 @dilippitchika Should we try to switch the data type if the user changes the table data manually as shown in the above video?

sbalaji1192:
@souma-ghosh We don't have to. We're only going to guess the type when a column is generated. we don't have to keep guessing every time the table data changes.

CC @dilippitchika

souma-ghosh:
If everything else is working fine we can go ahead with the current solution then @laveena-en

dilippitchika:
Agreed with balaji, we only guess the first time. After that, it's the users responsibility to manage them.

laveena-en:
Yes, we can go ahead with this. 
Tested and verified for data that consists of:
1. String
2. Number
3. Boolean values

dilippitchika:
@laveena-en what about dates?

souma-ghosh:
@dilippitchika Dates have not been handled yet. For us to confirm whether a column data is date type or not it has to be a valid ISO or UTC format or another format we explicitly specify. So should we cannot guess a column to be of type date unless we know what format/formats we are expecting

sbalaji1192 (dismissed):
Could you improve the Pr description?

laveena-en:
@souma-ghosh As discussed, it seems like for a specific date format (22/09/2022 18:30), the date seems to be seen as a number instead. Please do have a look.
image
https://docs.google.com/spreadsheets/d/1H-IEkVGRlD4W37TRq9bQ3j63YFAbvAYKSSomTp-zurA/edit?usp=sharing

souma-ghosh:
> @souma-ghosh As discussed, it seems like for a specific date format (22/09/2022 18:30), the date seems to be seen as a number instead. Please do have a look. image https://docs.google.com/spreadsheets/d/1H-IEkVGRlD4W37TRq9bQ3j63YFAbvAYKSSomTp-zurA/edit?usp=sharing

As we saw in the computed value for the api we are using here, we get a number not a string for `Date9` column. Since any number can be considered as a millisecond date format, whenever we get a number in the data we assign the column as number type column first. If the and only if we get a string type data we will assign the column as Date type or String type accordingly. So this behaviour is correct

laveena-en:
LGTM!

sbalaji1192 (approved):

PR #16709, opened 2022-09-13
feat: Progress Widget Reskinning

### LINKED ISSUE #10998: [Enhancement]: Progress bar reskinning

dhruvikn:
## Description
This PR fixes UI inconsistencies in the Progress widget - both linear and circular.

Fixes #10998 

## Type of change

## How Has This Been Tested?
- Manually

## Checklist:

### DIFF
--- a/app/client/src/widgets/ProgressWidget/component/index.tsx
+++ b/app/client/src/widgets/ProgressWidget/component/index.tsx
@@ -154,6 +154,7 @@ const DeterminateLinearProgress = styled.div<{
     left: 0;
     top: 0;
     bottom: 0;
+    border-radius: ${({ borderRadius }) => borderRadius};
   }
 `;
 
--- a/app/client/src/widgets/ProgressWidget/component/index.tsx
+++ b/app/client/src/widgets/ProgressWidget/component/index.tsx
@@ -274,9 +274,11 @@ export const Trail = styled.path`
 
 export const Path = styled.path<{
   fillColor: string;
+  borderRadius?: string;
 }>`
   stroke: ${(props) => props.fillColor};
-  stroke-linecap: butt;
+  stroke-linecap: ${({ borderRadius }) =>
+    borderRadius === "0px" ? "butt" : "round"};
   transition: stroke-dashoffset 0.5s ease 0s;
 `;
 
@@ -347,6 +349,7 @@ function RadialSeparators(props: { steps: number }) {
 // Pure circular progress (indeterminate/determinate)
 function CircularProgress(props: ProgressComponentProps) {
   const {
+    borderRadius,
     counterClockwise,
     fillColor,
     showResult,
@@ -426,6 +429,7 @@ function CircularProgress(props: ProgressComponentProps) {
           />
 
           <Path
+            borderRadius={borderRadius}
             d={drawPath()}
             data-testvalue={value}
             fillColor={fillColor}
@@ -453,12 +457,15 @@ function CircularProgress(props: ProgressComponentProps) {
 function CircularProgressWithSteps(
   props: ProgressComponentProps & { children?: React.ReactNode },
 ) {
-  const { children, ...circularProgressProps } = props;
+  const { borderRadius, children, ...circularProgressProps } = props;
 
   return (
     <CircularProgressWithStepsWrapper isScaleY={circularProgressProps.isScaleY}>
       <CircularProgressWithStepsContainer>
-        <CircularProgress {...circularProgressProps} />
+        <CircularProgress
+          borderRadius={borderRadius}
+          {...circularProgressProps}
+        />
         {children && (
           <CircularProgressWithStepsOverlay>
             {children}


jsartisan (approved):
LGTM. Just one question, we don't have any disabled state for progress widget right?

dhruvikn:
No @jsartisan, we don't have any disabled state for the progress widget.

dhruvikn:
The border radius for circular progress can only be either none or entirely round (L). No in-between (M).
If we were to achieve slightly round (M) edges in the arc, we'll need to change the way this filled arc is generated (via SVGs), which is a significant effort. This is as far as I could get without changing the core code and using the CSS stroke-linecap property. 

https://user-images.githubusercontent.com/22471214/190085739-cc069cfc-58ae-47e5-9ea3-83748809b36d.mp4

However, one issue when using `stroke-linecap="round"` for rounded edges is that it extends the length of the arc slightly on both ends of the arc. Which will cause this - 

https://user-images.githubusercontent.com/22471214/190087657-f42dd75c-249e-4c9b-85a9-84644a30b518.mp4

This is because ([source](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-linecap#example)) - 

> The round value indicates that at the end of each subpath the stroke will be extended by a half circle with a diameter equal to the stroke width.

CleanShot 2022-09-14 at 12 44 04@2x

cc: @dilippitchika @somangshu

somangshu:
I agree that the new effort does not seems worth while, What ever can be the smallest change should be taken forward

For the challenge @dhruvikn pointed out, Can we reduce the length by the amount of increase in edges while using stroke-linecap="round".

Also, since we can only have 2 modes, rounded / flat, would it be possible to remove the additional option in the property pane.

My last question. @dilippitchika do you think we really need this option? Even in terms of consistency, does it make sense to give such a control to the user?

laveena-en:
@dhruvikn The border of the unfilled blocks don't seem to have the border radius set. (The end of the progress bar and the individual blocks within) Can you please check? 

https://www.loom.com/share/951deccd1ced492ebb79461366c9616c

dhruvikn:
@somangshu, reducing the length would again mean changing the way the [SVG is generated](https://github.com/appsmithorg/appsmith/blob/29de2df4eb6610f1391d7fc985da908cce2ed909/app/client/src/widgets/ProgressWidget/component/index.tsx#L359). We'd not only need to subtract the extra length, but also reposition the arc.

Anything, apart from CSS styling changes, would mean changing the SVG generator code.

dhruvikn:
Also, @somangshu, the widget does not have the border-radius property in its style config. It is inheriting it from the app theme. Do we need to add the border-radius property to the widget as well?

CleanShot 2022-09-14 at 13 13 21@2x

cc: @dilippitchika

somangshu:
@dhruvikn the SVG inner-path is having a `stroke-dasharray` whose value can be altered to reduce the size it takes up, might not be a tidy solution. 
I get what you are saying, but the ux might be unacceptable to the end user, they will not have the right idea about the progress that is made. 

> It is inheriting it from the app theme. Do we need to add the border-radius property to the widget as well?

I still am wondering weather this widget really needs to inherit this attr, I kinda feel that it should not, Ill wait for @dilippitchika to respond on that

dhruvikn:
Addressed [unflied step blocks border radius issue](https://github.com/appsmithorg/appsmith/pull/16709#issuecomment-1246359688) in the latest commit, @laveena-en.

CleanShot 2022-09-14 at 13 18 16@2x

dhruvikn:
Yes @somangshu, you're right. Manipulating `stroke-dasharray` in this case would be a hacky solution. Let's wait for @dilippitchika

dilippitchika:
Ideally we should also curve the edges for a stepped circular property pane to make sure it matches with what the border radius is. If that is not possible though, then my suggestion is not to have rounded edges when there are steps. When there are no steps we should have rounded edges based on the theme. 

It's quite common to use rounded edges in progress bars.

dilippitchika:
Screenshot 2022-09-14 at 4 13 00 PM
Also i think there some wrong calculation happening here, when the progress is 99, it showing 100. @dhruvikn this might not be a good idea.

dhruvikn:
Hey @dilippitchika, two things, let's discuss this further in our call - 
1. For the circular progress widget, the calculation happening is correct, however, due to this issue (check the comment out) - https://github.com/appsmithorg/appsmith/pull/16709#issuecomment-1246352745, we are getting the wrong length of the arc.
2. Check out these states for linear progress - 
CleanShot 2022-09-14 at 16 31 46@2x
CleanShot 2022-09-14 at 16 31 35@2x
CleanShot 2022-09-14 at 16 31 28@2x
CleanShot 2022-09-14 at 16 30 50@2x
CleanShot 2022-09-14 at 16 30 36@2x
CleanShot 2022-09-14 at 16 30 26@2x

dhruvikn:
Pausing the work, and closing this PR for now in light of the above-mentioned issues in the circular progress widget. We, the Appsmith team, will discuss and plan out a proper way to have rounded edges (M and L) with correct progress calculation. Thanks.

cc: @somangshu @dilippitchika

PR #16747, opened 2022-09-14
fix: Extra Space at the end of the Canvas in Viewmode

### LINKED ISSUE #12672: [Bug] Empty space in bottom of the page in view mode

### Description

Current Experience
- The canvas does not clip to hug contents in the view mode, resulting in a lot of empty space at the bottom of the page and an unnecessary scrollbar.

Expected
- The canvas should clip to hug contents in the view mode.

### Steps To Reproduce

1. Add a widget to the top of the canvas.
2. Publish the application
3. Reduce the viewport size of the browser to notice a scrollbar appearing even though there is no content in the canvas.

### Public Sample App

_No response_

### Version

Cloud

rahulramesha:
## Description

We had logic to fix MainCanvas height when a Widget is dragged, resized or deleted. Extending that logic to When a widget is added to mainCanvas as well.

Also added logic to trim mainCanvas width in View Mode.

Fixes #12672 

## Type of change

## How Has This Been Tested?

Manual UI and Jest test

## Checklist:


app/client/src/components/editorComponents/DropTargetUtils.ts
```diff
@@ -18,7 +22,10 @@ export const calculateDropTargetRows = (
     }, widgetBottomRow);
   }
 
-  return Math.ceil(
-    Math.max(minBottomRow + GridDefaults.CANVAS_EXTENSION_OFFSET, defaultRows),
-  );
+  const canvasOffset =
```
marks0351, line 25: @rahulramesha  why a different offset for main canvas, if it is to decrease annoyance then its fine but i see the offset being applied in view mode as well which creates a scroll when a page is created exactly for the view port height.

app/client/src/constants/WidgetConstants.tsx
```diff
@@ -79,6 +79,7 @@ export const GridDefaults = {
   DEFAULT_GRID_COLUMNS: 64,
   DEFAULT_GRID_ROW_HEIGHT: 10,
   CANVAS_EXTENSION_OFFSET: 2,
+  MAIN_CANVAS_EXTENSION_OFFSET: 8,
```
riodeuno, line 82: @rahulramesha What's the rationale? Can't we use the `CANVAS_EXTENSION_OFFSET` here?

riodeuno (commented):
Just one comment, the rest looks good.

marks0351 (changes requested):

app/client/src/components/editorComponents/DropTargetUtils.ts
rahulramesha: @riodeuno @marks0351 The main rationale for `MAIN_CANVAS_EXTENSION_OFFSET` was that, `CANVAS_EXTENSION_OFFSET` is too small for mainCanvas. On edit mode, if the same `CANVAS_EXTENSION_OFFSET` is used, it looks as if the canvas grids abruptly ends at the last widget position and doesn't looks good while dragging. On View mode, it acts as a padding below the last widget, so that the bottom most widgets are easily accessible to interact rather than at the border of the browser, It does have a bit of scroll only if canvas goes beyond a certain length but not by much.

marks0351: why do we need this padding in view mode? did we have any feedback regarding usability? It's better if we do this only out of necessity else just skip and have the height equal to the bottom most widgets bottom row.

rahulramesha: @marks0351 This is the one i was mentioning, you could check the below image, with regular canvas offset (on release). It makes the button non interactive
image
Same app on this DP,
image

marks0351: yea it makes sense to implement this to accommodate this particular area of the canvas but in ee apps i think we have the provision to  hide the appsmith label and there it might be an unwanted feature and i dont recommend having a flag to deal with that. if we are going to fix this lets fix it as a feature by giving the ability to the user to set their bottom row. what do you guys think?

rahulramesha: Providing the feature to set their bottom row which is intuitive to the user might require designs and UX. We can pursue that approach as well but since this works for more cases and solves the long canvas issue, I think the extra 8 rows wont make a difference since most websites have a padding at the bottom.  thoughts?

riodeuno (dismissed):

marks0351 (dismissed):

app/client/src/sagas/WidgetOperationUtils.ts
```diff
@@ -1669,3 +1670,80 @@ export function mergeDynamicPropertyPaths(
 ) {
   return _.unionWith(a, b, (a, b) => a.key === b.key);
 }
+
+/**
+ * Note: Mutates finalWidgets[parentId].bottomRow for CANVAS_WIDGET
+ * @param finalWidgets
+ * @param parentId
+ */
+export function resizeCanvasToLowestWidget(
```
riodeuno, line 1679: @rahulramesha I am concerned that this is a mutating function. Do you think we could return a value from this? There are two concerns I have.
1. It makes testing easier, no need for cloning.
2. This function only changes one value. It should be fairly simple to add a line and update these in the `finalWidgets` where necessary.

app/client/src/sagas/WidgetOperationUtils.ts
```diff
@@ -1669,3 +1670,80 @@ export function mergeDynamicPropertyPaths(
...
+    if (!child.detachFromLayout && child.bottomRow > lowestBottomRow) {
+      lowestBottomRow = child.bottomRow;
+    }
+  });
+
+  const canvasOffset =
+    parentId === MAIN_CONTAINER_WIDGET_ID
+      ? GridDefaults.MAIN_CANVAS_EXTENSION_OFFSET
+      : GridDefaults.CANVAS_EXTENSION_OFFSET;
+  finalWidgets[parentId].bottomRow = Math.max(
```
riodeuno, line 1714: Should this be Math.min instead?

riodeuno (commented):
A few comments. Overall looks good to me.

app/client/src/constants/WidgetConstants.tsx
riodeuno: @rahulramesha I thought we were changing this to 5. Have we reconsidered?

app/client/src/sagas/WidgetOperationUtils.ts
rahulramesha: @riodeuno got it, I had just moved the method from another file. I agree with the points will change it to return the value instead of mutating it

app/client/src/sagas/WidgetOperationUtils.ts
rahulramesha: It is supposed to be Math.max, because it is comparing the maximum value of the lowest bottom row of the widget or the bottom row that is required to fill the entire height of the browser window.

riodeuno (approved):

PR #16749, opened 2022-09-14
fix: AST on BE (WIP)

### LINKED ISSUE #16338: [Task]: Consume AST apis on backend server

### SubTasks

The apis are added in RTS server. There are two apis for the AST backend usage :

- **Single script Identifiers**
  _Method_: `POST`
  _URL_: `/rts-api/v1/ast/single-script-data`
  _Body_: `{ "script": "(function(){ let a = 1; return [Api1.data, Api2.data, a]; })()", "evalVersion": 2 }` 
  _Response_: `{"success":true,"data":{"identifiers":["Api1.data","Api2.data"]}}`
  
- **Multiple script Identifiers**
  _Method_: `POST`
  _URL_: `/rts-api/v1/ast/multiple-script-data`
  _Body_: `{ "scripts": ["(function(){ let a = 1; return [Api1.data, Api2.data, a]; })()", "(function(){ let a = 1; return [XYZ.data, a]; })()"], "evalVersion": 2 }` 
  _Response_: `{"success":true,"data":[{"identifiers":["Api1.data","Api2.data"]},{"identifiers":["XYZ.data"]}]}`

> Where evalVersion is optional and default value is 2,  eval  version is the version we can find in the response of `/api/v1/pages?pageId=${pageId}` request of the application  where the path for it is `data.application. evaluationVersion`.

nidhi-nair:
## Description
This PR introduces a new way of parsing all the dynamic bindings present in an Appsmith application. 

Earlier, we used to break down the string representation of a binding into words and try to predict what possible entities we were referring to in any piece of code. An entity could be an action (query/API), a JS function, a widget, or an `appsmith` object reference. That last type is not considered in our dependency logic today so we'll skip talking about it.

However, this approach was ridden with bugs because this meant that if we had name clashes inside comments, strings, local variables, etc., we would still continue to recognize those as a reference to the global entity. A smaller concern that arose from this was additional runs of queries/APIs. But it was also possible that we end up creating cyclical references along the way, even the the user had done no such thing.

With this set of changes, we will now have a more reliable way of figuring out these entity dependencies. Now, we are relying on the AST representation of any piece of code to find proper global variable references. This essentially means we'll be ignoring all the comment/string/local variable/function parameter references that were identified before. This also brings us closer to how the client calculates dependencies and hence brings consistency in these calculations.

Outside of this, the PR also introduces a few optimizations to the way we were processing these references that we had identified. Before this PR, we did not have a sense of the nature of each of these references, and in the process of calculating the dependency graph, we needed to refer to the DB over and over to check for this information. Now, we have introduces types that retain knowledge of whether a particular reference is a candidate for a widget, action or JS action reference. We also attach a little more information about the type of reference in terms of whether it is an async function, a direct function call, etc. Please check the `EntityDependencyNode` class that captures this information. 

Any identified node is propagated through the logic with this information so that we can rely on this one source of information for further processing.

This solution makes use of APIs exposed on the RTS server to calculate the on page load dependencies. Unfortunately, older slim container setups would not be able to accommodate this change without manual intervention from admins to set up RTS and then expose it to the server. As a result, this PR continues to use the older way of doing things for such set ups. It is highly recommended that users of Appsmith upgrade to the fat container set up to ensure a seamless experience.

TL;DR, here is the change set:
1. On page load calculations will now be more accurate. Good bye, a-certain-type-of cyclical dependency errors.
2. This change is only applicable for fat container setups. If you are not on this, please consider migrating.

Fixes #16338 

## Type of change

## How Has This Been Tested?
- JUnit tests
- Manual testing

## Checklist:


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/MustacheHelper.java
```diff
@@ -322,21 +327,119 @@ public static String render(String template, Map<String, String> keyValueMap) {
         return StringEscapeUtils.unescapeHtml4(rendered.toString());
     }
 
-    public static void extractActionNamesAndAddValidActionBindingsToSet(Map<String, DynamicBinding> bindingNames, String mustacheKey) {
-        String key = mustacheKey.trim();
+    public static Mono<Map<String, Set<EntityDependencyNode>>> getPossibleEntityParentsMap(Set<String> mustacheBindings, int types) {
+
+        return Flux.fromStream(mustacheBindings.stream())
+                .flatMap(bindingString -> {
+                    return WebClient.create("http://localhost:8091/rts-api/v1/ast/single-script-identifiers")
```
trishaanand, line 334: QQ: What happens when our deployment scenario for rts changes? Could we do something more generic here so that RTS deployment is not associated with our code? Also, lets extract the path also into a static final string?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/MustacheHelper.java
```diff
@@ -322,21 +327,119 @@ public static String render(String template, Map<String, String> keyValueMap) {
...
+                            .post()
+                            .body(BodyInserters.fromValue(Map.of("script", bindingString, "evalVersion", 2)))
```
trishaanand, line 336: Super nit pick : Possible to extract the strings and replace them with variables?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/MustacheHelper.java
```diff
@@ -322,21 +327,119 @@ public static String render(String template, Map<String, String> keyValueMap) {
...
+                            .retrieve()
+                            .bodyToMono(Map.class)
+                            .map(response -> (List<String>) response.get("data"))
+                            .zipWith(Mono.just(bindingString));
+                }).collect(HashMap::new, (map, tuple) -> {
+                    String bindingValue = tuple.getT2();
+                    HashSet<EntityDependencyNode> totalParents = new HashSet<>();
+                    tuple.getT1().forEach(reference -> {
+                        if ((types & 0b1) == 1) {
```
trishaanand, line 345: Nit: Could we please add comments to explain these if conditions for future developers reading this? Thanks!

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Action.java
```diff
@@ -1,9 +1,6 @@
 package com.appsmith.server.domains;
 
-import com.appsmith.external.models.ActionConfiguration;
-import com.appsmith.external.models.BaseDomain;
-import com.appsmith.external.models.Datasource;
-import com.appsmith.external.models.Property;
+import com.appsmith.external.models.*;
```
trishaanand, line 3: Hey, could we remove the * and add back the individual imports? Thanks!

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/PageLoadActionsUtilCEImpl.java
```diff
@@ -19,17 +21,8 @@
...
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
```
trishaanand, line 24: Same as above. Thanks!

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/PageLoadActionsUtilCEImpl.java
```diff
@@ -261,6 +254,55 @@ private Mono<List<Set<DslActionDTO>>> filterAndTransformSchedulingOrderToDTO(Set
                 });
     }
 
+    private Mono<Set<EntityDependencyNode>> getPossibleEntityReferences(Mono<Map<String, ActionDTO>> actionNameToActionMapMono, Set<String> bindings) {
```
trishaanand, line 257: Nit: Please add function comments to all the newly added functions (and update any other function comments) to make this file more readable (this in all honesty, this is a very complicated peace of logic happening here.) Thanks!

trishaanand (commented):
Supremely minor comments added. Would love to review again post more context on the changes @nidhi-nair

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/MustacheHelper.java
nidhi-nair: Added a health check to rts to test for whether it is accessible on start up. If it is not, we'll assume this is not fat container and fall back to the string way of comparing and finding references. This is because we want to encourage folks to move to fat container and we'll only provide the existing experience on slim.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/CreateDBTablePageSolutionCEImpl.java
```diff
@@ -342,7 +342,7 @@ public Mono<CRUDPageResponseDTO> createPageFromDBTable(String defaultPageId,
                     }
 
                     log.debug("Going to update layout for page {} and layout {}", savedPageId, layoutId);
-                    return layoutActionService.updateLayout(savedPageId, layoutId, layout)
+                    return layoutActionService.updateLayout(savedPageId, page.getApplicationId(), layoutId, layout)
                             .then(Mono.zip(
```
subrata71, line 346: Sorry, I lack knowledge on this but I couldn't help asking this.
I see previously there's no `applicationId` as part of the `updateLayout` action but now we introduce this. Maybe there's a good reason to introduce the `applicationId` here. Can you please state the reason behind it?
@nidhi-nair

nidhi-nair: The AST logic varies depending on the evaluation version that the application is on. This happened after a change went in the modified how we parse new line characters. Hence, for every application on Appsmith now, we need to make sure we retain this context and parse the bindings in that way.

PR #16750, opened 2022-09-14
fix: chart widget color fontfamily

### LINKED ISSUE #16662: [Bug]: Chart widget's accent color and font family sets to default value for current user theme

### Description

For Chart widget, the accent color and font family sets to the default value for current user theme. On change of theme as well the accent color and font family of the chart does not change. Please find below the screenshot:

Screenshot 2022-09-09 at 12 24 03 PM

### Steps To Reproduce

1. Import the following via JSON file.
2. Open the app
3. Observe the chart widget's color and font family does not match with the current app theme

Here is the JSON file:
[Employee Time Tracker.txt](https://github.com/appsmithorg/appsmith/files/9532917/Employee.Time.Tracker.txt)

### Public Sample App

_No response_

### Version

Cloud

keyurparalkar:
## Description

This PR re-runs the function `migrateChartWidgetReskinningData` on version 61. We are executing this again since older apps didn't get the `accentColor` and the `fontFamily` properties due to this [change](https://github.com/appsmithorg/appsmith/blob/06c8d0c9720a72c6cf55bdec7f9c2f48cc7cca17/app/client/src/utils/DSLMigrations.ts#L1082).

Fixes #16662

## Type of change

## How Has This Been Tested?

Added the unit test named: `DSLMigration.test.ts`.
This test does the following:
1. Make sure that the migration function is executed.
2. Makes sure that migrations are executed in the correct order.
3. Notifies by failing the tests whenever a new migration is added to `transformDSL` but not in the test suite `DSLMigration.test.ts`

## Checklist:


app/client/src/utils/DSLMigration.test.ts
```diff
@@ -0,0 +1,4181 @@
...
+    moduleObj: checkboxMigration,
+    funcName: "migrateCheckboxGroupWidgetInlineProperty",
+    nextVersion: 48,
+  },
+  {
+    moduleObj: buttonWidgetMigrations,
+    funcName: "migrateRecaptchaType",
+    nextVersion: 49,
+  },
+  {
```
ashit-rath, line 4083: Here we skipped a version and I am guessing it's because there isn't an internal migration function.
For these case can we add dummy functions here and have an entry for the version?

Reason:
Let's say I add a new migration function in the `transformDSL` function and I don't add the function here to mock.
Correct me if I'm wrong, this test file won't throw an error and we want developers to always add the new function to this test file.

Now in order to do that we can do something like 
```
expect(migrationList.length).toEqual(LATEST_PAGE_VERSION)
```
^^ In order to make this work perfectly  we can't skip versions in `migrationList`

app/client/src/utils/DSLMigration.test.ts
```diff
@@ -0,0 +1,4181 @@
...
+
+  test.each(migrationList)(
+    "has $funcName ran successfully?",
+    ({ funcName, nextVersion }) => {
+      expect(
+        mockFnObj[funcName].mock.results[0].value.validationFuncName,
+      ).toEqual(funcName);
+      expect(mockFnObj[funcName].mock.results[0].value.version).toEqual(
+        nextVersion,
+      );
```
ashit-rath, line 4174: Can we also validate if the function was called with the correct version?

app/client/src/utils/DSLMigration.test.ts
```diff
@@ -0,0 +1,4181 @@
...
+import * as propertyPaneMigrations from "./migrations/PropertyPaneMigrations";
+import * as themingMigration from "./migrations/ThemingMigrations";
+import { LATEST_PAGE_VERSION } from "constants/WidgetConstants";
+
+interface IMigrationList {
+  moduleObj: any;
+  funcName: string;
+  nextVersion: number;
+}
+const originalDSL = {
```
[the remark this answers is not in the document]
ashit-rath: @keyurparalkar Can you address this comment?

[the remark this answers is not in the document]
keyurparalkar: @ashit-rath moved DSL into `testDSLs.ts` file

ashit-rath (approved):
Can you elaborate the PR description?

keyurparalkar:
> description

@ashit-rath I have updated the description of the PR.

app/client/src/utils/DSLMigration.test.ts
```diff
@@ -0,0 +1,703 @@
...
+  migrations.forEach((migration: Migration) => {
+    /**
+     * Generates mock fucntion for each migration function.
+     * Mocks the implementation
+     */
+    const version = migration.version ?? 0;
+    mockFnObj[version] = [];
+
+    migration.functionLookup.forEach(
+      (funcDetails: { moduleObj: any; functionName: string }) => {
```
ashit-rath, line 625: I don't think you need to define the typings for `funcDetails` here explicitly. I should infer directly.
Do you think we should rename `funcDetails` -> `lookup`?

keyurparalkar: @ashit-rath removed the typings from here. Renamed it to lookup.

ashit-rath (approved):

laveena-en:
Tested and verified.

PR #16785, opened 2022-09-15
chore: POC - List v2 - Introduces currentViewItems and currentRow + Refactoring

### LINKED ISSUE #16422: List V2 POC:  Meta child widgets properties on List widget property
Make the individual properties of the widgets defined in the Entity Definition be available in the List widget.
Eg. `text`, `value` property of the Input widget be made available to the List widget.

ashit-rath:
## Description
Introduces 2 properties and refactors the meta widget generation logic
`currentViewItems` - Belongs to list widget which returns the current view's widget's properties
`currentRow` - special keyword to refer any other widget in the same row, Eg - `{{currentRow.Input1.text}}``

Fixes #16422


Tooluloope:
@ashit-rath `currentItemsView` or `currentViewItems`

ashit-rath:
@Tooluloope  In code i did `currentViewItems` but in description I mentioned it as `currentItemsView`. Let me update

app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.ts
```diff
@@ -0,0 +1,675 @@
...
+};
+
+type GenerateMetaWidgetChildrenProps = {
+  index: number;
+  parentId: string;
+  templateWidget: FlattenedWidgetProps;
+  rowIndex: number;
+};
+
+type generateMetaWidgetReturn = {
```
Tooluloope, line 78: can this be `GeneratedMetaWidget`

app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.ts
```diff
@@ -0,0 +1,675 @@
...
+
+    const key = this.getPrimaryKey(rowIndex);
+    const metaWidget = klona(templateWidget) as MetaWidget;
+    const metaCacheProps = this.getCachedMetaWidgetProps(key, templateWidgetId);
+
+    if (!metaCacheProps) {
+      return {
+        childMetaWidgets: undefined,
+        metaWidgetId: undefined,
+        metaWidgetName: undefined,
```
Tooluloope, line 208: any reason why we are returning `undefined` here and `string` on line 198?

app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.ts
```diff
@@ -0,0 +1,675 @@
...
+  generateMetaWidgetChildren = ({
+    index,
+    parentId,
+    rowIndex,
+    templateWidget,
+  }: GenerateMetaWidgetChildrenProps) => {
+    const children: string[] = [];
+    let metaWidgets: MetaWidgets = {};
+
+    (templateWidget.children || []).map((childWidgetId: string) => {
```
Tooluloope, line 263: can we use `forEach` here instead of `map`

Tooluloope (changes requested):

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -167,40 +131,18 @@ class ListWidget extends BaseWidget<
       this.getMainContainer()?.bottomRow,
     );
 
-    // generate childMetaPropertyMap
-    // this.generateChildrenDefaultPropertiesMap(this.props);
-    // this.generateChildrenMetaPropertiesMap(this.props);
     this.generateChildrenEntityDefinitions(this.props);
```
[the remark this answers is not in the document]
ashit-rath: There is an autocomplete derived property which uses this. So I'm not 100% sure if we need it or not, we'd have to verify it's usage and then take the call.

Tooluloope (approved):

PR #16801, opened 2022-09-15
fix: signifiers cannot be dragged less than 4

ankurrsinghal:
signifiers cannot be dragged less than 4

### DIFF
--- a/app/client/src/components/editorComponents/DynamicHeightOverlay.tsx
+++ b/app/client/src/components/editorComponents/DynamicHeightOverlay.tsx
@@ -15,6 +15,7 @@ import {
 import { getParentToOpenIfAny } from "utils/hooks/useClickToSelectWidget";
 import { useWidgetSelection } from "utils/hooks/useWidgetSelection";
 import { BaseStyle, WidgetProps } from "widgets/BaseWidget";
+import { GridDefaults, WidgetHeightLimits } from "constants/WidgetConstants";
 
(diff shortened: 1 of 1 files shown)


app/client/src/components/editorComponents/DynamicHeightOverlay.tsx
```diff
@@ -419,7 +423,7 @@ const DynamicHeightOverlay: React.FC<DynamicHeightOverlayProps> = memo(
     }, [minDynamicHeight]);
 
     function onMinUpdate(dx: number, dy: number) {
-      if (minY + dy <= 10) {
+      if (minY + dy <= 40) {
```
riodeuno, line 426: Please use `WidgetHeightLimits.MIN_HEIGHT_IN_ROWS * GridDefaults.DEFAULT_GRID_ROW_HEIGHT`

riodeuno (changes requested):
Please use `WidgetHeightLimits.MIN_HEIGHT_IN_ROWS * GridDefaults.DEFAULT_GRID_ROW_HEIGHT`

app/client/src/components/editorComponents/DynamicHeightOverlay.tsx
ankurrsinghal: Done, but the WidgetHeightLimits.MIN_HEIGHT_IN_ROWS is still set to 2.

riodeuno (approved):

PR #16803, opened 2022-09-15
fix: Fix NGINX in start-https clashing with system conf

sharat87:
When the system NGINX conf has a `pid` directive, the additional `pid` directive in the value to `-g` argument causes a `duplicate pid directive` error.

In this fix, we _always_ specify a conf file for `nginx`, and never let it have the system config. This way, no matter what the system config contains or not contains, the behaviour of this script shouldn't be influenced by it.

### DIFF
--- a/app/client/start-https.sh
+++ b/app/client/start-https.sh
@@ -281,13 +281,19 @@ if type docker &>/dev/null; then
 fi
 
 if [[ -f $nginx_pid ]]; then
-    nginx -g "pid $nginx_pid;" -s quit
-    rm "$nginx_pid"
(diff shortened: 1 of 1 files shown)


mohanarpit (approved):

mohanarpit:
@sharat87 Minor question: Will this script work only on Mac/Linux devices or will it also work on Windows machines?

sharat87:
@mohanarpit We uncovered this bug only on WSL. So with this, yes, it'll work on Windows as well. 🙂

riodeuno (approved):

PR #16812, opened 2022-09-16
fix: Optimize reflows on the applications list page

### LINKED ISSUE #16012: [Task]: Optimize reflows on applications list page

### SubTasks

We have these two function in utils/helpers.tsx which trigger reflows.

- `isVerticalEllipsisActive`
- `isEllipsisActive`

They are used to decide if we have to show tooltips are not. I don't see the tooltips though.

**Possible solutions**

1. Always show the tooltips.
2. Since these are fixed width divs, make an approximation based on the length of the text.
3. Use the native browser tooltips by providing a title attribute.
4. If we can not remove these functions, make them async and wrap the calculation in `requestAnimationFrame` callback.
5. Figure if a tooltip needs to be shown on demand, that is on hover.?

This saves about 5-8ms per application card. If a user has 40 applications, thats about 300ms saved.

Screenshot 2022-08-15 at 3 45 41 PM

ankitakinger:
## Description

> Optimize reflows on the applications list page which is happening due to `isVerticalEllipsisActive` helper function.

Fixes #16012 

## Type of change

- Performance improvement

## How Has This Been Tested?

> Took a performance profile before and after making this change. Could see a difference in rendering time.

## Checklist:


SatishGandham (dismissed):

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/OtherUIFeatures/UpdateApplication_spec.js
```diff
@@ -111,6 +111,6 @@ describe("Update Application", () => {
       .first()
       .find(homePage.applicationCardName)
       .trigger("mouseover");
-    cy.get(".bp3-popover-target.bp3-popover-open").should("have.length", 1);
+    cy.get(".bp3-popover-target.bp3-popover-open").should("not.exist");
```
SatishGandham, line 114: Where are we checking that the application name has updated?

SatishGandham (changes requested):

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/OtherUIFeatures/UpdateApplication_spec.js
ankitakinger: I discussed this with @Parthvi12 and since we have removed the tooltip now, she asked me to add the check of it not existing.

SatishGandham (approved):

PR #16842, opened 2022-09-16
fix: Better support for disallowed hosts

sharat87:


mohanarpit:
@sharat87 Can you also make this change in `Elasticsearchplugin.java`?

mohanarpit (approved):

PR #16843, opened 2022-09-16
fix: Fix template filtering version mismatch

sharat87:
The `getReleasedVersion` method returns the current running version, _only_ if it dosen't end with `-SNAPSHOT`. If it does end with `-SNAPSHOT`, then it'll return the version from the most recent entry in the release notes.

The effect of this is that when fetching template applications, on version `v1.8.1-SNAPSHOT`, we pass in version as `v1.8.0` (which would be the most recent in release notes), which means the latest and greatest of templates won't ever be listed on prod.

The cloud-services API is actually capable of handling versions with `-SNAPSHOT` in them just fine. This server just isn't sending it.

This fix will send the current running version as is to cloud-services so we get more accurate results.


nidhi-nair:
QQ: For some more context to template apps, what does the latest running version have to do with them @sharat87 ?

sharat87:
@nidhi-nair, each template application has a `minVersion` (and soon, a `maxVersion`). This specifies that this template app is only functional on Appsmith versions above the `minVersion`. So, we need the correct value for this, from the running Appsmith, so we can show just those templates that will _work_ on their version of Appsmith.

nidhi-nair (approved):

PR #16171, opened 2022-08-20
feat: Linting in entity properties and methods

ohansFavour:
### Description

This PR introduces linting of entity's properties. Now, the linter picks up errors for invalid properties of an entity.

180985306-1748a218-6b79-49bf-b3da-f1ff7765915e

### Dependency map update timings

CleanShot 2022-09-03 at 19 55 21 2@2x

Major addition

 A new object (`unusedIdentifiersList`), that stores the list of unreferenced identifiers (unused identifiers) throughout the application. This is needed to prevent complete traversal of the dataTree when an entity or a new path is added to the dataTree while searching for references in old bindings.

Improvements.

1.   Performance

     - [x]  Prevent traversal of dataTree for every addition of entity/path while searching for references in old bindings.
     - [x]  Prevent extraction of references for all paths in the dependency map for every change.
     
###
 2. - [x] Linting in trigger fields only depends on top-level entities
 
 3. - [x] Javascript built-in classes should not be valid entity names. Eg, naming a JSObject as "Promise", makes Promise.all() unusable within it.

 4. - [x] Support linting of invalid properties

 5.   Incorrect extraction of identifiers from code  
       

         - [x] Remove all identifiers which are children of global dependencies like moment, lodash, xmlParser, forge.
         - [x] Remove all identifiers which are children of Appsmith functions like showAlert, setInterval.
         - [x] Remove all identifiers which are children of browser Apis like console.log, Object.keys, 
         - [x] Remove all member expressions which are derived from params.

         
        Eg. In code {{Api1.data.map((arg) => arg.data)}} ,  `arg.data` should not be a valid identifier.
   
        Edge case    
        - [ ] Only remove paths that are "actual" children of params
   
```
{{
function(){
 const test1  =   ["name", "school"].map((Api1)=>{
    return Api1.trim() // Api1.trim should not be a valid identifier since it's a child of param Api1
       })


 const test2 = ["name", "school"].map((detail)=>{
    return Api1.name + detail  // Api1.name should be a valid identifier since "Api1" here is not within the scope of that declared in test1
    })
   
   }()
}}

```
  
- [x] Remove all identifiers/member expressions which are derived from declared variables

Eg.

```
{{
function(){
 const test1  =   ["name", "school"].map((a)=>{
const Api1 = a.trim();
    return Api1.split("") // Api1.split should not be a valid identifier since it's a child of declared variable Api1
       })


 const test2 = ["name", "school"].map((b)=>{
    return test1.join()  // test1.join should not be a valid identifier
[...]

PR #16849, opened 2022-09-17
fix: update rts logic to use updated shared AST logic

### LINKED ISSUE #16848: [Bug]: AST logic used in the RTS folder not up to date.

### Description

The RTS APIs that exposes AST to the server are failing because code used in the RTS code has been modified due to changes in the AST logic as a result of this [PR](https://github.com/appsmithorg/appsmith/pull/16171)

### Steps To Reproduce

1. Call The endpoint `/rts-api/v1/ast/multiple-script-identifiers` in postman.
2. Observe the endpoint failing.

### Public Sample App

_No response_

### Version

Cloud

### LINKED ISSUE #17100: [Task]: Add Integration tests for RTS API and Rename AST Functions

### SubTasks

Add Jest Test cases
Add running integration test to Github workflow.

Irongade:
This PR does these things:

-  Ensures the RTS code makes use of the updated AST logic. The changes made can be found in this [PR](https://github.com/appsmithorg/appsmith/pull/16171).
- Writes Jest test cases to assert the results of the Calling the AST RTS endpoints
- Includes running the jest test cases in the github RTS workflows.

Fixes #17100 

Manual Tests

## Checklist:


app/rts/src/routes/ast_routes.ts
```diff
@@ -8,17 +8,17 @@ const astController = new AstController();
...
+  "/single-script-info",
   AstRules.getScriptValidator(),
   validator.validateRequest,
-  astController.getDependentIdentifiers
+  astController.getInfoFromScript
 );
 
 router.post(
-  "/multiple-script-identifiers",
+  "/multiple-script-infos",
```
AmanAgarwal041, line 18: We already have `multiple` attached in the route. Can we remove `s` from info?

app/rts/src/controllers/Ast/AstController.ts
```diff
@@ -46,7 +43,7 @@ export default class AstController extends BaseController {
       Promise.all(
         scripts.map(
           async (script) =>
-            await AstService.getIdentifiersFromScript(script, evalVersion)
+            await AstService.extractInfoFromScript(script, evalVersion)
```
ohansFavour, line 46: There should be a third argument here too

app/rts/src/controllers/Ast/AstController.ts
```diff
@@ -18,14 +18,11 @@ export default class AstController extends BaseController {
     super();
   }
 
-  async getDependentIdentifiers(req: Request, res: Response) {
```
nidhi-nair, line 21: Can we rename this method to something more descriptive? I am unable to understand what `info` indicates here.

app/rts/src/controllers/BaseController.ts
```diff
@@ -1,6 +1,9 @@
 import { Response } from "express";
 import { ValidationError } from "express-validator";
 import { StatusCodes } from "http-status-codes";
+import { ExtractInfoFromCode } from "@shared/ast";
+
+type Extractions = ExtractInfoFromCode;
```
ohansFavour, line 6: is there any reason for the re-assignment here?. If "Extractions" is a better name then we could change it in the @shared/ast file itself.

app/rts/src/services/AstService.ts
```diff
@@ -1,16 +1,10 @@
...
-    evalVersion
-  ): Promise<any> {
+  static async extractInfoFromScript(script, evalVersion): Promise<any> {
     return new Promise((resolve, reject) => {
       try {
-        const extractions = extractIdentifiersFromCode(
-          script,
-          evalVersion
-        );
+        const extractions = extractInfoFromCode(script, evalVersion);
```
ohansFavour, line 7: Third argument needed here too

nidhi-nair (changes requested):
Could we please add integration test cases here to make sure that CI fails if any PR tries to change the structure of how this library returns data in the future?

Also, does this change affect the API response in any way? If yes, could you please share the expected response structure in the PR description?

app/rts/src/controllers/Ast/AstController.ts
Irongade: We are returning a bunch of Information from the function, hence why we used Info, However on a second thought, we can make use of the word "Data". Data will be a good enough generic name.

Irongade:
> Also, does this change affect the API response in any way? If yes, could you please share the expected response structure in the PR description?

Expected change in the API response will follow the schema below.
 
```
 interface Extractions {
  references: string[];
  functionalParams: string[];
  variables: string[];
}
```

AmanAgarwal041:
> Could we please add integration test cases here to make sure that CI fails if any PR tries to change the structure of how this library returns data in the future?
> 
> Also, does this change affect the API response in any way? If yes, could you please share the expected response structure in the PR description?

The API routes also change.

app/rts/src/services/AstService.ts
```diff
@@ -1,15 +1,17 @@
-import { extractIdentifiersFromCode } from "@shared/ast";
```
nidhi-nair, line 1: Hey I'm sorry to nitpick, but this is still not informative. Info and data are quite generic and do not glean insight into what kind of data we expect to get here. Could we either revert to identifiers or use something more informative here?

I am guessing we will need to make this change to the ast library as well.

nidhi-nair (commented):
QQ, @Irongade could you please explain to me why we need to version control the yarn lock file?

Irongade:
> QQ, @Irongade could you please explain to me why we need to version control the yarn lock file?

@nidhi-nair, I'm not sure why yarn.lock file was version controlled as well, but my guess would be that there might be certain dependencies that we don't want changing.

.github/workflows/rts-build.yml
```diff
@@ -119,6 +119,16 @@ jobs:
           echo ::set-output name=version::$next_version-SNAPSHOT
           echo ::set-output name=tag::$(echo ${GITHUB_REF:11})
 
+      # Install all the dependencies
+      - name: Install dependencies
```
AmanAgarwal041, line 123: Build step already installs dependencies. Why do we need to install dependencies here? Can we not move this step after the build process ?

app/shared/ast/src/index.test.ts
```diff
@@ -1,85 +1,85 @@
...
-        script: 'Table5.data.map(c => ({ name: c.name }))',
-        expectedResults: ['Table5.data.map'],
+        script: "Table5.data.map(c => ({ name: c.name }))",
+        expectedResults: ["Table5.data.map"],
       },
       {
         // Literal property search
         script: "Table6['data']",
-        expectedResults: ['Table6'],
+        expectedResults: ["Table6"],
```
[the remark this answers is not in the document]
Irongade: Good, then there will be no merge conflicts. It was automatically done by Prettier.

.github/workflows/rts-build.yml
Irongade: 1. We need to install all dependencies in order to be able to run the Jest test cases, as Jest, Supertest and other dependencies are required for this step.
2. Ideally we should run jest test cases before commencing a build as this is the best practice. So I think we have to remove the extra yarn install in the build.sh file (as long as it does not break the docker deployment for RTS - I did not remove it for this reason).

AmanAgarwal041: Since both jest test cases and build are independent on each other here, we can move the running of jest test cases after build as both steps are important. We cannot remove the `yarn install` step in build.sh file as this is the file being called when the build is created.

nidhi-nair:
> I'm not sure why yarn.lock file was version controlled as well, but my guess would be that there might be certain dependencies that we don't want changing.

@Irongade I thought yarn allowed you to specify particular versions in the definition itself, no? Is that understanding incorrect? 🤔

Irongade:
> > I'm not sure why yarn.lock file was version controlled as well, but my guess would be that there might be certain dependencies that we don't want changing.
> 
> @Irongade I thought yarn allowed you to specify particular versions in the definition itself, no? Is that understanding incorrect? 🤔

Yes, but we would be hardcoding the versions in the package.json file for example `moment: 4.5.2`. It is often better to use the yarn.lock file instead to maintain versions. We make use of this pattern in the client folder as well.

AmanAgarwal041 (approved):

nidhi-nair (approved):

PR #16872, opened 2022-09-19
feat: Support row virtualization using react-window in table widget

### LINKED ISSUE #16567: [Task]: Row virtualization in Table widget

### SubTasks

Leverage react table virtualized row feature to improve performance in the table widget.

### LINKED ISSUE #17178: [Task]: Table widget performance optimisation - part 1

### SubTasks

 - [ ] Refactor autoToolTip component calculation to  improve the performance.

sbalaji1192:
Introducing the virtual row functionality to improve the performance of the table when there are a lot of rows to render.
React-window library is used to implement the virtual row functionality. This functionality will only be enabled if the cell wrapping of all the columns is disabled since the virtual expects each row of constant height.

Fixes #16567
Fixes #17178

## Type of change

## How Has This Been Tested?
Following cypress tests have been added 
1. should check that row is getting rendered.
2. should check that virtual rows are getting rendered when scrolling through the table
3. should check that the virtual rows feature is turned off when cell wrapping is enabled

## Checklist:


laveena-en:
Tested and verified for:
Row Virtualization
Regression on the table widget

app/client/src/widgets/TableWidgetV2/component/Table.tsx
```diff
@@ -360,67 +377,25 @@ export function Table(props: TableProps) {
...
-                  props.borderRadius,
-                )}
-            </div>
+              rows={subPage}
+              selectTableRow={props.selectTableRow}
+              selectedRowIndex={props.selectedRowIndex}
+              selectedRowIndices={props.selectedRowIndices}
+              tableSizes={tableSizes}
+              totalColumnWidth={totalColumnWidth}
+              useVirtual={!isCellWrappingAllowed}
```
ashit-rath, line 396: So we do not allow virtualization when any one of the columns have cell wrapping enabled?
Also should we enable only when it is server side pagination? as the client side one would always have fixed rows

app/client/src/widgets/TableWidgetV2/component/Table.tsx
```diff
@@ -228,6 +227,24 @@ export function Table(props: TableProps) {
...
+    return props.columns.reduce((prev, curr) => {
+      return prev + (curr.width || DEFAULT_COLUMN_WIDTH);
+    }, offset);
+  }, [columnString, props.multiRowSelection]);
+
+  const isCellWrappingAllowed = useMemo(() => {
+    return props.columns.some(
+      (column) => column.columnProperties.allowCellWrapping,
+    );
+  }, [columnString]);
```
ashit-rath, line 246: I don't think we need to memoise this as `isCellWrappingAllowed` is a primitive value and when it's passed to `Row` react can handle with shallow equals.

app/client/src/widgets/TableWidgetV2/component/Table.tsx
```diff
@@ -228,6 +227,24 @@ export function Table(props: TableProps) {
     [props.width],
   );
 
+  const totalColumnWidth = useMemo(() => {
```
ashit-rath, line 230: Same here, no need to memoise. The computation doesn't seem to be heavy and the return value is a primitive.

ashit-rath (commented):
@sbalaji1192 Can you update the PR description?

app/client/src/widgets/TableWidgetV2/component/Row.tsx
```diff
@@ -0,0 +1,193 @@
...
+  const isRowSelected = props.multiRowSelection
+    ? props.selectedRowIndices.includes(props.row.index)
+    : props.row.index === props.selectedRowIndex;
+  return (
+    <div
+      {...rowProps}
+      className={"tr" + `${isRowSelected ? " selected-row" : ""}`}
+      data-rowindex={props.index}
+      key={props.index}
+      onClick={(e) => {
```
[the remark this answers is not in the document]
sbalaji1192: This code is already part of an existing component and I'm just extracting it to create a row component. We're not adding accessibility for now.

app/client/src/widgets/TableWidgetV2/component/Row.tsx
```diff
@@ -0,0 +1,193 @@
...
+  (props: RowPropsType, ref: Ref<HTMLDivElement>) => {
+    return (
+      <div {...props.getTableBodyProps()} className="tbody" ref={ref}>
+        {props.rows.map((row, index) => {
+          return (
+            <BasicRow
+              accentColor={props.accentColor}
+              borderRadius={props.borderRadius}
+              index={index}
+              key={index}
```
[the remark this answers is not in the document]
sbalaji1192: No. It's a plain object.

app/client/src/widgets/TableWidgetV2/component/Table.tsx
sbalaji1192: Yeah. good point. Let me change the condition here.

app/client/src/widgets/TableWidgetV2/component/Table.tsx
sbalaji1192: We are memoizing it to avoid repeated calculations.

ashit-rath: I don't think the calculations are very expensive as the number of columns cannot be that very high. 
Also the `some` does early exit if it encounters a `true`

sbalaji1192: I have removed it.

app/client/src/widgets/TableWidgetV2/component/Table.tsx
```diff
@@ -228,6 +225,11 @@ export function Table(props: TableProps) {
     [props.width],
   );
 
+  const shouldUseVirtual =
+    !props.columns.some(
+      (column) => !!column.columnProperties.allowCellWrapping,
+    ) && props.serverSidePaginationEnabled;
```
ashit-rath, line 231: Can we have `props.serverSidePaginationEnabled` as the first condition and then the `columns.some`? 
If sever side is disabled then the 2nd condition doesn't need to run.

app/client/src/widgets/TableWidgetV2/component/cellComponents/AutoToolTipComponent.tsx
```diff
@@ -39,14 +39,13 @@ function useToolTip(
   const [showTooltip, updateToolTip] = useState(false);
 
   useEffect(() => {
-    requestAnimationFrame(() => {
```
ashit-rath, line 42: Why did we remove `requestAnimationFrame`?

app/client/src/widgets/TableWidgetV2/component/cellComponents/EmptyCell.tsx
```diff
@@ -41,26 +45,18 @@ export const renderEmptyRows = (
...
-                );
-              })}
-            </EmptyRow>
-          );
-        })}
-      </>
-    );
+    return rows.map((row: string, index: number) => {
+      return (
+        <EmptyRow className="tr" key={index} style={style}>
```
ashit-rath, line 50: Is it possible to use `primary column` value for the `key` here? and if it's not present then as a fallback the `index`?

app/client/src/widgets/TableWidgetV2/component/cellComponents/AutoToolTipComponent.tsx
sbalaji1192: requestAnimationFrame is creating an opposite effect here. its triggering a layout thrashing for each cell in the Table. hence we removed it.
Screenshot 2022-09-29 at 10 47 43 AM

app/client/src/widgets/TableWidgetV2/component/cellComponents/EmptyCell.tsx
sbalaji1192: This is an empty row we render to fill the table body. it doesn't have any data.

ashit-rath (dismissed):

laveena-en:
Tested and verified:

-Performance testing with large number of records (1000 records, 40 cols)
-Regression on table widget.
-Functionality in view and edit mode.

ashit-rath (approved):

PR #16883, opened 2022-09-20
fix: Make projection fields refresh when table heading row index selection is updated

### LINKED ISSUE #16868: [Bug]: Column filtering does not work when there are multiple row headings on a GSheet query

### Description

Column filters can only be applied to the first row on a Google Sheet query and even though columns show up on column filter dropdown, we cannot use them since running such a query returns an error as seen in the video attached

### Steps To Reproduce

1. Use a Google Sheet containing multiple header rows as seen in the video attached
2. Use dropdown option to filter data and observe that only when header index is `Row 1` would the query be successfully executed but the query does not execute when row is anything other than `Row 1` (see video)

https://www.loom.com/share/fa33edf500ad49ab858a8d3cbf1a1c29

### Public Sample App

_No response_

### Version

Cloud

nidhi-nair:
## Description

We weren't sending table heading row index information along with the sheet url and name parameters to fetch the columns in Google Sheets integration. This PR starts to send this information along. As a result, every time the table heading row index value changes, we can expect the projection field to get updated.

Fixes #16868 

## Type of change

## How Has This Been Tested?
- Manually tested on local, no possibility of unit tests or integration tests for GSheets.

## Checklist:

### DIFF
--- a/app/server/appsmith-plugins/googleSheetsPlugin/src/main/resources/editor/fetch_many.json
+++ b/app/server/appsmith-plugins/googleSheetsPlugin/src/main/resources/editor/fetch_many.json
@@ -17,16 +17,17 @@
       "fetchOptionsConditionally": true,
       "alternateViewTypes": ["json"],
       "conditionals": {
-        "enable": "{{!!actionConfiguration.formData.sheetUrl.data && !!actionConfiguration.formData.sheetName.data}}",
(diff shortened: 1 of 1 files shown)


sumitsum (approved):

nidhi-nair:
Added test plan approved label since QA has signed off on issue and automation is not possible.

PR #16912, opened 2022-09-20
fix: workspace email website validation

### LINKED ISSUE #8970: [Bug]: Email and website format is not validated on Org settings page

### Current Behavior

User can type in an incorrect email or website address of any format on Org settings page and this is not currently validated by the system. But on refreshing the page this invalid value is not saved (field is empty on page refresh) for email but incorrect website value is saved.

On prod, if there was a valid email format before the invalid email was entered then on refresh the valid format replaces r recent change.

### Steps To Reproduce

1. Go to Org settings from home page
2. Enter an incorrect format for email/website (with spaces for example) and observe the no validation error displayed to a user
3. Refresh page and observe behavior as mentioned above

### Environment

Release

### Version

Cloud

nsarupr:
## Description

Adding Validation on the Workspace's Email and Website on Backend APIs

Fixes #8970

## Type of change

## How Has This Been Tested?

- testValidateIncomingWorkspaceValidEmail
- testValidateIncomingWorkspaceInvalidEmail
- testValidateIncomingWorkspaceValidWebsite
- testValidateIncomingWorkspaceInvalidWebsite

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/WorkspaceServiceCEImpl.java
```diff
@@ -580,6 +586,16 @@ public Mono<Workspace> archiveById(String workspaceId) {
         });
     }
 
+    @Override
+    public void validateIncomingWorkspace(Workspace workspace){
```
AnaghHegde, line 590: Why do we need this to expose this? If it is just used in this service class can we make it private?

app/server/appsmith-server/src/main/java/com/appsmith/server/constants/PatternConstants.java
```diff
@@ -0,0 +1,7 @@
+package com.appsmith.server.constants;
+
+public class PatternConstants {
+    public final static String WEBSITE_PATTERN = "^(http://|https://)?(www.)?(([a-z0-9\\-]+)\\.)+([a-z]+)$";
```
AnaghHegde, line 4: Can we please add a comment here with valid and invalid cases? it would be hard to understand the use case just by looking at the regex.

AnaghHegde (commented):
Minor comments. Rest looks good to me.

app/server/appsmith-server/src/main/java/com/appsmith/server/constants/PatternConstants.java
nsarupr: I have added the test cases for users to look at.
I have covered all the valid and invalid use cases.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/WorkspaceServiceCEImpl.java
AnaghHegde: IMO, we can do that change when required. For now, I would say let's keep it private.

app/server/appsmith-server/src/main/java/com/appsmith/server/constants/PatternConstants.java
AnaghHegde: IMO, Test cases are to ensure that we covered all the cases. But comments are usually added in the actual implementation. Adding test cases would not justify not adding comments.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/WorkspaceServiceCEImpl.java
nsarupr: Although, won't be able to invoke it from the test case then.
In that case, we'll have to test the Update function of the repository directly.

AnaghHegde (dismissed):

AnaghHegde (approved):

AnaghHegde (dismissed):

AnaghHegde (approved):

PR #16914, opened 2022-09-20
fix: auto height limits scroll issue

ankurrsinghal:
Tried solving the scroll issue for widgets when there are limits

### DIFF
--- a/app/client/src/widgets/BaseWidget.tsx
+++ b/app/client/src/widgets/BaseWidget.tsx
@@ -49,6 +49,7 @@ import log from "loglevel";
 import { CanvasWidgetStructure } from "./constants";
 import { DataTreeWidget } from "entities/DataTree/dataTreeFactory";
 import Skeleton from "./Skeleton";
+import styled from "styled-components";
 
 /***
  * BaseWidget
@@ -62,6 +63,10 @@ import Skeleton from "./Skeleton";
  * 3) Call actions in widgets or connect the widgets to the entity reducers
  *
  */
+
+const DynamicHeightContainer = styled.div<{ isOverflow?: boolean }>`
+  overflow-y: ${(props) => (props.isOverflow ? "auto" : "unset")};
+`;
 
 abstract class BaseWidget<
   T extends WidgetProps,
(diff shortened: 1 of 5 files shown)


app/client/src/widgets/BaseWidget.tsx
```diff
@@ -63,6 +64,54 @@ import Skeleton from "./Skeleton";
  *
  */
 
+const StyledDynamicHeightContainer = styled.div<{ isOverflow?: boolean }>`
```
riodeuno, line 67: Please put this component in its own file.

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -63,6 +64,54 @@ import Skeleton from "./Skeleton";
...
+      }
+    };
+  }, [observer]);
+
+  const expectedHeightInRows = Math.ceil(
+    expectedHeight / GridDefaults.DEFAULT_GRID_ROW_HEIGHT,
+  );
+
+  return (
+    <StyledDynamicHeightContainer
```
riodeuno, line 105: Can we re-use the StyledDynamicHeightContainer instead of creating another div within it?

app/client/src/widgets/CheckboxGroupWidget/component/index.tsx
```diff
@@ -198,7 +198,7 @@ const CheckboxGroupComponent = React.forwardRef<
     optionCount += 1;
   }
 
-  const finalComponent = (
+  return (
```
riodeuno, line 201: This is beautiful! 😀 . We've cleaned up the components.

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -63,6 +64,54 @@ import Skeleton from "./Skeleton";
...
+}: {
+  children: ReactNode;
+  maxDynamicHeight: number;
+}) {
+  const [expectedHeight, setExpectedHeight] = useState(0);
+
+  const ref = useRef<HTMLDivElement>(null);
+
+  const observer = React.useRef(
+    new ResizeObserver((entries) => {
```
riodeuno, line 83: ResizeObserver seems like a valid solution for now. 

Even with the resizeobserver, the HOC seems like a good idea to separate the BaseWidget logic from the DynamicHeight logic. If I understand correctly, we're adding `DynamicHeightContainer` over all the other wrappers for `content`, in that case, we can safely use an HOC. Let's try to have the HOC created, once you've dealt with other pressing tasks.

riodeuno (changes requested):

app/client/src/widgets/BaseWidget.tsx
ankurrsinghal: We cannot, because we need to get the height of the component, I don't know if we can get it from the children prop, but using this, we can remove the dependency on isDynamicEnabled to set height auto in the widget component itself.

app/client/src/widgets/BaseWidget.tsx
ankurrsinghal: Cool.

app/client/src/widgets/BaseWidget.tsx
ankurrsinghal: Done

riodeuno (approved):

Issue #16590, opened 2022-09-06
[Bug]: Input Widgets with Auto height enabled when entering longer text, contents are not fully hugged


Sripriya93:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SubTasks

Drag and drop any of the Input widgets (Input, Phone Input, Currency Input)
Select Auto height in "Height" property 
In the "Label" property enter a longer text with more than four or five lines
Observe all the contents are not inside the border of the widget

Also when we move the widget downwards, we also observe that the contents are moved inside or the auto height is being increased for the widget. But then when we drag and drop another widget we are facing the same issue.

https://user-images.githubusercontent.com/104895602/188643207-f1323186-72a7-4344-a130-3c88b5fd9d75.mp4

ankurrsinghal:
This issue has been resolved in this PR, https://github.com/appsmithorg/appsmith/pull/16943

Sripriya93:
Working fine. Moving to done.
![image](https://user-images.githubusercontent.com/104895602/191473642-525e4728-e3b1-4822-a5bf-4843df2ac8b4.png)

PR #16943, opened 2022-09-21
fix: auto height input label oob

ankurrsinghal:
Auto height input label out of border problem mentioned in bug #16590


Sripriya93:
Working fine. Moving to done

image

PR #16949, opened 2022-09-21
fix: auto height text widget hide overflow

ankurrsinghal:
Hiding the Overflow options in the Text widget if Auto Height is enabled.


aswathkk (approved):

riodeuno (approved):

riodeuno:
@ankurrsinghal @Sripriya93 Adding a note regarding the feature missing from this:

The value of overflow property should become no-overflow when dynamic height is enabled, this is so that the dynamic height feature works correctly.

PR #16951, opened 2022-09-21
fix: Fix X-Forwarded-Host with multiple rev-proxies

sharat87:
When there's another reverse proxy in front of Appsmith container, in addition to the NGINX that comes with Appsmith, the Forwarded host is not carried over.

## Steps to reproduce

Start a new fat container with

```bash
docker run --rm -p 8001:80 -v "$PWD"/stacks-ee:/appsmith-stacks -e APPSMITH_OAUTH2_GOOGLE_CLIENT_ID="dummy-id" -e APPSMITH_OAUTH2_GOOGLE_CLIENT_SECRET="dummy-secret" -d --pull always appsmith/appsmith-ee
```

Now, once that is up, run the following curl command to see where the Google OAuth button would take you:

```bash
curl -sSI -H 'X-Forwarded-Host: another.com' -H 'X-Forwarded-Proto: https' http://localhost:8001/oauth2/authorization/google | grep -Eo 'redirect_uri=[^&]*'
```

The `redirect_uri` doesn't have the correct host. We expect `another.com`, but we see `localhost`.

### DIFF
--- a/deploy/docker/templates/nginx/nginx-app-http.conf.template.sh
+++ b/deploy/docker/templates/nginx/nginx-app-http.conf.template.sh
@@ -13,6 +13,12 @@ map \$http_x_forwarded_proto \$origin_scheme {
   default \$http_x_forwarded_proto;
   '' \$scheme;
 }
+
+map \$http_x_forwarded_proto \$origin_host {
+  default \$http_x_forwarded_host;
+  '' \$host;
+}
+
 # redirect log to stdout for supervisor to capture
 access_log /dev/stdout;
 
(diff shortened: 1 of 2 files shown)


sharat87:
Fix for issue reported in <https://community.appsmith.com/t/1361/11>.

mohanarpit (dismissed):

mohanarpit (approved):

trishaanand:
@sharat87 : Should we request @RakshaKShetty for a quick sanity with SAML fat container with custom domain?

sharat87:
@trishaanand, I'll apply the fix on the test instance directly and we can verify there. The test case needs a load balancer as well, otherwise we won't see the problem.

sharat87:
@trishaanand, I'm going to merge this after the freeze is lifted. Let's test the SAML scenario with the release Docker Image. Sound okay?

PR #16971, opened 2022-09-22
fix: Slow rendering of JSON and RAW response

### LINKED ISSUE #16705: [Bug]: API response JSON rendering is too slow

### Description

1. run api call on `https://restcountries.com/v3.1/all`
2. expand the bottom drawer
3. see that the response is truncated, and takes a long time to render in full
PS: this problem also exists for graphql

### Steps To Reproduce

https://user-images.githubusercontent.com/94514895/189814756-6b4c9fc3-62d9-467d-9d08-af66012fe629.mov

### Public Sample App

_No response_

### Version

Cloud

ChandanBalajiBP:
## Description

**API response JSON/RAW rendering is too slow.**

Codemirror editor was not rendering the required lines of code on increasing the viewport. ([video](https://github.com/appsmithorg/appsmith/issues/16705#issue-1370868446)).
Codeeditor upon the first render is set to 100% of container height. And editor also renders x extra lines above and below the viewport. (this is can be set by updating viewportMargin parameter in CodeEditor/index.tsx - now set to 10). Codemirror is using Observers to achieve this.

Solution 1

> Can increase viewportMargin to 50/100. Won't have much impact [performance](https://github.com/appsmithorg/appsmith/issues/16705#issuecomment-1248969534). 
But the approach is not an ideal solution. Because what if the screen is large and 100 is also fewer lines of code??

Solution 2

> Codemirror has viewportChange event, but the problem here is that the viewport of the container is resizing but not the editor itself.

Solution 3

> Refresh editor on resize of the container. This is currently implemented method. (edtior.refresh() is the function used to refresh)

Fixes #16705 

## Type of change

## Checklist:


app/client/src/components/editorComponents/ApiResponseView.tsx
```diff
@@ -265,6 +265,7 @@ export const responseTabComponent = (
   return {
     [API_RESPONSE_TYPE_OPTIONS.JSON]: (
       <ReadOnlyEditor
+        containerHeight={tableBodyHeight}
```
mohanarpit, line 268: QQ: Since `tableBodyHeight` is a nullable parameter, what happens if the value passed to this function is null?

I'm simply asking this question to further my understanding of how React works :)

ChandanBalajiBP: 1. The size of the editor is set to 100% in the constructor. So would render without any issue. 
2. This parameter is only considered when the container size is altered. So the editor will refresh only if the parameter is not null and has increased from its previous value.

app/client/src/components/editorComponents/CodeEditor/index.tsx
```diff
@@ -371,6 +372,15 @@ class CodeEditor extends Component<Props, State> {
         // handles case when inputValue changes from a truthy to a falsy value
         this.editor.setValue("");
       }
+
+      if (
+        prevProps.containerHeight &&
+        this.props.containerHeight &&
+        prevProps.containerHeight < this.props.containerHeight
```
ayushpahwa, line 379: Should we be debouncing this action? Instead of allowing refresh on every change, maybe we can reduce the number of calls to only when the user is done editing or maybe debounce it by some ms. Also, what are the performance impacts of this PR?

ChandanBalajiBP: I completely agree with you on debouncing. I will update the same. There is no change in performance w.r.t to loading of the component. And no observable performance impact on resizing.

app/client/src/components/editorComponents/CodeEditor/index.tsx
```diff
@@ -349,7 +350,21 @@ class CodeEditor extends Component<Props, State> {
     return true;
   }
 
+  //Debounce editor refresh request as container resizing triggers many change events.
+  debounceEditorRefresh = _debounce(async () => {
+    console.log("ondhu 3");
```
ayushpahwa, line 355: Please remove any console statements

app/client/src/components/editorComponents/CodeEditor/index.tsx
```diff
@@ -22,7 +22,7 @@ import "codemirror/addon/lint/lint.css";
 import { getDataTreeForAutocomplete } from "selectors/dataTreeSelectors";
 import EvaluatedValuePopup from "components/editorComponents/CodeEditor/EvaluatedValuePopup";
 import { WrappedFieldInputProps } from "redux-form";
-import _, { isString } from "lodash";
+import _, { isString, debounce as _debounce } from "lodash";
```
ayushpahwa, line 25: Please remove the import of the whole `_` from lodash and only use the functions that are needed.

ayushpahwa (changes requested):

app/client/src/components/editorComponents/CodeEditor/index.tsx
ChandanBalajiBP: @ayushpahwa  Codemirror editor linter is using the whole lodash package in getAnnotations function. (line num~248). So can we use _ in this case?

ayushpahwa (approved):

PR #16977, opened 2022-09-22
fix: MongoPlugin smart substitution for regex arguments

### LINKED ISSUE #11880: [Bug]-[1408]:Mongo plugin- smart substitution fails with $regex keyword when input is number

### Description

From community:
```
{$match: {
  $or:[
  {OrderNumber:{$regex:"{{orders_table.searchText}}"}}
  ]
}} 

Error: error $regex has to be a string. 

can you please share what evaluated value is being substituted for {{orders_table.searchText}}
If there is something in the search box will be “69” other wise “”
```

This issue was resolved upon turning off smart BSON substitution.

### Steps To Reproduce

.

### Public Sample App

_No response_

### Version

cloud

sondermanish:
## Description

Fixes #11880 

Server side smart substitution parses js bounded (numbers/string representation of number arguments) as  **whole numbers**, and **decimal numbers** regardless if they are placed in quotes **" "**. This feature creates a bug for mongo regex substitution as **mongo regex attributes** only takes string arguments.  This PR fixes before mentioned characteristic.

## Type of change

> Bug Fix: Now, digit/ number arguments could be provided to the $regex with smart substitution turned on. 

## How Has This Been Tested?

The backend changes has been tested:
- Junit test cases
- Manually

## Checklist:

### DIFF
--- a/app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
+++ b/app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
@@ -157,6 +157,14 @@ public class MongoPlugin extends BasePlugin {
     private static final String MONGO_URI_REGEX = "^(mongodb(?:\\+srv)?://)(?:(.+):(.+)@)?([^/?]+)/?([^?]+)?\\??(.+)?$";
 
     /**
+     * We use this regex to identify the $regex attribute and the respective argument provided:
+     * e.g. {"code" : {$regex: value, $options: value}} / {"code" : {$regex: value}}
+     * capturing the value group for regex.
+     * e.g {"code" : {$regex: 8777}}, this whole substring will be matched and 8777 is captured for further processing.
+     */
+    private static final String mongo$regexWithNumberIdentifier = ".*\\{[\\s\\n]*\\$regex[\\s\\n]*:([\\s\\n]*(?:(?:\\\"[/]*)|(?:/[\\\"]*)|)[\\d]*(?:(?:[/]*\\\")|(?:[\\\"]*/)|)[\\s\\n]*)(?:,|\\})";
+
+    /**
      * We use this regex to find usage of special Mongo data types like ObjectId(...) wrapped inside double quotes
      * e.g. "ObjectId(...)". Case for single quotes e.g. 'ObjectId(...)' is not added because the way client sends
      * back the data to the API server it would be extremely uncommon to encounter this case.
@@ -574,7 +582,94 @@ public class MongoPlugin extends BasePlugin {
                     params,
                     parameters);
 
+            updatedQuery = mongoRegexSmartSubstitutionHelper (updatedQuery);
+
             return updatedQuery;
+        }
+
+
+        private static StringBuilder makeValidForRegex(String value) {
+
+            // equivalent to new StringBuilder(value.trim()); but takes lesser memory
+            StringBuilder valueSB = trimAndBuildStringBuilderFromString(value);
+
+            char quote = '\"';
+            char forwardSlash = '/';
+            if ( valueSB.charAt(0) != quote && valueSB.charAt(0) != forwardSlash) {
+                valueSB.insert(0, quote );
+                // adds quote at the end of the line
+                valueSB.append(quote);
+            }
+
+            return valueSB;
+        }
+
+        private static StringBuilder trimAndBuildStringBuilderFromString(String input) {
+
+            int startIndex = 0;
+            int endIndex = input.length() -1;
+            for (int i = 0; i < input.length(); i++) {
+                if (!Character.isWhitespace(input.charAt(i))) {
+                    startIndex = i;
+                    break;
+                }
+            }
+
+            for  (int i = input.length()-1; i>= 0 ; i--) {
+                if (!Character.isWhitespace(input.charAt(i))) {
+                    endIndex = i;
+                    break;
+                }
+            }
+
+            return  endIndex >= startIndex ? new StringBuilder( input.substring(startIndex, endIndex+1)) : new StringBuilder("");
+        }
+
+
+        /**
+         * Smart Substitution Helper for mongo regex: Regex takes pattern as "<pattern>" or /<pattern>/,
+         * Rest are not illegal arguments to the $regex attribute in mongo.
+         * @Param: Smart substituted query string.
+         * @Returns: Updated query string.
+         */
+        private static String mongoRegexSmartSubstitutionHelper(String inputQuery) {
+
+
+
+            int startIndex = 0;
+            StringBuilder inputQuerySB = new StringBuilder();
+
+            Pattern mongo$regexIdentifierPattern = Pattern.compile(mongo$regexWithNumberIdentifier);
+            Matcher mongo$regexIdentifierMatcher = mongo$regexIdentifierPattern.matcher(inputQuery);
+
+            while (mongo$regexIdentifierMatcher.find()) {
+
+                StringBuilder valueSB;
+                try {
+                    valueSB = makeValidForRegex(mongo$regexIdentifierMatcher.group(1));
+
+                } catch(StringIndexOutOfBoundsException e) {
+                    // when the match group detected is of length zero this error is thrown;
+                    return inputQuery;
+                } catch (Exception e) {
+                    return inputQuery;
+                }
+
+
+                //groups are discovered in greedy manner hence it's sorted by default
+                // a sub-group within a big group could be due to the regex arguments provide by user hence will not parse that
+                if (startIndex > mongo$regexIdentifierMatcher.start()) {
+                    // the matcher walks greedily to find the pattern, if there is a group overlapping with other group means that it's a subquery, and it's not meant to be parsed.
+                    continue;
+                }
+                inputQuerySB.append(inputQuery.substring(startIndex, mongo$regexIdentifierMatcher.start(1)));
+                inputQuerySB.append(valueSB);
+                startIndex = mongo$regexIdentifierMatcher.end(1);
+            }
+
+            inputQuerySB.append(inputQuery.substring(startIndex));
+
+            return inputQuerySB.toString();
         }
 
         /**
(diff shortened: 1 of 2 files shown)


app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
```diff
@@ -574,9 +582,96 @@ private String smartSubstituteBSON(String rawQuery,
...
+                valueSB.insert(0, quote );
+                // adds quote at the end of the line
+                valueSB.append(quote);
+            }
+
+            return valueSB;
+        }
+
+        private static StringBuilder trimAndBuildStringBuilderFromString(String input) {
+
```
subrata71, line 608: There are a lot of instances where `String.trim` function is used. And looking at the implementation of trim function I found that there are two different trim functions used based on the coder e.g. `Latin1/UTF16`. I would vote for using the `String.trim` function. What do you think?

app/server/appsmith-plugins/mongoPlugin/src/test/java/com/external/plugins/MongoPluginTest.java
```diff
@@ -2848,5 +2848,84 @@ public void testValidateDatasource_withoutDefaultDBInURIString_returnsInvalid()
...
+        Mono<ActionExecutionResult> actionExecutionResultMono = dsConnectionMono.flatMap(clientConnection -> pluginExecutor.executeParameterized(clientConnection,
+                executeActionDTO,
+                dsConfig,
+                actionConfiguration));
+
+        StepVerifier.create(actionExecutionResultMono)
+                .assertNext(actionExecutionResult -> {
+                    assertNotNull(actionExecutionResult);
+                    assertTrue(actionExecutionResult.getIsExecutionSuccess());
+                    assertEquals(0, ((ArrayNode) actionExecutionResult.getBody()).size());
```
subrata71, line 2922: You are asserting size with 0. This doesn't essentially prove that regex with Number as value works. Can you populate some data first and apply the number regex? Please make sure that at least one document is returned by matching with that regex. This will also verify that the implementation is correct.

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
```diff
@@ -574,9 +582,96 @@ private String smartSubstituteBSON(String rawQuery,
...
+
+
+            int startIndex = 0;
+            StringBuilder inputQuerySB = new StringBuilder();
+
+            Pattern mongo$regexIdentifierPattern = Pattern.compile(mongo$regexWithNumberIdentifier);
+            Matcher mongo$regexIdentifierMatcher = mongo$regexIdentifierPattern.matcher(inputQuery);
+
+            while (mongo$regexIdentifierMatcher.find()) {
```
subrata71, line 645: Can you please state the reason behind calling `mongo$regexIdentifierMatcher.find()` multiple times? Could there be multiple occurrences of regex in a single query?

sondermanish: There could be multiple occurrences of `$regex` in different part of the same query. i.e `{"attribute":{$regex: /value/ }, "attribute": {$regex}:/value/}`, hence we require to match every such occurrences possible.

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
sondermanish: String.trim() gives us another object of string and then we are using it to build `StringBuilder` object, hence it will take tiny amount of extra memory for same computation, however the function I've used reduces that extra string object.
This however is very minor. Let me know if you want me to use `new StringBuilder(string.trim())`.

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
```diff
@@ -574,9 +582,96 @@ private String smartSubstituteBSON(String rawQuery,
                     params,
                     parameters);
 
+            updatedQuery = mongoRegexSmartSubstitutionHelper (updatedQuery);
```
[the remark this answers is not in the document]
sondermanish: Changing as suggested.

app/server/appsmith-plugins/mongoPlugin/src/test/java/com/external/plugins/MongoPluginTest.java
sondermanish: changing as suggested.

app/server/appsmith-plugins/mongoPlugin/src/test/java/com/external/plugins/MongoPluginTest.java
```diff
@@ -2848,5 +2848,84 @@ public void testValidateDatasource_withoutDefaultDBInURIString_returnsInvalid()
...
+                    assertNotNull(actionExecutionResult);
+                    assertTrue(actionExecutionResult.getIsExecutionSuccess());
+                    assertEquals(1, ((ArrayNode) actionExecutionResult.getBody()).size());
+                    assertEquals(List.of(new ParsedDataType(JSON), new ParsedDataType(RAW)).toString(), actionExecutionResult.getDataTypes().toString());
+                })
+                .verifyComplete();
+
+
+
+
```
mohanarpit, line 2889: Nit: Why do we have so many empty lines here? Can we please format the entire code to ensure that we are all following the same code guidelines?

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
subrata71: Yes, I understand your concern. But it would have an impact if the string size is much larger. And also the Latin1/UTF16 coder being used in the implementation of the trim function may have some other cases that we might miss if we go with our own implementation.
@sondermanish

app/server/appsmith-plugins/mongoPlugin/src/test/java/com/external/plugins/MongoPluginTest.java
sondermanish: removing the empty lines and formatting the code.

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
sondermanish: changing as suggested.

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
```diff
@@ -156,6 +156,14 @@ public class MongoPlugin extends BasePlugin {
      */
     private static final String MONGO_URI_REGEX = "^(mongodb(?:\\+srv)?://)(?:(.+):(.+)@)?([^/?]+)/?([^?]+)?\\??(.+)?$";
 
+    /**
+     * We use this regex to identify the $regex attribute and the respective argument provided:
+     * e.g. {"code" : {$regex: value, $options: value}} / {"code" : {$regex: value}}
+     * capturing the value group for regex.
+     * e.g {"code" : {$regex: 8777}}, this whole substring will be matched and 8777 is captured for further processing.
+     */
+    private static final String mongo$regexWithNumberIdentifier = ".*\\{[\\s\\n]*\\$regex[\\s\\n]*:([\\s\\n]*(?:(?:\\\"[/]*)|(?:/[\\\"]*)|)[\\d]*(?:(?:[/]*\\\")|(?:[\\\"]*/)|)[\\s\\n]*)(?:,|\\})";
```
subrata71, line 165: By this regex we are trying to match with an integer number only, right? Couldn't there be cases where non-integer numbers can also be part of the regex?

subrata71: Also, negative integers are not matched by this regex.

subrata71: Can you add this in the comment above it regarding what kind of values it can cover through this regex? @sondermanish

sondermanish: In that case the Smart substitution already sends those as strings as when the data type is being decided in `DataTypeStringUtils.jsonSmartReplacementPlaceholderWithValue().`

sondermanish: I can write test case to verify the same, or modify the existing test case for the same.

subrata71: Yes, but it removes the quotes around it. So you are left with a value like "{\"code\" : {$regex: -4, $options: value}}"; which is not matched by the number regex. Please correct me if I am wrong.

@sondermanish

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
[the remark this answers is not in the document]
sondermanish: @subrata71 changed as suggested

app/server/appsmith-plugins/mongoPlugin/src/main/java/com/external/plugins/MongoPlugin.java
sondermanish: modified as suggested

app/server/appsmith-plugins/mongoPlugin/src/test/java/com/external/plugins/MongoPluginTest.java
sondermanish: modified as suggested

subrata71 (approved):
Looks good to me.

PR #16983, opened 2022-09-22
fix: Fix escaping single quotes in env variable values

### LINKED ISSUE #16987: [Bug]: single quotes being escaped in env variables
@sharat87 Sending a value with quotes to the api(v1/admin/env) is resulting in quotes getting escaped. Added some points below which describes the issue.

1) 
https://user-images.githubusercontent.com/67054171/191439609-edad0640-93dd-4063-9cda-0a1a1b012e4e.mov

none

If I send the value as `'none'` from the client it stores it as `none` without the quotes in the env file. Checking on the browser also shows `none` without quotes. I believe it should be `APPSMITH_ALLOWED_FRAME_ANCESTORS="'none'"` instead.

2) Same case with sending `'self'`

3) Additionally if I send the value as `'self' release.app.appsmith.com` the env file stores it as 
release

In this case it does get applied properly though when I check the same in the browser

browser

4) This seems the case with all the fields but in case of APPSMITH_ALLOWED_FRAME_ANCESTORS we probably have to preserve/not escape  the quotes.

If you want the frontend to debug it is at https://github.com/appsmithorg/appsmith/pull/16629

_Originally posted by @akash-codemonk in https://github.com/appsmithorg/appsmith/issues/16597#issuecomment-1253345359_

sharat87:
Fixes #16987

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/EnvManagerCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/EnvManagerCEImpl.java
@@ -189,19 +189,25 @@ public class EnvManagerCEImpl implements EnvManagerCE {
                         return line;
                     }
                     remainingChangedNames.remove(name);
-                    String safeValue = changes.get(name);
-                    if (safeValue.contains(" ") || safeValue.contains("?") || safeValue.contains("*") || safeValue.contains("#")) {
-                        safeValue = "'" + safeValue.replace("'", "'\"'\"'") + "'";
-                    }
-                    return String.format("%s=%s", name, safeValue);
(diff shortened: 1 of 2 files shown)


mohanarpit (approved):

sharat87:
@akash-codemonk, thanks for checking. The first part is fine. This is the best way this can be predictably escaped. The problem with just wrapping the value with `"`, to get `"'none'"` is that, if the value contains a `$` character, there can be unintended variable interpolation. But since we use single quotes, `$` characters don't have any special meaning to them.

The second part is a problem. Let me take a look and fix. Thanks.

sharat87:
@akash-codemonk, the second point should now be fixed. Thanks.

sharat87:
@mohanarpit, can you approve again please? 🙂

mohanarpit (approved):

PR #16988, opened 2022-09-22
fix:auto height container container

ankurrsinghal:
Removed the scroll container for container-type widgets.

### DIFF
--- a/app/client/src/widgets/BaseWidget.tsx
+++ b/app/client/src/widgets/BaseWidget.tsx
@@ -554,7 +554,13 @@ abstract class BaseWidget<
 
   getCanvasView(): ReactNode {
     let content = this.getPageView();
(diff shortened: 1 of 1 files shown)


app/client/src/widgets/BaseWidget.tsx
```diff
@@ -554,7 +554,13 @@ abstract class BaseWidget<
 
   getCanvasView(): ReactNode {
     let content = this.getPageView();
-    if (isDynamicHeightEnabledForWidget(this.props)) {
+    if (
+      isDynamicHeightEnabledForWidget(this.props) &&
+      this.props.type !== "CONTAINER_WIDGET" &&
```
riodeuno, line 559: Did `this.props.children > 0` is not working?

riodeuno (approved):

riodeuno (approved):

PR #16995, opened 2022-09-22
Modal widget adheres to dynamic height

riodeuno:
- This PR enables dynamic height for modal widget

### DIFF
--- a/app/client/src/sagas/dynamicHeightSagas.ts
+++ b/app/client/src/sagas/dynamicHeightSagas.ts
@@ -14,7 +14,7 @@ import {
   GridDefaults,
   MAIN_CONTAINER_WIDGET_ID,
 } from "constants/WidgetConstants";
-import { groupBy } from "lodash";
+import { groupBy, uniq } from "lodash";
 import log from "loglevel";
 import {
   CanvasWidgetsReduxState,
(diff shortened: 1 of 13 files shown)


app/client/src/sagas/dynamicHeightSagas.ts
```diff
@@ -677,7 +715,7 @@ export default function* widgetOperationSagas() {
       batchCallsToUpdateWidgetDynamicHeightSaga,
     ),
     debounce(
-      200,
+      500,
```
riodeuno, line 718: Revert this change. 
I had added this to fix the undo-redo issue, but this doesn't help, and even if it did, this is not the right approach.

app/client/src/sagas/dynamicHeightSagas.ts
```diff
@@ -500,22 +512,44 @@ export function* updateWidgetDynamicHeightSaga() {
         },
       ];
     }
+  }
 
-    log.debug("Dynamic height: Widgets to update:", { widgetsToUpdate });
+  if (Object.keys(independentUpdates).length > 0) {
+    Object.keys(independentUpdates).forEach((updatedWidgetId) => {
```
rahulramesha, line 518: Minor NIT: In between Object.keys and forEach, it has looping through independentUpdates 3 times. This i feel can be done in a single loop

rahulramesha (approved):

app/client/src/sagas/dynamicHeightSagas.ts
riodeuno: Absolutely. Let me add a todo for this. I'll fix this today in a separate PR.

PR #17000, opened 2022-09-23
feat: meta widget management

### LINKED ISSUE #16583: Task: List v2 - Meta widget management
- [x] When a meta widget is deleted, all the child meta widgets should also be removed automatically
- [x] The modifyMetaWidgets actions should automatically add a property to each meta widget called `creatorId` which tells which List widget created the meta widgets
- [x] When calling `modifyMetaWidgets` with deleteIds, only the meta widgets belonging to the creator should get deleted.
- [x] When a page is changed or the List widget is removed, the meta widgets should also by removed.
- [x] Support the case where when the List widget is unmounted, the associated meta widgets are also removed.
- [x] Support n-nested Listwidgets

Tooluloope:
## Description

- [x] When a meta widget is deleted, all the child meta widgets should also be removed automatically
- [x] The modifyMetaWidgets actions should automatically add a property to each meta widget called `creatorId` which tells which List widget created the meta widgets
- [x] When calling `modifyMetaWidgets` with deleteIds, only the meta widgets belonging to the creator should get deleted.
- [x] When a page is changed or the List widget is removed, the meta widgets should also by removed.
- [x] When a List widget is unmounted

- Made changes to `widgetDeletionSaga` for bulk list widget and single deletion
- Removed all meta widgets on `INIT_CANVAS_LAYOUT`.
- Added `creatorId` to `MODIFY_META_WIDGETS` action
- Created a new Action to help with List widget Deletion.

Fixes #16583

## Type of change

## How Has This Been Tested?

- Test A
- Test B

## Checklist:


app/client/src/sagas/WidgetDeletionSagas.ts
```diff
@@ -217,13 +226,27 @@ function* deleteSaga(deleteAction: ReduxAction<WidgetDelete>) {
     if (widgetId && parentId) {
       const stateWidget: WidgetProps = yield select(getWidget, widgetId);
       const widget = { ...stateWidget };
+      let childMetaWidgetIds: string[] = [];
+      if (widget.type === "LIST_WIDGET_V2") {
```
ashit-rath, line 230: Let's not specifically add a case for List widget.
Add a flag to the List widget config defaults `hasMetaWidgets: true`
When any widget has metaWidgets then it should be delete.

app/client/src/sagas/WidgetDeletionSagas.ts
```diff
@@ -217,13 +226,27 @@ function* deleteSaga(deleteAction: ReduxAction<WidgetDelete>) {
...
+        const metaWidgets: MetaCanvasWidgetsReduxState = yield select(
+          getMetaCanvasWidgets,
+        );
+
+        const listChildMetaWidgetIds: string[] = yield call(
+          getMetaWidgetChildrenIds,
```
ashit-rath, line 236: Instead of lookup up for all the meta widgets recursively here, we should only get the meta widgets of this particular widget and the reducer should lookup the rest of its children and delete it.

app/client/src/sagas/WidgetOperationUtils.ts
```diff
@@ -301,6 +302,46 @@ export function getWidgetChildrenIds(
   return childrenIds;
 }
 
+export function getAllListWidgetIds(widgets: FlattenedWidgetProps[]): string[] {
```
ashit-rath, line 305: Let's always aim towards creating generic logics and names. As the meta widget is a platform level change and other future widgets like Grid/Form/Table would be using it.

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -240,6 +240,7 @@ class ListWidget extends BaseWidget<
       this.modifyMetaWidgets({
         addOrUpdate: metaWidgets,
         delete: removedMetaWidgetIds,
+        creatorId: this.props.widgetId,
```
ashit-rath, line 243: Let's move this `creatorId` from here to the implementation for `modifyMetaWidgets` in the `BaseWidget`. If we let widgets to add `creatorId`, then it can add any widgetId which is an abstraction leak.

app/client/src/reducers/entityReducers/metaCanvasWidgetsReducer.ts
```diff
@@ -27,19 +31,36 @@ const metaCanvasWidgetsReducer = createImmerReducer(initialState, {
...
     });
 
     return state;
   },
+  [ReduxActionTypes.BULK_DELETE_META_WIDGETS]: (
+    state: MetaCanvasWidgetsReduxState,
+    action: ReduxAction<BulkDeleteMetaWidgetPayload>,
+  ) => {
+    action.payload.metaWidgetIds.forEach((metaWidgetId) => {
+      delete state[metaWidgetId];
```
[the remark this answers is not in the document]
Tooluloope: This would be done in the widgetDeletionSaga, as we only get these Ids based on the creator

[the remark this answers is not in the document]
Tooluloope: oh, saw the other comment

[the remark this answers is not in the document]
ashit-rath: > This would be done in the widgetDeletionSaga, as we only get these Ids based on the creator

There is this rule of backend is to never trust what comes from the frontend, always double check the credibility of the request. So going by that logic, even if we were filtering in the saga, if I create another action which violates our delete logic; there  is nothing to stop it from executing that.

ashit-rath:
@Tooluloope 
1. We also need to support the case where when the List widget is unmounted, the associated meta widgets are also removed.

app/client/src/reducers/entityReducers/metaCanvasWidgetsReducer.ts
```diff
@@ -17,32 +18,45 @@ export type FlattenedWidgetProps<orType = never> =
...
+    } = action.payload;
 
+    if (addOrUpdate) {
+      Object.entries(addOrUpdate).forEach(([metaWidgetId, widgetProps]) => {
+        state[metaWidgetId] = widgetProps;
+        state[metaWidgetId].isMetaWidget = true;
+        state[metaWidgetId].creatorId = creatorId;
+      });
+    }
     deleteIds.forEach((deleteId) => {
```
ashit-rath, line 56: Add a new line before this

app/client/src/utils/metaReducerUtils.tsx
```diff
@@ -0,0 +1,29 @@
+import { MetaCanvasWidgetsReduxState } from "reducers/entityReducers/metaCanvasWidgetsReducer";
+
```
ashit-rath, line 2: Can we rename this file to metaWidgetReducerUtils?

app/client/src/sagas/metaReducerUtils.test.ts
```diff
@@ -0,0 +1,1070 @@
+import { MetaCanvasWidgetsReduxState } from "reducers/entityReducers/metaCanvasWidgetsReducer";
+import { getMetaWidgetChildrenIds } from "utils/metaReducerUtils";
+
```
ashit-rath, line 3: This file seems to be in the wrong folder.

app/client/src/reducers/entityReducers/metaCanvasWidgetsReducer.ts
```diff
@@ -17,32 +18,45 @@ export type FlattenedWidgetProps<orType = never> =
...
-};
-
 export type ModifyMetaWidgetPayload = {
   addOrUpdate: Record<string, FlattenedWidgetProps>;
   deleteIds: string[];
   propertyUpdates?: MetaWidgetPropertyUpdate[];
+  creatorId?: string;
+};
+export type DeleteMetaWidgetsPayload = {
+  creatorId: string[];
```
ashit-rath, line 28: Should it be `creatorIds`?

app/client/src/reducers/entityReducers/metaCanvasWidgetsReducer.ts
```diff
@@ -53,6 +67,24 @@ const metaCanvasWidgetsReducer = createImmerReducer(initialState, {
...
+
+    metaWidgetIds.forEach((metaWidgetId) => {
+      if (creatorId.includes(state[metaWidgetId].creatorId)) {
+        delete state[metaWidgetId];
+      }
+    });
+
+    return state;
+  },
+  [ReduxActionTypes.INIT_CANVAS_LAYOUT]: () => {
```
ashit-rath, line 85: Is this for resetting?

app/client/src/reducers/entityReducers/metaCanvasWidgetsReducer.ts
```diff
@@ -53,6 +67,24 @@ const metaCanvasWidgetsReducer = createImmerReducer(initialState, {
 
     return state;
   },
+  [ReduxActionTypes.DELETE_META_WIDGETS]: (
+    state: MetaCanvasWidgetsReduxState,
```
ashit-rath, line 71: Should we drop the word `Canvas` from here and call `MetaWidgetsReduxState`? In some places we are calling MetaWidgets but some MetaCanvasWidgets. I think we should choose one and stick with it everywhere. What do you think?

app/client/src/sagas/WidgetDeletionSagas.ts
```diff
@@ -311,7 +325,13 @@ function* deleteAllSelectedWidgetsSaga(
         mainCanvasMinHeight,
       );
     }
-
+    if (metaCreatorWidgetIds.length) {
+      yield put(
+        deleteMetaWidgets({
```
ashit-rath, line 330: How are we making sure that all the metaWidget's children (if any) are also being deleted?

app/client/src/utils/metaReducerUtils.tsx
```diff
@@ -0,0 +1,29 @@
...
+    const metaIds = getMetaWidgetByCreatorId(metaWidgets, parentId);
+    childrenIds.push(...metaIds);
+  });
+
+  return childrenIds;
+}
+
+function getMetaWidgetByCreatorId(
+  metaWidgets: MetaCanvasWidgetsReduxState,
+  parentId: string,
```
ashit-rath, line 19: rename this to `creatorId` as the function says `getMetaWidgetBy CreatorId`
Also this returns only ids not meta widgets so ideally it should be `getMetaWidgetIdsByCreatorId`

app/client/src/reducers/entityReducers/metaCanvasWidgetsReducer.ts
Tooluloope: Yes, for resetting.

app/client/src/reducers/entityReducers/metaCanvasWidgetsReducer.ts
Tooluloope: Yeah, I totally agree with that.

app/client/src/sagas/WidgetDeletionSagas.ts
Tooluloope: It's happening in the reducer

ashit-rath: @Tooluloope Can you point me to the place where this is happening? All I can see is we are looking at direct children only and not children's children and and so on recursively.

app/client/src/components/editorComponents/EditorContextProvider.tsx
```diff
@@ -61,6 +67,7 @@ export type EditorContextType = {
   // TODO (ashit) - Use generics instead of unknown
   setWidgetCache?: (widgetId: string, data: unknown) => void;
   getWidgetCache?: (widgetId: string) => unknown;
+  deleteMetaWidgets?: (creatorId: DeleteMetaWidgetsPayload) => void;
```
ashit-rath, line 70: `deletePayload: DeleteMetaWidgetsPayload`?

ashit-rath (commented):
@Tooluloope Have you considered nested list widgets as well?

app/client/src/sagas/WidgetDeletionSagas.ts
Tooluloope: Oh, apologies.
I changed the code after you said we should use `creatorId`.
Here was how it looked before.

Screenshot 2022-10-14 at 06 28 02

But with `creatorId`, we don't need to search recursively, since we have access to all `metaWidget` via the state, we'd just need to get them by their creator

ashit-rath: Ahh yes, got it
But I think there is an edge case here. 
Let's say we have nested list
List1 - outer list
List2 - inner list

All the widgets under List1 including List2 would have `creatorId `as List1
But the widgets under List2 would have the `creatorId `as List1. So when we delete List1, the List2's widgets might still stay around.

app/client/src/reducers/entityReducers/metaWidgetsReducer.ts
```diff
@@ -0,0 +1,102 @@
...
+  [ReduxActionTypes.DELETE_META_WIDGETS]: (
+    state: MetaWidgetsReduxState,
+    action: ReduxAction<DeleteMetaWidgetsPayload>,
+  ) => {
+    const { creatorIds } = action.payload;
+    const metaWidgetIds: string[] = getMetaWidgetChildrenIds(state, creatorIds);
+    const childMetaWidgetCreatorIds = getMetaWidgetCreatorIds(
+      state,
+      metaWidgetIds,
+    );
```
ashit-rath, line 83: We need to support n-nested lists and correct if I'm wrong, this logic seems to only deals with List > List

Tooluloope: Works n-levels deep

ashit-rath (approved):

PR #17017, opened 2022-09-23
feat: Implement data type handling in MySQL (#16621)

### LINKED ISSUE #10456: [Feature]: Add support for explicit data typecasting in MySQL prepared statement

### Summary

 If you try to add a number with leading zeros as a string into the database, it gets converted to an integer when the prepared statement is turned on. 

Here's an example where string "0001" gets converted to an integer

```
{
	"response":[
		0:{
			"affectedRows":1
		}
	]
	"request":{
		"requestParams":{
			"Query":{
				"value":"INSERT INTO MyTable (Name, Order, Code) VALUES ( $1, $2, $3 );"
				"substitutedParams":{
					"$1":{
						"value":"Local"
						"type":"STRING"
					}
					"$2":{
						"value":"1"
						"type":"INTEGER"
					}
					"$3":{
						"value":"0001"
						"type":"INTEGER"
					}
				}
			}
		}
	}
}
```

### LINKED ISSUE #15572: [Bug]: String "null" is treated as null data type instead of being treated as String data type

### Description

When I do execute a query to match a column with string "null" where the string "null" is coming from some binding with turning on the prepared statement then it doesn't return any data albeit there is data with the string "null".

When the prepared statement is turned on and binding has the value "null" as a string then it's treated as a null object instead of a string.

### Steps To Reproduce

1. Create Postgres/MySQL Datasource
2. Create an input widget and set the value to "null" //Input1
3. Write a query like `SELECT * FROM table where column = {{Input1.text}}`
4. Turn on the prepared statement
5. Insert some data in your table with the "null" string in your column
6. Execute the query and you will see no data is returned 

### Public Sample App

_No response_

### Version

Cloud / Self Hosted

### LINKED ISSUE #16600: [Bug] In MySQL prepared statements, it is impossible to use numeric values with leading zero as strings

### Description

Consider a MySQL query with the following SQL, where column `col` is of type `text`:

Prepared Statement: `ON`

```sql
update x set col = {{"0123"}}
```

This sets the value of the column to `0123` when **not** using stored procedures and `123` when using stored procedure. The reason seems to be that we are identifying this value as an _integer_, because `Integer.parseInt` succeeds, however, in doing so, we lose the leading zero. The leading zero here is important which is why the column was defined as `text` type instead of a numerical type.

### Steps To Reproduce

1. Go to the MySQL query pane and turn on the prepared statement
2. Write a MySQL update query like `update x set col = {{"0123"}}`
3. Now execute a select query on table x e.g. `select * from x`
4. You will notice that the col is having value of `123` instead of `0123`

### Public Sample App

_No response_

### Version

Cloud / Self Hosted

### LINKED ISSUE #16621: [Task]: Apply data type conversion logic to MySQL plugin

### SubTasks

_No response_

### LINKED ISSUE #16811: [Task]: Split for mysql datatypes cypress test cases

### SubTasks

Split the test cases as run time is > 6mins.

### LINKED ISSUE #17058: [Task]: Additional test cases of datatype handling in MySQL

### SubTasks

As implementation of data type handling in MySQL is about to complete. The specific cases which were handled in this [PR](https://github.com/appsmithorg/appsmith/pull/17017) should be updated in the test suite.

For example

- Ability to identify {{”null”}} as String data type.
- Ability to identify {{”098765”}} as String data type.

subrata71:
## Description
With this implementation we can now achieve the following under the prepared statement:
- Ability to distinguish between `null` object and `"null"` string
- Ability to distinguish values like `{{"098765"}}` and `{{098765}}`. The former is identified as a string and the latter is identified as an integer

Fixes #16621 
Fixes #16600 
Fixes #15572 
Fixes #10456
Fixes #16811
Fixes #17058

## Type of change

## How Has This Been Tested?

- JUnit
- Manual

## Checklist:


app/server/appsmith-plugins/mysqlPlugin/src/main/java/com/external/plugins/MySqlPlugin.java
```diff
@@ -1,11 +1,11 @@
 package com.external.plugins;
 
-import com.appsmith.external.constants.DataType;
+import com.appsmith.external.datatypes.*;
```
nidhi-nair, line 3: Could we please use complete references here?

app/server/appsmith-plugins/mysqlPlugin/src/main/java/com/external/plugins/MySqlPlugin.java
```diff
@@ -134,6 +137,31 @@ public class MySqlPlugin extends BasePlugin {
             "  and i.constraint_type in ('FOREIGN KEY', 'PRIMARY KEY')\n" +
             "order by i.table_name, i.constraint_name, k.position_in_unique_constraint;";
 
+    private static Map<ClientDataType, List<AppsmithType>> pluginSpecificTypes;
+
+    static {
```
nidhi-nair, line 142: Could we please separate out everything to do with datatypes into a different class? The `plugins/PluginClass/java` files should always only be entry points to all plugin related methods. Any logic can be separated out, for readability, extensibility, and maintainability.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/plugins/SmartSubstitutionInterface.java
```diff
@@ -39,7 +40,7 @@ default Object smartSubstitutionOfBindings(Object input,
                     String value = matchingParam.get().getValue();
 
                     input = substituteValueInInput(i + 1, key,
-                            value, input, insertedParams, args);
+                            value, input, insertedParams, append(args, matchingParam.get().getClientDataType()));
```
nidhi-nair, line 43: Hey sorry, I might be missing something here, but why did we decide to proceed with the varargs way of using a method call here?

nidhi-nair (changes requested):

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/plugins/SmartSubstitutionInterface.java
subrata71: The client data type is required in the substituteValueInInput method. One way we could pass this argument is by introducing a new variable just before the `varargs` but that would require changes along the way to all the plugins. I have found that in Postgres plugin `connectionPool` and `explicitCastDataTypes` objects are passed through varargs. @nidhi-nair

app/server/appsmith-plugins/mysqlPlugin/src/main/java/com/external/plugins/MySqlPlugin.java
subrata71: Yes, will do. Thanks

app/server/appsmith-plugins/mysqlPlugin/src/main/java/com/external/plugins/MySqlPlugin.java
subrata71: Do you suggest introducing a separate class where all the supported data types are accessible as a static member? @nidhi-nair

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/plugins/SmartSubstitutionInterface.java
nidhi-nair: Alright, I think we should keep these changes in mind for the future. Varargs are not great for visibility into the interface of a method because we end up not having any idea about what usage is like without dipping our fingers into code. The refactor here is a must have, put we can push it for late to a point when making the change becomes trivial.

app/server/appsmith-plugins/mysqlPlugin/src/main/java/com/external/plugins/MySqlPlugin.java
nidhi-nair: For the data type definitions as well as anything to do with substitution. Essentially anything outside of the pure functionality the is defined in the `PluginExecutor` interface should be split up.

subrata71: So, I am planning to introduce a new class like that and that will be accessed in the `PluginExecutor`

```
public class MySQLSpecificDataTypes {
    public static final Map<ClientDataType, List<AppsmithType>> pluginSpecificTypes;
    static {
      // populate the map
    }
}
```

**Usage**

`AppsmithType appsmithType = DataTypeServiceUtils.getAppsmithType(clientDataType, value, MySQLSpecificDataTypes.pluginSpecificTypes);`

Does it look okay? @nidhi-nair

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/plugins/SmartSubstitutionInterface.java
subrata71: Agreed with that but looking into the codebase I couldn't come up with a better choice. :( 
@nidhi-nair

app/server/appsmith-plugins/mysqlPlugin/src/main/java/com/external/plugins/MySqlPlugin.java
subrata71: There are a few methods implemented inside the PluginExecutor class which fit best inside a util class. I can keep this on my to-do list for future refactoring.
@nidhi-nair

app/server/appsmith-plugins/mysqlPlugin/src/main/java/com/external/plugins/datatypes/MySQLSpecificDataTypes.java
```diff
@@ -0,0 +1,35 @@
+package com.external.plugins.datatypes;
+
+import com.appsmith.external.datatypes.*;
```
nidhi-nair, line 3: Looks like the import here ended up being generic as well. :) There's a setting in IntelliJ to disable this. Auto-formatting will not introduce this after you do that.

subrata71: Yeah! 
Set 9999 to the following two fields (previously it was 5).

`Editor > Code Style > Java > Imports > "Class count to use import with '*" and "Names count to use static import with '*'"
`
@nidhi-nair

nidhi-nair (dismissed):

app/server/appsmith-plugins/mysqlPlugin/src/test/java/com/external/plugins/MySqlPluginTest.java
```diff
@@ -885,9 +885,12 @@ public void testStructure() {
         StepVerifier.create(structureMono)
                 .assertNext(structure -> {
                     assertNotNull(structure);
-                    assertEquals(2, structure.getTables().size());
+                    assertTrue(structure.getTables().size() > 0);
 
-                    final DatasourceStructure.Table possessionsTable = structure.getTables().get(0);
+                    final DatasourceStructure.Table possessionsTable =
+                            structure.getTables().
```
nidhi-nair, line 891: Nit: formatting seems off here.

app/server/appsmith-plugins/mysqlPlugin/src/test/java/com/external/plugins/MySqlPluginTest.java
```diff
@@ -885,9 +885,12 @@ public void testStructure() {
...
+                                    stream().
+                                    filter(table -> table.getName().equalsIgnoreCase("possessions")).findFirst().get();
                     assertEquals("possessions", possessionsTable.getName());
```
nidhi-nair, line 894: Seems like a redundant check

subrata71: In my local MySQL server there were a few other tables so the ordering of the tables that this test case expected was not happening. This made a tight coupling with the ordering of the tables. As we are only matching the structure of the two tables e.g. possession and users we are still achieving that but with no tight coupling with the order.
@nidhi-nair

app/server/appsmith-plugins/mysqlPlugin/src/test/java/com/external/plugins/MySqlPluginTest.java
```diff
@@ -885,9 +885,12 @@ public void testStructure() {
         StepVerifier.create(structureMono)
                 .assertNext(structure -> {
                     assertNotNull(structure);
-                    assertEquals(2, structure.getTables().size());
+                    assertTrue(structure.getTables().size() > 0);
```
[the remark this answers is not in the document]
subrata71: In the latter part of this test case, we are matching the structure of two tables e.g. possesions and users. I thought we shouldn't expect exactly two tables. If there exist more tables (depending on the environment) should we still stick to 2 tables' checking?
@nidhi-nair

app/server/appsmith-plugins/mysqlPlugin/src/test/java/com/external/plugins/MySqlPluginTest.java
nidhi-nair: Yes I meant that now that you are filtering on the basis of table name, the check with the name seems redundant. :) Instead the check should be in the previous step, that the `get` was successful.

subrata71: Ohh yes, that check now becomes redundant. Nice catch! I am removing this check and will update the PR accordingly. @nidhi-nair

app/server/appsmith-plugins/mysqlPlugin/src/test/java/com/external/plugins/MySqlPluginTest.java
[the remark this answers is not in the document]
subrata71: Got it. I am retaining the check then.

app/server/appsmith-plugins/mysqlPlugin/src/test/java/com/external/plugins/MySqlPluginTest.java
subrata71: Will format that.

nidhi-nair (approved):

app/client/cypress/integration/Smoke_TestSuite/ServerSideTests/Datatypes/MySQL_false_Spec.ts
```diff
@@ -0,0 +1,78 @@
...
+  });
+
+  it("5. Drop Table", () => {
+    query = inputData.query.dropTable;
+    dataSources.EnterQuery(query);
+    dataSources.RunQuery();
+  });
+
+  it("6. Verify Deletion of the datasource after all created queries are Deleted", () => {
+    dataSources.DeleteDatasouceFromWinthinDS(dsName, 409); //Since all queries exists
```
Aishwarya-U-R, line 68: This check can be removed from here since you are doing in spec1.
Directly delete the query, deploy, delete the ds - so few secs are saved.

ChandanBalajiBP: updated

nidhi-nair (approved):

Aishwarya-U-R (commented):
Cypress tests looks good!

app/client/cypress/integration/Smoke_TestSuite/ServerSideTests/Datatypes/MySQL_Spec.ts
```diff
@@ -1,5 +1,6 @@
 import { ObjectsRegistry } from "../../../../support/Objects/Registry";
 import inputData from "../../../../support/Objects/mySqlData";
+const isEqual = require("lodash/isEqual");
```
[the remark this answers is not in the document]
ChandanBalajiBP: Removed

Aishwarya-U-R (approved):
Cypress looks good!

PR #17054, opened 2022-09-26
fix: auto height text widget overflow none

ankurrsinghal:
Updated the `dynamicHeight` update property hook to set overflow none for text widget. Also referenced here in this PR, #16949.


app/client/src/utils/WidgetFeatures.ts
```diff
@@ -157,6 +158,15 @@ function updateMinMaxDynamicHeight(
     });
   }
 
+  console.log("updateMinMaxDynamicHeight", props);
```
riodeuno, line 161: Please remove the console log.

riodeuno (approved):

PR #17082, opened 2022-09-27
fix: Placeholder being shown when value is zero

### LINKED ISSUE #17063: [Bug]: Number and Range slider - Min Value field - Default value not set.

### Description

When I drag and drop the Range Slider, the default value is not set for Min Value field

### Steps To Reproduce

1. Drag and drop the Range and Number Slider widgets
2. In the Property pane, not that the Min Value field is blank (though evaluated value shows as 0)

Expected behavior - it should be populated with 1 (as per [this](https://www.notion.so/appsmith/Slider-Widgets-c8e9fcce6cdf4bcb98923669139b7087) Notion doc)

### Public Sample App

_No response_

### Version

Release

### LINKED ISSUE #17080: [Bug]:  Property Pane INPUT_TEXT if passed 0 as the default value for a widget property does not show the value

### Description

- For the `Number & Range` slider widgets, the `Min. Value` property uses the `INPUT_TEXT` control.
- When we pass the value `0` as the default value it does not show up in the `INPUT_TEXT` control instead we only see a placeholder.

### Steps To Reproduce

1. Drag and Drop the `Number / Range Slider` widget.
2. Check the `Min. Value` property in the property pane

### Public Sample App

_No response_

### Version

Release

aswathkk:
## Description

When the default value to INPUT_TEXT property control is 0, instead of showing the value 0, it is showing the placeholder text. This was because we were not properly checking for undefined value while rendering the component. This is fixed in this PR.

Fixes #17080
Fixes #17063 

## Type of change

## How Has This Been Tested?

- Manual

## Checklist:

### DIFF
--- a/app/client/src/components/propertyControls/InputTextControl.tsx
+++ b/app/client/src/components/propertyControls/InputTextControl.tsx
@@ -86,7 +86,7 @@ class InputTextControl extends BaseControl<InputControlProps> {
         onChange={this.onTextChange}
         placeholder={placeholderText}
         theme={this.props.theme}
-        value={propertyValue ? propertyValue : defaultValue}
+        value={propertyValue !== undefined ? propertyValue : defaultValue}
       />
     );
   }


yaldram (approved):

btsgh:
Tested this on the above DP. Default value is now 0 (instead of the placeholder) for both the Range and Number Sliders. 
Also verified - Notion doc has been updated.

PR #17085, opened 2022-09-27
POC: Dynamic height undo redo issue

riodeuno:
- Fixes issue where scroll height of container like widget's canvas was never resetting after we fixed the undo-redo issue
- Fixes issue where undo after move/add/delete/cut/paste/resize of a widget was not working.

### DIFF
--- a/app/client/src/utils/hooks/useDynamicAppLayout.tsx
+++ b/app/client/src/utils/hooks/useDynamicAppLayout.tsx
@@ -22,7 +22,6 @@ import { useWindowSizeHooks } from "./dragResizeHooks";
 import { getAppMode } from "selectors/entitiesSelector";
 import { updateCanvasLayoutAction } from "actions/editorActions";
 import { getIsCanvasInitialized } from "selectors/mainCanvasSelectors";
-import { calculateDynamicHeight } from "utils/DSLMigrations";
 
 const BORDERS_WIDTH = 2;
 const GUTTER_WIDTH = 72;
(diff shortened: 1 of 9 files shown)


app/client/src/widgets/ContainerWidget/component/index.tsx
```diff
@@ -86,7 +86,7 @@ function ContainerComponentWrapper(props: ContainerComponentProps) {
 }
 
 function ContainerComponent(props: ContainerComponentProps) {
-  useCanvasMinHeightUpdateHook(props.widgetId, props.minHeight);
+  // useCanvasMinHeightUpdateHook(props.widgetId, props.minHeight);
```
riodeuno, line 89: @marks0351 Please review this, and let me know of the consequences of removing this piece of code.

app/client/src/utils/hooks/useDynamicAppLayout.tsx
```diff
@@ -32,19 +31,19 @@ export const useDynamicAppLayout = () => {
...
   const isCanvasInitialized = useSelector(getIsCanvasInitialized);
   const appLayout = useSelector(getCurrentApplicationLayout);
 
-  /**
-   * calculates min height
-   */
-  const calculatedMinHeight = useMemo(() => {
-    return calculateDynamicHeight();
-  }, [mainCanvasProps]);
+  // /**
```
riodeuno, line 41: @marks0351 Please review this and please let me know the consequences of removing this piece of code.

marks0351: This was added to make sure the screen height of the device is taken into account to set the canvas's min height, when we resize the browser window the canvas min height also changes, removing this will not update it. its a slim usecase where user builds a small app(in terms of height) on a small screen and wishes not to have a scroll once he deploys the small app on a small device as well as a big device. In this dp if you resize your browser to make it small and build a small app, on deploying it wud show scroll in the resized small browser and wud not show scroll in normally resized browser.

app/client/src/widgets/ContainerWidget/component/index.tsx
marks0351: we do not have canvas min height available on the dsl(what we have is a dummy entry, it updates when someother update happens in the dsl) coz we compute it in memory and just supply to the view component which generates the children widgets but we wanted the minHeight values supplied to sagas as well as part of calculating resultant height of canvas when a widget is cut, deleted etc so we had them as meta props. we again didn't store them into dsl coz they were realtime and were bound to change.

marks0351: this doesn't seem to have any effect on the app for now although i am not sure if i have tested all the use cases.

marks0351: honestly i am finding it hard to remember all the usecases and the platform has changed a lot. so the new copy paste, reflow etc any of these features might be doing the job on the behalf of this hook.

app/client/src/utils/hooks/useDynamicAppLayout.tsx
riodeuno: Got it. Thanks. I'll look for an alternative approach, as this is causing an issue with the dynamic height. 
Wanted to know the reason for its existence, before messing with it. Chesterton's fence stuff.

rahulramesha: The latest changes in my PR could solve the issue in View mode from not having a scroll. But, These changes are required so the Canvas has the accurate screen height so that the canvas looks full instead of abruptly cut off to the last widget's bottom row.

riodeuno:
@marks0351 @rahulramesha Thanks for your inputs. I've noted the points down. We'll make sure to restore the fixes for issues before releasing this feature. I'm going to merge this in for now.

PR #17098, opened 2022-09-27
fix: Avoid closing a Toast message when auto-close is false

hetunandu:
## Description
It was noticed that the toast message component would close after some time even if `autoClose` option was false
This quick change will fix it

## Type of change

## Checklist:

### DIFF
--- a/app/client/src/components/ads/Toast.tsx
+++ b/app/client/src/components/ads/Toast.tsx
@@ -236,10 +236,12 @@ export const Toaster = {
         hideProgressBar: config.hideProgressBar,
       },
     );
-    // Update autoclose everytime to keep resetting the timer.
-    toast.update(toastId, {
-      autoClose: config.duration || 5000,
-    });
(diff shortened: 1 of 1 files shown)


tanvibhakta:
Hey @hetunandu  - could you please also (or maybe even instead?) make this change on https://github.com/appsmithorg/design-system/blob/feat/migrate-toast/packages/design-system/src/Toast/index.tsx ? We're in the process of migrating Toast as we speak and this component should land on release by EOW

tanvibhakta (changes requested):
Hey @hetunandu - could you please also (or maybe even instead?) make this change on https://github.com/appsmithorg/design-system/blob/feat/migrate-toast/packages/design-system/src/Toast/index.tsx ? We're in the process of migrating Toast as we speak and this component should land on release by EOW

hetunandu:
Updated on https://github.com/appsmithorg/design-system cc @tanvibhakta

PR #17137, opened 2022-09-28
fix: console log dx issue fixes

### LINKED ISSUE #17136: [Bug]: console log side effects in dx

### Description

The new console log changes cause some unwanted side effects which should be fixed
- Source of the logs is replaced with the override file. While it is necessary for the evaluations file but the exported functions and types can be moved to another file so that the main override is not imported to the main thread.
- For jest test, the function that replaces functions from nested objects with a name can sometimes cause the main object being logged to be mutated. Here, we need to make a clone of the object being logged and replace the custom logic with the json-fn lib.

### Steps To Reproduce

Reproduced when used with local env

### Public Sample App

_No response_

### Version

local release

ayushpahwa:
## Description

The new console log changes cause some unwanted side effects which should be fixed

- Source of the logs is replaced with the override file. While it is necessary for the evaluations file but the exported functions and types can be moved to another file so that the main override is not imported to the main thread. This PR introduces this change.
- For jest test, the function that replaces functions from nested objects with a name can sometimes cause the main object being logged to be mutated. Here, we are now making a clone of the object being logged and replacing the custom logic with the json-fn lib.

Fixes #17136 

## Type of change

## How Has This Been Tested?

- Test A
- Test B

## Checklist:


ayushpahwa:
Adding Test plan approved label since this is a dx change and doesn't introduce any new functionality. The existing tests should be sufficient to cover this.

app/client/src/workers/UserLog.ts
```diff
@@ -160,7 +85,9 @@ class UserLog {
     let returnData = [];
 
     try {
-      returnData = data.map((item: any) => {
+      // cloning the object to avoid mutation
+      const dataObject = klona(data);
```
arunvjn, line 89: TODO: Avoid deep cloning here if possible

Issue #16170, opened 2022-08-20
[Bug]: Tab title checks don't run when changing 


nzidol:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

When I type a long title into a tab, the tab shows a truncated name without dots:
![image](https://user-images.githubusercontent.com/13618973/185722866-aba7a7f0-2aaa-474c-af1b-841faf1d8d5c.png)
![image](https://user-images.githubusercontent.com/13618973/185722873-46e31a16-fc39-415e-8608-5fc08e87dd07.png)

After a screen refresh the tab shows dots (...) for missing characters
![image](https://user-images.githubusercontent.com/13618973/185722916-c63f7f32-8bac-4b9b-9870-717dc46fb80c.png)
(Nice to have: tab adjustable to text size auto or manual....)

When shortening the title, the tab shows the shortened name with dots, also note there is no error on the "Default Tab"
![image](https://user-images.githubusercontent.com/13618973/185723028-bc97508c-85eb-48a2-b93b-0e25e8529154.png)
![image](https://user-images.githubusercontent.com/13618973/185723210-581df3c1-2558-4a57-be77-916c3c08210f.png)

On page refresh the "Default Tab" shows an error and tab is as expected:
![image](https://user-images.githubusercontent.com/13618973/185723275-83172ce4-7d52-4027-8fcc-bbe4c73fb191.png)
![image](https://user-images.githubusercontent.com/13618973/185723284-f779b11c-c7ab-4652-9823-fa57d4312df0.png)

When correcting the tab name the error on "Default Tab" is not cleared.
![image](https://user-images.githubusercontent.com/13618973/185723627-259f3cd0-4548-44f0-a54d-2dc638122799.png)

### Steps To Reproduce

1. Create new page
2. Add Tab widget
3. Follow steps in description

### Version

Self Hosted 1.7.13

dilippitchika:
Thanks for bringing this to our attention @nzidol.

Both are valid issues and we will pick it up soon. We will soon fix the 1st one. I have linked them to the right issues to be fixed. Thanks.

PR #17138, opened 2022-09-28
feat: Add validation dependency

### LINKED ISSUE #15303: [Feature]: Revalidation when property is evaluated in the propertypane

### Summary

ReValidating properties in the property pane that have a dependency on each other.

1. Revalidation when the property is evaluated ( options ) 
2. Validation issue during createDataTree flow , because the defaultOptionValue gets evaluated & validated before options 
3. Creation of `validationDependency`
  - mainly will need to check different widget's validation functions.
  - understand the better approach to generate `validationDependency`:  is it parsing the `validationFunction` or giving static `validationDependency` array for each widget’s validation path.

e.g defaultValue in select has dependency on Options and ServerSideFiltering.

The Property should be re-validated.

https://theappsmith.slack.com/archives/CPG2ZTXEY/p1657372834459759

Removes unnecessary errors from the propertyPane/widget

### LINKED ISSUE #17159: [Bug]: validation and state updates for the inter-dependent controls in the property pane not working

### Description

The above bug was found during the development of Dynamic Feature called Auto Height with Limits. Here we have 2 values called min and max limit. Min cannot be greater than max, this is a validation we added. Now the min should check for its validations every time max is updated, which is currently not happening.

More findings:-
I added a property update for min control in the update hook for max control. Now after everytime max is updated and its last value was less than min, it sends an update for min control as well from its updateHook. But the error still remains, but as soon as I focus on the min control the red border goes away.
After more debugging I found that `CodeEditor.tsx` which is used in `InputTextControl` uses a state property called `AppState.evaluations.tree`, this get updates everytime we change a property of a widget. But I think there is no interdependcy logic built into this, so changing max will change the tree but will not re-render min control.

### Steps To Reproduce

1. Grab a text widget.
2. Enable the Auto Height with limits.
3. Set min to be greater than max.
4. Observe there is an error.
5. Now update the max to be greater than min.
6. Observe that the max control error is gone but min control still shows an error.

Please test it herem https://appsmith-mztipjmre-get-appsmith.vercel.app

### Public Sample App

https://appsmith-mztipjmre-get-appsmith.vercel.app

### Version

Cloud

Rishabh-Rathod:
## Problem

**Why do we need validation dependency?**

- When on change of particular property value if other property needs revalidation then current evaluation architecture do not support such validation run.

https://www.notion.so/appsmith/Validation-Dependency-d7623a5625bd4aa187a3ae2372d3ac07

**Issues to resolve**

Fixes #15303
Fixes #17159
Fixes https://github.com/appsmithorg/appsmith/issues/16170

## Solution

**New Validation Flow**

validate a property after its evaluation and check if there are other properties that need to revalidate on a change of this value. 

- if yes run validation for those properties.

How does validation dependency get created?

- every widget’s property will keep static `dependencies` array in `validation` property like shown below.

```jsx
{
	validation: {
    type: ValidationTypes.FUNCTION,
    params: {
      fn: defaultOptionValueValidation,
      expected: {
        type: 'value1 or { "label": "label1", "value": "value1" }',
        example: `value1 | { "label": "label1", "value": "value1" }`,
        autocompleteDataType: AutocompleteDataType.STRING,
      },
      dependentPaths: ["serverSideFiltering", "options"],
    },
    dependencies: ["serverSideFiltering", "options"],
  }
}
```

This `dependentPaths` array will be used to form validationDependencyMap.
 
#### Changes

- The code editor component wasn't re-rendering for a few cases. Due to this updates were not shown until the component state was changed.

- In CreateFirstTree, After evaluateTree we run `validateTree`. In `validateTree`, we validate all the properties in `validationPaths`. Here, if the errors were resolved on revalidation, it didn't reset the validation errors stored in data tree before. Now, this PR adds the reset logic. 
- Created `validationDependencyMap` to re-validate properties on change of the property their validation depends upon. 
- After each property evaluates in `updateDataTree`, we check if re-validation is needed, if yes then we re-validate dependent properties.  

## Type of change

## How Has This Been Tested?

- jest test

**Test Plan**
https://github.com/appsmithorg/TestSmith/issues/2078

## Checklist:


app/client/src/workers/DataTreeEvaluator/index.ts
```diff
@@ -1051,6 +1077,80 @@ export default class DataTreeEvaluator {
...
+        this.setParsedValue({
+          parsedValue,
+          currentTree,
+          entity: widget,
+          fullPropertyPath: fullPath,
+          propertyPath: getEntityNameAndPropertyPath(fullPath).propertyPath,
+          evalMetaUpdates,
+        });
+      });
+      // remove error if valid property
```
Rishabh-Rathod, line 1145: Handle both case of valid and invalid validation properly.

app/client/src/workers/DependencyMap/index.ts
```diff
@@ -106,11 +114,18 @@ export function createDependencyMap(
...
     dataTreeEvalRef.allKeys,
   );
-  return { dependencyMap, triggerFieldDependencyMap, invalidReferencesMap };
+
+  return {
+    dependencyMap,
+    triggerFieldDependencyMap,
+    invalidReferencesMap,
+    validationDependencyMap,
+  };
```
Rishabh-Rathod, line 128: Add updateValidationDependencyMap logic

Rishabh-Rathod (commented):
Dev review

app/client/src/workers/DataTreeEvaluator/validationUtils.ts
```diff
@@ -0,0 +1,167 @@
...
+            fullPath: false,
+          }),
+        );
+      }
+      // else {
+      //   resetValidationErrorsForEntityProperty(
+      //     tree,
+      //     `${entityKey}.${property}`,
+      //   );
+      // }
```
Rishabh-Rathod, line 163: #### Why we do not reset the validation error when revalidating in createFirstTree flow ? 

https://www.notion.so/appsmith/Validation-Dependency-d7623a5625bd4aa187a3ae2372d3ac07#afce596c20864bf0b65a0380e893c5e9

app/client/src/workers/evaluationUtils.ts
```diff
@@ -655,18 +570,33 @@ export const addErrorToEntityProperty = (
...
+export const resetValidationErrorsForEntityProperty = (
+  dataTree: DataTree,
+  fullPropertyPath: string,
+) => {
+  const { entityName, propertyPath } = getEntityNameAndPropertyPath(
+    fullPropertyPath,
+  );
+  if (propertyPath) {
+    const errorPath = `${entityName}.${EVAL_ERROR_PATH}['${propertyPath}']`;
+    const existingErrorsExceptValidation = (_.get(
```
Rishabh-Rathod, line 592: Just code refactor

Rishabh-Rathod (commented):
Initial self review

arunvjn (approved):
Looks good

PR #17168, opened 2022-09-29
feat: Static 404 page for chunk issue

### LINKED ISSUE #9740: [Bug] expected expression, got '<'
Sentry Issue: [APPSMITH-G5](https://sentry.io/organizations/appsmith/issues/2129275999/?referrer=github_integration)

```
UnhandledRejection: Non-Error promise rejection captured with keys: code, message, show
```

hetunandu:
## Description

Redirect to a static 404 page to avoid showing the user a blank screen

Screenshot 2022-09-29 at 11 26 27 AM

Fixes #9740

## Type of change

## Checklist:


app/client/public/404.html
```diff
@@ -0,0 +1,40 @@
...
+  </div>
+  <div class="main">
+    <img src="404-image.png" alt="404 image" class="not-found-img"/>
+    <p class="bold-text">We couldn’t load some JavaScript files for the operation.</p>
+    <p>A refresh usually works, but sometimes, it may take us a bit to get those files for you.</p>
+    <div class="button-container">
+      <span class="button primary" onclick="refresh()">Refresh</span>
+    </div>
+    <div class="try-again-container">
+      <p class="try-again-text">Trying again in 2 minutes</p>
```
mohanarpit, line 28: Any particular reason to keep such a big retry timeout?

kocharrahul7 (approved):
Output looks good, Let's hope this works 🙏

app/client/public/404.html
hetunandu: A user can refresh immediately, but if they don't it will automatically refresh in 2 minutes.
We chose 2 minutes as I thought if the server is down and causing this, it would get back up in around 2 minutes

mohanarpit: But will we hit this error screen if the server is down? Or will it happen if old chunks are being returned to the client?

hetunandu: For any reason, a chunk file is not downloaded, we will re attempt 3 times and then land on this screen.

Reasons why chunk is not download:
- The client has an old version of the app and hence an old link of a chunk, which is not present on the server to download
- The current server is unable to serve this file because it is down
- The server has not yet been updated but the client has so it does not have the file

mohanarpit (approved):

PR #17170, opened 2022-09-29
chore: Remove feature flags for GIT

AnaghHegde:
## Description

> GIT feature is available to all users. It does not make sense to keep this in feature flags anymore.


vishnu-gp (approved):

mohanarpit:
@AnaghHegde Has the client code base removed the feature flag as well? If not, then they will start seeing errors the minute we remove the feature flag from the server.

mohanarpit (approved):

AnaghHegde:
@IAmAnubhavSaini Could you please remove the feature flag from the FE code as well?

PR #17210, opened 2022-09-30
chore: minor refactor get datasource from action

nsarupr:


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCEImpl.java
```diff
@@ -182,6 +182,11 @@ public Mono<Datasource> populateHintMessages(Datasource datasource) {
                 });
     }
 
+    @Override
+    public Mono<Datasource> getValidDatasourceMono(Datasource datasource) {
```
trishaanand, line 186:
```suggestion
    public Mono<Datasource> getValidDatasourceFromActionMono(ActionDTO action, AclPermission aclPermission) {
```

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCEImpl.java
```diff
@@ -182,6 +182,11 @@ public Mono<Datasource> populateHintMessages(Datasource datasource) {
...
+        return Mono.just(datasource);
```
trishaanand, line 187:
```suggestion
        // Global datasource requires us to fetch the datasource from DB.
                    if (action.getDatasource() != null && action.getDatasource().getId() != null) {
                        return datasourceService.findById(action.getDatasource().getId(), aclPermission)
                                .switchIfEmpty(Mono.error(new AppsmithException(AppsmithError.NO_RESOURCE_FOUND,
                                        FieldName.DATASOURCE,
                                        action.getDatasource().getId())));
                    }
                    // This is a nested datasource. Return as is.
                    return Mono.justOrEmpty(action.getDatasource());
```

trishaanand (commented):
What do you think about something like this?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCEImpl.java
nsarupr: yeah, this makes sense. 
we have just moved this code from `NewActionServiceCEImpl`  ->  `DatasourceServiceCEImpl`
for better usage.

trishaanand (approved):

PR #17237, opened 2022-09-30
chore: added datasource-save event differentiator for analytics

### LINKED ISSUE #17099: [Task]: Differentiate between server side analytics for save datasource flow and other updates on datasource

### SubTasks

Today when a datsource is saved by a user by clicking on the Save button in the datasource editor, the server sends an update_DATASOURCE event to analytics. However, this same event is triggered even when the datasource gets updated by the backend as a result of refreshing tokens. We need to separate this event, by either pathing the code flow differently, or introducing a new event to only track these invisible updates so that we can remove them from our metrics.

sondermanish:
## Description

When we change the datasource, there is an explicit flow on the server side for saving those changes. The same flow is used by server to save the datasource and its configs when Authorisation tokens have to refreshed/ fetched. The both use case goes through the same flow and generates same analytics trace. However, this cannot be used to distinguish between both before mentioned use cases with same tag, therefore added an extra tag `isDatasourceUpdateServerInvoked` which is `True` when an internal flow triggered this event, `False`  otherwise.   

Fixes #17099

## Type of change

- Chore

## How Has This Been Tested?
- Manually

This doesn't require additional tests as the existing tests are adequate. Please add `testPlanApproved` tag

## Checklist:


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCEImpl.java
```diff
@@ -287,15 +305,13 @@ private Mono<Datasource> validateAndSaveDatasourceToRepository(Datasource dataso
...
-                .flatMap(unsavedDatasource -> {
-
-                    return repository.save(unsavedDatasource).map(savedDatasource -> {
-                        // datasource.pluginName is a transient field. It was set by validateDatasource method
-                        // object from db will have pluginName=null so set it manually from the unsaved datasource obj
-                        savedDatasource.setPluginName(unsavedDatasource.getPluginName());
-                        return savedDatasource;
-                    });
-                });
+                .flatMap(unsavedDatasource -> repository.save(unsavedDatasource).map(savedDatasource -> {
```
nidhi-nair, line 308: Nit: We usually prefer to use blocks instead of inline calls in cases like this for code readability. Could we please use that format? Also looks like there were some formatting issues with the previous code.

nidhi-nair (approved):

sondermanish:
Since the existing test cases are adequate, we don't need additional tests for this.
Checked segment for changes, found the intended flag working as desired.

image

Self approving the Test plan.

PR #17243, opened 2022-10-01
fix: Reflow while resizing does not reflow as expected

### LINKED ISSUE #17223: [Bug]: Reflow not working as expected

### Description

When app is generated via "Generate from a data table", Table is wrapped inside a container, when the container is resized, reflow is not happening.

https://user-images.githubusercontent.com/23132741/193230582-0a9e9e23-f7d6-4529-9eb2-706c0516c6e8.mov

check this loom too
https://www.loom.com/share/b3cfe8ad94fe4ab0a25fa9be48fd588dthe 

### Steps To Reproduce

1. Add a Table widget
2. Add a JSON form widget as shown in the above video
3. Try reflow and notice the behavior 

### Public Sample App

_No response_

### Version

1.8.4

rahulramesha:
## Description
   getIsReflowEffectedSelector was not returning for when widget was resized but only on while draging. Added the additional Condition to fix it.

Fixes #17223

## Type of change

## How Has This Been Tested?

Manual UI

## Checklist:

### DIFF
--- a/app/client/src/selectors/widgetReflowSelectors.tsx
+++ b/app/client/src/selectors/widgetReflowSelectors.tsx
@@ -1,6 +1,7 @@
 import { AppState } from "@appsmith/reducers";
 import { widgetReflow } from "reducers/uiReducers/reflowReducer";
 import { createSelector } from "reselect";
+import { getIsResizing } from "./widgetSelectors";
 
 export const getReflow = (state: AppState): widgetReflow =>
   state.ui.widgetReflow;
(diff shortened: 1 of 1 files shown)


app/client/src/selectors/widgetReflowSelectors.tsx
```diff
@@ -23,12 +24,14 @@ export const getIsReflowEffectedSelector = (
...
       return (
-        widgetId &&
-        dragDetails &&
-        !!dragDetails.draggedOn &&
-        dragDetails.draggedOn === widgetId &&
+        ((widgetId &&
+          dragDetails &&
+          !!dragDetails.draggedOn &&
+          dragDetails.draggedOn === widgetId) ||
+          isResizing) &&
```
marks0351, line 34: is it possible to write a test case for this scenario @rahulramesha ?

marks0351 (changes requested):

app/client/src/selectors/widgetReflowSelectors.tsx
rahulramesha: @marks0351, don't think we can, we will have to trigger reflow by dragging resize handle of the widget on cypress. That is similar to triggering reflow by dragging widget on canvas.

marks0351 (approved):

PR #17353, opened 2022-10-06
fix: replaced all the refs for simpler widgets

ankurrsinghal:
Since we are using ResizeObserver at the base widget level in the new DynamicHeightContainer, we can remove the dependency on ref passing down to widgets from base widget. This will make the API simpler for widgets as well.

### DIFF
--- a/app/client/src/widgets/BaseWidget.tsx
+++ b/app/client/src/widgets/BaseWidget.tsx
@@ -72,6 +72,7 @@ abstract class BaseWidget<
   K extends WidgetState
 > extends Component<T, K> {
   static contextType = EditorContext;
+  expectedHeight = 0;
 
   static getPropertyPaneConfig(): PropertyPaneConfig[] {
     return [];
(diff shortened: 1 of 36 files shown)


app/client/src/widgets/BaseWidget.tsx
```diff
@@ -500,10 +500,15 @@ abstract class BaseWidget<
   };
 
   addDynamicHeightContainer = (content: ReactNode) => {
+    const onHeightUpdate = (height: number) => {
+      this.updateDynamicHeight(height);
```
riodeuno, line 504: Let's use the same code here as we did in the componentDidUpdate, and remove the componentDidUpdate code.

ankurrsinghal: If we remove the componentDidUpdate, then other widgets like JSONForm, Container etc, won't work they still use the ref dependency.

ankurrsinghal:
Don't merge it yet, limits seem to be not working.

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -277,14 +276,7 @@ abstract class BaseWidget<
 
   /* eslint-disable @typescript-eslint/no-empty-function */
   /* eslint-disable @typescript-eslint/no-unused-vars */
-  componentDidUpdate(prevProps: T) {
-    requestAnimationFrame(() => {
-      const expectedHeight = this.contentRef.current?.scrollHeight;
-      if (expectedHeight !== undefined) {
-        this.updateDynamicHeight(expectedHeight);
```
ankurrsinghal, line 284: @riodeuno Since we have removed this piece of code, the AUTO_HEIGHT_WITH_LIMITS will not work as when min and max sends an update we are not running the `updateDynamicHeight`, and these functions are based on expectedHeight, which is also wrong as they are dependent on limits as well.

ankurrsinghal:
Since the whole QA is done and we are so close in fixing all the issues, and I cannot think right now, where these changes affect the current widgets, specially container ones. Let me know what you think? If you can look at the changes closely and you are also confident enough, we can go ahead.

app/client/src/widgets/BaseWidget.tsx
riodeuno: @ankurrsinghal Can we resolve this by calling `updateDynamicHeight` in the DynamicHeightOverlay component. So, when the positions of the min / max signifier changes (user stops changing it or the property pane sends a new update or when the component is mounted), we call `updateDynamicHeight` the value we send is going to come from the widget. We take the current widget `(bottomRow - topRow) * GridDefaults.DEFAULT_GRID_ROW_HEIGHT` and pass that as the expected height.

A few advantages with this approach:
1. Some code I had to add for the container widgets, can be removed.

Disadvantages of this approach:
1. We might need to do some special handling for some widgets. Not sure yet.

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -434,10 +432,22 @@ abstract class BaseWidget<
   ) {
     const onMaxHeightSet = (height: number) => {
       this.updateWidgetProperty("maxDynamicHeight", Math.floor(height / 10));
+      requestAnimationFrame(() => {
```
ankurrsinghal, line 435: @riodeuno is updateWidgetProperty async? Because running `updateDynamicHeight` after `updateWidgetProperty` it is not getting the latest min and max values correctly.

ankurrsinghal: Can I create a new function just for limits update and do some checks for the limits and run `updateWidgetDynamicHeight` from the context directly?

riodeuno: Can you wrap this requestAnimationFrame in `setTimeout(fn, 0)` and see if this helps?

Yes, as in your second point, we can check for changes in the min, max, and then call `updateDynamicHeight` in the `componentDidUpdate` of the `BaseWidget`. This is a viable solution.

riodeuno: Please wrap the requestAnimationFrame in setTimeout and not vice-versa.

riodeuno: @ankurrsinghal

riodeuno (approved):

PR #17395, opened 2022-10-08
chore: Refactor DB changelog to install plugins in workspace with single query

### LINKED ISSUE #16762: [Task]: Refactor: re-write method `installPluginToAllWorkspaces` so that it is run as a single command on MongoDB

### SubTasks

- re-write method `installPluginToAllWorkspaces` so that it is run as a single command on MongoDB
- ref: https://theappsmith.slack.com/archives/C02MUD8DNUR/p1663153042473789?thread_ts=1663092276.212609&cid=C02MUD8DNUR

bhuvanaindukuri:
## Description

> The method currently makes multiple DB calls
> 1) Fetch all workspace ids where the plugin id is not present
> 2) Fetch complete workspace details with these ids (1 query per workspace id)
> 3) Assign plugin and save each workspace

- Does not exactly fix the issue #16762 but does address it partially since the update happens in batches vs 1 by 1. Please check out the comment section to understand more. 

Fixes 

> The fix performs one single update query. The query adds the plugin to all workspaces where the plugin doesn't already exist.
> No data is fetched in java

## Type of change

## How Has This Been Tested?

> No tests found in the repo.
> Check if installation of plugin to all workspaces works when there are multiple workspaces and also when there are none.

- When there are no workspaces where the plugin is not installed
- There are multiple workspaces where the plugin is not installed

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
@@ -236,23 +236,16 @@ public class DatabaseChangelog {
     }
 
     public static void installPluginToAllWorkspaces(MongockTemplate mongockTemplate, String pluginId) {
-        Query queryToFetchAllWorkspaces = new Query();
+
+        Query queryToFetchWorkspacesWOPlugin = new Query();
         /* Filter in only those workspaces that don't have the plugin installed */
-        queryToFetchAllWorkspaces.addCriteria(Criteria.where("plugins.pluginId").ne(pluginId));
-        /* Fetch complete details of the workspaces */
-        List<Workspace> workspacesWithNoPlugin = mongockTemplate.find(queryToFetchAllWorkspaces, Workspace.class);
+        queryToFetchWorkspacesWOPlugin.addCriteria(Criteria.where("plugins.pluginId").ne(pluginId));
 
-        for (Workspace workspace : workspacesWithNoPlugin) {
+        /* Add plugin to the workspace */
+        Update update = new Update();
+        update.addToSet("plugins",new WorkspacePlugin(pluginId, WorkspacePluginStatus.FREE));
 
-            if (CollectionUtils.isEmpty(workspace.getPlugins())) {
-                workspace.setPlugins(new HashSet<>());
-            }
-
-            workspace.getPlugins()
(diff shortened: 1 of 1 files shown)


sumitsum:
@bhuvanaindukuri this does not look right. The goal of this issue is that we should not be fetching any data in the application memory. Instead we want to create a DB query (just one query) that simply runs on the DB and does the job - which means that we should not be requiring any result from the query.

bhuvanaindukuri:
> @bhuvanaindukuri this does not look right. The goal of this issue is that we should not be fetching any data in the application memory. Instead we want to create a DB query (just one query) that simply runs on the DB and does the job - which means that we should not be requiring any result from the query.

@sumitsum - Updated the code as per the requirement

sumitsum:
@bhuvanaindukuri can you please explain in the description what your change does ?

bhuvanaindukuri:
> @bhuvanaindukuri can you please explain in the description what your change does ? 

Updated the details of the fix in the PR description

sumitsum:
@bhuvanaindukuri the PR does not *exactly* do what was intended but does it partially. The bulk update operation seems to perform the updates in batches instead of all at once - which is still much better than doing it one by one as it was being done before. Can you please create a new GitHub issue mentioning the same and link this PR against it ? I am approving the PR. 
Ref: https://stackoverflow.com/questions/53296516/is-there-a-default-batch-size-that-is-used-by-mongodb-in-the-bulk-api
Ref: https://www.mongodb.com/community/forums/t/performance-bottleneck-in-bulk-insertion-java/8092/5

sumitsum:
@bhuvanaindukuri can you please merge with latest release and update the PR so that CI run may have a better chance at passing.

sumitsum (dismissed):

sumitsum:
@bhuvanaindukuri this may be helpful when updating the branch with latest release: https://stackoverflow.com/questions/7244321/how-do-i-update-or-sync-a-forked-repository-on-github 
Also, would it be possible for you to provide my account with push / edit access to your PR branch so that I may do such kind of updates(rebase) if required (in case you are not accessible) ?

sumitsum (approved):

sumitsum:
@bhuvanaindukuri seems like I have write access to the PR. I have rebased the PR. I will now run CI.

sumitsum:
Adding `test plan approved` label since this is a database migration related change. cc: @prapullac

PR #17412, opened 2022-10-09
fix: remove scroll from container widgets

ankurrsinghal:
Removed scroll contents option in property pane from container widgets when Dynamic Height is enabled.

### DIFF
--- a/app/client/src/widgets/ContainerWidget/widget/index.tsx
+++ b/app/client/src/widgets/ContainerWidget/widget/index.tsx
@@ -64,9 +64,6 @@ class ContainerWidget extends BaseWidget<
             controlType: "SWITCH",
             isBindProperty: false,
             isTriggerProperty: false,
-            hidden: (props: WidgetProps) =>
-              isDynamicHeightEnabledForWidget(props),
-            dependencies: ["dynamicHeight"],
           },
         ],
       },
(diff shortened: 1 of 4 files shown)


riodeuno:
@ankurrsinghal The scroll contents will be hidden only if the property is `AUTO_HEIGHT`, if it is `AUTO_HEIGHT_WITH_LIMITS`, it may still have scroll.

ankurrsinghal:
> @ankurrsinghal The scroll contents will be hidden only if the property is `AUTO_HEIGHT`, if it is `AUTO_HEIGHT_WITH_LIMITS`, it may still have scroll.

Yes, but it has to scroll if the expected height is greater than max height and not if it is less, so there is no point in having that options as it will always be redundant, same way we are handling the Text widget, from where the DynamicHeightContainer with ResizeObserver first created.

riodeuno:
@ankurrsinghal I see your point. We can disable this property for now, and revert if necessary in the future.
Also, I'm looking for a different mechanism to handle this. Converting this to draft for now. Also, our base branch now is `poc/dynamic-height-merge-release-1`

riodeuno:
@ankurrsinghal We have a new mechanism for this now. Please see the base branch's WidgetFeatures.ts file. I'm closing this PR.

PR #17415, opened 2022-10-09
fix: removed height auto default theme

ankurrsinghal:
Removed height auto css rule from the default theme.


ankurrsinghal:
@riodeuno I don't know why it was added by you, but the things are working fine I suppose.

PR #17448, opened 2022-10-10
fix: enabled dynamic height by default

ankurrsinghal:
Added default value for dynamic height = AUTO_HEIGHT in the widget register helpers.


riodeuno:
@ankurrsinghal I've added a different mechanism to enable this. Please see #17686 . Closing this PR.

Issue #17272, opened 2022-10-03
[Bug]: Forking selected pages from template from Add new page popover is crashing the app


Parthvi12:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

On forking selected pages from Add new page popover, crashes the app with page not found error

<img width="1440" alt="Screenshot 2022-10-03 at 4 03 00 PM" src="https://user-images.githubusercontent.com/80334441/193557237-f4e61338-55f7-41b2-82c3-2fa654ab0d65.png">

### Steps To Reproduce

1. Click on + on left pane of Pages
2. Select Add Page from Template option
3. Select 'Search' page of Customer Support Dashboard template 
Note : Happening on unchecking first page, and selecting other pages.
5. Observe the app crashes

### Version

Release

Parthvi12:
@akash-codemonk Can you try selecting only "Search" page and fork, it crashes for me

akash-codemonk:
Yeah tried that I see resource not found error as mention in https://github.com/appsmithorg/appsmith/issues/17270

Parthvi12:
@akash-codemonk Yes but there we are forking from start from template card and on refresh the app works fine, but here we are forking app from add new page popover and app goes in bad state, even on hard refresh it is not recovered. Hence marked as release blocker.

Richarex:
This is fixed now forking selected pages from template is not crashing now. https://www.loom.com/share/25ef55c41809485f93c4faacbd5abbda

Parthvi12:
Reopening as it's happening on release @akash-codemonk 
<img width="1440" alt="Screenshot 2022-10-11 at 2 43 38 PM" src="https://user-images.githubusercontent.com/80334441/195049636-7b76e99c-e1ee-492d-a1e2-b9e8ef7cfa3d.png">

akash-codemonk:
same as https://github.com/appsmithorg/appsmith/issues/17383? Did you unselect the first page?

AnaghHegde:
@Parthvi12 Raised a PR to fix issue - https://github.com/appsmithorg/appsmith/pull/17474

Richarex:
This is fixed now verified it on release.

PR #17474, opened 2022-10-11
fix: Post forking an application without the first page the application ends up having pages with page name as the page id

### LINKED ISSUE #17383: [Bug]: Post forking an application without the first page the application ends up having pages with page name as the page id

### Description

https://user-images.githubusercontent.com/67054171/194750905-94bd0186-39e0-4775-ae95-8af86b67e3c4.mov

page-id

If you fork a template by selecting pages to fork and if you don't select the first page. The application still ends having all the pages. Also the pages which were not selected seem to have the page name as the page id. We see the page not found error on the frontend when the invalid page(page which has the page name as the page id) ends up being the default page. Also it looks it is being set as the default page automatically post forking on the backend.

The response of GET `api/v1/applications/new` returns the unselected page(Dashboard) in the page list of the application. Also it is set as the default page and you can also see that the page id is the name of the page. Since it is set as the default page it tries to go to that page id on edit, but the page id is clearly invalid. The core issue seems to be the unselected page being added to the forked app.

### Steps To Reproduce

1. Create a new app and open the fork template modal. 
2. Fork the customer support dashboard by unselecting the first page.
3. You see the forked page. 
4. Go to home screen and edit the same app. It goes to a 404 page.

### Public Sample App

_No response_

### Version

Cloud

AnaghHegde:
## Description

When you fork a template by selecting pages to fork and if you don't select the first page. The application still ends up having all the pages. Also, the pages which were not selected seem to have the page name as the page id. We see the page not found error on the frontend when the invalid page(the page which has the page name as the page id) ends up being the default page. Also, it looks like it is being set as the default page automatically post forking on the backend.

The response of GET api/v1/applications/new returns the unselected page(Dashboard) in the page list of the application. Also, it is set as the default page and you can also see that the page id is the name of the page. Since it is set as the default page it tries to go to that page id on edit, but the page id is clearly invalid. The core issue is the reference inside the application.pages and application.publishedPages which was not updated based on the pageToImport parameter. Only the pageList in the JSON was updated after filtering. 

Fixes #17383 #17272 

## Type of change

## How Has This Been Tested?

- Locally
- JUnit Tests

## Checklist:


mohanarpit (approved):

wengad82:
why  I can't fork the template into a existing app?  i run appsmith v1.9.6

PR #17479, opened 2022-10-11
ci: Add a /command to run only perf tests

### LINKED ISSUE #14352: Running performance tests is linked to /ok-to-test, we need a new command to run only perf tests.

SatishGandham:
Adds a command to run only perf tests, and the command takes a ref parameter to choose specific code from performance infra.

Fixes #14352


.github/workflows/perf-test.yml
```diff
@@ -195,12 +193,19 @@ jobs:
         run: |
           ./cypress/setup-test.sh
 
+      - name: Get the performance-infra branch/ref
+        id: perf_infra_ref
+        run: |
+          ref=${{ github.event.client_payload.slash_command.args.named.ref }}
+          if [[ -z "$ref" ]]; then ref="main"; fi
+          echo ::set-output name=ref::$ref
+
```
SatishGandham, line 202: Does the above code return `main` as default if ref is not available?

PR #17482, opened 2022-10-11
feat: Resize Image option in Table Widget v2

### LINKED ISSUE #9815: [Feature]-[586]:There should be an option to resize image inside Table widget

### Summary

We should be able to set the height and width of the image inside table widget

The ability to provide height and width to images inside the table will give the images more visibility.

souma-ghosh:
## Description
Introduces Image Resize option under column settings for Image type columns.

Note:

- Default Row Height property (in Style tab for Table widget) does not clash with this feature
- When cell wrapping is turned on for a column then the final row height will be the bigger one among Image height and Wrapped cell height
- On reducing column width of an image column, the image covers all available width of the column while maintaining the aspect ratio

Fixes #9815 

## Type of change

## How Has This Been Tested?

- Cypress tests

## Checklist:

### DIFF
--- a/app/client/src/widgets/TableWidgetV2/component/Constants.ts
+++ b/app/client/src/widgets/TableWidgetV2/component/Constants.ts
@@ -36,9 +36,9 @@ export enum VerticalAlignmentTypes {
 }
 
 export enum ImageSizes {
-  DEFAULT = "32",
-  MEDIUM = "64",
-  LARGE = "128",
+  DEFAULT = "32px",
+  MEDIUM = "64px",
+  LARGE = "128px",
 }
 
 export const TABLE_SIZES: { [key: string]: TableSizes } = {
--- a/app/client/src/widgets/TableWidgetV2/component/Constants.ts
+++ b/app/client/src/widgets/TableWidgetV2/component/Constants.ts
@@ -35,6 +35,12 @@ export enum VerticalAlignmentTypes {
   CENTER = "CENTER",
 }
 
+export enum ImageSizes {
+  DEFAULT = "32",
+  MEDIUM = "64",
+  LARGE = "128",
+}
+
 export const TABLE_SIZES: { [key: string]: TableSizes } = {
   [CompactModeTypes.DEFAULT]: {
     COLUMN_HEADER_HEIGHT: 32,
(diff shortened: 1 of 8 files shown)


app/client/src/widgets/TableWidgetV2/widget/propertyConfig/PanelConfig/General.ts
```diff
@@ -231,5 +231,33 @@ export const GeneralStyle = {
...
+          value: "MEDIUM",
+        },
+        {
+          label: "Large",
+          value: "LARGE",
+        },
+      ],
+      isBindProperty: false,
+      isTriggerProperty: false,
+      validation: { type: ValidationTypes.TEXT },
```
sbalaji1192, line 257: We dont need validation

app/client/src/widgets/TableWidgetV2/component/Constants.ts
```diff
@@ -168,6 +175,7 @@ export interface BaseCellProperties {
   borderRadius: string;
   boxShadow: string;
   isCellVisible: boolean;
+  imageSize: ImageSize;
```
sbalaji1192, line 178: imageSize is not a base cell property. it is only applicable to image cell.

app/client/src/widgets/TableWidgetV2/component/Constants.ts
```diff
@@ -438,6 +447,7 @@ export type BaseCellComponentProps = {
   fontStyle?: string;
   textColor?: string;
   textSize?: string;
+  imageSize?: ImageSize;
```
sbalaji1192, line 450: its not a baseCellComponentprop. only specific to image

app/client/src/widgets/TableWidgetV2/component/Table.tsx
```diff
@@ -83,6 +84,7 @@ interface TableProps {
   boxShadow?: string;
   onBulkEditDiscard: () => void;
   onBulkEditSave: () => void;
+  imageSize?: ImageSize;
```
sbalaji1192, line 87: we don't need this type here.

sbalaji1192 (dismissed):

chandannkumar:
- Tested image size with all available size options
- Tested with cell wrapping
- Tested with row height
- Performed regression on table widget

sbalaji1192 (approved):

PR #17491, opened 2022-10-11
feat: Added analytics events for admin settings updates

vishnu-gp:
## Description

This PR adds analytics events for admin settings update events

## Type of change

## How Has This Been Tested?

- Manually

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/constants/FieldName.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/constants/FieldName.java
@@ -144,5 +144,5 @@ public class FieldName {
     public static final String ACTION_EXECUTION_TIME = "actionExecutionTime";
     public static final String WEBSITE = "website";
     public static final String TEMPLATE_APPLICATION_NAME = "templateAppName";
-    public static final String UPDATED_SETTINGS = "updatedSettings";
+    public static final String UPDATED_INSTANCE_SETTINGS = "updatedInstanceSettings";
 }
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/constants/FieldName.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/constants/FieldName.java
@@ -144,4 +144,5 @@ public class FieldName {
     public static final String ACTION_EXECUTION_TIME = "actionExecutionTime";
     public static final String WEBSITE = "website";
     public static final String TEMPLATE_APPLICATION_NAME = "templateAppName";
+    public static final String UPDATED_SETTINGS = "updatedSettings";
 }
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/EnvManagerCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/EnvManagerCEImpl.java
@@ -5,6 +5,7 @@ import com.appsmith.server.configurations.CommonConfig;
 import com.appsmith.server.configurations.EmailConfig;
 import com.appsmith.server.configurations.GoogleRecaptchaConfig;
 import com.appsmith.server.constants.EnvVariables;
+import com.appsmith.server.constants.FieldName;
 import com.appsmith.server.domains.User;
 import com.appsmith.server.dtos.EnvChangesResponseDTO;
 import com.appsmith.server.dtos.TestEmailConfigRequestDTO;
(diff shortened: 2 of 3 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/EnvManagerCEImpl.java
```diff
@@ -374,11 +375,16 @@ private Mono<Void> sendAnalyticsEvent(User user, Map<String, String> originalVar
         // Generate analytics event properties template(s) according to the env variable changes
         List<Map<String, Object>> analyticsEvents = getAnalyticsEvents(originalVariables, changes, new ArrayList<>());
 
+        // We cannot send sensitive information present as values in env to the analytics
+        // Values are filtered and only variable names are sent
+        Map<String, Object> analyticsProperties = Map.of(FieldName.UPDATED_SETTINGS, changes.keySet());
```
AnaghHegde, line 380: Can we add this to the `eventData` map and use it only in the AuditLogs? Similar to how we are sending the data in other places. It would be helpful to know what the updated values are.

AnaghHegde (commented):
Minor comments rest looks good to me

app/server/appsmith-server/src/main/java/com/appsmith/server/constants/FieldName.java
```diff
@@ -144,4 +144,5 @@ public class FieldName {
     public static final String ACTION_EXECUTION_TIME = "actionExecutionTime";
     public static final String WEBSITE = "website";
     public static final String TEMPLATE_APPLICATION_NAME = "templateAppName";
+    public static final String UPDATED_SETTINGS = "updatedSettings";
```
AnaghHegde, line 147: Request you to change this to `instanceSettingsUpdated` and the variable name to INSTANCE_SETTINGS_UPDATED

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/EnvManagerCEImpl.java
```diff
@@ -374,11 +375,16 @@ private Mono<Void> sendAnalyticsEvent(User user, Map<String, String> originalVar
...
+        Mono<User> sendAnalyticsMono = Mono.empty();
         // Currently supporting only one authentication method update in one env update call
```
AnaghHegde, line 382: Can we remove this comment as we are handling all the instance settings

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/EnvManagerCEImpl.java
vishnu-gp: @AnaghHegde I have thought about it initially but env values have this extreme sensitivity where we may not be in a position even to store it plain in the DB as Audit Log. Would like to hear insights from @trishaanand as well here.

app/server/appsmith-server/src/main/java/com/appsmith/server/constants/FieldName.java
vishnu-gp: @AnaghHegde So basically I would like to reuse this same field name inside Audit Log JSON as where we list all the env variables that were changed. Hence the more human-readable text. This is updated in the events Notion doc as well.

AnaghHegde: We have AuditLogConstants in ee repo. We can add a new variable there for this purpose. But let's keep it more generic in the FieldName class as this can be reused across different places.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/EnvManagerCEImpl.java
vishnu-gp: @AnaghHegde I think this comment would still be valid since the Authentication Analytics event is still a separate extra event. This is a futuristic thing where when we decide to update multiple Authentication methods from one single page in UI(single API call), we only track one.

app/server/appsmith-server/src/main/java/com/appsmith/server/constants/FieldName.java
vishnu-gp: Done

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/EnvManagerCEImpl.java
trishaanand: I think I agree here. We can right now begin with recording the field which was changed and by whom in admin settings and omit the exact values for these fields. If we get a request or feedback that the data is also interesting from audit logs pov, we can bring that in as well.

AnaghHegde (approved):

PR #17495, opened 2022-10-11
fix: Added null check for action dependency edge

### LINKED ISSUE #17494: [Bug]: Older applications on release throw internal server error as soon as it is clicked

### Description

When I click any older application on [release.app.appsmith.com](https://release.app.appsmith.com) it throws Internal Server Error provided that the application has some widgets in its `dsl`.

### Steps To Reproduce

1. Go to release.app.appsmith.com
2. Click on any application which has some widgets in its canvas
3. You will notice an Internal Server Error

### Public Sample App

_No response_

### Version

Release V1.8.4

nidhi-nair:
Since the older way of storing on load actions does not conform to the new type, this PR adds null checks to simply ignore all of the older values. These values are not used today and will be overridden as soon as an update layout is performed on the page, so they are inconsequential.

Fixes #17494

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/domains/ActionDependencyEdge.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/domains/ActionDependencyEdge.java
@@ -4,13 +4,11 @@ import com.appsmith.external.models.EntityDependencyNode;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.Setter;
-import lombok.ToString;
 import org.apache.commons.lang.builder.EqualsBuilder;
 import org.apache.commons.lang.builder.HashCodeBuilder;
 
 @Getter
 @Setter
-@ToString
 @AllArgsConstructor
 public class ActionDependencyEdge {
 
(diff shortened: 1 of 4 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/domains/ActionDependencyEdge.java
```diff
@@ -30,6 +32,10 @@ public boolean equals(Object obj) {
         if (obj instanceof ActionDependencyEdge) {
             final ActionDependencyEdge actionDependencyEdge = (ActionDependencyEdge) obj;
 
+            if (sourceNode == null || targetNode == null || actionDependencyEdge.sourceNode == null || actionDependencyEdge.targetNode == null) {
+                return false;
```
subrata71, line 36: This is for my better understanding

I see this check will essentially prevent the NPE from ever happening again but won't the older objects (i.e. previous structure where sourceNode/targetNode doesn't have any nested property) still persist in DB? 
@nidhi-nair

nidhi-nair: @subrata71 no, the string types were called `source` and `target`. I've changed the names of the properties so there will be no collision.

subrata71 (approved):

Issue #9403, opened 2021-11-26
[Bug]-[17]:Inconsistencies in Font style


aswathkk:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

In the table widget we have underline option in font style

![Screenshot 2021-11-26 at 3 41 32 PM](https://user-images.githubusercontent.com/10436935/143564376-16b60bb4-cc96-4f39-9a17-0fa6f1f25b9a.png)

In other widgets such as TextWidget and InputWIdget, we don't have underline option

![Screenshot 2021-11-26 at 3 42 31 PM](https://user-images.githubusercontent.com/10436935/143564495-746afd17-f34c-4a4e-bcf5-ef7a6a683063.png)

### Steps To Reproduce

1. In table widget, Select column type as Plain Text and look at font style
2. In other widgets, look at font style

## Fix to do

1. Add the underline option in the `Label styles `> `Emphasis` following widgets
  - Inputs : Input, Currency input, Phone input
  - Selects : Select, multi select, tree select, multi tree select
  - Datepicker
  - Rich text editor
  - Boolean widgets : Switch, checkbox
  - Group widgets : Switch group, Radio group, Checkbox group
  - JSON Form all column types
  - Range Sliders: Range Slider, Number Slider, Range Slider

2. Add the underline option in the Emphasis for the Text widget

### Environment

Production

### Version

Cloud

somangshu:
## Stats
| Stat | Values |
| ------------- | ------------- |
| Reach     | 17 |
| Effort (months)  | 0.5 |

sbalaji1192:
@dilippitchika What should be the expected action here? 
1. Should we remove the `underline` option in Table widget ?, or
2. Should we add the `underline` option in other widgets?

jdun28:
Would it be okay if I worked on this one? I can create a PR both ways so that you can decide which you want to merge in - either underline added to all or underline removed from the table font styles.

dilippitchika:
Missed this, we cannot remove this now from table since it's already being used and there is chance people can write JS to underline something. 

It's better if we only add the underline option in other places.

dilippitchika:
Updated the issue description to reflect the fix

aswathkk:
Hi @jdun28,
Thank you for showing interest in contributing. I've assigned this issue to you.

somangshu:
@ichik what call can we take here? 
If we are removing this prop all together, can we close the issue?

[...]

PR #17496, opened 2022-10-11
fix: Add Underline Font Style and Label Style

jdun28:
> Pull Request Template
>

## Description

Adds the icon and value for fontstyles and labelstyles in various widgets.

Fixes # (issue)

Fixes: #9403 

## Type of change

## How Has This Been Tested?

Updated `app/client/src/components/propertyControls/ButtonTabControl.test.tsx` to include passing in underline as a option that is passed in.

## Checklist:

### DIFF
--- a/app/client/src/widgets/MultiSelectWidget/widget/index.tsx
+++ b/app/client/src/widgets/MultiSelectWidget/widget/index.tsx
@@ -335,6 +335,10 @@ class MultiSelectWidget extends BaseWidget<
                 icon: "ITALICS_FONT",
                 value: "ITALIC",
               },
+              {
+                icon: "UNDERLINE",
+                value: "UNDERLINE",
+              },
             ],
             isJSConvertible: true,
             isBindProperty: true,
--- a/app/client/src/widgets/MultiSelectWidget/widget/index.tsx
+++ b/app/client/src/widgets/MultiSelectWidget/widget/index.tsx
@@ -335,10 +335,6 @@ class MultiSelectWidget extends BaseWidget<
                 icon: "ITALICS_FONT",
                 value: "ITALIC",
               },
-              {
-                icon: "UNDERLINE",
-                value: "UNDERLINE",
-              },
             ],
             isJSConvertible: true,
             isBindProperty: true,
--- a/app/client/src/widgets/TableWidget/widget/propertyConfig.ts
+++ b/app/client/src/widgets/TableWidget/widget/propertyConfig.ts
@@ -672,6 +672,10 @@ export default [
                       icon: "ITALICS_FONT",
                       value: "ITALIC",
                     },
+                    {
+                      icon: "UNDERLINE",
+                      value: "UNDERLINE",
+                    },
                   ],
                   isJSConvertible: true,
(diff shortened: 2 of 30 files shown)


jdun28:
@aswathkk I'm not sure what to do about the failures on this PR but I believe I have it all ready to go. I added to a few extra places that were not outlined in the GH issue but where label styles or font styles were defined. If you want me to remove those extra ones, just let me know - it's no problem.

app/client/src/widgets/TableWidget/widget/propertyConfig.ts
```diff
@@ -1705,6 +1705,10 @@ export default [
             icon: "ITALICS_FONT",
             value: "ITALIC",
           },
+          {
```
aswathkk, line 1708: This is a V1 of the Table widget that is deprecated now. We don't want to add this property here.

app/client/src/widgets/MultiSelectWidget/widget/index.tsx
```diff
@@ -335,6 +335,10 @@ class MultiSelectWidget extends BaseWidget<
                 icon: "ITALICS_FONT",
                 value: "ITALIC",
               },
+              {
```
aswathkk, line 338: MultiSelectWidget is deprecated we have a V2 for this widget. So, no need to add underline here!

aswathkk (changes requested):
Hi @jdun28 
Thanks for taking the time to contribute.

I found that you've added UNDERLINE to some of the deprecated widgets too. We don't want to do that since they are deprecated. I commented on a few places. There are some more widgets that are deprecated and I see that you've modified them too. Please take a look at those.

Also, when I select underline for some widgets like Input Widget, RichTextEditor, Checkbox etc, I found that the label is not getting underlined. I think you've just added property but didn't modify the underlying component to accept underline. Please take a look at this too

jdun28:
@aswathkk I removed the underlines I added on the widgets I saw that also had a V2. If there are others that are deprecated, I'd need those listed out as I wasn't seeing any other indication of deprecation in the code.

Anywhere in the other widgets where the component contained something similar to
```
font-size: ${labelTextSize ?? "inherit"};
  font-weight: ${labelStyle?.includes(FontStyleTypes.BOLD) ? "bold" : "normal"};
  font-style: ${
    labelStyle?.includes(FontStyleTypes.ITALIC) ? "italic" : "normal"
  };
```
I added the text-decoration piece for `FontStyleTypes.UNDERLINE` however not every component had a handler like this and based off of a review of the components it looked like some only needed the style added to the Emphasis. If there are still ones that aren't working properly but their Bold/Italic options work please point me at those specifically so I can dig in further.

Thanks!

aswathkk:
@jdun28 
The list of Deprecated widgets is available here: [src/utils/WidgetRegistry.tsx](https://github.com/appsmithorg/appsmith/blob/c3e807d246812e3f188e790722a78af4cd1efe60/app/client/src/utils/WidgetRegistry.tsx#L217-L225)

Whereever you added underline option, please verify yourself whether the option is working as expected or not. I found most of them aren't working.

PR #17515, opened 2022-10-12
feat: Ability to add new row in Table widget

### LINKED ISSUE #13672: [Feature]: Click to add a new row in the table

### Summary

Allow users to add a new row in the table with the data.
Link to PRD - https://www.notion.so/appsmith/Adding-a-new-row-in-the-table-0ec90544d26c4c25ab1f9b2504e919a1
Link to design - https://www.figma.com/file/aJ2Hwz9gH5AK6zaRYtZV8a/Table-widget-update?node-id=2119%3A132071

This will make tables support building full fledged apps

sbalaji1192:
## Description
Introducing the ability to add a new row in the table widget.
Link to PRD - https://www.notion.so/appsmith/Adding-a-new-row-in-the-table-0ec90544d26c4c25ab1f9b2504e919a1

Fixes #13672

## Type of change

## How Has This Been Tested?
Added a new spec file with the following cases

Table widget Add new row feature
 - Basic flow
	should test that allow Add new row property is present
	should test that Add new row link appears on the UI when the allow add new row property is enabled
	should test that onSave, onDiscard and default row are showing up only when the allow add new property is enabled
	should test that add new row link is disabled during the inline editing flow
	should test that clicking on add new row link adds an empty row at the top of the table
	should test that editing is disabled when in add new row in progress
	should test that new row is getting populated with the default row property value
	should test that inline editing, row selection, pagination, search, filters are actions cannot be performed while in add new row feature
	should test that only editable column cells are in editmode in the new row
	should test that newRow property holds the entered data
	should test that non data (iconBitton, button, menubutton) column cells are not showing up
 - Validation flow
	should test that validation is working for a new row cell
	should test that validation variable isNewRow is working
	should test that validation is working for more than one add new row cell at a time
	should test that validation error message only appears when a cell is in focus
	should test that save button is disabled when there is an error
 - Actions flow (save, discard)
	should test that discard button is undoing the add new feature
	should test that discard events is triggered when user clicks on the discard button
	should test that save event is triggered when user clicks on the save button

## Checklist:


app/client/src/utils/migrations/TableWidget.ts
```diff
@@ -567,3 +568,75 @@ export const migrateTableWidgetV2Validation = (currentDSL: DSLWidget) => {
...
+          return prev;
+        }, {}),
+    ${tableId}.editableCell.index)
+}}
+`;
+const newBindingSuffix = (tableId: string, columnName: string) => {
+  return `
+    ))
+    (
+      (${tableId}.isAddRowInProgress ? ${tableId}.newRow.${columnName} : ${tableId}.columnEditableCellValue.${columnName}) || "",
```
ashit-rath, line 599: This binding suffix seems to be same as https://github.com/appsmithorg/appsmith/pull/17515/files#diff-431c1c14916856d79951f5daf2b4e7d87cf1caf52e0f6956e13d5d08ced39c28R23

If they are the same can move such prefix and suffix to a constants file and refer it from there?

Reason - They seem to be very complex and if some changes takes place then it has to be changed at other places.

app/client/src/utils/migrations/TableWidget.ts
```diff
@@ -567,3 +568,75 @@ export const migrateTableWidgetV2Validation = (currentDSL: DSLWidget) => {
...
+            return prev;
+          }, {})),
+      ${tableId}.isAddRowInProgress ? -1 : ${tableId}.editableCell.index,
+      ${tableId}.isAddRowInProgress
+    )
+  }}
+  `;
+};
+
+export const migrateTableWidgetV2ValidationBinding = (
```
ashit-rath, line 614: Do we have tests for this?

app/client/src/utils/DSLMigrations.ts
```diff
@@ -1127,6 +1128,11 @@ export const transformDSL = (
 
   if (currentDSL.version === 64) {
     currentDSL = migrateRateWidgetDisabledState(currentDSL);
+    currentDSL.version = 65;
+  }
+
+  if (currentDSL.version === 65) {
```
ashit-rath, line 1134: Did adding a new migration here and not making an entry for the same didn't throw an error in `DSLMigration.test.ts`?

app/client/src/widgets/TableWidgetV2/component/cellComponents/InlineCellEditor.tsx
```diff
@@ -109,10 +128,16 @@ export function InlineCellEditor({
...
-  const onFocusChange = useCallback((focus: boolean) => !focus && onSave(), [
-    onSave,
-  ]);
+
+  const onFocusChange = useCallback(
+    (focus: boolean) => {
+      !focus && onSave();
+      setHasFocus(focus);
+    },
+    [onSave],
```
ashit-rath, line 139: Can we also add `setHasFocus`?

app/client/src/widgets/TableWidgetV2/component/cellComponents/InlineCellEditor.tsx
```diff
@@ -7,27 +7,38 @@ import { InputTypes } from "widgets/BaseInputWidget/constants";
 import { EditableCell } from "widgets/TableWidgetV2/constants";
 import { TABLE_SIZES, VerticalAlignment } from "../Constants";
 
+const FOCUS_CLASS = "has-focus";
+const PADDING_OFFSET = 8;
```
ashit-rath, line 11: Is this value common as I saw in some other component? If so then can we move to the constants file?

app/client/src/widgets/TableWidgetV2/component/cellComponents/SwitchCell.tsx
```diff
@@ -97,6 +100,7 @@ export const SwitchCell = (props: SwitchCellProps) => {
       cellBackground={cellBackground}
       compactMode={compactMode}
       horizontalAlignment={horizontalAlignment}
```
ashit-rath, line 102: Observation: 
It looks like we are doing a lot of prop drilling to get the props to the component. 
Is this avoidable probably by using context or some other mechanism?

app/client/src/widgets/TableWidgetV2/component/header/actions/PageNumberInput.tsx
```diff
@@ -0,0 +1,102 @@
...
+  borderRadius: string;
+  accentColor?: string;
+}) {
+  const [pageNumber, setPageNumber] = React.useState(props.pageNo || 0);
+  useEffect(() => {
+    setPageNumber(props.pageNo || 0);
+  }, [props.pageNo]);
+  const handleUpdatePageNo = useCallback(
+    (e) => {
+      const oldPageNo = Number(props.pageNo || 0);
```
ashit-rath, line 60: Why are we doing `Number(...)` if `pageNo: number` ?

app/client/src/widgets/TableWidgetV2/component/header/actions/PageNumberInput.tsx
```diff
@@ -0,0 +1,102 @@
...
+  updatePageNo: (pageNo: number, event?: EventType) => void;
+  disabled: boolean;
+  borderRadius: string;
+  accentColor?: string;
+}) {
+  const [pageNumber, setPageNumber] = React.useState(props.pageNo || 0);
+  useEffect(() => {
+    setPageNumber(props.pageNo || 0);
+  }, [props.pageNo]);
+  const handleUpdatePageNo = useCallback(
```
ashit-rath, line 58: Can we add some new lines here and in line 55?

app/client/src/widgets/TableWidgetV2/component/header/actions/PageNumberInput.tsx
```diff
@@ -0,0 +1,102 @@
...
+    (e) => {
+      const oldPageNo = Number(props.pageNo || 0);
+      let page = Number(e.target.value);
+      // check page is less then min page count
+      if (isNaN(page) || page < 1) {
```
ashit-rath, line 63: Can we put `1` into a constant?

app/client/src/widgets/TableWidgetV2/component/header/actions/PageNumberInput.tsx
```diff
@@ -0,0 +1,102 @@
...
+      }
+      // fire Event based on new page number
+      if (oldPageNo < page) {
+        props.updatePageNo(page, EventType.ON_NEXT_PAGE);
+      } else if (oldPageNo > page) {
+        props.updatePageNo(page, EventType.ON_PREV_PAGE);
+      }
+      setPageNumber(page);
+    },
+    [props.pageNo, props.pageCount],
```
ashit-rath, line 78: Can we add `setPageNumber`, `props.updatePageNo`?

app/client/src/widgets/TableWidgetV2/component/header/actions/PageNumberInput.tsx
```diff
@@ -0,0 +1,102 @@
...
+      max={props.pageCount || 1}
+      min={1}
+      onBlur={handleUpdatePageNo}
+      onKeyDown={(e: any) => {
+        if (e.keyCode === Keys.ENTER) {
+          handleUpdatePageNo(e);
+        }
+      }}
+      onValueChange={(value: number) => {
+        setPageNumber(value);
```
ashit-rath, line 97: Why are we not calling `handleUpdatePageNo` here?

app/client/src/widgets/TableWidgetV2/component/header/actions/PageNumberInput.tsx
```diff
@@ -0,0 +1,102 @@
...
+  pageNo: number;
+  pageCount: number;
+  updatePageNo: (pageNo: number, event?: EventType) => void;
+  disabled: boolean;
+  borderRadius: string;
+  accentColor?: string;
+}) {
+  const [pageNumber, setPageNumber] = React.useState(props.pageNo || 0);
+  useEffect(() => {
+    setPageNumber(props.pageNo || 0);
```
ashit-rath, line 56: Do we need to have this in `useEffect` if we are calling `setPageNumber` from `handleUpdatePageNo`?
Also why do we need to maintain `pageNumber` separately when we have `pageNo` and `updatePageNo` from props?

app/client/src/widgets/TableWidgetV2/widget/derived.js
```diff
@@ -691,49 +694,83 @@ export default {
...
     };
 
-    const editedColumn = Object.values(props.primaryColumns).find(
-      (column) => column.alias === props.editableCell.column,
-    );
-    const value = props.editableCell.value;
-
-    if (editedColumn && editedColumn.validation) {
-      const validation = editedColumn.validation;
+    let editableColumns = [];
```
ashit-rath, line 697: Do we have updated tests for the changes?

app/client/src/widgets/TableWidgetV2/widget/index.tsx
```diff
@@ -134,7 +143,7 @@ class TableWidgetV2 extends BaseWidget<TableWidgetProps, WidgetState> {
       updatedRowIndices: `{{(()=>{ ${derivedProperties.getUpdatedRowIndices}})()}}`,
       updatedRow: `{{this.triggeredRow}}`,
       pageOffset: `{{(()=>{${derivedProperties.getPageOffset}})()}}`,
-      isEditableCellValid: `{{(()=>{ ${derivedProperties.getEditableCellValidity}})()}}`,
+      isEditableCellsValid: `{{(()=>{ ${derivedProperties.getEditableCellValidity}})()}}`,
```
ashit-rath, line 146: Can we add test for this?

app/client/src/widgets/TableWidgetV2/widget/index.tsx
```diff
@@ -1899,6 +1959,115 @@ class TableWidgetV2 extends BaseWidget<TableWidgetProps, WidgetState> {
...
+    value: boolean,
+    alias: string,
+    originalIndex: number,
+    rowIndex: number,
+  ) => {
+    if (this.props.isAddRowInProgress) {
+      this.updateNewRowValues(alias, value, value);
+    } else {
+      this.updateTransientTableData({
+        __original_index__: originalIndex,
```
ashit-rath, line 1975: The value of `ORIGINAL_INDEX_KEY` and this one is different. Is this suppose to be like this?
[here](https://github.com/appsmithorg/appsmith/pull/17515/files#diff-464ac52b424732b03c62586f9b039c254d839d3a140b9120e2f842deb34ad916R1914) as well

app/client/src/components/propertyControls/TableInlineEditValidationControl.tsx
```diff
@@ -209,12 +191,12 @@ class TableInlineEditValidationControlProperty extends BaseControl<
   };
 
   static getControlType() {
-    return "TABLE_INLINE_EDIT_VALIDATION";
+    return "TABLE_INLINE_EDIT_VALIDATION_CONTROL";
```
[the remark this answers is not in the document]
sbalaji1192: TABLE_INLINE_EDIT_VALIDATION_CONTROL is used in all the validation properties (regex, required, min, and max). TABLE_INLINE_EDIT_VALID_PROPERTY_CONTROL is used only in the valid property of the validation.

app/client/src/utils/migrations/TableWidget.ts
sbalaji1192: We are intentionally not using the binding from the property control since we need to freeze the binding string for the migration. while the binding in the property control will evolve over time.

app/client/src/utils/migrations/TableWidget.ts
sbalaji1192: I had some thought and decided not to write tests for the migration function, this function is not gonna change any more. Whatever the test we write gonna end up running for eternity increasing the unit test timing without a purpose.what do you think?

app/client/src/widgets/TableWidgetV2/component/cellComponents/InlineCellEditor.tsx
sbalaji1192: We don't have to. setState function identity doesn't change on re-renders. refer - https://reactjs.org/docs/hooks-reference.html#usestate

app/client/src/widgets/TableWidgetV2/component/cellComponents/SwitchCell.tsx
sbalaji1192: I thought about it, but each cell requires a different set of props. So if we use context it will be difficult to figure out which prop is being used by which cell. At least now we have a proper segregation of props. We used to send an object to the cells but we started spreading the props to avoid re-renders.

app/client/src/widgets/TableWidgetV2/component/header/actions/PageNumberInput.tsx
sbalaji1192: I simply moved the pageNumberInput file. Let me try to address these comments.

app/client/src/widgets/TableWidgetV2/component/header/actions/PageNumberInput.tsx
sbalaji1192: I guess we're calling in the onKeydown. This is simply to update the state.

app/client/src/widgets/TableWidgetV2/component/header/actions/PageNumberInput.tsx
sbalaji1192: Not sure about the logic here. Will refactor this when we do the responsive header story.

app/client/src/widgets/TableWidgetV2/widget/index.tsx
sbalaji1192: They are two different things. But was meant to name them the same. Now changed it.

app/client/src/components/propertyControls/TableInlineEditValidationControl.tsx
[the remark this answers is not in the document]
ashit-rath: I wanted to know if there are any major differences between both of them; to make them one instead of 2.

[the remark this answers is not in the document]
sbalaji1192: Yeah. They are having different bindings. I tried creating one class as an extension of another. but Pretty much most of the functions need to be overridden because of the binding. So created them as sep so they can evolve on their own.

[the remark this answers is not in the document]
ashit-rath: If by binding we mean different suffix and prefix then can we pass these as parameters from the property config. If this is the only reason then having 2 may cause maintenance issue as a new dev might not look into both for a change.

app/client/src/utils/migrations/TableWidget.ts
ashit-rath: Interesting point. 
Thought: Not all applications migrate at the same time, some happen just after release and some at later point of time based on when the application is re-opened or self-hosted updates. If someone modifies this file accidentally then it would go unnoticed (very tiny chance of this happening), then it's gonna cause trouble.
Also having tests right now would validate the current migration as well.
So we might have to weigh the repercussions vs the running time of the test.

We could start a decommission process where we slowly remove tests for old migrations to speed up our test suite run time.

app/client/src/widgets/TableWidgetV2/component/header/actions/PageNumberInput.tsx
ashit-rath: Wouldn't this cause some perf issue as it would lead to double re-renders? Can we fix it in this PR?

sbalaji1192: I have not modified this code. It got carried from the v1 and now I have just moved the file. As there are already too many things to test in this PR. I'm a little worried about making changes unrelated to the issue.
Having said that, I don't think it would cause a re-render since react would bail out during the second setPageNumber

> If you update a State Hook to the same value as the current state, React will bail out without rendering the children or firing effects. - react doc

ashit-rath: With this logic, the props.pageNo and pageNumber are always kept in parity. Due to this 
Render 1 -  props.pageNo changes 
Render 2 - due to subsequent setState of `pageNumber`

bail out would only happen if same value but to reach to the value first it will do a setState.

sbalaji1192: we're calling `setPageNumber` while updating the page no. . so during the subsequent call in useEffect, pageNo and pageNumber should be in parity, right?

sbalaji1192: Let's keep this code as is for now.

app/client/src/utils/migrations/TableWidget.ts
sbalaji1192: I have added a test case to proof it from accidental changes.

app/client/src/utils/DSLMigrations.ts
sbalaji1192: I overlooked it. It did throw an error.

app/client/src/components/propertyControls/TableInlineEditValidationControl.tsx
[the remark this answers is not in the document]
sbalaji1192: I don't there will be changes that we need to make in the two controls. I see them evolving on their own. And also there are changes like binding, additional dynamic data, and prompts. I guess let's keep them as different controls.

ashit-rath (approved):
One minor callout

kamakshibhat-appsmith:
Issues observed:

Issue 1 : Add new row link is NOT disabled during the inline editing flow
       Loom --> https://www.loom.com/share/07f153726543464f859aea0d6d2d983c
Issue 2: after scrolling to the last data of the table ,can we scroll to the top to the selected row (on click of ADD NEW ROW )?
       Loom--> https://www.loom.com/share/9e902957d151471aa1aab7963b5a916b 
Issue 3: isNewRow variable doesn't work when we have empty string
       Loom-->https://www.loom.com/share/f759b5237dac413aa08162f360ec8a87
Issue 4: Issue with Pagination when we click on discard.
       Loom -->https://www.loom.com/share/a1360eba6a474b228902f4197a418bcb
@sbalaji1192 Lmk if more details are required

kamakshibhat-appsmith:
Issue 4: After clicking Save action on the add new row, it takes a while to execute the query , You can notice in that time, discard is also enabled.
   Loom--> https://www.loom.com/share/5d810e216ebe438da1cbcacd09a63c42
Issue 5: Fix alignment of the icons in the table header
Attachment--> Screen Shot 2022-10-28 at 11 07 10 PM

sbalaji1192:
@kamakshibhat-appsmith Have fixed all the issues except
#3 - Expected behavior. We treat the empty string as a valid value if the required field is not enabled.

kamakshibhat-appsmith:
Issue in migration
The validations are not
working on ADD NEW ROW flow due migration failure
      ( steps:)
      - add validatons in the older app
      - now test it on the new app (enable-->add new row-->check the validation)
      - you can see some bad code is appended to the bindings

kamakshibhat-appsmith:
tested and looks Good  . This fixes [#13672 ](https://github.com/appsmithorg/appsmith/issues/13672) @sbalaji1192

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/Widgets/TableV2/Inline_editing_spec.js
```diff
@@ -682,20 +682,20 @@ describe("Table widget inline editing functionality", () => {
     ).should("not.be.disabled");
   });
 
-  it("24. should check that doesn't grow taller when text wrapping is disabled", () => {
+  it.only("24. should check that doesn't grow taller when text wrapping is disabled", () => {
```
ashit-rath, line 685: Should the `only` be removed here and line 698?

ashit-rath (approved):

PR #17519, opened 2022-10-12
fix: Ends with operator not working as expected

### LINKED ISSUE #13334: [Bug]-[1320]:'Ends With' operator in Table filters not working in certain cases

### Description

It appears to not work when filtering the entire word in a column

Appears to work fine when filtering letters from the last word in the following situations. Note that I searched for 'son' and 'Benson' in the FullName column. Gives correct results

Screenshot 2022-04-26 at 7 57 09 PM

Screenshot 2022-04-26 at 7 56 58 PM

Search for 'son' in the LastName column yields correct results.
Screenshot 2022-04-26 at 8 07 00 PM

Now when I search for the whole word 'Benson' on the LastName column, it gives incorrect results. My guess is it is not returning correct results for anything more than 3 letters in a column with a single word (Eg: LastName column) and also if you search for the entire word in a column (Eg: Catherine Benson in FullName column)
Screenshot 2022-04-26 at 7 57 24 PM

### Steps To Reproduce

Detailed above with screenshots. Shared sample app below

### Public Sample App

https://app.appsmith.com/app/table-filters-ends-with-operator/page1-6268050000cba01cd0305d2d

### Version

Cloud/Self-Hosted - v1.6.20

souma-ghosh:
## Description

Ends with operator not working in certain scenarios.
Here the issue is that we are using `lastIndexOf` without considering the edge case where this method might return a `-1`. When `lastIndexOf` returns -1 the condition for endsWith operation returns true for certain string lengths where it should have returned false.

Handled this by updating the endsWith checking logic.

Fixes #13334 

## Type of change

## How Has This Been Tested?

- Cypress tests

## Checklist:

### DIFF
--- a/app/client/src/widgets/TableWidgetV2/widget/derived.js
+++ b/app/client/src/widgets/TableWidgetV2/widget/derived.js
@@ -444,9 +444,10 @@ export default {
         try {
           const _a = a.toString().toLowerCase();
           const _b = b.toString().toLowerCase();
-          return _a.lastIndexOf(_b) >= 0
-            ? _a.length === _a.lastIndexOf(_b) + _b.length
-            : false;
+          return (
+            _a.lastIndexOf(_b) >= 0 &&
+            _a.length === _a.lastIndexOf(_b) + _b.length
+          );
         } catch (e) {
           return false;
         }
--- a/app/client/src/widgets/TableWidgetV2/widget/derived.js
+++ b/app/client/src/widgets/TableWidgetV2/widget/derived.js
@@ -444,8 +444,9 @@ export default {
         try {
           const _a = a.toString().toLowerCase();
           const _b = b.toString().toLowerCase();
-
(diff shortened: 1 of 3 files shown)


app/client/src/widgets/TableWidgetV2/widget/derived.js
```diff
@@ -444,8 +444,9 @@ export default {
         try {
           const _a = a.toString().toLowerCase();
           const _b = b.toString().toLowerCase();
-
-          return _a.length === _a.lastIndexOf(_b) + _b.length;
+          return _a.lastIndexOf(_b) >= 0
```
sbalaji1192, line 447: return _a.lastIndexOf(_b) >= 0 &&  _a.length === _a.lastIndexOf(_b) + _b.length;

sbalaji1192 (approved):

laveena-en:
Tested and verified the `ends with` operator.
PR's behaviour is as expected.

PR #17522, opened 2022-10-12
feat: Dynamic height (final)

riodeuno:


app/client/src/widgets/MultiSelectWidget/widget/index.tsx
```diff
@@ -435,6 +436,7 @@ class MultiSelectWidget extends BaseWidget<
         dropdownStyle={{
           zIndex: Layers.dropdownModalWidget,
         }}
+        isDynamicHeightEnabled={isDynamicHeightEnabledForWidget(this.props)}
```
riodeuno, line 439: @ankurrsinghal We don't need to enable dynamic height for deprecated widgets. Let's check the widgets and remove this feature from the deprecated widgets.

app/client/src/widgets/SelectWidget/component/index.tsx
```diff
@@ -420,8 +422,10 @@ export interface SelectComponentProps extends ComponentProps {
   compactMode: boolean;
   selectedIndex?: number;
   options: DropdownOption[];
+  isDynamicHeightEnabled?: boolean;
   isLoading: boolean;
   isFilterable: boolean;
+  innerRef?: React.RefObject<HTMLDivElement>;
```
riodeuno, line 428: @ankurrsinghal `innerRef` may not be required here, please check

app/client/src/widgets/SingleSelectTreeWidget/component/index.tsx
```diff
@@ -131,7 +134,7 @@ function SingleSelectTreeComponent({
   const [filter, setFilter] = useState(filterText ?? "");
 
   const labelRef = useRef<HTMLDivElement>(null);
-  const _menu = useRef<HTMLElement | null>(null);
+  const _menu = useRef<HTMLDivElement>(null);
```
riodeuno, line 137: @ankurrsinghal This change may not be necessary, please check.

app/client/src/widgets/SwitchGroupWidget/component/index.tsx
```diff
@@ -52,14 +55,18 @@ export interface OptionProps {
   value: string;
 }
 
-function SwitchGroupComponent(props: SwitchGroupComponentProps) {
+const SwitchGroupComponent = React.forwardRef<
```
riodeuno, line 58: @ankurrsinghal Are refs necessary here?

app/client/src/widgets/TableWidget/component/Table.tsx
```diff
@@ -414,7 +415,7 @@ export function Table(props: TableProps) {
         </Scrollbars>
       </div>
       <ScrollIndicator
-        containerRef={tableBodyRef}
+        containerRef={tableBodyRef as React.MutableRefObject<HTMLDivElement>}
```
riodeuno, line 418: @riodeuno Revert this please.

app/client/src/widgets/TextWidget/component/FontLoader.tsx
```diff
@@ -6,10 +6,16 @@ interface Props {
   children?: React.ReactNode;
 }
 
-function FontLoader(props: Props) {
+const FontLoader = React.forwardRef<HTMLDivElement, Props>((props, ref) => {
```
riodeuno, line 9: @ankurrsinghal Do we need to add refs here anymore?

app/client/src/widgets/MultiSelectWidget/widget/index.tsx
ankurrsinghal: Noted
- [ ]

app/client/src/widgets/SelectWidget/component/index.tsx
ankurrsinghal: Noted.

app/client/src/widgets/SingleSelectTreeWidget/component/index.tsx
ankurrsinghal: Noted

app/client/src/widgets/SwitchGroupWidget/component/index.tsx
ankurrsinghal: No not needed will remove this change.

app/client/src/widgets/TableWidget/component/Table.tsx
ankurrsinghal: Noted

app/client/src/widgets/TextWidget/component/FontLoader.tsx
ankurrsinghal: No not needed.

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -56,7 +69,7 @@ abstract class BaseWidget<
   K extends WidgetState
 > extends Component<T, K> {
   static contextType = EditorContext;
-  context!: React.ContextType<typeof EditorContext>;
+  expectedHeight = 0;
```
riodeuno, line 72: @ankurrsinghal There are methods in this Class which use `expectedHeight` internally. Let's rename this variable, so that it doesn't conflict with those names. Perhaps `_expectedHeight`, if this isn't also being used in this file.

app/client/src/components/editorComponents/DynamicHeightOverlay/index.tsx
```diff
@@ -0,0 +1,322 @@
...
+    const showTableFilterPane = useShowTableFilterPane();
+    const { setIsAutoHeightWithLimitsChanging } = useAutoHeightUIState();
+    const isAutoHeightWithLimitsChanging = useSelector(
+      (state: AppState) => state.ui.autoHeightUI.isAutoHeightWithLimitsChanging,
+    );
+
+    const [isMinDotDragging, setIsMinDotDragging] = useState(false);
+    const [isMaxDotDragging, setIsMaxDotDragging] = useState(false);
+
+    const [maxY, setMaxY] = useState(maxDynamicHeight * 10);
```
aswathkk, line 81: @ankurrsinghal Is this value `10` same as GridDefaults.DEFAULT_GRID_ROW_HEIGHT?

If so, shouldn't we using that constant rather than literal value `10`?

riodeuno: 💯

PR #17535, opened 2022-10-13
fix: Incorrect page count when number of records in page is different from page size

### LINKED ISSUE #17187: [Bug]-[1072]:Page numbers are incorrect when records in a page are different than page size

### Description

When records in a page are different than page size, for example page size is 10 but records in a page are 50 then the no. of pages is still calculated using the page size but not the records on the page. This causes confusion to users.

### Steps To Reproduce

1. Add a table
2. Enable server side pagination
3. Add more records than the page size by setting the limit higher
4. Add total row count
5. Observe the number of pages are not accurate

### Public Sample App

_No response_

### Version

Production

Front logo [Front conversations](https://app.frontapp.com/open/top_52v4c)

souma-ghosh:
## Description

Total page count is incorrect when no. of records in a page is different from the visual page size of a table.

This is happening because we are calculating the total page count based on the page size of the table. The page size is calculated from the number of rows visible in the table without scrolling.

Fixed this by taking page size as the number of records in the current page.

Fixes #17187 

## Type of change

## How Has This Been Tested?

- Manually tested

## Checklist:

### DIFF
--- a/app/client/src/widgets/TableWidgetV2/component/Table.tsx
+++ b/app/client/src/widgets/TableWidgetV2/component/Table.tsx
@@ -136,7 +136,13 @@ export function Table(props: TableProps) {
   );
   const pageCount =
     props.serverSidePaginationEnabled && props.totalRecordsCount
-      ? Math.ceil(props.totalRecordsCount / props.pageSize)
+      ? /*
+      For serverSidePaginationEnabled we are taking props.data.length as the page size.
+      As props.pageSize is being set by the visible number of rows in the table (without scrolling),
+      it will not give the correct count of records in the current page when query limit
+      is set higher/lower than the visible number of rows in the table
+    */
+        Math.ceil(props.totalRecordsCount / props.data.length)
       : Math.ceil(props.data.length / props.pageSize);
   const currentPageIndex = props.pageNo < pageCount ? props.pageNo : 0;
   const {


Rhitottam:
/perf-test ref=rhito/feat/compare-with-lt-standard-deviation-for-regression

laveena-en:
@souma-ghosh The table seems to be crashing on reload. Can you please check? 
image

laveena-en:
Tested and verified.

app/client/src/widgets/TableWidgetV2/component/Table.tsx
```diff
@@ -135,8 +135,16 @@ export function Table(props: TableProps) {
     [columnString],
   );
   const pageCount =
-    props.serverSidePaginationEnabled && props.totalRecordsCount
-      ? Math.ceil(props.totalRecordsCount / props.pageSize)
+    props.serverSidePaginationEnabled &&
+    props.totalRecordsCount &&
+    props?.data?.length
```
sbalaji1192, line 140: No to do optional chaining. data is a required property in the props.

sbalaji1192 (approved):

app/client/src/widgets/TableWidgetV2/component/Table.tsx
```diff
@@ -138,8 +138,16 @@ export function Table(props: TableProps) {
...
   );
   const pageCount =
-    props.serverSidePaginationEnabled && props.totalRecordsCount
-      ? Math.ceil(props.totalRecordsCount / props.pageSize)
+    props.serverSidePaginationEnabled &&
+    props.totalRecordsCount &&
+    props.data.length
+      ? /*
+      For serverSidePaginationEnabled we are taking props.data.length as the page size.
+      As props.pageSize is being set by the visible number of rows in the table (without scrolling),
```
sbalaji1192, line 146: Can you move this comment to the start of the expression?

sbalaji1192 (approved):

PR #17536, opened 2022-10-13
fix: Fixed check for duplicate JSObject names in an application page

### LINKED ISSUE #13692: [Bug]-[400]:error js function already exists after deleting

### Description

to quote the user:
"It seems that the JS objects cache the functions you create, and there's no way to clear that cache.  I got an error that one of my controller functions already exists, even though I deleted it a while ago.  When I tried to delete and recreate the controller, it started saying that all the functions already existed, and the entire page started glitching.  I was able to get around this by cloning the page and deleting the old one, but this bug seems like something we'd want to fix."

### Steps To Reproduce

I was not able to reproduce this issue. Just posting here as reported by the user.

### Public Sample App

_No response_

### Version

Cloud

### LINKED ISSUE #14755: [Bug]: 400 error is returned on a newly defined JS function name

### Description

A newly defined function displays a message stating object and function name combination already exists when it does not.

### Steps To Reproduce

1. Go to app at [this](https://release.app.appsmith.com/app/scratch-pad/automation-test-cases-62987963c43c3d0bd6572185/edit/jsObjects/62b3f8abdd56222d3f8039b0?editName=false) location and observe the toast message and the browser logs display error for a newly defined function

image

### Public Sample App

_No response_

### Version

Cloud

## Test Plan
- [ ] [[Test Case]: verify that there is no 400 error appear on defining a new JS function name](https://github.com/appsmithorg/TestSmith/issues/1925)

Front logo [Front conversations](https://app.frontapp.com/open/top_5rtvw)

nidhi-nair:
## Description

Server side was allowing JS collection names to be duplicated in an app that took Appsmith apps to an inconsistent state. This PR fixes the logic for these checks.

Reproduction:
1. Create a new app, and open this app in another tab as well
2. Create a new JSObject in tab 1
3. Try to create a new JSObject in tab2. An error is seen on the page
4. Refresh the page
5. The app shows two JSObjects on the same name, but one of them won't have functions in the run dropdown.

Fixes #13692 
Fixes #14755

## Type of change

## How Has This Been Tested?
- JUnit test
- Manually tested

## Checklist:

### DIFF
--- a/app/server/appsmith-server/src/test/java/com/appsmith/server/services/LayoutActionServiceTest.java
+++ b/app/server/appsmith-server/src/test/java/com/appsmith/server/services/LayoutActionServiceTest.java
@@ -1,8 +1,10 @@
 package com.appsmith.server.services;
 
 import com.appsmith.external.models.ActionConfiguration;
+import com.appsmith.external.models.ActionDTO;
 import com.appsmith.external.models.Datasource;
 import com.appsmith.external.models.DatasourceConfiguration;
+import com.appsmith.external.models.PluginType;
 import com.appsmith.external.models.Property;
 import com.appsmith.server.constants.FieldName;
 import com.appsmith.server.domains.Application;
@@ -10,11 +12,9 @@ import com.appsmith.server.domains.GitApplicationMetadata;
 import com.appsmith.server.domains.Layout;
 import com.appsmith.server.domains.NewAction;
 import com.appsmith.server.domains.Plugin;
-import com.appsmith.external.models.PluginType;
 import com.appsmith.server.domains.User;
 import com.appsmith.server.domains.Workspace;
 import com.appsmith.server.dtos.ActionCollectionDTO;
-import com.appsmith.external.models.ActionDTO;
 import com.appsmith.server.dtos.DslActionDTO;
 import com.appsmith.server.dtos.LayoutActionUpdateDTO;
 import com.appsmith.server.dtos.LayoutDTO;
@@ -53,7 +53,6 @@ import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.test.StepVerifier;
 
-import java.io.File;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


app/server/appsmith-server/src/test/java/com/appsmith/server/services/LayoutActionServiceTest.java
```diff
@@ -544,6 +541,24 @@ public void tableWidgetKeyEscape() {
...
 
+    @Test
+    @WithUserDetails(value = "api_user")
+    public void testIsNameAllowed_withRepeatedActionCollectionName_throwsError() {
+        Mockito.doReturn(Flux.empty()).when(newActionService).getUnpublishedActions(Mockito.any());
+
+        ActionCollectionDTO mockActionCollectionDTO = new ActionCollectionDTO();
+        mockActionCollectionDTO.setName("testCollection");
+
+        Mockito.when(actionCollectionService.getActionCollectionsByViewMode(Mockito.any(), Mockito.anyBoolean()))
```
sondermanish, line 552: just for my info: why not use doReturn here as well, when both are spybeans?

sondermanish (approved):
when using doReturn can same stub be used more than once?

app/server/appsmith-server/src/test/java/com/appsmith/server/services/LayoutActionServiceTest.java
nidhi-nair: I could have, this is a mix of convention plus the fact that the `doReturn` way of mocking a method completely bypasses the stubbed method. I like the when/thenReturn way of doing things better though, because of the type checks and the fact that you have to be aware of all the calls being made inside the method in case of spies. It is easier to catch edge cases like that. Does that make sense?

Rhitottam:
/perf-test ref=rhito/feat/compare-with-lt-standard-deviation-for-regression

Testing new fixes to perf-infra tests

Rhitottam:
/perf-test ref=rhito/feat/compare-with-lt-standard-deviation-for-regression

Rhitottam:
/perf-test ref=rhito/feat/compare-with-lt-standard-deviation-for-regression

Issue #11639, opened 2022-03-04
[Feature]-[5000]:Clear store values


rohan-arthur:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

As an app developer, I can already:
1.  [store and retrieve](https://docs.appsmith.com/framework-reference/store-value) values across the application.
2. update the value of store variables

As an enhancement, I would like to remove("clear" or "reset") the keys from the store. 
There are two aspects to this:
**Clear entire store**
_I should be able to remove all store variables to initial by calling a function like **appsmith.store.clearAll()**_

**Clear single variable**
_I should be able to remove a specified variable by calling a function like **appsmith.store.clear(<<varName>>)**_

### Why should this be worked on?

This simple features rounds off the storing of global variables, by allowing a mechanism to clear or reset them. Users have asked for this feature several times on our support channel.

**TestPlan**

- [ ] https://github.com/appsmithorg/TestSmith/issues/2083
- [ ] https://github.com/appsmithorg/TestSmith/issues/2084

**Note**
The syntax for functions _clear_ and _clearAll_ is only for illustration, let us find the most effective way to present these functions.

rohan-arthur:
Related: #7827

Nikhil-Nandagopal:
@rohan-arthur should the function be at a global level to be consistent with storeValue? Otherwise, I feel this function signature is going to be confusing to users. Or we introduce storeValue underneath appsmith.store as well and deprecate it at the global level

rohan-arthur:
From [here](https://www.notion.so/appsmith/Simplify-the-experience-of-using-the-appsmith-store-630f9b63f0264b6b9e751dcd63a3f024#2cd9041f76bb4e4db555c071d48ccca0)

- [x ]  Appsmith users can use `storeValue(key: string, value: any)`  to create a store variable and assign any value to it using this function.
- [x ]  Appsmith users can use `storeValue("userID", undefined)`  to reset the value of a store variable.
- [ ] Appsmith users can use `clearValue('keyName')` to delete a store variable from the memory.
- [x ] Appsmith users can use `{{ appsmith.store.key }}` to reference/bind the store variable inside the application.

rohan-arthur:
## Stats
| Stat | Values |
| ------------- | ------------- |
| Reach     | 5000 |
| Effort (months)  | <<number>> |

satbir121:
[...]

PR #17568, opened 2022-10-14
feat: Clear store values

gitstart:
## Description

At the moment developer can add to store & update store

This Implementation is to allow developer
 - Clear all store value
 - Clear specifics value from store

Fixes https://github.com/appsmithorg/appsmith/issues/11639

## Type of change


rimildeyjsr (approved):

ramsaptami:
LGTM, can be merged to release

rimildeyjsr (approved):

rohan-arthur:
@rimildeyjsr can you share an example usage for both the new functions? Can they be used directly or with `appsmith.` accessor? 
If usage is direct, then `removeValue` may need a better naming, because every other function that is to do with the store has the word `store` used somewhere, and this would be the only exception. cc @bharath31

PR #17585, opened 2022-10-16
feat: ce tenant permissions

berzerkeer:
## Description

CE integrated with tenant permissions.

## Type of change


app/client/src/pages/Applications/index.tsx
```diff
@@ -397,6 +398,11 @@ function LeftPane() {
     userWorkspaces = loadingUserWorkspaces as any;
   }
 
+  const tenantPermissions = useSelector(getTenantPermissions);
+  const canCreateWorkspace = isPermitted(tenantPermissions, [
+    PERMISSION_TYPE.CREATE_WORKSPACES,
```
ankitakinger, line 403: why is the second parameter an array?

berzerkeer: Oh yeah, it doesn't need to be. Will change it.

berzerkeer: done

ankitakinger (approved):

PR #17621, opened 2022-10-17
chore: Add properties to analytics event

AnaghHegde:
## Description

> Fix NPE for datasource archive method analytics.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCEImpl.java
@@ -454,7 +454,14 @@ public class DatasourceServiceCEImpl extends BaseService<DatasourceRepository, D
                             .then(repository.archive(toDelete))
                             .thenReturn(toDelete);
                 })
-                .flatMap(analyticsService::sendDeleteEvent);
+                .flatMap(datasource -> {
+                    Map<String, String> eventData = Map.of(
+                            FieldName.WORKSPACE_ID, datasource.getId()
+                    );
+                    Map<String, Object> analyticsProperties = getAnalyticsProperties(datasource);
+                    analyticsProperties.put(FieldName.EVENT_DATA, eventData);
+                    return analyticsService.sendDeleteEvent(datasource, analyticsProperties);
+                });
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCEImpl.java
```diff
@@ -454,7 +454,14 @@ public Mono<Datasource> archiveById(String id) {
                             .then(repository.archive(toDelete))
                             .thenReturn(toDelete);
                 })
-                .flatMap(analyticsService::sendDeleteEvent);
+                .flatMap(datasource -> {
+                    Map<String, String> eventData = Map.of(
+                            FieldName.WORKSPACE_ID, datasource.getId()
```
nidhi-nair, line 459: Why are we using datasourceId for workspaceId? 🤔

AnaghHegde: My bad.

nidhi-nair (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCEImpl.java
```diff
@@ -454,7 +454,14 @@ public Mono<Datasource> archiveById(String id) {
                             .then(repository.archive(toDelete))
                             .thenReturn(toDelete);
                 })
-                .flatMap(analyticsService::sendDeleteEvent);
+                .flatMap(datasource -> {
+                    Map<String, String> eventData = Map.of(
+                            FieldName.WORKSPACE_ID, datasource.getWorkspaceId()
```
nidhi-nair, line 459: We actually have an `orgId` prop in the `getAnalyticsProperties` method. Do you want to replace that instead? Would that do the trick?

AnaghHegde: @nidhi-nair There are two parts, for AuditLogs we create new props named `eventData` in all the places. The other one is for Analytics. We won't need all the or in some cases, the data points we want won't be present. And another issue is the data points are already hardcoded and different service class has different name changing that would break our mixpanel dashboards.

trishaanand (approved):

PR #17658, opened 2022-10-18
feat: disable dropdown and left pane permissions check

berzerkeer:
## Description

Admin settings show only on isSuperUser. And api pane http method dropdown disable. Also added a check in categories to show/hide acl users tab.

## Type of change


ankitakinger (approved):
LGTM

PR #17683, opened 2022-10-18
fix: Replace reference error with sync field error

### LINKED ISSUE #12179: [Feature]-[762]:bring back error message `async functions in sync fields`

### Summary

When we use and async function (e.g. Query.run()) in a non-async field (e.g. `table.tableData`)

We used to throw an explicit error stating `Cannot use async functions in sync field`.
This was removed as having this check was hindering performance.

Right now, `TypeError: Api1.run is not a function` is shown which is misleading for the users.

We need to address this, so the users are not mislead.

### LINKED ISSUE #13273: [Bug]: API call inside a JS is displayed as an error when bound to a widget

### Description

Calling an API/datasource query in a JS object and binding this to a widget throws an error stating `Api1.run` is not a function. Error message needs to be improved since async function is not allowed in data fields.

### Steps To Reproduce

1. Go to app at [this](https://release.app.appsmith.com/app/git-connected-app-with-js-objects/page1-62668caa01f360366e25e4d2/edit) location
2. Observe the binding on the table widget display an error stating `mockAPI.run` is not a function

image

### Public Sample App

https://release.app.appsmith.com/app/git-connected-app-with-js-objects/page1-62668caa01f360366e25e4d2/edit

### Version

Cloud

Rishabh-Rathod:
## Description

We added logic to replace async function undefined error with "{{actionName}} cannot be used in this field".

Fixes #12179
Fixes #13273

Internal discussion for error message :- https://theappsmith.slack.com/archives/C02K0SZQ7V3/p1667457021297869?thread_ts=1667385039.225229&cid=C02K0SZQ7V3

## Type of change

## How Has This Been Tested?

Cypress test

*Test Plan*

- [ ] https://github.com/appsmithorg/TestSmith/issues/2086

## Checklist:


app/client/src/workers/evaluationUtils.ts
```diff
@@ -820,3 +821,35 @@ export const isATriggerPath = (
...
+      if (get(this.asyncFunctionsNameMap, referencedIdentifier)) {
+        return `${referencedIdentifier} action cannot be triggered from this field`;
+      }
+    }
+
+    const typeMatchResult = message.match(this.typeErrorRegex);
+    if (typeMatchResult) {
+      const referencedIdentifier = typeMatchResult[1];
+      if (get(this.asyncFunctionsNameMap, referencedIdentifier)) {
+        return `${referencedIdentifier} action cannot be triggered from this field`;
```
rimildeyjsr, line 847: this can be stored in a string and the string can be returned from line 839 and 847.

Also should we tell the user they are using an async function in a sync function? Might make the msg more clear?

app/client/src/workers/evaluate.ts
```diff
@@ -256,9 +257,9 @@ export default function evaluateSync(
 ): EvalResult {
   return (function() {
     resetWorkerGlobalScope();
-    const errors: EvaluationError[] = [];
-    let logs: LogObject[] = [];
-    let result;
+    const __errors: EvaluationError[] = [];
+    let __logs: LogObject[] = [];
+    let __result;
```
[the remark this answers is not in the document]
Rishabh-Rathod: `eval()` function when evaluating javascript code also passes the local scope variables which mean earlier when we where evaluating any property `errors`, `result`, `logs` were also accessible. 

Try using {{errors}} or {{logs}} in widget binding to experience it.

To avoid user's code evaluation from getting impacted 
1. we could either start using `Function()` for evaluation as it directly evaluates on the global scope.
2. name the variables as such that it would be rare usage in the user's code. 

I went with the second as 1st change is relatively more significant and also I still see a few use cases where the `eval()` method local scope could be a feature for us.

app/client/src/workers/evaluationUtils.ts
Rishabh-Rathod: Do you mean `async function in a sync field` ? we are trying to move away from sync field and async function terminology as they are confusing and not self-explanatory for users. 

Internal discussion - https://theappsmith.slack.com/archives/C02K0SZQ7V3/p1665998808737669?thread_ts=1665412058.786959&cid=C02K0SZQ7V3

Rishabh-Rathod: > this can be stored in a string and the string can be returned from line 839 and 847.

by `this` in the above sentence, do you mean the `action cannot be triggered from this field` string literal?

rimildeyjsr: > by this in the above sentence, do you mean the action cannot be triggered from this field string literal?

Yep, that string.

Also, forgot that we were moving away from sync/async - that case we can let it be

app/client/src/workers/evaluate.ts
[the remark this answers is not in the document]
rimildeyjsr: got it, thank you for the detailed explanation :)

NandanAnantharamu (approved):
LGTM

rimildeyjsr (dismissed):
LGTM

Rishabh-Rathod:
Will be changing the implementation for fixing the above issue to the older one after merging https://github.com/appsmithorg/appsmith/pull/18471.

PR #17686, opened 2022-10-18
fix: Auto Height Feature - Resolve issues and restructure code

riodeuno:
- [ ] Restructure code to make the platform agnostic of feature details when making changes to the property pane config and updating default values
- [ ] Add migrations but not enable it
- [ ] Update defaults for widgets with dynamic height
- [ ] Fix issue where Canvas widget's property pane configs were not defined
- [ ] Fix issue where invisible widgets were not reflowing correctly in preview mode
- [ ] Fix issue where max height signifier updates where not working correctly for container like widgets
- [ ] Fix issue where Tabs widget when dropped stayed in the loading state until refresh
- [ ] Fix issue where SelectComponent was showing errors in the console in dev mode.

### DIFF
--- a/app/client/src/sagas/dynamicHeightSagas.ts
+++ b/app/client/src/sagas/dynamicHeightSagas.ts
@@ -97,9 +97,8 @@ export function* getChildOfContainerLikeWidget(
     );
     if (tabsMeta) return tabsMeta.selectedTabWidgetId;
     return containerLikeWidget.tabsObj[0].widgetId;
-  } else {
-    if (Array.isArray(containerLikeWidget.children))
-      return containerLikeWidget.children[0];
+  } else if (Array.isArray(containerLikeWidget.children)) {
+    return containerLikeWidget.children[0];
   }
 }
 
--- a/app/client/src/sagas/dynamicHeightSagas.ts
+++ b/app/client/src/sagas/dynamicHeightSagas.ts
@@ -67,17 +67,17 @@ export function* getMinHeightBasedOnChildren(
     const { detachFromLayout } = stateWidgets[childWidgetId];
     // We ignore widgets like ModalWidget which don't occupy parent's space.
     // detachFromLayout helps us identify such widgets
-    if (!detachFromLayout) {
-      if (changesSoFar.hasOwnProperty(childWidgetId)) {
-        minHeightInRows = Math.max(
-          minHeightInRows,
-          changesSoFar[childWidgetId].bottomRow,
-        );
-        // If we need to get the existing bottomRow from the state
-      } else {
-        const childWidget: FlattenedWidgetProps = stateWidgets[childWidgetId];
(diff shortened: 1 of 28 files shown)


app/client/src/sagas/dynamicHeightSagas.ts
```diff
@@ -54,6 +54,55 @@ import { getWidgetMetaProps, getWidgets } from "./selectors";
...
+  let minHeightInRows = 0;
+  const stateWidgets: CanvasWidgetsReduxState = yield select(getWidgets);
+  const { children = [] } = stateWidgets[widgetId];
+  // For each child widget id.
+  for (const childWidgetId of children) {
+    // If we've changed the widget's bottomRow via computations
+    const { detachFromLayout } = stateWidgets[childWidgetId];
+    // We ignore widgets like ModalWidget which don't occupy parent's space.
+    // detachFromLayout helps us identify such widgets
+    if (!detachFromLayout) {
```
rahulramesha, line 70: Minor Nit: `if (!detachFromLayout) return minHeightInRows;` could be used instead to make the code look cleaner instead of nesting ifs

app/client/src/sagas/dynamicHeightSagas.ts
```diff
@@ -54,6 +54,55 @@ import { getWidgetMetaProps, getWidgets } from "./selectors";
...
+  // is not currently visible in the tabs widget.
+  if (containerLikeWidget.type === "TABS_WIDGET") {
+    const tabsMeta: { selectedTabWidgetId: string } | undefined = yield select(
+      getWidgetMetaProps,
+      containerLikeWidget.widgetId,
+    );
+    if (tabsMeta) return tabsMeta.selectedTabWidgetId;
+    return containerLikeWidget.tabsObj[0].widgetId;
+  } else {
+    if (Array.isArray(containerLikeWidget.children))
```
rahulramesha, line 101: Minor Nit: `else if `here can replace nesting

rahulramesha (approved):
Just a general comment going off on a tangent, `updateWidgetDynamicHeightSaga` has very long nested ifs and loops which make it a bit harder to understand which code is inside which loops or condition. Is it possible to break that function into smaller functions to reduce the size of the method and reduce nested conditions where ever possible.

app/client/src/sagas/dynamicHeightSagas.ts
rahulramesha: I was wrong again, basically if we use `if (detachFromLayout) continue;`  we don't have to nest the code inside `if (!detachFromLayout) {` condition. Lesser nesting of code leads for better readability.

riodeuno:
@rahulramesha I'm in the process of breaking it now. It is the core of our feature, so I'd like to make the code as clean as possible. All suggestions are welcome, please be as critical as you can be. I really appreciate it.

PR #17704, opened 2022-10-19
chore: Add recommended indexes

sharat87:
This PR adds indexes as recommended by MongoDB cloud.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
@@ -2744,6 +2744,9 @@ public class DatabaseChangelog2 {
 
     @ChangeSet(order = "037", id = "indices-recommended-by-mongodb-cloud", author = "")
     public void addTenantAdminPermissionsToInstanceAdmin(MongockTemplate mongockTemplate) {
+        dropIndexIfExists(mongockTemplate, NewPage.class, "deleted");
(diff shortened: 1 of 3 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
```diff
@@ -2741,4 +2741,14 @@ private <T extends BaseDomain> T fetchDomainObjectUsingId(String id, MongockTemp
         final T domainObject = mongockTemplate.findOne(query(where(fieldName(path)).is(id)), type);
         return domainObject;
     }
-}
\ No newline at end of file
+
+    @ChangeSet(order = "037", id = "indices-recommended-by-mongodb-cloud", author = "")
```
mohanarpit, line 2745: Can we also add the index for newPage collection?

sharat87: I was wondering if we should. If that index was the problem, I'm very surprised it didn't show up in the recommended indexes. 🤔 🤷‍♂️

Oh well, no harm I suppose. I'll go ahead and add it here.

Rhitottam:

mohanarpit (approved):

sharat87:
@Rhitottam, I'll wait for the result of the perf-test? Will it show up as a comment here?

Rhitottam:
> @Rhitottam, I'll wait for the result of the perf-test? Will it show up as a comment here?

Hi I ma just testing a change in our perf-infra code. No need to wait fro the result. 👍🏽

Issue #17739, opened 2022-10-20
[Task]: Use the new link component in the user auth flow


tanvibhakta:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SubTasks

- [x] Design a link component 
- [x] create the link component in the design system 
- [x] use that link component in the ~4 places we are directly styling `a` tags in the user auth flow.

PR #17740, opened 2022-10-20
chore: use link from ds

tanvibhakta:
## Description

Replaces custom link components with minimally styled ones from the design-system in the user auth flow. 
Fixes #https://github.com/appsmithorg/appsmith/issues/17739

In this PR, I have removed a global rule which says "no outlines". This allows our current design system to display the outlines, and makes it clear what other links we need to change/inputs we need to design etc, by the end of this effort. 

Design File: https://www.figma.com/file/4aZWeRRAyIBX660EbIQhXn/DS-Upgrade-Exploration?node-id=766%3A39347

Depends on https://github.com/appsmithorg/design-system/pull/242, https://github.com/appsmithorg/appsmith/pull/18231

## Type of change

## How Has This Been Tested?

Locally, and on the dp

### Flows to test
All the links on the UserAuth flow - 
- forgot password
- T&C + privacy policy
- Sign Up
- Back to Sign Up

## Checklist:


Rhitottam:

albinAppsmith (approved):

shadabbuchh:
Issues that were identified:

1. The font colour & size for `Forgot Password`, `Privacy Policy` & `Terms & Conditions` links is different from release/prod.
2. `Sign In`/`Sign Up` text is bold and looks interact-able even when email/password are not entered.

Screenshot 2022-10-25 at 12 10 00 PM

3. The underline is not continuous, it breaks for letters such as `g`, `p`, `y`.
Screenshot 2022-10-25 at 12 11 48 PM

4. Do we want the Pressed state to persist when user is redirected to a different tab? e.g. when we click on `Privacy Policy`.
https://www.loom.com/share/f23f4de4a2ff42628a89a0d19ee08fd2

tanvibhakta:
> The font color & size for Forgot Password, Privacy Policy & Terms & Conditions links is different from release/prod.

The goal is for these to follow the guidelines as set in the design files, not in release/prod. Could you please check if the guidelines are followed from the design? 
- color, hover, pressed behavior from design 
- font sizes from prod

> Sign In/Sign Up text is bold and looks interact-able even when email/password are not entered.

My code doesn't change this. In fact, while this isn't visible on release or prod environments, this is visible even on the release branch on local. I'm not sure what's happening here. 

>    The underline is not continuous, it breaks for letters such as g, p, y.

This is default browser behavior - typically we _want_ underline to break on descenders, but let me confirm with @vasanth-appsmith. 

> Do we want the Pressed state to persist 

This is also default browser behaviour to allow users to be able to resume context when they come back to a page, but I will confirm with design.

shadabbuchh:
> The goal is for these to follow the guidelines as set in the design files, not in release/prod. Could you please check if the guidelines are followed from the design?
> 
> * color, hover, pressed behavior from design
> * font sizes from prod

Font size is different from prod

> This is default browser behavior - typically we want underline to break on descenders, but let me confirm with @vasanth-appsmith.

As of now this behaviour is not happening in release/prod. And even the design file does not display this behaviour.

tanvibhakta:
3. We've decided to do away with breaks on descenders
4. We will be triggering the focus state only when a keyboard accesses it.

tanvibhakta:
@shadabbuchh - Issues 1, 3, and 4 should be fixed! I'm still not sure why 2 is happening, but since it's also happening on local-release and not anywhere else I think we can ignore it for now.

shadabbuchh:
Observations:

1. The gap between `Forgot Password`, `Back to Sign In`, `Privacy Policy` & `Terms & Conditions` links & the underline beneath each of these is more than what is expected.
2. Letter spacing for `Privacy Policy` & `Terms & Conditions` is greater in the DP than release/prod.
3. Space between `Sign In` & `Forgot Password` is greater than what is in release/prod.
4. The length of the whole `Sign In` & `Forgot Password` card is greater when compared to release/prod.
5. The placement of the `Sign Up` card as a whole is not congruent with release/prod.
6. Placeholder for Email/password is not placed in the exact center.

Please have a look at the video below for better reference:
https://www.loom.com/share/8faa6e8dc47b4ed6a817a43da421f789

tanvibhakta:
> Placeholder for Email/password is not placed in the exact center.

This is not within the scope of the current PR (I haven't changed anything about that here)

> The gap between Forgot Password, Back to Sign In, Privacy Policy & Terms & Conditions links & the underline beneath each of these is more than what is expected.

The space between the text and the underline is not determined by me, but determined by the line-height of the element which forms the bounding box of it. This will therefore change for different elements (this is one of the downsides of using something other that `text-decoration` to style our underlines). Since the difference is proportional to the font size, I propose that this is acceptable. 

> Letter spacing for Privacy Policy & Terms & Conditions is greater in the DP than release/prod.
> Space between Sign In & Forgot Password is greater than what is in release/prod.

This is an excellent catch, thank you! I have now implemented `TextType` within `Link` so this should be consistent everywhere.

Issue #17786, opened 2022-10-21
[Bug]: Update the left panel height on home page due to addition of admin settings


ankitakinger:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

When you have a long list of workspaces, the left panel on the home page is covered by the admin settings option at the bottom. Update the height of the left panel to handle this scenario.

### Steps To Reproduce

1. Go to the home page
2. Add a long list of workspaces
3. Check that on scrolling the left panel, you cannot view the last workspace in that list.

### Version

Self Hosted - latest

PR #17788, opened 2022-10-21
fix: Updating left panel height on home page due to addition of admin settings

ankitakinger:
## Description

> Updated left panel height on the home page due to the addition of admin settings to fix overlapping issue.

Fixes [#17786](https://github.com/appsmithorg/appsmith/issues/17786)

## Type of change

## How Has This Been Tested?

> Tested it locally and checked this fixes the issue.

## Checklist:

### DIFF
--- a/app/client/src/pages/Applications/index.tsx
+++ b/app/client/src/pages/Applications/index.tsx
@@ -326,7 +326,7 @@ const StyledAnchor = styled.a`
 
 const WorkpsacesNavigator = styled.div`
   overflow: auto;
-  height: calc(100vh - ${(props) => props.theme.homePage.header + 252}px);
+  height: calc(100vh - ${(props) => props.theme.homePage.header + 292}px);
   ${thinScrollbar};
   /* padding-bottom: 160px; */
 `;


Rhitottam:

IAmAnubhavSaini:
@ankitakinger I was reviewing it, just got side-tracked. Here's the review:

After this change, the left bar will have empty space for non-admin user.

Screenshot 2022-10-21 at 6 33 58 PM

Without this change, it didn't look **that** bad:

### before changes, for admin:
Screenshot 2022-10-21 at 6 37 38 PM


### after changes, for admin:
Screenshot 2022-10-21 at 6 38 08 PM

ankitakinger:
> @ankitakinger I was reviewing it, just got side-tracked. Here's the review:
> 
> After this change, the left bar will have empty space for non-admin user.
> 
> Screenshot 2022-10-21 at 6 33 58 PM
> 
> Without this change, it didn't look **that** bad:
> 
> ### before changes, for admin:
> Screenshot 2022-10-21 at 6 37 38 PM
> 
> ### after changes, for admin:
> Screenshot 2022-10-21 at 6 38 08 PM

I think extra space is better than the content getting covered. What do you suggest we should do here?

IAmAnubhavSaini:
The size of the container should be calculated by number of elements in it.

Could you share a screen shot of where the content is getting hidden? Because as I showed above, there is not much difference with earlier code.

IAmAnubhavSaini:
The extra admin option was hiding the last of the workspaces.

### with changes, the last most workspace is visible.

Screenshot 2022-10-21 at 6 54 04 PM

### without this PR, the last most workspace is hidden.
Screenshot 2022-10-21 at 6 53 48 PM

This was not clear in pr description or bug description.

PR #17808, opened 2022-10-21
feat: default admin path

berzerkeer:
## Description

deafult admin path based on permissions


app/client/src/ce/utils/adminSettingsHelpers.ts
```diff
@@ -29,3 +33,15 @@ export const saveAllowed = (settings: any) => {
...
+/* get default admin settings path */
+export const getDefaultAdminSettingsPath = (
+  isSuperUser = false,
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  tenantPermissions?: string[],
+): string => {
+  if (isSuperUser) {
+    return ADMIN_SETTINGS_CATEGORY_DEFAULT_PATH;
+  }
+  return ADMIN_SETTINGS_CATEGORY_AUDIT_LOGS_PATH;
```
ankitakinger, line 46: Audit logs is EE feature. I don't think this is needed here. Neither is the super user check needed coz by default this function should return that path.

berzerkeer: So since we show `Admin Settings` for everyone even those who are not a super user, they can still click on that right. So when they do, I have to check if they are a super user and redirect them to some place other than `Instance Settings` hence the `isSuperUser` check and why I chose `audit-logs` path was because in CE right now this redirect them to `Upgrade CTA for Audit Logs` . In EE though this function by default returns them `Groups`

ankitakinger: Okay but then we need to update this before we push the permission driven changes to release

ankitakinger (approved):

app/client/src/ce/utils/adminSettingsHelpers.ts
berzerkeer: Yes of course. This is only for this branch alone. In EE aswell we will update a few things when merging to release. If there are no other blockers , you can approve. thanks !

ankitakinger: approved already.

PR #17832, opened 2022-10-23
feat: Enable fetch API

### LINKED ISSUE #16326: [Feature]: Enable Fetch API

### Summary

The API editor is awesome and a secure way to make requests when combined with a datasource. While it has its place, it begins to feel too bulky while building large or complex apps. 

Enabling the Fetch API will give users more flexibility while building because it has a simple interface and is declared directly in code. It offers more control and can be used for edge case scenarios that wouldn't work in the API editor.

So many reasons, some of which are:
- Create API requests on the fly
- In memory file download
- Easier testing/debugging i.e before deciding to move API to datasource
- Better developer experience
- For custom JS libraries #5668 i.e thousands of packages [depend on fetch](https://www.npmjs.com/package/node-fetch?activeTab=dependents)

### LINKED ISSUE #18147: [Bug]: Unable to use some identifiers

### Description

In Appsmith, we are unable to use any identifiers that are already there inside [evaluate.ts](https://github.com/appsmithorg/appsmith/blob/14159c5593ea15945a0772ed6b5fa43c6d4f34af/app/client/src/workers/Evaluation/evaluate.ts) file since they are exposed to `evaluate` function. This leads to a lot of reserved keywords inside Appsmith.

https://user-images.githubusercontent.com/10436935/201354785-7a335004-1226-46f1-894b-9d437117274a.mov

### Steps To Reproduce

Add a binding with the global identifiers that we have on evaluate.ts file. example: `topLevelWorkerAPIs`,  `dataTree`,`beginsWithLineBreakRegex.toString()` etc.

### Public Sample App

_No response_

### Issue video log

_No response_

### Version

cloud

arunvjn:
## Description
Enables [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) browser API. 
Users can now write native fetch requests on Appsmith and have their apps directly communicate to their REST endpoints, bypassing Appsmith proxy server. 

Please note that _`fetch`_ API is overrode to never send cookies along with the request i.e fetch requests to the same domain will not include the cookies stored against that domain, for security reasons. The `credentials` attribute in fetch and [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) init object (optional second parameter that you can pass in), is always set to `omit`.

#### Cypress added for the following scenarios.
- Tests that fetch option `credentials: 'include'` is overrode before making the request.
- Tests to see if platform functions work with inside the `then` blocks of fetch

Fixes #18147 
Fixes #16326 

## Type of change

## How Has This Been Tested?
- Manual
- Cypress

### Test Plan

- [ ] https://github.com/appsmithorg/TestSmith/issues/2089

### Issues raised during DP testing

- [x] https://github.com/appsmithorg/appsmith/pull/17832#issuecomment-1312805369
- [x] https://github.com/appsmithorg/appsmith/pull/17832#issuecomment-1313281975

## Checklist:


app/client/package.json
```diff
@@ -166,6 +167,7 @@
     "url-search-params-polyfill": "^8.0.0",
     "webfontloader": "^1.6.28",
     "webpack-retry-chunk-load-plugin": "^3.1.1",
+    "whatwg-fetch": "^3.6.2",
```
Rishabh-Rathod, line 170: @arunvjn Could you please help in understanding where is `whatwg-fetch` library is being used?

arunvjn: It is used inside `fetch-intercept`

app/client/src/workers/evaluate.ts
```diff
@@ -119,6 +120,7 @@ export function setupEvaluationEnvironment() {
   });
   userLogs.overrideConsoleAPI();
   overrideTimeout();
+  interceptAndOverrideHttpRequest();
```
Rishabh-Rathod, line 123: [Out of scope] Just thinking out load, If we create a polyfill for `setInterval` and `clearInterval` using setTimeout like https://codesandbox.io/s/vbfgz?file=/src/index.js:367-377 will that help in resolving callback unable to access outer variables ( closure values ) issue?

app/client/cypress/integration/Smoke_TestSuite/ServerSideTests/JsFunctionExecution/fetch_xhr_spec.ts
```diff
@@ -0,0 +1,40 @@
+import { ObjectsRegistry } from "../../../../support/Objects/Registry";
+const jsEditor = ObjectsRegistry.JSEditor;
+const agHelper = ObjectsRegistry.AggregateHelper;
+
+describe("Tests fetch and xhr calls", () => {
```
sharat87, line 5: The title mentions XHR, but I don't see tests on XHR. There's two tests, but both are calling `fetch` only. Can we please add a test or two for XHR too?

app/client/package.json
sharat87: Isn't that needed only in environmens where `fetch` isn't available? Like old NodeJS versions? The `attach` call checks if `fetch` is already available in `env`, and if not, only then requires this module. From https://github.com/mlegenhausen/fetch-intercept/blob/develop/src/attach.js#L43.

In our case, `env` will be `self`, from https://github.com/mlegenhausen/fetch-intercept/blob/develop/src/browser.js#L4.

So, we should be okay not needing this dependency, no?

app/client/src/workers/HTTPRequestOverride.ts
```diff
@@ -0,0 +1,20 @@
...
+    //@ts-expect-error allow override
+    const obj = new XHRPrototype.constructor(args);
+    obj.withCredentials = false;
+    return obj;
+  };
+  self.XMLHttpRequest.prototype = XHRPrototype;
+
+  return fetchIntercept.register({
+    request: function(...args) {
+      const request = new Request(args[0], { ...args[1], credentials: "omit" });
```
sharat87, line 16: Curious, we are assuming `args` will be two arguments at least. What happens when there's a single request object passed here? How does it still work? 🤔

app/client/src/workers/HTTPRequestOverride.ts
```diff
@@ -0,0 +1,20 @@
...
+export default function interceptAndOverrideHttpRequest() {
+  const XHRPrototype = XMLHttpRequest.prototype;
+  //@ts-expect-error allow override
+  self.XMLHttpRequest = function(args) {
+    //@ts-expect-error allow override
+    const obj = new XHRPrototype.constructor(args);
+    obj.withCredentials = false;
+    return obj;
+  };
+  self.XMLHttpRequest.prototype = XHRPrototype;
```
sharat87, line 12: Can you elaborate a little on this please? I see two problems.

One, we are overriding the `withCredentials` to `false` in the constructor, but it can be changed before calling `open`, right? Like this:

```javascript
const xhr = new XMLHttpRequest();
xhr.open("GET", '/api/v1/applications/new', true);
xhr.withCredentials = true;
xhr.send();
```

This will still be possible, right?

Two, the original `XMLHttpRequest` class is still accessible here, with `XMLHttpRequest.prototype.constructor`. Check this out:
Screenshot 2022-10-28 at 12 16 11 PM

sharat87 (changes requested):

app/client/src/workers/HTTPRequestOverride.ts
arunvjn: fetch and Request constructor accepts similar arguments i.e both could take a url string or Request object as the first param and an options object as a second param. 
The override here always reconstructs a new Request object whose credentials attribute is set to 'omit'. 
The second param in both `Request` and `fetch` is always the options object. We override the 'credentials' attribute in this options object.

app/client/src/workers/HTTPRequestOverride.ts
arunvjn: Yes, nice catch :). We need to make this property immutable.

app/client/src/workers/HTTPRequestOverride.ts
sharat87: Yeah, that's where my question comes.

> both could take a url string or Request object as the first param and an options object as a second param

This means that `args` can be `[url: string, request: Record<string, unknown>]` or `[request: Record<string, unknown>]`, right?

But in line 16, we are using `args` as if we already know that it's the former, and not the latter. I'm wondering if that assumption is true here. 🙂

app/client/src/workers/HTTPRequestOverride.ts
sharat87: Instead, I'm wondering if we should just override the `open` method in `XMLHttpRequest`. Like this:

```javascript
// UNTESTED CODE
{
  const originalOpen = XMLHttpRequest.open;
  XMLHttpRequest.open = (...args) => {
    this.withCredentials = false;
    return originalOpen.apply(this, args);
  };
}
```

May be simpler? 🤔

app/client/src/workers/HTTPRequestOverride.ts
arunvjn: fetch's args could be `[ url: string, options: Record<string, unknown> ] `or `[ request: Request, options? : Record<string, unknown> ]` 

[Request](https://developer.mozilla.org/en-US/docs/Web/API/Request/Request) constructor args `[ url: string, options?: Record<string, unknown> ] ` or `[ request: Request, options? : Record<string, unknown> ]`. 

Second argument if provided will always be the options object that contains the 'credentials' attribute. 

What I am doing here is always creating a new Request using arg[0] (url or Request) and passing default options from arg[1] after overriding the credentials attribute.

app/client/src/workers/HTTPRequestOverride.ts
arunvjn: ^^ This should work.  It could solve the 2nd issue you mentioned as well.

sharat87: I'm hoping `send` is the _only_ way to actually execute a request. See if there's other ways to execute it, we'll have to patch those up as well. 🙂

app/client/src/workers/evaluate.ts
arunvjn: This might actually work. We could definitely try this out.

app/client/src/workers/HTTPRequestOverride.ts
arunvjn: Just found out that we cannot disable same origin cookies in XHR even with `withCredentials: false` . Reverting the changes around XHR.

app/client/package.json
arunvjn: Ah okay, I guess I looked at the source code of an older version of `fetch-intercept`.

app/client/cypress/integration/Smoke_TestSuite/ServerSideTests/JsFunctionExecution/fetch_xhr_spec.ts
arunvjn: Changed the file and the spec name

app/client/src/workers/HTTPRequestOverride.ts
sharat87: I see that we are disabling XHR completely. Does this mean that including any libraries that need XHR to run, won't be supported? 🤔 😞

Do you think we can create a custom XHR class, that's actually powered by `fetch` under the hood? Or is that too much? 🫤

arunvjn: Yes, libraries that needs XHR cannot be run. 
Custom implementation of XHR could be attempted but synchronous request feature via XHR might be slightly complicated to implement.

sharat87: Oh yeah, sync can't be done. But I doubt a very many libraries use that. Sync API calls not being possible, can actually be a good thing for us. May be not for v1, but let's see if we can create such an XHR. I suspect a _lot_ of libraries will need this instead of `fetch`.

app/client/src/workers/HTTPRequestOverride.ts
```diff
@@ -0,0 +1,12 @@
+import { register } from "fetch-intercept";
+
+const _originalXMLHttpRequest = self.XMLHttpRequest;
+
```
sharat87, line 4: This isn't needed may be?

```suggestion
```

app/client/src/workers/evaluate.ts
```diff
@@ -296,7 +298,7 @@ export default function evaluateSync(
     }
 
     try {
-      result = eval(script);
+      result = (1, eval)(script);
```
sharat87, line 301: This is a non-obvious piece of code. Can you add a comment with a link to MDN on what this does, and why this is important please?

sharat87 (commented):
Minor changes.

app/client/src/workers/HTTPRequestOverride.ts
arunvjn: Done

app/client/src/workers/HTTPRequestOverride.ts
sharat87: Really sorry, but I'm afraid my brain is back on this.

In the signatures you gave for `fetch` function, consider `[ request: Request, options? : Record<string, unknown> ]`. The `options?` means that the second argument is optional, right? That's what the `?` means right? If yes, then `args[1]` will de `undefined`, and this line will throw an error right?

(This is the problem I'm trying to hint at earlier also, but I confused myself. 😅)

sharat87: Ah, TIL. I recall seeing an NPE from that, but I can't reproduce it now, I was probably not drunk enough at the time.

Cool, thanks.

sharat87 (approved):
Thank you @arunvjn for sticking with me and tolerating my back-and-forth. I think this is now in a good shape. Please take it off 🛫.

app/client/src/workers/HTTPRequestOverride.ts
```diff
@@ -5,6 +5,10 @@ export default function interceptAndOverrideHttpRequest() {
     writable: false,
     configurable: false,
     value: function(...args: any) {
+      if (!self.ALLOW_ASYNC) {
+        self.IS_ASYNC = true;
+        return;
+      }
```
sharat87, line 11: @arunvjn, can you give me a brief on the significance of this block here please? 🤔

arunvjn: Currently, as we type away on JS Editor, we do a dry run(execute) of the function that is being edited to determine its nature. We mark a function async if it uses an Appsmith function or returns a promise. The function that is marked "async" is then disabled on data fields in property pane. `IS_ASYNC ` flag which gets set marks the function that contains fetch call as 'async'.

We will soon be removing this dry run. I believe there is a draft PR for it already.

sharat87: Thank you. Does this dry run also call `eval`, or is it a static AST traversal? Can you point me to the place in code where this happens please? 🙂

arunvjn: Here's the function. https://github.com/appsmithorg/appsmith/blob/d7fc4784911cac1a1846d4624012070195c730b0/app/client/src/workers/Evaluation/evaluate.ts#L403

We simply call the JSObject's edited function. No eval or AST. Note that this method is fairly primitive and doesn't work that well.

AnandiKulkarni:
@arunvjn when you give wrong api name it remain in executing mode for really long time, there is no error displayed or no error catched, no linting errors for some places such as if we write wrong method .jason, also expecting autocomplete for .json or res.text or res.blob
trycatch: async () => {
try{
	return fetch('https://dmmy.restapiexample.com/api/v1/employees').then(res=>res.json()).then().then((res2)=>{
			return res2.data
		})
}catch(error){
console.error(error.message)
}}

arunvjn:
> @arunvjn Sometimes getting error while JSobject function is mapped to table. JSobject function has fetch method
> 
>  Screen.Recording.2022-11-14.at.2.02.03.PM.mov

Are you able to replicate it consistently ? 
Could you also try replicating it with appsmith's mock users API ?

AnandiKulkarni:
> > @arunvjn Sometimes getting error while JSobject function is mapped to table. JSobject function has fetch method
> > Screen.Recording.2022-11-14.at.2.02.03.PM.mov
> 
> Are you able to replicate it consistently ? Could you also try replicating it with appsmith's mock users API ?

@arunvjn  yes It worked with Github api, it seems it is api issue

sharat87 (approved):

Aishwarya-U-R (commented):
Cypress looks good! 
**Note:** Not all cases from respective TestSmith is covered in Cypress since the coverage done via script is enough to proceed.

Rishabh-Rathod (approved):
LGTM

JhouXerox:
image
can't get http code in res

PR #17850, opened 2022-10-24
Enable feature to call an action after a property update

riodeuno:
## Description
- Add a feature to call a redux action after a property update.
- Along with `modify` and `remove` lists, `batchUpdateProproperty` action can also include `postUpdateActions`
- `postUpdateActions` is an array of `ReduxActionType`
- The default payload included in `postUpdateActions` is `widgetId`.

Fixes # TBD

## Type of change
- New feature (non-breaking change which adds functionality in codebase)
- Internal API Change

## How Has This Been Tested?
- If all existing tests pass, this feature does not break the current application
- New cypress test will be added to verify this in a subsequent PR
- Unit tests to be added

### Test Plan
- TBD

### Issues raised during DP testing
- NA

## Checklist:

- [NA] I have made corresponding changes to the documentation

### DIFF
--- a/app/client/src/sagas/WidgetOperationSagas.tsx
+++ b/app/client/src/sagas/WidgetOperationSagas.tsx
@@ -1,6 +1,7 @@
 import {
   ReduxAction,
   ReduxActionErrorTypes,
+  ReduxActionType,
   ReduxActionTypes,
   WidgetReduxActionTypes,
 } from "@appsmith/constants/ReduxActionConstants";
@@ -526,7 +527,7 @@ export function* getPropertiesUpdatedWidget(
 ) {
   const { dynamicUpdates, updates, widgetId } = updatesObj;
 
-  const { modify = {}, remove = [], triggerPaths } = updates;
+  const { modify = {}, remove = [], postUpdateActions, triggerPaths } = updates;
 
   const stateWidget: WidgetProps = yield select(getWidget, widgetId);
 
@@ -571,7 +572,10 @@ export function* getPropertiesUpdatedWidget(
   // If there exists another spot in this workflow, where we are iterating over the dynamicTriggerPathList and dynamicBindingPathList, after
   // performing all updates to the widget, we can piggy back on that iteration to purge orphaned paths
   // I couldn't find it, so here it is.
-  return purgeOrphanedDynamicPaths(widget);
+  return {
+    updatedWidget: purgeOrphanedDynamicPaths(widget),
+    actionsToDispatch: postUpdateActions,
+  };
 }
 
 function* batchUpdateWidgetPropertySaga(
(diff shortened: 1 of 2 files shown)


app/client/src/sagas/WidgetOperationSagas.tsx
```diff
@@ -633,6 +658,15 @@ function* batchUpdateMultipleWidgetsPropertiesSaga(
       updatedWidgetIds,
     }),
   );
+  for (const updatedWidgetAndActions of updatedWidgetsAndActionsToDispatch) {
+    const uniqueActions = uniq(updatedWidgetAndActions.actionsToDispatch);
+    for (const actionType of uniqueActions) {
+      yield put({
```
rahulramesha, line 664: @riodeuno, just to confirm. Does any of the possible actions being dispatched in loop update the the redux state? If so, it might cause a lot of performance issues.

riodeuno: @rahulramesha That's a good point. Technically, it is possible that multiple actions will be called. However, from the platform's point of view, there isn't a way to optimise this because the platform needs to be agnostic of which actions to run or batch. 
Today, it doesn't cause an issue, because we call it once when updating the minDynamicHeight and maxDynamicHeight

rahulramesha: My only concern is that, when we update state in a loop it just adds a potential leak for performance cuz it triggers every single selector if they have `deepEqual` it just adds a lot of milliseconds over multiple updates. We encountered a similar leak like this some time ago which just froze the UI for many seconds.

riodeuno: If we look at the logic, we can see that `postUpdateActions` is one single set for one `batchUpdateProperty`. So, to run into this issue is to intentionally call a large array of actions in one `batchUpdateProperty`. This is unlikely, unless we intend for such a thing to happen.

With that said, I agree with your concern, it is simple to call a large number of actions. This particular feature is necessary for us to maintain certain abstractions necessary for the auto height with limits to work correctly. Mostly because, I don't have a better idea. 

So, to make sure we avoid such a thing happening, we need to add some limits.
What we could do is make this a single action instead of an array. It will serve our current purpose. I'll make this change.

marks0351 (changes requested):

app/client/src/sagas/WidgetOperationSagas.tsx
riodeuno: This works if we're only looking at redux updates. But, here `redux-saga` will come into play. So, the concerns of @rahulramesha will still exist. I'll for now, make this into a single action (instead of an array) for every `batchUpdateProperty`

riodeuno:
@marks0351 This may work if we're only updating the redux store, however, once redux-sagas get involved, we might find this to not work as expected. So, I'll for now consider the original issue reported by @rahulramesha and make this into a single action for every batch update property action.

app/client/src/sagas/WidgetOperationSagas.tsx
riodeuno: Done. updated:

marks0351 (dismissed):

rahulramesha (dismissed):

marks0351 (approved):

PR #17851, opened 2022-10-24
feat: Ability to disable vertical resize handles

### LINKED ISSUE #17863: [Task] Add feature to disable vertical resize handles

### SubTasks

- Platform change which allows resizeable component to disable vertical resizing.

Note: A subsequent update is to be able to disable horizontal resizing

riodeuno:
## Description
- [ ] This feature adds ability to disable vertical resizing of widgets by disabling the handles

Fixes #17863 17863

## Type of change

## How Has This Been Tested?
- Manual - DONE
- Jest - DONE

## Checklist:

- [NA] I have made corresponding changes to the documentation


app/client/src/components/editorComponents/ResizableUtils.tsx
```diff
@@ -73,3 +74,29 @@ export const computeFinalRowCols = (
...
+ * A rudimentary function which based on horizontal and vertical resize enabled
+ * tells us whether a resize handle in a particular direction works
+ * Note: This works only if vertical or horizontal directions are provided.
+ * @param horizontalEnabled : boolean
+ * @param verticalEnabled : boolean
+ * @param direction : ReflowDirection
+ * @returns if resize is allowed in the direction provided
+ * Works only for vertical and horizontal directions
+ */
+export function isHandleResizeAllowed(
```
rahulramesha, line 88: NIT: I can see that we are using this is used to only indicate the dots. Could we possibly rename the method because it sounded a bit misleading as if the value is used to restrict someone from resizing.

app/client/src/components/editorComponents/ResizableUtils.tsx
```diff
@@ -73,3 +74,29 @@ export const computeFinalRowCols = (
...
+  horizontalEnabled: boolean,
+  verticalEnabled: boolean,
+  direction?: ReflowDirection,
+): boolean {
+  if (direction === ReflowDirection.TOP || direction === ReflowDirection.BOTTOM)
```
marks0351, line 93: do corner resize handles need special handling ?
just want to make sure that we have not missed them.

app/client/src/resizable/resize/index.tsx
```diff
@@ -318,13 +342,13 @@ export const Resizable = forwardRef(function Resizable(
     >
       {(_props) => (
         <ResizeWrapper
+          $prevents={pointerEvents}
```
marks0351, line 345: any reason for the dollar? looks like a new/odd convention

app/client/src/resizable/resizenreflow/index.tsx
```diff
@@ -454,22 +462,30 @@ export function ReflowResizable(props: ResizableProps) {
...
+      handle.handleDirection,
+    );
+    return (
+      <ResizableHandle
+        {...handle}
+        allowResize={props.allowResize}
+        checkForCollision={checkForCollision}
+        disableDot={disableDot}
+        key={index}
+        onStart={() => {
```
marks0351, line 478: should we have this function declared before the map it creates new function for each handle.

marks0351 (changes requested):

app/client/src/components/editorComponents/ResizableUtils.tsx
riodeuno: That's a point we've considered. Right now, we don't need to consider the corner resize handles and I don't forsee a scenario where we would as the resizable component today only handles vertical or horizontal resize. This means that if only horizontal resize is available, corner resize will only resize horizontally. If both are disabled, the resize handles will not show up at all.

app/client/src/components/editorComponents/ResizableComponent.tsx
```diff
@@ -269,20 +269,33 @@ export const ResizableComponent = memo(function ResizableComponent(
...
 
+  const snapGrid = useMemo(
+    () => ({
+      x: props.parentColumnSpace,
+      y: props.parentRowSpace,
+    }),
+    [props.parentColumnSpace, props.parentRowSpace],
+  );
+
+  const isVerticalResizeEnabled = useMemo(() => {
```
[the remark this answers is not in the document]
riodeuno: Yep, it was necessary when used with dynamic height. We can remove it for now.

app/client/src/components/editorComponents/ResizableUtils.tsx
riodeuno: This function is agnostic of its usage. So, we're using it for dots today, but it could be used for other scenarios as well. I'd like to keep this agnostic, what do you think?

app/client/src/resizable/resizenreflow/index.tsx
riodeuno: Yes absolutely, great point! This is however outside the scope of this PR. Would you mind creating a task for this?

marks0351 (approved):

app/client/src/components/editorComponents/ResizableComponent.tsx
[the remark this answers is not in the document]
rahulramesha: @riodeuno this still seems to be memoized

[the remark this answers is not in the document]
riodeuno: Right, fixing it now. ... and done.

marks0351 (approved):

PR #17854, opened 2022-10-24
feat: Use widget features during registration

riodeuno:
## Description
- [ ] Use widget features to enhance property pane configurations
- [ ] Use widget features to add default values to widget when they're created

## Type of change

## How Has This Been Tested?
- Manual
- Jest PENDING


app/client/src/utils/WidgetRegisterHelpers.tsx
```diff
@@ -50,9 +54,17 @@ export const registerWidget = (Widget: any, config: WidgetConfiguration) => {
...
+        {},
+        WidgetFeatureProps[registeredFeature as RegisteredWidgetFeatures],
+        WidgetFeaturePropertyEnhancements[
+          registeredFeature as RegisteredWidgetFeatures
+        ](config),
+      );
+    });
   }
   const _config = {
     ...features,
```
riodeuno, line 70: Move this spread below ...config.defaults, as we want the features to override the defaults.

app/client/src/utils/WidgetRegisterHelpers.tsx
```diff
@@ -50,13 +54,21 @@ export const registerWidget = (Widget: any, config: WidgetConfiguration) => {
 };
 
 export const configureWidget = (config: WidgetConfiguration) => {
-  let features = {};
-  if (config.features && config.features.dynamicHeight) {
-    features = Object.assign({}, WidgetFeatureProps.DYNAMIC_HEIGHT);
+  let features: Record<string, unknown> = {};
+  if (config.features) {
+    Object.keys(config.features).forEach((registeredFeature: string) => {
```
marks0351, line 59: suggestion: you cud use reduce instead of forEach.

marks0351 (approved):

app/client/src/utils/WidgetRegisterHelpers.tsx
riodeuno: @marks0351 could you elaborate on the advantages here? IMO, forEach is easier to grok.

marks0351: it's just that reduce is the actual function that does the exact same thing that is being done with the variable initiation and forEach. for me reduce and forEach both are readable. also there are no special advantages coz of this.

rahulramesha (approved):

PR #17873, opened 2022-10-25
fix: add connection pool to MySQL plugin and fix issues due to Spring upgrade

### LINKED ISSUE #8419: [Bug]-[1966]:Connection pools in Mysql plugin lead to runaway connections

### Current Behavior

When we try to invoke a MySQL query in a loop within Appsmith, the R2DBC library for MySQL is not using the existing connection object and ends up creating a new connection for each query. This should not be the case. We should pick a connection from a pre-specified pool of connections and only use those. Else, we might end up bringing down MySQL servers with the error "Too many connections"

### Steps To Reproduce

```markdown
1. Create a MySQL query action
2. Invoke this action using a for-loop on click of a button
3. Monitor the existing number of connections on the MySQL server.
```

### Environment

Production

### Version

Cloud

### LINKED ISSUE #17481: [Task]: MySQL plugin: Add connection pool

### SubTasks

- Introduce connection pool mechanism to MySQL plugin.

### LINKED ISSUE #19507: [Bug]: MySQL: fix failing JUnit TCs

### Description

Post the replacement of DB driver with MariaDB driver, all JUnit TC would fail. Currently all Junit TC have been commented out to make the CI work. Need to uncomment them and make sure that they work. 

### Steps To Reproduce

.

### Public Sample App

_No response_

### Issue video log

_No response_

### Version

cloud, self hosted

### LINKED ISSUE #19511: [Bug]: MySQL Preferred mode throws internal server error

### Description

When I connect to MySQL server with the SSL set to Preferred mode, it throws an internal server error. 

### Steps To Reproduce

1. Create a  My SQL new data source and enter all details
2. When we keep the SSL mode as default, the connection goes through ok
3. When we connect with SSL mode as Preferred, it throws an internal server error. 

### Public Sample App

_No response_

### Issue video log

_No response_

### Version

Release

sumitsum:
## Description
- This PR adds connection pool mechanism to MySQL plugin. So far, we used to create one connection per datasource instance and all the queries used to run on this one connection. This PR introduces a change to create a connection pool instead of a single connection for each datasource instance. The following configuration is used with the connection pool: 
  - max pool size = 5
  - max idle time for a connection = 10 min
- Minor code refactor to move some of the functions into helper utility classes. 
- Fix to support the new MariaDB driver that got added recently to replace the older MySQL r2dbc driver that was found incompatible with newer Spring version. 
  - Fixed all broken JUnit TCs. 
  - Added migration code to replace `Preferred` SSL option with `Default` since the new driver does not support it as MariaDB does not have this mode. 
    - Also, removed `Preferred` SSL option from the datasource config for the same reason.

Fixes #17481 
Fixes #8419
Fixes #19507 
Fixes #19511 

## Type of change

## How Has This Been Tested?
- Manual
- All Existing JUnit TCs pass

## Checklist:

Front logo [Front conversations](https://app.frontapp.com/open/top_5igy4)


Parthvi12:
@sumitsum Cypress test MySQLNoiseTest_spec.js in which we kill mySQL session is failing while asserting error status code in /api/v1/actions/execute call , if this is intentional, can we please update cypress test too.
Screenshot 2022-10-26 at 11 39 07 AM

app/server/appsmith-plugins/mysqlPlugin/src/main/java/com/external/plugins/MySqlPlugin.java
```diff
@@ -499,110 +510,27 @@ private boolean getIsSelectOrShowOrDescQuery(String query) {
...
-                                    "Appsmith server has found an unexpected SSL option: " + sslAuthType + ". Please reach out to" +
-                                            " Appsmith customer support to resolve this."
-                            )
-                    );
+        public Mono<ConnectionPool> datasourceCreate(DatasourceConfiguration datasourceConfiguration) {
+            ConnectionPool pool = null;
+            try {
+                pool = getNewConnectionPool(datasourceConfiguration);
+            } catch (AppsmithPluginException e) {
+                Mono.error(e);
```
nidhi-nair, line 525: Are we missing a return keyword here?

app/server/appsmith-plugins/mysqlPlugin/src/main/java/com/external/plugins/MySqlPlugin.java
```diff
@@ -614,258 +542,72 @@ public void datasourceDestroy(Connection connection) {
...
         @Override
-        public Mono<DatasourceStructure> getStructure(Connection connection, DatasourceConfiguration datasourceConfiguration) {
+        public Mono<DatasourceStructure> getStructure(ConnectionPool connectionPool,
+                                                      DatasourceConfiguration datasourceConfiguration) {
             final DatasourceStructure structure = new DatasourceStructure();
             final Map<String, DatasourceStructure.Table> tablesByName = new LinkedHashMap<>();
             final Map<String, DatasourceStructure.Key> keyRegistry = new HashMap<>();
 
-            return Mono.from(connection.validate(ValidationDepth.REMOTE))
+            return Mono.usingWhen(
```
nidhi-nair, line 555: Nice change with the using resource kind of usage! :)

app/server/appsmith-plugins/mysqlPlugin/src/main/java/com/external/plugins/MySqlPlugin.java
```diff
@@ -63,24 +59,23 @@
...
-import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 import static com.appsmith.external.constants.ActionConstants.ACTION_CONFIGURATION_BODY;
 import static com.appsmith.external.helpers.PluginUtils.MATCH_QUOTED_WORDS_REGEX;
 import static com.appsmith.external.helpers.PluginUtils.getIdenticalColumns;
 import static com.appsmith.external.helpers.PluginUtils.getPSParamLabel;
 import static com.appsmith.external.helpers.SmartSubstitutionHelper.replaceQuestionMarkWithDollarIndex;
-import static io.r2dbc.spi.ConnectionFactoryOptions.SSL;
+import static com.external.utils.MySqlDatasourceUtils.getNewConnectionPool;
```
nidhi-nair, line 69: Thanks for pulling this out, how do you think we can best structure this so it is a mandated repeatable pattern in different plugins?

app/server/appsmith-plugins/mysqlPlugin/src/main/java/com/external/plugins/MySqlPlugin.java
sumitsum: Thanks :)

app/server/appsmith-plugins/mysqlPlugin/src/main/java/com/external/plugins/MySqlPlugin.java
sumitsum: Hi @nidhi-nair , thanks for bringing up this question. It would be great to have such kind of a pattern. However, I don't have an immediate answer at the moment (please imagine tears here, for the lack of emoji), hence I have created an issue to track it : https://github.com/appsmithorg/appsmith/issues/17904 . Let us use this new issue to jot down our thoughts regarding the same :)

app/server/appsmith-plugins/mysqlPlugin/src/main/java/com/external/plugins/MySqlPlugin.java
sumitsum: @nidhi-nair nice catch ! Thanks a lot :). I have fixed this.

sumitsum:
> @sumitsum Cypress test MySQLNoiseTest_spec.js in which we kill mySQL session is failing while asserting error status code in /api/v1/actions/execute call , if this is intentional, can we please update cypress test too. Screenshot 2022-10-26 at 11 39 07 AM

@Parthvi12 thanks for explaining the cypress failure here. I have updated the test, so that should fix it hopefully.

nidhi-nair (approved):

app/server/appsmith-plugins/mysqlPlugin/src/main/resources/form.json
```diff
@@ -87,10 +87,6 @@
               "label": "Default",
               "value": "DEFAULT"
             },
-            {
```
nidhi-nair, line 90: For my understanding, why are we removing this option?

nidhi-nair: Ah okay, just saw your comment on the migration, never mind. :)

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
```diff
@@ -2912,4 +2913,21 @@ public void addIndicesRecommendedByMongoCloud(MongoTemplate mongoTemplate) {
 //            }
 //        }
 //    }
+
+    /**
+     * Since MySQL plugin's underlying driver has been changed to MariaDB driver, the `ssl-mode=preferred` is no
+     * longer supported. Hence, any such usage is being updated to `ssl-mode=default` by this method.
+     */
+    @ChangeSet(order = "040", id = "remove-preferred-ssl-mode-from-mysql", author = "")
```
nidhi-nair, line 2921: Could we please write this migration as a ChangeUnit?

nidhi-nair: Hmm .. Never mind actually, understanding the new format might take some time, let's restructure after release opens up.

app/client/cypress/integration/Smoke_TestSuite/ServerSideTests/Datasources/MySQLNoiseTest_spec.js
```diff
@@ -63,7 +63,7 @@ describe("MySQL noise test", function() {
       expect(response.body.data.statusCode).to.eq("200 OK");
     });
     cy.wait("@postExecute").then(({ response }) => {
-      expect(response.body.data.statusCode).to.eq("5004");
```
nidhi-nair, line 66: What does this change indicate?

nidhi-nair: Let's run Cypress to verify this in that case

nidhi-nair (dismissed):

btsgh:
Testing the PR branch on localhost - 
[19511](https://github.com/appsmithorg/appsmith/issues/19511) - Do not see the Preferred option at all in the dropdown now. Tried on creating a new datasource and also on an existing datasource.

trishaanand (approved):

PR #17879, opened 2022-10-26
fix: auto height min max revalidation

ankurrsinghal:
Used the new `dependentPaths` in the ValiationConfig to create a dependency between min and max so that changing max we can validate the min again.


riodeuno:
@ankurrsinghal Please create a new PR for this. This seems to have many conflicts. I've merged the latest release into the base branch, so a new PR should be cleaner.

PR #17883, opened 2022-10-26
fix: listv2 page no

### LINKED ISSUE #11739: [Bug]: List Widget's PageNo does not reflect correctly

### Description

.png :-
Screenshot 2022-03-09 at 5 17 30 PM

### Steps To Reproduce

1. Create a ListWidget with data spanning multiple pages
2. Map the ListWidget's PageNo to Input text to see the value changing as user moves pages
3. Observe that the values are not computed correctly & remains the same as page 1
4. **Expected**: Values should be computed rightly

### Public Sample App

_No response_

### Version

Cloud Release v1.6.14

Tooluloope:
## Description

Fix Listv2 page No and also PageSize

Page Size could change

- Infinite scroll is toggled since the Pagination controls are removed
- The Template container height changes 
- The List component changes
This causes us to recompute and update the `pageSize` meta property

Page number is current dependent on the pagination controls

NB: Page number doesn't work for infinite scrolling

Fixes #11739

## Type of change

## How Has This Been Tested?
- Manual

## Checklist:


app/client/src/widgets/ListWidgetV2/widget/derived.js
```diff
@@ -35,4 +35,8 @@ export default {
     return { currentItem: structure, currentIndex: "" };
   },
   //
+  getPageSize: (props, moment, _) => {
+    return props.currentViewItems ? props.currentViewItems.length : 0;
```
ashit-rath, line 39: I don't this this would be the right way to measure a page size.
Eg. If I am on the last page and it shows only one row but the view can support 3 rows. the `currentViewItems` would have only one item in the array and would thus show `pageSize` as 1 instead of 3.

Tooluloope: Oh yeah, I worked on this, 
I made `pageSize` to be the amount of items the List WIdget can take without scrolling.
Also added cypress test

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -201,27 +191,27 @@ class ListWidget extends BaseWidget<
...
-
     if (this.props.serverSidePaginationEnabled) {
+      if (!this.props.pageNo) this.props.updateWidgetMetaProperty("pageNo", 1);
+      // run onPageSizeChange if user resize widgets
+      if (
+        this.props.onPageSizeChange &&
+        this.props.pageSize !== prevProps.pageSize
+      ) {
+        super.executeAction({
+          triggerPropertyName: "onPageSizeChange",
```
ashit-rath, line 202: When onPageSize change is triggered, what is the value of the pageSize when used in a query `List1.pageSize`? old or new?
Note: Add a test case for this if not added

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -201,27 +191,27 @@ class ListWidget extends BaseWidget<
...
+          },
+        });
+      }
+    }
+
+    // Should this be triggered?
+    if (this.props.serverSidePaginationEnabled && this.props.onPageSizeChange) {
       if (
-        this.props.serverSidePaginationEnabled === true &&
+        this.props.serverSidePaginationEnabled &&
```
ashit-rath, line 214: Add a cypress test for when server side pagination in enabled and if the onPageSizeChange event is triggered or not.

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -450,6 +438,27 @@ class ListWidget extends BaseWidget<
...
 
+  updatePageSize = () => {
+    const pageSize = this.getPageSize();
+    if (pageSize !== this.props.pageSize) {
+      this.props.updateWidgetMetaProperty("pageSize", pageSize);
+    }
+  };
+
+  shouldUpdatePageSize = (prevProps: ListWidgetProps<WidgetProps>) => {
+    const hasListComponentHeightChanged =
```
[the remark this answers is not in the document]
Tooluloope: Yeah, we can, but I didn't want to compute `getPageSize` everytime

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -848,9 +838,9 @@ class ListWidget extends BaseWidget<
               accentColor={this.props.accentColor}
               borderRadius={this.props.borderRadius}
               boxShadow={this.props.boxShadow}
-              current={this.state.page}
+              current={this.props.pageNo}
```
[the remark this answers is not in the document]
Tooluloope: this.getPageSize() gives pageSize and not pageNo

app/client/src/widgets/ListWidgetV2/widget/index.tsx
Tooluloope: Should this even be triggered, 
All the user does is toggle `serverSidePaginationEnabled` and pageSize doesn't change.

The only time it'd change is when infiniteScroll is toggled due to the Pagination control being absent

ashit-rath: Even I thought it won't change but then drop a list widget and enable server side pagination and see the page size change from 3 -> 2 due to pagination controls

app/client/src/widgets/ListWidgetV2/widget/index.tsx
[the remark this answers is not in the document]
ashit-rath: I don't think the getPageSize computation is too high. The reason why I'm not feeling confident about this check is because the conditions here are different and in future there may be other variables that would dictate the change in page size calculation. If we directly check from source i.e the `getPageSize` then we don't have to worry about anything.

If you feel this is additional computation then add a property on the class `this.pageSize` and do the computation once in componentDidMount and componentDidUpdate and refer everywhere.

app/client/src/widgets/ListWidgetV2/widget/index.tsx
Tooluloope: This looks like we should call `onPageSizeChange` when the `serverSidePaginationEnabled` is true.
The page change occurred in clientSide and is initialized as any value in `serverSidePagination`.

https://user-images.githubusercontent.com/31691737/198381365-4f81220c-ee4a-4dfc-adb6-c0ab690823d9.mov

app/client/src/widgets/ListWidgetV2/widget/index.tsx
Tooluloope: The New Value

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -164,9 +155,13 @@ class ListWidget extends BaseWidget<
     });
     this.prevMetaContainerNames = [];
     this.componentRef = createRef<HTMLDivElement>();
+    this.pageSize = 0;
```
ashit-rath, line 158: In constructor itself, I think we have all the props required to calculate `pageSize`. Can we check that and implement if possible?

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -848,9 +815,9 @@ class ListWidget extends BaseWidget<
               accentColor={this.props.accentColor}
               borderRadius={this.props.borderRadius}
               boxShadow={this.props.boxShadow}
-              current={this.state.page}
+              current={this.props.pageNo}
```
ashit-rath, line 818: Can we rename `current` to `pageNo`? 
Let's continuously refactor our code to make it proper.

ashit-rath (approved):

laveena-en:
Hey @Tooluloope The DP doesn’t seem to be up yet. Can you please check?
image

Tooluloope:
@laveena-en you can view it now

laveena-en:
Tested and verified.

PR #17885, opened 2022-10-26
fix: error message for input widget updated for when the default value is more than the max chars allowed

### LINKED ISSUE #14830: [Bug]: Input widget issues

iamrkcheers:
## Description

Error message when Default Text has more characters than the Max Chars changed to "Default Text length must be less than or equal to 'Max Chars' allowed".

Fixes #14830

## Type of change

## How Has This Been Tested?

- Manual
- Jest

## Checklist:


sbalaji1192 (dismissed):

laveena-en:
Tested and verified.
The text will now be `Default text length must be less than ${max} characters`

app/client/src/widgets/InputWidgetV2/widget/index.test.tsx
```diff
@@ -212,3 +213,12 @@ describe("maxValueValidation - ", () => {
...
   });
 });
+
+describe("defaultTextMaxCharValidation", () => {
+  const values = { maxNum: 5 } as InputWidgetProps;
+  it("checks for INPUT_DEFAULT_TEXT_MAX_CHAR_ERROR string", () => {
+    expect(INPUT_DEFAULT_TEXT_MAX_CHAR_ERROR(Number(values.maxNum))).toBe(
+      `Default text length must be less than or equal to ${values.maxNum} characters`,
+    );
+  });
```
mohanarpit, line 223: Sorry, I don't understand this test case. This test seems to be testing Javascript's string interpolation rather than an Apspmith feature.

app/client/src/widgets/InputWidgetV2/widget/index.test.tsx
```diff
@@ -4,6 +4,7 @@ import {
   minValueValidation,
   maxValueValidation,
 } from "./index";
+import { INPUT_DEFAULT_TEXT_MAX_CHAR_ERROR } from "./../../../ce/constants/messages";
```
mohanarpit, line 7: Any file in the `ce` folder should ALWAYS be imported as `@appsmith/constants/messages` because the compiler will use different files in CE & EE. This import will break in our EE repo.

app/client/src/widgets/InputWidgetV2/widget/index.test.tsx
iamrkcheers: This makes sense. Removing the test case altogether.

app/client/src/widgets/InputWidgetV2/widget/index.test.tsx
iamrkcheers: Making the necessary change.

sbalaji1192 (dismissed):

sbalaji1192 (approved):

PR #17917, opened 2022-10-27
ci:  Add S3 credentials to environment and create run meta record at the time of checkout

### LINKED ISSUE #17916: [Task]: Create performance run meta record at the beginning of workflow

### SubTasks

Create run meta record at the beginning of workflow, otherwise if the perf tests or delayed or the build takes time, we will be comparing results with new code.

SatishGandham:
## Description

- Adds S3 credentials to an environment variable.
- Rename /perf-test command workflow 
- Create run meta when the code is checked out.

Fixes #17916

## Type of change
- Some changes and improvements to CI

## How Has This Been Tested?
To be tested on CI

## Checklist:


.github/workflows/client-build.yml
```diff
@@ -48,6 +48,25 @@ jobs:
...
+          cmd: 'jq .host ${{ secrets.APPSMITH_PERFORMANCE_DB_CONFIG }} -r'
+
+      - name: Extract DB Password
+        uses: sergeysova/jq-action@v2
+        id: get_db_password
+        with: 
+          cmd: 'jq .password ${{ secrets.APPSMITH_PERFORMANCE_DB_CONFIG }} -r'
+
+      - name: Create Perf Meta
+        run: PGPASSWORD=${{steps.get_db_password.outputs.value}} psql -h ${{steps.get_db_host.outputs.value}} -U aforce_admin -d perf-infra -c 'INSERT INTO public.run_meta (gh_run_id, gh_run_attempt, pull_request_id,is_active)
```
SatishGandham, line 67: @mohanarpit , @sharat87 
Can you please check if the query syntax is correct? Should the variable be in quotes?
Also is the jq syntax correct?

If these steps fail, will the job fail?

.github/workflows/client-build.yml
```diff
@@ -48,6 +48,25 @@ jobs:
...
+        id: get_db_host
+        with: 
+          cmd: 'jq .host ${{ secrets.APPSMITH_PERFORMANCE_DB_CONFIG }} -r'
+
+      - name: Extract DB Password
+        uses: sergeysova/jq-action@v2
+        id: get_db_password
+        with: 
+          cmd: 'jq .password ${{ secrets.APPSMITH_PERFORMANCE_DB_CONFIG }} -r'
+
```
SatishGandham, line 65: @mohanarpit , @sharat87
Can you please check if the query syntax is correct? Should the variable be in quotes?
Also is the jq syntax correct?

If these steps fail, will the job fail?

mohanarpit: My guess is that you will require double quotes for the variable because single quotes don't expand variables.

SatishGandham: @mohanarpit , Will make that change. Any way to ensure failure here doesn't stop the build process? If not I will create a separate workflow temporarily to test this code.

sharat87: Single quotes should work fine, but this will break because the JSON will be placed as part of the shell command, and _that_ will fail. Try something like this:

```yaml
          cmd: |
            jq .password '${{ secrets.APPSMITH_PERFORMANCE_DB_CONFIG }}' -r
```

But, even there, I'd _strongly_ recommend using two separate secrets, one for host, and one for password, instead of one secret with a JSON in it. One, we won't have to rely on this third party action. Two, you can just start using them, `${{ secrets.APPSMITH_PERFORMANCE_DB_PASSWORD }}` instead of `${{steps.get_db_password.outputs.value}}`. So, less code, less bugs. Three, it's explicit. Just from seeing this, I'd be a little hesitant to change the secret's value, since I don't know what else keys are in there, that I don't know of.

SatishGandham: I choose this as we already have a secret with DB creds in JSON format. 
Can DB user, DB Name and port be public?

Can you check the `Create Perf Meta` step below help with the quotes.

sharat87: Oh we do?! 😞

Oh well. Fine I suppose. We have a plan to move away from Secrets non-soon-ish, and this will get revised then.

User, name and port are fine. Host and password aren't. I'd prefer port to also be secret, but, not too strong. 🙂

.github/workflows/client-build.yml
```diff
@@ -48,6 +48,25 @@ jobs:
...
+      - name: Create Perf Meta
+        run: PGPASSWORD=${{steps.get_db_password.outputs.value}} psql -h ${{steps.get_db_host.outputs.value}} -U aforce_admin -d perf-infra -c 'INSERT INTO public.run_meta (gh_run_id, gh_run_attempt, pull_request_id,is_active)
+VALUES (${{env.GITHUB_RUN_ID}}, ${{env,GITHUB_RUN_ATTEMPT}}, ${{ inputs.pr }}, FALSE)'
```
sharat87:
```suggestion
        run: |
          PGPASSWORD="${{steps.get_db_password.outputs.value}}" psql -h "${{steps.get_db_host.outputs.value}}" -U aforce_admin -d perf-infra -c '
            INSERT INTO public.run_meta (gh_run_id, gh_run_attempt, pull_request_id,is_active)
            VALUES (${{env.GITHUB_RUN_ID}}, ${{env,GITHUB_RUN_ATTEMPT}}, ${{ inputs.pr }}, FALSE)
          '
```

sharat87 (approved):

sharat87 (approved):

PR #17929, opened 2022-10-28
feat: Code Scanner Enhancements

### LINKED ISSUE #17367: [Feature]: Code scanner enhancements

### Summary

- When working to scan multiple codes 
   - users have to close and open the code scanner frequently to scan items. Instead it could be opened just like the camera widget.
   - users always have to select the camera which they want to use to scan the item, this introduces an extra click.
- Code-scanner should also reverse the camera just like the camera widget.

Makes it more convenient to adopt the code scanner for power use-cases

Front logo [Front conversations](https://app.frontapp.com/open/top_571xo)

### LINKED ISSUE #17954: [Task]: Background detection should be disabled in Code Scanner Widget

### SubTasks

Check out this conversation - https://github.com/appsmithorg/appsmith/pull/17929#issuecomment-1294848438
 
image

dhruvikn:
## Description

Adds a new property Scanner Variant that sets how the code scanner will look and behave. If set to "Always on", the scanner will be visible and scanning all the time. If set to "Click to Scan", the scanner will pop up inside a modal and start scanning when the user clicks on the button. 

Also, mirroring the camera is now just a button away.

Moreover, the widget now will not scan for codes in the background.

Fixes #17367
Fixes #17954

## Type of change

## How Has This Been Tested?

- Manual
- Cypress

Cypress Test Suite for Code Scanner

## Checklist:


dhruvikn:
@dilippitchika, I have fixed the aspect ratio issue for the video. Now the video feed will maintain and will be fully visible at all times. However, I wasn't able to contain the scanning animation to the video only. Currently, the scanning animation is created using a [pseudo element](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements) on the video's container. To contain it in the video only would mean creating a pseudo-element on the `<video>` tag which, unfortunately, [is not possible](https://stackoverflow.com/a/27701731/1722994).

cc: @jsartisan any suggestions are welcome and appreciated. 🙌

laveena-en:
Tested and verified for:
-`Always on` property in the widget
- Mirroring
On edit and view mode

dilippitchika:
@laveena-en have you already tested this in phone? If not, please do

dilippitchika:
@dhruvikn few suggestions
1. Make always on the default and increase the size.
3. Change the label to "Scanner layout"
5. In the button group make always on the first one

Make sure you add doc ticket for this and make necessary changes

laveena-en:
@dhruvikn As discussed, the widget continues to capture QR/Bar codes in the background thus opening up multiple pages on both desktop and mobile. 
P.S The camera widget's behaviour is also the same. Please do check
cc: @dilippitchika

laveena-en:
Recalling my previous sign off post discussion with @dhruvikn

dhruvikn:
> the widget continues to capture QR/Bar codes in the background thus opening up multiple pages on both desktop and mobile.

@laveena-en, this is a great catch, super super work!

Just had a chat with @dilippitchika, **background detection should be disabled** for better UX. I'll be creating a new issue for this as this wasn't in the initial scope of #17367. However, we will solve the issue in this PR only. I'll keep you folks once I address this.

dhruvikn:
Created a new issue #17954 which is addressed now in this PR.

cc: @laveena-en @dilippitchika

keyurparalkar (approved):
@dhruvikn Can you please add the cypress test cases in the PR description as well?

laveena-en:
@dhruvikn Looks like the deployment isn’t up yet. Can you check this please? 
image

dhruvikn:
Can you please use this one? @laveena-en 

https://appsmith-e7secc8uk-get-appsmith.vercel.app/

laveena-en:
@dhruvikn A couple of callouts:
1. The QR/barcode isn’t getting scanned via the front camera when the widget is set to `always on` on mobile devices
2. Also, the scanner isn’t picking up the QR code on iOS while being on the `always on` mode via both the front and back camera. Works fine on the `click to scan ` mode though

Sharing an app that you can reproduce this on: https://appsmith-k9oqm8jw0-get-appsmith.vercel.app/app/new-notepad/page1-6361feb1ce5a0f64a008c08a

dhruvikn:
After a chat with @laveena-en, we can confirm that the code is being detected on iOS with both front and back.

However, the navigateTo function is not working when set to Always On on mobile. Will be looking into this once.

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/Widgets/CodeScanner/CodeScanner_spec.js
```diff
@@ -2,62 +2,245 @@ const explorer = require("../../../../../locators/explorerlocators.json");
...
+    cy.wait(200);
 
-  it("3. Code Scanner functionality to check enabled widget", function() {
-    cy.openPropertyPane(widgetName);
-    cy.togglebarDisable(commonlocators.disableCheckbox);
-    cy.PublishtheApp();
-    cy.get(publish.codescannerwidget + " " + "button").should("be.enabled");
-    cy.get(publish.backToEditor).click();
+    // Value will be undefined initially
+    cy.get(commonlocators.TextInside).should(
```
sbalaji1192, line 32: This assertion will pass even when the binding doesn't work

app/client/src/widgets/CodeScannerWidget/component/index.tsx
```diff
@@ -382,54 +482,88 @@ function CodeScannerComponent(props: CodeScannerComponentProps) {
...
+          isImageMirrored ? "mirror-video" : ""
+        }`}
+      >
+        {props.isDisabled && (
+          <DisabledOverlayer disabled={props.isDisabled}>
+            <CameraOfflineIcon />
+          </DisabledOverlayer>
+        )}
+
+        {!props.isDisabled && (
```
sbalaji1192, line 516: can you club this and the above interpolation with a ternary operator?

sbalaji1192 (dismissed):

laveena-en:
Tested and verified the functionality and callouts addressed in this DP.
- Verified the `Always ON` mode on desktop and mobile devices [Windows &Mac] [Android and iOS]
- Verified mirroring's functionality
- Tested the functionality's consistency via the front and back camera on mobile devices
- Tested onCodeDetected events workflow when `Always ON` mode is selected

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/Widgets/CodeScanner/CodeScanner_spec.js
dhruvikn: Let's just remove this case then?

sbalaji1192: Yeah. Remove it.

keyurparalkar (approved):

PR #17952, opened 2022-10-29
fix: Fix test email not respecting SSL option

### LINKED ISSUE #16273: [Bug]: Cannot change SMTP credentials without enabling TLS

### Description

In Admin Settings, if we want to change the SMTP username and password, we have to enable TLS, even if we don't want it. So, if we want to set credentials, but not use TLS, the flow is quite weird.

1. Enable TLS, so that the UI for credentials show up.
2. Set the credentials.
3. Disable TLS, since we don't want it.

With each of the above step requiring a "Save & Restart".

### Steps To Reproduce

Start a new Appsmith instance, and try to set SMTP credentials with TLS turned off.

Also see <https://theappsmith.slack.com/archives/C0385FH70E8/p1661138139809349>.

### Public Sample App

_No response_

### Version

Self Hosted

### LINKED ISSUE #16930: [Bug]: Test email doesn't respect TLS setting

### Description

In Admin Settings, the "Send Test Email" button doesn't send the value of whether TLS is enabled or not. The server, when sending the test email, always assumes TLS is disabled.

So, when the server doesn't expect TLS, the test email works, but actual emails from Appsmith fail.

This is a very confusing experience.

### Steps To Reproduce

Start maildev with the following configuration:

```bash
docker run --rm -p 9080:9080 -p 9025:9025 maildev/maildev --web 9080 --smtp 9025 --incoming-user rock --incoming-pass paper
```

Now start an Appsmith container, signup and go to Admin Settings, Email configuration. Fill up with the following

| Field | Value |
| --- | --- |
| Host | host.docker.internal |
| Port | 9025 |
| TLS | Enabled |
| Username | rock |
| Password | paper |

Now hit the `Send Test Email` button, and the email will be sent, and be visible at <http://localhost:9080>.

So since test email is working fine, go ahead and hit `Save and Restart`. Once Appsmith is back up and running, try to invite someone to a workspace, like `another@something.com`, and see that no email shows up at <http://localhost:9080>.

This is because TLS is turned on, and the Maildev server we started doesn't work with TLS. This was okay in Test email, since it ignored the fact that TLS is enabled, but the rest of Appsmith doesn't ignore this.

### Public Sample App

_No response_

### Version

Self Hosted

### LINKED ISSUE #20884: [Bug]: Appsmith shouldn't attempt to STARTTLS if TLS is disabled in the config.

### Description

When starting a server that has TLS disabled in the config, it will attempt STARTTLS anyway.

I expect the server to not try STARTTLS if TLS is disabled.

### Steps To Reproduce

1. Disable TLS in your instance config.
2. Start your instance and watch the logs.
3. Observer that STARTTLS is logged anyway.

### Public Sample App

_No response_

### Issue video log

_No response_

### Version

1.9.7

Front logo [Front conversations](https://app.frontapp.com/open/top_7mmfg)

sharat87:
Fixes #16930
Fixes #16273
Fixes #20884

For connecting to some SMTP servers with TLS, we need the `mail.smtp.ssl.enable` property to be set to `true`, in addition to the `mail.smtp.starttls.enable` property. We add that in this PR.

Also, the username/password fields are now _always_ visible. Previously, if you wanted to configure credentials, with TLS disabled, you have to enable TLS, set credentials, and then disable TLS again.

Also, the send-test-email API was ignoring the TLS setting in Admin Settings. This PR fixes that as well, but including that piece of information in the payload sent to the server, and the API to make use of it.


ankitakinger (dismissed):
FE changes looks good.

sharat87:
Above failure is marked as flaky.

shot-2023-05-18-08-16-22

sharat87:
Hey @ankitakinger @nidhi-nair, can one/both of you approve this, when you get a chance please? 🙏

ankitakinger (approved):

PR #17964, opened 2022-10-30
feat: auto height helpers

### LINKED ISSUE #17963: [Task]: Auto-height widget utils

### SubTasks

- [ ] Add helper functions to get the widget's max auto height in rows
- [ ] Add helper functions to get widget's min auto height in rows
- [ ] Add helper to check if widget has auto height enabled
- [ ] Add helper to check if widget has auto height enabled with limits

riodeuno:
## Description

- [ ] Add helper functions to get the widget's max auto height in rows
- [ ] Add helper functions to get widget's min auto height in rows
- [ ] Add helper to check if widget has auto height enabled
- [ ] Add helper to check if widget has auto height enabled with limits

Fixes #17963 

## Type of change

## How Has This Been Tested?
- Jest

## Checklist:

~~- [NA] I have made corresponding changes to the documentation~~


app/client/src/widgets/WidgetUtils.ts
```diff
@@ -715,3 +717,47 @@ export const flat = (array: DropdownOption[]) => {
...
+  props: WidgetProps,
+  withLimits = false,
+) => {
+  if (withLimits) {
+    return props.dynamicHeight === DynamicHeight.AUTO_HEIGHT_WITH_LIMITS;
+  }
+  return (
+    props.dynamicHeight === DynamicHeight.AUTO_HEIGHT ||
+    props.dynamicHeight === DynamicHeight.AUTO_HEIGHT_WITH_LIMITS
+  );
```
marks0351, line 736:
```suggestion
  return [DynamicHeight.AUTO_HEIGHT, DynamicHeight.AUTO_HEIGHT_WITH_LIMITS].includes(props.dynamicHeight);
```

marks0351: looks cleaner to me this way, just an opinion based suggestion not a review comment.

app/client/src/widgets/WidgetUtils.ts
```diff
@@ -715,3 +717,47 @@ export const flat = (array: DropdownOption[]) => {
...
+};
+
+/**
+ * Gets the max possible height for the widget
+ * @param props: WidgetProperties
+ * @returns: The max possible height of the widget (in rows)
+ */
+export function getWidgetMaxAutoHeight(props: WidgetProps) {
```
marks0351, line 744: both getWidgetMaxAutoHeight and getWidgetMinAutoHeight look similar can't we pass an additional prop and just use single function?

marks0351 (changes requested):

app/client/src/widgets/WidgetUtils.ts
riodeuno: I was actually conflicted about the `isDynamicHeightEnabledForWidget`'s `withLimits` argument. This pertains to what @IAmAnubhavSaini had posted regarding using "flag" arguments. I agree with his opinion, and would like to keep the function names descriptive and have them do one thing.

What is your opinion on the same?

app/client/src/widgets/WidgetUtils.ts
riodeuno: I like your suggestion. I usually also use an array when we have multiple values to check against. Here I left this as is, because we're never going to have more than these two checks. If we do, we might want to use an array.

What do you think?

marks0351: yea no issues then...

marks0351 (approved):

PR #17973, opened 2022-10-31
chore: dev workflow to benchmark builds

pratapaprasanna:
This is a dev workflow. Which will be deleted once we get the numbers comparison between buildjet runners and github-runners

### DIFF
--- a/.github/workflows/test-build-docker-image-git.yml
+++ b/.github/workflows/test-build-docker-image-git.yml
@@ -1,0 +1,1715 @@
+name: Test, build and push Docker Image on github-runner
+
+on:
+  # This line enables manual triggering of this workflow.
+  workflow_dispatch:
+
+  # trigger for pushes to release and master
+  push:
+    branches: [release, master]
+    paths:
+      - "app/client/**"
+      - "app/server/**"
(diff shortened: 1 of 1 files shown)


.github/workflows/test-build-docker-image-git.yml
```diff
@@ -0,0 +1,1706 @@
...
+      - name: Push server release image to Docker Hub
+        if: success() && github.ref == 'refs/heads/release'
+        uses: depot/build-push-action@v1
+        with:
+          context: app/server
+          push: true
+          build-args: |
+            APPSMITH_SEGMENT_CE_KEY=${{ secrets.APPSMITH_SEGMENT_CE_KEY_RELEASE }}
+          tags: |
+            ${{ secrets.DOCKER_HUB_ORGANIZATION }}/appsmith-server:${{steps.vars.outputs.tag}}
```
sharat87, line 1673: Since this is a test-workflow, can we either comment/disable pushing images, or push them to a separate organization? Like a Docker username may be? Instead of the official appsmith org?

sharat87: Same for the other push actions here.

PR #17994, opened 2022-10-31
feat: base widget Auto height enhancements

### LINKED ISSUE #17993: [Task] Auto - Height enhancements for BaseWidget

### SubTasks

The BaseWidget should do the following:
- [ ] Enable calls to update auto height of widget
- [ ] Check if this update should be applied
- [ ] Add the auto height overlay in edit mode
- [ ] Wrap widgets in auto height container

riodeuno:
## Description
The BaseWidget now does the following:
- [x] Enables calls to update auto height of widget
- [x] Checks if this update should be applied
- [x] Adds the auto height overlay in edit mode
- [x] Wraps widgets in auto height container

Fixes #17993

## Type of change

## How Has This Been Tested?

- Manual (Smoke test, should not break existing apps)
- Jest (For new functions added)
- Cypress (Existing tests)

## Checklist:

~~- [NA] I have made corresponding changes to the documentation~~

### DIFF
--- a/app/client/src/widgets/BaseWidget.tsx
+++ b/app/client/src/widgets/BaseWidget.tsx
@@ -5,12 +5,14 @@
  */
 import {
   CSSUnit,
+  GridDefaults,
   PositionType,
   RenderMode,
   RenderModes,
   WidgetType,
+  WIDGET_PADDING,
 } from "constants/WidgetConstants";
-import React, { Component, ReactNode } from "react";
+import React, { Component, CSSProperties, ReactNode } from "react";
 import { get, memoize } from "lodash";
 import DraggableComponent from "components/editorComponents/DraggableComponent";
 import SnipeableComponent from "components/editorComponents/SnipeableComponent";
@@ -37,6 +39,17 @@ import PreviewModeComponent from "components/editorComponents/PreviewModeCompone
 import { CanvasWidgetStructure } from "./constants";
 import { DataTreeWidget } from "entities/DataTree/dataTreeFactory";
 import Skeleton from "./Skeleton";
+
+import {
+  getWidgetMaxAutoHeight,
+  getWidgetMinAutoHeight,
+  isAutoHeightEnabledForWidget,
+  shouldUpdateWidgetHeightAutomatically,
+} from "./WidgetUtils";
+import AutoHeightOverlay from "components/editorComponents/AutoHeightOverlay";
+import log from "loglevel";
+import AutoHeightContainer from "./AutoHeightContainer";
+import { DynamicHeight } from "utils/WidgetFeatures";
 
 /***
  * BaseWidget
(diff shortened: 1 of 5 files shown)


riodeuno:
@rahulramesha @marks0351 The two components in this PR are placeholders, @ankurrsinghal will be creating separate PRs which implement them correctly.

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -344,9 +447,15 @@ abstract class BaseWidget<
       return <Skeleton />;
     }
 
-    return renderMode === RenderModes.CANVAS
-      ? this.getCanvasView()
-      : this.getPageView();
+    const content =
```
marks0351, line 450: better process content after the if check to avoid it's execution when not needed.

marks0351 (changes requested):

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -175,6 +188,35 @@ abstract class BaseWidget<
...
+    this change.
+    @param height number: Height of the widget's contents in pixels 
+    @return void
+  */
+  updateHeightAutomatically(height: number): void {
+    const shouldUpdate = shouldUpdateWidgetHeightAutomatically(
+      height,
+      this.props,
+    );
+    const { updateWidgetAutoHeight } = {
```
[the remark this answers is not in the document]
riodeuno: This is actually an issue due to lack of implementation. This function is supposed to come from the context. This context is available in another PR which when merged into release, will enable us to do something like `const {updateWidgetAutoHeight} = this.context`

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -175,6 +188,35 @@ abstract class BaseWidget<
...
+      updateWidgetAutoHeight: (widgetId: string, height: number) => {
+        log.debug("Updating widget height:", widgetId, height);
+      },
+    };
+    if (updateWidgetAutoHeight) {
```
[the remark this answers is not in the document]
riodeuno: Typescript makes this necessary -- as noted in the previous comment, when the context is available, the context is possibly undefined, leading us to have this check.

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -328,6 +370,67 @@ abstract class BaseWidget<
     );
   }
 
+  addAutoHeightOverlay(content: ReactNode, style?: CSSProperties) {
+    const onBatchUpdate = (height: number, propertiesToUpdate?: string[]) => {
+      if (propertiesToUpdate === undefined) {
+        propertiesToUpdate = ["minDynamicHeight", "maxDynamicHeight"];
+      }
```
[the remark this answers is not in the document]
riodeuno: Sounds good. I'll use this. Thanks for the suggestion.

app/client/src/widgets/BaseWidget.tsx
riodeuno: Thanks for pointing this out, the actual code should be `return this.addAutoHeightContainer(content);` instead of `return this.addAutoHeightContainer(this.getPageView());`

The processing is common for both scenarios. Regardless of the render mode, we need to make sure to wrap the autoheightoverlay and the error boundary. 

I'll make this change.

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -328,6 +370,67 @@ abstract class BaseWidget<
...
+      const modifyObj: Record<string, unknown> = {};
+      propertiesToUpdate.forEach((propertyName) => {
+        modifyObj[propertyName] = Math.ceil(
+          height / GridDefaults.DEFAULT_GRID_ROW_HEIGHT,
+        );
+      });
+      this.batchUpdateWidgetProperty({
+        modify: modifyObj,
+        // postUpdateAction: ReduxActionTypes.CHECK_CONTAINERS_FOR_DYNAMIC_HEIGHT,
```
rahulramesha, line 386: Will this be enabled in the other PRs?

riodeuno: @rahulramesha Yes. This is related to the postUpdateActions PR created.

app/client/src/widgets/BaseWidget.tsx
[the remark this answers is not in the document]
riodeuno: Done.

app/client/src/widgets/BaseWidget.tsx
riodeuno: Done.

marks0351 (approved):

PR #17995, opened 2022-10-31
chore: Remove example workspace creation in migration since its no longer used

trishaanand:


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog.java
```diff
@@ -765,253 +755,257 @@ public void giveInvitePermissionToOrganizationsAndPublicPermissionsToApplication
         }
     }
 
+        // Examples organization is no longer getting used. Commenting out the migrations which add/update the same.
     @SuppressWarnings({"unchecked", "rawtypes"})
```
mohanarpit, line 759:
```suggestion
    // @SuppressWarnings({"unchecked", "rawtypes"})
```

mohanarpit (approved):

PR #18001, opened 2022-10-31
fix: Number validation passthrough behaviour on 0

### LINKED ISSUE #17472: [Bug]: ValidationTypes.Number returning 'min' value (when min value validation is set) instead of the parsed value

### Description

When the 'min' value validation is set, eg: 
`validation: {
              type: ValidationTypes.NUMBER,
              params: { min: 1 },
            }`
The expected value is supposed to be the parsed value.
However, what is being returned is the minimum value.

### Steps To Reproduce

1. Open the 'index.tsx' file for any widget having a validation of the type 'Number' for any of its properties. eg: 'maxNumFiles' property for the 'FilePicker' widget.
2. Add a validation to set its minimum value. eg:
`validation: {
              type: ValidationTypes.NUMBER,
              params: { min: 1},
            }`
3. Enter 0 for the 'Max No. of files' property in the canvas for the widget.
4.  Debug to find the value returned. eg: in our case, 'this.props.maxNumFiles' returns the min value ie 1, instead of the evaluated value ie 0.

### Public Sample App

_No response_

### Version

Cloud - 1.8.5

aswathkk:
## Description

Fixes the behaviour where the Number with min and max validation doesn't passthrough the value when the value is zero.

A side effect is fixed as per the comment: https://github.com/appsmithorg/appsmith/issues/17472#issuecomment-1281818238

Fixes #17472 

## Type of change

## How Has This Been Tested?

- Jest

## Checklist:


app/client/src/workers/validations.ts
```diff
@@ -572,7 +572,13 @@ export const VALIDATORS: Record<ValidationTypes, Validator> = {
       if (parsed < Number(config.params.min)) {
         return {
           isValid: false,
-          parsed: parsed || config.params.min || 0,
+          parsed:
+            // passThroughOnZero is introduced to resolve a bug and to not break existing apps
+            // Refer: https://github.com/appsmithorg/appsmith/issues/17472#issuecomment-1281818238
+            config.params.passThroughOnZero !== undefined &&
+            config.params.passThroughOnZero === false
```
sbalaji1192, line 579: `config.params.passThroughOnZero === false` this should be fine right?

aswathkk: @sbalaji1192 I didn't get you

sbalaji1192: I meant we can refactor this condition config.params.passThroughOnZero !== undefined && config.params.passThroughOnZero === false to just config.params.passThroughOnZero === false.

app/client/src/constants/PropertyControlConstants.tsx
```diff
@@ -121,6 +121,7 @@ type ValidationConfigParams = {
   ignoreCase?: boolean; //to ignore the case of key
   type?: ValidationTypes; // Used for ValidationType.TABLE_PROPERTY to define sub type
   params?: ValidationConfigParams; // Used for ValidationType.TABLE_PROPERTY to define sub type params
+  passThroughOnZero?: boolean; // Used for ValidationType.NUMBER to allow 0 to be passed through
```
sbalaji1192, line 124: Can you also mention that default value is true

app/client/src/workers/validations.ts
aswathkk: Right. 💯

sbalaji1192 (approved):

PR #18018, opened 2022-11-01
fix: JS Objects save failures due to AST changes

### LINKED ISSUE #17619: [Bug]: Autosave failing for large JSObjects

### Description

To quote the user
"I have a massive issue here, I was editing JSObject in my app and while I was editing I constantly meet performance issues on the editor. The autosave feature keeps having issues and it takes so long to save. Sometimes it outright crashes or resets my codes to the previous save state. I kept using it anyway but a few moments later this happens: The entire server went down with it.
Please let me know what I can do to avoid it, I think we had the exact same issue before and we couldn't locate the issue, Our server went down multiple times before."

Server being used : AWS EC2 t3a.medium

More context can be found here, along with the logs : https://theappsmith.slack.com/archives/C0385FH70E8/p1665567331103289 

### Steps To Reproduce

1. Go to JSObject editor and try to edit a large JSObject
2. Make quick updates to query references and the autosave slows down.

### Public Sample App
App JSON : https://theappsmith.slack.com/archives/C0385FH70E8/p1665637419736139?thread_ts=1665567331.103289&cid=C0385FH70E8

_No response_

### Version

1.8.3

Front logo [Front conversations](https://app.frontapp.com/open/top_5bg98)

subrata71:
This PR is to deal with the following items
- Connections quickly get exhausted when AST single-script-data endpoint is invoked several times.
- Users face internal server errors when there's no connection available to call the AST API
- The multiple-script-data endpoint is supposed to take multiple bindings in one single request and to serve the need for multiple connections through one connection where applicable
- Now RTS server is accessed through 127.0.0.1 instead of localhost to avoid the domain name resolution before every API call ([Related SO post](https://stackoverflow.com/questions/7382602/what-is-the-difference-between-127-0-0-1-and-localhost))
- Max connections are increased to 100 which we can reduce later when we have a permanent solution to this problem
- Earlier for each action inside a JS collection the updateLayout would've been called. As a result, the AST API was called more than 38K times just for changing one character in the JS editor for a moderately complex app
- We have ensured one `updateLayout` call for a JS collection in this API and that reduces the number of API calls from 38k to 1.5k (Approximately)
- This also helps improve the response time of the `update JS collection API`
- For the binding values which start with `appsmith.theme` will be served directly from the application logic without calling the API

#### Disclaimer
The PR is not supposed to solve the problem entirely i.e. it will not prevent it from ever happening again. But it will increase the chances of not getting `internal server error` while editing large JS objects.

## Description

[RCA and possible solution on Notion](https://www.notion.so/appsmith/RCA-Internal-Server-Error-while-Editing-Large-JS-object-88017c9245dd4a6a9f9237eddec6f82c)

Fixes #17619 

## Type of change

## How Has This Been Tested?
- Manual
- JUnit

### Dev activity

### QA activity:


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AstServiceCEImpl.java
```diff
@@ -34,44 +34,53 @@ public class AstServiceCEImpl implements AstServiceCE {
...
+            return Flux.empty();
         }
 
         // If RTS server is not accessible for this instance, it means that this is a slim container set up
         // Proceed with assuming that all words need to be processed as possible entity references
         if (Boolean.FALSE.equals(instanceConfig.getIsRtsAccessible())) {
-            return Mono.just(new HashSet<>(MustacheHelper.getPossibleParentsOld(bindingValue)));
-        }
+            return Mono.just(bindingValues)
+                    .flatMapMany(Flux::fromIterable)
```
sondermanish, line 57: Why not start with Flux.fromIterable over here?

subrata71: Done

sondermanish (approved):

sondermanish (approved):

subrata71:
There are several factors/issues which are hindering the potential benefit that we could gain over this PR

- For a moderately complex app it requires thousands of connections to be available for consuming AST APIs. The binding values are quite distinct and there are only a few cases where multiple binding values are multiplexed into one single request (/ast/multiple-script-data). Thus we couldn't exploit the multiple-script-data endpoint.
- The AST API is consumed more by the appsmith-defined binding values e.g. {appsmith.theme.colors.primaryColor, appsmith.theme.boxShadow.appBoxShadow etc}. Can we work on that particular area so that AST API is consumed to get possible entity references for the user-defined binding values or can we introduce caching in the webclient?
- The debounce time is currently configured as `100ms` in the front-end. For every small change the update JS collection API is triggered and that requires a lot of connections to be available for consuming the AST API. Can we increase the debounce time to something larger or can we introduce a button for saving the JS object explicitly by the user?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/LayoutActionServiceCE.java
```diff
@@ -24,6 +24,8 @@ public interface LayoutActionServiceCE {
 
     Mono<ActionDTO> updateSingleAction(String id, ActionDTO action);
 
+    Mono<String> updatePageLayoutsGivenAction(String pageId);
```
nidhi-nair, line 27: This seems to be a misnomer, and looks like it should belong in another service altogether

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/LayoutActionServiceCEImpl.java
```diff
@@ -119,7 +120,8 @@ public Mono<ActionDTO> updateAction(String id, ActionDTO action) {
                     .flatMap(action1 -> {
                         log.debug("Action {} has been removed from its collection.", action1.getId());
                         action.setCollectionId(null);
-                        return this.updateSingleAction(id, action);
+                        return this.updateSingleAction(id, action)
```
nidhi-nair, line 123: Can we have two versions of this method, one that performs the update layout as well, and one that only does the update action? This is so that we are able to knowingly use one of the two all the time. It would also require lesser refactor in this PR.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/LayoutCollectionServiceCEImpl.java
```diff
@@ -208,6 +208,7 @@ public Mono<ActionCollectionDTO> createCollection(ActionCollectionDTO collection
                             })
                             .flatMapMany(Flux::fromIterable)
                             .flatMap(action -> layoutActionService.updateSingleAction(action.getId(), action))
+                            .flatMap(updatedAction -> layoutActionService.updatePageLayoutsGivenAction(updatedAction.getPageId()).thenReturn(updatedAction))
```
nidhi-nair, line 211: We seem to be calling update layout for each action here, why is that?

nidhi-nair (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/LayoutCollectionServiceCEImpl.java
```diff
@@ -587,6 +588,7 @@ public Mono<ActionCollectionDTO> updateUnpublishedActionCollection(String id,
                             });
                 })
                 .flatMap(actionCollection -> actionCollectionService.update(actionCollection.getId(), actionCollection))
+                .flatMap(savedActionCollection -> layoutActionService.updatePageLayoutsGivenAction(savedActionCollection.getPublishedCollection().getPageId()).thenReturn(savedActionCollection))
```
sondermanish, line 591: Since, the method updatePageLayoutsGivenAction require only pageId to work, could we refactor the method name?

sondermanish (dismissed):
The flow looks good to me, and since the updateSingleAction has been made to follow single responsibility principle, Here are some questions to ponder about:

-  Instead of linking this `updateSingleAction(id, action).flatMap(updatedAction -> updatePageLayoutsGivenAction(updatedAction.getPageId()).thenReturn(updatedAction))` snippet in all the places , Can create a method which aggregates these two together?

- Let's refactor the name of `updatePageLayoutGivenAction`  to `updatePageLayoutActionWithPageId` ?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AstServiceCEImpl.java
```diff
@@ -34,44 +36,58 @@ public class AstServiceCEImpl implements AstServiceCE {
     private final InstanceConfig instanceConfig;
 
     private final WebClient webClient = WebClientUtils.create(ConnectionProvider.builder("rts-provider")
-            .maxConnections(500)
-            .maxIdleTime(Duration.ofSeconds(5))
-            .maxLifeTime(Duration.ofSeconds(10))
-            .pendingAcquireTimeout(Duration.ofSeconds(5))
+            .maxConnections(100)
```
riteshkew, line 39: how did we finalise upon these numbers?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AstServiceCEImpl.java
```diff
@@ -34,44 +36,58 @@ public class AstServiceCEImpl implements AstServiceCE {
...
-            return Mono.empty();
+    public Flux<Tuple2<String, Set<String>>> getPossibleReferencesFromDynamicBinding(List<String> bindingValues, int evalVersion) {
+        if (bindingValues == null || bindingValues.size() == 0) {
+            return Flux.empty();
+        }
+        /*
+            For the binding value which starts with "appsmith.theme" can be directly served
+            without calling the AST API or the calling the method for non-AST implementation
+         */
+        if (bindingValues.size() == 1 && bindingValues.get(0).startsWith("appsmith.theme.")) {
```
riteshkew, line 57: can we consider this for other appsmith related binding values which are usually static, is it always appsmith.theme

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AstServiceCEImpl.java
```diff
@@ -34,44 +36,58 @@ public class AstServiceCEImpl implements AstServiceCE {
...
                 .post()
-                .uri(commonConfig.getRtsBaseDomain() + "/rts-api/v1/ast/single-script-data")
+                .uri(commonConfig.getRtsBaseDomain() + "/rts-api/v1/ast/multiple-script-data")
                 .contentType(MediaType.APPLICATION_JSON)
-                .body(BodyInserters.fromValue(new GetIdentifiersRequest(bindingValue, evalVersion)))
+                .body(BodyInserters.fromValue(new GetIdentifiersRequestBulk(bindingValues, evalVersion)))
                 .retrieve()
-                .bodyToMono(GetIdentifiersResponse.class)
+                .bodyToMono(GetIdentifiersResponseBulk.class)
+                .retryWhen(Retry.max(3))
```
riteshkew, line 78: what does this retry solve for

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AstServiceCEImpl.java
subrata71: We have reached this number through trial and error. We have noticed that with this number the latency of the API is less and the time spent on closing these connections is also less compared to the time when the number of connections is higher.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AstServiceCEImpl.java
subrata71: We will keep enriching this list in order to reduce the number of API calls. But for now, I have found the following binding parameters appear most of the time

```
appsmith.theme.colors.primaryColor
appsmith.theme.borderRadius.appBorderRadius
appsmith.theme.boxShadow.appBoxShadow

```
And these are appsmith-defined. There are a few other bindings that are not appsmith-defined but start with appsmith e.g. `appsmith.store.*`.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AstServiceCEImpl.java
subrata71: Let's say one UPDATE JS collection API call requires 600 AST API calls. If one of the 600 API calls fails the user will see the internal server error. Adding retry (3) ensures that one failed AST API call will be retried 3 times by the `Webclient` itself. Thus the chance of encountering internal server error is reduced.

riteshkew (approved):
The current Changes look good and the next set of changes for this would be covered in another PR.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/LayoutActionServiceCE.java
subrata71: This function needs to be called explicitly from that JS collection update part once all the actions are updated. Hence, keeping this as a separate method inside that interface.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/LayoutActionServiceCEImpl.java
subrata71: I have already done the refactor from the places where it's called. Besides if we introduce two functions one having the update page layout logic and another only having the update action logic, we require one more function to update the page layouts separately. Hence, we will end up introducing three functions.

btsgh:
Tested this fix  by trying the following scenarios - 

1. Cut / copy paste a code within a JS Object to a different area- works fine
2. Git connected apps JS Object editing - works fine
3. Import and export flow of git connected apps - works fine
4. Editor - adding lines in the js editor - works fine
5. Create  New JS object - works fine. (code from highlighted jsobject is copied, but this is being handled as part of different bug)
6.  Delete JS object  - works fine
7. Copy Js object  to same or different page - works fine
8.  Move JS object  to a different page - works fine
The internal server error was no longer seen.
The JS Object editor works fine, without hanging or crashing in any of the above scenarios.

PR #18039, opened 2022-11-01
feat: widget features test

ankurrsinghal:
Added unit tests for:-
- [x] WidgetFeatures/validateMinHeight
- [x] WidgetFeatures/validateMaxHeight
- [x] WidgetFeatures/findAndUpdatePropertyPaneControlConfig

### DIFF
--- a/app/client/src/utils/WidgetFeatures.ts
+++ b/app/client/src/utils/WidgetFeatures.ts
@@ -74,7 +74,7 @@ export function findAndUpdatePropertyPaneControlConfig(
         );
 
         if (
-          controlConfigIndex &&
+          controlConfigIndex !== undefined &&
           controlConfigIndex > -1 &&
           sectionConfig.children
         ) {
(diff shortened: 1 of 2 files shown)


app/client/src/utils/WidgetFeatures.ts
```diff
@@ -74,7 +74,7 @@ function findAndUpdatePropertyPaneControlConfig(
         );
 
         if (
-          controlConfigIndex &&
+          controlConfigIndex !== undefined &&
```
ankurrsinghal, line 77: @riodeuno Here if suppose we want to update any property whose index in the array is 0, it will not enter this if block, so updated the condition for undefined check explicit.

riodeuno: Thank you @ankurrsinghal ! Nice catch.

riodeuno (approved):

PR #18048, opened 2022-11-02
feat: update property name automatically when custom column name is changed

### LINKED ISSUE #17142: [Feature]: When user updates the name of the custom column also update the property name

### Summary

Since user is updating name of custom column, like other table columns they should not need to update the name again in the `property name` field.

In this case though, everytime user updates the name of the custom coloumn we will update the property name field, so their changes might go away and we have to trigger a migration.

The scenario where a user updates the property name and then updates the display name is rare

Today the users are not able to understand why the name of the custom column changed is not visible directly on the selectedRow binding. We had 2 internal users raise this which means the UX is not obvious.

souma-ghosh:
## Description

This feature is only for custom columns. Users needed to change property name of custom columns every-time they changed the column name.
Column name can be updated from 2 places
1. Table widget property pane > Columns
2. Column settings > title edit

This PR uses `updateHook` to change column `alias` of a custom column whenever the column `label` (column name) is changed.

Fixes #17142 

## Type of change

## How Has This Been Tested?

- Cypress tests
- Jest tests

### Dev activity

### QA activity:


app/client/src/widgets/TableWidgetV2/widget/propertyUtils.ts
```diff
@@ -624,3 +624,27 @@ export const replacePropertyName = (
   path.pop();
   return `${path.join(".")}.${targetPropertyName}`;
 };
+
+export const updateAliasOnLabelChange = (
+  props: TableWidgetProps,
+  propertyPath: string,
+  propertyValue: any,
```
sbalaji1192, line 631: use unknown

app/client/src/widgets/TableWidgetV2/widget/propertyUtils.ts
```diff
@@ -624,3 +624,27 @@ export const replacePropertyName = (
...
+): Array<PropertyHookUpdates> | undefined => {
+  // alias will be updated along with label change only for custom columns
```
sbalaji1192, line 633: Could you add jest case as well ?

app/client/src/widgets/TableWidgetV2/widget/propertyConfig/PanelConfig/Data.ts
```diff
@@ -93,6 +93,8 @@ export default {
       propertyName: "alias",
       label: "Property Name",
       controlType: "INPUT_TEXT",
+      helperText: () =>
+        "The property name will update automatically when custom column name is changed.",
```
sbalaji1192, line 97: Can you talk with Dilip and get a proper help text .This one looks like hint

app/client/src/widgets/TableWidgetV2/widget/propertyConfig/PanelConfig/index.ts
```diff
@@ -39,4 +40,5 @@ export default {
     discardButtonStyleConfig,
     BorderAndShadow,
   ],
+  updateHook: updateAliasOnLabelChange,
```
sbalaji1192, line 43: Let's rename this function to updateCustomColumnAliasOnLabelChange

app/client/src/widgets/TableWidgetV2/widget/propertyUtils.ts
```diff
@@ -624,3 +624,27 @@ export const replacePropertyName = (
...
   return `${path.join(".")}.${targetPropertyName}`;
 };
+
+export const updateAliasOnLabelChange = (
+  props: TableWidgetProps,
+  propertyPath: string,
+  propertyValue: unknown,
+): Array<PropertyHookUpdates> | undefined => {
+  // alias will be updated along with label change only for custom columns
+  const regex = /^primaryColumns\.(customColumn\d+)\.(.*)$/;
```
sbalaji1192, line 634: can be replaced to `/^primaryColumns\.(customColumn\d+)\.label$/`

app/client/src/widgets/TableWidgetV2/widget/propertyUtils.ts
```diff
@@ -624,3 +624,27 @@ export const replacePropertyName = (
...
+  const propertyToUpdate = [];
+  if (
+    propertyPath?.length &&
+    regex.test(propertyPath) &&
+    propertyPath.split(".").pop() === "label"
+  ) {
+    const propertyPathArr = propertyPath.split(".");
+    propertyPathArr.pop();
+    propertyPathArr.push("alias");
+    propertyToUpdate.push({
```
sbalaji1192, line 644: can replace this entire block with
<pre>
    propertyToUpdate.push({
      propertyPath: propertyPath.replace("label", "alias"),
      propertyValue,
    });
</pre>

app/client/src/widgets/TableWidgetV2/widget/propertyUtils.test.ts
```diff
@@ -454,3 +455,50 @@ describe("uniqueColumnAliasValidation", () => {
...
+  });
+
+  it("should generate proper array for propertyToUpdate for any custom column", () => {
+    expect(
+      updateAliasOnLabelChange(
+        {} as TableWidgetProps,
+        "primaryColumns.customColumn12345.label",
+        "customColumn12",
+      ),
+    ).toEqual([]);
```
sbalaji1192, line 502: This should have returned the the update array right?

app/client/src/widgets/TableWidgetV2/widget/propertyUtils.test.ts
```diff
@@ -454,3 +455,50 @@ describe("uniqueColumnAliasValidation", () => {
     });
   });
 });
+
+describe("updateAliasOnLabelChange", () => {
+  it("should generate correct array for propertyToUpdate", () => {
```
sbalaji1192, line 460: should return the propertyToUpdate array to update alias for the given custom column

app/client/src/widgets/TableWidgetV2/widget/propertyUtils.test.ts
```diff
@@ -454,3 +455,50 @@ describe("uniqueColumnAliasValidation", () => {
...
+      ),
+    ).toEqual([
+      {
+        propertyPath: "primaryColumns.customColumn1.alias",
+        propertyValue: "customColumn12",
+      },
+    ]);
+  });
+
+  it("should generate empty array for propertyToUpdate when the column is not a custom column", () => {
```
sbalaji1192, line 475: should not return propertyToUpdate array to update alias for the given column

app/client/src/widgets/TableWidgetV2/widget/propertyUtils.test.ts
```diff
@@ -454,3 +455,50 @@ describe("uniqueColumnAliasValidation", () => {
...
+    expect(
+      updateAliasOnLabelChange(
+        {} as TableWidgetProps,
+        "primaryColumns.resume_url.label",
+        "customColumn12",
+      ),
+    ).toEqual([]);
+  });
+
+  it("should generate empty array for propertyToUpdate when label property is not being changed", () => {
```
sbalaji1192, line 485: should not return the propertyToUpdate array to update alias when any property other than label property of the custom column gets changed.

app/client/src/widgets/TableWidgetV2/widget/propertyConfig/PanelConfig/Data.ts
```diff
@@ -93,6 +93,8 @@ export default {
       propertyName: "alias",
       label: "Property Name",
       controlType: "INPUT_TEXT",
+      helperText: () =>
+        "Changing the name of the column overrides any changes to this field",
```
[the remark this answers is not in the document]
souma-ghosh: Yes have checked with him. This is the correct text

app/client/src/widgets/TableWidgetV2/widget/propertyUtils.ts
```diff
@@ -624,3 +624,20 @@ export const replacePropertyName = (
...
+  // alias will be updated along with label change only for custom columns
+  const regex = /^primaryColumns\.(customColumn\d+)\.label$/;
+  const propertyToUpdate = [];
+  if (propertyPath?.length && regex.test(propertyPath)) {
+    propertyToUpdate.push({
+      propertyPath: propertyPath.replace("label", "alias"),
+      propertyValue,
+    });
+  }
+  return propertyToUpdate;
```
sbalaji1192, line 642: we should send undefined instead of an empty array when the condition is not met. otherwise, property Control will throw an error at this [line](https://github.com/appsmithorg/appsmith/blob/release/app/client/src/pages/Editor/PropertyPane/PropertyControl.tsx#L296)

sbalaji1192 (approved):

souma-ghosh:
@dilippitchika About updating other widget bindings when we change the name of a custom column; we should create a separate issue to work on that.
As far as I understand the logic here has to be very much similar to what is happening to bindings when we update the name of a widget. e.g. when we update the name of a table from `Table1` to `Table12` all the dependencies that had `Table1` in them get updated to `Table12`.
But we will need careful planning before going about this. I will create a separate issue and tag it here soon

chandannkumar:
Test Plan for this PR: https://github.com/appsmithorg/TestSmith/issues/2095

chandannkumar:
Tested this PR and LGTM
- Tested for updating custom column names
- Tested the same on container type widgets, modal
- Tested the same for all bindings including widgets, Query
- Tested on older apps

chandannkumar:
Observation: @souma-ghosh 
- Unable to retain same name as before of custom column 
https://bthrujcsw8.vmaker.com/record/fLK5yGK4KrUBDsCU

- On making duplicate custom column names, the populated data in Table is of the last column instead of the 1st custom column. What should be the correct behavior here?
https://bthrujcsw8.vmaker.com/record/g7AtuERxBkfKe3rQ

souma-ghosh:
> Observation: @souma-ghosh
> 
> * Unable to retain same name as before of custom column
>   https://bthrujcsw8.vmaker.com/record/fLK5yGK4KrUBDsCU
> * On making duplicate custom column names, the populated data in Table is of the last column instead of the 1st custom column. What should be the correct behavior here?
>   https://bthrujcsw8.vmaker.com/record/g7AtuERxBkfKe3rQ

1. For the first one, the issue already existed before and is not as a result of this PR. We should create an issue for this.
2. For the second one, @dilippitchika could you comment on the expected behaviour for this scenario. According to my understanding since there is a duplicate column label the table should not be able to identify which column to pick up and show in the table or in `TableName.selectedRow`. We are maintaining separate ids for each column but we are using column `label` to identify and make the columns unique. We show an error when non unique columns are encountered. Should this behaviour be sufficient?

dilippitchika:
For 1, I didn't understand how this is an existing issue @souma-ghosh. What are the steps to reproduce?

For 2, It's expected behavior to show the key which matches at the end. So not an issue.

souma-ghosh:
> For 1, I didn't understand how this is an existing issue @souma-ghosh. What are the steps to reproduce?
> 
> For 2, It's expected behavior to show the key which matches at the end. So not an issue.

The issue mentioned in point 1 happens for any column
Steps to reproduce 1
- Go into the column settings of any column
- Click on column title to edit it
- Delete the last character and type back the deleted letter again, e.g. backspace on `status` so it becomes `statu` then type back the `s` again (don't save the column title anywhere in between this step)
- Now save the title and come outside to the main property pane and observe that the column name is still `statu` and not `status`

souma-ghosh:
@chandannkumar Both the above mentioned issues are already happening in prod. We can move ahead

PR #18111, opened 2022-11-04
fix: Auto Height Fixes

riodeuno:
- [ ]  Fix JSONForm height discrepancy
- [ ] Fix issue where widgets moved below the other
- [ ] Fix dropTarget height after parent container resize


riodeuno:
@rahulramesha @marks0351 So, I finally figured out why we had all those special handling for height in multiple spots of the application code. The reason was the canvas height after resizing a container like widget. The original logic caused an eval for the min height update, which I'd like to avoid. Granted, it was batched and went along with other updates that came after resizing.

PR #18121, opened 2022-11-06
feat: Add saga to compute the auto height layout tree

riodeuno:
## Description
- [ ] Add saga to generate layout tree for computing auto height
- [ ] Add selector to get occupied spaces for above computations


app/client/src/selectors/editorSelectors.tsx
```diff
@@ -607,6 +607,85 @@ export const showCanvasTopSectionSelector = createSelector(
...
+    if (canvasWidgets) {
+      // Iterate through the list of canvas widgets
+      canvasWidgets.forEach((canvasWidget: FlattenedWidgetProps) => {
+        // Set the canvas widget id
+        const canvasWidgetId = canvasWidget.widgetId;
+
+        // Get the nesting level of this Canvas:
+        let parentId = canvasWidget.parentId;
+        let level = 0;
+        while (parentId) {
```
rahulramesha, line 648: There seem to be multiple nested loops, Can't we just use the DSL structure and recursively traverse the tree. It would be easier to calculate levels and we go over each node only once throughout

riodeuno: @rahulramesha Thanks for the pointer. I looked into the code. From a perspective of code clarity, we might want to have a separate function which does this using the DSL. When using the DSL, we have to either loop through each of the widget's children or do what we're doing here, where we pick a canvas widget and find how many canvas widget ancestors it has. 

When looping through children, we will have to filter out the non-canvas widgets, and ignore widgets like the Container widgets which have children but are not to be considered. Also, do this for all children canvases of a canvas widget.

What we're doing here, is going to be the cleaner logic for this computation. 

Although, we can improve the logic, by checking the map and stopping the upward walk earlier. I've added this change.

PR #18238, opened 2022-11-17
fix: Corrected usage of old soft deleted filter in BaseAppsmithRepository

### LINKED ISSUE #18161: [Task]: Check usage of deleted param for soft delete action in the entire codebase

Recently we found there were issues happening because of the deprecated `deleted` param in entities for soft-delete operation. In the latest version, we only use `deletedAt` field for soft-delete. This task is to verify we are only using `deletedAt` in all places in the code and `delete` field is not creating issues across.

### SubTasks
- Verify on FE
- Verify on BE

vishnu-gp:
## Description

Recently we found there were issues happening because of the deprecated `deleted` param in entities for soft-delete operation. In the latest version, we only use `deletedAt` field for soft-delete. This PR correct the usage of old `deleted` field only filter and adds the current field to the filter.

Fixes #18161

## Type of change

## How Has This Been Tested?

- Manual
- JUnit

### Dev activity

### QA activity:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/BaseAppsmithRepositoryCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/BaseAppsmithRepositoryCEImpl.java
@@ -95,9 +95,17 @@ public abstract class BaseAppsmithRepositoryCEImpl<T extends BaseDomain> {
     }
 
     public static final Criteria notDeleted() {
-        return new Criteria().orOperator(
-                where(fieldName(QBaseDomain.baseDomain.deleted)).exists(false),
-                where(fieldName(QBaseDomain.baseDomain.deleted)).is(false)
+        return new Criteria().andOperator(
+                //Older check for deleted
+                new Criteria().orOperator(
+                        where(FieldName.DELETED).exists(false),
+                        where(FieldName.DELETED).is(false)
+                ),
+                //New check for deleted
+                new Criteria().orOperator(
+                        where(FieldName.DELETED_AT).exists(false),
+                        where(FieldName.DELETED_AT).is(null)
+                )
         );
     }
 
--- a/app/server/appsmith-server/src/test/java/com/appsmith/server/services/ActionCollectionServiceTest.java
+++ b/app/server/appsmith-server/src/test/java/com/appsmith/server/services/ActionCollectionServiceTest.java
@@ -24,6 +24,7 @@ import com.appsmith.server.dtos.RefactorActionNameDTO;
 import com.appsmith.server.dtos.WorkspacePluginStatus;
 import com.appsmith.server.helpers.MockPluginExecutor;
 import com.appsmith.server.helpers.PluginExecutorHelper;
+import com.appsmith.server.repositories.ActionCollectionRepository;
 import com.appsmith.server.repositories.PermissionGroupRepository;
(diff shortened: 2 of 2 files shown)


app/server/appsmith-server/src/test/java/com/appsmith/server/services/ActionCollectionServiceTest.java
```diff
@@ -217,6 +222,42 @@ public void testCreateActionCollection() {
...
+        actionCollectionDTO.setPageId(pageId);
+        actionCollectionDTO.setPluginId(datasource.getPluginId());
+        actionCollectionDTO.setPluginType(PluginType.JS);
+        actionCollectionDTO.setDeletedAt(Instant.now());
+        layoutCollectionService.createCollection(actionCollectionDTO).block();
+        ActionCollection createdActionCollection = actionCollectionRepository.findByApplicationId(createdApplication.getId(), READ_ACTIONS, null).blockFirst();
+        createdActionCollection.setDeletedAt(Instant.now());
+        actionCollectionRepository.save(createdActionCollection).block();
+
+        StepVerifier.create(actionCollectionRepository.findByApplicationId(createdApplication.getId(), READ_ACTIONS, null))
```
sumitsum, line 256: Sorry, I could be understanding this wrong, but shouldn't we be checking here if `findByApplicationId` has not returned anything ?

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/BaseAppsmithRepositoryCEImpl.java
```diff
@@ -95,9 +95,17 @@ public static final String fieldName(Path path) {
...
+                new Criteria().orOperator(
+                        where(FieldName.DELETED).exists(false),
+                        where(FieldName.DELETED).is(false)
+                ),
+                //New check for deleted
+                new Criteria().orOperator(
+                        where(FieldName.DELETED_AT).exists(false),
+                        where(FieldName.DELETED_AT).is(null)
+                )
         );
```
abhvsn, line 109: @vishnu-gp as we have deprecated the use of deleted flag does it make sense to completely remove the dependency and only use deletedAt field instead?

vishnu-gp: @abhvsn Did you mean removing the filter from here or removing the entire deprecated field from all entities? The current setup is even if this field is deprecated we default it to false and on deletes update this field as well in addition to deletedAt. I would like to start a thread and hear everyone's opinion on this before making that change. You can see similar filters in `BaseRepositoryImpl` as well.

abhvsn: Fair enough, I will vote for completely removing the dependency on deleted flag, rational being we are deprecationg that field but still using in some places which creates confusion 😕.
I am treating this as a tech debt and we should get rid of this, earlier the better!

vishnu-gp: @abhvsn Let's track this here https://github.com/appsmithorg/appsmith/issues/18248 and move forward with PR for the time being.

PR #18293, opened 2022-11-21
fix: Authenticated datasource passwordExists field added for plugins

### LINKED ISSUE #14783: [Bug]-[7568]:Edit auth API datasource, I don't know if client secret is saved or not

### Summary

When I edit an authenticated API datasource, I see the datasource form with all the fields filled, except the client secret. But now I don't know if the secret is actually saved or not, because the placeholder text is the same both in create and edit mode: `Client Secret`.

Screenshot 2022-06-24 at 08 01 31

When I edit a datasource, I should be able to distinguish the saved password/secret/etc fields from the ones where I have not entered anything

This simple fix should allay the user's anxiety and make the datasource editing experience better.

vaibh1297:
## Description

> Authenticated datasource passwordExists  field added for plugins

Fixes #14783

## Type of change

## How Has This Been Tested?

- Manual

### Dev activity

### QA activity:


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/DBAuth.java
```diff
@@ -30,5 +31,13 @@ public enum Type {
     @JsonProperty(access = JsonProperty.Access.WRITE_ONLY)
     String password;
 
+    @JsonProperty
+    public boolean isPasswordExists() {
```
sumitsum, line 35: Instead of a method, we need to have a field named `isPasswordExists` that should be set by the API server during the code flow when password is persisted in the DB. Also, we need to add this field in the following classes as well since they have password like fields as well:
```
OAuth2
BasicAuth
ApiKeyAuth
BearerTokenAuth
```

Just a thought - should we instead override the `setter` method of password field to update this new variable whenever the password field is set ?

sumitsum: Alternatively, we may choose not to persist the `isPasswordExists` field at all because we only need it once when sending back the data to the client and it is a derived value i.e. can be derived on the fly from the password field.

sumitsum: IMO, we should go with the second one if possible.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/AuthenticationDTO.java
```diff
@@ -48,6 +51,14 @@ public enum AuthenticationStatus {
     @JsonIgnore
     AuthenticationResponse authenticationResponse;
 
+    protected abstract String getSecret();
+
+    @JsonProperty
+    protected boolean secretExists() {
+        String secret = getSecret();
+        return StringUtils.hasLength(secret);
```
mohanarpit, line 59: Nit: Any reason we are saving the secret in a variable before checking its length? Strings are immutable and this just creates unnecessary garbage in the JVM. In this particular case, it would be simpler to just write `StringUtils.hasLength(getSecret())`

vaibh1297: done

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/AuthenticationDTO.java
```diff
@@ -48,6 +51,13 @@ public enum AuthenticationStatus {
     @JsonIgnore
     AuthenticationResponse authenticationResponse;
 
+    protected abstract String getSecret();
+
+    @JsonProperty
```
sumitsum, line 56: Can you please check once if this does not persist the value in DB ?

vaibh1297: yes, value is not persisted in db.

sumitsum (dismissed):

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/SecretExists.java
```diff
@@ -0,0 +1,19 @@
...
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+@Getter
+@Setter
+@NoArgsConstructor
+//@JsonInclude(JsonInclude.Include.NON_NULL)
```
sumitsum: Please remove the commented out line.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/SecretExists.java
```diff
@@ -0,0 +1,19 @@
...
+public class SecretExists {
```
sumitsum, line 11: Please add a comment explaining what this class does - ie. essentially what problem this class solves.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/SecretExists.java
vaibh1297: already removed, please check the updated version of pr.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/SecretExists.java
vaibh1297: done

sumitsum (dismissed):

app/server/appsmith-plugins/amazons3Plugin/src/main/resources/form.json
```diff
@@ -62,7 +62,8 @@
           "dataType": "PASSWORD",
           "initialValue": "",
           "isRequired": true,
-          "encrypted": true
+          "encrypted": true,
+          "valueExistPath": "datasourceConfiguration.authentication.secretExists.password"
```
sumitsum, line 66: Can you please rename this to `isSecretExistsPath` since it seems to better explain what the key stands for ? Please mention in case you have other opinions on this.

sumitsum (approved):

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/SecretExists.java
```diff
@@ -0,0 +1,22 @@
...
+import lombok.Setter;
+
+/**
+ * This class contains boolean variables to indicate
+ * whether an authenticated datasource has secret or not
+*/
+@Getter
+@Setter
+@NoArgsConstructor
+public class SecretExists {
```
sumitsum, line 15: Can you please rename this to `SecretExistsIndicator` ?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ApiKeyAuth.java
```diff
@@ -34,4 +36,9 @@ public enum Type {
     @JsonProperty(access = JsonProperty.Access.WRITE_ONLY)
     @Encrypted
     String value;
+
+    @Override
+    protected void buildSecretExists(SecretExists secretExists) {
```
sumitsum, line 41: Can you please rename this to `configureSecretExistsIndicator` ?

sumitsum (changes requested):

sumitsum (dismissed):

PR #18298, opened 2022-11-21
fix: Label position gets deselected on selecting already selected option

aswathkk:
## Description

Fixes the issue pointed out in [this](https://www.loom.com/share/b02a1ed57b4342a7bace95462cc7edab) Loom recording where the label position gets deselected when an already selected item is selected.
This was happening since these controls didn't have a `defaultValue` associated to it.


yaldram (approved):

kamakshibhat-appsmith:
Working fine

aswathkk:
I'm working on adding migration so that the existing applications won't break on enabling dynamic height.

app/client/src/utils/DSLMigrations.ts
```diff
@@ -1117,6 +1128,19 @@ export const transformDSL = (currentDSL: DSLWidget, newPage = false) => {
   return currentDSL;
 };
 
+function migrateLabelPosition(currentDSL: DSLWidget) {
+  return traverseDSLAndMigrate(currentDSL, (widget: WidgetProps) => {
+    if (
+      widget.type === "PHONE_INPUT_WIDGET" ||
+      widget.type === "CURRENCY_INPUT_WIDGET"
+    ) {
+      if (widget.labelPosition === undefined) {
```
sbalaji1192, line 1137: Let's club this if block with enclosing if block.

PR #18307, opened 2022-11-21
chore: Trying out a few update layout scenarios (WIP)

nidhi-nair:
WIP PR, please don't merge.

Setting up a clean way to write microbenchmarks, gots to work on the following:
- [ ] - Are these benchmarks reliable? Do we need to use a blackhole on reactive methods?
- [ ] - Make these benchmarks easier to create
- [ ] - Check and isolate these tests from normal build process
- [ ] - Customize how results get generated, make it easier to define inputs
- [ ] - How do we make sure a benchmark is in the green?


nidhi-nair:
Closing because the code here is stale and is not being actively worked on. Was related to JMH based perf testing.

PR #18321, opened 2022-11-21
fix: Using an html tag within a Text widget to access a web page, without the https:// appended causes application to crash

### LINKED ISSUE #18204: [Bug]: Using an html tag within a Text widget to access a web page, without the https:// appended causes application to crash

### Description

When I use an html tag within a Text widget and try to access google web page using only www.google.co.in, ... application crashes

If I use with https:// appended at the start, it works fine

Expected behavior - even without the https:// appended, it should work. 

### Steps To Reproduce

1. Drag and drop a text widget.
2. In the property pane, for the text to be displayed, use an html tag within a Text widget and try to access google web page using only www.google.co.in
3. Click on Google in the text widget.  The application crashes.  
4. Note that the url that the application tries to access is - 
https://release.app.appsmith.com/app/release-nov-16/textwidget-6374e05855c68b4cc232c47b/www.google.co.in

Expected behavior: Application should not crash. The url that the application tries to access should be modified to the right one. Https:// should be automatically appended to the url.
Note: If we directly enter the link like this - www.google.co.in - in the text widget, it works fine.

### Public Sample App

_No response_

### Issue video log

https://ddpa7mfuvv.vmaker.com/record/z4bOHTvbS1c8xM6d

### Version

Production

jsartisan:
## Description
On using anchor with href without `https` or `http`, It creates invalid URL which leads to 404 page. This PR fixes that invalid URL.

Fixes #18204 

## Type of change

- Bug fix

## How Has This Been Tested?
- Manual

### Dev activity

### QA activity:

### DIFF
--- a/app/client/src/widgets/TextWidget/component/filters/LinkFilter.ts
+++ b/app/client/src/widgets/TextWidget/component/filters/LinkFilter.ts
@@ -1,0 +1,22 @@
+import { Filter } from "interweave";
+import { addHttpIfMissing } from "../helpers";
+
+class LinkFilter extends Filter {
+  attribute(name: string, value: string): string {
+    if (name === "href") {
+      return addHttpIfMissing(value);
+    }
+
+    return value;
+  }
+
+  node(name: string, node: HTMLElement): HTMLElement {
+    if (name === "a") {
+      node.setAttribute("target", "_blank");
+    }
+
+    return node;
+  }
+}
+
+export default LinkFilter;
--- a/app/client/src/widgets/TextWidget/component/filters/LinkFilter.ts
+++ b/app/client/src/widgets/TextWidget/component/filters/LinkFilter.ts
@@ -9,14 +9,6 @@ class LinkFilter extends Filter {
 
     return value;
   }
-
-  node(name: string, node: HTMLElement): HTMLElement {
-    if (name === "a") {
-      node.setAttribute("target", "_blank");
-    }
-
-    return node;
-  }
 }
 
 export default LinkFilter;
--- a/app/client/src/widgets/TextWidget/component/helpers.tsx
+++ b/app/client/src/widgets/TextWidget/component/helpers.tsx
@@ -1,0 +1,15 @@
+/**
+ * add http if missing
+ *
+ * @param url
+ * @returns
+ */
+export const addHttpIfMissing = (url: string) => {
+  if (!url) {
+    return url;
+  }
+  if (url.indexOf("http") === 0) {
+    return url;
+  }
+  return `http://${url}`;
+};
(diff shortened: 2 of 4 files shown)


app/client/src/widgets/TextWidget/component/filters/LinkFilter.ts
```diff
@@ -0,0 +1,22 @@
...
+    if (name === "href") {
+      return addHttpIfMissing(value);
+    }
+
+    return value;
+  }
+
+  node(name: string, node: HTMLElement): HTMLElement {
+    if (name === "a") {
+      node.setAttribute("target", "_blank");
```
keyurparalkar, line 15: Won't this line always open the link in new tab? Since we are setting `target='_blank'`. Should this be dependent on user?

app/client/src/widgets/TextWidget/component/helpers.tsx
```diff
@@ -0,0 +1,15 @@
...
+ * @returns
+ */
+export const addHttpIfMissing = (url: string) => {
+  if (!url) {
+    return url;
+  }
+  if (url.indexOf("http") === 0) {
+    return url;
+  }
+  return `http://${url}`;
```
keyurparalkar, line 14: Why are we returning `http` instead of `https`?

app/client/src/widgets/TextWidget/component/helpers.tsx
```diff
@@ -0,0 +1,15 @@
+/**
+ * add http if missing
+ *
+ * @param url
+ * @returns
+ */
+export const addHttpIfMissing = (url: string) => {
+  if (!url) {
+    return url;
```
keyurparalkar, line 9: @jsartisan does this `if block` returns string if it's not an URL?

app/client/src/widgets/TextWidget/component/filters/LinkFilter.ts
jsartisan: Right. Removed the target _blank.

app/client/src/widgets/TextWidget/component/helpers.tsx
jsartisan: If the URL is undefined, null, or false, it will just return whatever it was. So now we don't have to do `return null` or `return false` etc.

app/client/src/widgets/TextWidget/component/helpers.tsx
jsartisan: The main reason is some sites may not have HTTPS and when they have HTTPS, they always put a redirect from HTTP to HTTPS. We can't know if they have HTTP or HTTPS, so we can just put HTTP for the unknown.

app/client/src/widgets/TextWidget/component/filters/LinkFilter.ts
keyurparalkar: Have we considered scenarios apart from `anchor` tag? like tags mentioned in this [link](https://www.w3schools.com/tags/att_href.asp#:~:text=For%20and%20%3Carea,relative%20URLs%20on%20a%20page.). @jsartisan are there any use cases that you found were user might be referring to a file like the `link`tag in the above URL.

keyurparalkar (approved):

PR #18362, opened 2022-11-22
fix: auto height zero issue

ankurrsinghal:
Fixed the issue of widget not collapsing when it's height is zero.


riodeuno:
@ankurrsinghal  Closing this, as it is fixed in another PR.

riodeuno:
Fixed in #18366

PR #18426, opened 2022-11-24
fix: permission driven context menu for pages

### LINKED ISSUE #18425: [Bug]: When user tries to add the widget in the page which has view access and then click on the meatball icon on the page the menu icon doesn't pop up
### Description
- Page context menu is not permission driven on first load. It only became permission driven only when the user switches to the specific page.

https://www.loom.com/share/1b9cf318925d4634aee7fce0460c5489

berzerkeer:
## Description

Right now after fetching current page, we also fetch all pages to extract page DSL , in the process the response also returns the user permissions for each page, but the problem is we are not storing this permissions in redux store along with the pages array. Due to that we can't determine what to show/hide in the context menu of each pages. So with this PR added a new `Redux Action` UPDATE_PAGE_LIST which updates the pages array with its respective user permissions. 

TL;DR The page context menu is now permission driven. Show/hide options based on permission.

Fixes #18425

Media
https://www.loom.com/share/2842940b03064f27982d949431ee6900

## Type of change

## How Has This Been Tested?
- Manual

### Dev activity


app/client/src/sagas/PageSagas.tsx
```diff
@@ -950,6 +950,7 @@ function* fetchPageDSLSaga(pageId: string) {
       return {
         pageId: pageId,
         dsl: extractCurrentDSL(fetchPageResponse),
+        userPermissions: fetchPageResponse.data.userPermissions,
```
ankitakinger, line 953: add optional chaining before `userPermissions` here safe side

berzerkeer: Oh sure.

ankitakinger (approved):
LGTM

PR #18432, opened 2022-11-24
chore: Add logs from segment error retry handler

AnaghHegde:


sidhantgoel (approved):
thumbs up

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SegmentConfig.java
```diff
@@ -56,11 +57,15 @@ public Analytics analyticsRunner() {
         final Analytics analytics = Analytics.builder(analyticsWriteKey).log(logProcessorWithErrorHandler).build();
         logProcessorWithErrorHandler.onError(logData -> {
             final Throwable error = logData.getError();
-            analyticsOnAnalytics.enqueue(TrackMessage.builder("segment_error")
+            // TODO remove this log statement once the issue with analytics is resolved
+            log.error(" UserId is null or empty inside error. Message from log data {}, stack trace {}, message from error {}, args {}", logData.getMessage(), error == null ? "null" : ExceptionUtils.getStackTrace(error), error == null ? "" : error.getMessage(), ObjectUtils.defaultIfNull(logData.getArgs(), Collections.emptyList()));
+            // TODO set the userId before calling the segment build function
+            analyticsOnAnalytics.enqueue(TrackMessage.builder("segment_error").userId("segmentError")
```
mohanarpit, line 63: Why are we setting the `userId` to `segmentError` here? Will this lead to errors in our mixpanel dashboards?

AnaghHegde: From the error logs, we figured out this is where it is failing. This might break the dashboards of errors in mixpanel but I wanted to confirm if the error is because we are not setting the userId in this code block. 
We were earlier using 2.1.1 which only checks for the null for userId but with the latest upgrade, it also checks for empty strings. 
I will add the user details after verifying the fix in the release.

mohanarpit (approved):

PR #18440, opened 2022-11-24
chore: Switched to sequential checks instead of a parallel one for RTS check

nidhi-nair:
Added some logs as well as moved from zipped results to sequential requests for CS and RTS

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/InstanceConfig.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/InstanceConfig.java
@@ -39,9 +39,9 @@ public class InstanceConfig implements ApplicationListener<ApplicationReadyEvent
         configService.getByName(Appsmith.APPSMITH_REGISTERED)
                 .filter(config -> Boolean.TRUE.equals(config.getConfig().get("value")))
                 .switchIfEmpty(registerInstance())
-                .zipWith(performRtsHealthCheck())
-                .doOnSuccess(ignored -> this.printReady())
-                .doOnError(ignored -> this.printReady())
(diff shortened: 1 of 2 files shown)


mohanarpit:
@nidhi-nair QQ: Will this have any noticeable performance impact on the server startup time?

nidhi-nair:
Not really @mohanarpit , the RTS health check endpoint responds in <10ms and does not involve DB updates.

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/InstanceConfig.java
```diff
@@ -70,24 +70,31 @@ private Mono<? extends Config> registerInstance() {
                 }))
                 .flatMap(responseEntity -> {
                     if (responseEntity.getStatusCode().is2xxSuccessful()) {
-                        return Mono.justOrEmpty(responseEntity.getBody());
+                        return Mono.justOrEmpty(Objects.requireNonNull(responseEntity.getBody()).getData());
```
sharat87, line 73: This change is intentional, right? Asking because we are adding a `.getData` call here, which looks like it'll change the behaviour of this return value, but I only see logs changes and sequential check mentioned in the PR. So just checking if this is something that got unintentionally pushed.

sharat87 (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/InstanceConfig.java
nidhi-nair: Yeah we expect to get the instanceId all the time here, I was intending to log it to verify the response but decided against it. Not a required change, but shouldn't break anything here.

PR #18451, opened 2022-11-25
chore: Support using custom port(s) for listening in the start-https.sh script

sharat87:
Added `--https-port` and `--http-port`, so you can set a custom port instead of the default 80 and 443. This is useful if you want to test Appsmith when using a non-standard port, fox example, how `appsmith.URL.host` behaves, or how OAuth redirects behave etc.

We also add a shortcut to start this with release endpoint. Just using `release` is now the same thing as using `https://release.app.appsmith.com`. This is useful to people like me who are _very_ good with typos.

### DIFF
--- a/app/client/start-https.sh
+++ b/app/client/start-https.sh
@@ -22,13 +22,19 @@ if [[ ${1-} =~ ^-*h(elp)?$ ]]; then
 
         If neither of the above ar set, then we check if mkcert is available, and use https if yes, or http otherwise.
 
+--https-port: Port to use for https. Default: 443.
+ --http-port: Port to use for http. Default: 80.
+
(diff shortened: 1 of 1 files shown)


mohanarpit (approved):

satbir121:
@sharat87 Are we going to update documentation?

sharat87:
@satbir121, I'm treating the output of `-h` of this script to be the docs for this. Anything else will quickly go stale. The Client Setup document talks about using this script, but doesn't mention the arguments it supports, just basic usage. Custom ports is a very special use case, and I don't think is needed in a doc intended for new folks. What do you think?

mohanarpit (approved):

PR #18466, opened 2022-11-26
chore: Fix analytics NPE for checkout remote branch flow

AnaghHegde:
Noticed the NPE error on git checkout remote branch flow on a user call.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
@@ -1297,7 +1297,7 @@ public class GitServiceCEImpl implements GitServiceCE {
                             .flatMap(application1 -> addAnalyticsForGitOperation(
                                     AnalyticsEvents.GIT_CHECKOUT_REMOTE_BRANCH.getEventName(),
                                     application1,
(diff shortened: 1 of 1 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/GitServiceCEImpl.java
```diff
@@ -1297,7 +1297,7 @@ private Mono<Application> checkoutRemoteBranch(String defaultApplicationId, Stri
                             .flatMap(application1 -> addAnalyticsForGitOperation(
                                     AnalyticsEvents.GIT_CHECKOUT_REMOTE_BRANCH.getEventName(),
                                     application1,
-                                    application1.getGitApplicationMetadata().getIsRepoPrivate()
+                                    Optional.ofNullable(application1.getGitApplicationMetadata().getIsRepoPrivate()).isPresent() && application1.getGitApplicationMetadata().getIsRepoPrivate()
```
sharat87, line 1300: I think this can be simplified to:

```suggestion
                                    Boolean.TRUE.equals(application1.getGitApplicationMetadata().getIsRepoPrivate())
```

If value is `true`, we get `true`, if `false`, we get `false`, if `null`, we get `false`, which is the same behavior.

sharat87 (approved):

PR #18581, opened 2022-11-30
chore: Update NavigateTo tests to verify params

hetunandu:
## Description

Recently, navigation broke where params were not retained when using the navigateTo function. Adding an assertion here to make sure it does not break again

### DIFF
--- a/app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/ActionExecution/NavigateTo_spec.ts
+++ b/app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/ActionExecution/NavigateTo_spec.ts
@@ -32,7 +32,11 @@ describe("Navigate To feature", () => {
       .contains("Query Params")
       .siblings()
       .find(".CodeEditorTarget")
-      .then(($el) => cy.updateCodeInput($el, "{{ { test: '123' } }}"));
+      .then(($el) => cy.updateCodeInput($el, "{{{ test: '123' }}}"));
+    agHelper.ClickButton("Submit");
+    cy.url().should("include", "a=b");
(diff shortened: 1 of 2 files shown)


app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/ActionExecution/NavigateTo_spec.ts
```diff
@@ -29,9 +28,16 @@ describe("Navigate To feature", () => {
     cy.get(".t--open-dropdown-Select-Page").click();
     agHelper.AssertElementLength(".bp3-menu-item", 2);
     cy.get(locator._dropDownValue("Page2")).click();
+    cy.get("label")
+      .contains("Query Params")
+      .siblings()
+      .find(".CodeEditorTarget")
+      .then(($el) => cy.updateCodeInput($el, "{{ { test: '123' } }}"));
     deployMode.DeployApp();
```
rahulramesha, line 36: @hetunandu the issue was only in edit mode and this test in view mode, which did not break. Also, i remember @Aishwarya-U-R was also working on a similar spec.

rahulramesha (changes requested):

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/ActionExecution/NavigateTo_spec.ts
hetunandu: Updated to add edit mode assertion

rahulramesha (approved):

akash-codemonk (approved):

PR #18583, opened 2022-11-30
chore: Analytics for entity explorer recent action click

### LINKED ISSUE #18347: [Task]: Create an analytics event for navigating to files

### SubTasks

We require a mixpanel event which triggers whenever a user opens an action (jsObject, query, or API). This should be triggered when a user navigates to a page using the entity explorer. The event should have the following properties:
- Action name
- If the action has been opened in the last 5 actions already (true or false) (we maintain a queue of the last 5 actions which have been opened, and if the one being opened is in that queue, we return true, otherwise we return false)
- the length of the queue (if the user hasn't opened any items yet, this should return 0)

The max queue length we need to maintain is 5 actions. We don't need to consider what happens on the canvas, or what happens in each action page.

eco-monk:
## Description

- Used to keep track of recently accessed entities by the user. (upto 5)
- Triggered when a user navigates to a JsObject/Query/API using the entity explorer.
- Results are persisted in indexed db.

The event has the following properties:
- id
- type: JSOBJECT, Queries/Apis
- name
- isRecent: boolean
- recentActionslength: number (max 5)

Fixes #18347

## Type of change

## How Has This Been Tested?
- Manual

### Dev activity

### QA activity:


hetunandu (changes requested):
Why did we not just use this saga

https://github.com/appsmithorg/appsmith/blob/release/app/client/src/sagas/RecentEntitiesSagas.ts#L100?

eco-monk:
> Why did we not just use this saga
> 
> https://github.com/appsmithorg/appsmith/blob/release/app/client/src/sagas/RecentEntitiesSagas.ts#L100?

@hetunandu
Recent entities is used to keep track of all recently accessed entities. What we want here is to track only actions.

Or are you suggesting we move the below (including relevant methods) into `RecentEntitiesSagas.ts`. Instead of creating a new file?
```
takeLatest(
  ReduxActionTypes.UPDATE_RECENT_ACTION_ENTITY,
  updateRecentEntitySaga,
),
```

hetunandu:
@eco-monk 
Just add the Analytics code based on the type here. OR add for all, we can filter it when querying on the analytics dashboard
https://github.com/appsmithorg/appsmith/blob/release/app/client/src/sagas/RecentEntitiesSagas.ts#L100

eco-monk:
@hetunandu, we don't want this analytics event to be triggered on path change.
We want this to be triggered only when the user clicks on the entity explorer. 
Hence the actions are dispatched from entity explorer intentionally.

The former currently tracks through URL change, which is not ideal for this.
e.g. will be triggered on page load / cmd+click (in the future).

eco-monk:
@hetunandu, triggering this from the entity explorer itself sounds like a good idea.!

To find out if it is part of the recent entities, I don't think we can re-use the existing list.
- We don't want a list that updates whenever the path has changed.
- We want to refer to a list that only tracks entity explorer clicks.

Perhaps we should rename the event and sagas? So, it's purpose is clear.
`ACTION_CLICKED_FROM_EXPLORER` or something along those lines.

Issue #12308, opened 2022-03-28
[Bug]-[4000]:Table Widget Skeleton should not disappear till the data is loaded


danciaclara:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

On the deployed app, there is no way to indicate that the queries/functions are loading as the user looks at the page with blank widgets. On the edit mode, we have a spinner at the top but this doesn't show in the deployed version.  Would be nice to have the spinner on the top bar or have the cursor change to a clock icon. 

### Why should this be worked on?

Suggestion by a discord user - https://discord.com/channels/725602949748752515/760761686549463060/957322085242384446

Nikhil-Nandagopal:
@danciaclara the skeleton loading state is what indicates this right?
Cursor is not a good experience because it means the entire application is unresponsive and loading which is never the case. The respective widgets going into loading state seems to be the right UX to me so I'm closing this but please do re-open with more information if you think I missed something

danciaclara:
@Nikhil-Nandagopal Yeah that makes sense about the cursor. However, the skeleton screen disappears too quickly when the application is a bit heavy.  It takes a few seconds for the table to populate after the skeleton screen disappears. Probably that caused the user to ask for the loading state as the skeleton screen went unnoticed.  For eg: Here's a screenshot of the page showing the table without any data. <img width="1403" alt="Screenshot 2022-03-29 at 2 09 58 PM" src="https://user-images.githubusercontent.com/32227135/160570881-313e74e9-eba4-46d5-998c-cce60259b1a6.png"> 

Be it the skeleton screen or a circular loading state for individual widgets (as different widgets load times may differ based on the query) the data should appear immediately after the loading state.

Nikhil-Nandagopal:
@danciaclara yeah that makes sense. This is a performance problem and the loading state should not change till the evaluation is complete

dilippitchika:
I think this is a problem only with table widget, but not with any other widgets. Does anyone have any evidence for other widgets on this?

dilippitchika:
Needs triaging:

We need to understand if this is happening because of any code in the table widget or because of the evaluations. 

Please check our a-force app to understand the empty table issue.

dilippitchika:
Reach - 303
Impact - High

sbalaji1192:
[...]

PR #18599, opened 2022-12-01
fix: list v2 loading state

### LINKED ISSUE #18105: Task: List v2 - Server side pagination - flashing rows on page change
When server side pagination is enabled, and the next page is clicked; the rows flash multiple times before settling on the new page rows
https://www.loom.com/share/837e0f2ed750498ea89464d560d7b53b

ashit-rath:
## Description

- Fixes the loading state of the list widget when data changes in server side paginated list. 
- Isolates the loader to a separate component

Note: This has a monkey patch of `isDataLoading` which fails in multiple scenarios.
For for a simple use case where the list data is just the query, the patch should work fine
The actual fix is expected in #12308 and the fix we are introducing can be reverted after that.

Fixes #18105

Media
Please check issue for loom video

## Type of change

## How Has This Been Tested?
- Manual
### Test Plan

> Add Testsmith test cases links that relate to this PR

### Dev activity

### QA activity:

### DIFF
--- a/app/client/src/widgets/ListWidgetV2/widget/index.tsx
+++ b/app/client/src/widgets/ListWidgetV2/widget/index.tsx
@@ -1,10 +1,16 @@
 import equal from "fast-deep-equal/es6";
 import log from "loglevel";
 import React, { createRef, RefObject } from "react";
-import { get, range, isEmpty, floor } from "lodash";
+import { get, isEmpty, floor } from "lodash";
 import { klona } from "klona";
 
+import BaseWidget, { WidgetOperation, WidgetProps } from "widgets/BaseWidget";
 import derivedProperties from "./parseDerivedProperties";
+import ListComponent, { ListComponentEmpty } from "../component";
+import ListPagination, {
+  ServerSideListPagination,
+} from "../component/ListPagination";
+import Loader from "../component/Loader";
 import MetaWidgetContextProvider from "../../MetaWidgetContextProvider";
 import MetaWidgetGenerator, { GeneratorOptions } from "../MetaWidgetGenerator";
 import WidgetFactory from "utils/WidgetFactory";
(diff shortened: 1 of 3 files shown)


app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -439,6 +440,10 @@ class ListWidget extends BaseWidget<ListWidgetProps, WidgetState> {
     return this.props.serverSidePagination && this.props.onPageSizeChange;
   };
 
+  hasListDataUpdated = (prevListData: ListWidgetProps["listData"]) => {
+    return this.isDataLoading && prevListData !== this.props.listData;
```
Tooluloope, line 444: We might need to improve the logic here


https://user-images.githubusercontent.com/31691737/205031447-b78e9541-054d-4269-a7e2-588cfca11604.mov

Tooluloope (changes requested):

app/client/src/widgets/ListWidgetV2/widget/index.tsx
ashit-rath: That's a good find!
I'm wondering, how do we even detect this? I feel now we are better off without the monkey patch. What do you think?

Tooluloope: Yeah, It's might be tedious to detect all because anything could trigger a refresh of the Data on List widget. Even an onPageSizeChange could trigger a fresh of data.

And this might lead to inconsistent behavior throughout.

ashit-rath: Yeah, I'm reverting it and keeping just the loader changes.

Tooluloope (approved):

PR #18609, opened 2022-12-01
fix: update permission driven ctas to handle auto save update changes

### LINKED ISSUE #18600: [Bug]: When you create a new datasource, it is not editable

### Description

When I create a new datasource, the datasource form is in an uneditable state. We cannot update its name and the save, test & delete buttons at the bottom of the form are also disabled. If you click on back or any other page/action, then a new popup which got introduced recently pops up asking if you want to save this datasource or not.

Things to be handled in this issue:
1. If the datasource is just created (basically if you are in the create datasource flow), then the buttons and editing of this form should not work based on permissions and all of it should be enabled.
2. If the datasource is already existing (that is if you are simply viewing it or editing the existing datasource), then the form should be driven based on permissions.
3. Handle the new popup which got introduced recently. If the datasource is existing, then the save button on the popup should also be enabled based on permissions. If the datasource is newly created (datasource creation flow), then the save button should be enabled and not work based on permissions.

### Steps To Reproduce

1. Go to Edit application
2. Create a new datasource
3. Notice that the form is uneditable
4. If you click on back or anywhere which changes the current page/form open, then a popup will show up asking if you want to save or don't save the datasource.

### Public Sample App

_No response_

### Issue video log

_No response_

### Version

Release instance

berzerkeer:
## Description
There has been some change to the datasource create flow. Before as soon as the user click the `+` next to the datasource entity, it creates a datasource and then let the user edit the datasource form. But now, we only create datasource only when the user saves the datasource. 

Since the datasource only gets created on save, there will be no permissions. So the CTAs which are permission driven will all be disabled in EE.  So fixed that and handled it for the scenarios mentioned in the linked issue.

Fixes #18600 

Media

## Type of change

## How Has This Been Tested?

- Manual

### Dev activity

### DIFF
--- a/app/client/src/pages/Editor/DataSourceEditor/SaveOrDiscardDatasourceModal.tsx
+++ b/app/client/src/pages/Editor/DataSourceEditor/SaveOrDiscardDatasourceModal.tsx
@@ -10,16 +10,33 @@ import {
   DialogComponent as Dialog,
   Size,
 } from "design-system";
+import { TEMP_DATASOURCE_ID } from "constants/Datasource";
+import { hasManageDatasourcePermission } from "@appsmith/utils/permissionHelpers";
 
 interface SaveOrDiscardModalProps {
   isOpen: boolean;
   onDiscard(): void;
   onSave?(): void;
   onClose(): void;
+  datasourceId: string;
+  datasourcePermissions: string[];
 }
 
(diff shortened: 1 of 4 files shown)


ankitakinger (approved):
LGTM

PR #18629, opened 2022-12-02
fix: Add user permission in the response for update datasource api

### LINKED ISSUE #18601: [Bug]: The query button will be disabled  and edit icon disappears when user edits datasource name

### Description

The query button will be disabled  and edit icon disappears when user edits datasource name  image
image
image

### Steps To Reproduce

1. go to datasource page
2. click on new datasource
3. click on back button and save the datasource 
4. now edit the database name
5. The query button which was in enabled state will be disabled now
6. Now manually refresh the page
7. the query button will be enabled

### Public Sample App

_No response_

### Issue video log

_No response_

### Version

localhost

trishaanand:
When the datasource is updated, in the response the server did not set the current permissions of the user. This is now required as part of granular access control for the client to enable/disable actions on top of the datasource (like create a query, save, etc.). This PR adds the permissions in the response.

Fixes #18601


nsarupr (approved):
Looks good to me.

PR #18715, opened 2022-12-06
refactor: JSONForm eliminate the use of canvasWidgets and remove childStylesheet from dynamicBindingPathList

### LINKED ISSUE #16616: JSONForm Widget crashes inside List widget
Dragging a JSON form crashes the list widget

ashit-rath:
## Description

TL;DR
This refactors some of the code for JSONForm widget, which helps it work along with the upcoming List v2.0. It also contains some changes to theming's internal working which may improve some performance (very minor) but it originally spawned because of the first refactor.

This PR comprises 2 refactors.
The JSONForm used `canvasWidgets` directly to get the uneval `schema` and uneval `childStylesheet` as it used both to generate the next schema.

How are both the problems eliminated:
1. uneval `schema`
Instead of taking the old schema and overwriting the modified/added/deleted bits to it, the schema parser now only generates the first schema if not present and post that the parser provides paths of the properties that got deleted and the modifed/added paths with value. It still returns the amended schema using the prev but it's not used anymore. The paths that are returned are then directly used to update using the `batchUpdateWidgetProperty`. We realised that when something is modified or added to the source data, fresh config is always generated to old config was not in play so uneval old schema was not need anymore.

2. uneval `childStylesheet`
The reason for getting evaluated `childStylesheet` was when a new widget was dropped, the paths of `childStylesheet` was automatically added the the `dynamicBindingPathList`. This lead to the the property to come always in the evaluated form. 
This was fixed along with a migration that removes all the widgets where the paths are added to the dynamicBindingPathList. 

Fixes #16616

#### Note to QA
This PR does not contain any bug/feature and should not introduce any regression .

- This PR requires a full regression of the JSONForm
- Theming aspects of the JSONForm, Table widget and the Button Group widget needs to tested as well
- This PR contains a migration, post migration nothing new is introduced. Only internal properties are modified so everything should work as is.

## Type of change

- Refactor (non-breaking change which fixes an issue)

## How Has This Been Tested?
- Jest test 
  - for the migration
  - for the schema parser

### Dev activity

### QA activity:


app/client/src/utils/migrations/ThemingMigrations.ts
```diff
@@ -612,3 +613,31 @@ export const addPropertyToDynamicPropertyPathList = (
     ];
   }
 };
+
+export const migrateChildStylesheetFromDynamicBindingPathList = (
+  currentDSL: DSLWidget,
```
sbalaji1192, line 618: Could you have a look at this utility we wrote to simplify the migration codes `traverseDSLAndMigrate`

ashit-rath: Ohh nice, I was looking for something like this.

sbalaji1192 (approved):

chandannkumar:
Tested this PR and LGTM
- Regression on JSONForm widget
- Regression on Table, Button group widget related to theming aspects

sbalaji1192 (approved):

PR #18731, opened 2022-12-06
Fix/pagination issues in serverside pagination listv2

### LINKED ISSUE #11332: [Bug]: List widget - When using pagination, no way to go back to previous page after reaching last page

### Current Behavior

When using server-side pagination on the list widget, if you click on the Next button after the last page with data, it shows "No data to display" and then there is no way to go back to the previous page and the list widget stays stuck on the no data page.

Uploading Screenrecording_20220222_084207.mp4…

.

### Steps To Reproduce

1. Connect data to a list widget and use server-side pagination
2. Go to the last page with data and click Next
3. It shows a page with the text "No data to display" 
4. Now there is no way to go back to the previous page

### Environment

Production

### Version

Cloud

Tooluloope:
## Description

- Fixed serverside pagination code and css 
- Disabled pagination when isLoading
- 

Fixes #11332

## Type of change

## How Has This Been Tested?
- Cypress

### Dev activity

### QA activity:

### DIFF
--- a/app/client/src/widgets/ListWidgetV2/component/ListPagination.tsx
+++ b/app/client/src/widgets/ListWidgetV2/component/ListPagination.tsx
@@ -398,8 +398,18 @@ function ListPagination(props: ListPaginationProps) {
       current={props.pageNo}
       disabled={props.disabled}
       locale={locale}
+      nextIcon={() => (
+        <button aria-label="next page" className="rc-pagination-item-link">
+          <Icon name="right-arrow-2" size={IconSize.XXL} />
+        </button>
+      )}
       onChange={props.onChange}
       pageSize={props.pageSize}
+      prevIcon={() => (
+        <button aria-label="prev page" className="rc-pagination-item-link">
+          <Icon name="left-arrow-2" size={IconSize.XXL} />
+        </button>
+      )}
       showTitle={false}
       total={props.total}
     />
--- a/app/client/src/widgets/ListWidgetV2/component/ListPagination.tsx
+++ b/app/client/src/widgets/ListWidgetV2/component/ListPagination.tsx
@@ -431,6 +431,7 @@ interface ServerSideListPaginationProps {
   nextPageClick: () => void;
   pageNo: number;
   prevPageClick: () => void;
+  isLoading: boolean;
 }
 
 export function ServerSideListPagination(props: ServerSideListPaginationProps) {
--- a/app/client/src/widgets/ListWidgetV2/component/ListPagination.tsx
+++ b/app/client/src/widgets/ListWidgetV2/component/ListPagination.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import Pagination from "rc-pagination";
 import styled, { css } from "styled-components";
+import { Icon, IconSize } from "design-system";
 
 const locale = {
   // Options.jsx
@@ -17,7 +18,12 @@ const locale = {
   next_3: "Next 3 Pages",
 };
 
-const paginatorCss = css`
+const paginatorCss = css<{
+  borderRadius: string;
+  boxShadow?: string;
+  accentColor: string;
+  disabled?: boolean;
+}>`
   margin: 0 auto;
   padding: 0;
   font-size: 14px;
@@ -48,7 +54,7 @@ const paginatorCss = css`
     min-width: 28px;
     height: 28px;
     margin-right: 8px;
-    font-family: Arial;
+    font-family: inherit;
     line-height: 26px;
     text-align: center;
     vertical-align: middle;
@@ -59,6 +65,9 @@ const paginatorCss = css`
     outline: 0;
     cursor: pointer;
     user-select: none;
+    display: flex;
+    align-items: center;
+    justify-content: center;
   }
   .rc-pagination-item a {
     display: block;
@@ -74,6 +83,10 @@ const paginatorCss = css`
     border-color: #1890ff;
     transition: all 0.3s;
   }
+  .rc-pagination-item:focus a,
+  .rc-pagination-item:hover a {
+    color: #1890ff;
+  }
   .rc-pagination-item-active {
     font-weight: 500;
     background: #ffffff;
@@ -85,6 +98,10 @@ const paginatorCss = css`
   .rc-pagination-item-active:focus,
   .rc-pagination-item-active:hover {
     border-color: #40a9ff;
+  }
+  .rc-pagination-item-active:focus a,
+  .rc-pagination-item-active:hover a {
+    color: #40a9ff;
   }
   .rc-pagination-jump-prev,
   .rc-pagination-jump-next {
@@ -115,7 +132,7 @@ const paginatorCss = css`
     min-width: 28px;
     height: 28px;
     color: rgba(0, 0, 0, 0.85);
-    font-family: Arial;
+    font-family: inherit;
     line-height: 28px;
     text-align: center;
     vertical-align: middle;
@@ -140,7 +157,6 @@ const paginatorCss = css`
   }
   .rc-pagination-prev .rc-pagination-item-link,
   .rc-pagination-next .rc-pagination-item-link {
-    display: block;
     width: 100%;
     height: 100%;
     font-size: 12px;
@@ -150,6 +166,14 @@ const paginatorCss = css`
     border-radius: 2px;
     outline: none;
     transition: all 0.3s;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+
+    &::after,
+    &::before {
+      display: none;
+    }
   }
   .rc-pagination-prev:focus .rc-pagination-item-link,
   .rc-pagination-next:focus .rc-pagination-item-link,
@@ -170,6 +194,10 @@ const paginatorCss = css`
   .rc-pagination-disabled:hover,
   .rc-pagination-disabled:focus {
     cursor: not-allowed;
+
+    & > * {
+      pointer-events: none;
+    }
   }
   .rc-pagination-disabled .rc-pagination-item-link,
   .rc-pagination-disabled:hover .rc-pagination-item-link,
@@ -288,37 +316,41 @@ const paginatorCss = css`
       display: none;
     }
   }
-`;
 
-const StyledPagination = styled(Pagination)<{
-  disabled?: boolean;
-  borderRadius: string;
-  boxShadow?: string;
-  accentColor: string;
-}>`
-  ${paginatorCss}
   pointer-events: ${(props) => (props.disabled ? "none" : "all")};
   opacity: ${(props) => (props.disabled ? "0.4" : "1")};
 
   .rc-pagination-item {
     border-radius: ${({ borderRadius }) => borderRadius};
     box-shadow: ${({ boxShadow }) => `${boxShadow}`} !important;
-    color:  ${({ accentColor }) => accentColor} !important;
+    color:  var(--wds-color-text) !important;
     border-color:  transparent !important;
-
   }
 
   .rc-pagination-prev .rc-pagination-item-link, .rc-pagination-next .rc-pagination-item-link  {
     border-radius: ${({ borderRadius }) => borderRadius};
     box-shadow: ${({ boxShadow }) => `${boxShadow}`} !important;
     border-color:  transparent !important;
-    color:  ${({ accentColor }) => accentColor} !important;
+    color:  var(--wds-color-text) !important;
+
+    &:hover {
+      background-color: var(--wds-color-bg-hover) !important;
+
+      a {
+        color: var(--wds-color-text);
+      }
+    }
+  }
+
+  .rc-pagination-disabled .rc-pagination-item-link {
+    color: var(--wds-color-text-disabled) !important;
   }
 
   .rc-pagination-item:hover {
-    background-color: ${({ accentColor }) => accentColor} !important;
+    background-color: var(--wds-color-bg-hover) !important;
+
     a {
-      color: white;
+      color: var(--wds-color-text);
     }
   }
 
@@ -328,9 +360,22 @@ const StyledPagination = styled(Pagination)<{
       color: white;
     }
 
-    .rc-pagination-item-active:hover a {
-      color: white !important;
+    &:hover {
+      background-color: ${({ accentColor }) => accentColor} !important;
+
+      a {
+        color: white !important;
+      }
     }
+`;
+
+const StyledPagination = styled(Pagination)<{
+  disabled?: boolean;
+  borderRadius: string;
+  boxShadow?: string;
+  accentColor: string;
+}>`
+  ${paginatorCss}
 `;
 
 interface ListPaginationProps {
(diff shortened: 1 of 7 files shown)


app/client/src/widgets/ListWidgetV2/component/ListPagination.tsx
```diff
@@ -288,37 +316,41 @@ const paginatorCss = css`
...
-}>`
-  ${paginatorCss}
   pointer-events: ${(props) => (props.disabled ? "none" : "all")};
   opacity: ${(props) => (props.disabled ? "0.4" : "1")};
 
   .rc-pagination-item {
     border-radius: ${({ borderRadius }) => borderRadius};
     box-shadow: ${({ boxShadow }) => `${boxShadow}`} !important;
-    color:  ${({ accentColor }) => accentColor} !important;
+    color:  var(--wds-color-text) !important;
```
ashit-rath, line 326: Shouldn't `accentColor` be used for pagination control items for theming effects to take place?

app/client/src/widgets/ListWidgetV2/component/ListPagination.tsx
```diff
@@ -328,11 +360,24 @@ const StyledPagination = styled(Pagination)<{
...
     }
 `;
 
+const StyledPagination = styled(Pagination)<{
+  disabled?: boolean;
+  borderRadius: string;
+  boxShadow?: string;
+  accentColor: string;
+}>`
+  ${paginatorCss}
```
ashit-rath, line 378: Can we move all the styles here instead of defining it in a different variable?

app/client/src/widgets/ListWidgetV2/component/ListPagination.tsx
```diff
@@ -328,11 +360,24 @@ const StyledPagination = styled(Pagination)<{
...
+        color: white !important;
+      }
     }
 `;
 
+const StyledPagination = styled(Pagination)<{
+  disabled?: boolean;
+  borderRadius: string;
+  boxShadow?: string;
+  accentColor: string;
```
ashit-rath, line 376: Let's not define the typings inline. Move it to it's own `type`

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -182,15 +182,12 @@ class ListWidget extends BaseWidget<ListWidgetProps, WidgetState> {
...
-        this.triggerOnPageSizeChange();
-      }
     }
 
     if (this.isCurrPageNoGreaterThanMaxPageNo()) {
-      const maxPageNo = Math.ceil(
-        (this.props?.listData?.length || 0) / this.pageSize,
+      const maxPageNo = Math.max(
+        Math.ceil((this.props?.listData?.length || 0) / this.pageSize),
+        1,
```
ashit-rath, line 190: Does this also fixes the issue where when the data is removed and inserted back the list items never come back?

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -182,15 +182,12 @@ class ListWidget extends BaseWidget<ListWidgetProps, WidgetState> {
 
     if (this.shouldUpdatePageSize()) {
       this.updatePageSize();
```
ashit-rath, line 184: When pageSize is updated then I think we should change the pageNo to 1. 
As there is no event handler to trigger when page size changes, the query won't get fired unless we change the page to 1.

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -759,12 +742,24 @@ class ListWidget extends BaseWidget<ListWidgetProps, WidgetState> {
     this.context?.deleteWidgetProperty?.(widgetId, propertyPaths);
   };
 
+  shouldDisableNextPage = () => {
+    const { listData, serverSidePagination } = this.props;
+
+    return Boolean(serverSidePagination && !listData?.length);
```
ashit-rath, line 748: We have a similar issue with client side pagination as well. Are we fixing it here or is there another issue to do it?

app/client/src/widgets/ListWidgetV2/widget/index.tsx
Tooluloope: Yes, this fixes the issue

app/client/src/widgets/ListWidgetV2/widget/index.tsx
Tooluloope: Would this be for Client or only serverSide?

app/client/src/widgets/ListWidgetV2/widget/index.tsx
ashit-rath: Awesome! tag that issue in the description.

app/client/src/widgets/ListWidgetV2/widget/index.tsx
Tooluloope: Oh, I haven't seen the issue in client side.

The next button gets disabled automatically since there's no data. It doesn't even get to the no-data page, it disables it after the last item is displayed

app/client/src/widgets/ListWidgetV2/widget/index.tsx
ashit-rath: server side only as in client side `pageSize` and `pageNo` has no utility.

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -759,12 +742,24 @@ class ListWidget extends BaseWidget<ListWidgetProps, WidgetState> {
...
+      isLoading,
+      pageNo,
+      parentRowSpace,
+      serverSidePagination,
+    } = this.props;
     const templateHeight = this.getTemplateBottomRow() * parentRowSpace;
+    const disableNextPage = this.shouldDisableNextPage();
 
-    if (this.props.isLoading) {
+    if (isLoading) {
```
Tooluloope, line 762: @ashit-rath, Also I realized we're using `isLoading` here already. So should I remove the isLoading from the Pagination as it's redundant

app/client/src/widgets/ListWidgetV2/widget/index.tsx
ashit-rath: my bad! I misunderstood this. This is fine

app/client/src/widgets/ListWidgetV2/widget/index.tsx
ashit-rath: If we have `isLoading` here then how are we able to click the next button so many times? Is there a delay between the click and `isLoading` getting the value `true`?

ashit-rath: Whether it's loading or not I don't think in any case we should remove the pagination controls unless there's no data or no vertical space in which case it's fine.

Now there are 2 cases
1. Data from JSObject - `isLoading` is always false here and as the value is static, we cannot do much here. Unless we develop a system where we detected if meta widgets are regenerated over different render cycle (Too much effort)
2. Data from Query - In this case we can use the `isLoading` for disabling the button.

app/client/src/widgets/ListWidgetV2/component/ListPagination.tsx
Tooluloope: We used this variable in two places, `PaginationWrapper` and `StyledPagination`, so i think it's best to have it as a variable

app/client/src/widgets/ListWidgetV2/component/ListPagination.tsx
```diff
@@ -1,6 +1,7 @@
 import React from "react";
 import Pagination from "rc-pagination";
 import styled, { css } from "styled-components";
+import { Icon, IconSize } from "design-system";
```
ashit-rath, line 4: We aren't suppose to use design-system in widgets. Sadly we would have to wait for wds.

app/client/src/widgets/ListWidgetV2/component/ListPagination.tsx
ashit-rath: @Tooluloope Can you address this?

app/client/src/widgets/ListWidgetV2/component/ListPagination.tsx
Tooluloope: This was recently added to List WIdget V1

Tooluloope:
@ashit-rath, `accentColor` is not been used for backGround color and these wds variables were created during reskinning by @jsartisan

chandannkumar:
@Tooluloope Observation
- On deploying List v2 with SSP enabled, the app shows error of 'no data to display'
video of List v2 : https://bthrujcsw8.vmaker.com/record/nBTi9beuyWNT2CX4

- The issue on List v1 still remains same. Error of 'no data to display' after reaching last page
Video of List v1 : https://bthrujcsw8.vmaker.com/record/oqYd1AcWsqsADoe4

QQ: This PR will also fix the current issue in List v1 as well?

Tooluloope:
Hi @chandannkumar, 
1. The first issue occured because you enabled Inifinite Scroll.
I'll be pushing a fix for that now, as we're not enabling infinite scroll in this release.

2. The fix would only be for this List V2 not V1

chandannkumar:
> 1. The first issue occured because you enabled Inifinite Scroll.
This is also happening when Infinite Scroll is disabled

kamakshibhat-appsmith:
observations:
1. data stops rendering onPageChange during server-side pagination when Primary key is not set 
https://www.loom.com/share/25fea7e1f34d44bdaca13894feef6bcb
2. Primary key gets deselected when last page is reached but when we do the JS toggle, the reference seems to be there
https://www.loom.com/share/8abbe114318a4f649b92d8c55785d7a2
@Tooluloope

Tooluloope:
@kamakshibhat-appsmith 
No 2. is expected, as you can see once you toggle to JS you cant toggle back. and the reason the Primary Key goes is that there's no Data at that point

@ashit-rath should we cache the keys or is that going to be too much overhead?

kamakshibhat-appsmith:
@Tooluloope 
For number 2, I got the point why  primary key goes but on the JS toggle , shouldn't the reference also be removed?

ashit-rath:
@Tooluloope Do we have enough clarity to move this forward?

Tooluloope:
> @Tooluloope Do we have enough clarity to move this forward?

Sure, I wanted to finish up on the cypress Test first

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -633,6 +624,40 @@ class ListWidget extends BaseWidget<
...
+  shouldUpdatePrimaryKeyOptions = (prevProps: ListWidgetProps) => {
+    if (Array.isArray(this.props.listData) && this.props.listData.length) {
+      return !isEqual(
+        Object.keys(prevProps.listData?.[0] ?? {}),
+        Object.keys(this.props.listData?.[0] ?? {}),
+      );
+    }
+
+    return false;
+  };
```
ashit-rath, line 660: This will cause undo-redo issue.
Can you create a separate PR for the primary key changes? let's deal with it separately

app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.ts
```diff
@@ -984,6 +984,14 @@ class MetaWidgetGenerator {
     if (this.pageNo !== nextOptions.pageNo) {
       this.modificationsQueue.add(MODIFICATION_TYPE.PAGE_NO_UPDATED);
     }
+
+    // only needed when there's no set primaryKey and we have to regenerate.
```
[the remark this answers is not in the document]
Tooluloope: In serverside Pagination, 2 major things trigger the regeneration
1. Change in PageNo
2. Change in PrimaryKeys (Since Data Length remains constant, only PrimaryKeys signifies a change in data)

Once Primarykey is undefined, This change in data isn't known.

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -30,6 +30,14 @@ import {
 import { EventType } from "constants/AppsmithActionConstants/ActionConstants";
 import { ModifyMetaWidgetPayload } from "reducers/entityReducers/metaWidgetsReducer";
 import { WidgetState } from "../../BaseWidget";
+import { Stylesheet } from "entities/AppTheming";
+import { getBindingTemplate } from "../constants";
+
+export const isValidListData = (
```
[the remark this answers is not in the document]
Tooluloope: Got it.

app/client/src/widgets/ListWidgetV2/widget/index.tsx
Tooluloope: In what way would this cause an undo-redo issue?

ashit-rath: For undo-redo to work, each operation has to be atomic. Here when we change list data, we are going to update the property.

So that leads to 2 operation and when  the user would try to undo, they would effectively undo the update of this property rather than the undo of data change.

I think in list v2 there might be a few places more where we might have made the same mistake. We need to find and fix them if they can be avoided.

app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.ts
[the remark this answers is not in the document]
ashit-rath: But what are we regenerating here if data and primary keys are undefined? the forEach loop inside the generate will not run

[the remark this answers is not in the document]
Tooluloope: No, the data isn't undefined, only the Primary key is undefined
As you can see below, only `data.length` and `primaryKeys` signify changes in data.

`data.length` hardly changes, so `primaryKeys` must change when moving to new data, especially on serverSide. 

Screenshot 2022-12-29 at 06 31 31

app/client/src/widgets/ListWidgetV2/widget/index.tsx
Tooluloope: oh, just understood.
That's a nasty bug.

I'll create a separate issue for the PrimaryKeys I tackled here.
cc @kamakshibhat-appsmith @chandannkumar

app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.ts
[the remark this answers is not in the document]
ashit-rath: Okay so you are saying, 
if in the current page the data changes somehow
the data length is the same 
the primary keys are undefined
then we need to regenerate the meta widgets?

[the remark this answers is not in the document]
Tooluloope: Operation Moving to Next Page

2 Things would change at different times, 
-  page No would change immediately -> causes regeneration of widgets(In hindsight, this might not be necessary)
- Data would later change -> this is only communicated by `primaryKey` since the `data.length` is still constant.

So yeah, we should regenerate the meta widgets only at this point.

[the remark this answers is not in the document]
ashit-rath: Oh I see the point now. Can you add some more comments?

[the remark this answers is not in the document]
ashit-rath: Also move this condition in the if statement where we have 
```
if (hasRegenerationOptionsChanged) {
...
}
```

app/client/src/widgets/ListWidgetV2/widget/index.tsx
Tooluloope: https://github.com/appsmithorg/appsmith/issues/19299

ashit-rath (approved):

chandannkumar:
Tested this PR and LGTM

PR #18791, opened 2022-12-08
feat: [APPSMTH-30] When field is hidden in a json form allow user to set if they want to use source data as fallback

### LINKED ISSUE #14231: [Feature]: When field is hidden in a json form allow user to set if they want to use source data as fallback

### Summary

Add a new switch for dev to configure when visible is set to off, to use the source data in form data rather than user having to write JS code for this.

When i hide a field in json form, i want to access it's value from the source data so that my update query doesn't remove the fields altogether

gitstart-appsmith:
## Description
When fields is hidden in a json form, those fields value are no longer available on the widget formData. This implementation is meant to give dev the options to set if they want to use sourceData that still contains hidden fields value as fallback when some fields are hidden.

### Step To Reproduce
1. Drop a JSON form
2. create a field and its value is present in the Jsonform1.formdata
3. make this field invisible and its value is not present in the formdata anymore.
4. do the same thing with a simple form, the field is present regardless of the field visiibility.

NB: This would require updating the Appsmith Documentation

Fixes #14231 

## Type of change

## How Has This Been Tested?
- Cypress Tests

## Checklist:


app/client/src/widgets/JSONFormWidget/helper.ts
```diff
@@ -134,6 +134,18 @@ const convertArrayTypeToFormData = (
   return;
 };
 
+/**
+ * Functions merge hidden field values present in source data to form data values.
+ */
+export const mergeSourceAndFormData = (formValue: any, sourceData: any) => {
+  const formData: Record<string, unknown> = {};
+  Object.keys({ ...formValue, ...sourceData }).forEach((key) => {
+    formData[key] = formValue[key] ?? sourceData[key];
```
ashit-rath, line 143: The logic of merging this has some flaws.

The idea here is that when a field's `isVisible` property is `false` and the `useSourceData` is turned on, only then we use source data as the fallback for that particular field only and not others.

If the source data looks something like
```js
{
  name: "John",
  age: 50
}
```

We set the default value of the age field to be `null`

Now the form data would come out like this (with `useSourceData` disabled)
```js
{
  name: "John",
  age: null
}
```

with `useSourceData` enabled it would look something like
```js
{
  name: "John",
  age: 50
}
```
^ This should not happen as the age field is not hidden but due to the override logic it will happen.

The ideal way to solve this would be to do it in `convertSchemaItemToFormData` when when go field by field and check the `isVisible` and `useSourceData`.

app/client/src/widgets/JSONFormWidget/helper.test.ts
```diff
@@ -824,4 +824,48 @@ describe(".convertSchemaItemToFormData", () => {
 
     expect(result).toEqual(expectedOutput);
   });
+
+  it("set hidden form field value to available source field value when useSourceData is set to true", () => {
+    const formData = {
+      customField1: "male",
+      customField2: "demo",
+      array: [{ name: "test1" }, { name: null }],
```
ashit-rath, line 832: in the 2nd object of array add `date` and have an entry for the same in the `sourceData`. Here `data` field has `isVisible: false`

app/client/src/widgets/JSONFormWidget/helper.test.ts
```diff
@@ -824,4 +824,48 @@ describe(".convertSchemaItemToFormData", () => {
...
+
+    const expectedOutput = {
+      gender: "male",
+      age: "sourceDemo",
+      students: [{ firstName: "test1" }, { firstName: null }],
+      testHiddenArray: [
+        { firstName: "sourceTest1" },
+        { firstName: "sourceTest2" },
+      ],
+      testVisibleObject: { firstName: "test1" },
```
ashit-rath, line 854: Great test condition for  `testVisibleObject`.

app/client/src/widgets/JSONFormWidget/helper.test.ts
```diff
@@ -824,4 +824,48 @@ describe(".convertSchemaItemToFormData", () => {
 
     expect(result).toEqual(expectedOutput);
   });
+
+  it("set hidden form field value to available source field value when useSourceData is set to true", () => {
```
ashit-rath, line 828: We need to upgrade the `schema` object to add new schema items.
There are cases when `sourceData` object keys are non-ASCII, for example smileys, other language than english.
Checkout the file `src/widgets/JSONFormWidget/schemaTestData.ts`, there you would find test schemas with non-ASCII keys.
We need to have such schema items in our `schema` here to verify if the correct keys from the `sourceData` is picked up when the corresponding field is hidden

Example 
```
{
     __: {
          isDisabled: false,
          isRequired: false,
          labelTextSize: "0.875rem",
          label: "%%",
          isVisible: true,
          isSpellCheck: false,
          children: {},
          dataType: DataType.STRING,
          fieldType: FieldType.TEXT_INPUT,
          iconAlign: "left",
          sourceData: "%",
          isCustomField: false,
          accessor: "%%",
          identifier: "__",
          originalIdentifier: "%%",
          position: 5,
          boxShadow: "none",
        },
        xn__j2bd4cyac6f: {
          isDisabled: false,
          isRequired: false,
          labelTextSize: "0.875rem",
          label: "हिन्दि",
          isVisible: true,
          isSpellCheck: false,
          children: {},
          dataType: DataType.STRING,
          fieldType: FieldType.TEXT_INPUT,
          iconAlign: "left",
          sourceData: "हिन्दि",
          isCustomField: false,
          accessor: "हिन्दि",
          identifier: "xn__j2bd4cyac6f",
          originalIdentifier: "हिन्दि",
          position: 6,
          boxShadow: "none",
        },
}
```

And this would be the `dataSource` for the above

```
{
    "%%": "%",
    हिन्दि: "हिन्दि",
}

File: app/client/src/widgets/JSONFormWidget/helper.ts [hunk 5]
```diff
@@ -90,18 +92,34 @@ const convertObjectTypeToFormData = (
...
+        });
+      } else if (
+        !schemaItem.isVisible &&
+        options.useSourceData &&
+        sourceValue !== undefined
+      ) {
+        formData[toKey] = convertSchemaItemToFormData(schemaItem, sourceValue, {
+          ...options,
+          sourceValue,
+        });
```
ashit-rath, line 122: I don't think we should do all of these checks here.
The check for hidden should be done in `convertSchemaItemToFormData`
The method itself represents that it will convert 1 schema item to form data. 
So that essentially means if I pass 1 schema item with no children, then it should return me the correct form value based on the schemaItem's hidden state.

So let's remove all the code here except
```ts
const convertObjectTypeToFormData = (
  schema: Schema,
  formValue: unknown,
  options: ConvertFormDataOptions,
) => {
  if (formValue && typeof formValue === "object") {
    const formData: Record<string, unknown> = {};
    let sourceValue;
    
    if (options.sourceValue) {
        sourceValue = valueLookup(
          options.sourceValue as Record<string, unknown>,
          schemaItem,
          options.fromId,
        );
      }

    Object.values(schema).forEach((schemaItem) => {
        const toKey = schemaItem[options.toId];
        formData[toKey] = convertSchemaItemToFormData(schemaItem, value, {
          ...options,
          sourceValue,
        });
    });

    return formData;
  }

  return;
};

```

app/client/src/widgets/JSONFormWidget/helper.test.ts
```diff
@@ -824,4 +824,48 @@ describe(".convertSchemaItemToFormData", () => {
...
+      {
+        fromId: "identifier",
+        toId: "accessor",
+        sourceValue: sourceData,
+        useSourceData: true,
+      },
+    );
+
+    expect(result).toEqual(expectedOutput);
+  });
```
ashit-rath, line 870: Add another test case where we just pass 1 schema item 
```
xn__j2bd4cyac6f: {
          isDisabled: false,
          isRequired: false,
          labelTextSize: "0.875rem",
          label: "हिन्दि",
          isVisible: true,
          isSpellCheck: false,
          children: {},
          dataType: DataType.STRING,
          fieldType: FieldType.TEXT_INPUT,
          iconAlign: "left",
          sourceData: "हिन्दि",
          isCustomField: false,
          accessor: "हिन्दि",
          identifier: "xn__j2bd4cyac6f",
          originalIdentifier: "हिन्दि",
          position: 6,
          boxShadow: "none",
        }
        ```
and the `sourceData` would be `"हिन्दि"` and `formData` would be `"test"` and verify if it returns the correct value.

File: app/client/src/widgets/JSONFormWidget/helper.ts [hunk 7]
```diff
@@ -90,18 +92,34 @@ const convertObjectTypeToFormData = (
     const formData: Record<string, unknown> = {};
 
     Object.values(schema).forEach((schemaItem) => {
+      let sourceValue;
+      if (options.sourceValue) {
+        sourceValue = valueLookup(
+          options.sourceValue as Record<string, unknown>,
+          schemaItem,
+          options.fromId,
```
ashit-rath, line 100: The `options.fromId` should not be used here for lookup into the the `sourceData` as the `formId` can be anything like `accessor`, `identifier`, or any other key.
But the `sourceData` lookup can only by done by `originalIdentifier`.
Right now it's working because all the schemaItem have the same `identifier` and `originalIdentifier` value. Once they differ, eg. when non-ASCII keys are used in the source data, then this would start returning `undefined` instead of a value.

So hardcode this parameter to `"originalIdentifier"` and add some comments around it.

app/client/src/widgets/JSONFormWidget/widget/index.tsx
```diff
@@ -236,14 +237,34 @@ class JSONFormWidget extends BaseWidget<
...
+      this.updateAndMergeFormDataIfRequired();
+    }
     const { schema } = this.constructAndSaveSchemaIfRequired(prevProps);
     this.debouncedParseAndSaveFieldState(
       this.state.metaInternalFieldState,
       schema,
     );
   }
 
+  updateAndMergeFormDataIfRequired() {
```
ashit-rath, line 250: Instead of defining a new method we can directly call `updateFormData` method. Does the same thing

gitstart-appsmith:
Thank you @ashit-rath , we'd look into these reviews.

app/client/src/widgets/JSONFormWidget/helper.ts
gitstart-appsmith: @ashit-rath 
This couldn't implemented.

First off the snippet above would generate an error as the variable `schemaItem` is been used at 
```typescript
sourceValue = valueLookup(
          options.sourceValue as Record<string, unknown>,
          schemaItem,
          options.fromId,
         );
      }

    Object.values(schema).forEach((schemaItem) => {
```
outside of where it is been defined.

And again checking for hidden fields at the function where suggested seems not workable as that is what is been used inside this `forEach` array method.
```javascript
 Object.values(schema).forEach((schemaItem) => {
      let sourceData;
      if (options.sourceData) {
        sourceData = valueLookup(
          options.sourceData as Record<string, unknown>,
          schemaItem,
          // sourceData lookup can only be done using originalIdentifier
          // if sourceData lookup done using other id e.g. accessor or identify, the return
          // value could be undefined.
          "originalIdentifier",
        );
      }
      const toKey = schemaItem[options.toId];
      if (schemaItem.isVisible) {
```
to ignore hidden field.

ashit-rath: @gitstart-appsmith Yes, you are right; I see the problem.
Instead can we do something like this at the very first of `forEach`
```
if (schemaItem.hidden && !options.useSourceData) return;
```
^ Doing this will make sure that the previous behaviour is retained (i.e when a field is hidden it's key never shows up in the `formData`)
and move all the logic to find the right value to `convertSchemaItemToFormData` as that is the correct place to do it.

gitstart-appsmith: @ashit-rath A lot of possible options have been attempted and non seems to be working base on the feedback.
The suggested implementation are causing existing test cases to fail.

current implementation in `covertSchemaItemToFormData` function
```typescript
 let value: unknown;
  if (!schemaItem.isVisible && options.useSourceData) {
    value = options.sourceData;
  } else if (schemaItem.isVisible) value = formValue;

  if (schemaItem.fieldType === FieldType.OBJECT) {
    return convertObjectTypeToFormData(schemaItem.children, value, options);
  }

  if (schemaItem.fieldType === FieldType.ARRAY) {
    return convertArrayTypeToFormData(schemaItem.children, value, options);
  }

  return formValue;
```

and current implementation in `convertObjectTypeToFormData`

```typescript
if (formValue && typeof formValue === "object") {
    const formData: Record<string, unknown> = {};

    Object.values(schema).forEach((schemaItem) => {
      if (!schemaItem.isVisible && !options.useSourceData) return;
      let sourceData;
      if (options.sourceData) {
        sourceData = valueLookup(
          options.sourceData as Record<string, unknown>,
          schemaItem,
          "originalIdentifier",
        );
      }
      const toKey = schemaItem[options.toId];
      let value;
      /*
      * added this if else block because without it, it would still returns hidden field value of nested object, which the 
      * possibly only alternative i could see would be to add an `else return` block in `covertSchemaItemToFormData`
      * which would add undefined to the props of the return object.
      */
      if (!schemaItem.isVisible) value = sourceData;
      else
        value = valueLookup(
          formValue as Record<string, unknown>,
          schemaItem,
          options.fromId,
        );
      if (
        !schemaItem.isVisible &&
        options.useSourceData &&
        value === undefined
      ) {
        return;
      } // added this cause when shema item is hidden and `useSourceData` is set to `true` 
      // it set the object prop to `undefined`
      formData[toKey] = convertSchemaItemToFormData(schemaItem, value, {
        ...options,
        sourceData,
      });
    });

    return formData;
  }

  return;
```

I await feedback/suggestion on this

ashit-rath: @gitstart-appsmith Got it!

1. I think we should remove the condition
```js
if (
        !schemaItem.isVisible &&
        options.useSourceData &&
        value === undefined
      ) {
        return;
      }
```
As this would mean that the `sourceData` for that particular key is `undefined` and the user might want that to be present. So we shouldn't take any decision based on what the value is. 

2. Use curly braces around the if-else conditions
```js
if (!schemaItem.isVisible) {
  value = sourceData;
} else {
   value = valueLookup(
     formValue as Record<string, unknown>,
     schemaItem,
     options.fromId,
   );
}
```

3. As we are going with this approach, we can revert the changes in `covertSchemaItemToFormData` as it seems irrelevant anymore

app/client/src/widgets/JSONFormWidget/widget/index.tsx
```diff
@@ -237,7 +238,10 @@ class JSONFormWidget extends BaseWidget<
     ) {
       this.state.resetObserverCallback(this.props.schema);
     }
-
+    if (prevProps.useSourceData !== this.props.useSourceData) {
+      const { formData, sourceData } = this.props;
+      this.updateFormData(formData, isEmpty(formData) && isEmpty(sourceData));
```
ashit-rath, line 243: Why are we adding this `isEmpty(formData) && isEmpty(sourceData)`?

app/client/src/widgets/JSONFormWidget/widget/index.tsx
```diff
@@ -237,7 +238,10 @@ class JSONFormWidget extends BaseWidget<
     ) {
       this.state.resetObserverCallback(this.props.schema);
     }
-
+    if (prevProps.useSourceData !== this.props.useSourceData) {
```
ashit-rath, line 241: Can we add a new line above line 241 and below 244?

app/client/src/widgets/JSONFormWidget/widget/index.tsx
ashit-rath: Here the conversion implies to the conversion of keys not the computation of `formData`.

For our case we would always want our `formData` to be converted to the right keys so we shouldn't skip this otherwise we would save the incorrect structure of the `formData`.

Please revert this to `this.updateFormData(formData)`

For more information, you can check the use of `fromId`,`toId` and it's implications.

ashit-rath (approved):

chandannkumar:
Observation: cc: @ashit-rath 
- Resetting JSONForm widget twice, the formData gets empty
https://bthrujcsw8.vmaker.com/record/HApBAEIBwmz99jOZ

chandannkumar:
Tested this PR and LGTM
- Added toggle property in property pane to use Source data's hidden fields as fallback

PR #18824, opened 2022-12-09
test: Updated test for list widget

NandanAnantharamu:
Scenario's covered

- Widgets within list widget have no dynamic height
- Widgets within list widget in existing applications have no dynamic height
- Widgets when moved out of the list widget have dynamic height in property pane
- Widgets when moved into the list widget have no dynamic height
- Widgets which have dynamic height outside the list widget when moved into the list widget should no longer have dynamic height
- Widgets when copied and pasted into the list widget no longer have dynamic height
- Widgets when cut and pasted into the list widget no longer have dynamic height
- Widgets which were not possible to be added to list widget cannot be pasted/moved into the list widget.


app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/DynamicHeight/DynamicHeight_List_TextWidget_Spec.js
```diff
@@ -0,0 +1,114 @@
...
+
+describe("Dynamic Height Width validation list widget", function () {
+    afterEach(() => {
+        agHelper.SaveLocalStorageCache();
+    });
+
+    beforeEach(() => {
+        agHelper.RestoreLocalStorageCache();
+    });
+    it("Validate change with auto height width for list widgets", function () {
```
riodeuno, line 16: @NandanAnantharamu Let's add comments here where we describe what we're checking. The code is difficult to read for me.

riodeuno (changes requested):

app/client/cypress/integration/Smoke_TestSuite/ClientSideTests/DynamicHeight/DynamicHeight_List_TextWidget_Spec.js
NandanAnantharamu: I have added comments ,one thing drag and drop usecase were not possible so we have added tests using copy paste feature

rahulramesha (approved):

riodeuno (approved):

PR #18858, opened 2022-12-12
chore: Property pane search input focus

### LINKED ISSUE #18847: [Task]: Focus on search input when an existing widget on canvas is selected

### SubTasks

We were focusing on the search input when an existing widget is selected on the canvas.
This created some conflicting bugs with context preservation project. The test `/ClientSideTests/IDE/Canvas_Context_Property_Pane_spec.js` failed due to this.

I commented out the relevant code path in the PR #18164 

We need to identify the solution to this problem and ship this feature again.

aswathkk:
## Description

When a widget is opened for the first time, the focus will be on the property pane search.

Fixes #18847

## Type of change

## How Has This Been Tested?
- Manual
- Cypress

### Dev activity

### QA activity:


app/client/src/pages/Editor/PropertyPane/PropertyPaneSearchInput.tsx
```diff
@@ -37,28 +37,26 @@ type PropertyPaneSearchInputProps = {
...
-  //       }, 300);
-  //     }
-  //   }
-  // }, [shouldFocusSearch, shouldFocusPanelSearch, props.isPanel]);
+  useEffect(() => {
+    // Checks if the property pane opened not because of focusing an input inside a widget
+    // The same functionality is being used for context preservation. Need to check if we can piggy back on that.
+    const isActiveFocusNotFromWidgetInput = !isCurrentFocusOnInput();
+    if (shouldFocusSearch && isActiveFocusNotFromWidgetInput) {
+      if (!props.isPanel && !shouldFocusPanelSearch) {
```
rahulramesha, line 48: Instead of if else with same code, can you create a constant, like focusTimeout and set that to 0 or 300 and use the same setTimeout.

rahulramesha: I was mentioning about `wrapperRef.current?.focus();`, you can have something like, `const focusTimeout = (props.isPanel && shouldFocusPanelSearch) ? 300 : 0;` instead of writing same SetTimeout twice with some if condition.

aswathkk: In that case, the focus will happen with a timeout of `0` if either one of `props.isPanel` or `shouldFocusPanelSearch` is false.
I don't want to do that since it creates UI flickering.
I wanted to focus only if either both of `props.isPanel` and `shouldFocusPanelSearch` are true or both of them are false.

aswathkk: Final solution that works would be like 

```javascript
const focusTimeout = props.isPanel && shouldFocusPanelSearch ? 300 : 0;
if (
  (!props.isPanel && !shouldFocusPanelSearch) ||
  (props.isPanel && shouldFocusPanelSearch)
) {
  setTimeout(() => {
    wrapperRef.current?.focus();
  }, focusTimeout);
}
```
which is less readable than
```javascript
if (!props.isPanel && !shouldFocusPanelSearch) {
  setTimeout(() => {
    wrapperRef.current?.focus();
  });
} else if (props.isPanel && shouldFocusPanelSearch) {
  // Layered panels like Column Panel's transition takes 300ms
  setTimeout(() => {
    wrapperRef.current?.focus();
  }, 300);
}      
```

Which one you find more readable, @rahulramesha?

rahulramesha: I was looking at why `shouldFocusPanelSearch` was required, it seems like the logic is a bit incorrect for that particular selector. Instead of `selectedPropertyPanel` and you will have to check if the `FocusedProperty` string contains the `panelPropertyPath` this can be passed from `panelPropertiesEditor`. Also the if condition with just `!props.panel ||   shouldFocusPanelSearch` should work fine as well. Please check

rahulramesha (dismissed):

rahulramesha (approved):

PR #18873, opened 2022-12-12
ci: Add replicate set to the mongodb url for transaction PR

AnaghHegde:
This PR adds replicaSet to the mongodb url used in the CI. This is to support the mongoDB transaction PR - https://github.com/appsmithorg/appsmith/pull/14939

### DIFF
--- a/.github/workflows/test-build-docker-image.yml
+++ b/.github/workflows/test-build-docker-image.yml
@@ -504,7 +504,7 @@ jobs:
         if: steps.run_result.outputs.run_result != 'success'
         working-directory: app/server
         env:
-          APPSMITH_MONGODB_URI: "mongodb://localhost:27017/mobtools"
+          APPSMITH_MONGODB_URI: "mongodb://localhost:27017/mobtools?replicaSet=mobtools-test-cluster-shard-0"
           APPSMITH_REDIS_URL: "redis://127.0.0.1:6379"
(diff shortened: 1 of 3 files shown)


mohanarpit:
How does this affect our users running external MongoDB clusters?

.github/workflows/test-build-docker-image.yml
```diff
@@ -504,7 +504,7 @@ jobs:
         if: steps.run_result.outputs.run_result != 'success'
         working-directory: app/server
         env:
-          APPSMITH_MONGODB_URI: "mongodb://localhost:27017/mobtools"
+          APPSMITH_MONGODB_URI: "mongodb://localhost:27017/mobtools?replicaSet=mobtools-test-cluster-shard-0"
```
mohanarpit, line 507: Can you also make this change in `integration-tests-command.yml` (for `ok-to-test`), `perf-test.yml` (for performance infra) and `server-build.yml` (for JUnit)?

AnaghHegde: We do not need it for `server-build.yml` because it is already handled as the tests runs in JUnit container. I have verified it in the transactions PR. Server tests runs fine without this change.

AnaghHegde:
@mohanarpit This would not make any difference from what i have read. But i will test this scenario once before merging the transactions PR.

mohanarpit (approved):

Issue #18228, opened 2022-11-17
[Task]: [Custom JS Lib] Enable git sync flow with custom JS lib.


sumitsum:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SubTasks

- Ref: https://github.com/appsmithorg/appsmith/pull/18219

subrata71:
While communicating with the remote repository it succeeds for the first time only. In the subsequent communication, it fails to communicate with the remote repository. It fails with a message `"....Invalid Private Key....."`. 

### RCA
The issue is kind of sneaky. The private key was getting changed at some point. The private key is required to communicate with the remote repository and it's stored in DB being encrypted. There's an event listener which listens on saving documents into the DB event. It encrypts any field in the document if it's annotated with `@Encrypted` annotation.

The `Application.java` class contains this private key in a nested field. The application object was saved twice in two different areas of the codebase and that resulted in the double encryption of the private key.


### Solution
Converting the second save operation to an update solves the issue.

Issue #18227, opened 2022-11-17
[Task]: [Custom JS Lib] Enable import / export flow.


sumitsum:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SubTasks

Ref: https://github.com/appsmithorg/appsmith/pull/18219

Issue #18226, opened 2022-11-17
[Task]: [Custom JS Lib]: Modify `add` / `remove` endpoints to update data directly on DB. 


sumitsum:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SubTasks

Ref: https://github.com/appsmithorg/appsmith/pull/18219
May have to update the comment in class `CustomJSLib.java`

<img src="https://front.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_7eqdo)

Issue #18159, opened 2022-11-14
[Task]: [Custom JS Lib] Add JUnit TC for endpoints


sumitsum:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SubTasks

- Add JUnit TC for endpoints.

PR #18904, opened 2022-12-12
feat: add endpoints for custom JS library add / remove / get

### LINKED ISSUE #18155: [Feature]: [Custom JS Lib] Create empty controller

### Summary

Create a controller class so that the following endpoints can added later - add, delete, fetchAll.

To enable custom JS lib feature.

sumitsum:
## Description
- Please note that this PR `WILL NOT` enable the custom JS lib feature on its own. The client side changes are still pending. 
  - Client side code will be merged in separately. The server side changes in this PR is a blocker to bring in the client side changes hence merging it first. 
- The PR adds REST API endpoints to enable the following:
  - add a new custom JS library to application (`PATCH` request)
  - remove a previously added custom JS library from application (`PATCH` request)
  - get a list of all installed custom JS libraries in an application (`GET` request)
- List of custom installed JS libraries is added to the `Application` class. 
- This PR introduces new `controller` and `repository` classes for custom JS lib related operations.
- Instead of fetching / updating the entire `Application` object, only the custom JS lib related fields are fetched / updated.
- It also adds support for import - export flow and git sycn flow. 
- A feature flag has been introduced which has email based policy. Currently, server-side code is wrapped in this feature flag.
- Please note that the JUnit test cases will be added separately. It is [tracked here](https://github.com/appsmithorg/appsmith/issues/18159).
- Further enhancement tasks (non blockers): 
  - check if `add` or `remove` end point implementations can directly update the DB array instead of fetching the array, updating them and writing back. Otherwise, wrap the function in a lock so that simultaneous updates don't cause inconsistency. It is [tracked here](https://github.com/appsmithorg/appsmith/issues/18226). 

Fixes #18155 #18156 #18157 #18158 #18227 #18228

## Type of change

## How Has This Been Tested?
- Manual
- JUnit TC will be picked up ASAP. It is [tracked here](https://github.com/appsmithorg/appsmith/issues/18159).

### Dev activity

### QA activity:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
@@ -21,6 +21,7 @@ import com.appsmith.server.domains.ActionCollection;
 import com.appsmith.server.domains.Application;
 import com.appsmith.server.domains.ApplicationPage;
 import com.appsmith.server.domains.Layout;
+import com.appsmith.server.domains.CustomJSLib;
 import com.appsmith.server.domains.NewAction;
 import com.appsmith.server.domains.NewPage;
 import com.appsmith.server.domains.Theme;
@@ -31,6 +32,7 @@ import com.appsmith.server.dtos.ActionCollectionDTO;
 import com.appsmith.external.models.ActionDTO;
 import com.appsmith.server.dtos.ApplicationImportDTO;
 import com.appsmith.server.dtos.ApplicationJson;
+import com.appsmith.server.dtos.CustomJSLibApplicationDTO;
 import com.appsmith.server.dtos.ExportFileDTO;
 import com.appsmith.server.dtos.PageDTO;
 import com.appsmith.server.exceptions.AppsmithError;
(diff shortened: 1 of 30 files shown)


nidhi-nair (dismissed):

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -718,7 +718,8 @@ private Mono<Application> importApplicationInWorkspace(String workspaceId,
         List<NewPage> importedNewPageList = importedDoc.getPageList();
         List<NewAction> importedNewActionList = importedDoc.getActionList();
         List<ActionCollection> importedActionCollectionList = importedDoc.getActionCollectionList();
-        List<CustomJSLib> customJSLibs = importedDoc.getCustomJSLibList();
+        List<CustomJSLib> customJSLibs = importedDoc.getCustomJSLibList() == null ? new ArrayList<>() :
```
nidhi-nair, line 721: For my understanding, when would this be null?

sumitsum: @nidhi-nair this is to handle for the case when any exported data (that gets imported) has customJSLibs value set to `null` instead of empty list. Ideally this is never expected to happen, but in case it does happen for any reason this would handle for that case instead of crashing.

sumitsum: One potential use case would be when a user tries to import an older application that did not have custom JS libs introduced then.

nidhi-nair (approved):

PR #18968, opened 2022-12-15
test: add common test methods for RBAC

Parthvi12:
Adds common test methods that are to be used in RBAC and fix flaky test

### DIFF
--- a/app/client/cypress/support/index.js
+++ b/app/client/cypress/support/index.js
@@ -30,7 +30,6 @@ import "./queryCommands";
 import "./widgetCommands";
 import "./themeCommands";
 import "./AdminSettingsCommands";
-import "./RBACCommands";
 /// <reference types="cypress-xpath" />
 
 Cypress.on("uncaught:exception", () => {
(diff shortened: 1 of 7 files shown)


app/client/cypress/support/index.js
```diff
@@ -30,6 +30,7 @@ import "./queryCommands";
 import "./widgetCommands";
 import "./themeCommands";
 import "./AdminSettingsCommands";
+import "./RBACCommands";
```
ankitakinger, line 33: this RBACCommands file won't exist on CE right? This change is not required on CE.

Parthvi12: But we will use it in EE , adding it here to avoid merge conflicts in future.

ankitakinger: But this file doesn't exist in CE. This import line will lead to an error. We have done the same for OIDCCommands & SAMLCommands as well.

Parthvi12: Okay removing it, thanks.

ankitakinger (approved):

PR #19001, opened 2022-12-16
fix: auto height limits one row difference

### LINKED ISSUE #19079: [Feature]: Keep a buffer of min 1 row between min/max height in auto-height with limits

### Summary

Keep a buffer of min 1 row between min/max height in auto-height with limits

Keep a buffer of min 1 row between min/max height in auto-height with limits so that the min/max handles never collide

ankurrsinghal:
## Description

Issue:-
Currently, the min and max limits handle in Auto Height with Limits move with each other on collision, while we can change the max when they have the same values but we cannot change min because the max handle is on top of it so they move together if we want to decrease the max.

Solution:-
We added a one-row difference between the two. When you increase the min towards the max, the max will increase but with one row extra, similarly, the min will decrease with one row extra when the max is decreased towards the min.

Fixes #19079

Media

## Type of change

## How Has This Been Tested?

- Manual

### Dev activity

### QA activity:


app/client/src/components/autoHeightOverlay/index.tsx
```diff
@@ -219,13 +235,21 @@ const AutoHeightOverlay: React.FC<AutoHeightOverlayProps> = memo(
 
     function onMaxStop() {
```
riodeuno, line 236: @ankurrsinghal Let us avoid defining functions within a component or another function. This is an anti-pattern and can cause performance issues.

riodeuno (changes requested):
@ankurrsinghal A couple of requests to change. I haven't seen the logic in `autoHeightOverlay/index.tsx`. 

The `BaseWidget.tsx` changes look fine to me.

app/client/src/components/autoHeightOverlay/index.tsx
ankurrsinghal: Wanted to do this for a long time, to slice up this file into smaller components and use useCallback for the child props. Made those changes.

app/client/src/components/autoHeightOverlay/index.tsx
```diff
@@ -51,7 +51,7 @@ interface MinMaxHeightProps {
 interface AutoHeightOverlayContainerProps
   extends MinMaxHeightProps,
     WidgetProps {
-  batchUpdate: (height: number) => void;
+  batchUpdate: (options: { minHeight?: number; maxHeight?: number }) => void;
```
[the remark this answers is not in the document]
ankurrsinghal: Added!

riodeuno (approved):

kamakshibhat-appsmith:
Looks Good  @ankurrsinghal

PR #19024, opened 2022-12-19
feat: parent message listener

### LINKED ISSUE #15538: [Feature]: Appsmith app should be able to listen to messages sent from the parent webpage
When appsmith is embedded as an iframe, listen to messages from parent window.

JS equivalent -> `window.addEventListener('message', (ev) => {})`.

## Method signature

- Subscriptions to be unsubscribed:
  - On page change, because the callback can have references to entities from that page
  - On app switch/exit, because we want to restrict listening to app's only.

- `subscribeParentMessages`
```
subscribeParentMessages(
	"https://eco-monk.github.io", 
	(message) => { showAlert(message) }
)
```

- `unsubscribeParentMessages`
```
unsubscribeParentMessages("https://eco-monk.github.io")
```

---

Todo:
 - [x] decide on UI (Discussed: this will be handled purely in code for now) @kocharrahul7 @RoopKrrish9696 
 - [x] investigate and handle security concerns (we're good with adding domain checks)
 - [x] handle page switch unsubscribe
 - [x] unsubscribe and subscribe -> params remain the same
 - [x] No action creator -> since it's not a widely used method

eco-monk:
## Description
When appsmith is embedded as an iframe, listen to messages from parent window.

Fixes #15538

## Global methods added 
- (Note: "*" subscribes or unsubscribes all domains)
- Subscriptions are unsubscribed automatically on page change or when switching/exiting the app
  - On page change, because the callback can have references to entities from that page
  - On app switch/exit, because we want to restrict listening to app's only.

- `subscribeParentMessages`
```
subscribeParentMessages(
	"https://eco-monk.github.io", 
	(message) => { showAlert(message) }
)
```

- `unsubscribeParentMessages`
```
unsubscribeParentMessages("https://eco-monk.github.io")
```

## Misc
- Moved `routeChanged` dispatch to app level. (previously it was only for editor)
  - Added conditions in `NavigationSagas` -> `handleRouteChange`, so previous logic is only executed for editor based navigations.

## Type of change

## How Has This Been Tested?

Manual:
- Test by calling multiple subscribes and unsubscribes for same website (error message shown)
- Test with different origins (messages received only for subscribed origins)

## Checklist:


app/client/src/AppRouter.tsx
```diff
@@ -101,6 +102,7 @@ function AppRouter(props: {
           </>
         ) : (
           <>
+            <RouteChangeListener />
```
akash-codemonk, line 105: I am guessing this is required in edit and view mode of an application. Should this be moved to AppViewer/index.tsx to not have to go through this code flow outside an application?

akash-codemonk (changes requested):

app/client/src/AppRouter.tsx
eco-monk: Yes, I started out with something like that. But ended up here for two reasons.
- The redux event `ROUTE_CHANGED` becomes generic this way (which is implied in it's name) and `NavigationSaga` takes care of filtering out events according to it's needs.
- We want this event to be triggered when AppViewer and Editor is offloaded as well. 
e.g. when user exits an app (AppViewer/Editor is offloaded), we want to clear all subscriptions.

PR #19029, opened 2022-12-19
chore: restrict when /usage-pulse can be fired

### LINKED ISSUE #19025: [Task]: remove /usage-pulse from /applications page

### SubTasks

As of today when you load app.appsmith.com, a /usage-pulse is fired, that should not happen.

IAmAnubhavSaini:
## Description

Restricts /usage-pulse call only to urls containing "/app/". 

`sendPulse` fires a call to /usage-pulse endpoint in the server. Tracking metions of sendPulse:

1. [client/public/logger.js#L84 DOMContentLoaded](https://github.com/appsmithorg/appsmith/blob/release/app/client/public/logger.js#L84)
2. [client/public/logger.js#L44 pointerDown](https://github.com/appsmithorg/appsmith/blob/release/app/client/public/logger.js#L44)
3. Here we remove the event listener added on the line 44 (above). [client/public/logger.js#L47](https://github.com/appsmithorg/appsmith/blob/release/app/client/public/logger.js#L47)

First two result in calling `punchIn`, which calls `sendPulse` [here](https://github.com/appsmithorg/appsmith/blob/release/app/client/public/logger.js#L80)

It is defined [here](https://github.com/appsmithorg/appsmith/blob/release/app/client/public/logger.js#L27-L35)

sendPulse internally calls `sendHTTPPulse` [here](https://github.com/appsmithorg/appsmith/blob/release/app/client/public/logger.js#L27-L35)

It is just a POST `fetch` request.

Problematic /usage-pulse

1. login page
2. signup page
3. applications page

Instead of a blocklist, I am going with an allowlist that contains only one input: `/app/` in url.

Fixes #19025 

## Type of change

- Chore

## How Has This Been Tested?

- Manually

### Dev activity

### QA activity:

### DIFF
--- a/app/client/public/logger.js
+++ b/app/client/public/logger.js
@@ -15,22 +15,31 @@ function getCurrentUTCHourTimestamp() {
 
 const PULSE_API_ENDPOINT = "/api/v1/usage-pulse";
 
-// Use standard fetch to POST (fire and forget)
+/**
+ * Sends HTTP pulse to the server, when beaconAPI is not available.
+ * Fire and forget.
+ */
 function sendHTTPPulse() {
   fetch(PULSE_API_ENDPOINT, {
     method: "POST",
     credentials: "same-origin",
-  });
+  })
+    .then(() => {
+      // Fire and forget
+    })
+    .catch(() => {
+      // Ignore errors; fire and forget
+    });
 }
 
-// Use the Beacon API to POST (fire and forget)
+/**
+ * Sends a usage-pulse to the server using the Beacon API.
+ * If the Beacon API is not available, falls back to a standard fetch.
+ * Note: Only sends pulse when user is on "/app/" pages: editor and viewer.
+ */
 function sendPulse() {
-  const url = PULSE_API_ENDPOINT;
-
-  // In case the beacon fails to queue, use HTTPRequest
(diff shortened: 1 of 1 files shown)


app/client/public/logger.js
```diff
@@ -15,22 +15,31 @@ function getCurrentUTCHourTimestamp() {
...
+ * Sends HTTP pulse to the server, when beaconAPI is not available.
+ * Fire and forget.
+ */
 function sendHTTPPulse() {
   fetch(PULSE_API_ENDPOINT, {
     method: "POST",
     credentials: "same-origin",
-  });
+  })
+    .then(() => {
```
riodeuno, line 27: @IAmAnubhavSaini Could you explain the reason for adding the then and catch handlers here? Are we planning to add logic in here? If not, does it make a difference if we have them or not?

riodeuno (approved):

app/client/public/logger.js
IAmAnubhavSaini: Catch because I don't like dangling promises, and not catching a promise throws an error.
Then, `then` for completeness-sake. 

Reading this code tells me that it is complete and that developer didn't forget or missed then/catch and that the code is written/working as expected. Think documentation.

I do agree that these are not really required.

PR #19056, opened 2022-12-20
chore: table widget cell tooltip should be calculated only on hover

### LINKED ISSUE #18995: [Task]: Table widget cell tooltip should be initialised only when a cell is hovered

### SubTasks

- [ ] Currently, the cell tooltip-related events are registered when the table widget is mounted on the UI, as this is degrading the performance on large tables, we should move event registration to the on hover of the cell.

sbalaji1192:
## Description
Prior to this PR, we were calculating whether a cell has overflowing content or not when the table gets loaded, to show a tooltip on hover. But this is affecting the performance of the table, now we're calculating for a cell only when the user hovers the cell. 

Fixes #18995

## Type of change

- chore

## How Has This Been Tested?

- Manual
- Cypress

### Dev activity

### QA activity:


kamakshibhat-appsmith:
Observations :
1. Tooltips are not getting dissmissed in the first run 
https://www.loom.com/share/3f2d7a7116394b2e80ef62bb66b73343

app/client/src/widgets/TableWidgetV2/component/cellComponents/AutoToolTipComponent.tsx
```diff
@@ -29,40 +29,63 @@ export const Content = styled.span`
...
 
-    if (element && element.offsetWidth < element.scrollWidth) {
-      updateToolTip(true);
-    } else {
-      updateToolTip(false);
-    }
+    const mouseEnterHandler = () => {
+      const element = ref.current?.querySelector("div") as HTMLDivElement;
+
+      timeout = setTimeout(() => {
```
ashit-rath, line 48: Q
1. Why are we using `setTimeout` here?
2. Why are we removing the event listener in line 56
3. Is there any benefit of  using ref and native events  over using `onMouseEnter` and `onMouseLeave` to the `Content` component directly?

app/client/src/widgets/TableWidgetV2/component/cellComponents/AutoToolTipComponent.tsx
```diff
@@ -29,40 +29,63 @@ export const Content = styled.span`
...
   useEffect(() => {
-    const element = ref.current?.querySelector("div") as HTMLDivElement;
+    let timeout: number;
 
-    if (element && element.offsetWidth < element.scrollWidth) {
-      updateToolTip(true);
-    } else {
-      updateToolTip(false);
-    }
+    const mouseEnterHandler = () => {
```
ashit-rath, line 45: If I move my mouse very rapidly, would that make this `handler` trigger multiple times?

app/client/src/widgets/TableWidgetV2/component/cellComponents/AutoToolTipComponent.tsx
sbalaji1192: 1. The existing tooltip has [hoverOpenDelay](https://blueprintjs.com/docs/#core/components/tooltip), so we are simulating the same during the first render.
2. We're using the mouse enter and mouse leave to wrap the content with a tooltip, once wrapped we don't need these event handlers.
3. Since we're removing the event listeners we cannot use `onMouseEnter` and `onMouseLeave`

app/client/src/widgets/TableWidgetV2/component/cellComponents/AutoToolTipComponent.tsx
sbalaji1192: It will get triggered once for each mouse enter event.

laveena-en:
@sbalaji1192 looks like tooltips aren’t being dismissed intermittently. Can you please have a look?
 
https://ivP6ez0lJB.vmaker.com/record/XF7sj0CnNJGoCdxz

laveena-en:
Re-tested and LGTM.

ashit-rath (approved):

PR #19066, opened 2022-12-20
ci: Migrate runners from buildjet to github

pratapaprasanna:


.github/workflows/client-build.yml
```diff
@@ -23,7 +23,7 @@ defaults:
 
 jobs:
   build:
-    runs-on: buildjet-8vcpu-ubuntu-2004
+    runs-on: ubuntu-latest-4-cores
```
mohanarpit, line 26: Do we want this to be 4 cores or 8 cores on Github large runners? I'm asking because on Buildjet, we were using the 8 core option.

.github/workflows/server-build.yml
```diff
@@ -23,7 +23,7 @@ defaults:
 
 jobs:
   build:
-    runs-on: buildjet-8vcpu-ubuntu-2004
+    runs-on: ubuntu-latest
```
mohanarpit, line 26: Why are we using the default runner here instead of the 4 core or 8 core machine?

.github/workflows/client-build.yml
pratapaprasanna: https://github.com/appsmithorg/appsmith-ci/actions/runs/3739127454
Here we could run on it on 4-cores. So, didnt want to use an un-necessary resource

.github/workflows/server-build.yml
pratapaprasanna: We could https://github.com/appsmithorg/appsmith-ci/actions/runs/3739357063/jobs/6346468632
run it on a default runner successfully, So didn't want to use unnecessary resources

mohanarpit: We can run it on the default runner. But the time it takes for this job to complete on a default runner is approx 45 mins. On an 8-core machine, this number drops to 16 mins. Check https://github.com/appsmithorg/appsmith/actions/runs/3734511193 for a successful run. Hence, I'd highly recommend keeping this job on 8-core machines.

pratapaprasanna: On a default-one
[ubuntu-latest](https://github.com/appsmithorg/appsmith-ci/actions/runs/3739357063/jobs/6346468632) : 28 min
[ubuntu-latest-4-cores]( https://github.com/appsmithorg/appsmith-ci/actions/runs/3739508417) :  
[ubuntu-latest-8-cores](https://github.com/appsmithorg/appsmith-ci/actions/runs/3739683641/jobs/6347180095):

PR #19090, opened 2022-12-21
chore: route change listener

eco-monk:
Related to https://github.com/appsmithorg/appsmith/issues/15538
Need to listen to route change outside editor for that feature.

## Description
Made `ROUTE_CHANGE` redux event to be triggered on every route change.
Previously it was just for editor route changes. (Not for app viewer, home page etc...)

Made sure existing methods called on `NavigationSaga` -> `handleRouteChange` listen only to editor route events, as it was before.

## Type of change

## How Has This Been Tested?
- Existing cypress tests should pass, since there are no functional changes.
- Manually verified events are triggered

### Dev activity

### QA activity:


akash-codemonk (approved):
LGTM. Adding rahul to review as well.

app/client/src/RouteChangeListener.tsx
```diff
@@ -0,0 +1,16 @@
+import { routeChanged } from "actions/focusHistoryActions";
+import React from "react";
+import { useEffect } from "react";
+import { useDispatch } from "react-redux";
+import { useLocation } from "react-router-dom";
+import { AppsmithLocationState } from "utils/history";
+
+export default function RouteChangeListener() {
```
rahulramesha, line 8: Any reason this is an empty component instead of custom Hook?

rahulramesha (commented):
Most of the code looks good to me, just one query

app/client/src/RouteChangeListener.tsx
eco-monk: That's because of `useLocation` hook we're using from `react-router-dom`.
It needs to be used inside the `Router` element. And we only have <AppHeader /> component inside the <Router>.

Which is not an ideal place to do the route change listening.
Hence, I had created this component.

eco-monk:
`Fork_Template_Existing_app_spec.js` passes locally.
image

app/client/src/RouteChangeListener.tsx
```diff
@@ -0,0 +1,16 @@
...
+  const location = useLocation<AppsmithLocationState>();
+  const dispatch = useDispatch();
+  useEffect(() => {
+    dispatch(routeChanged(location));
+  }, [location.pathname, location.hash]);
+  // eslint-disable-next-line react/jsx-no-useless-fragment
+  return <></>;
```
hetunandu, line 15: Return Null instead 
```suggestion
  return null;
```

Explanation
https://reactjs.org/docs/jsx-in-depth.html#booleans-null-and-undefined-are-ignored

akash-codemonk (dismissed):

hetunandu (approved):

PR #19122, opened 2022-12-22
chore: send segment anonymous id

eco-monk:
Add segment's `anonymousId` as a header in all API calls.

cached id -> [details](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/identity/#segment-id-persistence)

On Page load actions:
- If segment is enabled:
  - and cached id exists -> trigger with cached id
  - if cached id doesn’t exist, we wait for max 2 seconds.
    - if segment init is success -> trigger with anonymous id
    - if failed/delayed -> trigger without anonymous id
- If segment is disabled we don’t wait at all and anonymous id is not sent.


SatishGandham:
/perf-test
Just testing perf infra

app/client/src/api/ApiUtils.ts
```diff
@@ -49,6 +50,12 @@ export const apiRequestInterceptor = (config: AxiosRequestConfig) => {
     config.timeout = 1000 * 120; // increase timeout for git specific APIs
   }
 
+  if (config.headers) {
```
Nikhil-Nandagopal, line 53: @eco-monk if config.headers is not there then we should add it instead of not sending the values right?

app/client/src/utils/AnalyticsUtil.tsx
```diff
@@ -477,6 +477,16 @@ class AnalyticsUtil {
     }
   }
 
+  static getAnonymousId() {
+    const windowDoc: any = window;
+    const { segment } = getAppsmithConfigs();
+    return (
+      windowDoc.analytics?.user?.().anonymousId() ||
```
Nikhil-Nandagopal, line 484: This is kind of unreadable. Can you split this and make it simple for anyone else to understand?

Nikhil-Nandagopal (commented):
@eco-monk left my comments

app/client/src/api/ApiUtils.ts
eco-monk: Yep, that's right.

app/client/src/utils/AnalyticsUtil.tsx
eco-monk: Done.

app/client/src/api/ApiUtils.ts
eco-monk: Done

hetunandu (changes requested):
Looks good. Just remove the comments

eco-monk:
Thanks @hetunandu.!
Will keep the logs for now, to test this.
Will remove once done.

Issue #14498, opened 2022-06-13
[Feature]: Make it easier to resolve the merge conflicts when user updates the any resource


abhvsn:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

As per current DB structure actions are embedded inside the JSObjects. Every new method in JSObjects results in new action getting created in DB. Consider user updates the JSObject.myFun1 then user will have changes at 2 locations
- Queries directory under myFun1 file
- Body of JSObject1
We should not commit these duplicates and JSObject body should be committed as is. These steps surely makes it easier to resolve the conflicts even if occurs at same line.

Steps to resolve this issue:
- Don’t commit action body to queries/JSObject1-myFun1.json as this is not adding any value
- Keep the text as is and avoid stringifying the body of JSObject

Discussion thread: https://theappsmith.slack.com/archives/C02K2MZERSL/p1655105041401639

### Why should this be worked on?

This will enhance the merge conflict resolution which occurs for every JSObject update by quite a bit, and also it will reduce the number of occurrences to end up in these conflicts.

- [x] #18514
- [x] #18515
- [x] #18516
- [x] https://github.com/appsmithorg/appsmith/issues/19791

<img src="https://front.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_6bu18)

abhvsn:
@Nikhil-Nandagopal I was digging through this and wanted to comeup with alternate approach as newline is not supported by JSON:
https://stackoverflow.com/a/52557585/13092339

Considering the current limitations we can split the string into array and then serialize. The major issue here is we end up attaching `""` for each line. 
Output for ref:
```
 "body": [
      "export default {",
      "	myVar1: [],",
      "	myVar2: {},",
      "	myFun1: () => {",
      "		return \"appsmith\";",
      "	},",
      "	myFun2: async () => {",
      "		//use async-await or promises",
      "	}",
      "}"
    ],
```

Will this be good enough? I'm looking for more ideas here.

abhvsn:
As per the discussion we have decided to use js files to store the code present in the JSObject.

Gabriel-Azevedo:
Storing JSObjects as JS files would significantly improve the code review experience. 

My biggest problem as a tech lead is that pull request reviews for Appsmith basically became auto-approves to the point that we turned off the approval requirement to merge.
[...]

PR #19156, opened 2022-12-23
feat: Separate the js object code and metadata when used with git sync

AnaghHegde:
## Description

> 
As per current DB structure actions are embedded inside the JSObjects. Every new method in JSObjects results in new action getting created in DB. Consider user updates the JSObject.myFun1 then user will have changes at 2 locations

Queries directory under myFun1 file
Body of JSObject1
We should not commit these duplicates and JSObject body should be committed as is. These steps surely makes it easier to resolve the conflicts even if occurs at same line.

Before this change - 
image

After this PR
image

Fixes https://github.com/appsmithorg/appsmith/issues/19791
https://github.com/appsmithorg/appsmith/issues/14498

## Type of change

## How Has This Been Tested?

- Manual
- JUnit

### Dev activity

### QA activity:


abhvsn:
@AnaghHegde any specific reason we are moving the models and constants to interface from the server? Asking this because we don't want to keep the appsmith-git module as loosely coupled as possible. 
This change seems to be too big right now to review!

app/server/appsmith-git/src/main/java/com/appsmith/git/helpers/FileUtilsImpl.java
```diff
@@ -287,19 +288,58 @@ public Mono<Path> saveApplicationToGitRepo(Path baseRepoSuffix,
...
+    }
+
+    private boolean saveJsObject(Object sourceEntity, Path path, Gson gson) {
+        try {
+            Files.createDirectories(path);
+            JsonObject jsonObject = gson.fromJson(gson.toJson(sourceEntity), JsonObject.class);
+
+            // Get the js Object code from the action collection
+            JsonObject unPublishedCollectionObject = jsonObject.get("unpublishedCollection").getAsJsonObject();
+            String body = unPublishedCollectionObject.get("body").getAsString();
```
abhvsn, line 308: As we have created a field `actionCollectionBody` in `ApplicationGitReference` model it will be better to extract body field from unpublishedCollection in `gitFileUtils.createApplicationReference`, this way it will be future proof if we decide to change anything around body field.

abhvsn (commented):
LGTM 
If we are also updating the structure for actions I propose to make that change in this PR itself to avoid adding another file format migration 🙂

app/server/appsmith-git/src/main/java/com/appsmith/git/helpers/FileUtilsImpl.java
```diff
@@ -502,6 +542,42 @@ private Map<String, Object> readFiles(Path directoryPath, Gson gson, String keyS
...
+            data = FileUtils.readFileToString(filePath.toFile(), "UTF-8");
+        } catch (IOException e) {
+            log.error(" Error while reading the file from git repo {0} ", e.getMessage());
+        }
+        return data;
+    }
+
+    /**
+     * This method is to read the content for action and actionCollection or any nested resources which has the new structure - v3
+     * Where the user queries and the metadata is split into teo different files
```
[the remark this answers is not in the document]
AnaghHegde: No we are not doing that now.

app/server/appsmith-git/src/main/java/com/appsmith/git/helpers/FileUtilsImpl.java
```diff
@@ -121,6 +121,30 @@ public class FileUtilsImpl implements FileInterface {
...
+        theme
+        publishedTheme.json
+        editModeTheme.json
+     pages
+        page1
+            canvas.json
+            queries
+                Query1
+                    Query1.txt
+                    metadata.json
```
abhvsn, line 138: If this is not happening in v3 let's remove this because it may create confusion in future.

AnaghHegde: Sorry my bad for giving wrong idea earlier. It is part of v3, i am just testing few cases and will push the changes in sometime.

abhvsn (approved):
Minor comment around the v3 file format described in the comment. We can merge this otherwise.

app/server/appsmith-git/src/main/java/com/appsmith/git/helpers/FileUtilsImpl.java
```diff
@@ -502,6 +567,42 @@ private Map<String, Object> readFiles(Path directoryPath, Gson gson, String keyS
         return resource;
     }
 
+    /**
+     * This method will read the content of the file as a plain text and does not apply the gson to json transformation
+     * @param filePath file path for files on which read operation will be performed
+     * @return content of the file in the path
+     */
+    private String readFileAsString(Path filePath) {
```
sidhantgoel, line 575: We can use this method from Apache

AnaghHegde: I am using the `FileUtils.readFileToString` from apache. I am not sure if i understand what you are referring here.

AnaghHegde:
@Parthvi12 Pushed a fix for the issue - https://github.com/appsmithorg/appsmith/issues/19734. Please verify the change once and let me know.

app/server/appsmith-git/src/main/java/com/appsmith/git/helpers/FileUtilsImpl.java
sidhantgoel: I meant we don't need this function, we can call directly FileUtils.readFileToString instead of calling this function, also we shouldn't consume the exception silently and return empty data, but the caller has to handle it in the right way, exception should be supplied to Nono.error if we can't continue after this error.

Parthvi12:
@AnaghHegde The above mentioned issue https://github.com/appsmithorg/appsmith/issues/19734 is fixed, can you look at this issue related to PR: https://github.com/appsmithorg/appsmith/issues/19779

app/server/appsmith-git/src/main/java/com/appsmith/git/helpers/FileUtilsImpl.java
AnaghHegde: We need this function because we will end up writing the same piece of code many times. I do agree that we should pass the error instead of just logging it here. I will change the implementation so that the error is passed to the caller.

sidhantgoel (approved):

PR #19188, opened 2022-12-23
fix: Logout user, if session deserialization fails

sharat87:
We handle this in the global error handler, looking for the error message that indicates this cause, and include a cookie deletion header, so the user can recover by logging in again.


trishaanand (dismissed):

sharat87:
@nidhi-nair @trishaanand, I really wish we can handle this somewhere other than the global exception handler, but I couldn't find the place... oh, you've approved it. Okay 🤐.

trishaanand (approved):

PR #19273, opened 2022-12-28
fix: Added projection for page call that only needs application id details

nidhi-nair:
The NewPage document contains DSL details for a page as well. When we don't need this information, we can use a projection that disables this field. User was facing a problem with this call taking 12 seconds to complete with those details. (document is 1 MB in size for them)


subrata71 (approved):
Even if we add projection for that query the problem may still prevail for the areas where all the fields are required for that page and that will end up fetching the page without any projection fields. 

We will take that up as a follow-up to this fix.

PR #19296, opened 2022-12-29
fix: helm install values checks

sum35h:
## Description
- Updated ngnix as default ingress class in `values.yaml`
- Upgraded mongo's deployment image to `v5.0.12` as the previous 4.4.11 is deprecated
- Added `imagePullSecrets` to enable pulling images from private docker registries
- Added conditional to check for k8s version >=1.1.8 to handle both old and new k8s version scenarios for setting up the Ingress controller class name.

### DIFF
--- a/deploy/helm/templates/ingress.yaml
+++ b/deploy/helm/templates/ingress.yaml
@@ -45,7 +45,7 @@ spec:
     {{- end }}
   {{- end }}
   {{- if semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion }}
-  ingressClassName: {{ .Values.ingress.className }}
+  ingressClassName: {{ .Values.ingress.className |  default "nginx" }}
   {{- end }}
   rules:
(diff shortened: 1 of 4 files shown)


deploy/helm/templates/ingress.yaml
```diff
@@ -44,7 +44,9 @@ spec:
       secretName: {{ .secretName }}
     {{- end }}
   {{- end }}
+  {{- if semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion }}
   ingressClassName: {{ .Values.ingress.className }}
```
pratapaprasanna, line 48: ingressClassName: {{ .Values.ingress.className |  default "nginx" }}

pratapaprasanna (approved):

pratapaprasanna:
Please elaborate on the description and we shall merge it.

We have added `imagePullSecrets` to enable pulling images from private docker registries as well.

PR #19412, opened 2023-01-01
fix: Added Firestore pagination paths to self referencing candidates

### LINKED ISSUE #19256: [Bug]: Incorrect Cyclic Dependency

### Description

We've introduced a new cyclic dependency issue with the AST parsing. We're incorrectly detecting cyclic dependency when a query is trying to perform cursor pagination.

### Steps To Reproduce

1. Drag a table
2. Connect to firestore
3. Create a select query
4. bind the query to the table {{fetchData.data}}
5. Bind the table to the query to fetch data after {{Table1.tableData[Table1.tableData.length - 1]}}

### Public Sample App

https://release.app.appsmith.com/app/crud-app-templates/firestore-61764fbeba7e887d03bc3641/edit#8o4npzin85

### Issue video log

_No response_

### Version

Cloud

nidhi-nair:
## Description

Firestore requires cursor based pagination which means that a query that needs to be paginated will most definitely also refer to the same widget that is consuming the query data. This introduced a cyclic dependency in the dependency graph for such apps. This fix uses the concept of self referencing paths to ignore the pagination related fields in Firestore for all dependency calculations.

Fixes #19256 

## Type of change

## How Has This Been Tested?
- Manual
- ?

### Dev activity

### QA activity:


subrata71:
@nidhi-nair Can you please explain how did you come up with this solution?

```
@Override
public Set<String> getSelfReferencingDataPaths() {
    return Set.of("formData.prev.data", "formData.next.data");
}
```
To be more specific how did you identify that `formData.prev.data` and `formData.next.data` are the paths that needed to be added in the above function?

The other day I thought of adding some paths in the above-overridden method but wasn't sure what to add.

nidhi-nair:
Sure. There's 3 clues to go on here. 
1. The CRUD page template is a fully configured query and shows us all possible connections that a Firestore query can make to a Table widget. From here, we gather that the possible bindings are:
    ````
    Order By -> ["Table1.sortOrder.order", "Table1.sortOrder.column"]
    Start After -> ["Table1.tableData"]
    End Before -> ["Table1.tableData"]
    Limit Documents -> ["Table1.pageSize"]
    ````
2. The editor json configuration for Firestore tells us which paths are getting used by the plugin. From here we can map the above to the property paths we need.
    ````
    Order By -> "formData.orderBy.data"
    Start After -> "formData.next.data"
    End Before -> "formData.prev.data" 
    Limit Documents -> "formData.limitDocuments.data"
    ````
3. In the dependency graph created for on page load calculations, if you insert a debug point where the error gets generated, you'll see that state of the graph is something akin to the following diagram. This clearly shows that the loop is getting created because of the pagination properties. 
IMG_1618BF0BB28B-1

The graph is limited by paths for clarity, but the point to note here is that the other properties with references to `Table1` are noticably missing because they don't enter the `Table1.data` path in the graph and would hence not cause a conflict.

Hope this helps. :)

subrata71 (dismissed):
The explanation makes sense.

btsgh:
Brought up the branch on local and tried with the application attached in the bug. The firestore page still brings up a cyclic dependency error. 
Screenshot 2023-01-04 at 5 23 16 PM

nidhi-nair:
Looks like older apps have the pre-UQI paths in the `dynamicBindingPathList` as well. This is tech debt we should get rid off. For now I have added these paths to the ignore list as well. Will create an issue for cleaning up data and attach it here.

btsgh:
1. Tried with an imported json of the CRUD template app. The first time as soon as I open the app, I still see the cyclic dependency error and also I see that the "TypeError: Cannot read properties of undefined (reading 'length')". However, after I run the queries again, the error is no longer seen. [Confirmed this with @nidhi-nair and this is as per expectation]
2. Tried with a new application, with the steps in the bug attached to the PR. The cyclic dependency error is not seen. [The error is easily reproducible on Release on a new app also, but on this fix branch, it is not reproducible]
3. Exported and imported the new application and checked for the error - not reproducible.
4. Gave a new user permission on the new app and the imported app, and tried opening the page - works fine
5. Forked the imported app to another workspace. No cyclic dependency error. Works fine

Issue #19424, opened 2023-01-02
[Task]: Separate health check api for checking the status of Databases.


nsarupr:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SubTasks

- `/api/health` api in order to check the Status of redis and mongo databases.

PR #19425, opened 2023-01-02
feat: separate health check api for redis and mongo

nsarupr:
## Description

> Use a distinct health check api for checking the health of Redis and Mongo.

Fixes https://github.com/appsmithorg/appsmith/issues/19424

Media

## Type of change

### Dev activity

### QA activity:


trishaanand (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/HealthCheckServiceCEImpl.java
```diff
@@ -0,0 +1,57 @@
...
+    public Mono<Health> getMongoHealth() {
+        Function<TimeoutException, Throwable> healthTimeout = error -> new AppsmithException(
+                AppsmithError.HEALTHCHECK_TIMEOUT, "Mongo");
+        MongoReactiveHealthIndicator mongoReactiveHealthIndicator = new MongoReactiveHealthIndicator(reactiveMongoTemplate);
+        return mongoReactiveHealthIndicator.health().timeout(Duration.ofSeconds(1)).onErrorMap(TimeoutException.class, healthTimeout);
+    }
+
+    private boolean isUp(Health health) {
+        if (health.getStatus().equals(Status.UP))
+            return Boolean.TRUE;
```
mohanarpit, line 54: Nit: We should always add braces in if-else conditions. This future-proofs the code to silly developer errors allowing us to extend the if-condition easily.

```suggestion
        if (health.getStatus().equals(Status.UP)) {
            return Boolean.TRUE;
        }
```

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/HealthCheckServiceCEImpl.java
```diff
@@ -0,0 +1,57 @@
...
+    public Mono<Health> getRedisHealth() {
+        Function<TimeoutException, Throwable> healthTimeout = error -> new AppsmithException(
+                AppsmithError.HEALTHCHECK_TIMEOUT, "Redis");
+        RedisReactiveHealthIndicator redisReactiveHealthIndicator = new RedisReactiveHealthIndicator(reactiveRedisConnectionFactory);
+        return redisReactiveHealthIndicator.health().timeout(Duration.ofSeconds(1)).onErrorMap(TimeoutException.class, healthTimeout);
+    }
+
+    @Override
+    public Mono<Health> getMongoHealth() {
+        Function<TimeoutException, Throwable> healthTimeout = error -> new AppsmithException(
```
mohanarpit, line 46: I'm new to this syntax. Can you help me understand this?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/HealthCheckServiceCEImpl.java
```diff
@@ -0,0 +1,57 @@
...
+    private final ReactiveRedisConnectionFactory reactiveRedisConnectionFactory;
+    private final ReactiveMongoTemplate reactiveMongoTemplate;
+    public HealthCheckServiceCEImpl(ReactiveRedisConnectionFactory reactiveRedisConnectionFactory,
+                                    ReactiveMongoTemplate reactiveMongoTemplate) {
+        this.reactiveRedisConnectionFactory = reactiveRedisConnectionFactory;
+        this.reactiveMongoTemplate = reactiveMongoTemplate;
+    }
+
+    @Override
+    public Mono<String> getHealth() {
```
mohanarpit, line 31: Any particular reason we are returning a string instead of a boolean value?

mohanarpit:
@nsarupr Have we tested this endpoint without headers? Do we also have to make modifications to `SecurityConfig.java` to ensure that the health check endpoint is publicly accessible? I'm asking this because the monitoring systems & load balancers that perform this health check will not have any session.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/HealthCheckServiceCEImpl.java
nsarupr: Represents a function that accepts one argument and produces a result.
This is a functional interface whose functional method is apply(Object).

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/HealthCheckServiceCEImpl.java
nsarupr: Only reason is that we can give a descriptive response, instead of a boolean value.

nsarupr:
> @nsarupr Have we tested this endpoint without headers? Do we also have to make modifications to `SecurityConfig.java` to ensure that the health check endpoint is publicly accessible? I'm asking this because the monitoring systems & load balancers that perform this health check will not have any session.

Thanks @mohanarpit, I didn't check on that part. I have added this new URL to the list of Urls in `securityconfig` so that this can be used even without any authentication.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/HealthCheckServiceCEImpl.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    public Mono<Health> getMongoHealth() {
+        Function<TimeoutException, Throwable> healthTimeout = error -> new AppsmithException(
+                AppsmithError.HEALTHCHECK_TIMEOUT, "Mongo");
+        MongoReactiveHealthIndicator mongoReactiveHealthIndicator = new MongoReactiveHealthIndicator(reactiveMongoTemplate);
+        return mongoReactiveHealthIndicator.health().timeout(Duration.ofSeconds(1)).onErrorMap(TimeoutException.class, healthTimeout);
+    }
+
+    private boolean isUp(Health health) {
+        if (health.getStatus().equals(Status.UP)) {
```
mohanarpit, line 53: Sorry, another nit: 
```suggestion
        if (Status.UP.equals(health.getStatus())) {
```

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/HealthCheckServiceCEImpl.java
mohanarpit: If that is the case, I'd recommend a structure like: 
```
{
  "success": true,
  "description": "All systems are up"
}
```

The above is just a suggestion, I'd recommend taking a look at `AppsmithError` to see how the error code & description is added to all responses.

I'm suggesting this change because this structure allows other systems to be able to read the boolean value `success` and take decisions to send out alerts. We don't have to do a string match.

mohanarpit (approved):

PR #19438, opened 2023-01-02
fix: Added handling for empty evaluated values

nidhi-nair:


nidhi-nair:
Merging in without review to unblock CI, all tests have passed.

PR #19464, opened 2023-01-03
ci: Revert changes to perf-tests job

SatishGandham:
## Description
Reverts the changes made to perf job in ui-build-workflow as the perf tests are getting skipped.


mohanarpit:
Instead of reverting the changes, do you think it's possible for us to triage & arrive at a solution? I'm asking because the re-factor that was done on this workflow file is actually the direction we wish to take. Hence, I'm in favour of finding the root cause of the failure here instead of reverting the change.

mohanarpit (approved):

PR #19475, opened 2023-01-04
chore: ee clean up

eco-monk:
Related to #15538

To enable adding a new action only for EE.

Refactored `ActionTriggerType` enum to a union type.
So we can extend this with a new action in EE repo.

Made sure type discrimination is handled in `ActionExecutionSaga` properly as before.

---

- Introduced `ActionTriggerKeys` union type which can be used for type checking the values.
- Refactored `ActionDescription` types to accommodate usage of the union type instead of enum.
- exported required types for usage in EE repo.

---

strings and payload are type checked as follows.

image

image


rimildeyjsr (approved):

eco-monk:
Thanks @rimildeyjsr.!
We now know, action selector refactor doesn't get affected by this change.

We are looking for alternate ways, instead of using union type. 
We'd like to use a enum/constant type. Since using string even though they're type checked looks sketchy.

Issue #11855, opened 2022-03-15
[Epic]: Application Configuration In Database (ACID)


sharat87:
### Objective

Move application configuration from environment variables into the database. The goal is to reduce our reliance on environment variables, and use the database to store such configuration information.

### Rationale

- Environment variables are powerful. Too powerful to control. They can be set via a very many places depending on the method of deployment.
- Changing them (very usually) needs a restart. So we end up handling these changes two-fold. Update the env file, as well as make the change at runtime, if possible.
- Repeating the environment variables in server/client, causes duplication and quite a few headaches, where something is configured in the client, but not on the server, causing confusions.

### Success Metrics

| Goal          | Metric         |
| ------------- | -------------- |
| Make static assets CDN friendly  | Remove all runtime environment variables in static assets  |
| Make Admin Settings more friendly  | Handle (most, if not all) changes to configuration without needing a restart        |

### Requirements

| Requirement   | User Story    |
| ------------- | ------------- |
| An API for the client to get application configuration from the server | As a user, I should be able to host Appsmith's static assets on a CDN.  |
| Affect application configuration changes without restart | As a user, I should be able to change application configuration without needing a server restart and potential downtime.  |

### Implementation Notes

- The backend server will take the values of whatever settings it needs from the DB and use them.
- The client will request what it needs and gets their values. The backend should maintain a list of what variables can be shared to the client and to what extent.
  - By “what extent”, I mean does the client need the value of a setting, or just to see whether a setting has a non-empty value. This is useful for things like OAuth configuration, where the client doesn’t really need the key, just that it exists and is non-empty.
  - Eventually, the client shouldn’t have any runtime environment variables. All such information should be fetched from the server.
- The Admin settings will need to be wired up to apply changes in the DB, for those settings that are powered by the DB.

### Documentation

For all environment variables that this change was implemented, the documentation needs to be updated to reflect the change. Let's list them in this section as such variables come up during implementation.

### Developer Handoff Document in Figma

No UI change.

### RACI matrix

| ------------- | ------------- |
| Responsible   | @sharat87               |
| Accountable   | @sharat87               |
| Consulted     | @Nikhil-Nandagopal, @mohanarpit, @areyabhishek|
| Informed      |               |

sharat87:
Related Slack thread: <https://theappsmith.slack.com/archives/CPQNLFHTN/p1646815186755609>.

sharat87:
[...]

PR #19486, opened 2023-01-04
chore: Remove Maps API Key env variable for client

sharat87:
Part of #11855.

Instead of getting the Google Maps API Key from runtime env variables, we get it from the server, as part of the response of `/api/v1/tenant/current`. This doesn't add a database call, just include the env variable name in the response, so shouldn't have any performance impact on the API.

On the client though, the Maps API key won't be available, until at least the first call to `/tenant/current` is finished.

Also, first big PR in client code. 🙂

Edit: not `/me` anymore, but from `/tenant/current`.


app/client/src/selectors/tenantConfigurationSelectors.tsx
```diff
@@ -0,0 +1,4 @@
+import { AppState } from "@appsmith/reducers";
+
+export const getMapsApiKey = (state: AppState): string | undefined =>
+  state.tenant?.tenantConfiguration?.googleMapsKey as (string | undefined);
```
sharat87, line 4: Is this okay? I'd like if there's a better way then using an `as` here. 🙂

deploy/template/nginx_app.conf.sh
```diff
@@ -46,7 +46,6 @@ $NGINX_SSL_CMNT    server_name $custom_domain ;
         sub_filter __APPSMITH_ALGOLIA_SEARCH_INDEX_NAME__ '\${APPSMITH_ALGOLIA_SEARCH_INDEX_NAME}';
         sub_filter __APPSMITH_ALGOLIA_API_KEY__ '\${APPSMITH_ALGOLIA_API_KEY}';
         sub_filter __APPSMITH_CLIENT_LOG_LEVEL__ '\${APPSMITH_CLIENT_LOG_LEVEL}';
-        sub_filter __APPSMITH_GOOGLE_MAPS_API_KEY__ '\${APPSMITH_GOOGLE_MAPS_API_KEY}';
```
ankitakinger, line 49: This exists twice in this file. Should we remove the one at the bottom as well?

ankitakinger (commented):
There are more files where we have this ENV variable added, should we remove it from there as well?
https://github.com/appsmithorg/appsmith/blob/c4da7e718016f333941fe58a33ca53c943e970d9/deploy/template/docker.env.sh#L48-L50
https://github.com/appsmithorg/appsmith/blob/49383715664ee2731f55c95fbc561212ae0ab210/deploy/heroku/README.MD#L34
https://github.com/appsmithorg/appsmith/blob/85719acd2def393d29fe122268d8f6f2574de7a5/deploy/heroku/default.conf.template#L28
https://github.com/appsmithorg/appsmith/blob/c41096ec51c08a45c2376192c62b8985ae17f7af/deploy/heroku/bootstrap.sh#L46
https://github.com/appsmithorg/appsmith/blob/7be99c8976152c3c88f448b9bd10e0e7997e603c/deploy/helm/values.yaml#L320
https://github.com/appsmithorg/appsmith/blob/fe19dffccd0b08e46bac83f182375455cac7db2d/deploy/helm/README.md#L180
https://github.com/appsmithorg/appsmith/blob/52be06bfe9054d3ea51d6f904b555a558a347899/deploy/docker/templates/docker.env.sh#L48
https://github.com/appsmithorg/appsmith/blob/c41096ec51c08a45c2376192c62b8985ae17f7af/deploy/docker/scripts/run-with-env.sh#L30
https://github.com/appsmithorg/appsmith/blob/de91b18c418c36349e0226c8935cdc42f1b34d58/deploy/docker/entrypoint.sh#L105
https://github.com/appsmithorg/appsmith/blob/21c999bd08a3fc0a0a6bf8ace680294d588ce2bd/app/server/appsmith-server/src/test/java/com/appsmith/server/services/AuditLogServiceTest.java#L3266
https://github.com/appsmithorg/appsmith/blob/21c999bd08a3fc0a0a6bf8ace680294d588ce2bd/app/server/appsmith-server/src/test/java/com/appsmith/server/services/AuditLogServiceTest.java#L136
https://github.com/appsmithorg/appsmith/blob/c4da7e718016f333941fe58a33ca53c943e970d9/app/server/appsmith-server/src/main/java/com/appsmith/server/constants/EnvVariables.java#L21

All `.yml `files
All `.env.example` files

Also, have we tested this out, when we add the Google Maps key via the Admin settings page, it works as expected?

sharat87:
@ankitakinger, this PR doesn't get rid of the env variable completely, I'm only removing the _client_'s reliance on env variables. The client will get it from the server. The server still gets it from an env variable, but that's abstracted away. Once this is merged, I'll look into making the server also get rid of the env variable, as a separate PR. Smaller PRs. 🙂

app/client/src/selectors/tenantConfigurationSelectors.tsx
```diff
@@ -0,0 +1,4 @@
+import { AppState } from "@appsmith/reducers";
+
+export const getMapsApiKey = (state: AppState): string | undefined =>
```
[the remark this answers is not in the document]
sharat87: Thanks. Didn't realize they were inside the `ce`/`ee` folders. Moved now.

ankitakinger (dismissed):
LGTM

ankitakinger (approved):

app/client/src/widgets/MapWidget/widget/index.tsx
```diff
@@ -486,4 +490,10 @@ export interface MapWidgetProps extends WidgetProps {
   boxShadow?: string;
 }
 
-export default MapWidget;
+const mapStateToProps = (state: AppState) => {
```
riodeuno, line 493: @ashit-rath @SatishGandham Could you check this out? I believe withWidgetProps would be a better spot for this connection. The idea being that widgets should ideally not subscribe to the redux state directly.

I realise that we're already ignoring this in the Modal Widget and the Tabs Widget and would prefer to restructure them as well.

SatishGandham: +1
It should get this as a prop.

ashit-rath: Agreed and I think we should make it generic so that widgets can access other tenant config when required.

riodeuno: @sharat87 What do you think about adding this selector in the `withWidgetProps` file. This will become a generic prop available to all Widgets. We can add a conditional for `widgetType === "MAP_WIDGET"`, but that would not be advised, as it would be a form of abstraction leak.

app/client/src/widgets/withWidgetProps.tsx
```diff
@@ -117,6 +118,12 @@ function withWidgetProps(WrappedWidget: typeof BaseWidget) {
     //merging with original props
     widgetProps = { ...props, ...widgetProps, renderMode };
 
+    // add google mapi key to widget props if it's amap widget
```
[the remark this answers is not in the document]
riodeuno: I was wondering if it is necessary for us to wrap this in the `if` clause. I suppose the api keys don't change often, and it is ok for any widget to have it. This means, that we won't be causing extra re-renders.

This way, if another widget comes in we don't have to worry about providing it the google maps key and avoid an abstraction leak.

app/client/src/selectors/propertyPaneSelectors.tsx
```diff
@@ -211,16 +212,23 @@ export const getWidgetPropsForPropertyName = (
...
       evaluatedWidget: DataTreeWidget,
+      googleMapsApiKey?: string,
     ): WidgetProperties => {
       const widgetProperties = populateWidgetProperties(
         widget,
         propertyName,
         dependencies,
       );
 
+      // if the widget has a googleMapsApiKey dependency, add it to the widget properties
```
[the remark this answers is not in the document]
riodeuno: @SatishGandham Could you review this? This looks ok to me, not sure if there is a better way or if it will cause any performance issues.

app/client/src/widgets/MapWidget/widget/index.tsx
sharat87: Thank you guys. You'll be happy to know, I've enlisted an artisan's (@jsartisan) help with this, and I'm sure you'll like his work! :P

app/client/src/selectors/propertyPaneSelectors.tsx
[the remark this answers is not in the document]
SatishGandham: Looks good.

SatishGandham (dismissed):

app/client/src/reducers/uiReducers/usersReducer.ts
```diff
@@ -116,6 +116,7 @@ const usersReducer = createReducer(initialState, {
   ) => ({
     ...initialState,
     error: action.payload.error,
+    loadingStates: { ...state.loadingStates, fetchingUser: false },
```
riodeuno, line 119: This reducers should be using `createImmerReducer` instead of `createReducer` to avoid these types of spreads.

app/client/src/ce/AppRouter.tsx
```diff
@@ -144,6 +153,23 @@ function AppRouter(props: {
 
   useBrandingTheme();
 
+  // hide the top loader once the tenant is loaded
+  useEffect(() => {
```
riodeuno, line 157: This doesn't seem to be in the PR description. Can we update the description?

app/client/src/reducers/uiReducers/usersReducer.ts
riodeuno: I have created a separate issue for this: https://github.com/appsmithorg/appsmith/issues/20035

riodeuno (approved):
Looks good to me!

sharat87:
Just ran all the tests in the above comment, and they all pass.

Screenshot 2023-02-04 at 7 48 09 AM
Screenshot 2023-02-04 at 7 36 25 AM
Screenshot 2023-02-04 at 8 09 49 AM
Screenshot 2023-02-04 at 8 13 26 AM
Screenshot 2023-02-04 at 9 10 11 AM

sharat87:
On clicking the "Save & Restart" in Admin Settings, *before this PR*, we are calling the following APIs:
Screenshot 2023-02-07 at 9 26 20 AM


In this PR, we are calling the following:
Screenshot 2023-02-07 at 9 26 32 AM


Notice that after we PUT the env changer, we POST to restart, and start waiting. But with this PR, we are making a call to `/admin/env` _after_ POST to restart. This is a race condition. If the restart has already killed the backend server, this `/admin/env` won't succeed, and will result in a 502 error.

Can we see if we need this call at all? If we do, can we wait for it to finish, before the POST to restart?

I think this call to `/admin/env` is coming from [ce/pages/AdminSettings/index.tsx](https://github.com/appsmithorg/appsmith/blob/ac5997cfe4fd0a29c6db54ee35ad6d476bfbcf4f/app/client/src/ce/pages/AdminSettings/index.tsx#L34-L36), but don't know enough to be sure.

Can I get some advise/pointers here please?

PR #19504, opened 2023-01-05
fix: Checkbox group widget fails to display the updated values 

### LINKED ISSUE #19474: [Bug]: Checkbox group widget fails to display the updated values

### Description

Even though the evaluated value of default selected values changes, the Checkbox group fails to update the state of options.

You can go back and forth between page 1 & 2 and you'll see the issue in the shared sample app.

### Steps To Reproduce

1. Add a checkbox group
2. Create fetch query using user data
3. Add checkbox options as - label: username, value: ids
4. Create a new query using user data and only fetch the first two rows. Create a default value array out of it and assign to the checkbox group.

### Public Sample App

https://app.appsmith.com/app/checkbox-group/page1-63b522bd0bc0a80b3001a30c

### Issue video log

_No response_

### Version

v1.8.16

jsartisan:
## Description

In the issue report by the user the component was not checked values for defaultSelectedValues. That was because we had componentDidUpdate which was checking the change in the options property. If there was a change, it was setting the selectedValues. We found out that this check is not required. Because of this check only, the component was re-rendering, and sometimes it was unchecking all the checked values. So we just removed the check and added the cypress check.

Fixes #19474

## Type of change
- Bug fix 

## How Has This Been Tested?
- Cypress

### Dev activity

### QA activity:


ashit-rath (commented):
@jsartisan Can you explain what was the problem and how removing a bunch code is fixing that issue?

jsartisan:
@ashit In the issue report by the user the component was not checked values for defaultSelectedValues. That was because we had `componentDidUpdate` which was checking the change in the `options` property. If there was a change, it was setting the `selectedValues`. We found out that this check is not required. Because of this check only, the component was re-rendering, and sometimes it was unchecking all the checked values. So we just removed the check and added the cypress check.

ashit-rath (dismissed):

chandannkumar:
Tested this PR and LGTM

dhruvikn (approved):

PR #19521, opened 2023-01-06
chore: Deny service, if not on MongoDB v5

sharat87:
Adds a migration function, that checks the version of the MongoDB being used. If it's older than v5, we print an informative message like this:

```
Caused by: com.appsmith.server.exceptions.UnsupportedMongoDBVersionException: Appsmith requires MongoDB v5. Please upgrade your MongoDB. You are running: '4.4.16'.
```

Then deny service. No requests will be accepted. This is to avoid potential data loss/damage due to a change in the way queries are run in whatever old version is being run.


mohanarpit (approved):

nidhi-nair (approved):

sharat87:
Not doing this anymore.

PR #19552, opened 2023-01-06
chore: Use rolling window in the usage pulse call and send segment anonymous id for anonymous users

sbalaji1192:
## Description
We have updated the user tracking to use a rolling window. Now the session starts when the user goes to the builder or viewer for the first time and the subsequent activity tracking will be checked only after an hour.
For anonymous users, we send the Segment anonymous id in the usage calls. When the telemetry is off, we still initiate the segment, get the id and then purge the analytics global object.

Fixed window (in release, as of now)- if the user starts a session at 01:15 pm, we take 01:00 pm as the session start time and we will check for the next user activity at 2:00 pm.

rolling window (in this PR) - if the user starts a session at 01:15 pm, we take 01:15 pm as the session start time and we will check for the next user activity at 2:15 pm.

Fixes https://github.com/appsmithorg/cloud-services/issues/183

## Type of change

## How Has This Been Tested?

- Manual

### Dev activity

### QA activity:


riodeuno:
@sbalaji1192 I'm afraid, I don't get the concept of rolling window vs fixed window. Would you mind elaborating in the description for posterity?

app/client/src/constants/routes/appRoutes.ts
```diff
@@ -3,12 +3,22 @@
 // All solutions from closed issues on their repo have been tried. Ref: https://github.com/pillarjs/path-to-regexp/issues/193
 const { match } = require("path-to-regexp");
 
+export const BUILDER_VIEWER_PATH_PREFIX = "/app/";
```
riodeuno, line 6: @sbalaji1192 I'm not sure I understand these changes. Could you explain?

app/client/src/usagePulse/index.ts
```diff
@@ -0,0 +1,97 @@
...
+      window.document.body.removeEventListener(
+        event,
+        UsagePulse.startTrackingActivity,
+      );
+    });
+  }
+
+  static watchForTrackableUrl(callback: () => void) {
+    UsagePulse.unlistenRouteChange = history.listen(() => {
+      if (UsagePulse.isTrackableUrl(window.location.href)) {
```
riodeuno, line 63: @sbalaji1192 Is this supposed to be `!UsagePulse.isTrackableUrl(window.location.href)`?

app/client/src/usagePulse/index.ts
```diff
@@ -0,0 +1,97 @@
...
+        UsagePulse.unlistenRouteChange();
```
riodeuno, line 64: @sbalaji1192 I'm not sure how this works. We're calling `unlistenRouteChange` within the setup of the `unlistenRouteChange`, seems like an anti-pattern. Can you describe what's happening?

app/client/src/usagePulse/index.ts
```diff
@@ -0,0 +1,97 @@
...
+      PULSE_INTERVAL * 1000,
+    );
+  }
+
+  static startTrackingActivity() {
+    if (UsagePulse.isTrackableUrl(window.location.href)) {
+      UsagePulse.sendPulse();
+      UsagePulse.scheduleNextActivityListeners();
+    } else {
+      UsagePulse.watchForTrackableUrl(UsagePulse.startTrackingActivity);
```
riodeuno, line 86: Is this why we're not doing a `!` in `UsagePulse.isTrackableUrl(window.location.href)`? (A previous comment)

app/client/src/ce/sagas/userSagas.tsx
```diff
@@ -134,19 +138,40 @@ export function* getCurrentUserSaga() {
...
+        //@ts-expect-error: response is of type unknown
+        response.data.isAnonymous &&
+        //@ts-expect-error: response is of type unknown
+        response.data.username === ANONYMOUS_USERNAME
+      ) {
+        /*
+         * We're initializing the segment api regardless of the enableTelemetry flag
+         * So we can use segement Id to fingerprint anonymous user in usage pulse call
+         */
+        AnalyticsUtil.blockTrackEvent = true;
```
riodeuno, line 164: Is there any other pattern we can use to do this? Seems like we're changing the mode of the `AnalyticsUtil`'s `logEvent`. 

If we can do something like `initialiseNonTelemetrySegment` or `initialiseSegmentFingerPrinting`?

I think that in this approach, we're not making things explicit, as the handling logic is moved away from the AnalyticUtils class

riodeuno (commented):
@sbalaji1192 I have a few comments on the code.
Regarding the Usagepulse class, I found it a little hard to understand. Would you mind adding comments on each of the methods to further elaborate what is happening?

app/client/src/constants/routes/appRoutes.ts
sbalaji1192: We are only sending the usage pulse call from the builder or viewer page. So instead of hard coding the path in the usage pulse file, we are creating a new constant BUILDER_VIEWER_PATH_PREFIX and using this in all the routes as well the usage pulse file.
And also trying to keep appRoutes.ts as the source of truth for paths

app/client/src/usagePulse/index.ts
sbalaji1192: We're checking for truthy value only here. We want to only call the callback when user goes to trackable url.

app/client/src/usagePulse/index.ts
sbalaji1192: This is not an anti-pattern. We're trying to listen for the route change until the user goes to a trackable URL. 
Since we don't want to track the user activity in pages other than builder or viewer, when its time to send a pulse, we check if a user is in a trackable URL otherwise we listen for history change event until goes to a trackable URL. once they do we unregister the listener (unlistenRouteChange) and then call the callback.

app/client/src/usagePulse/index.ts
sbalaji1192: yes. we want to watch for trackable URL when user is not in a trackable url.

PR #19747, opened 2023-01-12
feat: Adapt widgets to respond to viewport changes in auto layout mode

aswathkk:
> Pull Request Template
>

## Description

> and context. List any dependencies that are required for this change
> Links to Notion, Figma or any other documents that might be relevant to the PR

Fixes # (issue)

Media

## Type of change

- Manual
- Jest
- Cypress

### Dev activity

### QA activity:


app/client/src/widgets/BaseInputWidget/component/styles.module.css
```diff
@@ -0,0 +1,24 @@
+:global(.auto-layout) .inputComponentWrapper {
```
yaldram, line 1: We are additionally targeting the `.auto-layout` class so that it does not conflict in fix layouts, awesome.

app/client/src/selectors/mainCanvasSelectors.tsx
```diff
@@ -1,7 +1,17 @@
...
 export const getIsCanvasInitialized = (state: AppState) => {
   return state.ui.mainCanvas.initialized;
 };
 
 export const getIsMobile = (state: AppState) => state.ui.mainCanvas.isMobile;
+
+export const getUseAutoLayout = (state: AppState) =>
+  state.entities.canvasWidgets[MAIN_CONTAINER_WIDGET_ID].useAutoLayout;
+
+export function isAutoLayout() {
```
aswathkk, line 14: Check if this is reliable

app/client/src/utils/autoLayout/positionUtils.ts
```diff
@@ -222,6 +256,20 @@ export function placeWidgetsWithoutWrap(
     }
   }
 
+  // Trigger a position update for the widgets inside container widgets
+  for (const each of arr) {
+    for (const widget of each.children) {
+      if (widget.type === "CONTAINER_WIDGET" && widget.children?.length) {
```
prsidhu, line 262: @aswathkk would this apply to tabs widget and other widgets with canvases as well?

app/client/src/utils/autoLayout/positionUtils.ts
```diff
@@ -222,6 +256,20 @@ export function placeWidgetsWithoutWrap(
...
+        widgets = updateWidgetPositions(
+          widgets,
+          widget.children[0],
```
prsidhu, line 265: Due to container splitting, the number of canvases within a container will increase. So this won't scale.

app/client/src/utils/autoLayout/positionUtils.ts
```diff
@@ -222,6 +256,20 @@ export function placeWidgetsWithoutWrap(
     }
   }
 
+  // Trigger a position update for the widgets inside container widgets
+  for (const each of arr) {
```
prsidhu, line 260: @aswathkk What's the use case triggering a position update here?

app/client/src/utils/autoLayout/positionUtils.ts
```diff
@@ -56,6 +59,13 @@ export function updateWidgetPositions(
     const parent = widgets[parentId];
     if (!parent) return widgets;
 
+    const { columnSpace } = getCanvasDimensions(
```
prsidhu, line 62: @aswathkk The reason usage of this method was discontinued was because of minute inaccuracy in the calculation, which gets highlighted as the level of nesting of containers increases. Is it possible to pass the snapSpaces calculated in container / canvas widget to all points of usage of this method? If not, then I'll work on improving the accuracy of this method.

app/client/src/utils/autoLayout/positionUtils.ts
aswathkk: If there's a container like widget in the same row, the container shrinks/expands when we add or delete widgets next to it.
So, widgets inside those container will also shrinks/expands accordingly. We need to make sure that those widgets inside container are not shrink beyond minWidth.

app/client/src/utils/autoLayout/positionUtils.ts
aswathkk: Good catch.
I need to do this for other widgets with canvas

app/client/src/utils/autoLayout/positionUtils.ts
aswathkk: This operation has to be done for all the canvases.
Do you have any other idea to implement this?

app/client/src/utils/autoLayout/positionUtils.ts
aswathkk: Yeah. I've seen that inaccuracy.
This same method is being used in AutoLayoutUpdateSaga to update the bounding box of widget when its content changes.
When the widget is inside a container, the bounding box is not being accurate.

I tried to use `widget.parentColumnSpace` property. But, we can't rely on that since it won't be properly updated at this point.

Do you have any idea about the source of inaccuracy in this method?

prsidhu: Inaccuracy resides in calculation of the padding values to remove from the total width. Inaccuracy increases with nesting of containers.

app/client/src/utils/autoLayout/positionUtils.ts
prsidhu: @aswathkk  I'm not too keen about the positioning of this piece of code. Since positionUtils are called for all widgets upon switching to mobile viewport, this would result in redundant re-runs. 

In the saga that calls this function, we are calling updateWidgetPositions for all parents which in all cases would be a canvas. So doesn't that solve our problem?

app/client/src/components/designSystems/appsmith/autoLayout/AutoLayoutDimensionObeserver.tsx
```diff
@@ -0,0 +1,79 @@
+import { FLEXBOX_PADDING } from "constants/WidgetConstants";
+import React, { useState } from "react";
+import { PropsWithChildren, useEffect, useRef } from "react";
+import styled from "styled-components";
+
+const SimpleContainer = styled.div`
+  width: fit-content;
+  &.fill-widget {
+    width: 100%;
```
prsidhu, line 9: Do we need this? fill widgets should come with a column requirement of 64 or whatever is available.

aswathkk: This is the resize observer container.
If `width: 100%` is not provided here, the fill widgets will not take up 100% of the width available.

app/client/src/components/designSystems/appsmith/autoLayout/AutoLayoutDimensionObeserver.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  useEffect(() => {
+    if (currentDimension.width === 0 || currentDimension.height === 0) return;
+    const widthDiff = Math.abs(
+      props.width - 2 * FLEXBOX_PADDING - currentDimension.width,
+    );
+    const heightDiff = Math.abs(
+      props.height - 2 * FLEXBOX_PADDING - currentDimension.height,
+    );
+    if (widthDiff > 2 || heightDiff > 2) {
```
prsidhu, line 49: why 2?

app/client/src/utils/autoLayout/positionUtils.ts
```diff
@@ -661,3 +661,58 @@ export function getTotalRowsOfAllChildren(
...
+      const canvases: string[] = widget.children?.filter(
+        (id: string) => widgets[id] && widgets[id].type === "CANVAS_WIDGET",
+      );
+      if (canvases.length) {
+        widgetsToBeParsed = [...widgetsToBeParsed, ...canvases];
+      }
+    });
+  }
+  // Update positions of all the widgets.
+  for (const widgetId of widgetsToBeParsed) {
```
aswathkk, line 694: @prsidhu Here we are updating position only for the child.
We need to do the same for deeply nested elements too. Right?

app/client/src/utils/autoLayout/AutoLayoutUtils.ts
```diff
@@ -352,6 +391,72 @@ export function getAlignmentColumnInfo(
...
+        ? 1
+        : columns / GridDefaults.DEFAULT_GRID_COLUMNS;
+    width = width * factor - padding;
+  }
+
+  return width;
+}
+
+function getPadding(canvas: FlattenedWidgetProps): number {
+  let padding = 0;
```
aswathkk, line 447: We should investigate why the logic here is different from what we have at `getSnappedGrid` in WIdgetOperationUtils.ts
And make use of it.

app/client/src/components/designSystems/appsmith/autoLayout/AutoLayoutDimensionObeserver.tsx
aswathkk: Should've been 0.
fixing this

app/client/src/utils/autoLayout/positionUtils.ts
aswathkk: https://user-images.githubusercontent.com/10436935/224335617-7887e8e7-fe06-4852-beae-9dc4e55616e4.mov

app/client/src/utils/WidgetFactory.tsx
```diff
@@ -178,6 +182,28 @@ class WidgetFactory {
...
         ),
       );
+
+      autoLayoutConfig &&
+        this.autoLayoutConfigMap.set(widgetType, {
+          ...autoLayoutConfig,
+          widgetSize:
+            autoLayoutConfig.widgetSize?.map((sizeConfig) => ({
+              ...sizeConfig,
+              configuration: (props: WidgetProps) => {
```
riodeuno, line 192: Please elaborate on what is happening here. Also, let's move this to a separate function. Ideally, we'd want these properties to be available by default, without having to do these checks. If we can't have that, we can enhance the function in the widget register helpers to do so. Let's discuss.

app/client/src/utils/WidgetRegisterHelpers.tsx
```diff
@@ -61,6 +61,7 @@ export const registerWidget = (Widget: any, config: WidgetConfiguration) => {
     config.features,
     config.properties.loadingProperties,
     config.properties.stylesheetConfig,
+    config.autoLayout,
```
riodeuno, line 64: Shall we handle the undefined values here, instead of doing this in the WidgetFactory. The reason why we can't do that is due to the need for `props` which is only available at runtime, but I think there might be a better way to handle this. Also, I'm assuming that the function properties are not being sent to evaluations, as they're not serialisable and will fail to be transferred to the worker thread.

app/client/src/utils/autoLayout/AutoLayoutUtils.ts
```diff
@@ -352,6 +391,72 @@ export function getAlignmentColumnInfo(
...
+    mainCanvasWidth,
+    isMobile,
+  );
+
+  const columnSpace: number = canvasWidth / GridDefaults.DEFAULT_GRID_COLUMNS;
+
+  return { canvasWidth: canvasWidth, columnSpace };
+}
+
+function getCanvasWidth(
```
riodeuno, line 412: Please describe what this function does. It seems to be very close in name with getCanvasDimensions, and as we can see, this is substantially more complex than the previous function.

app/client/src/utils/autoLayout/AutoLayoutUtils.ts
```diff
@@ -352,6 +391,72 @@ export function getAlignmentColumnInfo(
...
+  while (widget.parentId) {
+    stack.push(widget);
+    widget = widgets[widget.parentId];
+  }
+  stack.push(widget);
+
+  let width = mainCanvasWidth;
+  while (stack.length) {
+    const widget = stack.pop();
+    if (!widget) continue;
```
riodeuno, line 433: I am curious as to how we reach this scenario where the widget is undefined. If we reach undefined, should we break instead of continue?

app/client/src/utils/autoLayout/AutoLayoutUtils.ts
```diff
@@ -352,6 +391,72 @@ export function getAlignmentColumnInfo(
...
+    const columns = getWidgetWidth(widget, isMobile);
+    const padding = getPadding(widget);
+    const factor =
```
riodeuno, line 436: What is this factor? Please elaborate in a comment.

app/client/src/utils/autoLayout/AutoLayoutUtils.ts
riodeuno: Have we investigate this? What were the findings?

app/client/src/utils/autoLayout/autoLayoutUtils.test.ts
```diff
@@ -4,16 +4,18 @@ import {
...
 import {
+  getCanvasDimensions,
   getFlexLayersForSelectedWidgets,
   getLayerIndexOfWidget,
   getNewFlexLayers,
   pasteWidgetInFlexLayers,
   updateFlexLayersOnDelete,
 } from "./AutoLayoutUtils";
-import { data } from "./testData";
+import { data, dataForgetCanvasDimensions } from "./testData";
```
riodeuno, line 14: I don't understand what this variable does. Seems like we meant `dataForGetCanvasDimensionsFunction`

app/client/src/utils/autoLayout/flexWidgetUtils.ts
```diff
@@ -1,3 +1,15 @@
+import { getIsAutoLayout } from "selectors/editorSelectors";
+import store from "store";
+import WidgetFactory from "utils/WidgetFactory";
+import { WidgetSizeConfig } from "widgets/constants";
+
+export interface MinMaxSize {
+  minHeight: number | string;
```
riodeuno, line 7: In which scenarios would this be a string and in which would this be a number. Ideally, we should use one type in all scenarios. If necessary, we can create two different types, one for strings and one for numbers.

app/client/src/utils/autoLayout/flexWidgetUtils.ts
```diff
@@ -111,3 +123,96 @@ export function getWidgetRows(widget: any, isMobile: boolean): number {
...
+ * @param canvasWidth | number : main canvas width.
+ * @returns MinSize | undefined
+ */
+function getMinMaxSize(
+  widget: any,
+  canvasWidth: number,
+): MinMaxSize | undefined {
+  // Get the widget size configuration.
+  const sizeConfig = getCurrentSizeConfig(widget, canvasWidth);
+  if (!sizeConfig) return;
```
riodeuno, line 139: When will this be a scenario? Could we elaborate in a comment?

app/client/src/utils/autoLayout/flexWidgetUtils.ts
```diff
@@ -111,3 +123,96 @@ export function getWidgetRows(widget: any, isMobile: boolean): number {
...
+
+  return { maxHeight, maxWidth, minHeight, minWidth };
+}
+
+export function getCurrentSizeConfig(
+  widget: any,
+  canvasWidth: number,
+): WidgetSizeConfig | undefined {
+  // Get the widget size configuration.
+  const sizeConfig = WidgetFactory.getWidgetAutoLayoutConfig(widget.type);
```
riodeuno, line 154: Let's discuss what the sizeConfigs are and how we're adding them to the configurations. I find that we're making too many `undefined` checks, which ideally should not be the same. It might be an issue with how we've designed the data flow and configurations.

app/client/src/utils/autoLayout/flexWidgetUtils.ts
```diff
@@ -111,3 +123,96 @@ export function getWidgetRows(widget: any, isMobile: boolean): number {
   const divisor = widget.parentRowSpace === 1 ? 10 : 1;
   return getBottomRow(widget, isMobile) / divisor - getTopRow(widget, isMobile);
 }
+
+/**
+ * Calculates the minimum & maximum size of a widget based on the widget type and the canvas width.
+ * @param widget | Widget props
+ * @param canvasWidth | number : main canvas width.
+ * @returns MinSize | undefined
```
riodeuno, line 131: Let's update the description, we're returning the MinMaxSize and not the MinSize.

app/client/src/utils/autoLayout/flexWidgetUtils.ts
```diff
@@ -111,3 +123,96 @@ export function getWidgetRows(widget: any, isMobile: boolean): number {
   const divisor = widget.parentRowSpace === 1 ? 10 : 1;
   return getBottomRow(widget, isMobile) / divisor - getTopRow(widget, isMobile);
 }
+
+/**
+ * Calculates the minimum & maximum size of a widget based on the widget type and the canvas width.
```
riodeuno, line 128: I really appreciate these function descriptions, let's make sure all the new functions we've added have these.

app/client/src/utils/autoLayout/flexWidgetUtils.ts
```diff
@@ -111,3 +123,96 @@ export function getWidgetRows(widget: any, isMobile: boolean): number {
...
+  // Find the most suitable breakpoint for the canvas width.
+  const sizes: WidgetSizeConfig[] = sizeConfig?.widgetSize;
+  let index = 0;
+  while (index < sizes?.length && canvasWidth > sizes[index].viewportMinWidth) {
+    index += 1;
+  }
+  return sizes[index - 1];
+}
+
+function getPxValue(val: string | number, factor: number): number | undefined {
```
riodeuno, line 166: Is there anyway, we can get away without having to extract numbers from string like these? I am wondering when we would need to use strings, and the scenario I think of, is when we're passing the value to the widget component styles, and we may be able to maintain numbers in all computations. Let's discuss this.

app/client/src/utils/autoLayout/flexWidgetUtils.ts
```diff
@@ -111,3 +123,96 @@ export function getWidgetRows(widget: any, isMobile: boolean): number {
...
+
+  returnValue.maxHeight = getPxValue(
+    minMaxSize.maxHeight,
+    widget.parentRowSpace,
+  );
+
+  return returnValue;
+}
+
+export function isAutoLayout() {
```
riodeuno, line 215: Do we need to access the store from here? I'm asking because, I've seen approaches like these to cause issues in unit tests. In general we should try to keep our functions pure.

app/client/src/utils/layoutPropertiesUtils.ts
```diff
@@ -193,82 +193,3 @@ export const generateVerticalAlignmentConfig = (
...
-  "BUTTON_GROUP_WIDGET",
-  "CHART_WIDGET",
-  "CHECKBOX_WIDGET",
-  "CHECKBOX_GROUP_WIDGET",
-  "CURRENCY_INPUT_WIDGET",
-  "CONTAINER_WIDGET",
-  "DATE_PICKER_WIDGET2",
-  "DIVIDER_WIDGET",
-  "FORM_WIDGET",
-  "FILE_PICKER_WIDGET_V2",
```
riodeuno, line 235: Nice! Always happy to see constants and types like these extracted from the right places.

app/client/src/widgets/AudioWidget/component/index.tsx
```diff
@@ -25,8 +27,6 @@ const ErrorContainer = styled.div`
   height: 100%;
 `;
 
-const Error = styled.span``;
```
riodeuno, line 28: Nice. Less code to compute.

app/client/src/widgets/BaseInputWidget/component/index.tsx
```diff
@@ -368,6 +368,10 @@ const TextInputWrapper = styled.div<{
   box-shadow: ${({ boxShadow }) => `${boxShadow}`} !important;
   min-height: 32px;
 
+  .auto-layout & {
```
riodeuno, line 371: How is this different from the "minHeight" in configurations?

app/client/src/widgets/BaseInputWidget/widget/index.tsx
```diff
@@ -43,6 +44,7 @@ class BaseInputWidget<
             label: "Position",
             controlType: "ICON_TABS",
             fullWidth: true,
+            hidden: isAutoLayout,
```
riodeuno, line 47: Let's use something like `hideInAutoLayout: true` and use the PropertyControl component to conditionally avoid showing this control.

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -614,9 +617,44 @@ abstract class BaseWidget<
         </AutoHeightContainerWrapper>
       );
     }
+    if (this.props.isFlexChild && !this.props.detachFromLayout) {
```
riodeuno, line 620: Can we move all of these computations into the AutoLayoutDimensionObserver?

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -614,9 +617,44 @@ abstract class BaseWidget<
...
+    }
     return this.addErrorBoundary(content);
   };
 
+  private updateWidgetDimensions = (width: number, height: number) => {
+    const { updateWidgetDimension } = this.context;
+    if (!updateWidgetDimension) return;
+    updateWidgetDimension(
+      this.props.widgetId,
+      width + 2 * FLEXBOX_PADDING,
```
riodeuno, line 653: Ideally, we should send this 2 * FLEXBOX_PADDING from the AutoLayoutDimensionObserver, unless there are other code paths through which this method is called.

app/client/src/widgets/ButtonWidget/component/DragContainer.tsx
```diff
@@ -45,6 +45,14 @@ const ButtonContainer = styled.div<ButtonContainerProps>`
     height: 100%;
   }
 
+  .auto-layout & > .bp3-button.bp3-fill {
```
riodeuno, line 48: These are some interesting values, please elaborate on how we arrived that these. Would using something like CSS clamp help?

app/client/src/widgets/ButtonWidget/index.ts
```diff
@@ -41,6 +42,30 @@ export const CONFIG = {
...
+    },
+    autoDimension: {
+      width: true,
+    },
+    widgetSize: [
+      {
+        viewportMinWidth: 0,
+        configuration: () => {
+          return {
+            minWidth: "120px",
```
riodeuno, line 58: I was wondering, if we can make our lives easier if we use numbers instead of strings here. For example, `minWidth: 120`

app/client/src/widgets/CheckboxWidget/index.ts
```diff
@@ -41,6 +42,24 @@ export const CONFIG = {
...
+        viewportMinWidth: 0,
+        configuration: () => {
+          return {
+            minWidth: "120px",
+          };
+        },
+      },
+    ],
+    disableResizeHandles: {
+      vertical: true,
```
riodeuno, line 60: If we only have `minWidth` or `minHeight` defined, can we always assume that the `vertical` or `horizontal` would be true respectively?

app/client/src/widgets/IconButtonWidget/index.ts
```diff
@@ -30,6 +30,29 @@ export const CONFIG = {
...
+      {
+        viewportMinWidth: 0,
+        configuration: () => {
+          return {
+            minWidth: "40px",
+          };
+        },
+      },
+    ],
+    disableResizeHandles: {
```
riodeuno, line 51: Ah, I see here that my previous question in a previous comment regarding the relation between `minWidth|minHeight` and `vertical|horizontal` is answered here.

app/client/src/widgets/constants.ts
```diff
@@ -15,7 +15,29 @@ import { DerivedPropertiesMap } from "utils/WidgetFactory";
 import { WidgetFeatures } from "utils/WidgetFeatures";
 import { WidgetProps } from "./BaseWidget";
 
+export type WidgetSizeConfig = {
+  viewportMinWidth: number;
+  configuration: (props: any) => { [key: string]: string | number };
```
riodeuno, line 20: Should we make this `(props: WidgetProps) => Record<string, string|number>`? I find this to be a little more readable.

riodeuno (commented):
I can see that a lot of work has gone into making this happen. I really appreciate all the work that we've put into this. 

I have quite a few questions in the review. There are a couple of things I would like to mention in general regarding this PR.

1. Please add function comments and comments to code which is hard to understand or we're computing something mathematically or algorithmically.

2. Please add unit tests for all functions we've created in this PR.

riodeuno:
@aswathkk Please fill in the details of the PR description. It is important that we include as much detail as possible. For example, there are a few things we've done in this PR which should mention in the PR description.

- We've created a new set of widget configurations specific to auto layout
- We've updated the styles of the widget components with very specific `min-height` and other styles based on the `.auto-layout` class selector.
- We're hiding specific widget properties based on the mode in which we are (auto layout or fixed)
- We're listening to widget dimension changes based on the content in some of the widgets, so they auto resize according to content size.

PR #19775, opened 2023-01-13
chore: send anonymous id from segment for anonymous users in usage pulse call

sbalaji1192:
## Description

For anonymous users, we need segment id, so if telemetry is off we're initiating the segment without tracking and once we get the id, the analytics object is purged.

Fixes https://github.com/appsmithorg/cloud-services/issues/231

## Type of change

- Manual
- Jest
- Cypress

### Dev activity

### QA activity:

### DIFF
--- a/app/client/src/ce/sagas/userSagas.tsx
+++ b/app/client/src/ce/sagas/userSagas.tsx
@@ -1,4 +1,4 @@
-import { call, put, race, select, take } from "redux-saga/effects";
+import { call, fork, put, race, select, take } from "redux-saga/effects";
 import {
   ReduxAction,
   ReduxActionWithPromise,
@@ -55,7 +55,10 @@ import {
   getFirstTimeUserOnboardingApplicationId,
   getFirstTimeUserOnboardingIntroModalVisibility,
 } from "utils/storage";
-import { initializeAnalyticsAndTrackers } from "utils/AppsmithUtils";
+import {
+  initializeAnalyticsAndTrackers,
+  initializeSegmentWithoutTracking,
+} from "utils/AppsmithUtils";
 import { getAppsmithConfigs } from "ce/configs";
 import { getSegmentState } from "selectors/analyticsSelectors";
 import {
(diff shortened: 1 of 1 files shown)


app/client/src/ce/sagas/userSagas.tsx
```diff
@@ -173,16 +188,16 @@ export function* getCurrentUserSaga() {
...
-
-        if (!enableTelemetry) {
-          AnalyticsUtil.removeAnalytics();
-        }
       }
 
-      UsagePulse.startTrackingActivity();
-
+      /*
+       * Forking it as we don't want to block application flow
```
riodeuno, line 194: 💯

riodeuno (dismissed):

app/client/src/ce/sagas/userSagas.tsx
```diff
@@ -124,6 +127,32 @@ export function* waitForSegmentInit(skipWithAnonymousId: boolean) {
...
+ */
+export function* initiateUsageTracking(payload: {
+  isAnonymousUser: boolean;
+  enableTelemetry: boolean;
+}) {
+  //To make sure that we're not tracking from previous session.
+  UsagePulse.stopTrackingActivity();
+
+  if (payload.isAnonymousUser) {
+    if (payload.enableTelemetry) {
```
eco-monk, line 144: We should add a check for `appsmithConfig.segment.enabled` as well.
Unless this is left out intentionally.

sbalaji1192: When does `appsmithConfig.segment.enabled` get disabled?

eco-monk: It's from the .env configuration.
This is checked inside `initializeAnalyticsAndTrackers` method.

So, even if `user/me` returns `enableTelemetry`: true for a user.
If the environment variable is set to false, we won't be initialising segment.

sbalaji1192: I think both user/me `enableTelemetry` and appsmithConfig.segment.enabled are controlled by the same env variable.
But not 100% sure. Let me add the `appsmithConfig.segment.enabled` check anyways to be sure.

eco-monk (dismissed):

app/client/src/ce/sagas/userSagas.tsx
eco-monk: I believe it will be different, because the env config is more granular.
e.g. appsmithConfig.sentry, appsmithConfig.smartlook and appsmithConfig.segment

Will check with someone on backend to be sure.
Would simplify things in client side if what you say is true.

eco-monk (approved):

PR #19777, opened 2023-01-13
chore: Adding user permissions in the response of get actions in view mode

### LINKED ISSUE #19055: [Bug]: The query execution fails in the edit mode when there is no view permission for the queries

### Description

The query execution fails in the edit mode when there is no view permission for the queries
image
image

### Steps To Reproduce

1. login as admin
2. in admin setting page
3. create a custom role
4. Add edit access on page but not on queries
5. Add view access on page not on queries 
6. Also give execute permission for all queries
7. Now  in the view mode the the page looks fine 
8. But when user open the app in edit mode the queries execution failed

### Public Sample App

_No response_

### Issue video log

_No response_

### Version

localhost

trishaanand:
Fixes #19055 
(backend fix. frontend changes shall follow)


mohanarpit:
@trishaanand Should both backend & frontend changes be made in this same PR? Will that help us test this fix out more reliably?

PR #19783, opened 2023-01-13
chore: Adding environmentName parameter in datasource test flow and required decomposition and refactors.

sondermanish:
## Description
- TL; DR , small Pr to add environmentName to Datasource-controller for `test`, `structure`, and `trigger` flows. In order to adapt for environments.  There is a method addition in authenticationValidator, and further refactor to avoid code duplicity.

- With Introduction of multiple environments we would have configurations corresponding to each environment. In order to test these configurations, we would require an environment parameter to be passed as an argument,  this could be done by passing an environment header with the current datasource `test`, `structure`, and `trigger` flows, however, since our community and enterprise flow uses the same code. We require our community code to include the header which however, will not be used in the community edition but it will have complete usage in enterprise edition.
- A `validateAuthentication` method in `authenticationValidator` interface has been loaded with an extra parameter (environmentId) for environment support, EE changes will follow post the merge.
- `Datasource trigger and Datasource Structure flows`, required some refactors to make it work with environments as well. 

## Type of change

## How Has This Been Tested?
- Manual
- Jest
- Cypress

Existing test cases are adequate, just tweaked a few test cases to accommodate changes.

### Dev activity

### QA activity:


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCE.java
```diff
@@ -15,7 +15,14 @@
...
 
-    Mono<DatasourceTestResult> testDatasource(Datasource datasource);
+    /**
+     *
+     * @param datasource - The datasource which is about to be tested
+     * @param environmentName - environmentName, name of the environment on which the datasource is getting tested,
+     *                         this variable is unused in the CE version of the code.
+     * @return Mono<DatasourceTestResult> - result whether the datasource secures a valid connection with the remote DB
+     */
+    Mono<DatasourceTestResult> testDatasource(Datasource datasource, String environmentName);
```
sumitsum, line 25: QQ: Why are we using environment name instead of environment id ?

sondermanish: Since we used the environment name in the execute action flow, I am using the environmentName and workspaceId to fetch the  environment object and  evaluate the datasource. This test flow will use the same evaluation method. Hence it requires the same environment name, though, agreed that we can also use environmentId instead of environmentName, however, both flow need to be using the same argument to keep DB calls to minimum. we are not using any environmentName then we are defaulting to the production environment. Furthermore, we are defaulting to "production" environment if we do not receive any string in the environmentName argument in the enterprise edition version.

sumitsum: @sondermanish how are we ensuring that the environment name will be unique within the environment collection ? Or do we intend to use `workspace Id x environment name` as the UID, even then we would have to ensure that the environment name is unique in a workspace. From a searching perspective i.e. finding in the DB, wouldn't it be better if we were to use the UID generated by Mongo  ?

sondermanish: we have a uniquenes index based on workspaceId, environmentName and deletedAt field. Hence no unarchived environments in the same workspace can have the same name.

sumitsum (approved):

btsgh:
@sondermanish - Checked the Test button for AmazonS3, MySQL, PostGreSQL, Dynamo DB, Firestore, Mongo. Test goes through fine. Able to Save the datasources and also create a simple query against each datasource without any issues.

nidhi-nair (dismissed):

nidhi-nair (approved):

PR #19798, opened 2023-01-16
ci: Add JAVA setup step to perf tests in test build workflow

SatishGandham:
Add JAVA setup step to perf tests in test build workflow.
This is a possible fix for perf tests on running because backend server was not up.


somangshu (approved):
LGTM

PR #19833, opened 2023-01-17
fix: Query and data source page - 404 error on reload

### LINKED ISSUE #19796: [Bug]: Query and datasource pages containing plugin in their slug returns a 404 after page refresh

### Description

Pages that have `plugin` present on their slug do not load on page refresh. This behaviour works fine on prod but not on release hence making this a blocker

### Steps To Reproduce

1. Go to a Google sheet/GraphQL datasource/query page and refresh the page to see a 404 as seen in the screenshots below

image
image
image

### Public Sample App
[Prod link](https://app.appsmith.com/app/nada-plugin/api/63c5119e04ebd915634211f7)
[query page crash.zip](https://github.com/appsmithorg/appsmith/files/10424095/query.page.crash.zip)
[Slack thread](https://theappsmith.slack.com/archives/C040LHZN03V/p1673855783712559)

### Issue video log

_No response_

### Version

Cloud

eco-monk:
Fixes #19796

## Description
- Internally we use expressions similar to JS’s regex, to identify url versions (old url with id, slug based url, custom slug url).
- Some have url’s have overlapped matching with multiple expressions. Hence, pageId was identified with the wrong value.
- Expressions: `VIEWER_PATH` (view mode with default slug) and `BUILDER_CUSTOM_PATH` (edit mode with custom slug)

Current approach:
- Change `getUpdatedRoute` conditions so it checks for pattern matches in the same order as `AppRouter.ts`. 
(Now checking `BUILDER_CUSTOM_PATH` first and then `VIEWER_PATH`)
- Because this order of pattern checks in, `AppRouter.ts` deals with this overlap properly. 

> Initial approach
> Added check to verify pageId properly when custom slug and regular slug path pattern overlaps.

TLDR; Translates to -> Page will not throw 404 on reload anymore.

Media
https://www.loom.com/share/993fec8a37ef4e2ca5e46d967efb83c9

---

TODO for the future:
- [ ] [Figure out a way to not solely rely in order of pattern checks in both AppRouter and getUpdatedRoute.](https://github.com/appsmithorg/appsmith/issues/19888)

---

## Type of change

## How Has This Been Tested?
- Jest

### Dev activity

### QA activity:


arunvjn:
@eco-monk Why does react-router parse the pageId properly in Editor/index.tsx and not inside `getUpdatedRoute` method ?

eco-monk:
So right now, the path `https://dev.appsmith.com/app/custom-63c63d944ae4345e31af12a7/edit/saas/google-sheets-plugin/api/63c63d984ae4345e31af12e5`.

Matches with both `VIEWER_PATH` and `BUILDER_CUSTOM_PATH`.

React router matches with `BUILDER_CUSTOM_PATH` because it's checked first before `VIEWER_PATH`.
```
      <SentryRoute component={EditorLoader} path={BUILDER_PATH} />
      <SentryRoute component={EditorLoader} path={BUILDER_CUSTOM_PATH} /> -> matches here first
      <SentryRoute component={AppViewerLoader} path={VIEWER_PATH} />
      <SentryRoute component={AppViewerLoader} path={VIEWER_CUSTOM_PATH} />
```
This is not the case with `getUpdatedRoute` method. We check whether there is a match with `VIEWER_PATH` first and then the `BUILDER_CUSTOM_PATH`.

---

If we were to switch around the order in AppRouter.tsx so it checks VIEWER_PATH first. 
We'd have the same issue.

```
      <SentryRoute component={EditorLoader} path={BUILDER_PATH} />
      <SentryRoute component={AppViewerLoader} path={VIEWER_PATH} /> -> matches here instead
      <SentryRoute component={EditorLoader} path={BUILDER_CUSTOM_PATH} /> 
      <SentryRoute component={AppViewerLoader} path={VIEWER_CUSTOM_PATH} />
```

eco-monk:
@arunvjn 
Do you think we can switch around the order in this method as well? Instead of checking the pageId?
Need to be careful that it doesn't break anything else. 
Because the ordering of checks seems to make or break things.

So I looked to adding checks that work regardless of the order.

The core problem is, we have to deal with paths that match with multiple patterns.

arunvjn:
> @arunvjn Do you think we can switch around the order in this method as well? Instead of checking the pageId? Need to be careful that it doesn't break anything else. Because the ordering of checks seems to make or break things.
> 
> So I looked to adding checks that work regardless of the order.
> 
> The core problem is, we have to deal with paths that match with multiple patterns.

Yes that's right. If you look at the routes declaration file, BUILDER_PATH is declared before BUILDER_CUSTOM_PATH.
Can you investigate how react-router rightly matches BUILDER_CUSTOM_PATH in this case ?

eco-monk:
> 

The conflict is with `VIEWER_PATH` (not `BUILDER_PATH`).
Since `BUILDER_CUSTOM_PATH` is placed before `VIEWER_PATH`. That's why It works fine.

hetunandu (dismissed):

hetunandu (approved):

ramsaptami:
Verified fix in areas where we saw a 404 being returned
- Google Sheet query page
- Google Sheet datasouce pages
- Adding a new GraphQL page

The fix is gtg to release @eco-monk

PR #19857, opened 2023-01-18
feat: Widget responsiveness without css container query

aswathkk:
CSS container query polyfill that we were making use of was found to be breaking the application styles on Firefox.
This POC shows how we can avoid the usage of CSS container queries to implement widget responsiveness


app/client/src/pages/Editor/Canvas.tsx
```diff
@@ -109,7 +110,9 @@ const Canvas = memo((props: CanvasProps) => {
     return (
       <Container
         background={backgroundForCanvas}
-        className="relative mx-auto t--canvas-artboard pb-52"
+        className={`relative mx-auto t--canvas-artboard pb-52 ${getViewportClassName(
+          canvasWidth,
```
aswathkk, line 114: This component already has `canvasWidth` available and is re-rendering whenever the canvas size changes.
That's why I added this extra class here itself.

yaldram (approved):

PR #19914, opened 2023-01-19
fix: selected and triggeredItem Cache

Tooluloope:
## Description
Cache SelectedItem and TriggeredItem

Fixes # (issue)

## Type of change

## How Has This Been Tested?
- Cypress

### Dev activity

### QA activity:


kamakshibhat-appsmith:
Looks good now @Tooluloope

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -683,11 +683,62 @@ class ListWidget extends BaseWidget<
     this.props.updateWidgetMetaProperty("selectedItemIndex", rowIndex);
   };
 
+  handleTriggeredAndSelectedItem = () => {
```
ashit-rath, line 686: Why are we calling this in `componentDidUpdate` but not in `onItemClickCapture`?

Tooluloope: TriggeredItem and Selecteditem are naturally derived Properties. They depend on selectedItemIndex and TriggeredItemIndex.

So `handleTriggeredAndSelectedItem` won't be fired every time. Only when a new selection is triggered.

Also, I had to remove `this.setRowDataCache();` from onItemClickCapture, to avoid race condition.  onItemClick and onItemClickCapture are called at the same time.

And what `this.setRowDataCache()`  uses triggeredIndex and SelectedIndex. So it's best to cache when the index has been updated.

ashit-rath: My only concern was the deep equal `equal(data, selectedItem)` and this being called in `componentDidUpdate`. When do you think doing it in `handleTriggeredAndSelectedItem` would cause a problem?

Tooluloope: the `equal(data, selectedItem)` is to prevent updating selectedItem when the data didn't change.

For example. In Serverside Pagination, it's possible that on navigating back to the page with the selectedItem, its data has been updated from the backend, It's only in such case we'd update the selectedItem, or if we choose a new row.

ashit-rath (approved):

app/client/src/widgets/ListWidgetV2/widget/index.tsx
Tooluloope: > When do you think doing it in handleTriggeredAndSelectedItem would cause a problem?

I'm sure there's not going to be a problem as there's a condition preventing `handleTriggeredAndSelectedItem`.  `shouldUpdateCacheKeys()` helps with this condition.

PR #19915, opened 2023-01-19
feat: Split `/me` to only return basic user information and use `/profile` used in home page for detailed view of the current user

trishaanand:
`/me` has become a very heavy API due to overloading of different functions into a single api since it gets called from all the pages. Splitting this API to make this faster and introducing a new `/profile` endpoint which should be used on the homepage (and only if required, on the edit user profile page.).

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/UserControllerCE.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/UserControllerCE.java
@@ -140,6 +140,12 @@ public class UserControllerCE extends BaseController<UserService, User, String> 
     }
 
     @GetMapping("/me")
+    public Mono<ResponseDTO<User>> getCurrentUser() {
+        return sessionUserService.getCurrentUser()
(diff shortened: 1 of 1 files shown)


sharat87 (commented):
@trishaanand, this is changing the contents of the `/me` response, right? Shouldn't that mean changes on the client too? Can we run an `ok-to-test` please? If Cypress tests pass, I'm fine, and will approve.

trishaanand:
> @trishaanand, this is changing the contents of the `/me` response, right? Shouldn't that mean changes on the client too? Can we run an `ok-to-test` please? If Cypress tests pass, I'm fine, and will approve.

@sharat87 : Hey Shri! ok-to-test is going to fail. I am awaiting changes/exploration on the frontend side to ascertain how many endpoints would be required. Just splitting it into two is not going to work here.

sharat87:
@trishaanand, okay, sorry then I'm further confused. I see that you've requested-for-review on this PR, but going by your message, I don't think this is ready for review? Can you help me understand what you need my review thoughts on please? 🤔

PR #19944, opened 2023-01-20
fix: Redis rdb persistance | User session persistance

### LINKED ISSUE #19768: [Bug]: Login Sessions Not Persisting After Updating Appsmith Version

### Description

For a self-hosted user, when a version update occurs, all the users' sessions are ended. As some of their apps are embed they users don't get the notification that they should refresh the page and only get a blank page. This happens at every update. Actually, if they stop the container and start it again all the users loose the session.
The user is using a fat Appsmith container in Docker. The don't use any external MongoDB or Redis.
Appsmith session cookies don't have an expiry at all today, and the version update should not cause logouts. 

[Discord thread](https://discord.com/channels/725602949748752515/1060864335842377748)

[Front thread](https://app.frontapp.com/open/msg_1kmlm9ho?key=https://app.frontapp.com/open/cnv_vb1juoc?key=XehLBPpwY-FI8hbZXfxIWfqFiX6Xk6Oq)

### Steps To Reproduce

The user is using a fat Appsmith container in Docker. The don't use any external MongoDB or Redis.

### Public Sample App

_No response_

### Issue video log

_No response_

### Version

Self-Hosted

Front logo [Front conversations](https://app.frontapp.com/open/top_6sado)

sum35h:
## Description
- Updated redis.conf to set the Redis rdb directory to `/appsmith-stacks/data/redis` bind-mount directory, so that the data is persisted after the container restarts/updates. ( This was causing user sessions on Appsmith installations using the embedded Redis to reset after every container restart, since the redis data was not being persisted)
- Updated redis.conf for the embedded Redis to increase the snapshot frequency to dump after every 15 sec if at least 1 key changed

Fixes #19768

## Type of change

## How Has This Been Tested?
- Manual

### DIFF
--- a/deploy/docker/entrypoint.sh
+++ b/deploy/docker/entrypoint.sh
@@ -248,8 +248,13 @@ configure_supervisord() {
     fi
     if [[ $APPSMITH_REDIS_URL == *"localhost"* || $APPSMITH_REDIS_URL == *"127.0.0.1"* ]]; then
       cp "$SUPERVISORD_CONF_PATH/redis.conf" /etc/supervisor/conf.d/
+      # Initialize Redis rdb directory
+      REDIS_DB_PATH="$stacks_path/data/redis"
+      mkdir -p "$REDIS_DB_PATH"
       # Enable saving Redis session data to disk more often, so recent sessions aren't cleared on restart.
-      sed -i 's/^# save 60 10000$/save 60 1/g' /etc/redis/redis.conf
+      sed -i 's/^save 60 10000$/save 15 1/g' /etc/redis/redis.conf
+      # Updating the default Redis rdb directory
+      sed -i "s|^dir /var/lib/redis$|dir ${REDIS_DB_PATH}|g" /etc/redis/redis.conf
     fi
     if ! [[ -e "/appsmith-stacks/ssl/fullchain.pem" ]] || ! [[ -e "/appsmith-stacks/ssl/privkey.pem" ]]; then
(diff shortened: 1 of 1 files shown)


deploy/docker/entrypoint.sh
```diff
@@ -248,8 +248,13 @@ configure_supervisord() {
...
     if [[ $APPSMITH_REDIS_URL == *"localhost"* || $APPSMITH_REDIS_URL == *"127.0.0.1"* ]]; then
       cp "$SUPERVISORD_CONF_PATH/redis.conf" /etc/supervisor/conf.d/
+      # Initialize Redis rdb directory
+      REDIS_DB_PATH="$stacks_path/data/redis"
+      mkdir -p "$REDIS_DB_PATH"
       # Enable saving Redis session data to disk more often, so recent sessions aren't cleared on restart.
-      sed -i 's/^# save 60 10000$/save 60 1/g' /etc/redis/redis.conf
+      sed -i 's/^save 60 10000$/save 15 1/g' /etc/redis/redis.conf
+      # Updating the default Redis rdb directory
+      sed -i "s|^dir /var/lib/redis$|dir ${REDIS_DB_PATH}|g" /etc/redis/redis.conf
```
sharat87, line 257: So, has this default configuration changed in Ubuntu's standard Redis? :(

Also, you can combine these two `sed` expressions like this:

```bash
      sed -i \
        -e 's/^save 60 10000$/save 15 1/g' \
        -e "s|^dir /var/lib/redis$|dir ${REDIS_DB_PATH}|g" \
        /etc/redis/redis.conf
```

PR #16904, opened 2022-09-20
fix: Table filter issues

souma-ghosh:
## Description

**Bug 12638**
After adding 3 filters to a table and removing the second filter, the current second filter `value` was showing the old second filter `value`. This was happening because we were not passing unique keys to the filters while rendering the filters list.
Hence added a unique key to filter in filters list.

**Bug 12642**
After adding 3 filters and then changing the operator of second filter from `AND` to `OR` or vice versa, then removing the **first** filter caused the operator to go back to the previous value.
This was happening because we are adding the `OR` operator to a filter by default. When we are changing the operator from the dropdown the value gets updated to the second filter only in the state. All other filters persisted with the previous operator value.
Hence this has been solved by updating the operator of all filters (after the second filter) whenever user changes the operator of the second filter from the operator dropdown.

Fixes #12642, #12638

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- Cypress tests

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

Issue #12638, opened 2022-04-06
[Bug]-[50]:Table Filter - Removing where clause does not rightly update the input text value but filtered result is fine


Aishwarya-U-R:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

https://user-images.githubusercontent.com/91450662/162014415-d2238737-a184-4563-9c91-d46dd5ad2904.mov

### Steps To Reproduce

1. Drag drop table with some table data
2. Filter table with some input text
3. Add one more filter with different input text
4. Remove 1st filter added
5. **Observe : Where clause input text is now not updated rightly**
6. **Expected:  Where clause input text should show respective filter value**

### Version

Cloud/Release v1.6.17

dilippitchika:
Great catch @Aishwarya-U-R 🏅 , we wouldn't have caught it. Probably there are some other items here which are undetected.

dilippitchika:
## Stats
| Stat | Values |
| ------------- | ------------- |
| Reach     | 50 |
| Effort (months)  | 0.5 |

souma-ghosh:
Fixed in PR
https://github.com/appsmithorg/appsmith/pull/16904

PR #19951, opened 2023-01-20
fix: Table filter text loses focus when user is typing

### LINKED ISSUE #18040: [Bug]: Table Filter Text Loses Focus When User Is Typing

### Description

I have an appsmith app that has a table view with filtering enabled. If a user tried to filter one of the text fields (eg `Address contains XYZ`), the filter text input loses focus after a short delay after the user stops typing. This is causing UX issues for our users because if they type slowly, they have to click back into the text field to continue typing. It seems possible that the field is losing focus on debounce, tho we are not 100% sure. See the example below

appsmith-table-filters

### Steps To Reproduce

1. create a table widget with a text field
2. enable filtering
3. try to filter text column and stop typing for a fraction of a second.

### Public Sample App

_No response_

### Version

Appsmith Community v1.8.0

Front logo [Front conversations](https://app.frontapp.com/open/top_6jdzg)

souma-ghosh:
## Description

TL;DR
When user types in the input box for table filter, the input loses focus after a 500ms while typing. This is a regression from https://github.com/appsmithorg/appsmith/pull/16904 PR (bug - https://github.com/appsmithorg/appsmith/issues/12638) 

We were stringifying the whole filter object and using it in the key for each table filter. But when we changed the filter value the key would get updated and hence the filter re-rendered and the text input lost focus.
The key was previously set this way to address another bug where the value field of the filter would not update in the DOM although the filter data was correct.
The solution to address both these issues was to assign a unique key to the filters. Since the filter data is totally front-end generated we do not have any unique id to use here as the key. So we generate unique id for each filter now and use it as the key.

Fixes #18040 

## Type of change

## How Has This Been Tested?

- Manual

### Dev activity

### QA activity:


sbalaji1192 (approved):

kamakshibhat-appsmith:
@souma-ghosh The table crashes as soon as we remove dummy data from Data items .

sbalaji1192 (approved):

souma-ghosh:
@laveena-en 
Page number should not reset when clearing a filter using Clear All btn or clearing a single filter with the cross btn in filter pane. Here is the loom explaining the behaviour using Clear All btn
https://www.loom.com/share/66019a500b4d4420b4c813823287897d

laveena-en:
@souma-ghosh 
On adding an partial evaluation, the results being fetched are incorrect: https://www.loom.com/share/f051a226e2484f3db350c6e5cc882d28

Also, unable to navigate to the next page when a filter is applied:
https://www.loom.com/share/515c6e815a7640eaab504ae7ce0f16ac

souma-ghosh:
> @souma-ghosh On adding an partial evaluation, the results being fetched are incorrect: https://www.loom.com/share/f051a226e2484f3db350c6e5cc882d28
> 
> Also, unable to navigate to the next page when a filter is applied: https://www.loom.com/share/515c6e815a7640eaab504ae7ce0f16ac

For the first one, the issue exists on release and prod. We should create an issue for this.
For the second one, if server side pagination is not integrated then this behaviour is correct as the number of results is less. That's why the the page count says 1 and the next button is disabled. Please correct me if I am wrong here.

laveena-en:
Created a separate issue for the partial condition in filters: https://github.com/appsmithorg/appsmith/issues/20435

laveena-en:
Tested and verified.

PR #19963, opened 2023-01-21
feat: added models for export

sidhantgoel:
Added models for export, this will be used in Import/Export V2 service.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/models/export/ActionMetadata.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/models/export/ActionMetadata.java
@@ -1,0 +1,29 @@
+package com.appsmith.server.modals;
+
+import java.util.Set;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
+import lombok.Data;
+
+@Data
+public class ActionMetadata {
+    @JsonIgnore
+    private String name;
+    private String pluginId;
+    private String fullyQualifiedName;
+    private String datasourceName;
+    private String datasourcePluginId;
+    private String pageId;
+    private String collectionId;
+    private ActionConfigurationMetadata actionConfiguration;
+    private Boolean executeOnLoad;
+    private Boolean clientSideExecution;
+    private Set<String> jsonPathKeys;
+    private Boolean userSetOnLoad = false; //TODO check this field
+    private Boolean confirmBeforeExecute = false;
+    @JsonIgnore
+    private String body;
+    private boolean deleted;
+    private String gitSyncId;
+}
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/models/export/ActionMetadata.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/models/export/ActionMetadata.java
@@ -1,4 +1,4 @@
-package com.appsmith.server.modals;
+package com.appsmith.server.models.export;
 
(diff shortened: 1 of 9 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/models/export/ActionMetadata.java
```diff
@@ -0,0 +1,29 @@
+package com.appsmith.server.modals;
```
mohanarpit, line 1: Is this a typo in the package name? Is this code compiling for you on your local? I see this typo in multiple files in this diff. Can you please check once?

app/server/appsmith-server/src/main/java/com/appsmith/server/models/export/ThemeMetadata.java
```diff
@@ -0,0 +1,11 @@
+package com.appsmith.server.models.export;
+
+import lombok.Data;
+
+@Data
+public class ThemeMetadata {
```
mohanarpit, line 6: Are these models going to be persisted in DB? Or are they more like DTOs? If they are going to be persisted in the DB, does it make sense to extend `BaseDomain` so that we get the fields for `deletedAt`, `deleted`, etc automatically?

app/server/appsmith-server/src/main/java/com/appsmith/server/models/export/ActionMetadata.java
sidhantgoel: This is a typo, corrected

app/server/appsmith-server/src/main/java/com/appsmith/server/models/export/ThemeMetadata.java
sidhantgoel: These models are for persistence, but for export and version control (this is how it will be saved in repository). There were no common properties so didn't extracted base class.

mohanarpit: Can you see if `BaseDomain` has the fields you need for all the persisted entities? I'm asking this because the idea of BaseDomain was to have uniformity of fields like createdAt, deletedAt etc.

sidhantgoel: Also we wanted to use Jackson instead of Gson for efficiency, DSO's are not compatible with Jackson.

sidhantgoel: We don't want any fields from the BaseDomain

sidhantgoel: BaseDomain is primarily for storage in MongoDB. This is completely new use case.

mohanarpit: If not in MongoDB, where are these entities being persisted then? File system? Do we want to keep a consistent data format for filesystem entities & DB entities?

Apologies for the noob questions, but I'm trying to understand the use case here.

mohanarpit (approved):

sidhantgoel:
This is temporary, it will be replaced by a method using which we can have single model and multiple views, so we don't have to define property at two places.

PR #19997, opened 2023-01-24
fix: Fix replicaset check to not require ClusterMonitor role

sharat87:
In the `entrypoint.sh` script, we check if the MongoDB in use, has replicaSet initiated or not. This is usually done with a `rs.initiate()` on the cluster.

We need the replicaSet to be enabled on MongoDB, since the backend server relies on MongoDB `changeStream`s, which is a feature, only available if replicaSet is enabled.

However, to use the `changeStream` APIs, having the `read` or `readWrite` role on MongoDB is enough. But the check we do in `entrypoint.sh`, runs `rs.status()` to see if `replicaSet` is initiated. This `rs.status()` call, unfortunately, requires the `ClusterMonitor` role, unlike the `changeStream` API.

To tackle this, we created the `appsmithctl check_replica_set` command. This command would attempt to use the `changeStream` API, and report success or failure. But this failed on certain configurations, where MongoDB was running as a single-node-cluster, on localhost, or a local/internal network. This was an edge case.

That edge case is solved by this PR. With this, we can use `appsmithctl check-replica-set` in the `entrypoint.sh` again.


sharat87:
@sum35h or @mohanarpit, can I get a review on this please?

sum35h (approved):

PR #19999, opened 2023-01-24
feat: Responsive Checkbox & Switch Widgets

aswathkk:
Sets minWidth to Checkbox & Switch widgets.
Label overflow behaviour updated for autoLayout


app/client/src/widgets/CheckboxWidget/index.ts
```diff
@@ -32,6 +32,18 @@ export const CONFIG = {
     responsiveBehavior: getDefaultResponsiveBehavior(Widget.getWidgetType()),
     minWidth: FILL_WIDGET_MIN_WIDTH,
```
yaldram, line 33: Do we need this `minWidth` here ?

yaldram (changes requested):

PR #20001, opened 2023-01-24
fix: Change watchtower to latest

sharat87:
We're `watchtower:latest-dev` image for the `--schedule` support. This was only available in the `latest-dev` version of Watchtower.

Now, this feature is available in the `latest` version of Watchtower, so we should be fine switching back to `latest`.

More importantly, the `latest-dev` version of Watchtower, doesn't have ARM images built. It only has AMD images built today. So this watchtower service fails to start on ARM architectures. This PR fixes that.

### DIFF
--- a/deploy/aws_ami/docker-compose.yml
+++ b/deploy/aws_ami/docker-compose.yml
@@ -17,7 +17,7 @@ services:
     #  com.centurylinklabs.watchtower.enable: "true"
 
   #auto_update:
-  #  image: containrrr/watchtower:latest-dev
+  #  image: containrrr/watchtower
   #  volumes:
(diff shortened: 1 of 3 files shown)


sharat87:
@sum35h, I think this'll need the AWS AMI to be published again, let's pick that up once this is merged. Thanks! 🙏

sum35h:
@sharat87  This change will not need any of the images to be updated, since those instances themselves pull the docker-compose.yaml from [bitly](https://bit.ly/3h08WqB)
On that note, should @Devops take up ownership of this bitly account going forward?

sum35h (approved):

PR #20086, opened 2023-01-25
feat: migrate JsonIgnore to JsonView

sidhantgoel:
In summary, the change comprises adding `JsonView` annotations on domain/model fields, controller methods, and arguments.
applying JsonView annotation on controller methods will include fields which are marked with the same view in domain objects.
Similarly, usage on controller method arguments will only deserialize the fields which are marked with same view in Domain object, which is helpful in case we don't want the client to send some fields for security reasons.
This change replaces the previous use of JsonIgnore for restricting fields in API response and allows for more flexibility and will enable us to have fine-grain control over fields serialized for different contexts, such as API response, export, and import.
The following views are defined as of now.
`View.Internal` - View used to serialize for internal use. It inherits from `Views.Public` and `Views.Export`, so it will also include those fields.
`Views.Public` - View used for API request/response.
`Views.Export` - View used for making field exportable, like in the case of Git sync and file export.
Also, we can define more views by adding an interface inside the Views class and Views can also inherit fields marked with different view by simple interface inheritance.

A small tutorial on JsonView - [Jackson JSON Views | Baeldung](https://www.baeldung.com/jackson-json-view-annotation)


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/Views.java
```diff
@@ -0,0 +1,21 @@
...
+    /**
+     * Used with JsonView annotation to denote that a field should be visible to the client.
+     */
+    interface Public {}
+
+    /**
+     * Used with JsonView annotation to denote that a field should be exported when exporting the Domain/DTOs.
+     */
+    interface Export {}
+}
```
abhvsn, line 21: Does it make sense to have a separate `view` package under `com.appsmith.server` and not mix with models?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/dtos/ExecuteActionDTO.java
```diff
@@ -15,41 +18,50 @@
...
+     * "k4": {
+     * "array": [
+     * "string",
+     * "number",
+     * "string",
+     * "boolean"
+     * ]
+     * },
+     * "k5": "boolean"
+     * }
```
abhvsn, line 48: With earlier indentation, it's more easy to understand the structure.

abhvsn (commented):
@sidhantgoel QQ: With JsonView we should not require separate models for export. If this assumption is correct why are we maintaining these models as well.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/dtos/ExecuteActionDTO.java
sidhantgoel: Auto format did this, will revert

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/Views.java
sidhantgoel: > @sidhantgoel QQ: With JsonView we should not require separate models for export. If this assumption is correct why are we maintaining these models as well.

Which model you are talking about?

sidhantgoel: We can have a separate package but that will contain only single class, I don't think it is a problem though, will move this

abhvsn: > > @sidhantgoel QQ: With JsonView we should not require separate models for export. If this assumption is correct why are we maintaining these models as well.
> 
> Which model you are talking about?

@sidhantgoel I was referring to models under `app/server/appsmith-server/src/main/java/com/appsmith/server/models/export` package.

abhvsn:
LGTM 
@sidhantgoel can you please resolve the merge conflicts before I approve?

nayan-rafiq (approved):

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ActionCollectionServiceImplTest.java
```diff
@@ -453,10 +454,10 @@ public void testUpdateUnpublishedActionCollection_withInvalidId_throwsError() th
     @Test
     public void testUpdateUnpublishedActionCollection_withModifiedCollection_returnsValidCollection() throws IOException {
         ObjectMapper objectMapper = new ObjectMapper();
-        final JsonNode jsonNode = objectMapper.readValue(mockObjects, JsonNode.class);
-        final ActionCollection actionCollection = objectMapper.convertValue(jsonNode.get("actionCollectionWithAction"), ActionCollection.class);
-        final ActionCollectionDTO modifiedActionCollectionDTO = objectMapper.convertValue(jsonNode.get("actionCollectionDTOWithModifiedActions"), ActionCollectionDTO.class);
-        final ActionCollection modifiedActionCollection = objectMapper.convertValue(jsonNode.get("actionCollectionAfterModifiedActions"), ActionCollection.class);
+        final JsonNode jsonNode = objectMapper.readerWithView(Views.Public.class).readValue(mockObjects, JsonNode.class);
+        final ActionCollection actionCollection = objectMapper.readerWithView(Views.Public.class).readValue(objectMapper.writerWithView(Views.Public.class).writeValueAsString(jsonNode.get("actionCollectionWithAction")), ActionCollection.class);
```
nidhi-nair, line 458: Nit, could you please create a utility function for this? I think we'll end up writing this kind of representation for a lot of our tests moving forward.

app/server/appsmith-server/src/main/java/com/appsmith/server/models/export/ActionConfigurationMetadata.java
```diff
@@ -1,64 +0,0 @@
-package com.appsmith.server.models.export;
```
nidhi-nair, line 1: QQ for my understanding, why are we removing all these classes?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/views/Views.java
```diff
@@ -0,0 +1,21 @@
...
+
+    /**
+     * Used with JsonView annotation to denote that a field should be visible to the client.
+     */
+    interface Public {}
+
+    /**
+     * Used with JsonView annotation to denote that a field should be exported when exporting the Domain/DTOs.
+     */
+    interface Export {}
```
nidhi-nair, line 20: I think you mentioned this before, but will consumption of this view be part of another PR?

Parthvi12:
Bugs filed during sanity testing:

Parthvi12:
Bugs found while sanitty testing:


- [ ] https://github.com/appsmithorg/appsmith/issues/21309


- [ ] On running actions such as JSObject/query gives internal server error

Screenshot 2023-03-02 at 2 16 45 PM

Parthvi12:
https://github.com/appsmithorg/appsmith/issues/21309 is observed on prod as well, hence not a blocker for this PR

nayan-rafiq (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/models/export/ActionConfigurationMetadata.java
sidhantgoel: These classes were used temporarily, now since we have multiple views, existing classes can be used.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/views/Views.java
sidhantgoel: Yes

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ActionCollectionServiceImplTest.java
nayan-rafiq: @nidhi-nair I've created a utility class to wrap the reading and writing methods of object mapper. Would you please review this once again?

nidhi-nair (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ObjectMapperUtils.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public <T> T readFromFile(File file, Class viewsClass, Class<T> resultClass) throws IOException {
+        return objectMapper
+                .readerWithView(viewsClass)
+                .readValue(file, resultClass);
+    }
+
+    public String writeAsString(Object src, Class viewsClass) throws JsonProcessingException {
+        return objectMapper
+                .writerWithView(Views.Public.class)
```
nidhi-nair, line 31: Oops looks like this method is not using the parameter?

nayan-rafiq: Sorry, I missed that. Thanks a lot for your review.

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ObjectMapperUtils.java
```diff
@@ -6,7 +6,12 @@
 import java.io.File;
 import java.io.IOException;
 
-public record ObjectMapperUtils(ObjectMapper objectMapper) {
+public class ObjectMapperUtils {
+    private final ObjectMapper objectMapper;
```
nidhi-nair, line 10: Should we autowire this from the default? There's a few settings there that might come in handy. Includes handling some types like `Instant` and such.

nayan-rafiq: @nidhi-nair we can do that in future if required. As of now, we don't need to do that because it's only used in tests where an external objectmapper is used.

nidhi-nair (approved):

nidhi-nair (approved):

Issue #18829, opened 2022-12-09
[Bug]: Segment adding to the script time


SatishGandham:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Segment is sending lot of tracking events, and adding 5-20% to the scripting time.
Review how we are using it and what we are tracking and remove unnecessary events.

Slack thread: https://theappsmith.slack.com/archives/C024GUDM0LT/p1670261140997809

### Steps To Reproduce

Record a performance profile, and in the call tree search for track and see how long it takes.

### Version

Both

somangshu:
@hetunandu pointed out 
| It is easy to figure this out. On mix panel, you can find the top events fired, find them on our codebase and then comment them out.

### To-do
- [ ] Find all the top events running on segment. @kocharrahul7 can you help us here?
- [ ] Identify the events that are not required
- [ ] Comment out those events on check the performance improvements
    - [ ] Report back the numbers
    - [ ] Final cleanup and Ship

kocharrahul7:
https://mixpanel.com/project/2032485/view/178891/app/dashboards#id=1658069&edited-bookmark=QvmPBjv1dAVr

Rhitottam:
@somangshu , @SatishGandham  Can check the following readings after checking the performance while running the following actions in the Performance Infra Dashboard in local environment
`Action: Reload Release Master Comparison Page of Performance Infra Dashboard and wait for the table data to load completely..
        This action involves loading a table containg data retrieved from two mysql apis
Reading 1: 
  Total Scripting Time: 3510
  Track Scripting Time: 40.9
  Percentage: 1.1%
Reading 2:
  Total Scripting Time: 3729
  Track Scripting Time: 30.4
  Percentage: 0.8% 
After Disabled Logging of events - BINDING_SUCCESS, DEBUGGER_NEW_ERROR, DEBUGGER_NEW_ERROR_MESSAGE
Reading 1:
  Total Scripting Time: 3499
  Track Scripting Time: 20.2
  Percentage: 0.8%
Reading 2:
  Total Scripting Time: 3530
  Track Scripting Time: 26.2
  Percentage: 0.7%

Action: Navaigate to PR Details page of Performance Infra Dashboard and wait for the table data to load completely.
        This action involves loading a table containg data from multiple apis and loading a chart displaying metrics data for multiple PRs
Reading 1: 
  Total Scripting Time: 7879
  Track Scripting Time: 178.8
  Percentage: 2.2%
[...]

PR #20087, opened 2023-01-25
fix: Add batching strategy to segment analytics tracking so that events are sent to segment in batches instead of individual

Rhitottam:
> Fixes https://github.com/appsmithorg/appsmith/issues/18829

> From Analytics 2.0 onwards, batching to deliver of multiple events with fewer API calls for faster performance
[https://segment.com/blog/analytics-js-2/](https://segment.com/blog/analytics-js-2/)
[https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/#batching](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/#batching)

## Description

> Add batching strategy to segment analytics while loading the script inside AnalyticsUtil
> [https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/#batching](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/#batching)

Why are we doing this, i.e, exact issue with the performance?
 - Whenever any operation is done inside Appsmith, multiple events are being tracked and sent to Segment for tracking. Due to this a lot apis calls are made for each event being tracked using the segment api. This can cause performance issues since for every event being tracked, segment runs a microtask for updating cookies  which only start running when the api is being called. Through batching these events in groups of 100, all microtasks for those events in a group can run without any waiting time in between, thus improving performance slightly.
Screenshot 2023-01-26 at 10 39 09 AM
Screenshot 2023-01-26 at 10 50 04 AM

What are we doing exactly?
- Adding a batching strategy to the Segment config so that events are tracked and delivered in batches of 100 instead of individually delivering them to Segment

Which areas of the application will be effected?
- This will affect the timing of the events being tracked by Segment. Because of this events in a batch of 100 will be received at the same time rather than then sequentially

Will the users see any effects?
- User will see a slight to moderate improvement in performance while doing operations like loading a page in Edit mode or reloading an application in Edit mode

 Fixes https://github.com/appsmithorg/appsmith/issues/18829

## Type of change

- Manual

### Dev activity

### QA activity:

### DIFF
--- a/app/client/src/utils/AnalyticsUtil.tsx
+++ b/app/client/src/utils/AnalyticsUtil.tsx
@@ -385,6 +385,7 @@ class AnalyticsUtil {
             resolve(false);
           }, 2000);
           analytics.SNIPPET_VERSION = "4.1.0";
+          // Ref: https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/#batching
           analytics.load(key, {
             integrations: {
               "Segment.io": {
--- a/app/client/src/utils/AnalyticsUtil.tsx
+++ b/app/client/src/utils/AnalyticsUtil.tsx
@@ -390,10 +390,10 @@ class AnalyticsUtil {
             integrations: {
               "Segment.io": {
                 deliveryStrategy: {
-                  strategy: "batching",
+                  strategy: "batching", // The delivery strategy used for sending events to Segment
                   config: {
-                    size: 100,
-                    timeout: 5000,
+                    size: 100, // The batch size is the threshold that forces all batched events to be sent once it’s reached.
+                    timeout: 5000, // The number of milliseconds that forces all events queued for batching to be sent, regardless of the batch size, once it’s reached
                   },
                 },
               },
--- a/app/client/src/utils/AnalyticsUtil.tsx
+++ b/app/client/src/utils/AnalyticsUtil.tsx
@@ -393,7 +393,7 @@ class AnalyticsUtil {
                   strategy: "batching", // The delivery strategy used for sending events to Segment
                   config: {
                     size: 100, // The batch size is the threshold that forces all batched events to be sent once it’s reached.
-                    timeout: 5000, // The number of milliseconds that forces all events queued for batching to be sent, regardless of the batch size, once it’s reached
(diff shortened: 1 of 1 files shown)


sbalaji1192 (approved):

riodeuno (commented):
@Rhitottam This looks good to me. Please add comments in the code to describe what each of the options do, particularly, `strategy`, `size` and `timeout`. Also a link to the API docs would be helpful. This way, I would quickly understand what these are in a year's time 😅

riodeuno:
@Rhitottam could you also mention in the PR description the answer to the following questions:
1. Why are we doing this, i.e, exact issue with the performance?
2. What are we doing exactly?
3. Which areas of the application will be effected?
4. Will the users see any effects?

Rhitottam:
Hi @riodeuno , I have the code comments and added an explanation for the changes done.

riodeuno (dismissed):

riodeuno:
@Rhitottam This is perfect. Thanks!. Looks good to me. I've approved the PR.

app/client/src/utils/AnalyticsUtil.tsx
```diff
@@ -385,7 +385,20 @@ class AnalyticsUtil {
...
-          analytics.load(key);
+          // Ref: https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/#batching
+          analytics.load(key, {
+            integrations: {
+              "Segment.io": {
+                deliveryStrategy: {
+                  strategy: "batching", // The delivery strategy used for sending events to Segment
+                  config: {
+                    size: 100, // The batch size is the threshold that forces all batched events to be sent once it’s reached.
+                    timeout: 5000, // The number of milliseconds that forces all events queued for batching to be sent, regardless of the batch size, once it’s reached
```
SatishGandham, line 396: Change the batching to one second.

SatishGandham (changes requested):

app/client/src/utils/AnalyticsUtil.tsx
Rhitottam: Done!!

eco-monk (approved):
Looks good to me.

SatishGandham (approved):

PR #20092, opened 2023-01-25
fix: Modal dropped inside a widget getting limited to parent size

aswathkk:
Regardless of where we drop the Modal widget, its parent should be MAIN_CONTAINER_WIDGET_ID
fixes: #19601


prsidhu (approved):
Tested. Looks good.

PR #20121, opened 2023-01-27
fix: renamed rts port env to APPSMITH_RTS_PORT

sum35h:
Issue: Nginx and RTS used the same env PORT for binding it's service, while the backend server had the rts port hardcoded on its rts uri.
- Renamed env PORT to APPSMITH_RTS_PORT for starting the rts server.
- Updated nginx config templates to use env `APPSMITH_RTS_PORT`
- Added appsmith.rts.port property in server to use env APPSMITH_RTS_PORT
- Updated CommonConfig.java rtsBaseDomain to use appsmith.rts.port

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/CommonConfig.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/CommonConfig.java
@@ -67,7 +67,7 @@ public class CommonConfig {
     @Value("${appsmith.rts.port:8091}")
     private String rtsPort;
 
-    private String rtsBaseDomain = "http://127.0.0.1:" + rtsPort;
+    private String rtsBaseDomain;
 
     private List<String> allowedDomains;
 
@@ -131,5 +131,8 @@ public class CommonConfig {
         // If `true`, then disable signup. If anything else, including empty string, then signups will be enabled.
         isSignupDisabled = "true".equalsIgnoreCase(value);
     }
-
+    
+    public String getRtsBaseDomain() {
+     return "http://127.0.0.1:" + rtsPort;
+}
 }
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/CommonConfig.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/CommonConfig.java
@@ -133,6 +133,6 @@ public class CommonConfig {
     }
     
     public String getRtsBaseDomain() {
-     return "http://127.0.0.1:" + rtsPort;
-}
+        return "http://127.0.0.1:" + rtsPort;
+    }
 }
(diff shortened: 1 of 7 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/CommonConfig.java
```diff
@@ -128,5 +131,8 @@ public void setSignupDisabled(@Value("${signup.disabled}") String value) {
         // If `true`, then disable signup. If anything else, including empty string, then signups will be enabled.
         isSignupDisabled = "true".equalsIgnoreCase(value);
     }
-
+    
+    public String getRtsBaseDomain() {
+     return "http://127.0.0.1:" + rtsPort;
+}
```
sharat87, line 137: Indentation.

```suggestion
    public String getRtsBaseDomain() {
        return "http://127.0.0.1:" + rtsPort;
    }
```

sharat87 (commented):
Code formatting fix, otherwise looks good.

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/CommonConfig.java
```diff
@@ -128,5 +131,8 @@ public void setSignupDisabled(@Value("${signup.disabled}") String value) {
         // If `true`, then disable signup. If anything else, including empty string, then signups will be enabled.
         isSignupDisabled = "true".equalsIgnoreCase(value);
     }
-
+    
+    public String getRtsBaseDomain() {
```
sharat87, line 135: Sorry, one more tiny change. Can we change this to `rtsBaseUrl` please? It doesn't contain any domain, it's a base URL, and this will cause confusion later. 🙂

sharat87 (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/CommonConfig.java
```diff
@@ -64,7 +64,10 @@ public class CommonConfig {
     @Value("${disable.telemetry:true}")
     private boolean isTelemetryDisabled;
 
-    private String rtsBaseDomain = "http://127.0.0.1:8091";
+    @Value("${appsmith.rts.port:8091}")
+    private String rtsPort;
+
+    private String rtsBaseDomain;
```
sharat87, line 70: Is this variable used anywhere? Can we remove it if it's unused?

sharat87 (changes requested):

mohamedgomran:
Hello @sharat87 👋  When this can be released in docker image appsmith-ce?

sum35h:
@mohamedgomran  Unfortunately this fix will only be out on Release v1.9.9

PR #20118, opened 2023-01-26
feat: Non auto height invisible widgets

riodeuno:
## Description
This PR adds another feature update we had planned for Auto Height
- [ ] For new applications, in View and Preview mode, any widget which is invisible will let go of its space and collapse if it's either on the main Canvas or a container-like widget which has Auto-height enabled.
- [ ] Widgets within a container-like Widget, say Tabs, that doesn't have Auto-height enabled, will now let go of their space if they're invisible.
- [ ] The experience in Edit mode has not changed.

TL;DR: In new applications, in the Preview and Published _AKA_ View modes, if a widget is invisible and within an Auto-height-enabled container like a Tab, a Modal, a Form, or the main Canvas, it will fully collapse, allowing widgets below it to move up and take its space. This changes the behavior today prior to the release of this PR for Auto-height-enabled widgets.

Fixes #19983
Fixes #18681

#### Before

https://user-images.githubusercontent.com/103687/214884767-0e7b0984-329b-4c99-8cf2-38a8aecc51d9.mov

#### After

https://user-images.githubusercontent.com/103687/214884513-55345213-9402-4f2f-b588-ea6de3094de8.mov

## Type of change
- New feature (non-breaking change which adds functionality)
- This change requires a documentation update

## How Has This Been Tested?
- Manual: See Test Plan for details
- Jest: New functions have jest tests
- Cypress: We check the positions of the widgets after trying to collapse a non-auto height widget within an auto height enabled container

### Test Plan
https://github.com/appsmithorg/TestSmith/issues/2176

### Issues raised during DP testing
https://github.com/appsmithorg/appsmith/pull/20118#issuecomment-1409937455
https://github.com/appsmithorg/appsmith/pull/20118#issuecomment-1410389053

### Dev activity
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [NA] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [NA] PR is being merged under a feature flag

### QA activity:
- [x] Test plan has been approved by relevant developers
- [x] Test plan has been peer reviewed by QA
- [ ] Cypress test cases have been added and approved by either SDET or manual QA
- [ ] Organized project review call with relevant stakeholders after Round 1/2 of QA
- [ ] Added Test Plan Approved label after reveiwing all Cypress test

PR #20142, opened 2023-01-27
feat: Collapse invisible widgets Field for Auto-height

### LINKED ISSUE #19983: [Feature]: Collapse all invisible widgets in view and preview mode

### Summary

- When a widget within a container -- which has auto height enabled -- goes invisible -- in view and preview mode -- they should fully collapse, allowing widgets beneath them to take up the vacated space.
- We cannot have this enabled by default without consequences to existing apps. The reason for this, is that this will effect exiting page layouts and cause unexpected behaviour without the knowledge of the application developer.
- To mitigate this, we should enable this only for newly created applications. This way, application developers will notice the behaviour during development.
- This requires the application API response, to provide the client with a flag that informs the client of the fact that the application is a newly created application, allowing the client to enable this feature without consequences like breaking apps.
- The application API response can have a flag `collapseInvisibleWidgets: true`.
- If this flag exists and is `true`, the client will collapse all invisible widgets within an auto height enabled container.
- If this flag doesn't exist or is `false`, the client will not collapse all invisible widgets within an auto height enabled container.

- Today, widgets which have auto height enabled, will collapse, allowing widgets below them to occupy their space.
- However, not all widgets have auto height. This leads to widgets like Button and Divider, always taking up space.
- Collapsing invisible widgets is a basic expectation when it comes to web UI.

NilanshBansal:
## Description

> This PR implements the backend APIs for Invisible widgets in Auto-height containers. It adds a new field `collapseInvisibleWidgets` which is set to True by default, when creating a new application.
> Notion doc here: https://www.notion.so/appsmith/Invisible-widgets-in-auto-height-containers-460c6bf4e0a342e4a1fa4955f7f6c461
> Frontend feature implementation here: https://github.com/appsmithorg/appsmith/pull/20118

Fixes #19983 

## Type of change

## How Has This Been Tested?

- Manual
- JUnit

### Dev activity

### QA activity:


app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Application.java
```diff
@@ -123,6 +123,8 @@ public class Application extends BaseDomain {
     @JsonIgnore
     AppPositioning unpublishedAppPositioning;
 
+    Boolean collapseInvisibleWidgets;
```
sumitsum, line 126: Can you please check what the behaviour would be if someone tries to import an application which does not have this attribute (i.e. the application was exported before this change was introduce) ? We might want to assign a default value for this attribute in `importApplicationInWorkspace` method as well.

sumitsum, line 126: I could not find any place where this is set to `False` via any user triggered action or otherwise - is this change supposed to be added later ?

NilanshBansal: @sumitsum for all new applications that will be created going forward this field will be set only to True. 
This is a default functionality that has to be maintained going forward with auto height coming in.
Why this has been Introduced?? --> To avoid hardcoding at the frontend and have flexibility in future, if the user wants to explicitly turn this on/off, but for now this always is set to True. 
For old applications, this field doesn't exist and hence the Frontend considers it as False by default, so that older applications work seamlessly as earlier.

NilanshBansal: If someone wants to Import any older applications, the attribute remains as it is.

sumitsum: Just thinking out loud, if no handling for this case is added, then the variable may be assigned a `null` value which may be fine in this case - in case there is a default handling for null.

NilanshBansal: @sumitsum as confirmed with @riodeuno, the import application will remain as the old structure, so, If an old application is imported it should work as per the old configs only.

sumitsum (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Application.java
riodeuno: @sumitsum @NilanshBansal This is under the assumption that we can change this behaviour, easily, if necessary, based on user feedback.

PR #20179, opened 2023-01-28
fix: QA callouts for list v2

### LINKED ISSUE #19941: TAsk: QA Callout-2
https://docs.google.com/spreadsheets/d/1kWosVCH622m9hohVKyFRdrr7TUa_exO8thwjo8Aqe2M/edit#gid=0

Tooluloope:
## Description

1. Issue with Pasting a copied List when the mouse pointer is in the second list item  https://www.loom.com/share/291b928a8b564cd6802e35968b5173be
2. Canvas crashes when we click on nested list Item in both Client side and Server Side pagination https://www.loom.com/share/64e36ec5d1b24f8cab45d68b218db222

https://docs.google.com/spreadsheets/d/1kWosVCH622m9hohVKyFRdrr7TUa_exO8thwjo8Aqe2M/edit#gid=0

Fixes #19941

## Type of change

## How Has This Been Tested?
- Jest
- Cypress

### Dev activity

### QA activity:


app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -736,6 +751,21 @@ class ListWidget extends BaseWidget<
     this.props.updateWidgetMetaProperty("selectedItemKey", null);
   };
 
+  /**
+   * Clearing selected Items and triggered items when the list widget is mounted
+   * because the MetaWidgetGenerator also clears all cached data when mounted or re-mounted
```
ashit-rath, line 756: Which part are you referring to where the cache is cleared in MetaWidgetGenerator?

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -736,6 +751,21 @@ class ListWidget extends BaseWidget<
     this.props.updateWidgetMetaProperty("selectedItemKey", null);
   };
 
+  /**
```
ashit-rath, line 754: Move this comment to the place this is called i.e in `componentDidMount`

app/client/src/widgets/ListWidgetV2/widget/index.tsx
Tooluloope: The `rowCacheData`. 
When we unmount the List Widget, the data in `rowCacheData` is lost and on-mount; since `SelectItemKey` is a meta property, the value is mounted with the List widget. 

So if the List widget is mounted on a page that doesn't have the `SelectItemKey`, there's no how we can grab the data since it is ServerSide Paginated.

This kind of case happens when you select a row or trigger a row and navigate away from the canvas to the query page.

ashit-rath: rowCacheData is not cleared when the Listwidget in umounted. We used to do that but I had to remove that logic because of nested list.
When the outer list widget changes page, the inner list in the prev page is unmounted and was removing all the cache data.
When we go back to the prev page, no cache for inner list and new ids were generated thus leading to meta props being lost.


Now with this logic when you switch pages, the inner list will mount/unmount and that would cause a reset.

ashit-rath (approved):

PR #20265, opened 2023-01-31
fix: primary key issues in list v2

### LINKED ISSUE #19941: TAsk: QA Callout-2

Tooluloope:
## Description

- Fix the Primary Key Error in Debugger
- Fix Blank Page when primary key doesn't exist

Fixes #19941 

## Type of change

## How Has This Been Tested?
- Jest
- Cypress

### Dev activity

### QA activity:


app/client/cypress/integration/Regression_TestSuite/ClientSideTests/Widgets/ListV2/Listv2_primaryKey_null_spec.js
```diff
@@ -0,0 +1,67 @@
+const dsl = require("../../../../../fixtures/Listv2/ListV2WithNullPrimaryKey.json");
```
ashit-rath, line 1: There is a primary key spec file. Can we move these test cases there?

app/client/cypress/integration/Regression_TestSuite/ClientSideTests/Widgets/ListV2/Listv2_primaryKey_null_spec.js
```diff
@@ -0,0 +1,67 @@
...
+const commonlocators = require("../../../../../locators/commonlocators.json");
+
+const widgetSelector = (name) => `[data-widgetname-cy="${name}"]`;
+
+describe(" Null Primary Key", function() {
+  before(() => {
+    cy.addDsl(dsl);
+  });
+  it("Widgets get displayed when PrimaryKey doesn't exist - SSP", () => {
```
ashit-rath, line 10: add a new line before this

app/client/cypress/integration/Regression_TestSuite/ClientSideTests/Widgets/ListV2/Listv2_primaryKey_null_spec.js
```diff
@@ -0,0 +1,67 @@
...
+    cy.wait(1000);
```
ashit-rath, line 11: Why wait(1000)?

app/client/cypress/integration/Regression_TestSuite/ClientSideTests/Widgets/ListV2/Listv2_primaryKey_null_spec.js
```diff
@@ -0,0 +1,67 @@
...
+    cy.get(commonlocators.listPaginateNextButton)
+      .first()
+      .click({
+        force: true,
+      });
+    cy.wait(1000);
+
+    cy.get(`${widgetSelector("Text2")} ${commonlocators.bodyTextStyle}`)
+      .first()
+      .should("have.text", `0`);
```
ashit-rath, line 35: use `""` instead of template literals

app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.ts
```diff
@@ -1175,16 +1175,14 @@ class MetaWidgetGenerator {
...
+     * as the PrimaryKey could be an array of null values or undefined. Hence, showing as equal whereas there was a change in
+     * data.
      */
 
     if (
       this.hasRegenerationOptionsChanged(nextOptions) ||
-      (this.primaryKeys === undefined &&
-        nextOptions.primaryKeys === undefined &&
-        this.serverSidePagination)
+      (!isEqual(nextOptions.data, this.data) && this.serverSidePagination)
```
ashit-rath, line 1185: Any specific reason to keep this condition here rather than moving inside the `hasRegenerationOptionsChanged`?

ashit-rath: Doing deep equal on the whole data can be very expensive. Since we just need to know if the current view's data has changed or not; so let's only get a piece of that dataset and compare that.

app/client/src/widgets/ListWidgetV2/widget/propertyConfig.ts
```diff
@@ -38,6 +38,23 @@ export const primaryColumnValidation = (
       };
     }
 
+    // when PrimaryKey is {{ currentItem["img"] }} and img doesn't exist in the data.
```
ashit-rath, line 41: Can you update this doc with these new error messages? 
https://www.notion.so/appsmith/List-v2-error-messages-acd96d5f5545452eaab7c625a5afc839
Add a new section with date as to when these new errors got added and tag parthacus on the same document to let him know about the updates

app/client/src/widgets/ListWidgetV2/widget/propertyConfig.ts
```diff
@@ -38,6 +38,23 @@ export const primaryColumnValidation = (
...
+          messages: ["Primary key cannot be null or undefined"],
+        };
+      }
+      return {
+        isValid: false,
+        parsed: inputValue, // undefined the chosen key doesn't exist.
+        messages: ["Chosen Primary key doesn't exist"],
+      };
+    }
+
```
ashit-rath, line 57: There are these conditions
1. If any one of the primarykey in null -> Primary key cannot be null or undefined
2. If all of them are null -> Chosen Primary key doesn't exist

Right now 2nd condition is fulfilled but not the first and let's not have the  `inputValue.length === 1` as it's very specific.

Tooluloope: So for `[null]` when the primaryKey exists, what would be the error message?

ashit-rath: Primary key cannot be null or undefined

Tooluloope: The difficulty here is getting the difference between primaryKey doesn't exist for a 1 page-sized List widget and primary key does exist, but the value is also null.

ashit-rath: Okay so here's how I see it. 
Priority 1 - primary key doesn't exist 
Priority 2 - cannot be null or undefined.

Tooluloope: Should it be `Primary key cannot be null or undefined` or `Primary key cannot contain null or undefined`

app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.ts
```diff
@@ -1339,7 +1326,9 @@ class MetaWidgetGenerator {
       nextOptions.pageSize !== this.pageSize ||
       nextOptions.primaryKeys !== this.primaryKeys ||
       nextOptions.serverSidePagination !== this.serverSidePagination ||
-      nextOptions.templateBottomRow !== this.templateBottomRow
+      nextOptions.templateBottomRow !== this.templateBottomRow ||
+      // Comparing a slice of the data as deep equal on the whole data can be very expensive.
+      (!isEqual(nextOptions.data[0], this.data[0]) && this.serverSidePagination)
```
ashit-rath, line 1331: What will happen if the the data length stays the same and only 2nd or 3rd item data gets modified. 
If feel this is not the right way to check if something changed.
Ideally primaryKey is our source of truth for any sort of change, now imagine if the first data (data[0]) changed but the id is still the same and used as primary key. We will try to regenerate this, where we shouldn't as the bindings that we have for currentItem should automatically pick up the updates which is one of the plus points of v2.

Now how I think we should probably do this.
As primary key is the source of truth here, if a PK is [1, null, 2] or [null, null, null]. At the very beginning we generate the primary keys for all the null values and store it in a class variable. We will do this only for the number of items that can be visible in the page. Once this is generated we deepequal this.

If the PK is correctly set by the user, then there is no need to generate PK from them out of hashes and deep equal can be done.

kamakshibhat-appsmith:
https://www.loom.com/share/a6b02e86d5b14a8782b002c865a2a2de @ashit-rath The meta part is still an issue in this PR, pls check Loom  cc: @Tooluloope

PR #20295, opened 2023-02-01
chore: Add version to analytics events

sharat87:
Currently, analytics events don't include Appsmith version information. This PR adds version, and edition information to all events.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/ProjectProperties.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/ProjectProperties.java
@@ -16,6 +16,8 @@ import org.springframework.context.annotation.PropertySource;
 @Getter
 public class ProjectProperties {
 
+    public static final String EDITION = "CE";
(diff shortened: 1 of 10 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/ProjectProperties.java
```diff
@@ -16,6 +16,8 @@
 @Getter
 public class ProjectProperties {
 
+    public static final String EDITION = "CE";
```
nayan-rafiq, line 19: Can we use Enum here instead of String constants?

sharat87: Sure, but it felt like over-engineering it. This doesn't go in the DB, doesn't ever change, I don't want to include other possible values for this here in the CE repo, so, a plain constant like how we have constants for URLs felt simpler and better. Not a very strong opinion though. What say?

PR #20301, opened 2023-02-02
chore: Fix Installation complete event not being sent

sharat87:
The problem was that the Mono returned by `analyticsService.sendEvent` was returned inside a `.map` call, which, (almost) always results in a bug.

This PR fixes that.


Nikhil-Nandagopal (approved):

Nikhil-Nandagopal:
@sharat87 let's merge this?

Issue #19692, opened 2023-01-11
[Bug]: Unable to connect to Amazon Congito IdP


pranavkanade:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Cognito IdP integration is failing for a user. The issue seems to be related to scopes.

### Steps To Reproduce

Follow the steps in the SSO doc for Amazon Cognito.

### Version

v1.9.2

infinitetrooper:
I'm reopening this as there's no QA validation for the fixes we've done. @RakshaKShetty  will update the status once QA is verified.

The user who reported this moved to AD but, we don't know yet if this is working for Cognito.

Slack Thread here - https://theappsmith.slack.com/archives/CPQNLFHTN/p1673362611104259

infinitetrooper:
The issue was reported by webmethod.com - https://theappsmith.slack.com/archives/C02K2MZERSL/p1675415041117339

PR #20433, opened 2023-02-07
fix: remove padding from state used in OIDC authorise

nsarupr:
## Description

> Cognito returns **Malformed URI** for redirect when User tries to Login. Due to this, browser ends up throwing `400 Bad Request`. In order to fix this, we have removed the padding from the `state` query parameter which we send to the OIDC, in order to avoid the cases where the additional `=` will lead to malformed URIs being created. Read [here](https://stackoverflow.com/questions/6916805/why-does-a-base64-encoded-string-have-an-sign-at-the-end) for Additional information on the Base64 encoding and padding.
> Also, we are changing the delimiter from `=` to `-` when server tries to find the redirect URI for other use cases.
> Server uses `,` in order to split the state to get the `origin value`. Now we will use `@` instead of `,`.

> TL;DR, remove `=` and `,` in order to avoid malformed URI strings.

Fixes https://github.com/appsmithorg/appsmith/issues/19692

Media

## Type of change

### Dev activity

### QA activity:


trishaanand:
Tests on EE with the same changes have passed (except an unrelated failure for upgrade) : https://github.com/appsmithorg/appsmith-ee/actions/runs/4113772980

PR #20463, opened 2023-02-08
fix: Check for AVX instruction, and produce a warning

sharat87:
When running on CPUs without an AVX instruction, the embedded MongoDB cannot run. It fails to run with a cryptic error message, and they are forced to reach out to support for help.

This PR adds a check, and a warning message, if trying to use embedded MongoDB on a CPU without the AVX instruction.

### DIFF
--- a/deploy/docker/entrypoint.sh
+++ b/deploy/docker/entrypoint.sh
@@ -5,17 +5,6 @@ set -o xtrace
 
 stacks_path=/appsmith-stacks
 
-system_check() {
-  if [[ -f /proc/cpuinfo ]] && ! grep --quiet avx /proc/cpuinfo; then
-    echo "===================================================================================================="
-    echo "=="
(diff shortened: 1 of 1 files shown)


deploy/docker/entrypoint.sh
```diff
@@ -173,6 +173,15 @@ init_replica_set() {
   fi
 
   if [[ $isUriLocal -gt 0 ]]; then
+    if [[ -f /proc/cpuinfo ]] && ! grep --quiet avx /proc/cpuinfo; then
```
mohanarpit, line 176: Do we also want to exit if we don't find AVX instructions on the target CPU?

sharat87: I considered that, but didn't want to be too drastic right away. If they face problems and share logs with our support team, with this error message, at least we know what's up. I want to see this in the wild for a week or two, and then make it exit with error message. Does that sound okay?

mohanarpit (approved):

Issue #20500, opened 2023-02-08
[Task - App Share]: Migrate defaultWorkspaceId in Permission Group model to defaultDomainId & defaultDomainReference


nsarupr:
### Is there an existing issue for this?

- [X] I have searched the existing issues

PR #20501, opened 2023-02-08
feat: migrate default workspace to default domain

nsarupr:
## Description

> Migrate the Default Workspace Id field in Permission Group to a combination of Default Domain Id and Default Domain Reference field.

Fixes https://github.com/appsmithorg/appsmith/issues/20500

Media

## Type of change

### Dev activity

### QA activity:


app/server/appsmith-server/src/main/java/com/appsmith/server/domains/PermissionGroup.java
```diff
@@ -25,8 +25,12 @@ public class PermissionGroup extends BaseDomain {
 
     //TODO: refactor this to defaultDocumentId, as we can use this to store associated document id for 
     //which we are auto creating this permission group.
+    @Deprecated
     String defaultWorkspaceId;
 
+    String defaultDomainId;
+    DomainReference defaultDomainReference;
```
trishaanand, line 32: Should we simply use the Class name string here instead of enum? Any reason for introducing this construct?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
```diff
@@ -1528,7 +1529,8 @@ private Set<PermissionGroup> generateDefaultPermissionGroupsWithoutPermissions(M
         // Administrator permission group
         PermissionGroup adminPermissionGroup = new PermissionGroup();
         adminPermissionGroup.setName(getDefaultNameForGroupInWorkspace(FieldName.ADMINISTRATOR, workspaceName));
-        adminPermissionGroup.setDefaultWorkspaceId(workspaceId);
+        adminPermissionGroup.setDefaultDomainId(workspaceId);
+        adminPermissionGroup.setDefaultDomainReference(WORKSPACE);
```
trishaanand, line 1533: Since these migrations have already run for old instances, we don't need to update this. Since the only instances who would be running these migrations are new instances or old ones who havent updated yet, for the former, it doesnt matter and for the latter, the newly written migrations would take care of introducing this.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/PermissionGroupDefaultWorkspaceIdMigration.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.data.mongodb.core.query.UpdateDefinition;
+
+import static com.appsmith.server.domains.DomainReference.WORKSPACE;
+import static com.appsmith.server.repositories.BaseAppsmithRepositoryImpl.fieldName;
+import static org.springframework.data.mongodb.core.query.Criteria.where;
+import static org.springframework.data.mongodb.core.query.Query.query;
+
+
+@ChangeUnit(order = "100", id="migrate-default-workspace-id-to-default-domain-id")
+public class PermissionGroupDefaultWorkspaceIdMigration {
```
trishaanand, line 21: I think the naming convention needs to be along the lines as suggested here : https://github.com/appsmithorg/appsmith-ee/blob/release/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/Migration101EeEnvVariableIndexes.java

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/PermissionGroupDefaultWorkspaceIdMigration.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.data.mongodb.core.query.Query;
+import org.springframework.data.mongodb.core.query.UpdateDefinition;
+
+import static com.appsmith.server.domains.DomainReference.WORKSPACE;
+import static com.appsmith.server.repositories.BaseAppsmithRepositoryImpl.fieldName;
+import static org.springframework.data.mongodb.core.query.Criteria.where;
+import static org.springframework.data.mongodb.core.query.Query.query;
+
+
+@ChangeUnit(order = "100", id="migrate-default-workspace-id-to-default-domain-id")
```
trishaanand, line 20: Could you check if the order needs to be "003" instead?

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/PermissionGroup.java
nsarupr: Making this construct, so that when a Model updates(Page / NewPage) / has diversified meaning but similar permission use cases(NewAction / ActionCollection), which are extended from BaseDomain, 1 single Identity is sent, so that the developers don't end up writing logics in multiple places, or use if else conditions.

The use case right now is limited, but this seems extensible for future as well.

trishaanand (approved):

trishaanand:
@nsarupr : Could we test the migrations with release db dump and ensuring data sanity? Thanks!

PR #20506, opened 2023-02-08
fix: Limit List widget nesting to a maximum of 3 levels

### LINKED ISSUE #20585: [Feat]: Limit Nesting in List Widget to 3 Levels

Tooluloope:
## Description

Limit the amount of nesting in the List widget to 3 levels.

The List widget could be nested in 2 ways.
1. Drag and Drop from Explorer
2. Copy and Paste.
3. Drag and Drop from Canvas

To cater to all these cases, I created 3 more blueprint operations that serve as a life cycle for the widget.
1. BEFORE_ADD
2. BEFORE_PASTE
3. BEFORE_DROP

This would enable validation of the level/nesting before the DSL is updated

Fixes #20585

## Type of change

## How Has This Been Tested?
- Cypress

### Dev activity

### QA activity:


app/client/src/sagas/WidgetOperationUtils.ts
```diff
@@ -1796,3 +1798,41 @@ const updateListWidgetBindings = (
...
+
+        return false;
+      }
+
+      return true;
+    });
+  }
+
+  return copiedWidgetGroups;
+};
```
ashit-rath, line 1838: I can see this similar logic is being repeated in 3 different places. Let's try to make this whole thing generic
1. Create a method in widget index config file blueprint operation and move this logic there
2. I think we can do this calculation of whether a level of nesting is more than allowed without looking at meta widget. You can have a recursive function that walks up the tree and counts how many list widget underneath it is as they are all template widgets. 
3. When creating this method, return error message that we want to show and call the Toast message inside the saga rather than in the method.

ashit-rath:
@Tooluloope Can you create an issue for it and tag it here?

Tooluloope:
> @Tooluloope Can you create an issue for it and tag it here?

Done @ashit-rath

app/client/src/sagas/CanvasSagas/DraggingCanvasSagas.ts
```diff
@@ -308,6 +312,17 @@ function* moveWidgetsSaga(
   try {
     const allWidgets: CanvasWidgetsReduxState = yield select(getWidgets);
 
+    for (const dragBlock of draggedBlocksToUpdate) {
+      yield call(
+        executeWidgetBlueprintBeforeOperations,
+        BlueprintOperationTypes.BEFORE_DROP,
```
ashit-rath, line 318: Can we pass the type as one argument and the rest as options?

app/client/src/sagas/WidgetBlueprintSagas.ts
```diff
@@ -238,3 +245,25 @@ export function* traverseTreeAndExecuteBlueprintChildOperations(
 
   return widgets;
 }
+
+export function* executeWidgetBlueprintBeforeOperations(
+  blueprintOperation: BlueprintOperationTypes,
```
ashit-rath, line 250: So here we should only be accepting the `before` operations types. Can we not use `BlueprintOperationTypes` but a subset of it so that if someone tries to use it for other operation, they won't be able to?

app/client/src/sagas/WidgetOperationSagas.tsx
```diff
@@ -1297,378 +1301,409 @@ function* pasteWidgetSaga(
   let widgets: CanvasWidgetsReduxState = canvasWidgets;
   const selectedWidget: FlattenedWidgetProps<undefined> = yield getSelectedWidgetWhenPasting();
 
-  let reflowedMovementMap,
-    gridProps: GridProps | undefined,
-    newPastingPositionMap: SpaceMap | undefined,
-    canvasId;
+  try {
```
ashit-rath, line 1304: Basically you have moved this whole block into a `try catch` and added the before blueprint operation?

app/client/src/widgets/ListWidgetV2/widget/helper.ts
```diff
@@ -25,3 +27,53 @@ export const extractTillNestedListWidget = (
 
   return extractedWidgets;
 };
+
+export const getNumberOfParentListWidget = (
```
ashit-rath, line 31: Can we add jest test for these functions?

app/client/src/sagas/WidgetOperationSagas.tsx
Tooluloope: Exactly.

ashit-rath (dismissed):

chandannkumar:
Observations: @Tooluloope 
1. On adding more than 3 Lists, 2 error message appears at same time, Can we only keep the appropriate one.


2. On cyclic dependency, the widget acts weird
https://bthrujCsW8.vmaker.com/record/dY2bEkINjqdXfRJq

Tooluloope:
@chandannkumar For the Cyclic Dependency, can you confirm if the parent of Text3 is List3?  If it's List3 then there would be a cyclic dependency as you're just meant to use `currentView`

chandannkumar:
Text3 and List3 are in Level 2. That is why i get the recommendation to use List3 in my currentView.
Since inside List3 widget i am trying to add List3, hence cyclic dependency

[Sample app](https://appsmith-git-fix-maximum-level-of-nesting-i-fbe564-get-appsmith.vercel.app/app/list-v2-phase-2/nested-listv2-63ef19dff62df12bd446e67d/edit)

chandannkumar:
@Tooluloope  Dp link is broken. Can you please check this
image

Tooluloope:
@chandannkumar The deploy link is ready

chandannkumar:
Tested this and LGTM:
- Tested nested list and their maximum level
- Tested in Modal, container type
- Tested with theming scenarios
- Tested copy paste scenarios

ashit-rath (approved):

PR #20524, opened 2023-02-09
Refactor/issue 19280/Application DTO Refactor

### LINKED ISSUE #19280: [Task]: Aggregate all published and unpublished Fields in a single type in Application.java

### SubTasks

- The `application` model has multiple fields which store the `published` and `unpublished` states such as-
    - `pages` and `publishedPages`
    - `publishedCustomJSLibs` and `unpublishedCustomJSLibs`
    - `publishedModeThemeId` and `editModeThemeId`
    - `publishedAppLayout` and `unpublishedAppLayout`
    - `unpublishedAppPositioning` and `publishedAppPositioning`
    - `unpublishedNavigationSetting` and `publishedNavigationSetting`

~~All the above fields need to be grouped together and stored as a single~~ `ApplicationDetail`

Update: As discussed, the newly created set of fields `AppPositioning` and `NavigationSetting` need to be grouped in the `ApplicationDetail` and the older fields remain in the `Application` as of now.

NilanshBansal:
## Description

> This PR refactors the backend code and the `Application` model to accumulate all published and unpublished fields into a single `ApplicationDTO`

Fixes #19280

## Type of change

## How Has This Been Tested?

- Manual
- JUnit

### Dev activity

### QA activity:


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog1.java
```diff
@@ -1962,20 +1962,20 @@ public void changeAppLayoutTypeDefinition(MongoOperations mongoOperations) {
         //     Tablet: Old - 1024, New 650 - 800
         //     Mobile: Old - 720, New 350 - 450
         final Criteria criteria = new Criteria().orOperator(
-                where(fieldName(QApplication.application.unpublishedAppLayout)).exists(true),
```
nidhi-nair, line 1965: We will need to change all these migrations to be by string path. For all instances that this migration will affect, the data would have been in the old path itself.

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Application.java
```diff
@@ -182,18 +155,22 @@ public String getLastDeployedAt() {
     public Application(Application application) {
         super();
         this.workspaceId = application.getWorkspaceId();
-        this.pages = new ArrayList<>();
-        this.publishedPages = new ArrayList<>();
+        this.getUnpublishedApplication().setPages(new ArrayList<>());
```
nidhi-nair, line 158: Instead of initializing individual properties here, what do you think of using the constructor for `ApplicationDTO` or a similar construct to do this for all fields?

nidhi-nair: Also asking because all these references will now need to be checked for null references as well.

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Application.java
```diff
@@ -216,32 +193,42 @@ public void sanitiseToExportDBObject() {
...
-            publishedAppLayout = appLayout;
+            getPublishedApplication().setAppLayout(appLayout);
         } else {
-            unpublishedAppLayout = appLayout;
+            getUnpublishedApplication().setAppLayout(appLayout);
         }
     }
 
+    public ApplicationDTO getUnpublishedApplication() {
+        return unpublishedApplication == null ? new ApplicationDTO() : unpublishedApplication;
```
nidhi-nair, line 221: Nit: I'm usually averse to abstracting the null logic like this because it takes away the responsibility of knowing when an applicationDTO is expected to not be null. In some such cases I would expect to error out, and in others I'd want to create a new application object entirely. Is there a third? If not, what kind of value does this abstraction add? It does not even remove LOC, because in either of the cases I have mentioned, we would need to do the null check anyway to return with the appropriate handling.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
```diff
@@ -2535,8 +2536,8 @@ public void migratePublicAppsSinglePg(MongoTemplate mongoTemplate, @NonLockGuard
                     // Update application themes
                     Criteria nonSystemThemeCriteria = Criteria.where(fieldName(QTheme.theme.isSystemTheme)).is(false);
                     Criteria idCriteria = Criteria.where(fieldName(QTheme.theme.id)).in(
-                            application.getEditModeThemeId(),
-                            application.getPublishedModeThemeId()
+                            application.getUnpublishedApplication().getThemeId(),
+                            application.getPublishedApplication().getThemeId()
                     );
                     Criteria queryCriteria = new Criteria().andOperator(nonSystemThemeCriteria, idCriteria);
                     Set<Policy> themePolicies = policyGenerator.getAllChildPolicies(application.getPolicies(), Application.class, Theme.class);
```
nidhi-nair, line 2543: Looks like WIP, will wait for DB and JSON migration for another round of review.

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/ApplicationDTO.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @Transient
+    private String id;
+
+    @JsonIgnore
+    Application.AppLayout appLayout;
+
+    Set<CustomJSLibApplicationDTO> customJSLibs;
+
+    @JsonIgnore
```
nidhi-nair, line 29: We shouldn't need these json ignores anymore

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Application.java
```diff
@@ -111,19 +98,11 @@ public class Application extends BaseDomain {
...
-    @JsonIgnore
-    NavigationSetting unpublishedNavigationSetting;
-
-    @JsonIgnore
-    NavigationSetting publishedNavigationSetting;
+    Boolean collapseInvisibleWidgets;
 
-    @JsonIgnore
-    AppPositioning publishedAppPositioning;
+    ApplicationDTO unpublishedApplication;
```
mohanarpit, line 103: @NilanshBansal For my understanding, isn't this a leaky abstraction? Why is the `ApplicationDTO` data structure being used in the persistence layer? 

I'm asking this because a DTO (Data Transfer Object) is typically used to communicate with external clients. There are obviously trade-offs we make when sticking to such a definition of the DTO. But if we are changing the definition of what this phrase means, it needs to be consistently applied to the rest of the code base as well.

NilanshBansal: @mohanarpit this PR is a WIP. We plan to replace all the return types of APIs from `Application` to `ApplicationDTO` subject to its requirement.

mohanarpit: I understand that this is a draft PR. My question is if persisting the DTO structure in MongoDB is a leaky abstraction or not?

NilanshBansal: @mohanarpit I get your point, there is a similar implementation in the codebase already done for `NewPage.java` with the `PageDTO` which I was taking as a reference point for this implementation.

NilanshBansal: I can update the name for the current implementation from `ApplicationDTO`, need your review on this @mohanarpit

NilanshBansal:
Closing PR, ApplicationDTO, API changes are not required as discussed. Another PR will be generated to solve the above issue

PR #20612, opened 2023-02-14
chore: remove feature flags

berzerkeer:
### Description
 - Removes feature flags for usage & billing.
 - moves pricing url to constant.

### DIFF
--- a/app/client/src/utils/hooks/useOnUpgrade.ts
+++ b/app/client/src/utils/hooks/useOnUpgrade.ts
@@ -1,5 +1,6 @@
-import { getTenantConfig } from "@appsmith/selectors/tenantSelectors";
 import { useSelector } from "react-redux";
+import { getInstanceId } from "@appsmith/selectors/tenantSelectors";
+import { PRICING_PAGE_URL } from "constants/ThirdPartyConstants";
 import AnalyticsUtil, { EventName } from "utils/AnalyticsUtil";
 
 type Props = {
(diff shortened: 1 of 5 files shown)


app/client/src/utils/hooks/useOnUpgrade.ts
```diff
@@ -1,45 +1,29 @@
...
   intercomMessage?: string;
   logEventName?: EventName;
   logEventData?: any;
 };
 
-const useOnUpgrade = (props: Props) => {
-  const { intercomMessage, logEventData, logEventName } = props;
-  const features = useSelector(selectFeatureFlags);
-  const instanceId = useSelector(getInstanceId);
+const PRICING_PAGE_URL =
```
sbalaji1192, line 11: move this to the file where we have ASSET_CDN_URL

app/client/src/utils/hooks/useOnUpgrade.ts
```diff
@@ -1,45 +1,29 @@
...
-    if (features.USAGE_AND_BILLING) {
-      window.open(
-        `https://www.appsmith.com/api/preview?secret=8JPsJRnSkt6Va8FzxUPFhZezxZuHRnSU&slug=pricing-preview?source=CE&instance=${instanceId}`,
-        "_blank",
-      );
-    } else {
-      triggerIntercom(intercomMessage || createMessage(UPGRADE_TO_EE_GENERIC));
-    }
+    window.open(
+      `${PRICING_PAGE_URL}?source=CE&instance=${tenantConfig?.instanceId}`,
```
sbalaji1192, line 24: move the entire thing to the constant

berzerkeer: The params too ? The params take value from the hook, only the base URL is going to be the same. So shouldn't we just move the base the URL to constant and append the params in the file from where we are actually redirecting to the URL.

sbalaji1192: create a function that accepts this value and returns the constant file.

berzerkeer: Yeah thats fine, but do we know whether this pricing URL will always take `source` and `instance` as search params?
Even though we can conditionally append or not, it just adds more checks . Thats why I was wondering.

sbalaji1192: Let add those two keys always for now. Will think about later how to do that conditionally

berzerkeer: Sure 👍

sbalaji1192 (approved):

PR #20660, opened 2023-02-16
chore: Remove OAuth env variables from client

sharat87:
Client will get supported OAuth list from the tenant API, instead of from injected env variables like `APPSMITH_OAUTH2_GOOGLE_CLIENT_ID`.

This is a step towards moving OAuth configuration out of env variables completely, and into the backend database, so their configuration can be tenant-wide, instead of instance-wide.

NOTE: Please review/approve only. Please do not merge. It needs an accompanying EE PR to go togather. Thanks!


app/client/src/ce/configs/index.ts
```diff
@@ -26,7 +24,6 @@ export interface INJECTED_CONFIGS {
     licenseKey: string;
   };
   enableMixpanel: boolean;
-  google: string;
```
sharat87, line 29: This is maps API key, got left out, is unused.

app/client/src/ce/pages/UserAuth/Login.tsx
```diff
@@ -85,7 +86,10 @@ export function Login(props: LoginFormProps) {
   const { emailValue: email, error, valid } = props;
   const isFormValid = valid && email && !isEmptyString(email);
   const location = useLocation();
-  const socialLoginList = ThirdPartyLoginRegistry.get();
+  const socialLoginList = [
+    ...useSelector(getThirdPartyAuths),
+    ...ThirdPartyLoginRegistry.get(),
+  ];
```
sharat87, line 92: Once auth on EE is also moved, we'll only use the selector here.

ankitakinger (approved):
LGTM

ankitakinger (approved):
LGTM

PR #20771, opened 2023-02-20
chore: Move Maps API Key to database

sharat87:
1. Changing the Maps API Key doesn't need restart anymore.
2. The `isRestartRequired` field in the response of updating env settings, was being ignored. The client owns the decision of when to restart (which is correct), so removed this from the server.
3. Write Maps API Key to the database, in the tenant configuration.
4. The Settings page for Maps Ke gets the current value from `/tenant/current` response, and not `/admin/env`.
5. Removed `APPSMITH_GOOGLE_MAPS_API_KEY` from `/admin/env` response.
6. Tests.

DO NOT MERGE. Please only review/approve. This is expected to break EE once it goes there, which I intend to solve alongside merging this.

Changing the Maps API Key will update it both in the tenant config in the database, as well as in the `docker.env` file. This is predominantly for backwards compatibility, and phased rollout. As part of a separate PR, we'll have a migration that proactively copies the env variable value to the database, and comment out the value in the `docker.env` file. Then we can stop updating the `docker.env` file as well.

## New
Screenshot 2023-02-25 at 7 30 14 AM

## Old
Screenshot 2023-02-25 at 7 23 05 AM


app/client/src/pages/Settings/config/googleMaps.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const [mapsKey, setMapsKey] = React.useState<string>(googleMapsApiKey);
+  const dispatch = useDispatch();
+
+  const isModified = googleMapsApiKey === mapsKey;
+
+  function onSave() {
+    dispatch(
+      saveSettings(
+        {
+          APPSMITH_GOOGLE_MAPS_API_KEY: mapsKey,
```
ankitakinger, line 40: isn't this going to send the updated key to `/admin/env` API itself? Should we use `/updateDefaultTenantConfiguration` instead? We are using this for `APPSMITH_ENABLE_SINGLE_SESSION_PER_USER `and `APPSMITH_SHOW_ROLES_AND_GROUPS` `assingleSessionPerUserEnabled` and `showRolesAndGroups` resp. from tenant config.

tanvibhakta (dismissed):

app/client/src/pages/Settings/config/googleMaps.tsx
sharat87: Interesting. Looks like it was added very recently, and, unfortunately, isn't in a position to use here. Alright, here's what I have.

1. The `/updateDefaultTenantConfiguration`'s backend implementation is missing in this repo. It's only in EE. Frontend implements the reducer for calling this API, but doesn't use it.
2. Even if we do that, in addition to doing that, we want to write to the env variable for a while longer, for backwards compatibility.

Considering this API doesn't exist here, I'm thinking we'll take this forward, then bring that API to CE here, and then leverage that API for Maps configuration as well. Is that okay?

app/client/src/pages/Settings/config/googleMaps.tsx
```diff
@@ -0,0 +1,108 @@
...
+                to: GOOGLE_MAPS_SETUP_DOC,
+              },
+            ]}
+          >
+            {createMessage(() => "How to configure?")}
+          </Callout>
+          <Input
+            id="APPSMITH_GOOGLE_MAPS_API_KEY"
+            isValid={mapsKey.trim() !== ""}
+            label="Google Maps API Key"
```
ankitakinger, line 80: 1. can we update the `k` for `key` to lowercase?
2. Do we really need the isValid here? I'm just asking because this wasn't used in the old implementation. And this is making the textbox have an orange border instead of grey/black. While the other textboxes on admin settings still use grey/black.
image
image

sharat87: Hey, sorry just seeing this.

1. Sure.
2. That validation actually is there. If you type something in to set an API key, it won't let you clear it. Empty value shows up as an error. So, once you set a Maps key, you can't remove it. That's why I'm actually considering removing it to fix that problem.

sharat87:
All seems to be working well now, except for one thing. Now that the Maps API key can be updated without a restart, it means that the new key should take affect without a page refresh. But that's not happening. The Redux state does get updated with the new value, but the widget doesn't seem to pick it up. The widget seems to be getting it from a _dependency_, instead of directly invoking the selector in a functional component... that's where I'm stuck.

Here's a video of what I see:

https://github.com/appsmithorg/appsmith/assets/120119/fe645982-5449-4d57-8489-df756aae5766

Can I get some help with this please @ankitakinger?

ankitakinger:
@sharat87 I can help you look into this on Wednesday once I'm back from my leave. Meanwhile, I checked your current changes and the entire `googleMaps` file need not be changed. Please refer to: https://theappsmith.slack.com/archives/CPQNLFHTN/p1687161636989599?thread_ts=1686629890.316679&cid=CPQNLFHTN

sharat87:
Thanks @ankitakinger, that sounds good. I tried making the change you suggested there, and it seems to be working well. I had to add it to `tenantConstants.tenantConfigConnection` list as well.

But, the save button shows us as "Save & Restart", until there's an actual change, when it changes to "Save". Not a big thing, but it does look odd.


https://github.com/appsmithorg/appsmith/assets/120119/3706be98-2eeb-41d4-9db6-b85007e573b2

sharat87:
Reg., https://github.com/appsmithorg/appsmith/pull/20771#issuecomment-1615857856, turns out that if Maps is loaded with one key, it can't be loaded with a different key on the same page, without a page refresh. So, going from no-key to valid-key, actually works fine. The problem is only limited to going from one-key to another-key, and that, is unblocked by a quick page refresh.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/TenantServiceCEImpl.java
```diff
@@ -55,15 +55,19 @@ public Mono<String> getDefaultTenantId() {
...
-                oldtenantConfiguration = new TenantConfiguration();
-            }
-            AppsmithBeanUtils.copyNestedNonNullProperties(tenantConfiguration, oldtenantConfiguration);
-            tenant.setTenantConfiguration(oldtenantConfiguration);
-            return repository.updateById(tenantId, tenant, MANAGE_TENANT);
-        });
+        return repository
+                .findById(tenantId, MANAGE_TENANT)
+                .switchIfEmpty(Mono.error(
+                        new AppsmithException(AppsmithError.ACL_NO_RESOURCE_FOUND, FieldName.TENANT, tenantId)))
```
sharat87, line 61: The diff is just adding these two lines, rest is formatting diff.

sharat87:
Cypress tests have actually all passed. The failure is due to this:

shot-2023-07-18-08-05-46

Here's Cypress dashboard:

shot-2023-07-18-08-07-12

ankitakinger (dismissed):
FE changes looks good to me

sharat87:
Changing the Maps key crashes frontend. 😭


https://github.com/appsmithorg/appsmith/assets/120119/94355368-8d39-48ea-ae4c-7f93fa1b3dd4

[Slack thread here](https://theappsmith.slack.com/archives/C02JV8G1MP0/p1689740931583169).

ankitakinger (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/ce/TenantConfigurationCE.java
```diff
@@ -34,11 +37,15 @@ public void addThirdPartyAuth(String auth) {
...
-    }
 
-    public License license;
+        if (tenantConfiguration == null) {
+            return;
+        }
+
+        googleMapsKey = ObjectUtils.defaultIfNull(tenantConfiguration.getGoogleMapsKey(), googleMapsKey);
+        isFormLoginEnabled = ObjectUtils.defaultIfNull(tenantConfiguration.getIsFormLoginEnabled(), isFormLoginEnabled);
+        instanceName = ObjectUtils.defaultIfNull(tenantConfiguration.getInstanceName(), instanceName);
```
nidhi-nair, line 49: LGTM. QQ about these values, what is the utility of these lines? Why are we retaining older values?

nidhi-nair (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/ce/TenantConfigurationCE.java
sharat87: This function is used to copy values from the object from DB, into the object we're about to send to the client. So we only copy those, that should be available to the client.

Now, for backwards compatibility, the client object is pre-filled with values from env variables. So for those that don't have a value in DB, which will end up being `null`, will retain their fallback value that was taken from env variables. That's why protect against the value from DB being `null`, and if so, we just keep what we have in the object.

Does that make sens? 🙂

nidhi-nair: Gotcha, thanks!

PR #20845, opened 2023-02-21
feat: Total Record Count field for Server-side pagination in List widget

### LINKED ISSUE #14551: [Feature]: Need Total Record Count field for Server-side pagination in List widget

### Summary

When a user decides to use server side pagination in the List widget, it is unaware of how many records will be present.

This causes issue where a user can keep clicking next page, and see List items which have undefined values.
We need a field like "Total Record Count" to be shown, whenever server side pagination is enabled, done similarly in Table Widget.

Tooluloope:
## Description

Created a new property, Total Record Count. This helps the user navigate the list items.
It prevents landing on an empty page since the app developer sets a Total Record Count.

Fixes #14551 

## Type of change

## How Has This Been Tested?
- Manual
- Cypress

### Dev activity
- [ ] I have commented on my code, particularly in hard-to-understand areas

### QA activity:

### DIFF
--- a/app/client/src/widgets/ListWidgetV2/widget/index.tsx
+++ b/app/client/src/widgets/ListWidgetV2/widget/index.tsx
@@ -786,6 +786,18 @@ class ListWidget extends BaseWidget<
     this.resetTriggeredItemView();
   };
 
+  getPageCount = () => {
+    const defaultValue = 0;
+    const { serverSidePagination, totalRecordsCount } = this.props;
+
+    if (!serverSidePagination) return (this.props.listData || []).length;
+
+    if (typeof totalRecordsCount === "number" && totalRecordsCount > 0)
+      return totalRecordsCount;
+
+    return defaultValue;
+  };
+
   shouldPaginate = () => {
     /**
      * if client side pagination and not infinite scroll and data is more than page size
@@ -943,7 +955,7 @@ class ListWidget extends BaseWidget<
     const disableNextPage = this.shouldDisableNextPage();
     return (
       this.shouldPaginate() &&
-      (serverSidePagination ? (
+      (serverSidePagination && !this.getPageCount() ? (
         <ServerSideListPagination
           accentColor={this.props.accentColor}
           borderRadius={this.props.borderRadius}
@@ -965,7 +977,7 @@ class ListWidget extends BaseWidget<
           onChange={this.onPageChange}
           pageNo={this.props.pageNo}
           pageSize={this.pageSize}
-          total={(this.props.listData || []).length}
+          total={this.getPageCount()}
         />
       ))
     );
@@ -1089,6 +1101,7 @@ export interface ListWidgetProps<T extends WidgetProps = WidgetProps>
   primaryKeys?: (string | number)[];
   serverSidePagination?: boolean;
   nestedViewIndex?: number;
+  totalRecordsCount?: number | string;
 }
 
 export default ListWidget;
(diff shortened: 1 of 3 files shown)


app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -965,7 +977,7 @@ class ListWidget extends BaseWidget<
           onChange={this.onPageChange}
           pageNo={this.props.pageNo}
           pageSize={this.pageSize}
-          total={(this.props.listData || []).length}
+          total={this.getPageCount()}
```
vsvamsi1, line 980: nothing major..... the page count could be evaluated earlier and the result could be referenced....

```
const pageCount = this.getPageCount()
...
serverSidePagination && !pageCount ? (
...
total={pageCount}
```

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -786,6 +786,18 @@ class ListWidget extends BaseWidget<
...
 
+  getPageCount = () => {
+    const defaultValue = 0;
+    const { serverSidePagination, totalRecordsCount } = this.props;
+
+    if (!serverSidePagination) return (this.props.listData || []).length;
+
+    if (typeof totalRecordsCount === "number" && totalRecordsCount > 0)
+      return totalRecordsCount;
+
```
vsvamsi1, line 797: Since the type of totalRecordsCount is a number | string. Should we attempt to cast a string to a number to support a possibility like "12", "1".
If we do maybe we can change the if block to something like this 
```
if (+totalRecordsCount > 0)
      return +totalRecordsCount;
```

ashit-rath (commented):
@Tooluloope The checklist in the PR description is not marked. Please do update that.
The documentation also needs to be updated and we might need @theparthacus 's help with the copy of the tooltip text so add this entry in the notion doc.

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -943,7 +955,7 @@ class ListWidget extends BaseWidget<
     const disableNextPage = this.shouldDisableNextPage();
     return (
       this.shouldPaginate() &&
-      (serverSidePagination ? (
+      (serverSidePagination && !this.getPageCount() ? (
```
ashit-rath, line 958: Can we add a cypress test to check this?

app/client/src/widgets/ListWidgetV2/widget/index.tsx
Tooluloope: For this, the value we get when the user enters a `string number` is a number, as seen below.

Screenshot 2023-02-22 at 09 36 46

It could be a  string when the user enters something that doesn't have a number representation, as shown below.

Screenshot 2023-02-22 at 09 38 35

vsvamsi1: Makes sense

ashit-rath: This is quite strange, if the uneval value is `{{}}` then ideally the evaluated value should be null/undefined or empty string and the validation should convert empty strings to null/undefined.
@aswathkk Is this a bug?

ashit-rath (approved):

app/client/src/widgets/ListWidgetV2/widget/index.tsx
aswathkk: So far I thought validation is done after doing an evaluation.
`{{}}` is a string value like `abc` or `{{"xyz"}}` and it should be returned as it like @Tooluloope mentioned.
I don't think this is a bug.

ashit-rath: If we consider the moustache syntax as a way to represent JS code then `{{}}` is similar to executing nothing. It's like calling eval with empty string or undefined.
IMO this scenario shouldn't be handled by the widgets or the validation.

@ApekshaBhosale @Rishabh-Rathod  Need your opinion on this

chandannkumar:
Tested this PR and LGTM
- Tested with multi level List widget
- Tested Total record count and their calculation of showing pages

PR #20864, opened 2023-02-22
fix: column ordering when server side pagination is enabled

### LINKED ISSUE #20858: [Bug]: Unable to reorder column positions through column headers in edit mode on the table when the first column is frozen

### Description
Unable to reorder column positions through column headers in edit mode on the table when the first column is frozen

### Steps To Reproduce

1. DnD a table
2. Turn on Server-side pagination
3. freeze the first column 
4. try reordering the unfreezed column positions

### Public Sample App

_No response_

### Issue video log

https://www.loom.com/share/b2e2b6001486406aabbe123cffa511be

### Version

Release

keyurparalkar:
## Description

Whenever server-side pagination was enabled and a column was frozen/ or a column was sorted then, columns were not orderable via dragging. This was happening because the `Table.tsx` component was not getting re-rendered when a column was frozen or sorted.

Fixes #20858 

## Type of change

## How Has This Been Tested?
- Manual

### Dev activity

### QA activity:


app/client/src/widgets/TableWidgetV2/component/index.tsx
```diff
@@ -270,6 +270,8 @@ function ReactTableComponent(props: ReactTableComponentProps) {
     });
   }, [
     props.columns.map((column) => column.alias).toString(),
+    props.columns.map((column) => column.sticky).toString(),
+    props.columns.map((column) => column.isAscOrder).toString(),
```
ashit-rath, line 274: Here we are the result of the map would look something like 
```
[ true, false, true]
```
and the `toString()` would convert it to
```
"true, false, true"
```

If due to some future changes we are able to re-order multiple columns at once 
After re-arranging the result of this array remains the same but the Column's position have changed. Wouldn't that cause some issue?

ashit-rath (commented):
Is it possible to add a cypress test?

app/client/src/widgets/TableWidgetV2/component/index.tsx
keyurparalkar: @ashit-rath for column's position, I don't think there will be any issue since we are also comparing the column order on this line.

keyurparalkar:
> Is it possible to add a cypress test?

No, we cannot write a cypress test for this since it involves dragging of the columns

app/client/src/widgets/TableWidgetV2/component/index.tsx
ashit-rath: It's a bit hard to grok what we are checking here. Although you are correct about the comparison not causing an issue, do you think we can make it better (the 3 conditions) so that future changes are easier to do and other who lack context can understands the implications of this better?

ashit-rath (approved):

laveena-en:
Tested and verified the fix. LGTM.

Raised a separate issue for the re-ordering bug on resizing the widget here: https://github.com/appsmithorg/appsmith/issues/20922

keyurparalkar:
Closing this issue, as this was causing regression on other table features

PR #20867, opened 2023-02-22
chore: Update healthcheck to use health endpoint

sharat87:
The health endpoint in the Docker container's healthcheck command is currently pointing to `/users/me`. But we have a designated endpoint for this, at `/health`. This PR switches to using this endpoint instead.


sharat87:
@sum35h @pratapaprasanna, can I get a review on this please? 🙂

pratapaprasanna (approved):

PR #20929, opened 2023-02-24
feat: Support for proxy authentication

### LINKED ISSUE #16330: [Bug]: Java properties for proxy not working as expected.

### Description

1. The system proxy is not working when the trailing slash is configured in the http_proxy or https_proxy.

2. There were no fields for proxy username and proxy password.

3. NO_PROXY is not working for the CIDR range for example (`10.*.*.*`) or (`10.171.49.0/24`), even not working if the domain name is specified like `.example.com`.

### Steps To Reproduce

1. Configure the docker http_proxy and https_proxy with a trailing slash.
2. Configure the NO_PROXY with an IP.
3. Now run the appsmith container with proxy environment variables. 
4. Log in to appsmith and create a data source API that is behind the proxy.

The below image is for the NO_PROXY reference 

image

The below image is for http_proxy and https_proxy with trailing slash.
Proxy slash

The reference implementation for NO_PROXY and proxyUser and proxyPassword.

### Public Sample App

_No response_

### Version

1.7.11

sharat87:
Appsmith supports running with a HTTP proxy, that can be configured with `HTTP_PROXY` or `HTTPS_PROXY` env variables. Like this:

```sh
HTTP_PROXY=http://myproxy:8080
HTTPS_PROXY=http://myproxy:8443
```

However, this proxy support didn't support authentication. This PR implements that. Now, proxy with authentication can be configured like this:

```sh
HTTP_PROXY=http://user:password@myproxy:8080
HTTPS_PROXY=http://user:password@myproxy:8080
```

This is not syntax or standards invented by Appsmith. This is the standard way proxy is usually configured.

Fixes #16330

🍰

### DIFF
--- a/deploy/docker/scripts/run-java.sh
+++ b/deploy/docker/scripts/run-java.sh
@@ -8,13 +8,39 @@ set -o noglob
 declare -a proxy_args
 proxy_configured=0
 
-if [[ ${HTTP_PROXY-} =~ ^http://(.*):([[:digit:]]*)/?$ && ${BASH_REMATCH[2]} != 0 ]]; then
-  proxy_args+=(-Dhttp.proxyHost="${BASH_REMATCH[1]}" -Dhttp.proxyPort="${BASH_REMATCH[2]}")
+match-proxy-url() {
+  # Examples:
+  #   http://proxy.example.com:8080/
+  #   http://user:pass@proxyhost:123
+  #   http://proxyhost:123
+  [[ $1 =~ ^http://(([^@:]*):([^@]*)?@)?([^@:]*):([0-9]+)/?$ ]]
+  proxy_user="${BASH_REMATCH[2]-}"
+  proxy_pass="${BASH_REMATCH[3]-}"
+  proxy_host="${BASH_REMATCH[4]-}"
+  proxy_port="${BASH_REMATCH[5]-}"
+  [[ -n $proxy_host ]]
+}
+
+echo "HTTP_PROXY = ${HTTP_PROXY-}"
+if match-proxy-url "${HTTP_PROXY-}"; then
+  proxy_args+=(-Dhttp.proxyHost="$proxy_host" -Dhttp.proxyPort="$proxy_port")
+  if [[ -n $proxy_user ]]; then
+    proxy_args+=(-Dhttp.proxyUser="$proxy_user")
+  fi
+  if [[ -n $proxy_pass ]]; then
+    proxy_args+=(-Dhttp.proxyPassword="$proxy_pass")
+  fi
   proxy_configured=1
 fi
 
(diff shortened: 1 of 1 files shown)


deploy/docker/scripts/run-java.sh
```diff
@@ -8,13 +8,38 @@ set -o noglob
...
 proxy_configured=0
 
-if [[ ${HTTP_PROXY-} =~ ^http://(.*):([[:digit:]]*)/?$ && ${BASH_REMATCH[2]} != 0 ]]; then
-  proxy_args+=(-Dhttp.proxyHost="${BASH_REMATCH[1]}" -Dhttp.proxyPort="${BASH_REMATCH[2]}")
+match-proxy-url() {
+  # Examples:
+  #   http://proxy.example.com:8080/
+  #   http://user:pass@proxyhost:123
+  #   http://proxyhost:123
+  [[ $1 =~ ^http://(([^@:]*):([^@]*)?@)?([^@:]*):([0-9]+)/?$ ]]
```
mohanarpit, line 16: Do we not have to do this for https proxy as well?

sharat87: For HTTPS proxies, we're already doing this, down below.

For proxies on HTTPS, that is, if the proxy URL starts with `https://`, I'd say that's a whole new feature. I'm not sure if people use proxies on HTTPS, and if it's even possible. But, it isn't supported at all so far, and I'd rather take it up separately.

mohanarpit: Understood. I was asking the question about authenticated https proxies. Currently, this PR handles authentication for HTTP proxies, but not for HTTPS proxies. As you mentioned, if we aren't supporting it right now within the product, I'd say we make this information public and merge this change in.

mohanarpit (approved):

pratapaprasanna (approved):

deploy/docker/scripts/run-java.sh
sharat87: This is very confusing dude, let me rephrase and try again.

We supported both HTTP and HTTPS proxies for quite some time now. This PR adds authentication support for _both_ HTTP and HTTPS proxies.

Screenshot 2023-02-24 at 1 49 48 PM

What's missing, is _proxy on HTTPS_. This is where the proxy itself, has an HTTPS URL, and runs with HTTPS. Here, the proxy URL will be configured like `HTTPS_PROXY=https://proxy:8080` (notice the `https://`). This is something I haven't seen in the wild, ever, I'm not aware if that's even possible, and no one's asked for it yet. We don't support this, and that's why I'm only checking for `http://` match here.

If I'm still not making sense, happy to get on a Huddle to talk it over. 🙂

mohanarpit: Thanks for the explanation. I think I wasn't too clear myself. My question was directed towards `HTTPS_PROXY=https://proxy:8080` itself. I was asking this because the regex was `http://...`. Any way, it's cleared up. I've approved the PR. We should be good to go.

PR #20933, opened 2023-02-24
feat: Add a restore point for Applications

### LINKED ISSUE #19720: [Task]: Restore point for Auto Layout Conversion BE
Terminology:
fixed apps - present day app where widgets are absolutely positioned
auto layout apps - apps that hold widgets in order

As part of Auto layout initiatives, we will be building a way to convert out present day fixed apps to auto layout apps. This conversion will be lossy because of the nature of auto layout.
So in order to make up for that, we will be providing a way to restore to the last point in time where the app was in fixed mode.

Miro Link: https://miro.com/app/board/uXjVPz_eBXA=/?share_link_id=986081276045

Requirements of the BE:

- an api to store app data as a snapshot.
- an api to use the snapshot to restore the app.
- this information should hold even if the app is deployed.
- an api to update layouts of all pages at once during app conversion.
- add snapshot information as part of application view api.
- all apis should with git sync without any issues.

nayan-rafiq:
## Description
This PR adds a way to take a snapshot of an application on the database. That snapshot can be restored to the application. 

Under the hood it uses the Export-Import feature.

Fixes #19720

Media

## Type of change

## How Has This Been Tested?
- Manual
- JUnit tests

### Dev activity

### QA activity:


app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
```diff
@@ -844,14 +843,24 @@ private Mono<Application> importApplicationInWorkspace(String workspaceId,
                                                     // The isPublic flag has a default value as false and this would be confusing to user
                                                     // when it is reset to false during importing where the application already is present in DB
                                                     importedApplication.setIsPublic(null);
+                                                    importedApplication.setPolicies(null);
```
AnaghHegde, line 846: Why are we setting this to null? We don't export the policies right.

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationControllerCE.java
```diff
@@ -178,6 +184,33 @@ public Mono<ResponseEntity<Object>> getApplicationFile(@PathVariable String id,
                 });
     }
 
+    @PostMapping("/snapshot/{id}")
+    public Mono<ResponseDTO<String>> createSnapshot(@PathVariable String id,
+                                                           @RequestHeader(name = FieldName.BRANCH_NAME, required = false) String branchName) {
```
AnaghHegde, line 189: Nit - Indentation seems to be off.

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationControllerCE.java
```diff
@@ -178,6 +184,33 @@ public Mono<ResponseEntity<Object>> getApplicationFile(@PathVariable String id,
...
+                                                                                    @RequestHeader(name = FieldName.BRANCH_NAME, required = false) String branchName) {
+        log.debug("Going to get snapshot with application id: {}, branch: {}", id, branchName);
+
+        return applicationSnapshotService.getWithoutApplicationJsonByApplicationId(id, branchName)
+                .map(applicationSnapshot -> new ResponseDTO<>(HttpStatus.OK.value(), applicationSnapshot, null));
+    }
+
+    @PostMapping("/snapshot/{id}/restore")
+    public Mono<ResponseDTO<Application>> restoreSnapshot(@PathVariable String id,
+                                                                  @RequestHeader(name = FieldName.BRANCH_NAME, required = false) String branchName) {
```
AnaghHegde, line 207: Same as above. Indentation seems off.

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationControllerCE.java
```diff
@@ -178,6 +184,33 @@ public Mono<ResponseEntity<Object>> getApplicationFile(@PathVariable String id,
                 });
     }
 
+    @PostMapping("/snapshot/{id}")
+    public Mono<ResponseDTO<String>> createSnapshot(@PathVariable String id,
+                                                           @RequestHeader(name = FieldName.BRANCH_NAME, required = false) String branchName) {
+        log.debug("Going to create snapshot with application id: {}, branch: {}", id, branchName);
+
+        return applicationSnapshotService.createApplicationSnapshot(id, branchName)
+                .map(snapshotId -> new ResponseDTO<>(HttpStatus.OK.value(), snapshotId, null));
```
AnaghHegde, line 193: Since its a POST. shouldn't we be using the HttpStatus.CREATED.value() here?

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationControllerCE.java
```diff
@@ -178,6 +184,33 @@ public Mono<ResponseEntity<Object>> getApplicationFile(@PathVariable String id,
...
+                .map(applicationSnapshot -> new ResponseDTO<>(HttpStatus.OK.value(), applicationSnapshot, null));
+    }
+
+    @PostMapping("/snapshot/{id}/restore")
+    public Mono<ResponseDTO<Application>> restoreSnapshot(@PathVariable String id,
+                                                                  @RequestHeader(name = FieldName.BRANCH_NAME, required = false) String branchName) {
+        log.debug("Going to restore snapshot with application id: {}, branch: {}", id, branchName);
+
+        return applicationSnapshotService.restoreSnapshot(id, branchName)
+                .map(application -> new ResponseDTO<>(HttpStatus.OK.value(), application, null));
```
AnaghHegde, line 211: Same. Shouldn't we be using the HttpStatus.CREATED.value() here?

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/ApplicationSnapshot.java
```diff
@@ -0,0 +1,28 @@
...
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import org.springframework.data.mongodb.core.mapping.Document;
+
+/**
+ * This stores a snapshot of an application.
+ */
+@Getter
+@Setter
+@NoArgsConstructor
```
AnaghHegde, line 15: With the `@Document` we don't need the Getter and Setter right?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
```diff
@@ -2807,4 +2809,13 @@ public void addSslModeSettingsForExistingMssqlDatasource(MongoTemplate mongoTemp
...
     }
+
+    @ChangeSet(order = "042", id = "add-index-to-application-snapshot", author = "")
+    public void addIndexToApplicationSnapshot(MongoTemplate mongoTemplate) {
+        Index applicationIdIndex = makeIndex(fieldName(QApplicationSnapshot.applicationSnapshot.applicationId))
+                .named("applicationId_unique_index")
+                .unique();
+        ensureIndexes(mongoTemplate, ApplicationSnapshot.class, applicationIdIndex);
+    }
+
```
AnaghHegde, line 2820: With the Spring boot 3.x upgrade, i don't think this migration will work. We should be creating a separate class for each migration. Please refer this - https://github.com/appsmithorg/appsmith/blob/release/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration003PermissionGroupDefaultWorkspaceIdMigration.java

Notion doc- https://www.notion.so/appsmith/Mongock-upgrade-3b4e5f32dcda47c08ef9d96cfedf15a9

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/CustomApplicationSnapshotRepositoryImpl.java
```diff
@@ -0,0 +1,13 @@
+package com.appsmith.server.repositories;
+
+import com.appsmith.server.repositories.ce.CustomApplicationSnapshotRepositoryCEImpl;
+import org.springframework.data.mongodb.core.ReactiveMongoOperations;
+import org.springframework.data.mongodb.core.convert.MongoConverter;
+import org.springframework.stereotype.Component;
+
+@Component
```
AnaghHegde, line 8: Can you please help me in understanding on why we are using the @component annotation here?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationSnapshotServiceCEImpl.java
```diff
@@ -0,0 +1,86 @@
...
+                .map(BaseDomain::getId);
+    }
+
+    private Mono<ApplicationSnapshot> createOrUpdateSnapshot(String applicationId, ApplicationJson applicationJson) {
+        return applicationSnapshotRepository.findWithoutApplicationJson(applicationId)
+                .defaultIfEmpty(new ApplicationSnapshot())
+                .flatMap(applicationSnapshot -> {
+                    String json = gson.toJson(applicationJson);
+                    applicationSnapshot.setApplicationJson(json);
+                    applicationSnapshot.setApplicationId(applicationId);
```
AnaghHegde, line 49: The maximum size of BSON document is 16MB with nested level of 100. If we have a huge application then this could lead to data corruption or data loss issues right? Can we add a check to ensure we don't run into this issue.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImplV2.java
```diff
@@ -887,14 +888,24 @@ private Mono<Application> importApplicationInWorkspace(String workspaceId,
...
                                                     // so that these won't be lost when we are pulling changes from remote and
                                                     // rehydrate the application. We are now rehydrating the application with/without
                                                     // the changes from remote
                                                     // We are using the save instead of update as we are using @Encrypted
                                                     // for GitAuth
-                                                    return applicationService.findById(existingApplication.getGitApplicationMetadata().getDefaultApplicationId())
+                                                    Mono<Application> parentApplicationMono;
+                                                    if (existingApplication.getGitApplicationMetadata() != null) {
+                                                        parentApplicationMono = applicationService.findById(
+                                                                existingApplication.getGitApplicationMetadata().getDefaultApplicationId()
```
AnaghHegde, line 902: Can you help me understand the logic here? How are we handling the non git connected Apps?

AnaghHegde (changes requested):
Can we Unit tests for ImportExport flow as well?

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImplV2.java
nayan-rafiq: @AnaghHegde for non git connected apps, we don't need to fetch the parent aka default application. Hence in the else part of this `if`, I'm returning the current application using `Mono.just()`

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/CustomApplicationSnapshotRepositoryImpl.java
nayan-rafiq: @AnaghHegde Actually this is how we add custom implementation for repository classes in spring data. In spring boot, we don't need to implement the interface, the framework itself creates an implementation when application is started. But it it sees that the repository interface extends some other interfaces, it needs their implementations as well. Using `Component` annotation, we're marking the implementation class as a bean. The spring boot will automatically pick up this as the base class when it'll implement the repository interface.

Thanks for this question because I actually followed the existing pattern without knowing much. You'll find that all the `Impl` classes under the repository package has this annotation. I explored about this just to answer this question :)

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/ApplicationSnapshot.java
nayan-rafiq: No, we need both actually. The one you're talking about is `@Data`. `Getter, Setter, NoArgsConstructor, Data` they are lombok annonations. We use them to generate Getters, Setters, Constructors on the fly. `@Document` annotation is from mongodb and it marks a class as persistable to DB.

The [`@Data`](https://projectlombok.org/features/Data) annotation is actually a wrapper that generates getters, setters and a constructor along with few other methods.  We don't use `@Data` for the domain classes because we don't need all args constructor for domain classes. That's why we use `Getter, Setter, NoArgsConstructor` annotations separately.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
nayan-rafiq: @AnaghHegde I was facing trouble because the `importedApplication` had empty policies instead of null. As a result, the `copyNestedNonNullProperties` method was not copying the policies from existing app to imported app. That's why I think better set it to null to ensure the `copyNestedNonNullProperties` works as expected.

nayan-rafiq:
> Can we Unit tests for ImportExport flow as well?

@AnaghHegde I've added unit tests for the new services. As this calls the import-export service, I didn't add another test for the import-export services. I'll try adding unit tests for that as well.

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationControllerCE.java
nayan-rafiq: @AnaghHegde, no. We're not creating any new resource through this endpoint. We're updating an existing application. That's why `HttpStatus.CREATED` should not be returned from this endpoint response.

AnaghHegde:
@nayan-rafiq Since this PR adds a support for merging the application json for non git connected apps, we can add a test case specifically for this flow.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImpl.java
AnaghHegde: @nayan-rafiq How are we getting this data? Can we ensure that this is handled in the same flow?

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/ApplicationSnapshot.java
AnaghHegde: Yes you are right. I got confused with those two. My bad.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/CustomApplicationSnapshotRepositoryImpl.java
AnaghHegde: Thanks a lot for explaining this ❤️

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationSnapshotServiceCEImpl.java
```diff
@@ -0,0 +1,96 @@
...
+    private Mono<ApplicationSnapshot> createOrUpdateSnapshot(String applicationId, ApplicationJson applicationJson) {
+        return applicationSnapshotRepository.findWithoutApplicationJson(applicationId)
+                .defaultIfEmpty(new ApplicationSnapshot())
+                .flatMap(applicationSnapshot -> {
+                    String json = gson.toJson(applicationJson);
+                    // check the size of the exported json before storing to avoid mongodb document size limit
+                    byte[] utf8JsonString = json.getBytes(StandardCharsets.UTF_8);
+                    if(utf8JsonString.length > MAX_SNAPSHOT_SIZE) {
+                        // file may exceed 16 MB document size limit of mongodb, throw error
+                        return Mono.error(new AppsmithException(AppsmithError.GENERIC_BAD_REQUEST, "Application too large for snapshot"));
```
AnaghHegde, line 56: Do we need to think of any other solutions here? Reason i am asking is because the user cannot try this feature on a large application. Is that correct?

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ApplicationSnapshotServiceTest.java
```diff
@@ -0,0 +1,186 @@
...
+import reactor.core.publisher.Mono;
+import reactor.test.StepVerifier;
+import reactor.util.function.Tuple2;
+
+import java.util.UUID;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@SpringBootTest
+public class ApplicationSnapshotServiceTest {
```
AnaghHegde, line 23: Can we assert for a failure case as well?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationSnapshotServiceCEImpl.java
nayan-rafiq: @AnaghHegde I've updated the code. Instead of storing the applicationJson in a single collection, now I'm splitting it into multiple objects when the size limit exceeds.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationSnapshotServiceCEImpl.java
```diff
@@ -0,0 +1,137 @@
...
+                    String applicationJsonString = objects.getT1();
+                    Application application = objects.getT2();
+                    ApplicationJson applicationJson = gson.fromJson(applicationJsonString, ApplicationJson.class);
+                    return importExportApplicationService.importApplicationInWorkspace(
+                            application.getWorkspaceId(), applicationJson, application.getId(), branchName
+                    );
+                });
+    }
+
+    private Mono<String> getApplicationJsonString(String applicationId) {
```
AnaghHegde, line 97: Nit - can we rename this to `getApplicationJsonStringFromSnapShot`

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationSnapshotServiceCEImpl.java
```diff
@@ -0,0 +1,137 @@
...
+                        Mono.zip(
+                                importExportApplicationService.exportApplicationById(branchedAppId, SerialiseApplicationObjective.VERSION_CONTROL),
+                                Mono.just(branchedAppId)
+                        )
+                )
+                .flatMapMany(objects -> {
+                    String branchedAppId = objects.getT2();
+                    ApplicationJson applicationJson = objects.getT1();
+                    return applicationSnapshotRepository.deleteAllByApplicationId(branchedAppId)
+                            .thenMany(createSnapshots(branchedAppId, applicationJson));
```
AnaghHegde, line 51: Can you help me in understanding why do we need to use `thenMany` here?

AnaghHegde (approved):
Minor comments. Rest Looks good to me

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationSnapshotServiceCEImpl.java
nayan-rafiq: The `createSnapshots` method returns a Flux. We use `flatMap` on a `Mono` when we return another `Mono` from the `flatMap` method. But if we need to return a `Flux` from the `flatMap` method, we need to use `flatMapMany`. The `createSnapshots` method returns a `Flux` so we need to use `flatMapMany`.

PR #20964, opened 2023-02-26
perf: switch to yarn 3 + dedupe dependencies

iamakulov:
## Description

This PR:
- upgrades Appsmith to Yarn v3
  - see what’s new: https://yarnpkg.com/getting-started/qa#why-should-you-upgrade-to-yarn-modern
  - this upgrade also makes it possible to enable Yarn Plug'n'Play, which will make Yarn commands even faster. I haven’t attempted this (it needs a bit of work) but I’d recommend trying it out: https://yarnpkg.com/features/pnp
- migrates the `shared/ast` to Yarn Workspaces
  - this makes managing shared packages much simpler (you don’t need the whole `build-shared-dep.js`, `install-dependencies.js`, etc logic)
  - and makes Yarn runs cheaper (`shared/ast` is now rebuilt only when needed instead of on every `yarn add`)
- replaces `patch-package` with Yarn’s native package patching
- dedupes all packages
  - (did you know we had 4 versions of `@blueprintjs/core` installed?)
- makes `yarn dedupe` run on every install

Apart from DX improvements, this also makes the bundle almost 10% smaller (−870 kB, 9.01 → 8.14 MB minified).

## Type of change

## How Has This Been Tested?

- Manual: smoke testing (clicking around the dashboard and the editor)

### Dev activity

### QA activity:


app/client/package.json
```diff
@@ -300,18 +297,5 @@
     "hooks": {
       "pre-commit": "lint-staged && git-secrets --scan --untracked && git-secrets --scan -r"
     }
-  },
-  "resolutions": {
```
yaldram, line 304: These resolutions are necessary, because we want to use the following package versions, that have no security vulnerabilities, I fear removing these resolutions might install packages that have vulnerabilities.

SatishGandham, line 304: Why did we move these to app/package.json?

app/client/patches/@blueprintjs+core+3.36.0.patch
```diff
@@ -1,35 +0,0 @@
-diff --git a/node_modules/@blueprintjs/core/lib/esm/components/editable-text/editableText.js b/node_modules/@blueprintjs/core/lib/esm/components/editable-text/editableText.js
-index 84f03fa..5e5488a 100644
---- a/node_modules/@blueprintjs/core/lib/esm/components/editable-text/editableText.js
-+++ b/node_modules/@blueprintjs/core/lib/esm/components/editable-text/editableText.js
-@@ -188,7 +188,16 @@ var EditableText = /** @class */ (function (_super) {
-         if (this.state.isEditing && !prevState.isEditing) {
-             (_b = (_a = this.props).onEdit) === null || _b === void 0 ? void 0 : _b.call(_a, this.state.value);
-         }
--        this.updateInputDimensions();
-+        // updateInputDimensions is an expensive method. Call it only when the props
```
SatishGandham, line 10: We are removing this patch, is it fixed in the version of blue-print we are using?

SatishGandham (changes requested):

app/client/package.json
```diff
@@ -8,7 +8,7 @@
   },
   "cracoConfig": "craco.dev.config.js",
   "dependencies": {
-    "@blueprintjs/core": "^3.36.0",
+    "@blueprintjs/core": "^3.43.0",
```
yaldram, line 11: Blueprint is updated here, might need to check if everything is working as expected.

yaldram (approved):

app/client/package.json
iamakulov: Oh, that’s a great q – I wasn’t quite sure myself when I was submitting the PR, but I have since figured out why. 

Basically:

1. CRA [sets a webpack flag](https://github.com/facebook/create-react-app/blob/d960b9e38c062584ff6cfb1a70e1512509a966e7/packages/react-scripts/config/webpack.config.js#L306-L312) called `resolve.modules`:
  
   ```js
   module.exports = {
     resolve: {
       modules: ['node_modules', 'src']
     }
   }
   ```
  
   This tells webpack where to look up package imports like `import 'foo/bar.js'`. With this config, webpack will try to look for `foo/bar.js` in `node_modules/foo/bar.js` and then in `src/foo/bar.js`. 
      
2. We have multiple versions of the `entities` package installed as a transitive dependency of `htmlparser2`, `domutils`, and more.

   Before this PR, the version of the package hoisted into `app/client/node_modules/entities` was `2.0.0`. After moving to Yarn workspaces and deduping dependencies, the version of the package hoisted into `app/node_modules/entities` became `4.4.0`.

   The key difference between `2.0.0` and `4.4.0` is that `4.4.0` has [an `exports` field](https://github.com/webpack/webpack/issues/9509) in its `package.json`:
   
    ```json
    {
        "exports": {
            ".": {
                "require": "./lib/index.js",
                "import": "./lib/esm/index.js"
            },
            "./lib/decode.js": {
                "require": "./lib/decode.js",
                "import": "./lib/esm/decode.js"
            },
            "./lib/escape.js": {
                "require": "./lib/escape.js",
                "import": "./lib/esm/escape.js"
            }
        }
    }
    ```
    
3. When the `exports` field is present, webpack (and Node.js) start verifying package imports more strictly. Without the `exports` field, when we had this import:

   ```js
   import { APP_MODE } from "entities/App";
   ```
   
   webpack would try to resolve it `node_modules/entities/App`, see that that file doesn’t exist, and try `src/entities/App` next.
   
   Now, with the `exports` field, webpack additionally verifies whether `entities/App` is an allowed export. And since the `exports` object above ↑ doesn’t include `App`, webpack throws an error:
   
    ```
    ERROR in ./src/RouteBuilder.ts 5:0-40
    Module not found: Error: Package path ./App is not exported from package /Users/iamakulov/Repos/appsmith/app/node_modules/entities (see exports field in /Users/iamakulov/Repos/appsmith/app/node_modules/entities/package.json)
    ```
    
4. That’s why the change is needed. It puts `src` as the first item in `require.modules` – and makes webpack try importing `src/entities/App` first. This resolves the error.

   The alternative is to rename `entities` to something else, but I’d rather solve the root issue. What if we install eg the `reflow` package next?

iamakulov: Linked to this from the comment to provide more background 👍

app/client/patches/@blueprintjs+core+3.36.0.patch
iamakulov: Correct! https://github.com/palantir/blueprint/pull/4723 was released in v3.44.3. Due to `yarn dedupe`, we’re now using Blueprint v3.47.0 (instead of 3.36.0 + 3.33.0 + 3.43.0 + 3.47.0, [like before](https://share.cleanshot.com/CMml4mCf)).

app/client/craco.common.config.js
```diff
@@ -64,5 +65,19 @@ module.exports = {
         css: false,
       },
     },
+    {
+      plugin: {
```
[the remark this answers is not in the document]
iamakulov: Oh, that’s a great q – I wasn’t quite sure myself when I was submitting the PR, but I have since figured out why. 

Basically:

1. CRA [sets a webpack flag](https://github.com/facebook/create-react-app/blob/d960b9e38c062584ff6cfb1a70e1512509a966e7/packages/react-scripts/config/webpack.config.js#L306-L312) called `resolve.modules`:
  
   ```js
   module.exports = {
     resolve: {
       modules: ['node_modules', 'src']
     }
   }
   ```
  
   This tells webpack where to look up package imports like `import 'foo/bar.js'`. With this config, webpack will try to look for `foo/bar.js` in `node_modules/foo/bar.js` and then in `src/foo/bar.js`. 
      
2. We have multiple versions of the `entities` package installed as a transitive dependency of `htmlparser2`, `domutils`, and more.

   Before this PR, the version of the package hoisted into `app/client/node_modules/entities` was `2.0.0`. After moving to Yarn workspaces and deduping dependencies, the version of the package hoisted into `app/node_modules/entities` became `4.4.0`.

   The key difference between `2.0.0` and `4.4.0` is that `4.4.0` has [an `exports` field](https://github.com/webpack/webpack/issues/9509) in its `package.json`:
   
    ```json
    {
        "exports": {
            ".": {
                "require": "./lib/index.js",
                "import": "./lib/esm/index.js"
            },
            "./lib/decode.js": {
                "require": "./lib/decode.js",
                "import": "./lib/esm/decode.js"
            },
            "./lib/escape.js": {
                "require": "./lib/escape.js",
                "import": "./lib/esm/escape.js"
            }
        }
    }
    ```
    
3. When the `exports` field is present, webpack (and Node.js) start verifying package imports more strictly. Without the `exports` field, when we had this import:

   ```js
   import { APP_MODE } from "entities/App";
   ```
   
   webpack would try to resolve it `node_modules/entities/App`, see that that file doesn’t exist, and try `src/entities/App` next.
   
   Now, with the `exports` field, webpack additionally verifies whether `entities/App` is an allowed export. And since the `exports` object above ↑ doesn’t include `App`, webpack throws an error:
   
    ```
    ERROR in ./src/RouteBuilder.ts 5:0-40
    Module not found: Error: Package path ./App is not exported from package /Users/iamakulov/Repos/appsmith/app/node_modules/entities (see exports field in /Users/iamakulov/Repos/appsmith/app/node_modules/entities/package.json)
    ```
    
4. That’s why the change is needed. It puts `src` as the first item in `require.modules` – and makes webpack try importing `src/entities/App` first. This resolves the error.

   The alternative is to rename `entities` to something else, but I’d rather solve the root issue. What if we install eg the `reflow` package next?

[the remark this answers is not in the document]
iamakulov: Linked to this from the comment to provide more background 👍

app/client/package.json
iamakulov: Yarn requires resolutions to be present in the root `package.json` – because they’re used to resolve all dependencies across all workspaces.

This is what you’ll see if you try to move `resolutions` back to `app/client/package.json`:

```sh
$ yarn
➤ YN0000: ┌ Project validation
➤ YN0057: │ appsmith: Resolutions field will be ignored
➤ YN0000: └ Completed
...
```

app/.yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable */
```
SatishGandham, line 1: Should this file be committed?

iamakulov: Yes! That’s by design; it prevents accidental version mismatches between environments: https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored

PR #20965, opened 2023-02-26
feat: Add static header for CSRF protection

sharat87:
We add a static constant value header, to all requests made by the client, whose method isn't `GET` or `HEAD`.

More discussion in [this Slack thread](https://theappsmith.slack.com/archives/C03RPDB936Z/p1669701485071579?thread_ts=1669460732.544859&cid=C03RPDB936Z).


abhvsn (dismissed):
LGTM

app/server/appsmith-server/src/main/java/com/appsmith/server/filters/CSRFFilter.java
```diff
@@ -0,0 +1,51 @@
...
+        if (!isGetOrHead && !EXEMPT.contains(request.getPath().value())) {
+            // For POST requests, either a `X-Requested-By: Appsmith` header or a `Content-Type: application/json`
+            // is required. If neither is present, reject the request. This is to prevent CSRF attacks.
+            if (MediaType.APPLICATION_JSON.equals(request.getHeaders().getContentType())
+                    || "Appsmith".equals(request.getHeaders().getFirst("X-Requested-By"))) {
+                return chain.filter(exchange);
+            }
+
+            log.error("CSRF header requirements not satisfied to {}. Rejecting request.", request.getPath());
+            return Mono.error(new RuntimeException("CSRF header requirements not satisfied. Rejecting request."));
```
abhvsn, line 45: @sharat87 any reason we are not throwing the Appsmith custom error here?

app/server/appsmith-server/src/main/java/com/appsmith/server/filters/CSRFFilter.java
```diff
@@ -0,0 +1,51 @@
...
+    public Mono<Void> filter(ServerWebExchange exchange, WebFilterChain chain) {
+        final ServerHttpRequest request = exchange.getRequest();
+        final HttpMethod method = request.getMethod();
+        final boolean isGetOrHead = HttpMethod.GET.equals(method) || HttpMethod.HEAD.equals(method);
+
+        if (!isGetOrHead && !EXEMPT.contains(request.getPath().value())) {
+            // For POST requests, either a `X-Requested-By: Appsmith` header or a `Content-Type: application/json`
+            // is required. If neither is present, reject the request. This is to prevent CSRF attacks.
+            if (MediaType.APPLICATION_JSON.equals(request.getHeaders().getContentType())
+                    || "Appsmith".equals(request.getHeaders().getFirst("X-Requested-By"))) {
```
abhvsn, line 40: Can we move these Strings to class-level constants?

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
```diff
@@ -112,8 +114,9 @@ public SecurityWebFilterChain securityWebFilterChain(ServerHttpSecurity http) {
         ServerAuthenticationEntryPointFailureHandler failureHandler = new ServerAuthenticationEntryPointFailureHandler(authenticationEntryPoint);
 
         return http
-                // This picks up the configurationSource from the bean corsConfigurationSource()
+                // The native CSRF solution doesn't work with WebFlux, yet, but only for WebMVC. So we make our own.
```
abhvsn, line 117: @sharat87 Is the CSRF token not getting generated for webflux is the issue here?
https://github.com/spring-projects/spring-security/issues/6046

abhvsn (commented):
Minor comments but overall LGTM

app/server/appsmith-server/src/main/java/com/appsmith/server/filters/CSRFFilter.java
sharat87: Good call. Done.

app/server/appsmith-server/src/main/java/com/appsmith/server/filters/CSRFFilter.java
sharat87: I thought this should be a secret internal error when I was working on this, which, honestly doesn't make the least sense to me now. 🤔

You're right, it should be an Appsmith error, and should be a 4xx error. Changed this. Thanks.

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
sharat87: Yes. But more than even that, the trouble of going through Spring's solution for this, and seeing the support they give for customization of these solutions for WebFlux, just didn't inspire confidence in me. Static header value is solution enough for us. 🙂

Does that sound okay?

abhvsn (approved):
LGTM

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
abhvsn: Agreed 👍

PR #20972, opened 2023-02-27
perf: move evaluation worker dependencies to the evaluation worker

iamakulov:
## Description

The Evaluation worker has several large dependencies:
- `linkedom/worker`
- `node-forge`
- tern JSON definitions

This PR removes them from the main bundle:

#### cce52da36b4629ba5cd13793d41e4b326b6fc5d0, 39634282598f2f9b9e0bc94510de792b5421ff45: move `linkedom/worker` and `node-forge` out of the main bundle

Issues:

- `linkedom/worker` is a DOM mock for the worker environment. It’s supposed to be used only in the evaluation worker. However, it was bundled to the main bundle because `EvaluationsSaga.ts` [calls `validate()`](https://github.com/appsmithorg/appsmith/blob/94caa07a38627f51f32fdbe1e014a61d29c28024/app/client/src/sagas/EvaluationsSaga.ts#L659) → [calls `VALIDATORS.FUNCTION()`](https://github.com/appsmithorg/appsmith/blob/94caa07a38627f51f32fdbe1e014a61d29c28024/app/client/src/workers/Evaluation/validations.ts#L318) → [calls `evaluate()`](https://github.com/appsmithorg/appsmith/blob/94caa07a38627f51f32fdbe1e014a61d29c28024/app/client/src/workers/Evaluation/validations.ts#L1096) → [calls `resetWorkerGlobalScope()`](https://github.com/appsmithorg/appsmith/blob/94caa07a38627f51f32fdbe1e014a61d29c28024/app/client/src/workers/Evaluation/evaluate.ts#L243) → imports `DOM_APIS` [which relies on `linkedom/worker`](https://github.com/appsmithorg/appsmith/blob/94caa07a38627f51f32fdbe1e014a61d29c28024/app/client/src/workers/Evaluation/SetupDOM.ts#L4). `resetWorkerGlobalScope()` is supposed to be used only in the worker, but it’s actually used in the main thread as well.

- Similarly, `node-forge` is a Node.js crypto library. It’s supposed to be used only in the evaluation worker. However, it’s bundled into the main bundle for similar reasons: `resetWorkerGlobalScope()` relies on the `JSLibraries` definitions, and these definitions include the libraries themselves:

  https://github.com/appsmithorg/appsmith/blob/1cf9790a4634345ca2dbb248167119a1ee82e611/app/client/src/workers/common/JSLibrary/index.ts#L14-L45

  If `resetWorkerGlobalScope()` was used only in the worker, this wouldn’t be an issue. But the main thread also relies on it, which leads to `node-forge` being bundled into the main bundle.

These two commits solve this with a minimal, practically-effective-but-could’ve-been-better refactoring. Ideally, this should be solved with a holistic approach:
- Reworking the `resetWorkerGlobalScope()` function to not rely on any existing definitions at all. For example, what if we dropped `resetWorkerGlobalScope()` completely – and instead recorded all global values right before the function invocation + restored them right after?

  ```ts
  export default function evaluateSync(...): EvalResult {
    return (function() {
      const globalScope = getGlobalScope();
      //
      restoreGlobalScope(globalScope);
      return { result, errors };
    })();
  }
  ```

  **Question:** should I do this? Is there any reason this is a bad idea? I’m not intimately familiar with evaluator’s requirements, so I’m worried you thought of this already, and it didn’t work for some reason.
  
- And/or more strictly separating the worker code from the main thread code, to prevent the main thread from accidentally importing `resetWorkerGlobalScope()`. (Note: RFC incoming.)

  This isn’t really a possibility here since `resetWorkerGlobalScope()` is actually used by the main thread. (The main thread evaluates code too, through [the `EvaluationsSaga`](https://github.com/appsmithorg/appsmith/blob/94caa07a38627f51f32fdbe1e014a61d29c28024/app/client/src/sagas/EvaluationsSaga.ts#L659). But perhaps, with a stricter main/worker thread separation, the developers would notice the accidentally bundled libraries sooner.

## Type of change

## How Has This Been Tested?

- Manual (smoke testing): changing stuff in the app, adding buttons with functional onClick listeners, observing that the app doesn’t crash and the console doesn’t have any (new) errors

### Dev activity

### QA activity:


Rishabh-Rathod:
Hey @iamakulov, Thanks for working on this.  



From the above image of the main bundle, even `jsHint` could be removed from the main bundle as it is only used in `lint.worker` and need not be present in the main bundle or `eval.worker` bundle.

iamakulov:
Hey @Rishabh-Rathod, great catch! That’s coming up in a separate PR indeed.

SatishGandham (approved):
LGTM. I will wait for review from @arunvjn  or @Rishabh-Rathod

Rishabh-Rathod (approved):
Changes LGTM.

iamakulov:
> That’s coming up in a separate PR indeed.

I thought I’d have a separate PR, but it turns out that (because I’m using a fork) I can’t base a new PR upon this one, so I’m just pushing changes right here.

I’d appreciate your review of two new commits: https://github.com/appsmithorg/appsmith/pull/20972/commits/5dbcad87765320560e6fbdec6efba2c6e9780ca4 and https://github.com/appsmithorg/appsmith/pull/20972/commits/7cbb12af886621256224be0c93e6a465dd710ad3!

Aishwarya-U-R:
Please merge latest & run. Git case failures should go away by that.

iamakulov:
(Although nevermind, just pushed. However, this is still blocked by https://github.com/appsmithorg/appsmith/pull/21269 and won’t build/pass tests without it.)

PR #21031, opened 2023-02-28
feat: default selected item

### LINKED ISSUE #13214: [Feature]: List widget - default selected item

### Summary

As a user, I should be able to set a default selected item while using the list widget

Use Case:
Depending on the item selected a Query / API functions, It would be helpful if programatically I can set a default selected item in the list that onPageLoad is set.

Tooluloope:
## Description

`defaultSelectedItem` property was added to the List widget. This enables the User to Select any item when the list widget loads. This selection makes use of the Data identifier property and it currently for ClientSide pagination.

Fixes #13214 

## Type of change

## How Has This Been Tested?
- Cypress
- Manual

### Dev activity

### QA activity:
- [x] Added Test Plan Approved label after reveiwing all Cypress test


ashit-rath:
@Tooluloope This is sitting in the Review pipeline with no reviewer assigned. If this is still WIP, then move this to the appropriate pipeline

app/client/src/widgets/ListWidgetV2/widget/propertyConfig.ts
```diff
@@ -114,6 +121,41 @@ export const primaryColumnValidation = (
   };
 };
 
+export function defaultSelectedItemValidation(
```
ashit-rath, line 124: add jest tests for this

app/client/src/widgets/ListWidgetV2/widget/propertyConfig.ts
```diff
@@ -82,7 +85,11 @@ export const primaryColumnValidation = (
 
     const areKeysUnique = _.uniq(inputValue).length === listData.length;
 
-    if (!areKeysUnique) {
+    const isDataTypeUnique =
+      _.uniqBy(inputValue, (item: any) => item.toString()).length ===
+      listData.length;
+
+    if (!areKeysUnique || !isDataTypeUnique) {
```
ashit-rath, line 92: Can we update the jest test cases to accommodate this?

app/client/src/widgets/ListWidgetV2/widget/propertyConfig.ts
```diff
@@ -279,37 +324,31 @@ export const PropertyPaneContentConfig = [
...
+    sectionName: "Item Selection",
     children: [
       {
-        propertyName: "isVisible",
-        label: "Visible",
-        helpText: "Toggles the visibility of this List to end users",
-        controlType: "SWITCH",
-        isJSConvertible: true,
+        propertyName: "defaultSelectedItem",
+        helpText: "Selects Item by default",
```
ashit-rath, line 331: I feel we should have some information to give a correlation between `Data Identifier` and `Default Selected Item`

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -524,6 +538,52 @@ class ListWidget extends BaseWidget<
...
+      this.updateSelectedItemView(rowIndex);
+    }
+  };
+
+  getRowIndexOfSelectedItem = () => {
+    const { defaultSelectedItem, primaryKeys } = this.props;
+
+    if (!primaryKeys || isNil(defaultSelectedItem)) return -1;
+
+    return map(primaryKeys, (key) => String(key)).indexOf(
```
ashit-rath, line 578: Since the utility of stringified keys are a requirement here, should we convert primary keys to string and then pass to metaWidgetGenerator and use that here as well?

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -258,6 +262,16 @@ class ListWidget extends BaseWidget<
     }
 
     this.setupMetaWidgets(prevProps);
+
+    // Should come after setting up MetaWidgets for that page.
```
ashit-rath, line 266: Why do we need to setup meta widgets and then apply selectedItem? Is it because we use `this.metaWidgetGenerator.getPrimaryKey(rowIndex);` in `updateSelectedItemFromDefaultSelectedKey`.

If yes then why do we need to use that method to get primaryKey?

app/client/src/widgets/ListWidgetV2/widget/index.tsx
Tooluloope: That would mean moving the `generatePrimaryKeys` function out of the `MetaWidgetGenerator` to the list widget?

app/client/src/widgets/ListWidgetV2/widget/index.tsx
Tooluloope: It's solely for the `SelectedItemView`, `SelectedItem` is dependent on data and not on the `metaWidgetGenerator`, but I decided to put the two together to make it readable. So it's clear we're updating `SelectedItem`, `SelectedItemKey` and `SelectedItemView` together

app/client/src/widgets/ListWidgetV2/widget/index.tsx
ashit-rath: I'm just referring to the strigification of the primary keys before passing on to the meta widget generator. Not the complete logic

app/client/src/widgets/ListWidgetV2/widget/index.tsx
ashit-rath: if we update the `selectedItem` and `selectedIndex` along with `pageNo` at the same time do we get some issue with `selectedItem` update?

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -258,6 +262,16 @@ class ListWidget extends BaseWidget<
...
+    if (this.pageChangeEventTriggerFromSelectedKey) {
+      this.updateSelectedItemFromDefaultSelectedKey();
+      this.pageChangeEventTriggerFromSelectedKey = false;
+    }
+
+    if (this.props.defaultSelectedItem !== prevProps.defaultSelectedItem) {
+      this.updatePageNoFromDefaultSelectedKey();
```
ashit-rath, line 273: This method does more than just updating the `pageNo`, can we do a better name?

app/client/src/widgets/ListWidgetV2/widget/index.tsx
vsvamsi1, line 578: We don't have to use lodash map, es6 supports it. You could try using findIndex.
`
primaryKeys.findIndex((val)=> String(val)===defaultSelectedItem.toString())

`

app/client/src/widgets/ListWidgetV2/widget/index.tsx
Tooluloope: I didn't get the question. We don't use `SelectedIndex` anymore. Do you mean `selectedKey`.

Also, The main reason why I'm doing it this way is for when we start working on default Selected for serverSide. We won't know the key or anything. So it's best to select them after they've been rendered

ashit-rath: The POV that I'm thinking is, the selectedKey and selectedItem is sort of independent of meta widgets that have been generated or not.

Client side
-------------
we have default key
we have all the data
we find the key in the data and set the key and the item; also move to that page
if we don't find the key, we don't do anything


Server side
------------
we have default key
we don't have all the data
we cannot find the data item and we cannot move to that page (unless we have some more info)


in both the cases do we need to wait for the meta widgets to get generated before we set these? The meta widget generator is just going to cache the keys we pass them and nothing more.
Do you think this is going to cause problem in setting selectedItemView?

ashit-rath (approved):

kamakshibhat-appsmith:
The default selected item doesn't work on page refresh with hardcoded data https://www.loom.com/share/879508765f0e4e8dae086624830fca36

kamakshibhat-appsmith:
Also, The total records count is not updating when we refresh the page in a certain case https://www.loom.com/share/5795caf78e9248cfb829aa8d05cc863e

kamakshibhat-appsmith:
@Tooluloope Please check the following
1. The default selected Item doesn't work on query refresh [LOOM](https://www.loom.com/share/a15e7ec20b7f4ade8e349532c30bdab9)
2. The app is crashing when you configure a query that did not run in the first place, then you re-run it to get the data [LOOM](https://www.loom.com/share/7508237a9c404636abf89e5f7daa2dfd)
3. Configuring the default selected item first and then configuring the data leads to failure of default selection [LOOM](https://www.loom.com/share/bb73dbb8a87b4c0182c4e6ba09452ef1)
 4. List Widget crashes - STEPS : make the default selected item of the PAGE 2, Change the page, Go to DataSources and come back, widget crashes [LOOM](https://www.loom.com/share/5cac5ca784b049c9a0ee367a6a836d0f)
 5. The list behaves weirdly when we have any wrong data identifier [LOOM](https://www.loom.com/share/87ff4ebdb44842d19dbf641c1fe01f8e)
 6. Default selected item doesn't work properly for nested lists [LOOM](https://www.loom.com/share/1aa1697a80e947a681ec28e707ed633c)

kamakshibhat-appsmith:
@Tooluloope this is in Needs QA but i dont see the deploy preview triggered for it. Can you pls confirm?

kamakshibhat-appsmith:
1. Unable to see item selected state (Blue border) when Primary key is empty/dummy [LOOM](https://www.loom.com/share/f9128a2e1d57472684277cb7854851b4)
2. UI behaving weirdly when we have server side pagination ON but NOT onPageChange Configured  [LOOM](https://www.loom.com/share/82bb61fef2df4d5c8e8b8a65082a7848)

kamakshibhat-appsmith:
> @Tooluloope
> 
> 1. Unable to see item selected state (Blue border) when Primary key is empty/dummy [LOOM](https://www.loom.com/share/f9128a2e1d57472684277cb7854851b4)
> 2. UI behaving weirdly when we have server side pagination ON but NOT onPageChange Configured  [LOOM]
>    (https://www.loom.com/share/82bb61fef2df4d5c8e8b8a65082a7848)
> 3. When you slowly type the default selected Item, it doesn't update the List to the the Page where it is actually located however , the item is selected

 @Tooluloope Created a separate issue for Point 2 here https://github.com/appsmithorg/appsmith/issues/21488

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -718,39 +903,43 @@ class ListWidget extends BaseWidget<
...
+  };
+
+  getItemViewBindingByRowIndex = (rowIndex: number) => {
+    const container =
       this.metaWidgetGenerator.getRowContainerWidgetName(rowIndex);
 
-    const selectedItemViewBinding = triggeredContainer
-      ? `{{ ${triggeredContainer}.data }}`
-      : "{{{}}}";
+    const itemViewBinding = container ? `{{ ${container}.data }}` : {};
```
ashit-rath, line 913: We are returning `{}` from here. Shouldn't it be "{{{}}}"?

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -625,14 +799,19 @@ class ListWidget extends BaseWidget<
...
    */
   shouldUpdateCacheKeys = (prevProps: ListWidgetProps) => {
     return (
       this.props.triggeredItemKey !== prevProps.triggeredItemKey ||
       this.props.selectedItemKey !== prevProps.selectedItemKey ||
-      (!prevProps.serverSidePagination && this.props.serverSidePagination)
+      (!prevProps.serverSidePagination && this.props.serverSidePagination) ||
+      (this.props.selectedItemKey &&
+        !this.metaWidgetGenerator
+          .getCurrCachedRows()
```
ashit-rath, line 813: We still use the word `Row` at a bunch of different places. We should slowly refactor these and call them `item`. Don't do it here, take this up in some other smaller PR

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -625,14 +799,19 @@ class ListWidget extends BaseWidget<
...
+          .has(this.props.selectedItemKey))
```
ashit-rath, line 814: Since `shouldUpdateCacheKeys` is called in `componentDidUpdate`. Wouldn't this always return true for any kind of component update? Does this have any side-effect?

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -536,6 +621,95 @@ class ListWidget extends BaseWidget<
     super.updateWidgetProperty("pageSize", this.pageSize);
   };
 
+  // This is only for client-side data
+  updatePageNumber = (key: string) => {
```
ashit-rath, line 625: shouldn't we make key optional? as in the next line we are checking of `!key`

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -536,6 +621,95 @@ class ListWidget extends BaseWidget<
...
+    if (this.props.serverSidePagination || !key) return;
```
ashit-rath, line 626: if a key has a falsey value like `null`, undefined` or `""` and it is unique among the primary keys. Should it be a valid primary key and we shouldn't return from here?

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -338,6 +405,24 @@ class ListWidget extends BaseWidget<
     };
   };
 
+  generatePrimaryKeys = () => {
```
ashit-rath, line 408: Why do we have this outside?

kamakshibhat-appsmith:
@Tooluloope 
-  When I try to reset the List with SSP on, it is highlighting the default selected item [LOOM](https://www.loom.com/share/0de6ffc014a641fca7ac773ff032f43f)

app/client/src/widgets/ListWidgetV2/widget/index.tsx
Tooluloope: I realized it was extra work for eval, since eval would have to evaluate this, So why not just pass the right value

ashit-rath: The reason why this was done because, when we have a non binding value; the path is removed from `dynamicBindingPathList` but when we add a genuine binding back, the path does not get added to `dynamicBindingPathList`. This is what I noticed before but I might be wrong so double check this.

app/client/src/widgets/ListWidgetV2/widget/index.tsx
Tooluloope: Got it, I'll had a task for this.

app/client/src/widgets/ListWidgetV2/widget/index.tsx
Tooluloope: It still remains in the dynamicBindingPathList


https://user-images.githubusercontent.com/31691737/225845545-17b6745b-ee12-4c36-a5f9-4ca13ad40f8b.mov

app/client/src/widgets/ListWidgetV2/widget/index.tsx
Tooluloope: undefined or null cant be unique, this would be picked up in validations. Also it's guaranteed that the key would be a string. The only edge case would be empty string.

kamakshibhat-appsmith:
@Tooluloope 
-The UI of the list is not updating to default selected Item when we reset the widget (the new selected item should be in the same page where default selected Item is) [LOOM](https://www.loom.com/share/42f36197f80d4c5d8e73b40090dffa8f)
- Some issue with the auto height logic in text widget failing when resetting the widget (Not sure if its related to Reset but its consistent) [LOOM](https://www.loom.com/share/f7b7ad9836e94f7495b1b9b3a96c3b66)

Tooluloope:
@kamakshibhat-appsmith 

> Some issue with the auto height logic in text widget failing when resetting the widget (Not sure if its related to Reset but its consistent) [LOOM](https://www.loom.com/share/f7b7ad9836e94f7495b1b9b3a96c3b66)

This is an auto-height issue, not a list widget issue.

ashit-rath (approved):

kamakshibhat-appsmith:
Hi @pratapaprasanna , The following [DP](https://ce-21031.dp.appsmith.com) is giving me a 503 error when I am trying to access APP. This is happening with the accounts which were used earlier for the same DP . any idea what's happening?https://ce-21031.dp.appsmith.com

kamakshibhat-appsmith:
@Tooluloope I tested the pending items on a different account on the DP and it looks good

PR #21036, opened 2023-02-28
poc: Increase widgets height in factor of 10px

aswathkk:
Increase widgets height in factor of 10px


aswathkk:
Closing this since this is no longer relevant

PR #21078, opened 2023-03-01
ci: build lite version of deploy preview

pratapaprasanna:
**Case-1:**

**`/build-deploy-preview skip-tests=true`**

**Client-jest skipped**
Screenshot 2023-03-01 at 3 40 04 PM
**Server Build taking less time**
Screenshot 2023-03-01 at 3 40 46 PM

**Case-2:**

**`/build-deploy-preview `**
Screenshot 2023-03-02 at 11 54 37 AM

**Case-3**

**`/build-deploy-preview  env=release`**

Screenshot 2023-03-07 at 8 05 38 PM

**github-action**
Screenshot 2023-03-07 at 8 05 58 PM

**final-dp:**

Screenshot 2023-03-07 at 8 06 46 PM

### DIFF
--- a/.github/workflows/client-build.yml
+++ b/.github/workflows/client-build.yml
@@ -9,6 +9,10 @@ on:
         description: "This is the PR number in case the workflow is being called in a pull request"
         required: false
         type: number
+      skip-tests:
+        description: "This is a boolean value in case the workflow is being called in build deploy-preview"
+        required: false
+        type: string
 
   pull_request:
     branches: [release, master]
@@ -63,7 +67,7 @@ jobs:
         run: |
           PGPASSWORD='${{secrets.APPSMITH_PERFORMANCE_DB_PASSWORD}}' psql -h '${{secrets.APPSMITH_PERFORMANCE_DB_HOST}}' \
           -U aforce_admin -d perf-infra -c \
-          "INSERT INTO public.run_meta (repo, gh_run_id, gh_run_attempt, is_active) 
+          "INSERT INTO public.run_meta (repo, gh_run_id, gh_run_attempt, is_active)
           VALUES ('${{github.repository}}', '${{github.run_id}}', '${{github.run_attempt}}', FALSE)"
 
       # In case this is second attempt try restoring status of the prior attempt from cache
@@ -117,7 +121,7 @@ jobs:
 
       # Run the Jest tests only if the workflow has been invoked in a PR and the previous re-run has failed
       - name: Run the jest tests
-        if: steps.run_result.outputs.run_result != 'success' && inputs.pr != 0
+        if: steps.run_result.outputs.run_result != 'success' && inputs.pr != 0 && inputs.skip-tests != 'true'
         run: REACT_APP_ENVIRONMENT=${{steps.vars.outputs.REACT_APP_ENVIRONMENT}} yarn run test:unit
 
       # We burn React environment & the Segment analytics key into the build itself.
--- a/.github/workflows/on-demand-build-docker-image-deploy-preview.yml
+++ b/.github/workflows/on-demand-build-docker-image-deploy-preview.yml
@@ -17,35 +17,74 @@ jobs:
           body: |
             Deploying Your Preview: <${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}>.
             Workflow: `${{ github.workflow }}`.
+            skip-tests: `${{ github.event.client_payload.slash_command.args.named.skip-tests }}`.
+            env: `${{ github.event.client_payload.slash_command.args.named.env }}`.
             PR: ${{ github.event.client_payload.pull_request.number }}.
 
   server-build:
+    if: github.event.client_payload.slash_command.args.named.env != 'release'
     name: server-build
     uses: ./.github/workflows/server-build.yml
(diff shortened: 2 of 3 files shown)


.github/workflows/client-build.yml
```diff
@@ -9,6 +9,10 @@ on:
         description: "This is the PR number in case the workflow is being called in a pull request"
         required: false
         type: number
+      skip-tests:
```
mohanarpit, line 12: What's the default value for this variable? I'm asking because this workflow file is invoked in multiple places like "Test, Build, Docker image", "External integration tests", etc.

.github/workflows/server-build.yml
```diff
@@ -133,12 +137,23 @@ jobs:
...
           APPSMITH_IS_SELF_HOSTED: false
           APPSMITH_ENVFILE_PATH: /tmp/dummy.env
+          APPSMITH_SKIP_TESTS: ${{ inputs.skip-tests }}
         run: |
-          mvn --batch-mode versions:set \
-            -DnewVersion=${{ steps.vars.outputs.version }} \
-            -DgenerateBackupPoms=false \
-            -DprocessAllModules=true
-          ./build.sh
+          if [ $APPSMITH_SKIP_TESTS = "true" -o -z "$APPSMITH_SKIP_TESTS" ]
```
mohanarpit, line 142: Should `$APPSMITH_SKIP_TESTS = "true"` actually be `$APPSMITH_SKIP_TESTS == "true"`?

.github/workflows/on-demand-build-docker-image-deploy-preview.yml
```diff
@@ -17,24 +17,30 @@ jobs:
...
     name: client-build
     uses: ./.github/workflows/client-build.yml
     secrets: inherit
     with:
       pr: ${{ github.event.client_payload.pull_request.number }}
+      skip-tests: ${{ github.event.client_payload.slash_command.args.named.skip-tests }}
 
   rts-build:
     name: rts-build
+    if: github.event.client_payload.slash_command.args.named.env != 'release'
```
mohanarpit, line 43: For my understanding, why we are we not building or using rts if the environment specified is not release?

.github/workflows/server-build.yml
pratapaprasanna: Yea fixed it

.github/workflows/client-build.yml
pratapaprasanna: By default, it will be empty. only if it is set to true then skip-tests will work

.github/workflows/server-build.yml
```diff
@@ -133,12 +139,22 @@ jobs:
...
+              -DgenerateBackupPoms=false \
+              -DprocessAllModules=true
+            ./build.sh -DskipTests
+          else
+            mvn --batch-mode versions:set \
+              -DnewVersion=${{ steps.vars.outputs.version }} \
+              -DgenerateBackupPoms=false \
+              -DprocessAllModules=true
+            ./build.sh
+          fi
```
sharat87, line 157: Minor refactoring.

```sh
          args=
          if [[ "${{ inputs.skip-tests }}" == true ]]; then
            args=-DskipTests
          fi
          mvn --batch-mode versions:set \
            -DnewVersion=${{ steps.vars.outputs.version }} \
            -DgenerateBackupPoms=false \
            -DprocessAllModules=true
          ./build.sh $args
```

So we don't have to duplicate the command, and don't have to define the env var `APPSMITH_SKIP_TESTS`. 🙂

.github/workflows/on-demand-build-docker-image-deploy-preview.yml
```diff
@@ -17,35 +17,74 @@ jobs:
           body: |
             Deploying Your Preview: <${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}>.
             Workflow: `${{ github.workflow }}`.
+            skip-tests: `${{ github.event.client_payload.slash_command.args.named.skip-tests }}`.
+            env: `${{ github.event.client_payload.slash_command.args.named.env }}`
```
sharat87, line 21:
```suggestion
            env: `${{ github.event.client_payload.slash_command.args.named.env }}`.
```

.github/workflows/on-demand-build-docker-image-deploy-preview.yml
```diff
@@ -17,35 +17,74 @@ jobs:
...
+          cache-dependency-path: "app/client/yarn.lock"
+
+      - name: Pull Vercel Environment Information
+        run: vercel pull --yes --token=${{ secrets.VERCEL_TOKEN }}
+
+      - name: Build Project Artifacts
+        run: vercel build --yes --token=${{ secrets.VERCEL_TOKEN }}
+
+      - name: Deploy Project Artifacts to Vercel
+        run: vercel deploy --prebuilt --token=${{ secrets.VERCEL_TOKEN }}
```
sharat87, line 82: What DP URL does this deploy to? Asking because I don't see any configuration here. Will all DPs be deployed to the same URL? If not, can we get that URL and post a comment on the PR with the link?

sharat87 (commented):
Looks good. Suggested a minor refactoring.

PR #21086, opened 2023-03-01
task: side nav poc

### LINKED ISSUE #21064: [Task]: side nav POC

### SubTasks

- [x] Click action to open entity explorer to the side
- [x] Remove hover logic for entity explorer
- [x] Split Entity explorer, datasources and libraries
- [x] Move Settings pane CTA
- [x] Add documentation CTA (dummy)

eco-monk:
Fixes #21064

https://www.loom.com/share/1e5ada4f4ea241f68a3304a14e8d81ac

---
Also will work with both two and three pane layout.
https://www.loom.com/share/56f15a6db33348ff8e0151ca136e4957


app/client/src/selectors/multiPaneSelectors.ts
```diff
@@ -3,8 +3,8 @@ import { AppState } from "ce/reducers";
 export const getTabsPaneWidth = (state: AppState) =>
   state.ui.multiPaneConfig.tabsPaneWidth;
 
-export const isMultiPaneActive = (state: AppState) =>
-  state.ui.users.featureFlag.data.MULTIPLE_PANES === true;
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const isMultiPaneActive = (state: AppState) => true;
```
eco-monk, line 7:
```suggestion
export const isMultiPaneActive = (state: AppState) =>
  state.ui.users.featureFlag.data.MULTIPLE_PANES === true;
```

akash-codemonk:
Please do not merge. We will need the existing UI to work since we should be able to switch between multipane view and default view(as in how it works on release without multipane for public users). I can take it over from here.

PR #21094, opened 2023-03-01
test: Merge table.ts & tableV2.ts pages

Vijetha-Kaja:
## Description

- Merged Table.ts & TableV2.ts into single page Table.ts and refactored scripts accordingly
Flaky fixes below:
- Listv2_Button_Widget_spec.ts
- CreateNewApp_spec.js
- GuidedTour_spec.js

## How Has This Been Tested?
- Cypress test runs

### QA activity:


app/client/cypress/support/Pages/Table.ts
```diff
@@ -99,6 +124,17 @@ export class Table {
     // this.agHelper.Sleep(500);
   }
 
+  public WaitUntilTableLoadV2() {
```
Aishwarya-U-R, line 127: Can we make single method taking 2 different parameters? Like :
```
  public WaitUntilTableLoad(rowIndex = 0, colIndex = 0, tableVersion : 'v1' | 'v2' = 'assign a mostly used default value here') {
  if(tableVersion == ''v1)
  {
  ...v1 definition
  }
  else if(tableVersion == 'v2')
  {
  ... v2 definition
  }```

Comment (Vijetha-Kaja) [reply app/client/cypress/support/Pages/Table.ts:127]: yes, we can do that. But then we will need to modify many spec files with this value

Comment (Aishwarya-U-R) [reply app/client/cypress/support/Pages/Table.ts:127]: Yeah - I get its a painstaking job once - but after completed, Table helper will look more clean. We will not get confused with which method to use. Problem now is - people may not know another method for V2 exists while calling V1 methods. But adding it as parameter to one single method will do good.
For any other v2 versions like this in future - lets learn lesson from this activity & incorporate the same!

File: app/client/cypress/support/Pages/Table.ts [hunk 2]
```diff
@@ -157,73 +236,135 @@ export class Table {
...
+      if (serverSide == "On")
+        cy.get(this._pageNumberServerSidePagination).should(
+          "have.text",
+          Number(pageNo),
+        );
+      else {
+        cy.get(this._pageNumberClientSidePagination).should(
+          "have.value",
+          Number(pageNo),
+        );
```
Aishwarya-U-R, line 361: I see the current definition of table v2 for serverSide on & off is below:

```
public AssertPageNumber(pageNo: number, serverSide: "Off" | "On" = "On") {
    if (serverSide == "On")
      cy.get(this._pageNumber).should("have.text", Number(pageNo));
    else {
      cy.get(this._pageNumberServerSideOff).should(
        "have.value",
        Number(pageNo),
      );
      cy.get(this._previousPage).should("have.attr", "disabled");
      cy.get(this._nextPage).should("have.attr", "disabled");
    }
    if (pageNo == 1) cy.get(this._previousPage).should("have.attr", "disabled");
  }
```

In the above - seems both v1 & v2 having same denfinition isnt it?

app/client/cypress/support/Pages/Table.ts
```diff
@@ -39,20 +40,36 @@ export class Table {
     "//div[@class='thead']//div[@class='tr'][1]//div[@role='columnheader']//span[text()='" +
     columnName +
     "']/parent::div/parent::div/parent::div";
+  private _columnHeaderV2 = (columnName: string) =>
+    this._tableWrap +
+    "//div[@class='thead']//div[@class='tr'][1]//div[@role='columnheader']//div[text()='" +
+    columnName +
+    "']/parent::div/parent::div";
   private _nextPage = ".t--widget-tablewidget .t--table-widget-next-page";
+  private _nextPageV2 = ".t--widget-tablewidgetv2 .t--table-widget-next-page";
```
Aishwarya-U-R, line 49: All locators I see are different only by version v2 being added. Can we try something like below? - Single locator for both:

```
private _tableWidgetVersion = (version : "v1" | "v2") => `.t--widget-tablewidget(${version == "v1"? "" : version})`; 
//If v1, then empty string added else v2 added to tablewidget to selector, will also work if v3 sent in future
private _nextPage = (version : "v1" | "v2") => this._tableWidgetVersion(version) + " .t--table-widget-next-page";
```

Only if entire definition is different - we will have segregation in methods eg: AssertPageNumber. 
Else call the respective locator with version passed from caller with same definition eg WaitForTableEmpty, AssertTableLoaded etc

app/client/cypress/support/Pages/Table.ts
Aishwarya-U-R: Cross-check other methods also once.

Aishwarya-U-R (approved):

PR #406, opened 2020-08-23
Feature/explorer test fixes

NandanAnantharamu:
updated explorer test related to the widget with a text widget

PR #21183, opened 2023-03-06
fix: welcome page get started button

### LINKED ISSUE #13093: [Bug]-[40]:Disabled button can be clicked on the welcome page

### Description

Disabled button registers a click on the welcome page and hence user is navigated to the next page without needing to enter details

### Steps To Reproduce

1. Sign up to Appsmith
2. On the welcome page, click enter without entering any details and observe user navigated to next page even though button is disabled.

[LOOM DEMO](https://www.loom.com/share/86ed770d47a74f5599ff485365a49a7b)

### Public Sample App

_No response_

### Version

Cloud

eco-monk:
## Description

Fixes #13093

When disabled, `Get started` button in welcome page will not redirect on click

Media
https://www.loom.com/share/85755d7e71e64428871a1183d18595ba

## Type of change

## How Has This Been Tested?

- Manual

### Dev activity

### QA activity:


app/client/src/ce/sagas/userSagas.tsx
```diff
@@ -471,7 +471,7 @@ export function* logoutSaga(action: ReduxAction<{ redirectURL: string }>) {
       const currentUser: User | undefined = yield select(getCurrentUser);
       yield put(logoutUserSuccess(!!currentUser?.emptyInstance));
       localStorage.clear();
-      yield put(flushErrorsAndRedirect(redirectURL || AUTH_LOGIN_URL));
+      yield put(flushErrorsAndRedirect(redirectURL || AUTH_LOGIN_URL, true));
```
hetunandu, line 474: I always figured we will clear the history and not create a new tab

```suggestion
history.go(-history.length) // This will take it to the first page of the session
history.replace(redirectURL || AUTH_LOGIN_URL) // This will show the correct page
```

app/client/src/sagas/ErrorSagas.tsx
```diff
@@ -293,7 +293,12 @@ export function* flushErrorsAndRedirectSaga(
   }
   if (!action.payload.url) return;
 
-  history.push(action.payload.url);
+  if (action.payload.inNewTab) {
```
hetunandu, line 296: This is a bad idea and setting created just for this particular scenario. I would avoid it

hetunandu (changes requested):

app/client/src/ce/sagas/userSagas.tsx
eco-monk: I actually started with that. But it doesn't seem to help.
Even though it successfully goes back to the login url, the history still exists.

Also tried `history.go(-(history.length - 1))` to move out of appsmith, but the result is same. Since the tab is still open the history persists.

https://www.loom.com/share/06243c58335f4ccf90588b37f604ff45

eco-monk: Hence I thought we'll close the tab instead.

eco-monk: As an alternative,
Perhaps we can restrict going back from signin page?
Instead of looking to clear history.

eco-monk: Using alternate approach to prevent back button instead of opening a new tab

akash-codemonk: Is it possible to clear history from the browser programmatically? Also I didn't get what Id's are sensitive in the url that we have to clear. Is it page id, app id etc? Also other apps don't clear anything in history I think for e.g notion post signout it momentarily show the page name and id, either ways this can seen from the browser history.

eco-monk:
Note: need to account for redirections specific to EE repo (if any)

hetunandu (approved):

hetunandu (approved):

Richarex:
Verified this on DP url, working as expected. There are few issues with the button widget added in this bug id [#406](https://github.com/appsmithorg/design-system/issues/406)

eco-monk:
Discussion on why DP flow is different:
https://theappsmith.slack.com/archives/CGBPVEJ5C/p1678694313317649

PR #21204, opened 2023-03-06
Fix list widget height update issue.

prsidhu:
## Description

Issue: Increasing height of a list widget, when inside a form widget, causes the app to crash due to an infinite loop.
Fix: Update componentDidUpdate check to compare the new and old values before calling the update.

## Type of change

## How Has This Been Tested?

- Manual

### Dev activity
- [x ] My code follows the style guidelines of this project
- [ x] I have performed a self-review of my own code

### DIFF
--- a/app/client/src/widgets/ListWidgetV2/widget/index.tsx
+++ b/app/client/src/widgets/ListWidgetV2/widget/index.tsx
@@ -231,7 +231,10 @@ class ListWidget extends BaseWidget<
 
     this.pageSize = this.getPageSize();
 
-    if (this.shouldUpdatePageSize()) {
+    if (
+      this.shouldUpdatePageSize() &&
+      this.props.pageSize !== prevProps.pageSize
+    ) {
       this.updatePageSize();
       if (this.props.serverSidePagination && this.pageSize) {
         this.executeOnPageChange();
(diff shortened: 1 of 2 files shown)


app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -231,7 +231,10 @@ class ListWidget extends BaseWidget<
 
     this.pageSize = this.getPageSize();
 
-    if (this.shouldUpdatePageSize()) {
+    if (
+      this.shouldUpdatePageSize() &&
+      this.props.pageSize !== prevProps.pageSize
+    ) {
```
Tooluloope, line 237: This would stop the loop, but the update won't occur, i.e. `this.props.pageSize` would never get the updated value.

You could put a text widget and change the text to`List1.pageSize` to observe if the page size changes

prsidhu: shouldn't the page size update only if the value has changed?

Tooluloope: Here's the fix I proposed for this issue, although `updateWidgetProperty` shouldn't behave this way.

https://github.com/appsmithorg/appsmith/pull/21234/files#diff-4f8873018de0a2a6d1e475b019cbacc5b755186ed137dbd7f12974b758e4e850

PR #21238, opened 2023-03-07
chore: Add a fallback UUID for usage pulse calls

sbalaji1192:
## Description
Added a fallback value (crypto.randomUUID) when there is an issue with generating the key from segment

Fixes #https://github.com/appsmithorg/cloud-services/issues/582

Media

## Type of change

- Manual
- Jest
- Cypress

### Dev activity

### QA activity:


berzerkeer (approved):
nit: any reason to add a polyfill and use crypto to generate random uuid when we have `nanoId` package ?

sbalaji1192:
@berzerkeer Crypto randomUUID should suffice for all modern browsers. I was adding the polyfill as a fail-safe in older browsers. I didn't look into nano UUID, so as not to bloat the bundle.

berzerkeer:
> @berzerkeer Crypto randomUUID should suffice for all modern browsers. I was adding the polyfill as a fail-safe in older browsers. I didn't look into nano UUID, so as not to bloat the bundle.

Okay, asked because we already have nanoId as a dependency in our repo.

sbalaji1192:
Oh. then in that case lets use nanoID. didn't know that.

berzerkeer (dismissed):

app/client/src/ce/sagas/userSagas.tsx
```diff
@@ -147,6 +148,18 @@ export function* initiateUsageTracking(payload: {
     } else {
       yield initializeSegmentWithoutTracking();
```
sbalaji1192, line 149: 1. remove this line and corresponding functions.
2. when telemetry is enabled, we need to fetch the anonymous in every sendPulse of usage pulse file.

PR #21317, opened 2023-03-10
fix: Selected Widget Visibility

### LINKED ISSUE #1282: [Bug]-[500]:Entity explorer does not navigate to widget on other tab
Steps to reproduce:
1) Add a tab widget 
2) Add Multiple tabs 
3) Add a dropdown widget on Tab1 
4) Copy the Dropdown widget 
5) Navigate to Tab2 and paste it without selection of the area 
and observe 

Observation:
Observation 1:
The dropdown widgets gets placed on Tab1 even when the user is on Tab2 but this is visible on the entity explorer

Observation 2:
Now clicking on the entity explorer for dropdown's it is observed that the user is not navigate to respective tabs for display 

Expectation:
Expectation 1:
User must be navigated to Tab where the widget is placed when the user is on another tab on click from the entity explorer 

Expectation 2:
On Copy the widget must be placed on the current tab where the user is 

Video link:
[LOOM DEMO](https://www.loom.com/share/4f073d08002f4c96a4fbbefba5252785)

### LINKED ISSUE #18173: [Task]: Make widget selection happen if the URL changes
## Description

Widget selection is a complicated feature. Depending on how and where a selection happens, the system handles and controls differently. There is also a hierarchy in widgets with containers, hidden via UI states with modals and tabs, and ability to select multiple widgets at once. 

Overtime we have added a lot of these features in a way that there is not a single way for the platform to select a widget. There are also specialised methods for different places that facilitate widget selection. Tabs widget is has an old issue regarding this 

We need to achieve the following:

- Have a streamlined way to select widget(s)
- Make widget selection work with browser back/forward buttons
- Have a single way to select/focus on any entity of Appsmith

### Solution
Refactor the widget selection to make it easier to select a widget. Make it as easy as it is to navigate to an action/datasource. Make selection happen via a url navigation so that browser navigation can work with platform navigation.

### SubTasks

- [x] #19570
- [x] #19571
- [x] #1282

hetunandu:
### Part 3 of selected widget refactor

As part of context switching and selected widget refactor, we saw that widgets that are inside modals or tabs and are hidden cannot be switched to without updating some meta properties. The meta properties are actually owned by the end user and the developer user would create some default values for it as well. This becomes a problem soon when the platform also tries to update it. So as part of this refactor, we will use the selected widget ancestry (the chain of widgets from the top to the currently selected widget) to handle if widgets need to be visible or not. It will also indicate the widgets in the path of selection to "make way" for the selected widget to be seen.

Media

https://user-images.githubusercontent.com/12022471/224916943-b10e8694-0c95-4157-bb93-288d7c0bf60a.mov

- This works on any number of levels of hirarchy
- The logic is supposed to handled by each widget that can potentially hide other widgets inside it
- Improves some platform perf as the handing so widget meta is not done by the platform anymore

Affected widgets:
- Modal Widget
- Tabs Widget

> tl;dr: Update the platform's way to show widgets that can be hidden. Makes sure a selected widget is always shown.

Fixes #1282
Resolves #18173

## Type of change

## How Has This Been Tested?

- Manual
- Cypress

### Dev activity

### QA activity:


app/client/src/widgets/TabsWidget/widget/index.tsx
```diff
@@ -387,6 +394,17 @@ class TabsWidget extends BaseWidget<
     return WidgetFactory.createWidget(childWidgetData, this.props.renderMode);
   };
 
+  private getSelectedTabWidgetId() {
+    let returnValue = this.props.selectedTabWidgetId;
+    if (this.props.children) {
+      returnValue =
+        this.props.children.find((tab) =>
+          this.props.selectedWidgetAncestry.includes(tab.widgetId),
+        )?.widgetId ?? this.props.selectedTabWidgetId;
```
ashit-rath, line 403: Not 100% sure but this may cause some problems in certain widget operation and auto height.

Here we are trying to derive the selectedTabWidgetId from either the meta prop `selectedTabWidgetId` or the `ancestry`. While this is fine in the context of Tabs widget rendering, there are other places where certain operations depend on the meta value `selectedTabWidgetId`.

We should set the `selectedTabWidgetId` meta value probably onMount.

app/client/src/sagas/ModalSagas.ts
ashit-rath, line 1: Now when the `isVisible` property is not a factor for the modal's open/close state. Are we suppose to update the logic of `closeModalSaga` as well?

ashit-rath: We should also look at 
`generateWidgetProps` in `WidgetPropsUtils.tsx`
default value of isVisible in `src/widgets/ModalWidget/index.ts`
assignment of widgetProps.isVisible in `withWIdgetProps`. We so not need to check for MODAL WIDGET.

app/client/src/widgets/TabsWidget/widget/index.tsx
hetunandu: Can you point out how I can check this?

The main goal of this change it to do it without updating the meta value of the tabs widget since it can be controlled by the developer user / end user

app/client/src/widgets/TabsWidget/widget/index.tsx
```diff
@@ -444,4 +462,8 @@ class TabsWidget extends BaseWidget<
   }
 }
 
-export default TabsWidget;
+const mapDispatchToProps = (state: AppState) => ({
```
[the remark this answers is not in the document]
hetunandu: Good point! I totally missed the withWidgetProps ability to pass in props. Updated the code

app/client/src/widgets/withWidgetProps.tsx
```diff
@@ -61,6 +62,7 @@ function withWidgetProps(WrappedWidget: typeof BaseWidget) {
       getWidget(state, widgetId),
     );
     const metaWidget = useSelector(getMetaWidget(widgetId));
+    const selectedWidgetAncestry = useSelector(getSelectedWidgetAncestry);
```
ashit-rath, line 65: I think this may cause perf regression. Since all widgets have now subscribed to this selector, any time we click a widget; all the other widgets will receive this update even if it may not be relevant to that widget.

I think if we abstract the logic of the use of `selectedWidgetAncestry` in the `withWidgetProps` by doing something like 
```
// get only the bool value
const isWidgetPresentInSelectedAncestry = useAncestryHook()
```
We can save un-necessary re-renders.

That being said, we also need to pass `selectedWidgetAncestry` as prop to widget but not all widgets require this. So we can do a similar treatment `childWidgets` or `flattenedChildCanvasWidgets`

app/client/src/widgets/TabsWidget/widget/index.tsx
riodeuno: @ashit-rath Yes, we're relying on the `selectedTabWidgetId` for computations in auto height. The reason behind is that we need to make sure that we're computing the height of the tabs widget based on the currently selected tab. Here is a function which does this: https://github.com/appsmithorg/appsmith/blob/bc431d01629995b9f5f667da91f514b036f4e525/app/client/src/sagas/autoHeightSagas/helpers.ts#L36

If meta property is not the way to get this, we can always update this function to use the logic you've described. 
We could potentially add this logic into the widget configuration. We don't have the option of getting any "derived props" which have nothing to do with evaluations for widgets. The idea would be that we add a widget configuration that is a function to compute this. The function can probably be run in the `withWidgetProps` and the same configuration function can be used in the `helper.ts` I've linked to in this comment.

An example of this configurations could be something like this: https://github.com/appsmithorg/appsmith/blob/bc431d01629995b9f5f667da91f514b036f4e525/app/client/src/widgets/TabsWidget/index.ts#L22

riodeuno: Also, we have this being used in the paste workflow as well. However, this is a strange case. Any "selectable" widget's parent cannot be anything other than a canvas widget. So, this code flow doesn't seem to be doing what I think it is doing. https://github.com/appsmithorg/appsmith/blob/bc431d01629995b9f5f667da91f514b036f4e525/app/client/src/sagas/WidgetOperationUtils.ts#L459
If I am right, then it is possible that we can safely remove this reference. We just need to make sure of this.

app/client/src/widgets/withWidgetProps.tsx
hetunandu: Updated code to make this only work for widgets with child widgets

ashit-rath (approved):

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -645,14 +645,11 @@ abstract class BaseWidget<
       // return this.getCanvasView();
       case RenderModes.PAGE:
         content = this.getWidgetComponent();
-        if (this.props.isVisible) {
```
marks0351, line 648: @riodeuno isn't this change needed for hiding invisible widgets?

app/client/src/widgets/withWidgetProps.tsx
```diff
@@ -101,6 +102,11 @@ function withWidgetProps(WrappedWidget: typeof BaseWidget) {
       }
     }, equal);
 
+    const selectedWidgetAncestry: string[] = useSelector((state: AppState) => {
+      if (!WIDGETS_REQUIRING_SELECTED_ANCESTRY.includes(type)) return [];
```
marks0351, line 106: please add {} for readability

marks0351 (changes requested):

app/client/src/widgets/BaseWidget.tsx
hetunandu: This is already handled in withWidgetProps and hence not needed here

marks0351 (approved):

PR #21346, opened 2023-03-12
chore: add pricing url and customer portal url to env var

berzerkeer:
## Description

 - The customer portal URL and pricing page URL should come from env vars. So that release can point to release customer portal and prod can use prod customer portal URL.

Fixes [#487](https://github.com/appsmithorg/cloud-services/issues/487)

## Type of change

- Chore

## How Has This Been Tested?

- Manual

### Dev activity

### QA activity:


app/client/src/ce/configs/index.ts
```diff
@@ -131,6 +133,8 @@ export const getConfigsFromEnvVars = (): INJECTED_CONFIGS => {
       .APPSMITH_DISABLE_IFRAME_WIDGET_SANDBOX
       ? process.env.APPSMITH_DISABLE_IFRAME_WIDGET_SANDBOX.length > 0
       : false,
+    pricingUrl: process.env.REACT_APP_PRICING_URL || "",
```
dipyamanbiswas07, line 136: do we need "" as fallback? might result in some breakages

berzerkeer: Here, we check for the pricing url within the node process. i.e the react client `env` file or env variable.  So if this results in `""` then in the `getAppsmithConfigs` function, we check `pricingUrl: ENV_CONFIG.pricingUrl || APPSMITH_FEATURE_CONFIGS.pricingUrl` where ENV_CONFIG.pricingUrl is the result of the above , so if its `""` the fallback is right from the window object `APPSMITH_FEATURE_CONFIGS` which has a fallback value of PROD urls.

dipyamanbiswas07 (approved):

Issue #21298, opened 2023-03-09
[Bug]: Limit Gsheet Access - selected files are not being seen in the query dropdown


sneha122:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

In case of specific google sheets, file selected while creation of datasource, are not being seen in the gsheet query dropdown

### Steps To Reproduce

1. Create Gsheet Datasource with specific sheets modality
2. Complete oauth2 flow
3. Select files from file picker
4. Create query for gsheet datasource
5. observe the spreadsheet dropdown, selected file is not there

### Environment

Release

### Version

Cloud - v1.9.11

PR #21361, opened 2023-03-13
fix: fetch gsheet project id from cs added

sneha122:
## Description

File picker implementation for Limiting Gsheet Access requires google sheet project id. The changes in this PR gets the gsheet project id from cloud-services and returns it back to client. Client then uses this project id to open file picker and select required files.

Fixes https://github.com/appsmithorg/appsmith/issues/21298, https://github.com/appsmithorg/appsmith/issues/21362

Media

## Type of change

## How Has This Been Tested?
- Manual

### Dev activity

### QA activity:


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/AuthenticationResponse.java
```diff
@@ -28,4 +28,6 @@ public class AuthenticationResponse implements AppsmithDomain {
 
     @Encrypted
     Object tokenResponse;
+
+    String projectID;
```
sumitsum, line 32: Is `projectID` something that we want to store as encrypted on the DB ? If so, we should use @Encrypted annotation for it.

sumitsum (changes requested):

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/AuthenticationResponse.java
sneha122: there is no need to store projectID on DB as we are fetching it on demand and its the same for all datasources, so should I add Transient annotation instead so we do not save it in DB at all? Also in that case @encrypted is required?

sumitsum: Agree, let us mark it as @Transient

sumitsum: Also, can you please mention a comment here saying that this field is not something that is returned as a response from the authorization server but it is something that the cloud service is providing.

sumitsum (changes requested):

sneha122:
`Also, should we rename it to projectIDPrefix as I am assuming it is not the full project id but only the prefix part of it ?` - Project ID is the prefix part of client Id, essentially it already is a prefix @sumitsum

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/AuthenticationResponse.java
sneha122: ProjectID is prefix part of clientId so it already is a prefix.

sumitsum:
Server side changes LGTM.

Irongade (approved):
Generally, I'd like to encourage you to add more comments in your code. Especially for new features, this will help you and anyone else looking into this in the future.

sneha122:
> Generally, I'd like to encourage you to add more comments in your code. Especially for new features, this will help you and anyone else looking into this in the future.

Thanks @Irongade for the feedback!! Usually I add comments in the hard to understand areas, but is there any guidelines that you follow for comments?

sumitsum (dismissed):

sneha122:
The code review is done for this but keeping it in code review pipeline because in order to move this to needs QA, cs PR needs to be merged, and code review for that is pending as of now

Irongade (dismissed):

sumitsum (approved):

sneha122:
This PR cannot be tested on DP, as DP generated with `build-deploy-preview` does not have cloud-services configured correctly. Hence as per this [thread](https://theappsmith.slack.com/archives/C04Q58L86RW/p1679318148732489), we have decided to merge the PR and test the changes on release. CC: @prapullac @rohan-arthur

prapullac:
As per above comment adding the label.

PR #21442, opened 2023-03-15
ci: Add lint check to the build step

SatishGandham:
Add a step to verify lint rules in client build workflow.


riodeuno:
@SatishGandham Does this have any dependency on husky? I'm asking because today, `husky` is installed as a dependency and was planning on creating a PR to move this to the dev dependencies as well as upgrade the package.

SatishGandham:
@riodeuno , I don't think so. Any way, I will test and check.

.github/workflows/client-build.yml
```diff
@@ -100,6 +100,19 @@ jobs:
         if: steps.run_result.outputs.run_result != 'success'
         run: yarn install --frozen-lockfile
 
+      - name: Install Eslint
+        if: steps.run_result.outputs.run_result != 'success'
+        run: yarn global add eslint@latest
+      
+      - name: Run linters
+        uses: wearerequired/lint-action@v2
```
KelvinOm, line 108: Do we really need to use `wearerequired/lint-action@v2`?  

I mean, usually such things are done through separate scripts in the `package.json`. Something like:
```
    "lint:ci": "eslint --ext .ts,.tsx ./src && eslint --ext .js ./cypress",
    "prettier:ci": "prettier --check ./src && prettier --check ./cypress",
```

Then we can run them like this `yarn run lint:ci` and `yarn run prettier:ci`

1. This approach will allow developers to fully control the invoked scripts without changing the yml config.
2. Developers will understand which scripts are run in CI and will be able to easily reproduce their results locally to fix errors.

.github/workflows/client-build.yml
```diff
@@ -100,6 +100,19 @@ jobs:
         if: steps.run_result.outputs.run_result != 'success'
         run: yarn install --frozen-lockfile
```
KelvinOm, line 101: Let's consider the possibility of splitting the build and lint/prettier into different stages and running them in parallel. This way we can't slow down the build and it will be cleared exactly where the problem occurred.

.github/workflows/client-build.yml
SatishGandham: Good point. I really didn't know how to do it, picked the first result!!
Made the necessary changes.

.github/workflows/client-build.yml
SatishGandham: Our client build is faster than server build. And linting adds 2-3 minutes to the workflow.
I don't think we need to separate it out. I'm assuming failure here is not a regular occurrence, if it is separating would save from building the server unnecessarily.

SatishGandham:
@riodeuno , we don't depend on husky.

.github/workflows/client-build.yml
```diff
@@ -100,6 +100,19 @@ jobs:
...
 
+      - name: Install Eslint & Prettier
+        if: steps.run_result.outputs.run_result != 'success'
+        run: yarn global add eslint@latest prettier@latest
```
KelvinOm, line 105: Why do we install packages globally? On line 101 we are installing packages and everything we need is already there.

.github/workflows/client-build.yml
```diff
@@ -100,6 +100,19 @@ jobs:
...
+      
+      - name: Run Eslint
+        if: steps.run_result.outputs.run_result != 'success'
+        run: eslint --ext .ts,.tsx ./src && eslint --ext .js ./cypress
```
KelvinOm, line 109: Why can't we run scripts from package.json directly? Does this `yarn run lint:ci` not work here?

.github/workflows/client-build.yml
KelvinOm: Good point. We can return to this if necessary.

app/client/package.json
```diff
@@ -10,6 +10,28 @@
     "packages/*"
   ],
   "cracoConfig": "craco.dev.config.js",
+  "scripts": {
+    "analyze": "yarn cra-bundle-analyzer",
+    "start": "BROWSER=none EXTEND_ESLINT=true REACT_APP_ENVIRONMENT=DEVELOPMENT REACT_APP_CLIENT_LOG_LEVEL=debug HOST=dev.appsmith.com craco start",
+    "build": "./build.sh",
```
KelvinOm, line 16: I would like to add type-checking before building 
```suggestion
    "build": "yarn run check-types && ./build.sh",
```

I think we can add a `type-check` call right before the build, because if there is something wrong with the types, then there is no point in doing the build. Also in this way we can remove [this code](https://github.com/appsmithorg/appsmith/pull/21442/files#diff-7600caa85f1c89ff1fcd0b4b73e379816ef5a59cd7e61fdb90feb496e704d16aR111-R113).

KelvinOm:
I suppose this [PR](https://github.com/appsmithorg/appsmith/pull/21586) should be here because without fixes we can't merge this one.

If we are not going to fix JSON files in cypress, then in addition to [this line](https://github.com/appsmithorg/appsmith/pull/21586/files#diff-1a24c392dd3c34edf67a9d69969395006fe6ae294fb8fa4d5ba3bf4ba07f954fR6) we should also add
```
cypress/fixtures
cypress/locators/
```

app/client/package.json
```diff
@@ -10,6 +10,28 @@
...
+    "cytest": "REACT_APP_TESTING=TESTING REACT_APP_ENVIRONMENT=DEVELOPMENT craco start & ./node_modules/.bin/cypress open",
+    "test:unit": "$(npm bin)/jest -b --colors --no-cache --silent --coverage --collectCoverage=true --coverageDirectory='../../' --coverageReporters='json-summary'",
+    "test:jest": "$(npm bin)/jest --watch",
+    "generate:widget": "plop --plopfile generators/index.js",
+    "postinstall": "patch-package && CURRENT_SCOPE=client node ../shared/install-dependencies.js",
+    "preinstall": "CURRENT_SCOPE=client node ../shared/build-shared-dep.js",
+    "install": "node cypress/apply-patches.js",
+    "storybook": "yarn workspace @design-system/storybook storybook",
+    "lint:ci": "eslint --ext .ts,.tsx ./src && eslint --ext .js ./cypress",
+    "prettier:ci": "prettier --check ./src",
```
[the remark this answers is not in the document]
SatishGandham: As discussed, there are lot of JSON files in cypress which don't pass this check. We will add it after QA team makes those changes.

KelvinOm:
Let's change [this file](https://github.com/appsmithorg/appsmith/blob/release/app/client/cypress/.eslintrc.json) (app/client/cypress/.eslintrc.json) like this:
```
{
  "extends": [
    "../.eslintrc.json"
  ],
  "env": {
    "cypress/globals": true
  },
  "rules": {
    "cypress/no-unnecessary-waiting": 0
  }
}
```

If we do not do this, we will continue to have [such discussions](https://theappsmith.slack.com/archives/CPG2ZTXEY/p1678804912000339).

P.S. `plugin:cypress/recommended` - already present in the root config

app/client/package.json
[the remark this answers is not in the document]
KelvinOm: Let's add them to ignore as I described in [this message](https://github.com/appsmithorg/appsmith/pull/21442#issuecomment-1476037317). Otherwise, we need to remove the checks in cypress folder on pre-commit hook.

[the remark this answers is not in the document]
SatishGandham: Done.

KelvinOm (approved):
Please check before merging that all the rules work in CI, and builds really fall.

app/client/package.json
SatishGandham: Build doesn't happen if we the previous three checks we added fail. By adding it as a part of build, we are unnecessarily running jest tests.

PR #21543, opened 2023-03-17
fix: corrections for eslint and prettier

KelvinOm:
Correction of eslint and printer errors/warnings. It is associated with [this activity](https://theappsmith.slack.com/archives/CGBPVEJ5C/p1678968535728579) in order to avoid conflicts with others.

  1. Fix eslint warnings in unit tests. All unit tests are passed.
  2. Fix prettier at cypress.


coverage-summary.json
KelvinOm, line 1: This file is in [gitignore](https://github.com/appsmithorg/appsmith/blob/release/.gitignore#L13), but it continues to track since it was added to the repo, so after each local test run, this file has to be deleted from git diff.

app/client/src/actions/recentEntityActions.ts
tanvibhakta, line 1: Of course this file needed to be deleted, not just the import. Got it :)

coverage-summary.json
tanvibhakta: Iirc you can prevent this file from being generated by adding `--coverage=false` to the command

tanvibhakta (approved):
LGTM!

coverage-summary.json
KelvinOm: In any case, this file should not be here.

PR #21578, opened 2023-03-20
ci:use same PVC for dps rather than a cleanup

pratapaprasanna:


scripts/deploy_preview.sh
```diff
@@ -45,15 +45,12 @@ kubectl config set-context --current --namespace=default
...
-
-echo "Use kubernetes secret to Pull Image"
-kubectl create ns $NAMESPACE
+echo "Create namespace and secret to Pull Image"
+kubectl create ns $NAMESPACE || echo "true"
 kubectl create secret docker-registry $SECRET \
   --docker-server=https://index.docker.io/v1/ \
   --docker-username=$DOCKER_HUB_USERNAME \
-  --docker-password=$DOCKER_HUB_ACCESS_TOKEN -n $NAMESPACE
+  --docker-password=$DOCKER_HUB_ACCESS_TOKEN -n $NAMESPACE || echo "true"
```
sharat87, line 53: The idea is that the `create ns` will fail if the namespace already exists, and the `create secret` will also fail similarly, right? If yes, do you think something like this is better?

```
if kubectl create ns "$NAMESPACE"; then
  kubectl create secret docker-registry "$SECRET" \
    --docker-server=https://index.docker.io/v1/ \
    --docker-username="$DOCKER_HUB_USERNAME" \
    --docker-password="$DOCKER_HUB_ACCESS_TOKEN" \
    -n "$NAMESPACE"
fi
```

Reason I'm recommending this is that the `|| true` escape hatch stops _all_ errors from throwing up. This way, if the `create secret` fails because of whatever reason, the script will stop and fail, instead of proceeding.

scripts/deploy_preview.sh
```diff
@@ -45,15 +45,14 @@ kubectl config set-context --current --namespace=default
...
-  --docker-server=https://index.docker.io/v1/ \
-  --docker-username=$DOCKER_HUB_USERNAME \
-  --docker-password=$DOCKER_HUB_ACCESS_TOKEN -n $NAMESPACE
+echo "Create namespace and secret to Pull Image"
+if kubectl create ns "$NAMESPACE";
+then
+   kubectl create secret docker-registry "$SECRET" \
+   --docker-server=https://index.docker.io/v1/ \
+   --docker-username="$DOCKER_HUB_USERNAME" \
+   --docker-password="$DOCKER_HUB_ACCESS_TOKEN" \ -n "$NAMESPACE"
```
sharat87, line 54: Syntax error. The backslash should be followed by a newline, or just not have a backslash at all. 🙂

Like, use this:
```
  --docker-password="$DOCKER_HUB_ACCESS_TOKEN" -n "$NAMESPACE"
```

Or this:
```
  --docker-password="$DOCKER_HUB_ACCESS_TOKEN" \
  -n "$NAMESPACE"
```

sharat87 (changes requested):

PR #21602, opened 2023-03-20
perf: code-split codemirror & lottie + optimize the build for modern apps

iamakulov:
This PR might be easier to review commit-by-commit.

## Description

This PR
- switches the build target to modern browsers; this means stuff like `class`, `async`/`await`, etc is not transpiled anymore
- code-splits away the code editor + sets up ESLint rules to enforce that
- code-splits away Lottie + sets up ESLint rules to enforce that

In total, this saves 1.34 minified MBs (8.34 → 7 MB), getting rid of these modules (+ making others smaller, thanks to the modern build target):

CleanShot 2023-03-20 at 13 45 02@2x

## Type of change

- Manual:
  - testing the code editor (including by emulating slow connections);
  - testing the lottie player (confetti animation + welcome tour)

### Dev activity

### QA activity:


app/client/package.json
```diff
@@ -187,10 +188,7 @@
     "storybook": "yarn workspace @design-system/storybook storybook"
   },
   "browserslist": [
-    ">0.2%",
-    "not dead",
-    "not ie <= 11",
-    "not op_mini all"
+    "defaults and supports es6-module"
```
iamakulov, line 191: Why the switch?

If you take the old Browserslist query (`>0.2%, not dead, not ie <= 11, not op_mini all`) and paste it into the Babel REPL, you’ll see that this query makes classes and async functions [transpile down to ES5](https://babeljs.io/repl#?browsers=%3E0.2%25%2C%20not%20dead%2C%20not%20ie%20%3C%3D%2011%2C%20not%20op_mini%20all&build=&builtIns=false&corejs=false&spec=false&loose=false&code_lz=IYZwngdgxgBAZgV2gFwJYHsL3egFAShgG8YBfAKHKgBtQQYAxHY8mASBnNKA&debug=false&forceAllTransforms=false&modules=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env&prettier=false&targets=&version=7.21.3&externalPlugins=&assumptions=%7B%7D):

Screenshot 2023-03-20 at 14 50 29

The reason for this can be seen [at browsersl.ist](https://browsersl.ist/#q=%3E0.2%25%2C+not+dead%2C+not+ie+%3C%3D+11%2C+not+op_mini+all). This query matches Android 4, which has a 0.3% browser share and does not support any of the modern JS features:

CleanShot 2023-03-20 at 14 55 19

To solve this, instead of adding yet another exception, I decided to switch to `defaults and supports es6-module`. `defaults` means `> 0.5%, last 2 versions, Firefox ESR, not dead`, whereas `supports es6-module` excludes weird ancient browsers like Opera Mini or KaiOS 2.5.

This switch keeps all the stuff like `async/await` as-is and saves 0.67 minified MBs in the main bundle alone.

app/client/src/components/editorComponents/LazyCodeEditor/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+  // The state when the CodeEditor chunk is loading, but the user has interacted with the placeholder
+  | "loading-interacted"
+  // The state when we’re waiting for the idle callback to fire to render CodeEditor
+  | "waiting-idle-callback"
+  // The state when the CodeEditor must be rendered
+  | "active"
+  // The state when the CodeEditor must be rendered and made focused
+  | "active-focused"
+  // The state when the CodeEditor chunk failed to load
+  | "error";
```
iamakulov, line 27: To code-split the code editor, I re-used the existing `LazyCodeEditorWrapper` component.

The component already had logic to
- rendering the code editor with `requestIdleCallback()`
- and to speed up that render when the user interacts with the placeholder

With the addition of lazy-loading, the logic got really complex, so I decided to implement it as a barebones state machine. The comments should (hopefully) make it clear what’s going on, but just for the reference, here’s a full diagram of all states: https://stately.ai/registry/editor/b6ae19f0-6adb-47f7-a00b-dac5cf9a61bd?machineId=8e0d268e-57b2-4fe5-8f65-cb363a57f141

app/client/src/components/editorComponents/LazyCodeEditor/index.tsx
```diff
@@ -0,0 +1,236 @@
+import React, { useState, useEffect, useRef, useCallback } from "react";
```
KelvinOm, line 1: @hetunandu Could you review this a code?

app/client/.eslintrc.js
```diff
@@ -0,0 +1,179 @@
+/** @type {import('eslint').Linter.Config} */
```
KelvinOm, line 1: Why is this necessary?

app/client/.eslintrc.js
```diff
@@ -0,0 +1,179 @@
...
+            message:
+              "Reason: Please don’t import Remix icons statically. (They won’t always be needed, but they *will* always be present in the bundle and will increase the bundle size.) Instead, please use the importRemixIcon wrapper from design-system-old (e.g. const StarIcon = importRemixIcon(() => import('remixicon-react/Star'))).",
+          },
+        ],
+      },
+    ],
+    // Annoyingly, the `no-restricted-imports` rule doesn’t allow to restrict imports of
+    // `editorComponents/CodeEditor` but not `editorComponents/CodeEditor/*`: https://stackoverflow.com/q/64995811/1192426
+    // So we’re using `no-restricted-syntax` instead.
+    "no-restricted-syntax": [
```
KelvinOm, line 99: We inherit this config [here](https://github.com/appsmithorg/appsmith/blob/release/app/client/packages/storybook/.eslintrc.json#L3) and [here](https://github.com/appsmithorg/appsmith/blob/release/app/client/cypress/.eslintrc.json#L2), and in the future we will also use the general rules for the rest of the packages. So I have to ask do we really need these rules for other dependencies?

I would suggest separating the general rules and inheriting them here and in packages.

app/client/src/components/editorComponents/LazyCodeEditor/index.tsx
hetunandu: Looks good to me

app/client/src/Globals.d.ts
```diff
@@ -1 +1,6 @@
 declare module "*.module.css";
+
+declare module "*.txt" {
```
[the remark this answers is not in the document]
iamakulov: For this: https://github.com/appsmithorg/appsmith/pull/21602/files/0b750838576786578dd2b095acab183cb3f60e65#diff-4463e6b504253024bc05a7214c25934113b5c13ef05e2e6020cdb463c9574d76

This is needed to give import like

https://github.com/appsmithorg/appsmith/blob/2cf0a35e7b358026deb90fd87a5a0d256504f93e/app/client/src/utils/helpers.tsx#L7

correct types.

app/client/.eslintrc.js
iamakulov: In VSCode (and maybe other IDEs?), this makes the IDE provide type-based autocomplete for the ESLint config, which is much better.

With `/** @type **/`:

Screenshot 2023-03-22 at 11 00 04

Without `/** @type **/`:

Screenshot 2023-03-22 at 11 00 24

app/client/.eslintrc.js
iamakulov: How about this? https://github.com/appsmithorg/appsmith/pull/21602/commits/0f463db1e912121d33e609783c4fd7d24aafa34b

This splits out the base ESLint config (`src/.eslintrc.base.json`) and makes the following packages use it:
- `client`
- `client/cypress`
- `client/packages/storybook`
- `shared/ast` (🆕)

This definitely creates a bunch of new ESLint errors (all or almost all of them formatting-based), but I’d rather resolve them when we’re merging this upstream, since https://github.com/appsmithorg/appsmith/pull/21442 also tuned the ESLint configs somewhat.

KelvinOm: Yes, it's better that way. I think `shared/ast` can be skipped for now. I would like to move it to packages.

As for formatting, please pay attention to https://github.com/appsmithorg/appsmith/pull/21679.

KelvinOm (approved):

app/client/.eslintrc.js
iamakulov: (I’ll keep the ESLint changes for `shared/ast` for now. I added them for consistency [to make every project with `.eslintrc.json` inherit a shared config], and, as I understand, they won’t complicate moving `shared/ast` to `packages`.

But let me know if I misunderstood, and the inheritance there is actively harmful.)

KelvinOm: Thanks! The config will not complicate the movement. )

app/client/src/components/editorComponents/LazyCodeEditor/index.tsx
```diff
@@ -0,0 +1,236 @@
...
+   * and (possibly) other conditions. It’s intended to not have any side effects –
+   * all side effects should go into this.handleStateChange() */
+  transition(event: LazyCodeEditorEvent) {
+    switch (this.state) {
+      case "idle":
+        if (event === "RENDERED") {
+          if (CachedCodeEditor) {
+            this.state = "waiting-idle-callback";
+          } else {
+            this.state = "loading";
```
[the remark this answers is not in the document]
iamakulov: Here!

https://github.com/appsmithorg/appsmith/blob/4504a62005ff395a2a09394cf28efa51502c43c0/app/client/src/components/editorComponents/LazyCodeEditor/index.tsx#L186-L188

Note that I intentionally trigger the event from `useEffect`. It’s not the fastest approach (we’re delaying loading until `useEffect()` fires), but running side effects from `render()` is not 100% safe, especially with concurrent React 18 features.

(Although I just tried, and I can’t come up with a case where that will actually lead to bugs in practice. But better safe than sorry I guess.)

PR #21621, opened 2023-03-21
chore: Check for AVX instruction earlier than the first mongod run

sharat87:
We're already checking for the AVX instruction, and producing a helpful message about it. But, we're doing that check _after_ we invoke the `mongod` process, which fails on such processes, and the error message is never shown.

This PR changes this, by doing the check before that `mongod` command is run.


sharat87:
Hey @sum35h, can you review/approve please? 🙂

sum35h (approved):

PR #21699, opened 2023-03-22
feat: Added file id mapping in datasource config

### LINKED ISSUE #21074: [Feature]: Gsheets: Store File Ids in datasource configuration

### Summary

We need to save file Ids of the files selected from file picker in the datasource configuration, as file picker aggregates all files selected in all datasources/workspaces/applications as per https://github.com/appsmithorg/appsmith/issues/20835. 

This will help us in showing users only those files that they selected from file picker for any particular datasource, instead of showing all files they have selected across all datasources.

sneha122:
## Description

This PR adds:
- File Id mapping of the google sheets selected by user, in datasource configuration, so that when creating queries on top of such gsheet datasource, only the selected spreadsheets can be seen in the spreadsheet dropdown.

Changes done on client side:
- As soon as user selects file in file picker popup, the callback will get the file ids and update the datasource to contain file ids as a part of datasource configuration properties.
- If user cancels the file selection, file ids is sent as empty array and datasource is updated.

Changes done on server side:
- In `GoogleSheetPlugin.java` where we have defined execute and trigger methods for gsheet query, here I have added a new variable allowedFileIds, which gets the list of authorised file ids from datasource configuration object and the same list is passed to functions like `transformTriggerResponse` and `transformExecutionResponse`, which returns  file list data based on the allowedFileIds. In FileListMethod class, these methods contain the logic to send only authorised file data.
- Since these two methods are a part of triggerMethod and executionMethod interfaces, all gsheet query operation classes that extend this method, their function definition needed to be updated with this third allowedFileIds parameter.
- Similarly all gsheet query operations test classes were using these two methods, and hence this third parameter needed to be added there as well.

How to test:
- With this improvement, when we select `file1` and `file2` for one datasource and `file3` and `file4` for another datasource, In the query dropdown for first ds, we should only see `file1` and `file2`, for second datasource, we should only see `file3` and `file4`.
- Please check following gsheet operations:
    - Fetch Many
    - Fetch Details
    - Update One
    - Update Many
    - Insert Many

Fixes #21074

Media

## Type of change

## How Has This Been Tested?

- Manual
- JUnit

### Dev activity

### QA activity:


app/client/src/sagas/DatasourcesSagas.ts
```diff
@@ -1180,18 +1185,38 @@ function* filePickerActionCallbackSaga(
...
 
+    // Once users selects/cancels the file selection,
+    // Sending sheet ids selected as part of datasource
+    // config properties in order to save it in database
+    set(
+      datasource,
+      "datasourceConfiguration.properties[0].key",
+      createMessage(GSHEET_AUTHORISED_FILE_IDS_KEY),
+    );
+    set(datasource, "datasourceConfiguration.properties[0].value", fileIds);
```
ChandanBalajiBP, line 1208: Can we merge these two set call to one?

app/client/src/entities/Datasource/index.ts
```diff
@@ -13,6 +13,12 @@ export enum AuthenticationStatus {
   SUCCESS = "SUCCESS",
   FAILURE = "FAILURE",
 }
+
+export enum FilePickerActionStatus {
+  CANCEL = "cancel",
```
ChandanBalajiBP, line 18: can you please update these to Uppercases. You can refer above enum.

ChandanBalajiBP (changes requested):
Really appreciate the changes you have added in this PR. You have used enum and constants at required places👏

app/client/src/entities/Datasource/index.ts
sneha122: @ChandanBalajiBP These values `cancel` and `picked` are coming from google's file picker api, hence cannot change it.

app/client/src/sagas/DatasourcesSagas.ts
sneha122: @ChandanBalajiBP Can we set two different properties in the same set call? If yes could you please point me to the reference code for the same?

ChandanBalajiBP: > ```
> set(datasource, "datasourceConfiguration.properties[0]", { 
> key: createMessage(GSHEET_AUTHORISED_FILE_IDS_KEY,
> value: fileIds
> })
> ```

We are reducing a call to setter method this way.

sneha122: Thanks a lot!! This was really helpful, have made the corresponding changes. Please check @ChandanBalajiBP

ChandanBalajiBP (approved):
LGTM!

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/FileListMethod.java
```diff
@@ -89,10 +104,22 @@ public JsonNode transformTriggerResponse(JsonNode response, MethodConfig methodC
         List<Map<String, String>> filesList = StreamSupport
                 .stream(response.get("files").spliterator(), false)
                 .map(file -> {
-                    final String spreadSheetUrl = "https://docs.google.com/spreadsheets/d/" + file.get("id").asText() + "/edit";
-                    return Map.of("label", file.get("name").asText(),
-                            "value", spreadSheetUrl);
+                    if (allowedFileIds != null && !allowedFileIds.isEmpty()) {
```
sumitsum, line 107: Is this code section a duplicate of the code in `transformExecuteResponse` ? If so, should we extract the common code in a static function in say `SheetsUtil.java` and call that function instead ?

app/server/appsmith-plugins/googleSheetsPlugin/src/test/java/com/external/config/FileListMethodTest.java
```diff
@@ -120,4 +122,49 @@ public void testTransformTriggerResponse_withSheets_returnsDropdownOptions() thr
...
+        assertNotNull(jsonNode);
+
+        MethodConfig methodConfig = new MethodConfig(new HashMap<>());
+
+        TriggerMethod fileListMethod = new FileListMethod(objectMapper);
+        JsonNode result = fileListMethod.transformTriggerResponse(jsonNode, methodConfig, null);
+
+        assertNotNull(result);
+        assertTrue(result.isArray());
+        assertEquals(result.size(), 2);
```
sumitsum, line 142: Can you please add an assert here to match the exact expected content of `result` ?

app/server/appsmith-server/src/test/java/com/appsmith/server/solutions/ExamplesWorkspaceClonerTests.java
```diff
@@ -886,7 +886,8 @@ public void cloneApplicationWithActionsThrice() {
                             "header prefix",
                             Set.of(
                                     new Property("custom token param 1", "custom token param value 1"),
-                                    new Property("custom token param 2", "custom token param value 2")
+                                    new Property("custom token param 2", "custom token param value 2"),
+                                    new Property("custom token param 3", Set.of("file1", "file2", "file3"))
```
sumitsum, line 890: Sorry, I couldn't understand why this change is required here. Can you please provide details for this change ?

app/server/appsmith-plugins/googleSheetsPlugin/src/test/java/com/external/config/FileListMethodTest.java
```diff
@@ -120,4 +122,49 @@ public void testTransformTriggerResponse_withSheets_returnsDropdownOptions() thr
         expectedObjectNode.put("value", "https://docs.google.com/spreadsheets/d/testId/edit");
         assertEquals(expectedObjectNode, result.get(0));
     }
+
+    @Test
+    public void testTransformTriggerResponse_withAllSheetsAccess_returnsAllSheets() throws JsonProcessingException {
```
sumitsum, line 127: Can you please add similar methods for `transformExecuteResponse` method as well ?

sumitsum (changes requested):

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/FileListMethod.java
```diff
@@ -55,11 +57,24 @@ public JsonNode transformExecutionResponse(JsonNode response, MethodConfig metho
...
+                        String fileId = file.get("id").asText();
+                        // This will filter out and send only authorised google sheet files to client
+                        if (allowedFileIds.contains(fileId)) {
+                            final String spreadSheetUrl = "https://docs.google.com/spreadsheets/d/" + file.get("id").asText() + "/edit";
+                            return Map.of("id", file.get("id").asText(),
+                                    "name", file.get("name").asText(),
+                                    "url", spreadSheetUrl);
+                        }
+                    } else {
```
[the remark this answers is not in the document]
sneha122: In this case, the secondary if checks if the file is present in authorised list, if not it will come out of the if block, and return null by default [line 75], and in the end while aggregating all data as list, I am not taking these null objects [line 77]

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/GetStructureMethod.java
```diff
@@ -101,7 +101,7 @@ private List<String> validateInputs(MethodConfig methodConfig) {
     }
 
     @Override
-    public JsonNode transformExecutionResponse(JsonNode response, MethodConfig methodConfig) {
+    public JsonNode transformExecutionResponse(JsonNode response, MethodConfig methodConfig, Set<String> allowedFileIds) {
```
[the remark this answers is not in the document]
sneha122: By defining, do you mean in the interface, we would define it like this, and overriding classes can then decide whether to have 2/3 params?

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/utils/SheetsUtil.java
```diff
@@ -18,4 +23,15 @@ public static int getColumnNumber(String columnName) {
         }
         return 1;
     }
+
+    public static Set<String> getAuthorizedSheetIds(DatasourceConfiguration datasourceConfiguration) {
+        if (datasourceConfiguration.getProperties() != null
```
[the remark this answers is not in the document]
sneha122: @sumitsum isEmpty only checks if array is empty or not, if properties is null, it will throw NullPointerException right? Also I checked the code, properties is by default set to empty array so using isEmpty should work, as this will never be null in that case. Reference: https://stackoverflow.com/questions/11512034/does-java-util-list-isempty-check-if-the-list-itself-is-null. Let me know what you think.

app/server/appsmith-server/src/test/java/com/appsmith/server/solutions/ExamplesWorkspaceClonerTests.java
sneha122: This is not needed, I think this test was failing on server while coding, so I had added that, but checked now, removing it does not fail the test, hence removing this code. Thanks @sumitsum

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/plugins/GoogleSheetsPlugin.java
```diff
@@ -171,6 +172,9 @@ public Mono<ActionExecutionResult> executeCommon(Void connection,
             final OAuth2 oauth2 = (OAuth2) datasourceConfiguration.getAuthentication();
             assert (oauth2.getAuthenticationResponse() != null);
 
+            // This will get list of authorised sheet ids from datasource config, and transform execution response to contain only authorised files
+            final Set<String> allowedFileIds = SheetsUtil.getAuthorizedSheetIds(datasourceConfiguration);
```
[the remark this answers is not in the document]
sumitsum: Hi @sneha122 I am not able to see this change. Can you please check once ?

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/plugins/GoogleSheetsPlugin.java
```diff
@@ -171,6 +172,9 @@ public Mono<ActionExecutionResult> executeCommon(Void connection,
             final OAuth2 oauth2 = (OAuth2) datasourceConfiguration.getAuthentication();
             assert (oauth2.getAuthenticationResponse() != null);
 
+            // This will get list of authorised sheet ids from datasource config, and transform execution response to contain only authorised files
+            final Set<String> userAuthorizedSheetIds = SheetsUtil.getAuthorizedSheetIds(datasourceConfiguration);
```
sumitsum, line 176: Can you please update this usage to `getAuthorizedSheetIds` and the import to `import ...SheetsUtil.getAuthorizedSheetIds` ? Also, please update the same way in rest of the files wherever applicable.

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/utils/SheetsUtil.java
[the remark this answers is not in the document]
sumitsum: Hi @sneha122 can you please check for `CollectionUtils.isEmpty(properties)` in the code base and mention if you were thinking about the same thing ?

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/GetStructureMethod.java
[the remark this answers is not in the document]
sumitsum: Hi @sneha122 yes, we would define the default implementation in the interface. This way any class which does not override the 3 param method will still have a default implementation calling the 2 param method. The idea here is to save use form having to add any extra `null` param or any extra implementation at all at the individual class level.

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/FileListMethod.java
```diff
@@ -55,11 +57,24 @@ public JsonNode transformExecutionResponse(JsonNode response, MethodConfig metho
         List<Map<String, String>> filesList = StreamSupport
                 .stream(response.get("files").spliterator(), false)
                 .map(file -> {
-                    final String spreadSheetUrl = "https://docs.google.com/spreadsheets/d/" + file.get("id").asText() + "/edit";
-                    return Map.of("id", file.get("id").asText(),
-                            "name", file.get("name").asText(),
-                            "url", spreadSheetUrl);
+                    if (allowedFileIds != null && !allowedFileIds.isEmpty()) {
```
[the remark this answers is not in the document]
sumitsum: @sneha122 can you please check for this and mention if you were thinking about the same thing: `StringUtils.isEmpty(datasourceConfiguration.getUrl())` ?

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/FileListMethod.java
```diff
@@ -55,11 +58,18 @@ public JsonNode transformExecutionResponse(JsonNode response, MethodConfig metho
         List<Map<String, String>> filesList = StreamSupport
                 .stream(response.get("files").spliterator(), false)
                 .map(file -> {
-                    final String spreadSheetUrl = "https://docs.google.com/spreadsheets/d/" + file.get("id").asText() + "/edit";
-                    return Map.of("id", file.get("id").asText(),
-                            "name", file.get("name").asText(),
-                            "url", spreadSheetUrl);
+                    if (userAuthorizedSheetIds != null && !userAuthorizedSheetIds.isEmpty()) {
```
sumitsum, line 61: Hi @sneha122 in case the entire block from the line 61 to 70 is still identical, can we refactor this into another function as well ?

app/server/appsmith-plugins/googleSheetsPlugin/src/test/java/com/external/config/FileListMethodTest.java
```diff
@@ -120,4 +122,116 @@ public void testTransformTriggerResponse_withSheets_returnsDropdownOptions() thr
...
+
+        final ObjectNode expectedObjectNode1 = objectMapper.createObjectNode();
+        expectedObjectNode1.put("label", "test1");
+        expectedObjectNode1.put("value", "https://docs.google.com/spreadsheets/d/id1/edit");
+
+        final ObjectNode expectedObjectNode2 = objectMapper.createObjectNode();
+        expectedObjectNode2.put("label", "test2");
+        expectedObjectNode2.put("value", "https://docs.google.com/spreadsheets/d/id2/edit");
+
+        assertEquals(expectedObjectNode1, result.get(0));
```
sumitsum, line 152: QQ: Are we sure about the order of values here i.e. if we are using a `set` datastructure anywhere to hold values anywhere the order would get distured resulting in flaky tests. Can you please confirm this once ? Same for the test for `transformExecuteResponse`.

sumitsum (changes requested):

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/utils/SheetsUtil.java
```diff
@@ -18,4 +26,34 @@ public static int getColumnNumber(String columnName) {
...
+    }
+
+    public static Map<String, String> getSpreadsheetData(JsonNode file, Set<String> userAuthorizedSheetIds) {
+        if (userAuthorizedSheetIds == null) {
+            return extractSheetData((JsonNode) file);
+        }
+
+        String fileId = file.get("id").asText();
+        // This will filter out and send only authorised google sheet files to client
+        if (userAuthorizedSheetIds.contains(fileId)) {
```
sumitsum, line 47: Minor: could we combine the two if conditions into one like: 
```
if (userAuthorizedSheetIds == null || userAuthorizedSheetIds.contains(fileId))
```
I feel like you had mentioned something like this during our conversation yesterday but for some reason we couldn't get back to it.

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/GetStructureMethod.java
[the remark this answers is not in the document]
sumitsum: As discussed yesterday, we are dropping this idea because it is making readability worse.

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/utils/SheetsUtil.java
sneha122: @sumitsum We could but lets not do it, this way its clear that null condition is for all sheets modality and contains condition is for specific sheets. Also we did discuss about combining the if blocks, but instead we refactored the code to avoid nested if and wrote like this. LMK what you think

sumitsum: Hi @sneha122 agree with the reasoning here. In that case, to make the reason more clear can you please add a comment on top of both the if conditions explaining which Gsheet dropdown condition they correspond to ?

sneha122: Yup will do that

sumitsum (approved):

sumitsum (approved):

sneha122:
> Deploy-Preview-URL: https://ce-21699.dp.appsmith.com

@prapullac

PR #21717, opened 2023-03-23
fix: Conversion algo not respecting the min-height

### LINKED ISSUE #21656: [bug][Conversion Algorithm][Widget responsivness] the switchbox on running conversion algorithm is not adhering properly with the boundary boxes
https://www.loom.com/share/adae2b64a925457ab32029915c46146a

aswathkk:
For the widgets that are not vertically resizable, after conversion we need to set their height to be that of the min-height defined in the widgets config.

Fixes #21656


app/client/src/utils/autoLayout/positionUtils.ts
```diff
@@ -331,6 +334,17 @@ export function extractAlignmentInfo(
...
 
+    // Make sure that the height of the widget that cannot be vertically resized is
+    // set to the minimum height after converting from fixed to auto layout.
+    if (
+      firstTimeDSLUpdate &&
+      isFillWidget &&
+      disableResizeHandles?.vertical &&
+      minHeight
+    ) {
+      rows = minHeight / rowSpace;
```
rahulramesha, line 345: Is it possible that any of these widgets might have height greater than minHeight? This could end up cutting off the widget's full height. Can you please try Math.max(rows, minHeight / rowSpace)?

rahulramesha: Also, please try to use Math.round to get numerical values rather than in decimals.

rahulramesha: ideally in the grid system, we keep numerical values. In Autolayout, columns make sense to have decimal values but rows i feel should only be numericals not decimals. Please check with Preet or Ashok on this.

rahulramesha (approved):

PR #21756, opened 2023-03-24
chore: Include user-agent in events from backend

sharat87:
We get the `User-Agent` header for frontend events today, but not for backend events. This PR get the `User-Agent` header value, from the current request context, if any, and sends it along with any analytics events.

[Relevant Slack sonversation](https://theappsmith.slack.com/archives/C02MUD8DNUR/p1679579820266469?thread_ts=1679496974.642199&cid=C02MUD8DNUR).

Sample backend event with `userAgent`:

Screenshot 2023-03-24 at 5 14 35 PM

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ExchangeUtils.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ExchangeUtils.java
@@ -21,7 +21,7 @@ public class ExchangeUtils {
      * @return a Mono that resolves to the value of the given header, if present. Else, an empty Mono.
      * @param headerName The header name to look for.
      */
-    public static Mono<String> getHeaderFromCurrentRequest(String headerName) {
+    private static Mono<String> getHeaderFromCurrentRequest(String headerName) {
         return Mono.deferContextual(Mono::just)
                 .flatMap(contextView -> Mono.justOrEmpty(
                         contextView.get(ServerWebExchange.class).getRequest().getHeaders().getFirst(headerName)
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ExchangeUtils.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ExchangeUtils.java
@@ -8,6 +8,28 @@ import reactor.core.publisher.Mono;
 public class ExchangeUtils {
 
     public static final String HEADER_ANONYMOUS_USER_ID = "X-Anonymous-User-Id";
+    public static final String USER_AGENT = "User-Agent";
+
+    private ExchangeUtils() {
+        // This is a utility class. Instantiation is not allowed.
+    }
+
+    /**
+     * Returns the value of the given header, from the _current_ request. Since this gets the header from
+     * the current request, it has to be called from a request context. It won't work in new background contexts, like
+     * when calling `.subscribe()` on a Mono.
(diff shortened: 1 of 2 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ExchangeUtils.java
```diff
@@ -8,22 +8,43 @@
...
+     * @return a Mono that resolves to the value of the given header, if present. Else, `FieldName.ANONYMOUS_USER`.
+     * @param headerName The header name to look for.
      */
-    public static Mono<String> getAnonymousUserIdFromCurrentRequest() {
+    public static Mono<String> getHeaderFromCurrentRequest(String headerName) {
         return Mono.deferContextual(Mono::just)
                 .map(contextView -> ObjectUtils.defaultIfNull(
-                        contextView.get(ServerWebExchange.class).getRequest().getHeaders().getFirst(HEADER_ANONYMOUS_USER_ID),
+                        contextView.get(ServerWebExchange.class).getRequest().getHeaders().getFirst(headerName),
                         FieldName.ANONYMOUS_USER
```
vivonk, line 28: If this method is now useful for providing header value for any header name, we should instead return a null value from here if headerName is not present in headers and not `FieldName.ANONYMOUS_USER`. So that `getHeaderFromCurrentRequest()` will be a generic function.

And in `getAnonymousUserIdFromCurrentRequest` provide `FieldName.ANONYMOUS_USER` as a value if the header is null.

vivonk (dismissed):
LGTM

sharat87:
@vivonk, sorry, once more please, made that method `private`, since we don't need it outside. We can make it public in the future, if and when needed.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCEImpl.java
```diff
@@ -196,16 +196,22 @@ public Mono<Void> sendEvent(String event, String userId, Map<String, ?> properti
...
+                    final String instanceId = tuple.getT3();
                     String userIdToSend = finalUserId;
                     if (FieldName.ANONYMOUS_USER.equals(finalUserId)) {
                         userIdToSend = StringUtils.defaultIfEmpty(userIdFromClient, FieldName.ANONYMOUS_USER);
                     }
-                    TrackMessage.Builder messageBuilder = TrackMessage.builder(event).userId(userIdToSend);
+                    TrackMessage.Builder messageBuilder = TrackMessage.builder(event)
+                        .userId(userIdToSend)
+                        .context(Map.of(
+                            "userAgent", userAgent
```
vishnu-gp, line 213: Super minor, if we could move this to FieldName class

vishnu-gp (approved):
LGTM. Left a super minor comment.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCEImpl.java
sharat87: It's not really a _field name_ for a domain object though, no? 🤔 It's the key that Segment Javascript library uses to send the user-agent. I'd rather not make this a constant elsewhere, cause, I don't think we want this to change even if some such reused constant changes it's value in the future. Does that sound okay?

infinitetrooper:
@vishnu-gp  - Not right now, we'll add it if users ask for it.

PR #21849, opened 2023-03-29
feat: wds button refactoring

KelvinOm:
## Description

1. Moved `wds` to the `design-system` folder. 
2. Added `headless` package and `HeadlessButton`.
3. Added `theming` package. `ThemeProvider` and two classes were created to work with tokens([TokensAccessor](https://github.com/appsmithorg/appsmith/pull/21849/files#diff-58832e8b8e8017648929473a90eb716e6a2671ba1749be3d4c5665b093bc3dc3)) and colors([ColorsAccessor](https://github.com/appsmithorg/appsmith/pull/21849/files#diff-f515e0eefc418c8bfc0710572e83a0029bd94f2fb975853f71730e5b11c774bd))
5. The token structure has been changed. The same class(TokensAccessor) is now used to create CSS variables and tokens for Figma([themeTokens.json](https://github.com/appsmithorg/appsmith/pull/21849/files#diff-5ad75b848cb4254c0bd0bb0bf6a89eeccb628dc0012752172654e12e62d570d9)) 

The final storybook is [here](https://widget-design-system-b1p6g13iq-get-appsmith.vercel.app/?path=/story/design-system-widgets--button).

## Type of change

## How Has This Been Tested?
- Manual

### Dev activity


app/client/packages/design-system/headless/src/components/HeadlessButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { useHover } from "@react-aria/interactions";
+import type { RefObject } from "react";
+import type { FocusableRef } from "@react-types/shared";
+import type { ButtonProps } from "@react-types/button";
+
+export interface HeadlessButtonProps extends ButtonProps {
+  className?: string;
+}
+
+export type HeadlessButtonRef = FocusableRef<HTMLElement>;
```
KelvinOm, line 16: I'll write docs and tests in another task.

app/client/packages/design-system/theming/src/components/ThemeProvider.tsx
```diff
@@ -0,0 +1,36 @@
...
+export interface ThemeProviderProps {
+  theme: Theme;
+  children: ReactNode;
+}
+
+/**
+ * creates locally scoped css variables
+ *
+ */
+const StyledProvider = styled.div<ThemeProviderProps>`
```
KelvinOm, line 18: Here we can get rid of the `styled components` dependency and use [setProperty](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/setProperty) instead. @jsartisan do you think it's worth it?

app/client/packages/design-system/theming/src/tokens/themeTokens.json
```diff
@@ -0,0 +1,96 @@
+{
+  "color": {
```
KelvinOm, line 2: @jsartisan @ichik Please look at the structure. Are there any comments on it?

It seems to fit well on the token studio model
Снимок экрана 2023-03-29 в 15 20 51

app/client/packages/storybook/.storybook/addons/theming/register.js
```diff
@@ -61,6 +61,9 @@ addons.register("wds/theming", () => {
         });
       };
 
+      const colorChange = (value) => updateGlobal("accentColor", value);
```
KelvinOm, line 64: Since quite a lot of events are triggered when choosing a color and color generation also requires resources, I added a debounce here.

app/client/packages/design-system/wds/src/components/Button/Button.stories.mdx
```diff
@@ -0,0 +1,28 @@
...
+
+# Button
+
+A button is a clickable element that is used to trigger an action.
+
+<Canvas>
+  <Story name="Button">{Template.bind({})}</Story>
+</Canvas>
+
+<ArgsTable story="Button" of={WdsButton} />
```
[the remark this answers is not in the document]
ichik: Previewing all variants at once can be useful for development when e.g. refining color algo.

app/client/packages/design-system/theming/src/tokens/themeTokens.json
ichik: Think it's fine for now.

KelvinOm: @ichik Can we build a button on these tokens and make sure everything works?

KelvinOm: I'm talking about colors. I'm not really familiar with how token studio works. How do I understand that I need to use `bg-accent-subtle-active`(from [themeTokens.json](https://github.com/appsmithorg/appsmith/pull/21849/files#diff-5ad75b848cb4254c0bd0bb0bf6a89eeccb628dc0012752172654e12e62d570d9))? Now I see `accent-light-hover` for example.

ichik: Apologies, based on your screenshot with the initial question I thought you were interested in the spacing tokens specifically, the linked prototype was relevant with regards to spacing only, color naming wasn't settled by then. 

But we've already agreed on the names for the colors, didn't we? You wrote [the breakdown](https://miro.com/app/board/uXjVPkSIDg0=/?share_link_id=723090747453) yourself and this file follows as far as I can see. [Figma POC5](https://www.figma.com/file/GpPBFUsa7FvFY8p6E4xumK/Color-Generator-POC5%3A-Clamped-accent-color%2C-multiple-widgets-application%2C-Dark-and-Light-modes?node-id=1%3A22&t=UWFLETNLNzffJE8p-1) uses similar names, but it wasn't fully settled at the time of creation, so there are minor differences.
POC6 is gonna use this naming scheme fully.

app/client/packages/design-system/theming/src/tokens/themeTokens.json
```diff
@@ -0,0 +1,96 @@
...
+    },
+    "fg-accent": {
+      "value": "#553de9",
+      "type": "color"
+    },
+    "fg-on-accent": {
+      "value": "#fff",
+      "type": "color"
+    },
+    "focus": {
```
ichik, line 35: This is the only one that I see difference with the above linked breakdown: should've been `bd-focus`.

app/client/packages/design-system/wds/src/components/Button/Button.stories.mdx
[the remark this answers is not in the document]
jsartisan: +1 to what vadim said.

app/client/packages/design-system/headless/src/components/HeadlessButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export const HeadlessButton = forwardRef(
+  (props: HeadlessButtonProps, ref: HeadlessButtonRef) => {
+    const { autoFocus, children, className, isDisabled } = props;
```
jsartisan, line 20: What are the use-cases of auto focus on buttons for us? It makes sense for inputs, but i haven't seen much use autofocus on buttons.

app/client/packages/design-system/headless/src/components/HeadlessButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+          className={classNames(className, {
+            "is-disabled": isDisabled,
+            "is-active": isPressed,
+            "is-hovered": isHovered,
+          })}
+          ref={domRef}
+        >
+          {children}
+        </button>
+      </FocusRing>
```
jsartisan, line 38: I will read more about it and come back but should we use useMemo here as on each render a new object will created. We can tell react to use the same className object if disabled, active, hovered is not changed at all.

```
import { useMemo } from "react";

// ...

const buttonClassNames = useMemo(() => ({
  [className]: true,
  "is-disabled": isDisabled,
  "is-active": isPressed,
  "is-hovered": isHovered,
}), [className, isDisabled, isPressed, isHovered]);

return (
  <FocusRing autoFocus={autoFocus} focusRingClass="focus-ring">
    <button
      {...mergeProps(buttonProps, hoverProps)}
      className={classNames(buttonClassNames)}
      ref={domRef}
    >
      {children}
    </button>
  </FocusRing>
);
```

I will need to check more on this if useMemo will be worth it or not.

app/client/packages/design-system/theming/src/tokens/themeTokens.json
KelvinOm: Got it, thank you! I suppose after this PR will be merged, we should start using token studio and integrate tokens into Figma as we planned.

app/client/packages/design-system/headless/src/components/HeadlessButton.tsx
KelvinOm: This is useful, for example, when you open a confirmation dialog and the focus sets immediately on the "OK" button.

app/client/packages/design-system/headless/src/components/HeadlessButton.tsx
KelvinOm: There is no performance issues here. Since there is a situation of calling useMemo VS creating an object. 

If we use `useMemo` and `useCallback` wherever, then in the end we will get performance issues because we will call tons of unnecessary functions where it is not necessary. `useMemo` and `useCallback` are recommended to be used where we fix performance problems and point-by-point.
[Here](https://kentcdodds.com/blog/usememo-and-usecallback) is an interesting article about this.

app/client/packages/design-system/wds/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export { WdsButton } from "./components/Button";
```
jsartisan, line 1: we can simply export Button instead of WdsButton.
`export { WdsButton as Button } from "./components/Button"`;

reason: let the user decid the import name. we should export the generic name.
If user wants the import name as WdsButton, he will do:
`import { Button as WdsButton }` from `@wds/button"`.

Another approach can be just changing the name to Button instead of WdsButton in `wds/src/components/Button/Button.tsx` itself

app/client/packages/design-system/theming/src/components/ThemeProvider.tsx
jsartisan: Both works according to me. It's just that you will need to do everything manually without styled component ( which styled-components do for you ) if you

app/client/packages/design-system/wds/src/components/Button/index.styled.tsx
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  & [data-component="text"] {
+    width: 100%;
+
+    span {
+      display: block;
+      text-overflow: ellipsis;
+      overflow: hidden;
+      white-space: nowrap;
```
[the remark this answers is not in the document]
KelvinOm: Created a [issue](https://github.com/appsmithorg/appsmith/issues/21925) for this

app/client/packages/design-system/theming/src/tokens/themeTokens.json
KelvinOm: fixed

app/client/packages/design-system/theming/src/components/ThemeProvider.tsx
KelvinOm: I'll leave the `styled-components` because if I use `setProperty`, all the variables are put in style attribute and make  DOM messy. We will keep in mind that we have the opportunity to get rid of `styled-components`.

app/client/packages/design-system/wds/src/index.ts
KelvinOm: In any other case, I would agree with you. But since we are dealing with several design systems in our project, I would like all imports to be transparent. This will make it easier for us to find and replace components in the future. 

Just compare

```
import {AdsButton} from "@design-system/ads";
import {WdsButton} from "@design-system/wds";
import {HeadlessButton} from "@design-system/headless";
```

```
import {Button} from "@design-system/ads";
import {Button} from "@design-system/wds";
import {Button} from "@design-system/headless";
```

app/client/packages/design-system/wds/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,56 @@
...
+      ...rest
+    } = props;
+
+    return (
+      <StyledButton
+        data-fit-container={isFitContainer}
+        data-loading={isLoading}
+        data-variant={variant}
+        isDisabled={isDisabled}
+        {...rest}
```
riodeuno, line 43: Let us try not to spread props like this. It is difficult to debug. How many props are we dealing with here?

app/client/packages/design-system/wds/src/components/Button/index.styled.tsx
```diff
@@ -0,0 +1,93 @@
...
+  overflow: hidden;
+  white-space: nowrap;
+  gap: var(--spacing-4);
+  padding: var(--spacing-2) var(--spacing-4);
+  min-height: calc(var(--sizing-root-unit) * 8);
+  border-radius: var(--border-radius-1);
+  user-select: none;
+
+  // TODO: remove this when we use only flex layout
+  &[data-fit-container="true"] {
```
riodeuno, line 20: This is interesting. Instead of using props directly we have used the 'data-' prefix attributes. Could you share this in the style guide or engineering guide for the benefit of the team?

app/client/packages/design-system/wds/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,56 @@
...
+        ref={ref}
+      >
+        {isLoading && <Spinner />}
+
+        {!isLoading && (
+          <Text data-component="text" fontFamily={fontFamily}>
+            {children}
```
riodeuno, line 50: Will there be non text children in a button? If we can be sure that the children is going to be a Unicode string, we might be able to avoid using the ReactNode type.

app/client/packages/design-system/wds/src/index.ts
riodeuno: @KelvinOm  Will we be also be exposing this to widget developers? If so, Widget developers can be external. Do we want this to be our public facing API?

KelvinOm: @riodeuno Yes we plan to supply our components to widget developers so I suppose this is should be a public API.

app/client/packages/design-system/headless/src/components/HeadlessButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+  (props: HeadlessButtonProps, ref: HeadlessButtonRef) => {
+    const { autoFocus, children, className, isDisabled } = props;
+    const domRef = useFocusableRef(ref) as RefObject<HTMLButtonElement>;
+    const { buttonProps, isPressed } = useButton(props, domRef);
+    const { hoverProps, isHovered } = useHover({ isDisabled });
+
+    return (
+      <FocusRing autoFocus={autoFocus} focusRingClass="focus-ring">
+        <button
+          {...mergeProps(buttonProps, hoverProps)}
```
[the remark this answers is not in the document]
KelvinOm: This component is almost completely taken as is from Spectrum. I am sure that it is well tested. Just look at the [history](https://react-spectrum.adobe.com/blog/building-a-button-part-1.html) of the component creation.

app/client/packages/design-system/headless/src/components/HeadlessButton.tsx
```diff
@@ -0,0 +1,41 @@
+import React, { forwardRef } from "react";
+import { useFocusableRef } from "@react-spectrum/utils";
```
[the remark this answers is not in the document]
KelvinOm: This type is bit expanded.

app/client/packages/design-system/theming/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@design-system/theming",
```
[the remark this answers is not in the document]
KelvinOm: Under the circumstances, I don't see a better naming. We are also developing a design system and I see no reason to avoid this in the name.

Also Related to [this](https://github.com/appsmithorg/appsmith/pull/21849#discussion_r1155596156).

cc @albinAppsmith.

app/client/packages/design-system/theming/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "dependencies": {
+    "react": "*",
+    "react-dom": "*"
+  },
+  "devDependencies": {
+    "@types/react": "*",
+    "@types/react-dom": "*"
+  },
+  "peerDependencies": {
+    "react": "^16.8.0 || ^17.0.0-rc.1 || ^18.0.0"
```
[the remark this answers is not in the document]
KelvinOm: These versions of react have slightly different requirements for the code, and so for greater transparency, all three versions are indicated.

app/client/packages/design-system/wds/src/components/Button/Button.tsx
KelvinOm: Nice catch! This is absolutely not necessary here. All props here must be passed explicitly.

app/client/packages/design-system/wds/src/components/Button/index.styled.tsx
KelvinOm: Please guide me. Where should I add this information?

app/client/packages/design-system/wds/src/components/Button/Button.tsx
ichik: Icons, although those don't necessarily need to be wrapped in Text.

KelvinOm:
@riodeuno Thank you for the review and valuable comments!

app/client/packages/design-system/theming/package.json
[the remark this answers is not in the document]
albinAppsmith: Yeah I'm also little skeptical about this naming since Appsmith Design System is aliased as `design-system` in main repo.
So now the places Appsmith Design System is being used is going to look like below
`import { Button} from "design-system"`
and wds will be
`import something from "@design-system/wds"`

@KelvinOm Can we name this as `widget-design-system`?

app/client/packages/design-system/headless/src/components/HeadlessButton.tsx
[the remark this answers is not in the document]
albinAppsmith: Since you are using react-aria buttons, be careful about the props they are using. For eg: they have `onPress` instead of `onClick`. And when a user passes `onClick` as far as I know they will not handle the `isDisabled` flag and all. Can you try this out once? Try passing `onClick` instead of `onPress` and check if `isDisabled` flag is working or not?

app/client/packages/design-system/wds/src/index.ts
riodeuno: If this is a public API, shall we expose these as something which doesn't have an abbreviated prefix? Basically, the context of what WDS is might be lost with our external contributors. So, if we really have to, let's use another more understandable prefix. Or, if we're all aligned we can go with `Button` as well.

app/client/packages/design-system/wds/src/components/Button/index.styled.tsx
riodeuno: I was thinking to compile everything here, and then partition this out for easier navigation. So, for now, let's add any coding guidelines we come across to this doc here: https://www.notion.so/appsmith/Appsmith-front-end-code-style-guide-86db29586bac4ba3b2e044c315cae892

app/client/packages/design-system/wds/src/index.ts
KelvinOm: 1. I don't really like the name `Button` as it will greatly complicate the work in our main application. If I search for `Button`, which button will I find? 
2. We are creating a design system for developing widgets, so in my opinion, this prefix is quite appropriate. 

@jsartisan @albinAppsmith WDYT?

app/client/packages/design-system/headless/src/components/HeadlessButton.tsx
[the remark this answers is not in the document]
KelvinOm: In TS context  onClick will cause an error


In Storybook you can use onClick but there is a depreciation warning.



With `isDisabled` flag, no events are called. Everything works as expected.

app/client/packages/design-system/wds/src/components/Button/Button.tsx
KelvinOm: Here we are talking about these parameters. Added them explicitly. 

```
onBlur,
onFocus,
onFocusChange,
onKeyDown,
onKeyUp,
onPress,
onPressChange,
onPressEnd,
onPressStart,
onPressUp,
```

app/client/packages/design-system/theming/package.json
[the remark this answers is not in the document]
KelvinOm: The issue here is that we will import not something but a button so it will be like 
```
import { Button } from "design-system"
import { Button } from "@design-system/wds"
```

We also have headless components that need to be imported in a similar way

`import { Button } from "@design-system/headless"`

We can name package `widget-design-system`, and we can and probably should remove the `Wds` prefix. If we remove the prefix, then we must understand that further use of WDS, ADS and headless in the monolith will lead to confusion. I think the solution here is to move widgets from the monolith to packages as well. If we use components in understandable isolated parts, then there will be no issue.

I gonna rename new packages and remove prefixes from components
```
@design-system/headless » @headles-design-system
@design-system/wds » @widget-design-system
```
so imports will be
```
import { Button } from "@headles-design-system"
import { Button } from "@widget-design-system"
```

Also, apparently the `theming` packages should go to `widget-design-system`. 

As a separate task, I think we can use the same approach as [Spectrum](https://github.com/adobe/react-spectrum/tree/main/packages/%40react-spectrum) when each component is designed as a separate package. So consumers of the design system will be able to import/export only what they need.

@jsartisan  @albinAppsmith  @riodeuno Does it make sense?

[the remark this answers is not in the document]
albinAppsmith: This sounds good to me and will give a clear picture to the developers on from where and what are they importing.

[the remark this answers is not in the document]
jsartisan: IMO, ads should rename `design-system` to `appsmith-design-system`

The way i see it @design-system will contain both ads and wds in future and if that happens, `@design-system/wds` and `@design-system/ads` makes more sense to me
For me, `@design-system` is umbrella for all the things of design system that is done at appsmith, whether is ads, wds, headless, theming etc.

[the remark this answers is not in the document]
riodeuno: I agree with @jsartisan 's suggested nomenclature. Sounds clean.
Regarding widgets being in a separate package. I vehemently agree. This will move us in the right direction for the future.
I'll add @rajatagrawal to the conversation, as he has recently made some improvements that will help us properly isolate widgets from the core platform.

app/client/packages/design-system/wds/src/components/Button/Button.tsx
riodeuno: Thank you. This may be more verbose, but I think it is easier to debug.

app/client/packages/design-system/wds/src/components/Button/index.styled.tsx
riodeuno: Thank you!

app/client/packages/design-system/theming/package.json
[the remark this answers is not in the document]
riodeuno: Also, on a side note, let's refrain from using the abbreviation `ads`, as it is likely to be misconstrued for advertisements.

@design-system/widgets, @design-system/core or @design-system/appsmith, etc, could also work here.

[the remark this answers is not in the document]
KelvinOm: In conclusion, we will have such imports with such names

```
import { Button } from "@design-system/ads/button"
import { Button } from "@design-system/wds/button"
import { Button } from "@design-system/headless/button"
```

We will also be able to create a separate package that will import everything like this
```
import { Button } from "@design-system/ads"
import { Button } from "@design-system/wds"
import { Button } from "@design-system/headless"
```

Put your thumbs up 👍  if you agree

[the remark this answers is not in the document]
albinAppsmith: Like I mentioned before since Appsmith Design System is separate repo, while making the above changes we might need to change the alias in packag.json
Screenshot 2023-04-05 at 1 19 00 PM

or it might can cause confusion in developers when they see
`import { Button } from "@design-system/wds/button"`
`import { Button } from "design-system`

Also, We should not be exposing headless components to outside wds, if this happen people might start using headless components and not use Appsmith Design System components.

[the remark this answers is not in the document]
KelvinOm: @albinAppsmith First of I wold like to suggest to move into local packages.This will help us simplify the work and implementation in the main app. It will also help us maintain the consistency of the code.

> Also, We should not be exposing headless components to outside wds, if this happen people might start using headless components and not use Appsmith Design System components.

We may not add it to the main package.json. But headless components should be a separate package, it can also be used in the future to create custom widgets.

app/client/packages/design-system/theming/src/components/ThemeProvider.tsx
```diff
@@ -0,0 +1,36 @@
...
+  ${({ theme }) => {
+    return css`
+      ${Object.keys(theme).map((key) => {
+        return Object.keys(theme[key]).map((nestedKey) => {
+          return `--${kebabCase(key)}-${kebabCase(nestedKey)}: ${
```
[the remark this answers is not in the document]
KelvinOm: Our variables will be lower down the DOM and will have a higher priority, so we should not conflict with them. We can add a prefix prop to the provider to generate unique variables. But I really like what variables we have now. So clean.

```
--color-bg-accent: #553de9;
--color-bg-accent-hover: #6e61ff;
--color-bg-accent-active: #8f85ff;
--color-bg-accent-subtle-hover: #b6adff;
--color-bg-accent-subtle-active: #d7d0ff;
--color-bd-accent: #553de9;
--color-fg-accent: #553de9;
--color-fg-on-accent: #fff;
--color-bd-focus: #2a82ea;
--sizing-root-unit: 4px;
--spacing-0: 0px;
--spacing-1: 4px;
--spacing-2: 8px;
--spacing-3: 12px;
--spacing-4: 16px;
--spacing-5: 20px;
--border-radius-1: 0px;
--box-shadow-1: 0 0 0 0 rgba(0,0,0,0.2);
--border-width-1: 1px;
--opacity-disabled: 0.3;
```

app/client/packages/design-system/theming/src/tokens/defaultTokens.json
```diff
@@ -0,0 +1,18 @@
+{
```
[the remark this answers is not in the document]
KelvinOm: I will [write](https://github.com/appsmithorg/appsmith/issues/21918) documentation about all of this

app/client/packages/design-system/theming/src/tokens/defaultTokens.json
```diff
@@ -0,0 +1,18 @@
...
+  "seedColor": "#553de9",
+  "focusColor": "#2a82ea",
+  "rootUnit": 4,
+  "isDarkMode": false,
+  "borderRadius": {
+    "1": "0px"
```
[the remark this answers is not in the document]
KelvinOm: The key has a particular meaning. In design systems, we can use t-shirt sizes(for example) to indicate a larger or smaller value. We decided to use numeric values in case a fractional value appears between 1 and 2, then we can write 2.5 if we use S and M, then we will not be able to do it. 

There will be more of these values in the future.

app/client/packages/design-system/theming/src/tokens/themeTokens.json
```diff
@@ -0,0 +1,96 @@
...
+  "sizing": {
+    "root-unit": {
+      "value": "4px",
+      "type": "sizing"
+    }
+  },
+  "spacing": {
+    "0": {
+      "value": "0px",
+      "type": "spacing"
```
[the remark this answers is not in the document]
KelvinOm: Both of them are based on the base value of the grid (`rootUnit`). Spacing is used to represent padding and margin and can have a non-linear gradation (see example). Sizing is used to set the height and width of elements.

Снимок экрана 2023-04-05 в 14 17 21

app/client/packages/design-system/theming/src/utils/ColorsAccessor.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  constructor(color: ColorTypes) {
+    this.color = this.parse(color);
+  }
+
+  private parse = (color: ColorTypes) => {
+    try {
+      return new Color(color);
+    } catch (error) {
+      return new Color(defaultsTokens.seedColor);
```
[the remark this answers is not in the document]
KelvinOm: I think we don't need it now.

app/client/packages/design-system/theming/src/utils/ColorsAccessor.ts
```diff
@@ -0,0 +1,95 @@
...
+    const contrast = this.color.contrast(new Color("#fff"), "APCA");
+
+    // if contrast is less than -35 then the text color should be white
+    if (contrast < -60) return "#fff";
+
+    return "#000";
+  };
+
+  getFocus = () => {
+    return defaultsTokens.focusColor;
```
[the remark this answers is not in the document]
KelvinOm: Now we have decided that we will use a certain color for the focus, but if we decide to calculate it based on seedColor, then we will have such an opportunity

app/client/packages/design-system/theming/src/utils/TokensAccessor.ts
```diff
@@ -0,0 +1,165 @@
...
+};
+
+type ThemeTokens = {
+  [key in TokenType]: { [key: string]: Token };
+};
+
+type TokenObj = { [key: string]: string | number };
+
+export class TokensAccessor {
+  constructor(
```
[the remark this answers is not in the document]
KelvinOm: This is a TS feature called [Parameter Properties](https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties)

app/client/packages/design-system/theming/src/utils/ColorsAccessor.ts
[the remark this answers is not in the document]
riodeuno: I am curious to know if we anticipate any scenario where the catch clause will trigger.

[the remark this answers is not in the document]
KelvinOm: Since we can have just a string with an invalid color at the input this code will help us not to fall.

PR #21938, opened 2023-03-31
ci: Remove license in ci-test.yml

sharat87:
The License key is unused in this repo. Setting it here is misleading, and can cause confusions at best, or missed accidental License failures at worst.

This PR removes the License key occurrences from `ci-test.yml`.

Note: Please review/approve only. Don't merge. This will likely cause a conflict, so I'd like to be available when this is merged so I can resolve the conflict. Thanks!

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -128,8 +128,6 @@ jobs:
 
       - name: Create folder
         if: steps.run_result.outputs.run_result != 'success'
-        env:
-          APPSMITH_LICENSE_KEY: ${{ secrets.APPSMITH_LICENSE_KEY }}
         working-directory: "."
         run: |
           mkdir -p cicontainerlocal/stacks/configuration/
@@ -146,8 +144,6 @@ jobs:
 
       - name: Run Appsmith & TED docker image
         if: steps.run_result.outputs.run_result != 'success'
-        env:
-          APPSMITH_LICENSE_KEY: ${{ secrets.APPSMITH_LICENSE_KEY }}
         working-directory: "."
         run: |
           sudo /etc/init.d/ssh stop ;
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-test.yml
```diff
@@ -158,7 +154,7 @@ jobs:
           -v ~/git-server/repos:/git-server/repos  appsmith/test-event-driver:latest
           cd cicontainerlocal
           docker run -d --name appsmith -p 80:80 -p 9001:9001 \
-            -v "$PWD/stacks:/appsmith-stacks" -e APPSMITH_LICENSE_KEY=$APPSMITH_LICENSE_KEY \
+            -v "$PWD/stacks:/appsmith-stacks" \
```
Aishwarya-U-R, line 157: Will the EE repo runs not get affected with this change?

sharat87: EE repo runs will use the workflow from the EE repo, right? We'll keep this argument in that repo's workflow file.

Aishwarya-U-R: When this file is merged to release & flowing to EE repo - will the changes not get over-written?
And now we may solve the conflicts, so every-time we make changes to this file - we need to do this then?

sharat87: Unless we make changes in this line or around this line, it shouldn't cause a conflict. It's the same with any other file in the repo.

Aishwarya-U-R: I see.. Because we have been making quite a few changes in these files for various things - git comments update, ci-merge-check functionality etc. Do not want this to affect EE repo runs in such cases.
So whats the merge conflict you expect when this is merged to CE release?

sharat87: Oh I think this line was changed separately in the EE repo, so I'm expecting a conflict. If not, there won't be a conflict. I just wanted to be around when this is merged, so _if_ there's a conflict, I'm available to resolve. 🙂

Aishwarya-U-R (approved):

PR #21951, opened 2023-03-31
chore: Add Cache-Control header for static assets

### LINKED ISSUE #10503: Set expiry headers on static resources

sharat87:
Fix #10503

### DIFF
--- a/deploy/docker/templates/nginx/nginx-app-http.conf.template.sh
+++ b/deploy/docker/templates/nginx/nginx-app-http.conf.template.sh
@@ -76,6 +76,7 @@ server {
   # If the path has an extension at the end, then respond with 404 status if the file not found.
   location ~ ^/(?!supervisor/).*\.[a-z]+$ {
     try_files \$uri =404;
+    add_header Cache-Control "max-age=31104000, immutable";  # 360 days
   }
 
   location /api {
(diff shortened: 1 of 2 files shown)


SatishGandham (approved):
LGTM

SatishGandham:
@sharat87, All of our static resources except few are in `/static` folder.
Will configuring rules for this path excluding the `supervisor` negation make things simpler?

sharat87:
@SatishGandham, good point, let me look into that. I don't recall why I added that supervisor check, have to dig in once.

But, your comment got me thinking. The `Cache-Control` header we are adding here, will apply to all JS, CSS and images etc., right? But not all are hashed, right? Like in `build/static/img/*`, I see they aren't hashed (not that they'd change any time soon anyway), but also the `build/pageService.js` etc.

Do you think we should make this a little more strict, like, only send that header for the `/static/js/*`, `/static/css/*` and `/static/media/*` paths? I'm quite leaning towards that, than a blanket application of this header. What say?

SatishGandham:
Yes @sharat87. What you are suggesting is a better approach with the high expiry time we set. The small number of files that will miss caching won't be a big issue, they are also cached by service worker anyway [not 100% sure, but will be easy to add them if they are not]

sharat87:
Sorry, merged by accident, meant to merge a different PR. Reverted in https://github.com/appsmithorg/appsmith/pull/22174.

PR #21933, opened 2023-03-30
perf: preload route chunks for view and edit modes

iamakulov:
## Description

This PR implements route preloading for edit and view modes. In my tests on an emulated 4G connection, this makes edit and view modes load 1-1.5s sooner¹:

Screenshot 2023-03-30 at 01 20 44

_↑ Edit mode. (Each line is a separate test run. `es5` is the `bundle-optimization` build, `es5-preload` is this branch’s build. `spinnerRender` is time until the spinner became visible; `fmp` is time until widgets became visible._

Screenshot 2023-03-30 at 01 25 45

_↑ View mode_

_¹ — Unfortunately, a large chunk of the delay is slow waterfalling API responses which I can’t control (but Appsmith is working on separately). Hence, while the absolute improvement is big, the relative one is small._

### How this works

- During the build time, we now emit two new files (`build/edit.html` and `build/view.html`). These files are identical to `build/index.html`, with the only difference – they include additional `<link rel="preload">` tags:

   ```html
   <link rel="preload" as="script" href="/static/js/20204.4e1d68a3.chunk.js" />
  <link rel="preload" as="script" href="/static/js/85342.2ee5493d.chunk.js" />
  <link rel="preload" as="script" href="/static/js/69141.80219ba3.chunk.js" />
  <link rel="preload" as="style" href="/static/css/92067.2c8c92d5.chunk.css" />
  <link rel="preload" as="script" href="/static/js/92067.02797c81.chunk.js" />
  <link rel="preload" as="script" href="/static/js/79624.63f1a5a9.chunk.js" />
  <link rel="preload" as="script" href="/static/js/54102.1274415f.chunk.js" />
  <link rel="preload" as="script" href="/static/js/29549.e40bd2ec.chunk.js" />
  <link rel="preload" as="script" href="/static/js/47128.6dea5b2b.chunk.js" />
  <link rel="preload" as="script" href="/static/js/7207.8709ca68.chunk.js" />
  <link rel="preload" as="script" href="/static/js/39427.3928a9cf.chunk.js" />
  <link rel="preload" as="script" href="/static/js/99662.8131f01c.chunk.js" />
  <link rel="preload" as="script" href="/static/js/editor.3b1b5510.chunk.js" />
  <link rel="preload" as="script" href="/static/js/69141.80219ba3.chunk.js" />
  <link rel="preload" as="script" href="/static/js/75280.41f43bef.chunk.js" />
[...]

PR #21986, opened 2023-04-01
chore: Render tailored html files in some cases

sharat87:
When opening Appsmith in edit mode, we _try_ to render the specialized `edit.html`, and for opening in view mode, we _try_ to render the specialized `view.html` files, instead of the `index.html`. If those files aren't available, we fallback to `index.html`, so there's no disruption in service.

These specialized files are, today, identical to `index.html`, but contain a few extra preloads that are tuned to work better for those pages.

Learn more at https://github.com/appsmithorg/appsmith/pull/21933.


deploy/docker/templates/nginx/nginx-app-http.conf.template.sh
```diff
@@ -73,6 +73,14 @@ server {
     try_files \$uri /index.html =404;
   }
 
+  location ~ ^/app/[^/]+/[^/]+/edit\b {
```
SatishGandham, line 76: I don't know the exact regex but, if there is something after the edit, then it is preceded by a `/`
This regex matches `/app/hello/123/editfoo`, it is not a problem now, but there may be a case in future.

SatishGandham (changes requested):

iamakulov (approved):
This will work for Cypress tests as well, right? (We have a Cypress test that verifies that the preloads in `view.html`/`index.html` weren’t broken.)

sharat87:
@iamakulov, yes. Cypress tests should be able to verify if things continue to work as is with the changes here.

SatishGandham (approved):

sharat87:
Just checking in, we still want/need this, right? Anything I can do to help move this faster?

sharat87:
Changes are already in `release`.

PR #21987, opened 2023-04-01
fix: clickable containers in the List widget

### LINKED ISSUE #21409: [Bug]: List V2 onItemClick Event Not Run When Clicking on Widgets Placed Inside The List Item

### Description

If a user clicks on a list item that contains other widgets such as images or text, the onItemClick event does not get fired. It only gets fired when clicking on the container/blank space.

### Steps To Reproduce

1. Add a List widget (List v2).
2. Add some images and text inside the list items.
3. Add an onItemClick event.
4. Click on the images or text and note that the onItemClick event does not run.

### Public Sample App

_No response_

### Environment

Production

### Issue video log

https://cdn.discordapp.com/attachments/1082691051929735179/1082696640504287363/vokoscreenNG-2023-03-07_16-03-30.webm

### Version

v1.9.12

Front logo [Front conversations](https://app.frontapp.com/open/top_83tjg)

Tooluloope:
## Description

This fix primarily targets the List Widget. It enables the List Widget Containers/Items to be selected when clicking on an empty space or a plain widget/element. For instance, if a Text widget is clicked, the onItemClick action on the list container is triggered. However, if the Text widget contains a clickable link, only the link is activated, and the OnClick event doesn't propagate to the List Widget Container. The triggeredItems functionality remains unaffected.

Fixes #21409

## Type of change

## How Has This Been Tested?

- Cypress

### Dev activity

### QA activity:


app/client/src/widgets/ContainerWidget/component/index.tsx
```diff
@@ -111,6 +112,27 @@ function ContainerComponentWrapper(
...
+      if ((isContainerWidget || isMainCanvas) && !isTargetMainCanvas) {
+        // This is to prevent every click/Resize from propagating outside it's bounding container.
+        e.stopPropagation();
+      }
+
+      props.onClick?.(e);
+    },
+    [props.onClick],
+  );
+
```
Tooluloope, line 135: @riodeuno This way, the widget doesn't lose it's focus when dragging, resizing or changing the auto height with limits

ashit-rath (commented):
Can we try something like this?
https://codesandbox.io/s/react-17-forked-ochg14?file=/src/App.js

Tooluloope:
@ashit-rath There's a slight issue with this. 
https://codesandbox.io/s/react-17-forked-7n8jqn?file=/src/components/Hello.js

ashit-rath:
> @ashit-rath There's a slight issue with this. https://codesandbox.io/s/react-17-forked-7n8jqn?file=/src/components/Hello.js

@Tooluloope I didn't get the issue you faced, could you elaborate?

Tooluloope:
Many widgets utilizing the onClick event have their click handlers defined within the widget itself, which are then passed down to the respective component as a property.

For instance, the onClick event handler was passed as a prop to the `Hello1` component in the provided sandbox. Upon examining instanceProps.onClick, it can be observed that the handler consistently exists.

app/client/src/widgets/ContainerWidget/component/index.tsx
riodeuno: The attribute `type` is kind of accidental here, because we were spreading props. I cannot guarantee that this attribute will exist in the future. So, if possible today, please use another mechanism to identify the type of the widget. If necessary, please add a classname to server your purpose.
Also, is the `props.onClick?.(e)` doing the right thing? Please confirm.

ashit-rath:
@Tooluloope Is this PR ready for review?

Tooluloope:
> @Tooluloope Is this PR ready for review?

Yes it's ready  @ashit-rath

ashit-rath:
@Tooluloope As now we have added the logic of `findReactInstanceProps`, do we still need to do `e.stopPropagation` on individual widgets?

Tooluloope:
> @Tooluloope As now we have added the logic of `findReactInstanceProps`, do we still need to do `e.stopPropagation` on individual widgets?

Yes, we still need it.

Majorly because we can't determine the target.

For example, in the Chart widget, It's not possible to determine where the user would click.

Since `findReactInstanceProps` only checks `onClick` on a target.

ashit-rath:
@Tooluloope 
>For example, in the Chart widget, It's not possible to determine where the user would click.

I have a few concerns, let's connect once you are back

app/client/src/components/editorComponents/DropTargetComponent.tsx
```diff
@@ -309,7 +318,7 @@ export function DropTargetComponent(props: DropTargetComponentProps) {
         className={`t--drop-target drop-target-${
           props.parentId || MAIN_CONTAINER_WIDGET_ID
         }`}
-        onClick={handleFocus}
+        onClick={!props.parentId ? handleFocus : undefined}
```
ashit-rath, line 321: Why do we have this condition?

app/client/src/widgets/ButtonGroupWidget/component/index.tsx
```diff
@@ -636,9 +636,13 @@ class ButtonGroupComponent extends React.Component<
               disabled={isButtonDisabled}
               key={button.id}
               loading={!!loadedBtnId}
-              onClick={() => {
-                this.onButtonClick(button.onClick, button.id);
-              }}
+              onClick={
+                button.onClick
```
ashit-rath, line 640: Can we do something like to make it DRY?
```
generateOnClick = (button) => {
  if (!button.onClick) return
  
  return () => { this.onButtonClick(button.onClick, button.id);  }
}

...

onClick={generateOnClick(button)}
```

app/client/src/widgets/IconButtonWidget/component/index.tsx
```diff
@@ -276,16 +276,18 @@ function IconButtonComponent(props: IconButtonComponentProps) {
     return width - WIDGET_PADDING * 2;
   }, [width, height]);
 
+  const hasOnClick = useMemo(
```
ashit-rath, line 279: why `useMemo`? It's not an expensive computation

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -1050,6 +1051,24 @@ class ListWidget extends BaseWidget<
     );
   };
 
+  shouldCallOnItemClick = (e: React.MouseEvent<HTMLElement>) => {
```
ashit-rath, line 1054: Can we make this as helper function and put it in common widgets util/helper file?

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -40,6 +40,7 @@ import type {
   TabsWidgetProps,
 } from "widgets/TabsWidget/constants";
 import { getMetaFlexLayers } from "./helper";
+import { checkForOnCLick } from "widgets/WidgetUtils";
```
ashit-rath, line 43: typo `checkForOnCLick` capital `L` in `click`

app/client/src/widgets/NumberSliderWidget/component/SilderRoot.tsx
```diff
@@ -28,6 +28,13 @@ const SliderWrapper = styled.div<SliderRootProps>(
 
 export const SliderRoot = forwardRef<HTMLDivElement, SliderRootProps>(
   (props, ref) => {
-    return <SliderWrapper {...props} ref={ref} tabIndex={-1} />;
+    return (
+      <SliderWrapper
+        {...props}
+        onClick={(e) => props.onClick?.(e)}
```
ashit-rath, line 34: Do we need to do this?

app/client/src/widgets/WidgetUtils.ts
```diff
@@ -891,3 +894,30 @@ export const scrollCSS = css`
...
+    }
+  }
+  return null;
+};
+
+export const checkForOnCLick = (e: React.MouseEvent<HTMLElement>) => {
+  let target = e.target as HTMLElement | null;
+  const currentTarget = e.currentTarget as HTMLElement;
+  while (
+    !target?.classList.contains(WIDGET_CLICK_CHECK_CLASS) &&
```
ashit-rath, line 911: Since `PositionedContainer` gets added when `isFlexChild` is false, this check will fail in responsive layout.

Also if `detachFromLayout` is `true` then also `PositionContainer` will not get applied.

We might have to add a new div on top of widgets

app/client/src/widgets/WidgetUtils.ts
```diff
@@ -891,3 +894,30 @@ export const scrollCSS = css`
...
+      return domElement[key];
+    }
+  }
+  return null;
+};
+
+export const checkForOnCLick = (e: React.MouseEvent<HTMLElement>) => {
+  let target = e.target as HTMLElement | null;
+  const currentTarget = e.currentTarget as HTMLElement;
+  while (
```
ashit-rath, line 910: Add a new line before this

app/client/src/widgets/WidgetUtils.ts
```diff
@@ -891,3 +894,30 @@ export const scrollCSS = css`
...
+    !target?.classList.contains(WIDGET_CLICK_CHECK_CLASS) &&
+    target &&
+    target !== currentTarget
+  ) {
+    const hasOnClick = Boolean(findReactInstanceProps(target)?.onClick);
```
ashit-rath, line 915: I think we should also check for `target.onclick` to make sure we look at native events as well

app/client/src/components/editorComponents/DropTargetComponent.tsx
Tooluloope: DropTarget is used by other Canvas Widgets as well, but we only want to deselect all widgets If it's the Main Canvas widget

app/client/src/widgets/ButtonGroupWidget/component/index.tsx
Tooluloope: if we do this, We are stating it would always have an `onclick` function irrespective of if `button.onClick` exist or not

app/client/src/components/editorComponents/DropTargetComponent.tsx
ashit-rath: Got it
Can we make it more descriptive by either moving the `!props.parentId` to a variable `mainContainer or checking the `widgetId === MAIN_CONTAINER_WIDGET_ID`?

app/client/src/widgets/ButtonGroupWidget/component/index.tsx
Tooluloope: So `findReactInstanceProps` would always see an `onClick`

ashit-rath: When the `button.onClick` is not present it returns undefined and that would the value of `onClick`

app/client/src/widgets/NumberSliderWidget/component/SilderRoot.tsx
Tooluloope: Yeah, I think we do, The slider does an action and it shouldn't trigger onItemClick

ashit-rath: When we do `...props`, if the `props` has an onClick it will automatically pass to the `SliderWrapper. Why do we need to explicitly define it here?

Tooluloope: So the `SliderWrapper` is just a DIV. The events that are passed are `onMouseDownCapture` and `onKeyDownCapture`. This was a way of explicitly making it have an onClick, as I don't think we should fire onItemCllick when the slider is clicked. 

Another way to do it would be to check for `onMouseDownCapture` in the `findReactInstanceProps`

ashit-rath (approved):

chandannkumar:
Tested this PR and LGTM
- Tested with multiple widgets inside List widget
- Tested with nested List widget
- Tested onItemClick event by disabling widgets, readonly scenarios
- Tested on older apps

app/client/src/components/editorComponents/WidgetComponentBoundary.tsx
```diff
@@ -0,0 +1,28 @@
...
+const WidgetComponentBoundaryWrapper = styled.div`
+  height: 100%;
+  width: 100%;
+`;
+
+function WidgetComponentBoundary(props: Props) {
+  return (
+    <WidgetComponentBoundaryWrapper
+      className={
+        props.widgetProps.type !== "CANVAS_WIDGET"
```
ashit-rath, line 18: What is the rationale behind this?

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -588,6 +589,15 @@ abstract class BaseWidget<
       </FlexComponent>
     );
   }
+  addWidgetComponentBoundary = (
+    content: ReactNode,
+    widgetProps: WidgetProps,
+  ) => (
+    <WidgetComponentBoundary widgetProps={widgetProps}>
```
ashit-rath, line 596: Don't pass the whole props, just pass the type.

ashit-rath (approved):

PR #21994, opened 2023-04-02
chore: Fix unnecessary re-shading of jars

sharat87:
Solves a single thing in the build configurations, resulting in a few wins.

1. Reduced number of warnings in the output.
   1. In release branch:
      ```
      mvn clean package -DskipTests | grep --fixed-strings --count '[WARNING]'
      3233
      ```
   1. In this PR's branch:
      ```
      mvn clean package -DskipTests | grep --fixed-strings --count '[WARNING]'
      172
      ```

2. All uber-jar files are shaded twice, currently. Once with the default execution of `maven-shade-plugin`, and again with the `shade-plugin-jar` execution in these `pom.xml` files. This is double-work, and is the cause of most of the warnings we see.
   1. This `shade-plugin-jar` was added to have the plugin information included in the `/META-INF/MANIFEST.MF` file, since we can't configure the default execution of the shade plugin (it comes to us from Spring Boot).
   2. Instead, we switch to configuring plugin information in a `/plugin.properties` file.
   3. Previously, we used `/plugin.properities` for plugin information in dev time, and `/META-INF/MANIFEST.MF` in production. This PR will change it so that we use `/plugin.properties` all the time. We configure PF4J with a custom plugin manager to achieve this.

3. Moved all `plugin.properties` into `src/main/resources`, so that they land up in the root of the final jar files. But this means, during development, loading the plugin fails since it looks for a `plugin.properties` at the root of the plugin module, i.e., next to the `src` folder.
   1. For this, in the custom plugin manager class, we change where we look for the `plugin.properties` file during development mode. In this mode, we look at the `target/classes/plugin.properties` file, which is where maven saves this file, taken from `src/main/resources/plugin.properties`.
   2. This also solves the duplication of the plugin properties that's currently present, between `plugin.properties` and the `<properties>` section of `pom.xml` files.

Here's the shade plugin's default execution and configuration, from Spring Boot: https://github.com/spring-projects/spring-boot/blob/v3.0.1/spring-boot-project/spring-boot-starters/spring-boot-starter-parent/build.gradle#L174.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/PluginConfiguration.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/PluginConfiguration.java
@@ -1,5 +1,6 @@
 package com.appsmith.server.configurations;
 
+import org.pf4j.PropertiesPluginDescriptorFinder;
 import org.pf4j.spring.SpringPluginManager;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
@@ -9,7 +10,16 @@ public class PluginConfiguration {
 
     @Bean
     public SpringPluginManager pluginManager() {
-        return new SpringPluginManager();
+        return new CustomPluginManager();
+    }
+
+    private static class CustomPluginManager extends SpringPluginManager {
+        public CustomPluginManager() {
+            super();
+            pluginDescriptorFinder = "development".equals(System.getProperty("pf4j.mode"))
+                ? new PropertiesPluginDescriptorFinder("target/classes/plugin.properties")
+                : new PropertiesPluginDescriptorFinder();
+        }
     }
 
 }
--- a/app/client/src/api/ActionAPI.tsx
+++ b/app/client/src/api/ActionAPI.tsx
@@ -196,6 +196,7 @@ class ActionAPI extends API {
       headers: {
         accept: "application/json",
         "Content-Type": "multipart/form-data",
+        Expect: "100-continue",
       },
       cancelToken: ActionAPI.abortActionExecutionTokenSource.token,
     });
--- a/app/client/src/api/ApiUtils.ts
+++ b/app/client/src/api/ApiUtils.ts
@@ -1,7 +1,9 @@
 import {
   createMessage,
   ERROR_0,
+  ERROR_413,
   ERROR_500,
+  GENERIC_API_EXECUTION_ERROR,
   SERVER_API_TIMEOUT_ERROR,
 } from "@appsmith/constants/messages";
 import type { AxiosRequestConfig, AxiosResponse } from "axios";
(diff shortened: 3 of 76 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/PluginConfiguration.java
```diff
@@ -9,7 +10,16 @@ public class PluginConfiguration {
...
     public SpringPluginManager pluginManager() {
-        return new SpringPluginManager();
+        return new CustomPluginManager();
+    }
+
+    private static class CustomPluginManager extends SpringPluginManager {
+        public CustomPluginManager() {
+            super();
+            pluginDescriptorFinder = "development".equals(System.getProperty("pf4j.mode"))
+                ? new PropertiesPluginDescriptorFinder("target/classes/plugin.properties")
```
nidhi-nair, line 20: What are we doing here with this condition Shri?

sharat87: Hey, thanks, missed this. I added a third point to the PR description with details. Essentially, since the `plugin.properties` is moved to a different location, this breaks the development mode. So we fix the path here, if the mode is dev, otherwise, we stick to PF4J's defaults.

nidhi-nair: Thanks, you mentioned that in production previously the default behaviour was to look at the `/META-INF/MANIFEST.MF` file. Does this mean that the default constructor you have used here will not be expecting details at `/META-INF/MANIFEST.MF` but understands the `target/classes/plugin.properties` path automatically?

sharat87: The default constructor looks into `/plugin.properties`. Which will get included in the jars, because we moved the `plugin.properties` into `src/main/resources`, and so will be included at the root of jars built by Maven. This doesn't look at the `/META-INF/MANIFEST.MF` file at all, and we don't include this information there at all either.

I'm not sure I got your question... does the above detail help? 🤔

nidhi-nair: I did not phrase that well, my bad. My question was about the intended functionality for `MANIFEST.MF`. What were we using it for before, why are we okay with not creating it anymore?

sharat87: Oh okay. We were using it to put information about the plugin. Like here:

https://github.com/appsmithorg/appsmith/blob/977eda638848ca9468924bc1e2f961d5d22d18f4/app/server/appsmith-plugins/amazons3Plugin/pom.xml#L72

The PF4J library would then pick up plugin metadata from this. But this information is also already there in the `plugin.properties` file, essentially duplicated. In this PR, we now have PF4J take this information from `plugin.properties` as well, and so we don't need to add this information to the manifest file. The manifest file will still be present, we just won't be adding plugin metadata to it.

nidhi-nair (approved):

nidhi-nair:
Is that magic, is it super charged workflow, or a faux pas? Did you just trigger an ok to test without the sha?

sharat87:
@nidhi-nair, you will be entertained: https://theappsmith.slack.com/archives/C0134BAVDB4/p1679297087315919?thread_ts=1679291707.639179&cid=C0134BAVDB4 🙂

PR #21996, opened 2023-04-02
perf: work around the V8 service worker regression

iamakulov:
## Description

This pull request works around [a problematic V8 heuristic](https://bugs.chromium.org/p/chromium/issues/detail?id=1428605) that makes bundle init much slower. This heuristic affects both `release` and `bundle-optimization` branches, but seem to be triggering more frequently in `bundle-optimization`.

The PR works around the heuristic by moving the pre-caching step from the `install` event to the `activate` one (which, per [this V8 blog post](https://v8.dev/blog/code-caching-for-devs#use-service-worker-caches), disables the heuristic).

On repeat visits, this seems to make main bundle init **8× faster (!)** compared to the `release` branch. (See the last screenshot below for more details.)

### More info

We use the [`workbox`](https://github.com/GoogleChrome/workbox) library in our service worker. `workbox` fetches and caches all build chunks whenever the service worker is installed for the first time.

Coincidentally, V8 (Chromium’s JS engine) [generates a “full code cache”](https://v8.dev/blog/code-caching-for-devs#use-service-worker-caches) for scripts cached while a service worker is installing. "Full code cache” is a result of compiling a whole script into bytecode. It makes execution of the app faster – now, you don’t need to compile scripts every time the app loads; you can just restore the compiled bytecode from the cache.

“It makes execution faster” – except it doesn’t! For whatever reason, in Appsmith, we observed a huge regression when the service worker-generated code cache would kick in. The cache will be extremely slow to deserialize, making total bundle execution 2-3 times longer:

| First load (no sw cache) | Subsequent load (with sw cache) |
| --- | --- |
| The bundle takes 880 ms to execute (compiling the necessary functions on the fly) <br> Screenshot 2023-04-02 at 22 09 47 | The bundle now takes just 500 ms to execute, but we spend 1870 ms in “Compile Code” (which actually just deserializes service worker’s code cache) <br> Screenshot 2023-04-02 at 22 09 40 |

This sounds like a V8 issue, so we filed a bug: https://bugs.chromium.org/p/chromium/issues/detail?id=1428605

### Results

To work around this, in Chromium, we now cache all assets from the `activate` event listener. This is non-canonical but seems to work well. 

Here’s how this affects performance:

Screenshot 2023-04-02 at 21 18 26 2

The code caches from the service worker are still used in all branches. However, while `release` and `bundle-optimization` branches get slower (because they use the “full” code cache), this branch gets _faster_ (because it uses a “normal” code cache). As a result, we get an 8× improvement between the branches!

See this V8 article for more details about different code caches: https://v8.dev/blog/code-caching-for-devs#use-service-worker-caches

## Type of change

## How Has This Been Tested?

- Manual: checking that the service worker works in Chromium, Firefox, and Safari

### Dev activity

### QA activity:

### DIFF
--- a/app/client/craco.build.config.js
+++ b/app/client/craco.build.config.js
@@ -16,6 +16,7 @@ plugins.push(
     mode: "development",
     swDest: "./pageService.js",
     maximumFileSizeToCacheInBytes: 11 * 1024 * 1024,
+    exclude: [/\.map$/, /^manifest.*\.js$/, /index\.html/],
   }),
 );
 
--- a/app/client/craco.build.config.js
+++ b/app/client/craco.build.config.js
@@ -17,8 +17,12 @@ plugins.push(
     swDest: "./pageService.js",
     maximumFileSizeToCacheInBytes: 11 * 1024 * 1024,
     exclude: [
+      // Don’t cache source maps and PWA manifests.
+      // (These are the default values of the `exclude` option: https://developer.chrome.com/docs/workbox/reference/workbox-build/#type-WebpackPartial,
+      // so we need to specify them explicitly if we’re extending this array.)
       /\.map$/,
       /^manifest.*\.js$/,
+      // Don’t cache the root html file
       /index\.html/,
       // Don’t cache LICENSE.txt files emitted by CRA
       // when a chunk includes some license comments
--- a/app/client/craco.build.config.js
+++ b/app/client/craco.build.config.js
@@ -16,7 +16,16 @@ plugins.push(
     mode: "development",
     swDest: "./pageService.js",
     maximumFileSizeToCacheInBytes: 11 * 1024 * 1024,
-    exclude: [/\.map$/, /^manifest.*\.js$/, /index\.html/],
+    exclude: [
+      /\.map$/,
+      /^manifest.*\.js$/,
+      /index\.html/,
+      // Don’t cache LICENSE.txt files emitted by CRA
+      // when a chunk includes some license comments
+      /LICENSE\.txt/,
+    ],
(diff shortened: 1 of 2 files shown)


app/client/craco.build.config.js
```diff
@@ -16,6 +16,16 @@ plugins.push(
     mode: "development",
     swDest: "./pageService.js",
     maximumFileSizeToCacheInBytes: 11 * 1024 * 1024,
+    exclude: [
```
SatishGandham, line 19: @iamakulov, What does this exclude do?

iamakulov: The same thing this `.filter()` was doing in a previous version of the service worker:

https://github.com/appsmithorg/appsmith/blob/44f323e813bdc05e39e13622c9201ea52421ed2c/app/client/src/serviceWorker.js#L30-L32

([Docs](https://developer.chrome.com/docs/workbox/reference/workbox-build/#type-WebpackPartial))

`/\.map$/` and `/^manifest.*\.js$/` are default values (so we need to put them into the array explicitly if we’re extending it) – let me add a comment about that. `index.html` excludes the same file the previous code was excluding. `/LICENSE\.txt/` just excludes a few extra chunks with license info that webpack emits but the app doesn’t need.

The reason I replaced `.filter()` with the `exclude` plugin option is that `exclude` makes the list of files smaller _before_ the list is inlined into the service worker. It doesn’t really matter right now (the difference is minuscule) but will matter much more eg if we decide to exclude icons from precaching. (There are lots of icons, so the list – and the service worker – will become significantly smaller.)

PR #22049, opened 2023-04-03
fix: set initialState from metaProps

Rishabh-Rathod:
## Description

Fix the meta value code to set the initialMetaState correctly. 

## Type of change

## How Has This Been Tested?

- Manual
- Jest
- Cypress

### Dev activity

### QA activity:


app/client/src/widgets/MetaHOC.tsx
```diff
@@ -55,7 +55,7 @@ function withMeta(WrappedWidget: typeof BaseWidget) {
       super(props);
       const metaProperties = WrappedWidget.getMetaPropertiesMap();
       this.initialMetaState = fromPairs(
-        Object.values(metaProperties).map((key, value) => {
+        Object.entries(metaProperties).map((key, value) => {
```
arunvjn, line 58: shouldn't this be `([key, value])` instead of (key, value)? 

Also, next like should be just [key, this.props[key]] right ?

Can we simply use `Object.keys` instead ?

Rishabh-Rathod: No, we want to pick values from meta properties object. So the code would be this  `[key, metaPropeties[key]]` 

I have made the changes.

app/client/src/widgets/MetaHOC.tsx
```diff
@@ -55,8 +55,8 @@ function withMeta(WrappedWidget: typeof BaseWidget) {
       super(props);
       const metaProperties = WrappedWidget.getMetaPropertiesMap();
       this.initialMetaState = fromPairs(
-        Object.values(metaProperties).map((key, value) => {
-          return [key, this.props[value]];
+        Object.keys(metaProperties).map((key) => {
+          return [key, metaProperties[key]];
```
arunvjn, line 59: Isn't this equivalent to `this.initialMetaState = fromPairs(Object.entries(metaProperties));` which is also just
`this.initialMetaState = metaProperties` ? 😵‍💫

arunvjn (approved):

ApekshaBhosale (approved):

PR #22080, opened 2023-04-04
feat: util to serve images locally or via remote url

### LINKED ISSUE #22004: [Task]: Bundle and serve images locally for air-gapped instances.

### SubTasks

_No response_

berzerkeer:
## Description

- On air-gapped instances we can't fetch appsmith assets from S3, that will result in broken which is not desirable. 
- So this adds a script and util function which searches the client and server codebase for the assets url and downloads the image and puts it in the `public` folder so that the browser can access those even in an airgapped instance since the assets are being served locally.

> Way to serve assets locally.

Fixes #22004 

Media
https://vdqm24wed6.vmaker.com/record/IquS90WbWgS1I0bz - blocked certain api routes from getting called on airgap

## Type of change

## How Has This Been Tested?

- Manual

### Dev activity

### QA activity:


app/client/src/pages/Editor/APIEditor/ProviderTemplates.tsx
```diff
@@ -318,7 +318,7 @@ class ProviderTemplates extends React.Component<ProviderTemplatesProps> {
             {" Back"}
           </span>
           <br />
-
+          {/* TODO: Check image URls */}
```
dipyamanbiswas07, line 321: Could you add your name to the todo comment.

app/client/src/ce/utils/airgapHelpers.tsx
```diff
@@ -0,0 +1,7 @@
+export const getAssetUrl = (src = "") => {
+  return src;
+};
+
+export const isAirgapped = () => {
```
dipyamanbiswas07, line 5: One question around this. 

Since this is going to be used in a lot of places, could we do the computation once and store it? either in redux or as a static class property?

dipyamanbiswas07 (dismissed):

app/client/src/ce/utils/airgapHelpers.tsx
berzerkeer: AFAIK this won't make any difference since there is no computation here, it just returns false here and on EE also it actually just returns what we have in the window object ( so basically like redux store ) and these values are set when we initialize the app and not going to change afterwards so we don't really need state management stuff. If we do it in redux there will be a lot of boilerplate code. So I guess what we have here is jut fine. Let me know if this makes sense , I could be wrong. Thanks !

app/client/src/pages/Editor/APIEditor/ProviderTemplates.tsx
berzerkeer: They are not needed anymore, removed it.

app/client/src/ce/utils/airgapHelpers.tsx
dipyamanbiswas07: Got it. Since there are no computations required, we are good.

dipyamanbiswas07 (approved):

dipyamanbiswas07 (approved):

app/client/src/pages/Editor/QueryEditor/EditorJSONtoForm.tsx
```diff
@@ -576,13 +577,15 @@ export function EditorJSONtoForm(props: Props) {
   }
 
   function SingleValue(props: SingleValueProps<OptionTypeBase>) {
+    {
```
ankitakinger, line 580: why is this needed?

app/client/src/constants/ImagesURL.ts
```diff
@@ -1,5 +1,7 @@
 import { ASSETS_CDN_URL } from "./ThirdPartyConstants";
+import { getAssetUrl } from "@appsmith/utils/airgapHelpers";
+
 export const getInfoThumbnail = () =>
-  `${ASSETS_CDN_URL}/crud/crud_info_thumbnail.png`;
-export const getInfoImage = (): string =>
-  `${ASSETS_CDN_URL}/crud/working-flow-chart.png`;
+  getAssetUrl(`${ASSETS_CDN_URL}/crud/crud_info_thumbnail.png`);
+export const getInfoImage = () =>
```
ankitakinger, line 6: will `getInfoImage` return type not be a string anymore?

app/client/src/pages/Editor/QueryEditor/EditorJSONtoForm.tsx
berzerkeer: Oh this is just a typo, I added a console log to check the props ,only the console log removed and not the braces , I will remove this now. Good catch, Thanks !

app/client/src/constants/ImagesURL.ts
berzerkeer: It will always be string , will add it back. I must've removed it when testing.

ankitakinger (approved):

ankitakinger (approved):

app/client/src/selectors/editorSelectors.tsx
```diff
@@ -342,10 +343,9 @@ export const getWidgetCards = createSelector(
   getWidgetConfigs,
   getIsAutoLayout,
   (widgetConfigs: WidgetConfigReducerState, isAutoLayout: boolean) => {
-    const cards = Object.values(widgetConfigs.config).filter(
-      (config) => !config.hideCard,
-    );
-
+    const cards = Object.values(widgetConfigs.config).filter((config) => {
+      return isAirgapped() ? config.widgetName !== "Map" : !config.hideCard;
```
iamenrique, line 347: This means that air-gapped instances allow "hidden cards" to be displayed? 
image

berzerkeer: Nope, it was a bug. This has been fixed in a different PR. https://github.com/appsmithorg/appsmith/pull/22440

PR #22091, opened 2023-04-04
ci: add recreate option for dps

pratapaprasanna:
fixes: [248](https://github.com/appsmithorg/cloud-deployment/issues/248)

https://github.com/appsmithorg/appsmith-ci/pull/14#issuecomment-1504569726
This doesnt delete the existing stacks 

https://github.com/appsmithorg/appsmith-ci/pull/14#issuecomment-1504842238
this deletes the stack, database in mongo and therefore creating a complete new stack

### DIFF
--- a/scripts/deploy_preview.sh
+++ b/scripts/deploy_preview.sh
@@ -45,14 +45,14 @@ kubectl config set-context --current --namespace=default
 echo "Getting the pods"
 kubectl get pods
 
-if [ ! -z "$RECREATE" ]
+if [[ -n "${RECREATE-}" ]]
 then
-  kubectl delete ns $NAMESPACE || echo "true"
+  kubectl delete ns $NAMESPACE || true
   mongosh "mongodb+srv://$DB_USERNAME:$DB_PASSWORD@$DB_URL/$DBNAME?retryWrites=true&minPoolSize=1&maxPoolSize=10&maxIdleTimeMS=900000&authSource=admin" --eval 'db.dropDatabase()'
 fi
 
 echo "Use kubernetes secret to Pull Image"
-kubectl create ns $NAMESPACE || echo true
+kubectl create ns $NAMESPACE || true
 
 kubectl create secret docker-registry $SECRET \
   --docker-server=https://index.docker.io/v1/ \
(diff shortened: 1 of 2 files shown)


scripts/deploy_preview.sh
```diff
@@ -45,11 +45,15 @@ kubectl config set-context --current --namespace=default
...
-kubectl delete ns $NAMESPACE || echo "true"
+if [ ! -z "$RECREATE" ]
+then
+  kubectl delete ns $NAMESPACE || echo "true"
+  mongosh "mongodb+srv://$DB_USERNAME:$DB_PASSWORD@$DB_URL/$DBNAME?retryWrites=true&minPoolSize=1&maxPoolSize=10&maxIdleTimeMS=900000&authSource=admin" --eval 'db.dropDatabase()'
+fi
 
 echo "Use kubernetes secret to Pull Image"
-kubectl create ns $NAMESPACE
+kubectl create ns $NAMESPACE || echo true
```
sharat87, line 55: This can just be `true`.

```suggestion
kubectl create ns $NAMESPACE || true
```

scripts/deploy_preview.sh
```diff
@@ -45,11 +45,15 @@ kubectl config set-context --current --namespace=default
 echo "Getting the pods"
 kubectl get pods
 
-echo "Delete previously created namespace"
-kubectl delete ns $NAMESPACE || echo "true"
+if [ ! -z "$RECREATE" ]
```
sharat87, line 48: Prefer something like this.

1. `-n` is the inverse of `-z`.
2. Double square brackets are almost _always_ better than single square brackets.
3. Using `${RECREATE-}` instead of just `$RECREATE`, means that is `nounset` is set in the shell, and this env variable is not defined, the script will break. Just a good habit/practice. 🙂

```suggestion
if [[ -n "${RECREATE-}" ]]
```

scripts/deploy_preview.sh
```diff
@@ -45,11 +45,15 @@ kubectl config set-context --current --namespace=default
...
+then
+  kubectl delete ns $NAMESPACE || echo "true"
```
sharat87, line 50: Also just `true`.

```suggestion
  kubectl delete ns "$NAMESPACE" || true
```

sharat87 (changes requested):
Minor refactoring, but otherwise looks good.

sharat87 (approved):

Issue #21499, opened 2023-03-17
[Task]: Provide a way for plugins to opt out of airgap Appsmith version


abhvsn:
## Description
Some of the Appsmith supported plugins like redshift, firestore, etc. for which self-hosted versions are not available, and SaaS plugins depend on internet connectivity, we need to provide a scalable way for existing as well as plugins that will be introduced in the future to `opt out` of airgap version.

PR #22098, opened 2023-04-04
feat: Add functionality to opt out plugin from airgap instance

abhvsn:
## Description

As a part of supporting airgap instances we want to restrict the plugins which have any public dependency. This PR introduces a config setting for plugins to opt out of airgap. 
Also as a part of this exercise we are also adding a setting for CS dependency which can be utilised in future if our customers wants to opt out of CS dependent plugins.     
Corresponding EE PR: https://github.com/appsmithorg/appsmith-ee/pull/1258

> TL;DR: Provide a way for plugins supported in Appsmith to opt-out of airgap instances 

Fixes https://github.com/appsmithorg/appsmith/issues/21499

## Type of change

## How Has This Been Tested?

- Manual

### Dev activity

### QA activity:


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration005OptOutUnsupportedPluginsForAirGap.java
```diff
@@ -0,0 +1,62 @@
...
+        final Set<String> unsupportedPluginPackageNameInAirgap = Set.of(
+                SAAS_PLUGIN, RAPID_API_PLUGIN, FIRESTORE_PLUGIN, REDSHIFT_PLUGIN, DYNAMO_PLUGIN,
+                AMAZON_S3_PLUGIN, GOOGLE_SHEETS_PLUGIN
+        );
+
+        final Set<PluginType> cloudServicesDependentPluginTypes = Set.of(PluginType.SAAS, PluginType.REMOTE);
+
+        List<Plugin> plugins = mongoTemplate.findAll(Plugin.class);
+        for (Plugin plugin : plugins) {
+            if (unsupportedPluginPackageNameInAirgap.contains(plugin.getPackageName())
```
sumitsum, line 49: Hi Abhijeet, can you please check if we need the outer `if` clause i.e. if we completely remove the outer `if` clause even then the inner `if` clauses should produce the same result. 
Also, IMO, we could have written this as a update query as well so that we don't have to read the data in memory to update them, but since the list of plugins is usually a very small list I think it doesn't make much of a difference, so we could let this be.

abhvsn: @sumitsum outer if part is used to check if the mongo op is really required. If the plugin doesn't falls under any condition then save operation is not performed. 
About the update part I agree but again I thought the same, it doesn't make any difference considering the collection size and document size is pretty low.

sumitsum (approved):

PR #22103, opened 2023-04-04
fix: Fix annoying issue with Entity form control in google sheets

### LINKED ISSUE #21700: [Feature]: In Gsheet datasource, The entity dropdown in query creation should default to a single enabled option if all other options are disabled

### Summary

In google sheet query creation, when I select operation as `insert many`, `update one`, and `update many`, in the entity dropdown only sheet rows option is enabled by default, other options of spreadsheet and sheet are disabled, Since a single option is enabled for these operations we can default the dropdown to this option for above three operations. 

When I create gsheet query, for some operations, only single value is enabled in entity dropdown, for such operations I should be able to see that single option value by default in the dropdown, so that I do not have to waste my time into opening the dropdown and looking at other option values which are disabled and I can move ahead with query creation faster

Irongade:
In google sheet query creation, when I select operation as insert many, update one, and update many, in the entity dropdown only sheet rows option is enabled by default, other options of spreadsheet and sheet are disabled, Since a single option is enabled for these operations we can default the dropdown to this option for above three operations. This PR fixes that.

Fixes #21700 

- Manual

### Dev activity

### QA activity:

### DIFF
--- a/app/client/src/components/formControls/DropDownControl.tsx
+++ b/app/client/src/components/formControls/DropDownControl.tsx
@@ -211,7 +211,7 @@ function renderDropdown(
       }
 
       const isOptionDynamic = options.some((opt) => "disabled" in opt);
-      if (isOptionDynamic) {
+      if (isOptionDynamic && !!props?.isRequired) {
         const isCurrentOptionDisabled = options.some(
           (opt) => opt?.value === selectedValue && opt.disabled,
         );
--- a/app/client/src/components/formControls/DropDownControl.tsx
+++ b/app/client/src/components/formControls/DropDownControl.tsx
@@ -209,6 +209,18 @@ function renderDropdown(
         selectedValue = tempValues;
         props.input?.onChange(tempValues);
       }
+
+      const isOptionDynamic = options.some((opt) => "disabled" in opt);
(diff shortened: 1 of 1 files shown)


app/client/src/components/formControls/DropDownControl.tsx
```diff
@@ -209,6 +209,22 @@ function renderDropdown(
         selectedValue = tempValues;
         props.input?.onChange(tempValues);
       }
+
+      const isOptionDynamic = options.some((opt) => "disabled" in opt);
+      if (isOptionDynamic) {
+        const isCurrentOptionDisabled = options.some(
```
sneha122, line 215: @Irongade Since this code is added in dropdowncontrol component, which is common form control and this change will be applicable to wherever we use dropdown control and if they have disabled property, In google sheets case, since entity field is mandatory it makes sense, but what if the field is optional, in that case we would still be setting it to some default value, we should not do that right?

sneha122: Please check conversation: https://theappsmith.slack.com/archives/CNQ9Q91C0/p1679642422610159

sneha122 (approved):

PR #22177, opened 2023-04-07
ci: Use vars for DOCKER_HUB_ORGANIZATION

sharat87:
GitHub's Variables, unlike Secrets, don't get masked in the output, and are ideal for non-secret... _variables_. I'm switching on such secret here, and depending on our experience with this, we'll look to moving more.

Of course, goes without saying, do NOT use these variables for secrets. When in doubt, use Secrets.

[Learn more about Variables](https://docs.github.com/en/actions/learn-github-actions/variables).

- I'd like approvals from all three of you before I merge this please.
- Please approve only, don't merge. I'll merge this _after_ we tag the upcoming version.

### DIFF
--- a/.github/workflows/github-release.yml
+++ b/.github/workflows/github-release.yml
@@ -237,7 +237,7 @@ jobs:
           context: app/client
           push: true
           tags: |
-            ${{ secrets.DOCKER_HUB_ORGANIZATION }}/appsmith-editor:${{needs.prelude.outputs.tag}}
+            ${{ vars.DOCKER_HUB_ORGANIZATION }}/appsmith-editor:${{needs.prelude.outputs.tag}}
 
       # Only build & tag with latest if the tag doesn't contain beta
       - name: Build and push client image latest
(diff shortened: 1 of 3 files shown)


.github/workflows/github-release.yml
```diff
@@ -237,7 +237,7 @@ jobs:
           context: app/client
           push: true
           tags: |
-            ${{ secrets.DOCKER_HUB_ORGANIZATION }}/appsmith-editor:${{needs.prelude.outputs.tag}}
+            ${{ vars.DOCKER_HUB_ORGANIZATION }}/appsmith-editor:${{needs.prelude.outputs.tag}}
```
Aishwarya-U-R, line 240: We will also remove the variable from `Secrets` right? Since it's moved to Variables?

sharat87: Yes, but I want to see this working for a week or so, before removing it from there. 🙂

Aishwarya-U-R (approved):

pratapaprasanna (approved):

Issue #21796, opened 2023-03-27
[Bug]: Missing tolerations and node selector in helm charts


ErlendFax:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

When we set the `nodeSelector` and `tolerations` in the `values.yaml` file, the config is only applied to the Appsmith Statefulset. This should also be applied to _all_ relevant resources that the helm chart creates, for example the MongoDB, and Redis.

Not a helm expert but a **solution** would be something like adding this:

```
{{- with .Values.nodeSelector }}
      nodeSelector: {{- toYaml . | nindent 8 }}
    {{- end }}
    {{- with .Values.tolerations }}
      tolerations: {{- toYaml . | nindent 8 }}
    {{- end }}
```
... to all resources that will be running on a node.

_We have unfortunately needed to take down appsmith because of a performance issue [(see this)](https://github.com/appsmithorg/appsmith/issues/21622), and we thought placing all Appsmith stuff in a different node pool would be a workaround, but this does not work either as described above._

### Steps To Reproduce

1. Create a k8s node pool in GKE/AWS, with a "taint"
2. Add a `nodeSelector` that match the node you just made and `tolerations` to match the "taint", and add them to `values.yaml`
3. Run `helm install ...`

Appsmith Statefulset will be installed on the correct node, while the others instances (like MongoDB and Redis, etc) will be assign to another node (if any).

### Environment

Production

### Version

Self-hosted, helm - latest

ssichynskyi:
The issue is not fully fixed and has to be reopened. @pratapaprasanna @Nikhil-Nandagopal 
Issue describes missing node selector AND tolerations. The submitted MR fixed only node selector but not tolerations.

pratapaprasanna:
We shall address it in a different PR and new helm chats release

ssichynskyi:
Perfect, thanks @pratapaprasanna !

PR #22195, opened 2023-04-10
helm: add node selectors for appsmith sub-charts

pratapaprasanna:
Fixes: [21796](https://github.com/appsmithorg/appsmith/issues/21796)


sharat87 (changes requested):
Should we also increment the Helm Chart version?

sharat87:
@pratapaprasanna, can we take this forward if we still need it please?

sharat87 (approved):

PR #22239, opened 2023-04-10
fix: Remove devDependencies from RTS build

ChandanBalajiBP:
> Pull Request Template
>

## Description

> and context. List any dependencies that are required for this change
> Links to Notion, Figma or any other documents that might be relevant to the PR

Fixes # (issue)

Media

## Type of change

- Manual
- Jest
- Cypress

### Dev activity

### QA activity:


app/rts/build.sh
```diff
@@ -4,12 +4,22 @@ set -o errexit
 
 cd "$(dirname "$0")"
 rm -rf dist/
-yarn install --frozen-lockfile
+# Install node modules if not already installed
+# This is required for the first time build as node_modules is not present in the image
+! test -d "/node_modules" && yarn install --frozen-lockfile
```
sharat87, line 9: This test will always be `false`, since it's looking for a `node_modules` folder in the root of the filesystem. This is equivalent to just `yarn install --frozen-lockfile`, which is what we want here, right? Irrespective of whether a `node_modules` already exists or not? 🤔

```suggestion
yarn install --frozen-lockfile
```

app/rts/build.sh
```diff
@@ -4,12 +4,22 @@ set -o errexit
...
 npx tsc && npx tsc-alias
+# Keep copy of all dependencies in node_modules_bkp
+mv node_modules node_modules_bkp
+# Install only production dependencies
+yarn install --production --frozen-lockfile
 tsc_exit_code=$?
```
sharat87, line 15: This gets the exit code of the last command, which was `tsc-alias`, but the last command has changed. In fact, this is actually not needed, let's remove this line, as well as the last `exit $tsc_exit_code` line. Neither of them are needed.

sharat87 (changes requested):
Thanks @ChandanBalajiBP, looks good, just a few minor changes and we should be ready to push this. Also, two more questions please:

1. Can you confirm that the `formidable` package is not present in the `node_modules` when built this way. I'm sure you checked, just double-confirming.
2. We've made some changes to the shared AST `package.json`. Does the client build have any impact too? Is that build working fine?

app/rts/build.sh
ChandanBalajiBP: Right, i have updated.

app/rts/build.sh
ChandanBalajiBP: Updated

sharat87 (dismissed):

ChandanBalajiBP:
Adding test plan approved as the Ci has passed and from the build folder there are no dev dependencies.

sharat87 (approved):

PR #22321, opened 2023-04-12
feat: adding email property in google sheets

### LINKED ISSUE #21902: [Task]:  Retrieve email Id of the google account for which the google sheet datasource is being created

### SubTasks

In the google sheet datasource review page, we should be able to see the account email id from which the datasource is getting authorised

### Task
- Find a datasourceConfiguration field where we should store the emailId
- Fetch the account email id from a designated endpoint. (at what stage of the flow this should be done is a question)
- save it in the datasource.
- write a migration for the existing google sheet datasources to fetch the account id and update the datasource.

sondermanish:
## Description
As a part of google sheet scope limitation, the email address is being added in datasource configuration.  The email address 
is being stored in a property, an element in properties (`List<Property>`) with key: 'emailAddress' and value: (<emailId>) .

In order to retrieve this property a GET call to endpoint `https://www.googleapis.com/drive/v3/about?fields=user` is being made.

This property is being saved while google sheet datasource gets authorised, after access toke retrieval from cloud services.
in the endpoint `/saas/{datasourceId}/token`.

Since the flow is specific to google sheets, the main code is implemented in googlesheetPluginExecutor. The flow has three steps

- Getting access token from datasourceConfiguration
- making the call to googleApis and obtaining the email address
- saving it back into datasource configuration

The authorisation flow itself saves the datasource later on.  

> TL;DR Adding emailAddress in properties of datasource configuration when google sheets is authorised.

Fixes #21902 

## This has been tested
- Manual
- JUnit

### Dev activity


app/server/appsmith-plugins/googleSheetsPlugin/src/test/java/com/external/config/GetDatasourceMetadataMethodTest.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.List;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.spy;
+
+public class GetDatasourceMetadataMethodTest {
+
+    @Test
+    public void DsConfigPropertiesWithNoEmailTest() {
```
sumitsum, line 21: Can you please rename both test names to something that indicates in a more understandable way what the test is doing - e.g. something along the lines of `testGetUserEmailWhenPropertiesListIsEmpty()` ?

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/AuthenticationServiceCEImpl.java
```diff
@@ -457,7 +463,22 @@ public Mono<OAuthResponseDTO> getAccessTokenFromCloud(String datasourceId, Strin
                                             .getAuthentication()
                                             .setAuthenticationStatus(AuthenticationDTO.AuthenticationStatus.SUCCESS);
                                 }
-                                return Mono.zip(Mono.just(datasource), Mono.just(accessToken), Mono.just(projectID));
+
+                                Mono<String> accessTokenMono = Mono.just(accessToken);
+                                Mono<String> projectIdMono = Mono.just(projectID);
+
+                                return featureFlagService.check(FeatureFlagEnum.LIMITING_GOOGLE_SHEET_ACCESS)
```
sumitsum, line 470: Minor: Should we instead wrap the feature flag specific code inside the `if` clause and leave the default implementation out ? This should be useful when say another feature flag specific change needs to be added here, in which case we can use another `if` clause so overall it will look like: 
```
if (featureFlag_1) {...}
if (featureFlag_2) {...}
if (featureFlag_3) {...}
default implementation
```

sumitsum (changes requested):

app/server/appsmith-plugins/googleSheetsPlugin/src/test/java/com/external/config/GetDatasourceMetadataMethodTest.java
```diff
@@ -0,0 +1,36 @@
...
+        List<Property> properties = new ArrayList<>();
+        String emailAddress = "mockEmailAddress";
+        List<Property> returnedProperties= GetDatasourceMetadataMethod.setPropertiesWithEmailAddress(properties, emailAddress);
+        assert(!returnedProperties.isEmpty());
+        assert(returnedProperties.get(0).getKey().equals(FieldName.EMAIL_ADDRESS));
+        assert(returnedProperties.get(0).getValue().equals(emailAddress));
+    }
+
+    @Test
+    public void verifyEmailUsageSamePropertyObjectForStorage() {
```
sumitsum, line 23: Hi Manish, this is something very minor, hence it is not a binding requirement, I would still approve the code, however, I just want to mention that the name this method name is still not understandable at all. e.g. could this be a more suggestive name : `verifyIfOldEmailGetsReplacedWhenUpdatedWithNewEmail`

sumitsum (approved):

sumitsum (approved):

nidhi-nair (approved):

PR #22406, opened 2023-04-14
[Blocked] fix: Checking for stale connection when Oauth token expires (#21769)

### LINKED ISSUE #21769: [Bug]: OAuth 2.0 token is NOT refreshed

### Description

We are not authenticating a data source in this case with Zoho. Zoho provides a token that expires approximately an hour after it is obtained. Once the token expires, Appsmith does not automatically refresh this token, which causes the API to fail.

### Steps To Reproduce

1.  Enter the [[Zoho development console](https://api-console.zoho.com/add)](https://api-console.zoho.com/add).
2. Create a new client and use the client called 'Server-based Applications'.
3. Get your Homepage URL in Appsmith.
Configure the application as follows:
```yml
Client Name: <<Appname>>
Homepage URL: <<YourAppsmithApplicationURL>>
Authorized Redirect URIs: https://app.appsmith.com/api/v1/datasources/authorize
```

4. Press the `CREATE` button.
5. After creating your application, you will obtain your Client ID and Client Secret credentials on a dashboard like this.
6. Add a new data source of `Authenticated API` type
7. Data source configuration
8. 
```YAML
URL: https://expense.zoho.com/
Authentication Type: OAuth 2.0
Grant Type: Authorization Code
Add Access Token To: Request Header
Header Prefix: Bearer
Access Token URL: https://accounts.zoho.com/oauth/v2/token
Client ID: <<YourClientID>>
Client Secret: <<YourClientSecret>>
Scope(s): ZohoExpense.fullaccess.ALL
Client Authentication: Send as Basic Auth header
Authorization URL: https://accounts.zoho.com/oauth/v2/auth
```
9. Press the `Save and Authorize` button.
10. Verify that everything is working by running this query: `api/v1/organizations` from this data source using this endpoint. You should receive this output if everything went well.

After an hour and a half the token will expire and appsmith will not refresh it

### Public Sample App

https://app.appsmith.com/app/cnv-y3rxhu4/page1-641a223433970d1990c88202

### Environment

Production

### Issue video log

https://www.loom.com/share/22988ee151e7486396caf325a15aa67e

### Version

Appsmith v1.9.12-SNAPSHOT

Front logo [Front conversations](https://app.frontapp.com/open/top_9gklo)

sanveer-osahan:
## Description
For checking if the connection is stale or not for authenticated api, only the datasource `updatedAt` field was checked.
If `updatedAt` time is greater than connection `creationTime`, the connection is considered stale.
Added an additional check for Oauth token expiry. If token is expired, the connection will be considered stale and a new connection will be created using the `refresh_token` 

Fixes #21769

## Type of change

## How Has This Been Tested?

- Manual

### Dev activity

### QA activity:

### DIFF
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/AuthenticationDTO.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/AuthenticationDTO.java
@@ -2,6 +2,7 @@ package com.appsmith.external.models;
 
 import com.appsmith.external.constants.Authentication;
 import com.appsmith.external.views.Views;
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonSubTypes;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.annotation.JsonView;
@@ -58,6 +59,11 @@ public class AuthenticationDTO implements AppsmithDomain {
     @JsonView(Views.Public.class)
     public Mono<Boolean> hasExpired() {
         return Mono.just(Boolean.FALSE);
+    }
+
+    @JsonIgnore
+    public boolean isExpired() {
+        return false;
     }
 
 }
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/AuthenticationDTO.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/AuthenticationDTO.java
@@ -61,6 +61,12 @@ public class AuthenticationDTO implements AppsmithDomain {
         return Mono.just(Boolean.FALSE);
     }
 
+    /**
+     * This function has been added to check if Auth token is expired.
+     * By default, it is assume that the token is not expired.
+     * Override to provide custom check for token expiry.
+     * @return true if auth token is expired, else false
+     */
     @JsonIgnore
     public boolean isExpired() {
         return false;
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/OAuth2.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/OAuth2.java
@@ -109,4 +109,11 @@ public class OAuth2 extends AuthenticationDTO {
 
         return Mono.just(authenticationResponse.expiresAt.isBefore(Instant.now().plusSeconds(60)));
     }
+
+    @Override
+    public boolean isExpired() {
+        return this.authenticationResponse == null
+                || this.authenticationResponse.expiresAt == null
+                || this.authenticationResponse.expiresAt.isBefore(Instant.now().plusSeconds(60));
+    }
 }
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/OAuth2.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/OAuth2.java
@@ -112,8 +112,8 @@ public class OAuth2 extends AuthenticationDTO {
 
     @Override
     public boolean isExpired() {
-        return this.authenticationResponse == null
-                || this.authenticationResponse.expiresAt == null
-                || this.authenticationResponse.expiresAt.isBefore(Instant.now().plusSeconds(60));
+        return this.authenticationResponse != null
+                && this.authenticationResponse.expiresAt != null
+                && this.authenticationResponse.expiresAt.isBefore(Instant.now().plusSeconds(60));
     }
 }
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/plugins/BaseRestApiPluginExecutor.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/plugins/BaseRestApiPluginExecutor.java
@@ -79,6 +79,11 @@ public class BaseRestApiPluginExecutor implements PluginExecutor<APIConnection>,
     }
 
     @Override
+    public boolean isConnectionStale(DatasourceConfiguration datasourceConfiguration) {
+        return datasourceConfiguration.getAuthentication().isExpired();
+    }
+
+    @Override
     public Mono<DatasourceTestResult> testDatasource(DatasourceConfiguration datasourceConfiguration) {
         // At this point, the URL can be invalid because of mustache template keys inside it. Hence, connecting to
         // and verifying the URL isn't feasible. Since validation happens just before testing, and since validation
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/plugins/BaseRestApiPluginExecutor.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/plugins/BaseRestApiPluginExecutor.java
@@ -79,7 +79,7 @@ public class BaseRestApiPluginExecutor implements PluginExecutor<APIConnection>,
     }
 
     @Override
-    public boolean isConnectionStale(DatasourceConfiguration datasourceConfiguration) {
+    public boolean isConnectionStale(APIConnection connection, DatasourceConfiguration datasourceConfiguration) {
         return datasourceConfiguration.getAuthentication().isExpired();
     }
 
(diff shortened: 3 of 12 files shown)


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/AuthenticationDTO.java
```diff
@@ -60,4 +61,9 @@ public Mono<Boolean> hasExpired() {
         return Mono.just(Boolean.FALSE);
     }
 
+    @JsonIgnore
+    public boolean isExpired() {
```
sumitsum, line 65: Can you please add a comment here explaining that this method has primarily been added to check if OAuth token has expired ?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/OAuth2.java
```diff
@@ -109,4 +109,11 @@ public Mono<Boolean> hasExpired() {
 
         return Mono.just(authenticationResponse.expiresAt.isBefore(Instant.now().plusSeconds(60)));
     }
+
+    @Override
+    public boolean isExpired() {
+        return this.authenticationResponse == null
```
sumitsum, line 115: Is there any reason for adding the first 2 conditions checking for null ? IMO, the first condition implies a failure of authentication and the second null check implies that the authentication didn't have refresh scope.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/OAuth2.java
```diff
@@ -109,4 +109,11 @@ public Mono<Boolean> hasExpired() {
...
+                || this.authenticationResponse.expiresAt == null
```
sumitsum, line 116: Can you please refactor the `OAuth2ClientCredentials.java` and `OAuth2AuthorizationCode.java` files and replace with this function anywhere a check for expiry is being done ?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/plugins/BaseRestApiPluginExecutor.java
```diff
@@ -78,6 +78,11 @@ public Set<String> validateDatasource(DatasourceConfiguration datasourceConfigur
         return datasourceUtils.validateDatasource(datasourceConfiguration);
     }
 
+    @Override
+    public boolean isConnectionStale(DatasourceConfiguration datasourceConfiguration) {
```
sumitsum, line 82: Can you please change the method name to `isAuthenticationExpired` ? One reason being that `isConnectionStale` implies that we are checking the connection object, but here we are only passing the datasource config as parameter, not the actual connection object or datasourceContext object. Another reason being that in some datasources like MySQL we do a stale connection check during the action execution code flow and this function name might confuse someone with that flow. Please mention if you think otherwise.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
```diff
@@ -212,6 +212,15 @@ public boolean getIsStale(Datasource datasource, DatasourceContextIdentifier dat
                 && datasource.getUpdatedAt().isAfter(datasourceContextMap.get(datasourceContextIdentifier).getCreationTime());
     }
 
+    public boolean getIsStale(
+            Datasource datasource, DatasourceContextIdentifier datasourceContextIdentifier, PluginExecutor<Object> pluginExecutor) {
+        DatasourceContext<?> datasourceContext = datasourceContextMap.get(datasourceContextIdentifier);
+        return getIsStale(datasource, datasourceContextIdentifier)
```
sumitsum, line 218: QQ: Could we only have one implementation / definition of the method ?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
```diff
@@ -212,6 +212,15 @@ public boolean getIsStale(Datasource datasource, DatasourceContextIdentifier dat
...
+                || (datasource.getId() != null
+                    && datasourceContext != null
+                    && pluginExecutor.isConnectionStale(datasource.getDatasourceConfiguration()));
```
sumitsum, line 221: Also, as discussed, please add a JUnit TC for this. IMO, adding a TC for the `getStale` method should be ok here. One way to do this would be to create and configure a new datasource config object within the test case and pass the same to the getStale method. We should be able to get rest api plugin executor object via :
```
plugin = pluginRepository.findByPackageName("restapi-plugin").block()
// now use  pluginExecutorHelper.getPluginExecutor(...) to get pluginExecutor object using plugin id.
```

sumitsum (changes requested):

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/plugins/BaseRestApiPluginExecutor.java
sanveer-osahan: Initially I was using the connection object as parameter, but we connected and checked that the datasource config contains the info when the token will be expired.
I can change the method name to `isAuthenticationExpired` but will it be still relevant to other plugins? As all plugins has connections but not all plugins has authentication.
Alternatively can I also pass the connection object along with datasource config in this method?
It might not be used for Rest API but may be used in case of other plugins

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/OAuth2.java
sanveer-osahan: I have changed it to the following as per the requirement mentioned in below comments for `OAuth2ClientCredentials.java` and `OAuth2AuthorizationCode.java`

```
return this.authenticationResponse != null
                && this.authenticationResponse.expiresAt != null
                && this.authenticationResponse.expiresAt.isBefore(Instant.now().plusSeconds(60));

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/OAuth2.java
sanveer-osahan: Done

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
sanveer-osahan: This can be done. But now it has a high impact where any areas of code calling this are affected. I have made those changes.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/AuthenticationDTO.java
sanveer-osahan: Done

nidhi-nair:
QQ: Wa the `UpdatableConnection` logic not taking care of this?

sanveer-osahan:
> QQ: Was the `UpdatableConnection` logic not taking care of this?

Hi @nidhi-nair I tried searching in the codebase if refreshing of expired token is already handled in `UpdateableConnection`. There is only one place were I found the code for refresh - [`AuthenticationServiceCEImpl`](https://github.com/appsmithorg/appsmith/blob/release/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/AuthenticationServiceCEImpl.java) 

But here the refresh is only for SAAS and Remote PluginTypes

And `UpdateableConnection` only comes into picture when connection is identified as stale in [`DatasourceContextServiceCEImpl`](https://github.com/appsmithorg/appsmith/blob/release/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java). Here there is a check if connection is instance of `UpdateableConnection` and then updated.
But the stale logic doesn't handle expiry of token i

sanveer-osahan:
Hi @nidhi-nair, thanks for your input. 
It's clear now that in some apis, there are additional params required to fetch the refresh_token in response. 
If they are provided in datasource config, the refresh logic is working as expected without any change in the code.

PR #22418, opened 2023-04-14
fix: List v2 child widgets event defined with action selector

### LINKED ISSUE #22373: [Bug]: For nested list, the onItemClick action is throwing Syntax error

### Description

https://www.loom.com/share/f28977e6ea044fee9c0e90cecc77ddc2

### Steps To Reproduce

1. DnD a list and drop another child List
2. Configure the onItemClick of the child list
3. Click on child item
### Public Sample App

_No response_

### Environment

Production

### Issue video log

_No response_

### Version

Release

ashit-rath:
## Description

TL;DR
For list widget's child widget, when events like `onClick` were defined using the action selector, it throws syntax error.

Reason:
Underneath the hood the new action selector by default applies a semi-colon (`;`) at the end of every action selected. This does not play well with the List v2 as the widget adds a wrapper on top of this action which leads to a syntax error.

Fix:
 For trigger paths the binding is ow enclosed with `{ }`
 where as the binding paths are not enclosed with `{ }`

 Example
```js
 (() => { showAlert("Hello") })() // For trigger paths
 (() => currentItem.name )() // For binding paths
```

 It is expected for binding paths to return a value from the binding so to make it return by default js binding cannot be wrapped around `{ }`.

In case of trigger paths, it is not expected for the js binding to return anything but only execute other actions. When wrapped around `{ }`, it gives an advantage when action selectors are used to define a trigger/event. Action selectors by default adds a semi-colon(;) at the end of every action. If the wrapper `{}` is not present then a binding
```
  (() => showAlert("Hello"); )()

```
  would throw an error as the addition of a semi-colon is an invalid JavaScript syntax.
  Hence we convert the above IIFE to 
```
(() => { showAlert("Hello"); } )()
```
 to make the semi-colon work.

Fixes #22373

## Type of change

## How Has This Been Tested?
- Cypress

### Dev activity

### QA activity:
- [x] Cypress test cases have been added and approved by either SDET or manual QA
- [x] Added Test Plan Approved label after reveiwing all Cypress test


Tooluloope (dismissed):

kamakshibhat-appsmith:
Looks good @ashit-rath

Tooluloope (approved):

Issue #6325, opened 2021-08-02
[Bug]-[3000]:Page must scroll with cursor movement


ramsaptami:
## Summary
When user is editing a data source query and presses cursor to move to the next line, then page must scroll to show the line the cursor is on.

[![LOOM DEMO](http://cdn.loom.com/sessions/thumbnails/ab5df7ccfba94f7a9670de8178c3cb06-00001.gif)](https://www.loom.com/share/ab5df7ccfba94f7a9670de8178c3cb06)

BhandarkarPawan:
I'd like to work on this. Can it be assigned?

ramsaptami:
@BhandarkarPawan it's all yours. 
Don't forget to read the [contributor guidelines](https://github.com/appsmithorg/appsmith/blob/769ee31f87cfe7866e5cd87a2c2df9da48f778cc/CONTRIBUTING.md) before working on this issue. Good luck!

Nikhil-Nandagopal:
This happens in the property pane as well so tagging it as design system

kocharrahul7:
## Stats
| Stat | Values |
| ------------- | ------------- |
| Reach     | 6000 |
| Effort (months)  | 1 |

ginilpg:
Issue still persists with query pane and property pane. Need to prioritise this @kocharrahul7

eco-monk:
Notes: 
Bottom tabs have been moved to design system now. Might need changes in design system for this.
For property pane, the content/style tabs on top is the overly that needs to be checked.

royalpinto007:
Hello, can you please assign this issue to me?

ramsaptami:
Hey @royalpinto007 , the fix here requires a minor design change so it would be best if you could connect with the team on this to help guide you through the changes. Our engineer @eco-monk can help you through this.

royalpinto007:
Yes, it would be really great if the file due to which the design problem happens is mentioned. It would help me solve quickly.

eco-monk:
Hey @royalpinto007, Thank you so much for your interest.!
I am afraid, this is not a straight css fix.

Because we generate the query pane though a configuration object. 
We'd need to ensure other pages are not affected when making this change.

Removing tag, I'd suggest you look for other issues tagged `Good first issue`.

eco-monk:
Created a separate issue for property pane. #19293

Issue #19293, opened 2022-12-29
[Bug]: Property pane should scroll with cursor movement


eco-monk:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

https://www.loom.com/share/de9fb7deeaca486b8ab876f4ff449de5

### Steps To Reproduce

Add multiple lines in any code editor.
Move cursor to the top.
Notice the pane doesn't scroll.

### Version

Appsmith v1.8.14-SNAPSHOT

PR #22494, opened 2023-04-18
fix: codeditor scroll cursor into view 

akash-codemonk:
## Description

Was able to repro the issue on an isolated project where codemirror has auto height and has a scrollable container which does not take the full height of the viewport. Apparently scrolling for this scenario isn't handled in codemirror, related issue https://github.com/codemirror/codemirror5/issues/1104. https://github.com/codemirror/codemirror5/issues/2912#issuecomment-62161976 led me to figure out a solution using `scrollCursorIntoView` event.  
The original source of the solution is from https://github.com/WebKit/WebKit-http/blob/817c46e152af795d735678386db68805d0aa505e/Source/WebInspectorUI/UserInterface/Views/CodeMirrorAdditions.js#L354 .

The solution here partially works in the property pane while scrolling down, but the issue remains while scrolling up. The property pane is wrapped with a position static wrapper and has some sticky headers which might be causing some issues. The fix for the property pane will have to be figured out separately in https://github.com/appsmithorg/appsmith/issues/19293

The solution involves scrolling to the cursor(which is an html element). We use the scroll library to only scroll when the target element is out of view.

Fixes https://github.com/appsmithorg/appsmith/issues/6325

DP: https://appsmith-7fo141hgf-get-appsmith.vercel.app/

Media

https://user-images.githubusercontent.com/67054171/233363358-8622b59d-d573-451b-ab84-22181395f394.mov

## Type of change

## How Has This Been Tested?

- Manual

### Dev activity

### QA activity:


Richarex:
Tested the bug fix on the DP url and it looks good to me.

hetunandu (approved):

PR #22587, opened 2023-04-20
fix: Supervisord event listener health check exception Handling

sum35h:
Event Listener Python script fixes:
- Added exception handling in health check request
- Updated backend health check point to `http://localhost:8080/api/vi/health`

### DIFF
--- a/deploy/docker/scripts/supervisor_event_listener.py
+++ b/deploy/docker/scripts/supervisor_event_listener.py
@@ -20,16 +20,19 @@ def write_stderr(s):
 def wait_until_backend_healthy():
     sleep_sec = 3
     timeout_sec = 120
-    for _ in range(timeout_sec//sleep_sec):
-        if requests.get(BACKEND_HEALTH_ENDPOINT).ok:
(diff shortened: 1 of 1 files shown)


sharat87 (dismissed):

deploy/docker/scripts/supervisor_event_listener.py
```diff
@@ -21,16 +22,21 @@ def wait_until_backend_healthy():
     sleep_sec = 3
     timeout_sec = 120
     for _ in range(timeout_sec//sleep_sec):
-        if requests.get(BACKEND_HEALTH_ENDPOINT).ok:
-            write_stderr('\nBackend is healthy\n')
+        try:
+            if requests.get(BACKEND_HEALTH_ENDPOINT).ok:
+                write_stderr('\nBackend is healthy\n')
+                break
+        except ConnectionError:
```
sum35h, line 29: Now that the request is not going through nginx, requests throws connection refused error instead of 502.
Handling that here.

sharat87 (approved):

PR #22632, opened 2023-04-21
fix: Error handling added with proper message when try to create application with duplicate name

### LINKED ISSUE #7401: [Bug] Change the error message for duplicate app name in collaborated org
## Description
When multiple users are trying to create an app on a common organization, the changes for newly added app is not real time hence there's a conflict on app name causing an error. This error asks user to contact Appsmith support, but a simple page refresh loads the new app. Ideal scenario would be to have newly added app changes to reflect real time, but in the absence, the error message can be changed to inform the user of a page refresh instead of contacting support.

1. Share organization with multiple users and create an app from User 1
2. Now try to create an app from User 2 simultaneously and observe the error message displayed (as seen in screenshot)

### Important Details
- Version: Cloud
- OS: Win10
- Browser: Chrome
- Environment: Release

image

abhiappsmith:
## Description

>TL;DR: Added suffix retry if name clashes on creating an application.

## Issue: 
When multiple users are trying to create an app on a common organization, the changes for newly added app is not real-time hence there's a conflict on app name causing an error. This error asks user to contact Appsmith support, but a simple page refresh loads the new app. Issue: #7401 

## Solution: 
Added suffix retry if name clashes on creating an application. With this change, the above issue will not happen and in case of name conflict new application will be created with a suffix added in the name. E.g UntitledApplication 1 (1).

## Changes:

1. Added retry suffix method in the ApplicationServiceCEImpl for name clash only.
2. Renamed createDefault method to createDefaultApplication for better readability.
3. Removed AppsmithError and AppsmithErrorCode. Not required with current change.
4. Updated cloneApplicationDocument method in ExamplesWorkspaceClonerCEImpl to directly call createDefaultApplication with contains suffix retry logic.
5. Removed same retry suffix method which is used in ExamplesWorkspaceClonerCEImpl. Not required.
6. Updated test case for this change.

Fixes #7401

## Type of change

## How Has This Been Tested?

- Manual :  Tested by using backend api.
- JUnit

### Dev activity

### QA activity:


app/server/appsmith-server/src/main/java/com/appsmith/server/exceptions/AppsmithErrorCode.java
```diff
@@ -108,6 +108,7 @@ public enum AppsmithErrorCode {
     TOO_MANY_REQUESTS("AE-TMR-4029", "Too many requests"),
     USER_NOT_FOUND("AE-USR-4004", "User not found"),
     CSRF_TOKEN_INVALID("AE-APP-4039", "CSRF token missing/invalid"),
+    DUPLICATE_KEY_OBJECT_CREATION("AE-APP-5043", "Duplicate key object creation"),
```
mohanarpit, line 111: For my understanding, how did we come up with the error code, 5043? Is there a heuristic we are following here?

app/server/appsmith-server/src/main/java/com/appsmith/server/exceptions/AppsmithError.java
```diff
@@ -71,6 +71,8 @@ public enum AppsmithError {
     UNAUTHORIZED_ACCESS(403, AppsmithErrorCode.UNAUTHORIZED_ACCESS.getCode(), "Unauthorized access", AppsmithErrorAction.DEFAULT, "Unauthorized access", ErrorType.AUTHENTICATION_ERROR, null),
     DUPLICATE_KEY(409, AppsmithErrorCode.DUPLICATE_KEY.getCode(), "Duplicate key error: An object with the name {0} already exists. Please use a different name or reach out to Appsmith customer support to resolve this.",
             AppsmithErrorAction.DEFAULT, "Duplicate key", ErrorType.BAD_REQUEST, null),
+    DUPLICATE_KEY_OBJECT_CREATION(409, AppsmithErrorCode.DUPLICATE_KEY_OBJECT_CREATION.getCode(), "Duplicate key error: An object with the name {0} already exists. Please refresh the page and try again",
```
mohanarpit, line 74: Sorry, I'm a little confused. What is the difference between this error message and the one above this? Why do we have duplicate error messages and codes here?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationServiceCEImpl.java
```diff
@@ -223,7 +223,9 @@ public Mono<Application> createDefault(Application application) {
         if (!StringUtils.hasLength(application.getColor())) {
             application.setColor(getRandomAppCardColor());
         }
-        return super.create(application);
+        return super.create(application).onErrorMap(DuplicateKeyException.class, error ->
```
mohanarpit, line 226: Please format.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationServiceCEImpl.java
```diff
@@ -223,7 +223,9 @@ public Mono<Application> createDefault(Application application) {
...
+                new AppsmithException(AppsmithError.DUPLICATE_KEY_OBJECT_CREATION,
```
mohanarpit, line 227: Is application name conflict the only duplicate key error possible? I'm asking because we are catching all DuplicateKeyExceptions and converting it into an AppsmithException.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationServiceCEImpl.java
abhiappsmith: Will do. Are using any code formatter?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationServiceCEImpl.java
abhiappsmith: Yes, correct. Will change this.

abhiappsmith: Done.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationServiceCEImpl.java
abhiappsmith: Done

app/server/appsmith-server/src/main/java/com/appsmith/server/exceptions/AppsmithError.java
mohanarpit: If `DUPLICATE_KEY_OBJECT_CREATION` is only going to be used for duplicate application name, why not name the key and the description message to be more specific? The error message displayed to the user is near identical to the previous one. In fact it has lesser information than the previous error message to help the user understand how to fix the problem.

app/server/appsmith-server/src/main/java/com/appsmith/server/exceptions/AppsmithErrorCode.java
mohanarpit: Minor: Is it possible for us to group related error message codes together? It makes discovery and addition of new error codes much easier than simply adding it to the end of the file in no particular order.

app/server/appsmith-server/src/main/java/com/appsmith/server/exceptions/AppsmithError.java
abhiappsmith: Sorry, I don't fully understand your above comment. 
1. DUPLICATE_KEY_OBJECT_CREATION is currently used when creating an application but can be used in other places. It can be named specifically but then can't be used other than the application name. 
2. The error message displayed to the user is nearly identical but the action item is totally different. Now user knows what to do and proceed further. Please let me know if that is not the expected outcome.

app/server/appsmith-server/src/main/java/com/appsmith/server/exceptions/AppsmithErrorCode.java
abhiappsmith: Yes, we can do the grouping. Just one thing you are talking about only my code or the whole file?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationServiceCEImpl.java
```diff
@@ -216,14 +216,43 @@ public Mono<Application> create(Application object) {
...
             application.setColor(getRandomAppCardColor());
         }
-        return super.create(application);
+        return super.create(application)
+                .onErrorResume(DuplicateKeyException.class, error -> {
+                    if (error.getMessage() != null
+                            // Catch only if error message contains workspace_application_deleted_gitApplicationMetadata_compound_index mongo error
+                            && error.getMessage().contains("workspace_application_deleted_gitApplicationMetadata_compound_index")) {
+                        // The duplicate key error is because of the `name` field.
+                        return createSuffixedApplication(application, name, suffix + 1);
```
mohanarpit, line 241: Is there an exit condition for this loop? From the looks of it, we are risking an infinite loop here.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationServiceCEImpl.java
```diff
@@ -216,14 +216,43 @@ public Mono<Application> create(Application object) {
...
+    /**
+     * Tries to create the given application with the name, over and over again with an incremented suffix, but **only**
+     * if the error is because of a name clash.
+     * @param application Application to create.
+     * @param name Name of the application, to which numbered suffixes will be appended.
+     * @param suffix Suffix used for appending, recursion artifact. Usually set to 0.
+     * @return A Mono that yields the created application.
+     */
+    private Mono<Application> createSuffixedApplication(Application application, String name, int suffix) {
+        final String actualName = name + (suffix == 0 ? "" : " (" + suffix + ")");
```
mohanarpit, line 228: Can we keep this suffix logic consistent with what the client does when creating a new application? Ideally, to the user, it shouldn't matter how the suffix got appended as long as the application was created.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ce/ApplicationServiceCETest.java
```diff
@@ -319,9 +318,13 @@ public void createApplicationWithNullName() {
...
-    public void createValidApplication() {
+
+    /**
+     * Create an application with name "ApplicationServiceTest TestApp" and validate it.
+     * @param applicationFinalName This is the application resultant name and it can be different from original name
+     *                             due to retry if there is name clash.
+     */
+    private void createAndVerifyValidApplication(String applicationFinalName){
         Application testApplication = new Application();
         testApplication.setName("ApplicationServiceTest TestApp");
```
mohanarpit, line 329: Does this test pass? I'm asking because we seem to be passing the `applicationFinalName` into the function but then also defining a particular name inside the test function. This can get a little confusing when writing a new test in the future.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationServiceCEImpl.java
abhiappsmith: I picked this suffix pattern which is used in other multiple places.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationServiceCEImpl.java
abhiappsmith: There are 2 exit conditions when the application is created successfully or it will throw an exception other than the duplicate name.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ce/ApplicationServiceCETest.java
abhiappsmith: Yes, this test passed. applicationName can be changed due to retry logic and that's why applicationFinalName is also passed to verify that logic. If it's expected applicationFinalName will not change then passed the same name here. You can see this in line no. 403, 410 and 414.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationServiceCEImpl.java
mohanarpit: I'm referring to the exit from the recursion. I understand that in the scenarios you mentioned, the loop will exit. I would highly recommend adding a max retry number. I know it's a corner case, but any retries should have a max retry limit by default. This ensures that we aren't caught by surprise in the future.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ce/ApplicationServiceCETest.java
mohanarpit: Sorry, I wasn't able to explain myself clearly. In the function `createAndVerifyValidApplication` we are using the base name "ApplicationServiceTest TestApp". In other test functions, `createApplicationWithDuplicateName` and `createValidApplication`, we have to use the exact same name as the base name. Else the test will fail. This means  we have created a coupling between 2 test functions and this is liable to break in the future. 

I'd highly recommend creating the suffix programmatically and/or passing it to `createAndVerifyValidApplication` to ensure that, as a developer, I can write a test scenario without having to read the fine print inside `createAndVerifyValidApplication`. 

Does this make sense?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationServiceCEImpl.java
abhiappsmith: Sure, I will add maxRetry. Also after maxRetry is exhausted I will add the error message to refresh the page and try again as this is happening due to sync lag.

app/server/appsmith-server/src/main/java/com/appsmith/server/exceptions/AppsmithErrorCode.java
abhiappsmith: @subrata71  so you suggest to change it to the below error code:
1. DUPLICATE_KEY("AE-APP-4024", "Duplicate key"), - > AE-APP-4091
2. (Newly added) DUPLICATE_KEY_OBJECT_CREATION("AE-APP-5043", "Duplicate key during object creation"), - > AE-APP-4092

subrata71: That's correct ☝️ @abhiappsmith

subrata71 (approved):
LGTM

PR #22684, opened 2023-04-24
feat: Update existing app via JSON import

### LINKED ISSUE #22075: Update app through JSON import
When a user wants to develop in an instance and deploy in another, the only way to update an app is via Git. In scenarios where this is not possible (either in Airgapped scenarios or due to the user's choice), the user cannot update the deployed app.

This feature allows a user to update an app by exporting and importing a new version of the app. The original references will be maintained where applicable. More details [here](https://www.notion.so/appsmith/Update-app-through-JSON-Import-c8d50ab0ad4c4a5da2802ba0bb9a2b6b)

https://app.frontapp.com/inboxes/teams/folders/15573132/archived/74473297292

abhvsn:
## Description

When a user wants to develop in an instance and deploy in another, the only way to update an app is via Git. In scenarios where this is not possible (either in Airgapped scenarios or due to the user's choice), the user cannot update the deployed app. This PR allows a user to update an existing app via JSON import. The original references will be maintained where applicable

> TL;DR: Provide a way to update the existing application via JSON import in situations where Git sync is not an option.

Fixes #22075 

### How to test?
1. Create a new workspace `WS1`
2. Create new app
3. Generate CRUD page
4. Export the application
5. Create new workspace lets's call this as `WS2` and import the exported app from step 4
6. Create another page `pageCreatedInWS2` in imported application in `WS2`
7. Export the updated application from `WS2`
8. Update existing application created in step 2 in `WS1` by importing the application which is exported from `WS2` in step 7
9. Updated application should have `pageCreatedInWS2` page

```
Curl for ref:
 curl --location '{{base_url}}/api/v1/applications/import/{{workspaceId}}?applicationId={{applicationId}}' \
--header 'Cookie: SESSION={{cookieSessionId}}' \
--header 'X-Requested-By: Appsmith' \
--form 'file=@{{JSON_FILE_TO_IMPORT}}'
```

## Type of change

## How Has This Been Tested?
- Manual
- JUnit

### Dev activity

### QA activity:


app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImplV2.java
```diff
@@ -1252,7 +1252,7 @@ private Mono<Application> importApplicationInWorkspace(String workspaceId,
                 })
                 .onErrorResume(throwable -> {
                     log.error("Error while importing the application, reason: {}", throwable.getMessage());
-                    return Mono.error(new AppsmithException(AppsmithError.GENERIC_JSON_IMPORT_ERROR, workspaceId, ""));
+                    return Mono.error(new AppsmithException(AppsmithError.GENERIC_JSON_IMPORT_ERROR, workspaceId, throwable.getMessage()));
```
AnaghHegde, line 1255: Sharing this error with the user will confuse the user. For example - if the mongoDB transaction fails user will see a long message and bad UX.

AnaghHegde (commented):
We should add a cypress test when the client changes goes live.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImplV2.java
abhvsn: Thanks, I wanted to discuss this. Is there a way to make this error reporting actionable? As we have heard from multiple stakeholders users, A-force members etc. that error message doesn't give any information. Is there a plan to address this in near future? Also, do we have the list of unreadable exceptions, if it's only related to transactions, in that case, we can filter them out. Let me know your views

AnaghHegde: Hmm yes, we can filter the transaction error. But IMO it does not make any difference. If the import flow fails we need support from devs. I am okay with either, as long as the transaction errors are handled properly.

AnaghHegde:
@abhvsn We should handle the version mismatch case as well. I don't think we have this check present in the Import flow (When the instances are on a different version).

abhvsn:
> @abhvsn We should handle the version mismatch case as well. I don't think we have this check present in the Import flow (When the instances are on a different version).

@AnaghHegde This is already handled in the downstream method 
```
importApplicationInWorkspace(String workspaceId, ApplicationJson applicationJson, String applicationId, String branchName, boolean appendToApp)
```

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ImportExportApplicationServiceCEImplV2.java
abhvsn: IMO this will be specifically useful for the airgap instances as getting the logs can be tricky in some situations at least this will speed up the debugging process once we have actionable error responses.

AnaghHegde: Make sense. Lets add this change.

AnaghHegde (approved):

PR #22718, opened 2023-04-25
fix: Invalidated error state connection mono from cache (#20538)

### LINKED ISSUE #20538: [Bug]: OAuth with client credentials fail for UiPath OAuth service

### Description

OAuth with client credentials fail for UiPath OAuth service
Ref: https://theappsmith.slack.com/archives/C0385FH70E8/p1671153589146359

### Steps To Reproduce

Ref: https://theappsmith.slack.com/archives/C0385FH70E8/p1671153589146359

### Public Sample App

_No response_

### Issue video log

_No response_

### Version

self hosted, cloud

sanveer-osahan:
## Description
When the token fetched from client credentials grant type in oauth2 is expired, a request to the token endpoint is made to fetch a new token.
But sometimes the token endpoint fail. It can be intermittent failure, a retry will fetch the desired token. 
When the failure occurs, the datasource context mono goes into an error state. As this mono is cached, all the subsequent authenticated api calls from the user fetches the cached error mono resulting in failures and thus the token endpoint is never retried.
For the fix, while checking if datasource context is valid, added the check if its in a failed state. If yes, then the context will be considered invalid.

Fixes #20538 

## Type of change

## How Has This Been Tested

- Manual

### Dev activity

### QA activity:


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
```diff
@@ -99,7 +99,8 @@ public Mono<? extends DatasourceContext<?>> getCachedDatasourceContextMono(Datas
              * value would directly be returned to further evaluations / subscriptions.
              */
             if (datasourceContextIdentifier.getDatasourceId() != null
-                    && datasourceContextMonoMap.get(datasourceContextIdentifier) != null) {
+                    && datasourceContextMonoMap.get(datasourceContextIdentifier) != null
+                    && !datasourceContextMonoMap.get(datasourceContextIdentifier).toFuture().isCompletedExceptionally()) {
```
sumitsum, line 103: Hi Sanveer, I was just thinking, would it be better to move it to line no 82 ? i.e. the check where we remove the values in the map - reason being that once we have discovered that a value is in error state it doesn't seem meaningful to not remove it.

sumitsum (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
sanveer-osahan: makes sense

sumitsum (approved):

PR #22836, opened 2023-04-28
fix: Remove console log on MetaWidgetUpdate

### LINKED ISSUE #22527: [Bug]: App becomes very slow when typing text in widgets

### Description

Issue : At first glance the app seems to be working good. But whenever you start typing a few texts it becomes very slow and it is slower and slower the more you type text in input fields or in richtext editor (or even when you select dropdown menu).

### Steps To Reproduce

- Create an app with a list widget that fetch data from API Query
- Populate the list widget with container widget that contains itself a few widgets (6-10 inputs, checkboxs, dropdown menus...)
- On production, try to write at human pace (fast one) some texts in inputs, rich text editor.. Slow typing often happens when you try to type in a widget of first list's item then quickly try to type in other item's widget

### Public Sample App

https://app.appsmith.com/app/outils-xmind-copy/gamemap-643ea052687d2c3120d3d653

### Environment

Production

### Issue video log

https://www.loom.com/share/61b6f6ebfb2741cc9f04668a1d8d1d91

### Version

Cloud - Appsmith v1.9.15-SNAPSHOT

rajatagrawal:
Fixes #22527 

Based on debugging the performance logs, it seems like for now that the console log on every character type is causing a lag in the performance.

This becomes more accentuated for a list widget. Commenting out the logging for now since the issue is marked critical

In the meanwhile, @SatishGandham will help to triage the further cause of this issue.

### DIFF
--- a/app/client/src/widgets/MetaHOC.tsx
+++ b/app/client/src/widgets/MetaHOC.tsx
@@ -136,19 +136,6 @@ function withMeta(WrappedWidget: typeof BaseWidget) {
       propertyValue: unknown,
       actionExecution?: DebouncedExecuteActionPayload,
     ): void => {
-      AppsmithConsole.info({
-        logType: LOG_TYPE.WIDGET_UPDATE,
-        text: "Widget property was updated",
-        source: {
-          type: ENTITY_TYPE.WIDGET,
-          id: this.props.widgetId,
-          name: this.props.widgetName,
-          propertyPath: propertyName,
-        },
-        state: {
-          [propertyName]: propertyValue,
-        },
-      });
       this.handleUpdateWidgetMetaProperty(
         propertyName,
         propertyValue,


app/client/src/widgets/MetaHOC.tsx
```diff
@@ -136,19 +136,6 @@ function withMeta(WrappedWidget: typeof BaseWidget) {
       propertyValue: unknown,
       actionExecution?: DebouncedExecuteActionPayload,
     ): void => {
-      AppsmithConsole.info({
```
SatishGandham, line 139: @Rishabh-Rathod , is this code required in view mode?
What is the purpose of logging this to the debugger? It just creates noise?

SatishGandham (approved):
If this is a critical issue, I don't see a problem removing it. Else I would wait to investigate why this is causing a problem.

PR #22854, opened 2023-04-30
chore: Add rapid mode to skip new app creation in cypress test

### LINKED ISSUE #22878: [Task]: Add flexibility to skip app creation during re-run of cypress tests while writing/debugging test cases

### SubTasks

This task is important because creating app every time while writing/debugging test isn't required and reduces the speed of writing tests.

rajatagrawal:
Fixes #22878

This pr adds a flag called 'rapidMode' in cypress index file.

```
cypress/support/index.js

let rapidMode = {
  enabled: false, // Set it to true to skip app creation during a test re-run
  appName: "4a9b62a0", // Replace it with your app name
  pageName: "page1", // Replace it with the page name
  pageID: "644f8d6e21506c33a366854b", // Replace it with pageID

  url: function () {
    return `app/${this.appName}/${this.pageName}-${this.pageID}/edit`;
  },
};
```
When rapid mode is enabled, the cypress test will

1. Skip relogin if the test user is already logged in.
2. Skip creation of a new app and use a previously created app, passed as param.

The flag needs to be disabled when the code needs to be merged to release, and is meant for debugging/testing while writing tests on local development setup only.

When enabled, a cypress test saves around 25 sec per rerun.

Future fixes would include skipping multiple workspace page visit if a developer is using dsl for setting up fixtures. It would save around another 10-15 seconds per test re-run.

### DIFF
--- a/app/client/cypress/integration/Regression_TestSuite/ClientSideTests/Widgets/TableV2/Add_new_row_spec.js
+++ b/app/client/cypress/integration/Regression_TestSuite/ClientSideTests/Widgets/TableV2/Add_new_row_spec.js
@@ -193,6 +193,7 @@ describe("Table widget Add new row feature's", () => {
 
   describe("Validation flow", () => {
     before(() => {
+      cy.startServerAndRoutes();
       agHelper.RestoreLocalStorageCache();
       cy.addDsl(dsl);
     });
@@ -350,6 +351,7 @@ describe("Table widget Add new row feature's", () => {
 
   describe("Actions flow (save, discard)", () => {
     before(() => {
+      cy.startServerAndRoutes();
       agHelper.RestoreLocalStorageCache();
       cy.addDsl(dsl);
     });
--- a/app/client/cypress/support/commands.js
+++ b/app/client/cypress/support/commands.js
@@ -272,11 +272,14 @@ Cypress.Commands.add("Signup", (uname, pword) => {
 
 Cypress.Commands.add("LoginFromAPI", (uname, pword) => {
   cy.location().then((loc) => {
+    let baseURL = Cypress.config().baseUrl;
+    baseURL = baseURL.endsWith("/") ? baseURL.slice(0, -1) : baseURL;
+
     cy.visit({
       method: "POST",
       url: "api/v1/login",
       headers: {
-        origin: loc.origin,
+        origin: baseURL,
       },
       followRedirect: true,
       body: {
@@ -287,9 +290,15 @@ Cypress.Commands.add("LoginFromAPI", (uname, pword) => {
       .then(() => cy.location())
       .then((loc) => {
         expect(loc.href).to.equal(loc.origin + "/applications");
+        cy.wait("@getMe");
+        cy.wait("@applications").should(
+          "have.nested.property",
+          "response.body.responseMeta.status",
+          200,
+        );
+        cy.wait("@getReleaseItems");
       });
   });
-  cy.wait(2000); //for the page elements to load!
 });
 
 Cypress.Commands.add("DeleteApp", (appName) => {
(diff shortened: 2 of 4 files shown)


app/client/cypress/integration/Regression_TestSuite/ClientSideTests/Widgets/TableV2/Add_new_row_spec.js
```diff
@@ -193,6 +193,7 @@ describe("Table widget Add new row feature's", () => {
 
   describe("Validation flow", () => {
     before(() => {
+      cy.startServerAndRoutes();
```
Aishwarya-U-R, line 196: Is this explicit call needed here?

rajatagrawal: yeah, because in this test, a describe has been nested inside another describe, so `startServerAndRoutes` needs to be initialized again. The previous initialization gets invalidated in this test.

`startServerAndRoutes` mentioned in beforeEach() mentioned in `index.js` cypress file doesn't get run until this command.

app/client/cypress/support/index.js
```diff
@@ -85,12 +114,14 @@ before(function () {
...
 before(function () {
+  if (rapidMode.enabled) {
+    return;
+  }
   //console.warn = () => {};
   Cypress.Cookies.preserveOnce("SESSION", "remember_token");
   const username = Cypress.env("USERNAME");
   const password = Cypress.env("PASSWORD");
   cy.LoginFromAPI(username, password);
-  cy.wait("@getMe");
```
keyurparalkar, line 93: @rajatagrawal why are we removing the wait here?

app/client/cypress/support/commands.js
```diff
@@ -597,18 +606,16 @@ Cypress.Commands.add("generateUUID", () => {
...
-    appId = localStorage.getItem("applicationId");
+
     //Fetch the layout id
     cy.request("GET", "api/v1/pages/" + pageid).then((response) => {
       const respBody = JSON.stringify(response.body);
-      layoutId = JSON.parse(respBody).data.layouts[0].id;
-      cy.log("appid:" + appId);
+      const data = JSON.parse(respBody).data;
+      layoutId = data.layouts[0].id;
+      appId = data.applicationId;
```
keyurparalkar, line 618: Is there a reason why are we using the `applicationId` from the API response rather than using it from the localStorage?

_Reason:_Previously we were using `applicationId` from the localStorage.

app/client/cypress/support/commands.js
```diff
@@ -287,9 +290,15 @@ Cypress.Commands.add("LoginFromAPI", (uname, pword) => {
...
+        cy.wait("@getMe");
+        cy.wait("@applications").should(
+          "have.nested.property",
+          "response.body.responseMeta.status",
+          200,
+        );
+        cy.wait("@getReleaseItems");
       });
   });
-  cy.wait(2000); //for the page elements to load!
```
keyurparalkar, line 292: Have we tested this before removing the wait?

rajatagrawal: yes, the test suite is passing on CI. Actually we are now waiting for the actual ajax call that takes time right before this instead of a fixed 2 second wait. This can save time if the call response is received before 2 seconds.

app/client/cypress/support/commands.js
rajatagrawal: The test is making the api call to get the layout ID already and the response contains the appID. So I grabbed the app ID from the response and tried to remove additional code since the app ID was already available.

app/client/cypress/support/index.js
rajatagrawal: Because I have moved this wait inside loginFromAPI function where it seemed more appropriate since this wait was waiting for getMe ajax call made inside loginFromAPI function.

app/client/cypress/integration/Regression_TestSuite/ClientSideTests/Widgets/TableV2/Add_new_row_spec.js
Aishwarya-U-R: I see this spec is existing with multiple describe blocks. Why is this call needed now?

rajatagrawal: Because I have added cy.wait("@getWorkspace") to the function definition of addDSL function.

Aishwarya-U-R (approved):

PR #22927, opened 2023-05-02
fix: display name validation (server)

### LINKED ISSUE #22578: [Bug]: Input field validation is missing for display name

### Description

There is no input field validation for display name. (Note: Add validations in backend as well)
image

### Steps To Reproduce

1.  Got to profile page from top right menu
2. Edit the display name

### Public Sample App

_No response_

### Environment

Production

### Issue video log

_No response_

### Version

Cloud

NilanshBansal:
## Description

> This PR adds a validation regex for display name, it allows for Accented characters such as (ä ö ü è ß) etc, and alphanumeric with some special characters dot (.), apostrophe ('), hyphen (-) and spaces.
> It also allows for chinese characters, eg. 황현미

Fixes #22578 

## Type of change

## How Has This Been Tested?

- JUnit

### Dev activity

### QA activity:


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserServiceCEImpl.java
```diff
@@ -659,16 +659,31 @@ public Flux<User> get(MultiValueMap<String, String> params) {
...
     }
 
+    private Boolean validateName(String name){
+        /*
+            Regex allows for Accented characters and alphanumeric with some special characters dot (.), apostrophe ('),
+            hyphen (-) and spaces
+         */
+        Boolean isValidName = name.matches("^[A-Za-zÀ-ÖØ-öø-ÿ0-9 .'-]+$");
+        return isValidName;
+    }
```
sharat87, line 669: This will never result in a `null`, so we can prefer to use the non-boxed version `boolean`.

```suggestion
    private boolean validateName(String name){
        /*
            Regex allows for Accented characters and alphanumeric with some special characters dot (.), apostrophe ('),
            hyphen (-) and spaces
         */
        boolean isValidName = name.matches("^[A-Za-zÀ-ÖØ-öø-ÿ0-9 .'-]+$");
        return isValidName;
    }
```

sharat87: Also, let's move this regex pattern to a static constant, not as a `String`, but a compiled `Pattern`, so the JDK doesn't have to compile the regex on every function call.

Example: https://github.com/appsmithorg/appsmith/blob/439ceb54064aec4c7170ae701c28e1e2267fefb4/app/server/appsmith-git/src/main/java/com/appsmith/git/helpers/FileUtilsImpl.java#L82

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserServiceCEImpl.java
```diff
@@ -659,16 +659,31 @@ public Flux<User> get(MultiValueMap<String, String> params) {
...
         Mono<User> updatedUserMono;
         Mono<UserData> updatedUserDataMono;
+        String inputName;
 
         if (allUpdates.hasUserUpdates()) {
             final User updates = new User();
-            updates.setName(allUpdates.getName());
+            inputName = allUpdates.getName();
+            Boolean isValidName = validateName(inputName);
+            if (isValidName != true){
```
sharat87, line 683: When using non-boxed versions, there's no worry of `null`s creeping in. 🙂

```suggestion
            boolean isValidName = validateName(inputName);
            if (isValidName) {
```

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserServiceCEImpl.java
```diff
@@ -659,16 +659,31 @@ public Flux<User> get(MultiValueMap<String, String> params) {
...
         List<Mono<Void>> monos = new ArrayList<>();
 
         Mono<User> updatedUserMono;
         Mono<UserData> updatedUserDataMono;
+        String inputName;
 
         if (allUpdates.hasUserUpdates()) {
             final User updates = new User();
-            updates.setName(allUpdates.getName());
+            inputName = allUpdates.getName();
```
sharat87, line 681: Minor, this can be declared inside here, right? 🤔

```suggestion
            String inputName = allUpdates.getName();
```

app/server/appsmith-server/src/test/java/com/appsmith/server/services/UserServiceTest.java
```diff
@@ -415,12 +415,36 @@ public void createUserWithInvalidEmailAddress() {
...
     public void updateNameOfUser() {
         UserUpdateDTO updateUser = new UserUpdateDTO();
-        updateUser.setName("New name of api_user");
+        updateUser.setName("New name of api user");
         StepVerifier.create(userService.updateCurrentUser(updateUser, null))
                 .assertNext(user -> {
                     assertNotNull(user);
                     assertThat(user.getEmail()).isEqualTo("api_user");
-                    assertThat(user.getName()).isEqualTo("New name of api_user");
+                    assertThat(user.getName()).isEqualTo("New name of api user");
```
sharat87, line 423: Let's prefer `assertEquals` and slowly move towards having more tests use that. Why? It produces better diffs and better logs when the test fails, so we can spot the problem easier.

```suggestion
                    assertEquals("New name of api user", user.getName());
```

Note that it takes the _expected value_ as first parameter, and the _actual value_ as the second parameter.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/UserServiceTest.java
```diff
@@ -415,12 +415,36 @@ public void createUserWithInvalidEmailAddress() {
...
+
+    @Test
+    @WithUserDetails(value = "api_user")
+    public void updateNameOfUser_WithNotAllowedSpecialCharacter_InvalidName() {
+        UserUpdateDTO updateUser = new UserUpdateDTO();
+        updateUser.setName("invalid name@symbol");
+        StepVerifier.create(userService.updateCurrentUser(updateUser, null))
+                .expectErrorMatches(throwable -> throwable instanceof AppsmithException && throwable.getMessage()
+                .contains(AppsmithError.INVALID_PARAMETER.getMessage(FieldName.NAME)))
+                .verify();
```
sharat87, line 436: This formatting was hard to follow. The way the `.extectErrorMatches`, `.contains` and `.verify` were lined up, I immediately though that `.verify` is called on the return value of `.contains`, and `.contains` is called on return value of `.expectErrorMatches`, which isn't the case.

Let's fix the indentation to express the intent better, and make reading the code easier. 🎉

```suggestion
        StepVerifier.create(userService.updateCurrentUser(updateUser, null))
            .expectErrorMatches(throwable ->
                throwable instanceof AppsmithException
                    && throwable.getMessage().contains(AppsmithError.INVALID_PARAMETER.getMessage(FieldName.NAME))
            )
            .verify();
```

What say?

sharat87 (commented):
Few minor things, but overall looks good.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserServiceCEImpl.java
```diff
@@ -668,7 +677,12 @@ public Mono<User> updateCurrentUser(final UserUpdateDTO allUpdates, ServerWebExc
 
         if (allUpdates.hasUserUpdates()) {
             final User updates = new User();
-            updates.setName(allUpdates.getName());
+            String inputName = allUpdates.getName();
+            boolean isValidName = validateName(inputName);
+            if (isValidName != true){
```
sharat87, line 682:
```suggestion
            if (!isValidName) {
```

app/server/appsmith-git/src/main/java/com/appsmith/git/helpers/FileUtilsImpl.java
```diff
@@ -80,6 +80,7 @@ public class FileUtilsImpl implements FileInterface {
     private static final String VIEW_MODE_URL_TEMPLATE = "{{viewModeUrl}}";
 
     private static final Pattern ALLOWED_FILE_EXTENSION_PATTERN = Pattern.compile("(.*?)\\.(md|git|gitignore|yml)$");
+    public static final Pattern ALLOWED_ACCENTED_CHARACTERS_PATTERN = Pattern.compile("^[\\p{L} 0-9 .\'\\-]+$");
```
sharat87, line 83: Can we move this to `UserServiceCEImpl`, since I don't think it's really relevant to file utilities, and also since we're only using it there today.

akash-codemonk (approved):

PR #22975, opened 2023-05-03
fix: Editor loses focus when ESC is pressed

### LINKED ISSUE #22960: [Bug]: Code editor loses focus when ESC is pressed

### Description

The JS code editor loses focus when ESC is pressed. This issue has been seen in Chrome/Firefox.

Cause - This behaviour stems from the fact that the "type / for commands" in editors found in the JS pane has to be closed when escape is pressed. Since no such menus show up on the editor for JS objects, the focus is lost without menu closing

### Steps To Reproduce

1. Go to the JS Editor
2. Type something - make sure the autocomplete hinter is not showing
3. Press escape
4. Notice that the focus is lost and one needs to click on editor again to regain focus,

### Public Sample App

_No response_

### Environment

Production

### Issue video log

_No response_

### Version

1.6.6

rimildeyjsr:
## Description

The JS object editor loses focus when ESC is pressed. This behavior was introduced to close the binding prompt in editors found in the JS panes. But since there are no binding prompts in the JS object editor, it removed the focus from the editor. 

To solve this, we are checking if the binding prompt is open or not, in addition to the other checks, to decide if the editor should lose focus.

Fixes #22960 

## Media

https://www.loom.com/share/855a3a463ad647709173a08a5c016930

## Type of change

## How Has This Been Tested?

- Manual
- Cypress

### Dev activity
- [x] I have performed a self-review of my code

### QA activity:
- [ ] Test plan has been peer-reviewed by QA
- [x] Cypress test cases have been added and approved by either SDET or manual QA
- [x] Added Test Plan Approved label after reviewing all Cypress test


app/client/src/components/editorComponents/CodeEditor/index.tsx
```diff
@@ -1225,6 +1238,26 @@ class CodeEditor extends Component<Props, State> {
     };
   };
 
+  // show features like evaluatedvaluepopup or binding prompts
+  showFeatures = () => {
+    return (
+      this.state.isFocused &&
+      !this.props.hideEvaluatedValue &&
+      ("evaluatedValue" in this.props ||
+        ("dataTreePath" in this.props && !!this.props.dataTreePath))
```
ravikp7, line 1247: Are we sure that we should be using the `in` keyword instead of `!this.props.evaluatedValue` or `!this.props.dataTreePath` here? Because `in` keyword can evaluate to `true` even if the property value is `undefined`.

For example:
```js
const a = { key: undefined };
"key" in a; // this will return true
```

app/client/src/components/editorComponents/CodeEditor/index.tsx
```diff
@@ -1241,7 +1274,6 @@ class CodeEditor extends Component<Props, State> {
       height,
       hideEvaluatedValue,
       hoverInteraction,
-      input,
```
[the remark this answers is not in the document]
rimildeyjsr: @ravikp7 - I moved the logic of showing binding prompt and show features out into functions and this value was being used there, so removed it from here

app/client/src/components/editorComponents/CodeEditor/index.tsx
rimildeyjsr: @ravikp7 - i just copied and moved this code out of the render block to its own function, and since this has been regressed and tested I did not change the logic here

ravikp7 (approved):

berzerkeer (approved):

PR #22989, opened 2023-05-04
feat: ci changes & cypress tests with cypress tags

### LINKED ISSUE #22007: [Task]: Cypress tests for Airgap

### SubTasks

_No response_

### LINKED ISSUE #22814: [Task]: Workflow changes for airgap

### SubTasks

- TBP airgap workflow
- client build workflow
- integration tests workflow
- ok-to-test command workflow
- build docker image workflow
- ci-test workflow

berzerkeer:
## Description

This includes

> Building a new image for airgapped instances
> Running ci-tests on airgapped image
> Running cypress tests selectively ignoring non supported features for airgap like Templates, Custom JS lib and also alternating test behaviours for some tests like tests using mock db, since it doesn't work on airgap we have to create a ds. So this selective testing was done using cypress-tags 
> Having a new client build for airgapped images which bundles all the assets.
> And changes in the workflow files to account for all the above. 

 With airgap, we can ignore certain tests and also need to account for tests using mock datasources and such by creating new datasources instead of mock datasources. Since those are blocked. So to perform a selective testing we are using a plugin called `cypress-tags` and to perform conditional testing when required we use the `AIRGAPPED` cypress env.  This PR introduces both and also modified the codebase to support this new way of running cypress. 

Since we can't trigger `/ok-to-test` on this because ci-test needs the CYPRESS_EXCLUDE_TAGS and slash command doesn't dispatch from current branch, 

I manually triggered the `TBP` workflow to run ci-test on this branch. 
And the new `TBP airgap` workflow to run ci-test on airgapped docker image on this branch.

Here is the link to the run 
https://github.com/appsmithorg/appsmith/actions/runs/4882041416

Fixes #22007 
Fixes #22814 

## Type of change

## How Has This Been Tested?

- Manual
- Cypress

### Dev activity

### QA activity:


app/client/cypress/integration/Regression_TestSuite/ClientSideTests/Widgets/ListV2/Listv2_BasicServerSideData_spec.js
```diff
@@ -9,64 +9,131 @@ const toggleJSButton = (name) => `.t--property-control-${name} .t--js-toggle`;
...
-
-    cy.get(publishLocators.containerWidget).each(($containerEl) => {
-      cy.wrap($containerEl)
-        .get(publishLocators.textWidget)
-        .eq(1)
-        .find("span")
-        .invoke("text")
-        .should("have.length.gt", 0);
-    });
+    if (!Cypress.env("AIRGAPPED")) {
```
Aishwarya-U-R, line 13: We cant use the new tags in before(), after() hooks? 
Hence you have conditioned this in `if` here?

app/client/cypress/support/Pages/DataSources.ts
```diff
@@ -197,6 +197,20 @@ export class DataSources {
     else if (mode == "View") this.agHelper.AssertElementExist(this._editButton);
   }
 
+  public GeneratePageWitPostgress(datasourceName: any) {
```
Aishwarya-U-R, line 200: Why new method?
I thought we are editing the GeneratePageWithMockDB()?

app/client/cypress/support/Pages/Onboarding.ts
```diff
@@ -17,7 +19,18 @@ export class Onboarding {
...
-    cy.get(OnboardingLocator.datasourceMock).first().click();
+    if (Cypress.env("AIRGAPPED")) {
+      const cyObj = cy as any;
+      cyObj.get(datasource.MongoDB).click();
+      cyObj.fillMongoDatasourceForm();
+      cyObj.generateUUID().then((uid: any) => {
+        datasourceName = `Mongo CRUD ds ${uid}`;
+        cyObj.renameDatasource(datasourceName);
+      });
+      cyObj.testSaveDatasource();
```
Aishwarya-U-R, line 30: Use _.dataSources.Createdatasource('Mongo')

app/client/cypress/support/commands.js
```diff
@@ -293,12 +293,17 @@ Cypress.Commands.add("LoginFromAPI", (uname, pword) => {
...
-        cy.wait("@applications").should(
-          "have.nested.property",
-          "response.body.responseMeta.status",
-          200,
-        );
-        cy.wait("@getReleaseItems");
+
+        if (CURRENT_REPO === REPO.EE) {
+          cy.wait(2000);
+        } else {
```
Aishwarya-U-R, line 299: If EE repo, we need to only wait? Need not assert for `@ applications` n/w call to complete?

.github/workflows/ci-smoke.yml
```diff
@@ -216,6 +216,8 @@ jobs:
           CYPRESS_APPSMITH_OAUTH2_OIDC_TOKEN_URL: ${{ secrets.CYPRESS_APPSMITH_OAUTH2_OIDC_TOKEN_URL }}
           CYPRESS_APPSMITH_OAUTH2_OIDC_USER_INFO: ${{ secrets.CYPRESS_APPSMITH_OAUTH2_OIDC_USER_INFO }}
           CYPRESS_APPSMITH_OAUTH2_OIDC_JWKS_URL: ${{ secrets.CYPRESS_APPSMITH_OAUTH2_OIDC_JWKS_URL }}
+          CYPRESS_EXCLUDE_TAGS: "airgap"
+          CYPRESS_AIRGAPPED: false
```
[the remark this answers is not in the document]
berzerkeer: Yes @Aishwarya-U-R

app/client/cypress/integration/Regression_TestSuite/ClientSideTests/AdminSettings/Admin_settings_spec.js
```diff
@@ -42,69 +42,120 @@ describe("Admin settings page", function () {
...
-    cy.url().should("contain", "/settings/authentication");
-    cy.get(adminsSettings.emailTab).click();
-    cy.url().should("contain", "/settings/email");
-    cy.get(adminsSettings.googleMapsTab).click();
-    cy.url().should("contain", "/settings/google-maps");
-    cy.get(adminsSettings.versionTab).click();
-    cy.url().should("contain", "/settings/version");
-  });
+  it(
+    "excludeForAirgap",
```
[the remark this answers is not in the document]
berzerkeer: This spec file is present in both CE and EE, so to not have conflicts on every spec files which are present in both CE and EE, i have added the excludeForAirgap in CE. Also in CE these tests will run, since we are only ever excluding the `airgap` tag on CE.

app/client/cypress/integration/Regression_TestSuite/ClientSideTests/Widgets/ListV2/Listv2_BasicServerSideData_spec.js
berzerkeer: Yes @Aishwarya-U-R.  I should look more into whether we can use it in before after hooks. But from the docs AFAIK we can only use it on describe and it blocks.

app/client/cypress/support/Pages/DataSources.ts
berzerkeer: The name first of all. And since this is anyways something we will be using often in feature if there are more tests which needs this. This will be easy and more obvious to use.

app/client/cypress/support/Pages/Onboarding.ts
berzerkeer: Ahh made the change in EE, will make the same change here.

app/client/cypress/support/commands.js
berzerkeer: Yeah basically in EE we have license check, so if the user doesn't have the license, it will redirect them to the /license page which won't call applications. Actually this change is already there in EE , I will remove it from here since its likely not something we need in CE.

app/client/cypress/support/Pages/DataSources.ts
Aishwarya-U-R: No, the GeneratePageWithMockDB is maximum used only by postgress or mongo here. Having one separate for Postgress is overhead & not required. Use the same method for both.

app/client/cypress/support/commands.js
Aishwarya-U-R: Yes please.

app/client/cypress/support/Pages/DataSources.ts
berzerkeer: No that function itself won't work for airgapped, MockDB will be hidden which means `Users` and `Movies` DB won't be shown at all. hence we have a separate function to create a postgresDB from scratch and not the mock postrgres or mongo DB.

Aishwarya-U-R: Oh got it now. You are using the Postgress Ds from locally running TED. Since all were users - I dint realise you are using users from TED. If possible - please change the table name, there are many table in TED Postgress - to avoid this confusion in future.

berzerkeer: Will have to decide about that. Because the problem is with keeping the same table name
1. There won't be any conflicts because this table will only ever created for airgapped instances otherwise we would still use mockDB
2. There are other tests which depend on DS with this table name, so if I change the table name that means other tests should be changed for airgap aswell to refer to the changed table name, which means same functionality but more code just because we have changed one table name.
3. I don't think this will lead to confusions because for mockDB we have every methods and selectors with `mock` attached in its name like the function `GenerateMockDb` , anything which was explicitly stating `createDatasource(Postgres) , means its something we create from scratch and not from mock. 

So I don't think this needs to be done as part of this PR. Because the changes will be too big and will take time.

Aishwarya-U-R: I see only one testcase calling GeneratePageWitPostgress() twice.. 
Which other tests are dependent on the table name? Can you please point me to the spec/dependent cases?

berzerkeer: I was talking about changing table names take the `ListV2_BasicServerSideData_spec.js` for example. In that we create the mockDb the `Users` table. with query command and everything getting data from that table. If I were to change the name of the table for airgap there that would mean I have to make changes in the rest of the steps aswell and since this is like per application per run , I don't see any good reason to keep different table name when creating postgres DB for airgapped cypress runs.

Also if I were to edit this GenerateMockDB command , that means its definition is wrong. The function name is `GeneratePageWithMockDB` but inside I should add a `Cypress.env(AIRGAPPED` check and based on that I should either click the mock Users db and generate page from it or I should click the datasourceName and generate page from it also the new function takes a param which is the datasource name. With having separate functions one for generating page via mock db and one for generating page via postgress it makes its intentions clear and also the function signature differs aswell. If its like a big complicated function which introduces a lot of code I am with you for creating a new function but since its a small function and only used when needed , with a different intent and different signature I wanted to have it as the separate function.

berzerkeer: > Oh got it now. You are using the Postgress Ds from locally running TED. Since all were users - I dint realise you are using users from TED. If possible - please change the table name, there are many table in TED Postgress - to avoid this confusion in future.

Maybe I understood this wrongly. Let's connect tomorrow and you can explain it to me. 😓

app/client/cypress/support/Pages/DataSources.ts
```diff
@@ -197,6 +194,20 @@ export class DataSources {
...
+    this.ee.AddNewPage("generate-page");
+    this.agHelper.GetNClick(this._selectDatasourceDropdown);
+    this.agHelper.GetNClickByContains(
+      this.locator._dropdownText,
+      datasourceName,
+    );
+    this.agHelper.GetNClick(this._selectTableDropdown);
+    this.agHelper.GetNClick("[data-cy='t--dropdown-option-public.users']");
```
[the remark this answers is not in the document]
berzerkeer: Ahhhhh my bad got it. my brain has been wired to think table name -> DS name. now I get it. Yes this makes sense, I will pass the table name as param

PR #23024, opened 2023-05-05
feat: eslint plugin appsmith

### LINKED ISSUE #23025: [Task]: Create a custom eslint rule to enforce wrapping asset urls in airgap helpers

### SubTasks

_No response_

berzerkeer:
## Description

- eslint plugin to enfore wrapping src attributes in getAssetUrl function to support airgap.

Fixes #23025 

Media

https://user-images.githubusercontent.com/74818788/236258005-e40a991e-4808-432a-ab71-299f277d17ee.mov

## Type of change

## How Has This Been Tested?

- Manual

### Dev activity

### QA activity:


berzerkeer:
@KelvinOm is the `author` field necessary ? This is something many people will contribute on in the future if we need more set of custom rules. Should I remove it all together and keep it and as people add rules, they can add them to the author field . Thanks !

app/client/packages/eslint/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "eslint-plugin-appsmith",
+  "version": "1.0.0",
+  "private": true,
+  "description": "Custom ESLint plugin rules for appsmith",
+  "main": "index.js",
+  "author": "Sangeeth Sivan <sangeeth@appsmith.com>",
+  "license": "MIT"
```
KelvinOm, line 8: You need to add this otherwise the build will fall. This is how yarn works now.
```
  "scripts": {
    "lint:ci": "echo 'No lint specified'",
    "prettier:ci": "echo 'No prettier specified'"
  },
```

Снимок экрана 2023-05-05 в 13 34 19

app/client/.eslintrc.json
```diff
@@ -38,9 +46,13 @@
...
     "react/jsx-no-useless-fragment": "error",
-    "sort-destructure-keys/sort-destructure-keys": ["error", {"caseSensitive": false}],
+    "sort-destructure-keys/sort-destructure-keys": [
+      "error",
+      { "caseSensitive": false }
+    ],
     "no-console": "warn",
-    "no-debugger": "warn"
+    "no-debugger": "warn",
+    "appsmith/airgap-asset-urls": "warn"
```
KelvinOm, line 55: Let's make this an error and fix the problems. From experience, no one will pay attention to warnings, especially when there are already 46.

Снимок экрана 2023-05-05 в 13 35 43

KelvinOm (commented):
Let's call the folder with the package `eslint-custom-rules`. I think it will be more clear what is inside.

KelvinOm:
>@KelvinOm is the author field necessary ? This is something many people will contribute on in the future if we need more set of custom rules. Should I remove it all together and keep it and as people add rules, they can add them to the author field . Thanks !

[Only name and version are required](https://docs.npmjs.com/creating-a-package-json-file#required-name-and-version-fields). In addition, there is a contributors field, so if someone wants to add information about themselves, they can modify the `package.json` file. I don't see a problem here.

app/client/.eslintrc.json
berzerkeer: The reason why I made it as a warning is because, this only looks for `src` attributes in a JSX element, but there are multiple ways to actually have the src attribute value wrapped in `getAssetUrl` function. So example one could have a variable which is a function which returns string. Like

```typescript
const IMAGE_URL. = () => getAssetUrl("example.com");
<ImageElement src={IMAGE_URL}/>
```  
I can't evaluate the function and get its value and then assert, so I just took the easy way and only asserted all `src` attributes since whatever the implementation can be , it has to go to the `src` attribute in the end. 

So hence I kept it as a warning.  They can still disable it but with a warning they have the flexibility to ignore without polluting the codebase with `eslint-disable/asset-airgap-url` checks everywhere. So I think warning makes more sense.

berzerkeer:
> > @KelvinOm is the author field necessary ? This is something many people will contribute on in the future if we need more set of custom rules. Should I remove it all together and keep it and as people add rules, they can add them to the author field . Thanks !
> 
> [Only name and version are required](https://docs.npmjs.com/creating-a-package-json-file#required-name-and-version-fields). In addition, there is a contributors field, so if someone wants to add information about themselves, they can modify the `package.json` file. I don't see a problem here.

Nice then, I will remove the author field and also update the plugin name to eslint-custom-rules

app/client/.eslintrc.json
KelvinOm: I don't agree with that. These warnings will appear in every PR and there is a high probability of missing something important. It's always better to keep the output clean. So I would suggest suppressing these warnings if we are really sure that they are just warnings.

PR #23031, opened 2023-05-05
chore: Update Spring to v3.0.6

### LINKED ISSUE #1407: [Feature] Ability to set style for "Row button" in table
## Summary

Ability to set different styles (colors) for "Row buttons" in the table widget.

## Motivation

This would be useful to show the buttons intent. For instance an "edit"-button could be green or blue and a "delete"-button could be red.

Blueprints different intents could be used: "primary", "success", "danger" and "warning."

The button widget has this type of function already but only with "primary", "secondary" and "danger" intents.

### Example
buttons

sharat87:
Fixes [#1407](https://github.com/appsmithorg/appsmith-ee/issues/1407)


app/server/pom.xml
```diff
@@ -26,10 +26,7 @@
         <project.version>1.0-SNAPSHOT</project.version>
         <!-- By default skip the dockerization step. Only activate if necessary -->
         <skipDockerBuild>true</skipDockerBuild>
-        <spring-boot.version>3.0.1</spring-boot.version>
-        <!-- Required because Spring Boot v3.0.1 does not come with this version that solves a High priority CVE -->
-        <netty.version>4.1.86.Final</netty.version>
```
sharat87, line 31: We now get `4.1.91`, so removing this override.

app/server/pom.xml
```diff
@@ -26,10 +26,7 @@
...
-        <log4j2.version>2.17.1</log4j2.version>
```
sharat87, line 32: We now get 2.19.0, so removing this as well.

nidhi-nair (dismissed):

abhvsn (approved):

PR #23084, opened 2023-05-08
feat: Auto cleanup application logs

sum35h:
- Added cron task to cleanup log files older than 7 days to run weekly.
- Updated supervisor log rotation config to store only 2 backups for all processes.


deploy/docker/templates/cron.d/cleanup-logs
```diff
@@ -0,0 +1 @@
+@weekly root find /appsmith-stacks/logs/backend /appsmith-stacks/logs/rts/ /appsmith-stacks/logs/editor/ -name "*.log*" -type f -mtime +7 -delete
```
sharat87, line 1: Can we define a specific time every week, instead of `@weekly` please? Reason is, if I recall correctly, the way `@weekly` works, is that it'll run once at system startup, then it'll wait 7 days to run again. This is fine, but this means that the time of logs cleanup is not predictable. If a user comes to us asking when this is going to run, I'd like to be able to give a simple answer, like "Sunday mornings in UTC" or something. Does that sound okay?

I see that we are only specifying backend, rts, and editor, instead of the whole logs folder. I'm fine with this, but curious on any particular reason for this? 🤔

sum35h: Sure, I'll update the time.
Only backend, rts and editor logs are cleaned up since their processes name keeps changing. On the other hand for the rest mongo, postfres, etc, the supervisor log rotation would take care of it.

sum35h, line 1:
```suggestion
0 0 * * SUN root find /appsmith-stacks/logs/backend /appsmith-stacks/logs/rts/ /appsmith-stacks/logs/editor/ -name "*.log*" -type f -mtime +7 -delete
```

sharat87 (approved):

PR #23119, opened 2023-05-09
fix: CURL empty command error msg fix

### LINKED ISSUE #12008: [Bug]-[250]:cURL command - On empty box the user is displayed with inappropriate error message

### Description

When Import is clicked on an empty box of cURL inappropriate error message is displayed to the user stating  "Enter a valid  parameter cURLcode "

Expectation : 
Since the user has not added any cURL code we can let the user know by stating error as "Please add the cURL code"

### Steps To Reproduce

1) Navigate to cURL  
2) click on import option 
and observe the error message 

Note : Ensure the cURL code is empty 

### Public Sample App

_No response_

### Version

Release

PiyushPushkar02:
## Description

- Changed the error message returned in case of empty input from the user for CURL import

Fixes #12008 

## Type of change

## How Has This Been Tested?

- Manual

### Dev activity

### QA activity:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/CurlImporterServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/CurlImporterServiceCEImpl.java
@@ -80,7 +80,7 @@ public class CurlImporterServiceCEImpl extends BaseApiImporter implements CurlIm
         ActionDTO action;
 
         try {
-            if (input == null) {
+            if (StringUtils.isBlank((String) input)) {
                 throw new AppsmithException(AppsmithError.EMPTY_CURL_INPUT_STATEMENT, FieldName.CURL_CODE);
             }
             action = curlToAction((String) input, name);
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/CurlImporterServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/CurlImporterServiceCEImpl.java
@@ -37,6 +37,7 @@ import java.util.Base64;
 import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
+import static org.apache.commons.lang3.StringUtils.isBlank;
 
 @Slf4j
 public class CurlImporterServiceCEImpl extends BaseApiImporter implements CurlImporterServiceCE {
@@ -80,7 +81,7 @@ public class CurlImporterServiceCEImpl extends BaseApiImporter implements CurlIm
         ActionDTO action;
 
         try {
-            if (StringUtils.isBlank((String) input)) {
+            if (isBlank((String) input)) {
                 throw new AppsmithException(AppsmithError.EMPTY_CURL_INPUT_STATEMENT, FieldName.CURL_CODE);
             }
             action = curlToAction((String) input, name);
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/CurlImporterServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/CurlImporterServiceCEImpl.java
@@ -81,7 +81,7 @@ public class CurlImporterServiceCEImpl extends BaseApiImporter implements CurlIm
 
         try {
             if (input == null) {
-                throw new AppsmithException(AppsmithError.INVALID_CURL_INPUT_PARAMETER, FieldName.CURL_CODE);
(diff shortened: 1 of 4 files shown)


PiyushPushkar02:
> Hi @PiyushPushkar02 , can you please add a JUnit TC for this in `CurImporterServiceTests.java` if possible ?

Done @sumitsum

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/CurlImporterServiceCEImpl.java
```diff
@@ -81,7 +81,7 @@ public Mono<ActionDTO> importAction(Object input, String pageId, String name, St
 
         try {
             if (input == null) {
```
sumitsum, line 83: Hi Piyush, can you please update this if condition to `StringUtils.isBlank(input)` - this will check for two input being null and input being empty string as well. Also, when using the function please use it like `isBlank(...)` instead of `StringUtils.isBlank(...)` (by using appropriate import statement).

app/server/appsmith-server/src/test/java/com/appsmith/server/services/CurlImporterServiceTest.java
```diff
@@ -180,6 +180,28 @@ public void testImportAction_EmptyLex() {
                 .verify();
     }
 
+    @Test
+    @WithUserDetails(value = "api_user")
+    public void testImportAction_NullLex() {
```
sumitsum, line 185: Can you please update the test case name to be more verbose and direct e.g. something along the lines of  `testImportActionErrorOnNullInput`. Also, can you please add another test case to test empty string input. ?

sumitsum (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/CurlImporterServiceCEImpl.java
PiyushPushkar02: Here the input is not a string Object. Its an object of the Java Parent Object class.

sumitsum: Hi @PiyushPushkar02 , APMU, in the very next line the input object is being cast as a string : `action = curlToAction((String) input, name);` - i.e. the object is actually intended to be a string object. Could we do this type cast early and check for empty string as well ?

app/server/appsmith-server/src/test/java/com/appsmith/server/services/CurlImporterServiceTest.java
```diff
@@ -176,7 +176,29 @@ public void testImportAction_EmptyLex() {
         StepVerifier
                 .create(action)
                 .expectErrorMatches(throwable -> throwable instanceof AppsmithException &&
-                        throwable.getMessage().equals(AppsmithError.INVALID_CURL_COMMAND.getMessage()))
+                        throwable.getMessage().equals(AppsmithError.EMPTY_CURL_INPUT_STATEMENT.getMessage()))
```
sumitsum, line 179: Hi Piyush, it seems like this test is named as `emptyLex` but the input string passed is not actually empty - it contains a single quote - so I am not sure if it is indeed testing for empty string case - can you please check this ? If required, please add a new test case for empty input string case.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/CurlImporterServiceCEImpl.java
```diff
@@ -80,8 +80,8 @@ public Mono<ActionDTO> importAction(Object input, String pageId, String name, St
         ActionDTO action;
 
         try {
-            if (input == null) {
-                throw new AppsmithException(AppsmithError.INVALID_PARAMETER, FieldName.CURL_CODE);
+            if (StringUtils.isBlank((String) input)) {
```
sumitsum, line 83: Hi Piyush, as mentioned earlier, if possible, can you please write it as `isBlank(...)` and add import statement like `import ...StringUtils.isBlank`

sumitsum (changes requested):

app/server/appsmith-server/src/test/java/com/appsmith/server/services/CurlImporterServiceTest.java
PiyushPushkar02: As suggested @sumitsum I have checked the mentioned method. It seems to be checking for invalid input rather than empty input. I have added a test case for empty input and also changed the name for above mentioned to to InvalidInput.

sumitsum (approved):

PR #23224, opened 2023-05-11
fix: Use spread operator to make a copy of action dynamic bindings from re…

### LINKED ISSUE #23194: [Bug]: Unable to bind {{Table.pageSize}} to SQL Query

### Description

Unable to bind Table.pageSize in Users SQL query.

### Steps To Reproduce

1. Add sample Users DB Query.
2. Add a table and bind its data to the this query.
3. Update the select query to `SELECT * FROM users ORDER BY id LIMIT {{Table1.pageSize}}`
4. The binding fails to get attached and the saving spinner keeps rotating.

This issue happens most of the times but not all the times. Refresh the page/Replace the binding with a hard coded value and put the binding back again if the issue doesn't repro.

Happening on release tree starting from 8854493ff0.

Error screenshot:
Screenshot 2023-05-10 at 7 34 20 PM

### Public Sample App

_No response_

### Environment

Production

### Issue video log

https://www.loom.com/share/0d9ba48a5968430ea74067382cdac979

### Version

Cloud

rajatagrawal:
…dux state

Fixes #23194 

The code is trying to modify a value (i.e action.dynamicBindings) from redux store, which isn't allowed. The fix makes a copy of this variable so that modifications can be made to it later on.

Test Plan:

1. Add a table.
2. Add a postgres select query and bind table's pageSize and pageOffset property to the query.

The binding should work.

### DIFF
--- a/app/client/cypress/e2e/Regression/ServerSide/QueryPane/Postgres_Spec.js
+++ b/app/client/cypress/e2e/Regression/ServerSide/QueryPane/Postgres_Spec.js
@@ -5,6 +5,7 @@ import * as _ from "../../../../support/Objects/ObjectsCore";
 import { ObjectsRegistry } from "../../../../support/Objects/Registry";
 
 let ag = ObjectsRegistry.AggregateHelper;
+let ee = ObjectsRegistry.EntityExplorer;
 
 let datasourceName;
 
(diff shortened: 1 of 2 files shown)


sneha122 (approved):

ohansFavour (dismissed):

sneha122 (approved):

btsgh:
I think the original issue is because of the slight delay in Saving of the query. The error that you on the video that you have attached, I can also reproduce it on my release - and the error states that it's an issue with Syntax. This is because the save was not completed. Hence on re-run because save has got completed, the query passes.
I have taken a pull of this fix on localhost, and even here, when I run the query quickly without waiting for the query to be saved, I get an error [Though the error is slightly different in this case] But if I wait a second or two longer and run the query, it runs correctly, and on any re-run after that, it passes.
Please let me know how you want me to go about this?

btsgh:
As per above observations, and discussions with @rajatagrawal  - I am unable to reproduce the issue consistently on Release. However, on the local pull of this fix, I cannot reproduce the specific issue mentioned in this bug.

PR #23246, opened 2023-05-11
chore: Renamed datasource storage doc for indexing needs

nidhi-nair:
This PR is simply a refactor to make sure index names are smaller.


nidhi-nair:
Merging to avoid double migration in release

PR #23380, opened 2023-05-16
fix: Disable redis and mongo by default

pratapaprasanna:
The reason we are running all off them inside fat-container is that, we have observed a couple of people who are trying to deploy appsmith fail because of not having enough resources to schedule mongo and redis. Thereby, we thought its best kept them disabled to install appsmith .

To deploy appsmith on minikube also it is best to keep them inside fat-container-pod

### DIFF
--- a/deploy/helm/Chart.yaml
+++ b/deploy/helm/Chart.yaml
@@ -11,7 +11,7 @@ sources:
   - https://github.com/appsmithorg/appsmith
 home: https://www.appsmith.com/
 icon: https://assets.appsmith.com/appsmith-icon.png
-version: 2.0.2
+version: 2.0.3
 dependencies:
 - condition: redis.enabled
   name: redis
--- a/deploy/helm/values.yaml
+++ b/deploy/helm/values.yaml
@@ -1,6 +1,6 @@
 ## Redis parameters
 redis:
-  enabled: true
+  enabled: false
   auth:
     enabled: false
   master:
@@ -10,7 +10,7 @@ redis:
     nodeSelector: {}
 
 mongodb:
-  enabled: true
+  enabled: false
   service:
     nameOverride: appsmith-mongodb
   auth:


sharat87 (changes requested):
Can you add details to the PR description on what we're _fixing_ please? Why are we disabling these? Does this mean they won't run outside the container, but be running inside?

sharat87:
> we have observed a couple of people who are trying to deploy appsmith fail because of not having enough resources to schedule mongo and redis
Not sure I understand this. If there's no resources to run MongoDB and Redis as separate pods, we can't expect to have resources to run them in the same pod, no? In fact, when run as separate pods, Kubernetes can run them on separate nodes, distributing the resources better, but if we're running the embedded MongoDB and Redis, they all have to run on the same node as a single pod, making it harder, right? Am I missing something?

> To deploy appsmith on minikube also it is best to keep them inside fat-container-pod
Should we tune the defaults to Minikube? And, can you elaborate on why this is the best way for Minikube please?

sharat87 (changes requested):
> we have observed a couple of people who are trying to deploy appsmith fail because of not having enough resources to schedule mongo and redis

Not sure I understand this. If there's no resources to run MongoDB and Redis as separate pods, we can't expect to have resources to run them in the same pod, no? In fact, when run as separate pods, Kubernetes can run them on separate nodes, distributing the resources better, but if we're running the embedded MongoDB and Redis, they all have to run on the same node as a single pod, making it harder, right? Am I missing something?

> To deploy appsmith on minikube also it is best to keep them inside fat-container-pod

Should we tune the defaults to Minikube? And, can you elaborate on why this is the best way for Minikube please?

pratapaprasanna:
> > we have observed a couple of people who are trying to deploy appsmith fail because of not having enough resources to schedule mongo and redis
> 
> Not sure I understand this. If there's no resources to run MongoDB and Redis as separate pods, we can't expect to have resources to run them in the same pod, no? In fact, when run as separate pods, Kubernetes can run them on separate nodes, distributing the resources better, but if we're running the embedded MongoDB and Redis, they all have to run on the same node as a single pod, making it harder, right? Am I missing something?
> 
> > So, when we enable mongo and redis as different sub-helm charts, they will create/require pv and pvc which can cause some issues in scheduling. If, we have them all as one in fat-container we allocate a common pvc of size `10Gi` and we have seen them work without any issue.  more like this
```
| data-appsmith-0                                  Bound             pvc-e70dcfa0-ef55-4a68-8353-78f562337082             10Gi                RWO                      gp2                      105s
| datadir-appsmith-mongodb-0                       Bound             pvc-1c63acfc-8465-4644-b54e-bd6959db9390             8Gi                  RWO                      gp2                      43s               │
│ datadir-appsmith-mongodb-1                       Bound             pvc-76550e0e-7591-4f4c-9231-7bfb8814ffe4             8Gi                  RWO                      gp2                      12s               │
│ redis-data-appsmith-redis-master-0               Bound             pvc-06c9fd71-1a74-4b49-a3de-8b609bc3171e             8Gi                  RWO                      gp2                      43s               │
│ redis-data-appsmith-redis-replicas-0             Bound             pvc-ed369adb-2964-45d6-950b-2089d93f84f0             8Gi                  RWO                      gp2                      43s
```

and not just on the storage also works for the same CPU and Memory allocations for redis and mongo pods as well

> > To deploy appsmith on minikube also it is best to keep them inside fat-container-pod
> 
> Should we tune the defaults to Minikube? And, can you elaborate on why this is the best way for Minikube please?
> > The reason is because of the resources seperate mongo/redis  require to launch on minikube. If a user who is using minikube to deploy appsmith shouldnt be stuck with an issue like resources which barely is a barrier to launch appsmith

sharat87 (commented):
> when we enable mongo and redis as different sub-helm charts, they will create/require pv and pvc which can cause some issues in scheduling

Are these problems specific to Minikube?

My problem with this change is, is this the best for EKS users? For AKS users? Is this the best method of deploying for someone running a proper cluster? From what I understand, that's not the case. Disabling them by default is the preference only for Minikube.

If that is correct, I'd highly encourage adding a note in Minikuce installation docs to disable these two in the `values.yml` before installing, instead of changing the defaults. If we're changing the defaults, we have to add instructions for all other deployment methods to _enable_ them before installation, right?

pratapaprasanna:
cc: @Nikhil-Nandagopal @jnikhila  are off the opinion that, It is best we disable mongo and redis by default in both CE and EE . 
In docs 
for CE: we tell users that they can opt for external mongo and redis. 
for EE: 
Non-HA: we tell users that they can opt for external mongo and redis.
But, by default redis and mongo are disabled

HA:
we tell them incase you want HA 
Pre-requisite:
* You need a EKS cluster
* You need a shared storage

enable the following in values.yaml
- redis
- mongo
- postgres
- autoscaling

PR #23395, opened 2023-05-16
chore: move local dependency to packages

### LINKED ISSUE #23333: Clean up client workspaces
Move yarn env to client folder

KelvinOm:
## Description
1. Move everything related to client from app folder to client folder (`.yarn`, `yarn.lock`, package.json, .gitignore)
2. Move `ast` and `rst` to client packages
3. Fix running scripts in packages
4. Add running unit tests in packages in CI

TODO: It is necessary to consider enabling the `nmHoistingLimits: workspaces` option, since now all packages are hoisted to the root, there may be issues with dependencies in workspaces. Also, there is a possibility of implicit use of packages.

https://yarnpkg.com/configuration/yarnrc#nmHoistingLimits

Fixes #23333

#### Type of change

- [x] Manual
- [x] Jest
- [x] Cypress

#### Dev activity


KelvinOm:
> @KelvinOm , Can you find out what the convention is on where to put the builds of individual workspace packages.
> We are putting it in workspace directory here, should the builds be put in root, i.e client?

@SatishGandham If there is a need to build something, then the artifacts should be at the workspace level as if it were an isolated package. In any case, the point of workspaces is not to build anything except the main application.

app/client/package.json
```diff
@@ -4,9 +4,13 @@
   "private": true,
   "engines": {
     "node": "^16.14.0",
-    "npm": "^8.5.5"
+    "yarn": "^3.5.1"
```
iamakulov, line 7: 😍

app/client/packages/rts/build.sh
```diff
@@ -7,8 +7,6 @@ rm -rf dist/
 # This is required for the first time build as node_modules is not present in the image
 yarn install --immutable
 yarn tsc && yarn tsc-alias
-# Keep copy of all dependencies in node_modules_bkp
-mv node_modules node_modules_bkp
```
iamakulov, line 11: ❤️‍🔥

iamakulov (dismissed):
Love this cleanup!

> TODO: It is necessary to consider enabling the `nmHoistingLimits: workspaces` option, since now all packages are hoisted to the root, there may be issues with dependencies in workspaces. Also, there is a possibility of implicit use of packages.

Not sure what the performance implications of this are (might be minuscule, might be significant). But just in case, if you’re only worried about some environments (eg CI builds) or some packages, it’s also possible to set that flag per command, with the `YARN_NM_HOISTING_LIMITS` variable. E.g.:

```sh
YARN_NM_HOISTING_LIMITS=workspaces yarn install
```

(This is [a convention](https://yarnpkg.com/configuration/yarnrc) that works for all config flags.)

KelvinOm:
> Love this cleanup!
> 
> > TODO: It is necessary to consider enabling the `nmHoistingLimits: workspaces` option, since now all packages are hoisted to the root, there may be issues with dependencies in workspaces. Also, there is a possibility of implicit use of packages.
> 
> Not sure what the performance implications of this are (might be minuscule, might be significant). But just in case, if you’re only worried about some environments (eg CI builds) or some packages, it’s also possible to set that flag per command, with the `YARN_NM_HOISTING_LIMITS` variable. E.g.:
> 
> ```shell
> YARN_NM_HOISTING_LIMITS=workspaces yarn install
> ```
> 
> (This is [a convention](https://yarnpkg.com/configuration/yarnrc) that works for all config flags.)

@iamakulov Implicit use of dependencies can lead to unexpected consequences. Besides, this may get worse when we use workspaces more widely. 

As for performance, I still think that there is nothing faster than pnpm. At least we should try using the `nodeLinker: pnpm` option.

KelvinOm:
[build for 8fc4f9f](https://github.com/appsmithorg/appsmith/actions/runs/5016976616)

KelvinOm:
[build for e941933](https://github.com/appsmithorg/appsmith/actions/runs/5043720342)

KelvinOm:
[build for bced199](https://github.com/appsmithorg/appsmith/actions/runs/5043863959)

KelvinOm:
[build for 672a417](https://github.com/appsmithorg/appsmith/actions/runs/5043933055)

KelvinOm:
[build for 7ea4a1f](https://github.com/appsmithorg/appsmith/actions/runs/5044264947)

Issue #21272, opened 2023-03-08
[Bug]: appsmith-0 pod stuck in Init:0/2 because redis-init-container fails to install redis-cli


schradert:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Sometimes when I deploy the helm chart to my Kubernetes cluster (haven't tested other methods besides the following):

```bash
helm repo add appsmith https://helm.appsmith.com
helm install appsmith appsmith/appsmith --namespace appsmith --create-namespace
```

the `appsmith-0` pod gets stuck in `Init:0/2` status, and the logs from:

```bash
kubectl -n appsmith logs pods/appsmith-0 -c redis-init-container
```

just show on repeat:

```plain
waiting for redis
sh: redis-cli: not found
```

It will also randomly do this when the Pod has to restart for any reason that isn't a new deployment. I leave it running in a fully ready state at the end of the day, and when I check back the next day, it's stuck again without anyone having touched it. My expectation is that it would never be stuck in this situation.

### Steps To Reproduce

The two ways I have found to fix this interactively are:

```bash
kubectl -n appsmith delete pods/appsmith-0
```

and, more explicitly:

```bash
kubectl -n appsmith exec appsmith-0 -c redis-init-container -- apk add redis
```

This doesn't always happen clearly, as deleting the pod to force initialization usually works successfully. From my perspective, for some reason, it seems `apk add redis` is failing when the init container starts. The first logs from `redis-init-container` show:

```plain
fetch https://dl-cdn.alpinelinux.org/alpine/v3.17/main/x86_64/APKINDEX.tar.gz
ERROR: https://dl-cdn.alpinelinux.org/alpine/v3.17/main: temporary error (try again later)
WARNING: Ignoring https://dl-cdn.alpinelinux.org/alpine/v3.17/main: No such file or directory
fetch https://dl-cdn.alpinelinux.org/alpine/v3.17/community/x86_64/APKINDEX.tar.gz
ERROR: https://dl-cdn.alpinelinux.org/alpine/v3.17/community: temporary error (try again later)
WARNING: Ignoring https://dl-cdn.alpinelinux.org/alpine/v3.17/community: No such file or directory
ERROR: unable to select packages:
  redis (no such package):
    required by: world[redis]
```

This is from `apk add redis`. The fact that it says `temporary error (try again later)` a few times makes me think there's some kind of connectivity issue.

I haven't looked into anything yet, but I think two ways I would try to address this is to either:

[...]

PR #23401, opened 2023-05-16
chore: update init-container images and update probes

pratapaprasanna:
Fixes: [21271](https://github.com/appsmithorg/appsmith/issues/21272)
Fixes: [221](https://github.com/appsmithorg/cloud-deployment/issues/221)
Fixes: [24020](https://github.com/appsmithorg/appsmith/issues/24020)
This will help update the init-container image and also updates the probes so that user doesnt find the necessity to restart manually. Rather, kubelet restarts it by self.

**Case-1:**

```
(base) ➜  helm git:(update/cosmetic/changes/ce) ✗ cat values.yaml | grep probes
(base) ➜  helm git:(update/cosmetic/changes/ce) ✗ helm install appsmith . -n goutham --create-namespace --dry-run  > temp.yaml
(base) ➜  helm git:(update/cosmetic/changes/ce) ✗ tail -n 60 temp.yaml
            {}
          image: "index.docker.io/appsmith/appsmith-ce:latest"
          imagePullPolicy: IfNotPresent
          ports:
            - name: http
              containerPort: 80
              protocol: TCP
            - name: https
              containerPort: 443
              protocol: TCP
            - name: supervisord
              containerPort: 9001
              protocol: TCP
          startupProbe:
            # The `livenessProbe` and `readinessProbe` will be disabled until the `startupProbe` is successful.
            httpGet:
              port: http
              path: "/api/v1/health"
            failureThreshold: 3
            periodSeconds: 60
          livenessProbe:
            httpGet:
              port: http
              path: "/api/v1/health"
            failureThreshold: 3
            periodSeconds: 10
          readinessProbe:
            httpGet:
              port: http
              path: "/api/v1/health"
            failureThreshold: 3
            periodSeconds: 10
```

**Case-2**
```
(base) ➜  helm git:(update/cosmetic/changes/ce) ✗ cat values.yaml | grep probes
probes: {}
(base) ➜  helm git:(update/cosmetic/changes/ce) ✗ helm install appsmith . -n goutham --create-namespace --dry-run  > temp.yaml
(base) ➜  helm git:(update/cosmetic/changes/ce) ✗ tail -n 60 temp.yaml
            {}
          image: "index.docker.io/appsmith/appsmith-ce:latest"
          imagePullPolicy: IfNotPresent
          ports:
            - name: http
              containerPort: 80
              protocol: TCP
            - name: https
              containerPort: 443
              protocol: TCP
            - name: supervisord
              containerPort: 9001
              protocol: TCP
          startupProbe:
            # The `livenessProbe` and `readinessProbe` will be disabled until the `startupProbe` is successful.
            httpGet:
              port: http
              path: "/api/v1/health"
            failureThreshold: 3
            periodSeconds: 60
          livenessProbe:
            httpGet:
              port: http
              path: "/api/v1/health"
            failureThreshold: 3
            periodSeconds: 10
          readinessProbe:
            httpGet:
              port: http
              path: "/api/v1/health"
            failureThreshold: 3
            periodSeconds: 10
```

**Case-3:**
```
(base) ➜  helm git:(update/cosmetic/changes/ce) ✗ head -n 15 values.yaml
probes:
  livenessProbe:
    api: "/api/v1/health"
    failureThreshold: 8
    periodSeconds: 40
  readinessProbe:
    api: "/api/v1/health"
    failureThreshold: 9
    periodSeconds: 50
  startupProbe:
    api: "/api/v1/health"
    failureThreshold: 10
    periodSeconds: 60
## Redis parameters
redis:
(base) ➜  helm git:(update/cosmetic/changes/ce) ✗ helm install appsmith . -n goutham --create-namespace --dry-run  > temp.yaml
(base) ➜  helm git:(update/cosmetic/changes/ce) ✗ tail -n 60 temp.yaml
            {}
          image: "index.docker.io/appsmith/appsmith-ce:latest"
          imagePullPolicy: IfNotPresent
          ports:
            - name: http
              containerPort: 80
              protocol: TCP
            - name: https
              containerPort: 443
              protocol: TCP
            - name: supervisord
              containerPort: 9001
              protocol: TCP
          startupProbe:
            # The `livenessProbe` and `readinessProbe` will be disabled until the `startupProbe` is successful.
            httpGet:
              port: http
              path: "/api/v1/health"
            failureThreshold: 10
            periodSeconds: 60
          livenessProbe:
            httpGet:
              port: http
              path: "/api/v1/health"
            failureThreshold: 8
            periodSeconds: 40
          readinessProbe:
            httpGet:
              port: http
              path: "/api/v1/health"
            failureThreshold: 9
            periodSeconds: 50
          resources:
            limits: {}
            requests: {}
          volumeMounts:
```
**Case-4**
```
(base) ➜  helm git:(update/cosmetic/changes/ce) ✗ head -n 15 values.yaml
probes:
  livenessProbe:
    api: "/api/v1/health"
    # failureThreshold: 8
    periodSeconds: 40
  readinessProbe:
    # api: "/api/v1/health"
    failureThreshold: 9
    periodSeconds: 50
  startupProbe:
    api: "/api/v1/health"
    failureThreshold: 10
    # periodSeconds: 60
## Redis parameters
redis:
(base) ➜  helm git:(update/cosmetic/changes/ce) ✗ helm install appsmith . -n goutham --create-namespace --dry-run  > temp.yaml
(base) ➜  helm git:(update/cosmetic/changes/ce) ✗ tail -n 60 temp.yaml
            {}
          image: "index.docker.io/appsmith/appsmith-ce:latest"
          imagePullPolicy: IfNotPresent
          ports:
            - name: http
              containerPort: 80
              protocol: TCP
            - name: https
              containerPort: 443
              protocol: TCP
            - name: supervisord
              containerPort: 9001
              protocol: TCP
          startupProbe:
            # The `livenessProbe` and `readinessProbe` will be disabled until the `startupProbe` is successful.
            httpGet:
              port: http
              path: "/api/v1/health"
            failureThreshold: 10
            periodSeconds: 60
          livenessProbe:
            httpGet:
              port: http
              path: "/api/v1/health"
            failureThreshold: 3
            periodSeconds: 40
          readinessProbe:
            httpGet:
              port: http
              path: "/api/v1/health"
            failureThreshold: 9
            periodSeconds: 50
          resources:
            limits: {}
            requests: {}
```

### DIFF
--- a/deploy/helm/templates/statefulset.yaml
+++ b/deploy/helm/templates/statefulset.yaml
@@ -49,16 +49,16 @@ spec:
       {{- if ((.Values.initContainer.redis).image) }}
         image: {{ .Values.initContainer.redis.image }}
       {{- else }}
-        image: "alpine"
+        image: "docker.io/library/redis:6.2-alpine"
       {{- end }}
-        command: ['sh', '-c', "apk add redis; until redis-cli -h {{ include "appsmith.fullname" . }}-redis-master.{{.Release.Namespace}}.svc.cluster.local ping; do echo waiting for redis; sleep 2; done"]
+        command: ['sh', '-c', "until redis-cli -h {{ include "appsmith.fullname" . }}-redis-master.{{.Release.Namespace}}.svc.cluster.local ping ; do echo waiting for redis; sleep 2; done"]
       {{- end }}
       {{- if .Values.mongodb.enabled }}
       - name: mongo-init-container
       {{- if ((.Values.initContainer.mongodb).image) }}
         image: {{ .Values.initContainer.mongodb.image }}
       {{- else }}
-        image:  "docker.io/bitnami/mongodb:4.4.11-debian-10-r12"
+        image:  "docker.io/bitnami/mongodb:5.0.17-debian-11-r5"
       {{- end }}
         command: ['sh', '-c', "until mongo --host appsmith-mongodb.{{.Release.Namespace}}.svc.cluster.local --eval 'db.runCommand({ping:1})' ; do echo waiting for mongo; sleep 2; done"]
       {{- end }}
@@ -78,19 +78,26 @@ spec:
             - name: supervisord
               containerPort: 9001
               protocol: TCP
+          {{- $probes := .Values.probes | default dict }}
           startupProbe:
             # The `livenessProbe` and `readinessProbe` will be disabled until the `startupProbe` is successful.
             httpGet:
-              path: /
               port: http
+              path: {{ dig "startupProbe" "api" "/api/v1/health" $probes }}
+            failureThreshold: {{ dig "startupProbe" "failureThreshold" 3 $probes }}
+            periodSeconds: {{ dig "startupProbe" "periodSeconds" 60 $probes }}
           livenessProbe:
             httpGet:
-              path: /
               port: http
+              path: {{ dig "livenessProbe" "api" "/api/v1/health" $probes }}
+            failureThreshold: {{ dig "livenessProbe" "failureThreshold" 3 $probes }}
+            periodSeconds: {{ dig "livenessProbe" "periodSeconds" 60 $probes }}
           readinessProbe:
             httpGet:
-              path: /api/v1/users/me
               port: http
+              path: {{ dig "readinessProbe" "api" "/api/v1/health" $probes }}
+            failureThreshold: {{ dig "readinessProbe" "failureThreshold" 3 $probes }}
+            periodSeconds: {{ dig "readinessProbe" "periodSeconds" 60 $probes }}
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
           volumeMounts:
--- a/deploy/helm/Chart.yaml
+++ b/deploy/helm/Chart.yaml
@@ -11,7 +11,7 @@ sources:
   - https://github.com/appsmithorg/appsmith
 home: https://www.appsmith.com/
 icon: https://assets.appsmith.com/appsmith-icon.png
-version: 2.0.2
+version: 2.0.3
 dependencies:
 - condition: redis.enabled
   name: redis
--- a/deploy/helm/templates/NOTES.txt
+++ b/deploy/helm/templates/NOTES.txt
@@ -26,5 +26,5 @@ Please update your DNS records with your domain by the address. You can use foll
   echo "Visit http://127.0.0.1:8080 to use your application"
   kubectl --namespace {{ .Release.Namespace }} port-forward $POD_NAME 8080:$CONTAINER_PORT
 
-To expose your Appsmith service to be accessible from the Internet, please refer to our docs here https://docs.appsmith.com/setup/kubernetes#expose-appsmith.
+To expose your Appsmith service to be accessible from the Internet, please refer to our docs here https://docs.appsmith.com/getting-started/setup/installation-guides/kubernetes/publish-appsmith-online.
 {{- end }}


deploy/helm/templates/configMap.yaml
```diff
@@ -3,6 +3,8 @@
 {{- $mongoUser := .Values.mongodb.auth.rootUser -}}
 {{- $mongoPassword := .Values.mongodb.auth.rootPassword -}}
 {{- $mongoServicename := .Values.mongodb.service.nameOverride -}}
+{{- $encryptionSalt := randAlphaNum 14 -}}
+{{- $encryptionPassword := randAlphaNum 14 -}}
```
sharat87, line 7: Since we are generating them when rendering this template, does that mean they'll be lost when they update the Helm charts? They won't be stored anywhere, right? Not in `values.yml`, and not in `docker.env`, so it'll be super hard to recover, right?

pratapaprasanna: 1- While updating from appsmith-ce-helm-chart-2.0.2 to  appsmith-ce-helm-chart-2.0.3 
for the first time we recommend users to give the existing encryptionSalt and encryptionPassword in values.yaml
  1.1- If the user is updating the helm charts. he usually does it with `helm upgrade` and as a practice it is recommended to keep configmap and secrets backedup.
  1.2- We can even suggest user to do `helm get values appsmith > values_existing.yaml`and preserve them before updating
2- If the values are present in configmap pod restarts also wont cause any disturbance
3- Incase, if some one uninstalls the helm charts without taking backup of salt and password and does a helm install only then the user may face issues.

deploy/helm/values.yaml
```diff
@@ -104,6 +104,20 @@ securityContext: {}
...
+    failureThreshold: 8
+    periodSeconds: 10
+  readinessProbe:
+    api: "/api/v1/health"
+    failureThreshold: 3
+    periodSeconds: 10
+  startupProbe:
+    api: "/api/v1/health"
+    failureThreshold: 3
+    periodSeconds: 60
```
sharat87, line 119: Can you put a note on why we want to make these configurable please?

sharat87 (changes requested):

deploy/helm/values.yaml
pratapaprasanna: In our cloud-helm charts these values are configurable. To make them consistent across all places we have brought in these changes. 
You want me to add a note in values.yaml as well ?

sharat87: Okay, then next question is why do we have them configurable in cloud? Can we not hard-code them there as well? 🤔

pratapaprasanna: The reason why we made it configurable is because, during times of things being unstable, we wanted a way to operate them via values.yaml and not touch the actual deployment.yaml. 

In this PR. we have hardcoded them [here](https://github.com/appsmithorg/appsmith/pull/23401/files#diff-58ba2866b40688c464b75a1e2b79b702ef338a9eacc84170aa9dcde3a4c331e6R84) . Do you think it makes sense to remove them from values.yaml ?

sharat87: Oh okay. Yeah, that the next thing I was coming at. Even if we have them configurable, I'd rather not advertise it. Let's not have it in `values.yaml`, and in fact, I would say name it as `_probes`, instead of just `probes`, to signify that this is an internal configuration section, and users aren't expected to fiddle with it.

What say?

sharat87 (changes requested):
As discussed, let's remove the `probes` in the published `values.yml` file and then merge this in.

sharat87 (changes requested):
@pratapaprasanna, there's some miscommunication happening here. I see that we've requested for review again on this PR, but the `probes` is still there in the diff. Am I missing something? 🤔

deploy/helm/templates/statefulset.yaml
```diff
@@ -81,16 +81,76 @@ spec:
...
             httpGet:
-              path: /
               port: http
+              {{- if .Values.probes }}
+              {{- if .Values.probes.startupProbe.api | quote }}
+              path: {{ .Values.probes.startupProbe.api | quote }}
+              {{- end }}
+              {{- else }}
+              path: "/api/v1/health"
+              {{- end }}
```
sharat87, line 91: We are writing an `else` block only against the first `if` statement. This means, if I understand correct, that if we have `probes: {}` in `values.yaml`, then there will be no `path:` key in the resulting configuration, right? In terms of improving resiliency here, let's ensure that our template implementation works in these three scenarios at least:

1. No `probes` key.
2. Empty `probes` key, like `probes: {}`.
3. Full probes configuration with all nested configuration defined.

Can you confirm if this template works in the second case as well please?

pratapaprasanna: Updated the comment [here](https://github.com/appsmithorg/appsmith/pull/23401#issue-1711838081)

pratapaprasanna: However, this will fail for this scenario
```
probes:
   livenessProbe:
     api: "/api/v1/health"
     periodSeconds: 40
```
then 
failureThreshold doesnt appear by default. therefore

deploy/helm/templates/statefulset.yaml
```diff
@@ -81,16 +81,76 @@ spec:
...
             # The `livenessProbe` and `readinessProbe` will be disabled until the `startupProbe` is successful.
             httpGet:
-              path: /
               port: http
+            {{- if .Values.probes }}
+            {{- if .Values.probes.startupProbe.api | quote }}
+              path: {{ .Values.probes.startupProbe.api | quote }}
+            {{- else }}
+              path: "/api/v1/health"
+            {{- end }}
```
sharat87, line 90: Can we simplify this with the [`default`](https://helm.sh/docs/chart_template_guide/function_list/#default) or [`dig`](https://helm.sh/docs/chart_template_guide/function_list/#dig) functions? It'll be easier to read and reason with I think. What say?
Similarly in other such cases as well.

```suggestion
              path: {{ dig "probes" "startupProbe" "api" "/api/v1/health" .Values | quote }}
```

sharat87 (changes requested):
@pratapaprasanna, reg including the `probes: {}`, I know we talked about this, but a follow-up question. For existing users of our helm charts, if they update their Helm chart version, it'll fail, unless they go add the `probes: {}` to their `values.yaml`, right?

If we can't use `.Values` with `dig` directly, how about a variable instead? Like in here: https://stackoverflow.com/a/59800232/151048. Like:

```
{{- $probes := .Values.probes | default dict -}}
```

Somewhere at the top of the template, and then we use `$probes` with `dig` instead of `.Values.probes`. Would that work?

Sorry to send this back yet again, but I'm just looking to reduce the friction for folks updating their Helm charts. What say?

pratapaprasanna:
> @pratapaprasanna, reg including the `probes: {}`, I know we talked about this, but a follow-up question. For existing users of our helm charts, if they update their Helm chart version, it'll fail, unless they go add the `probes: {}` to their `values.yaml`, right?
> 
> If we can't use `.Values` with `dig` directly, how about a variable instead? Like in here: https://stackoverflow.com/a/59800232/151048. Like:
> 
> ```
> {{- $probes := .Values.probes | default dict -}}
> ```
> 
> Somewhere at the top of the template, and then we use `$probes` with `dig` instead of `.Values.probes`. Would that work?
> 
> Sorry to send this back yet again, but I'm just looking to reduce the friction for folks updating their Helm charts. What say?

I couldnt get it to work. I think we have to update docs to `helm show values` and then update

sharat87 (commented):
@pratapaprasanna, can you verify if the changes in https://github.com/appsmithorg/appsmith/pull/23401/commits/bfcd2a1acc361ae25391b258edc6fe581306666a can work if `probes` is not included in `values.yaml` please?

pratapaprasanna:
> @pratapaprasanna, can you verify if the changes in [bfcd2a1](https://github.com/appsmithorg/appsmith/commit/bfcd2a1acc361ae25391b258edc6fe581306666a) can work if `probes` is not included in `values.yaml` please?

It doesnt work if the probes are commented out

pratapaprasanna:
> Seems to be working, no? Am I doing something wrong? 🤔
> 
> shot-2023-07-06-11-58-52

I have no clue what is happening now. 

I tried the same thing and I was getting the following error. 

```
(base) ➜  helm git:(update/cosmetic/changes/ce) ✗ helm install testappsmith . -n test --create-namespace --dry-run > temp.yaml
Error: INSTALLATION FAILED: template: appsmith/templates/statefulset.yaml:85:23: executing "appsmith/templates/statefulset.yaml" at <dig "startupProbe" "api" "/api/v1/health" .Values.probes>: error calling dig: interface conversion: interface {} is nil, not map[string]interface {}
```

Now, its working .

sharat87 (approved):

PR #23408, opened 2023-05-16
chore: Remove index search in omnibar tests

hetunandu:
This change removes the index search in omnibar tests to reduce flakiness


Aishwarya-U-R (commented):
Changes look ok. But please run this spec 2 more times in CI.

PR #23426, opened 2023-05-17
chore: Remove unused datasource query endpoint

nidhi-nair:
## Description
This PR removes an unused endpoint to reduce the scope of changes for multiple environments

#### Type of change

## Testing
No testing required for this change, regression would cover testing for this.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/DatasourceStructureSolutionCE.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/DatasourceStructureSolutionCE.java
@@ -1,12 +1,8 @@
 package com.appsmith.server.solutions.ce;
 
-import com.appsmith.external.models.ActionExecutionResult;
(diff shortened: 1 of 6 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/DatasourceStructureSolutionCE.java
```diff
@@ -16,7 +12,4 @@ Mono<DatasourceStructure> getStructure(
             Datasource datasource,
             boolean ignoreCache,
             String environmentName);
-
```
sondermanish, line 19: Should we also get rid of PluginExecutors getDatasourceMetadata method and it's overrides?

nidhi-nair: You're right, thanks for catching that!

sondermanish (approved):
LGTM

PR #23427, opened 2023-05-17
chore: Fix path for running cypress binary from node modules

### LINKED ISSUE #23428: [Task]: Fix path for cypress binary for running cypress test suite locally

### SubTasks

`yarn install` isn't able to symlink cypress binary to `node_modules` folder inside `app/client` folder. Thus a developer is unable to run `yarn run cytest` command to run the cypress test suite.

`yarn install` should ideally create a binary for cypress inside `node_modules` as it is added as a dependency in package.json file.

Currently, cypress may have been installed globally on developer system which might be getting invoked when a developer runs `yarn run cytest`.

A new developer who setups up their project from scratch won't be able to run the test suite locally since cypress binary is unavailable in node_modules folder inside client folder.

To find out if cypress is installed globally, run : `npm list -g | grep -i cypress`

Check this link for documentation of workspaces setup for npm packages : https://classic.yarnpkg.com/lang/en/docs/workspaces/

Sample project to demonstrate cypress binary working in workspaces setup

[workspace-exp.zip](https://github.com/appsmithorg/appsmith/files/11496573/workspace-exp.zip)

rajatagrawal:
Fixes #23428

This doesn't require a test plan or a test run, since these changes are for running the cypress test suite locally.

### DIFF
--- a/app/client/package.json
+++ b/app/client/package.json
@@ -18,7 +18,7 @@
     "test:ci": "CYPRESS_BASE_URL=https://dev.appsmith.com cypress/test.sh --env=ci",
     "eject": "react-scripts eject",
     "start-prod": "REACT_APP_ENVIRONMENT=PRODUCTION craco start",
-    "cytest": "REACT_APP_TESTING=TESTING REACT_APP_ENVIRONMENT=DEVELOPMENT craco start & ./node_modules/.bin/cypress open",
+    "cytest": "REACT_APP_TESTING=TESTING REACT_APP_ENVIRONMENT=DEVELOPMENT craco start & ../node_modules/.bin/cypress open",
     "test:unit": "jest -b --colors --no-cache --silent --coverage --collectCoverage=true --coverageDirectory='../../' --coverageReporters='json-summary'",
     "test:jest": "jest --watch",
     "generate:widget": "plop --plopfile generators/index.js",


mohanarpit:
@rajatagrawal QQ: Shouldn't node_modules be at the same level as `app/client`? Why do we have to traverse one level up?

rajatagrawal:
> @rajatagrawal QQ: Shouldn't node_modules be at the same level as `app/client`? Why do we have to traverse one level up?

I think with bundle optimization changes, we have started to use node_modules workspace feature. As part of this feature, the node modules is shared between different workspaces and the package.json file for each workspace is in their individual folder.

Workspace configuration : https://github.com/appsmithorg/appsmith/blob/d1f1a19b662b3e1ff143b9644ac251fadaf35fe6/app/package.json#L8

This is my understanding. @SatishGandham can add more to it.

SatishGandham:
Hi Rajat, when you run yarn install, node modules are installed in the root of the project, but symbolic links to npm packages are created inside each package.

So, I don't think you need this change.

Aishwarya-U-R (approved):

rajatagrawal:
> Hi Rajat, when you run yarn install, node modules are installed in the root of the project, but symbolic links to npm packages are created inside each package.
> 
> So, I don't think you need this change.
> 
> Try this locally and add cypress as dependency to one of the package and see what happens in node-modules of the package, https://classic.yarnpkg.com/lang/en/docs/workspaces/

@SatishGandham 

I went through the link shared and have a question : 

The symlinking might only work for imports in js and ts files. We are trying to invoke the cypress binary for running the cypress suite. I think accessing the binary in the parent folder might be necessary ?

SatishGandham:
@rajatagrawal , symlinks are created for binaries.

rajatagrawal:
Had an offline discussion with @SatishGandham and looks like the cypress symlink should get created inside node_modules folder in app/client but it isn't. This non creation is the root cause of the issue.

cypress may be isntalled globally on developer's systems which might be getting invoked instead of the one mentioned as a dependency in package.json file.

Created an issue for this : https://github.com/appsmithorg/appsmith/issues/23428

Additional details for the bug are in this issue above.

Closing the PR for now.

PR #23441, opened 2023-05-17
feat: widget property setters

Druthi:
## Description

- This PR adds setter methods to update widget property programmatically. 

Example:-

`Input1.setText("setter methods are cool!");`

Docs link : 
https://docs.appsmith.com/reference/widgets
For any selected widget check the `Methods` section

Fixes 

#### Type of change

- [x] Manual
- [x] Jest
- [x] Cypress
>
>
#### Test Plan
https://github.com/appsmithorg/TestSmith/issues/2409

#### Issues raised during DP testing
- [x] [Errors are not logged in the debugger](https://github.com/appsmithorg/appsmith/pull/23441#issuecomment-1564017346) separate GitHub issue  https://github.com/appsmithorg/appsmith/issues/24609
- [x] https://github.com/appsmithorg/appsmith/pull/23441#issuecomment-1564155545 ( `setVisibility("false")`  )
- [x] https://github.com/appsmithorg/appsmith/pull/23441#issuecomment-1580525843
- [x] https://github.com/appsmithorg/appsmith/pull/23441#issuecomment-1576582825 - Blocker for testing
- [x] https://github.com/appsmithorg/appsmith/pull/23441#issuecomment-1577956441
- [x] https://github.com/appsmithorg/appsmith/pull/23441#issuecomment-1577930108 - Not a issue (lint error query)
- [x] https://github.com/appsmithorg/appsmith/pull/23441#issuecomment-1593471791
- [x] https://github.com/appsmithorg/appsmith/pull/23441#issuecomment-1591440488
- [x] https://github.com/appsmithorg/appsmith/pull/23441#issuecomment-1586747864
- [x] https://github.com/appsmithorg/appsmith/pull/23441#issuecomment-1596738201
- [x] https://github.com/appsmithorg/appsmith/pull/23441#issuecomment-1598541537
- [x] https://github.com/appsmithorg/appsmith/pull/23441#issuecomment-1611413076
- [x] https://github.com/appsmithorg/appsmith/pull/23441#issuecomment-1612621567
- [ ] https://github.com/appsmithorg/appsmith/pull/23441#issuecomment-1619654507
- [ ] https://github.com/appsmithorg/appsmith/pull/23441#issuecomment-1621256722

>
>
#### Dev activity

#### QA activity:
- [x] Test plan covers all impacted features and [areas of interest](https://github.com/appsmithorg/TestSmith/wiki/Guidelines-for-test-plans/_edit#areas-of-interest)
- [x] Test plan has been peer reviewed by project stakeholders and other QA members


ramsaptami:
- [ ] Errors are not logged in the debugger
https://www.loom.com/share/a9b21cfb888f4dfca9a5191707520100

AnandiKulkarni:
@Druthi string value is getting added in boolean property, example setVisibility is set as "false"


https://github.com/appsmithorg/appsmith/assets/80756091/077d1501-70d3-4c8f-87f4-934709ff56be

Druthi:
> @Druthi string value is getting added in boolean property, example setVisibility is set as "false" myVar1: [true,false,true], trying to give array in value - func3(){ setInterval(()=>{ Button1.setVisibility(this.myVar1[1]) },3000) }
> 
>  Screen.Recording.2023-05-26.at.3.26.01.PM.mov

Looking into these

AnandiKulkarni:
When I update the value in the function below it is executed without even running the function`func11(){
		Button1.setVisibility(false).then((s)=>{Button1.isVisible? showAlert("visible"):showAlert("not visible")})
}`

https://github.com/appsmithorg/appsmith/assets/80756091/f28c1298-9543-4600-85f6-cb93172a8c63

AnandiKulkarni:
@Rishabh-Rathod I do understand that lint error can not be provided for non boolean value and should be given during run time but what about if we provide string value example - setVisible('true')/(11) we clearly see that its string and number.

AnandiKulkarni:
CurrencyInput1.seproperty 
Screenshot 2023-06-06 at 11 19 24 AM

ramsaptami:
- [ ] Select options are not updated when variable is modified within the method. Call below function on a text widget and observe the select widget options are not updated
```
let l = {label: 'c', value: 'c'}
setTimeout(() => {newWidgets.setOptions(this.myVar1.push(l))}, 4000)
```

Rishabh-Rathod:
> * [ ]  Select options are not updated when variable is modified within the method. Call below function on a text widget and observe the select widget options are not updated
> 
> ```
> let l = {label: 'c', value: 'c'}
> setTimeout(() => {newWidgets.setOptions(this.myVar1.push(l))}, 4000)
> ```

@ramsaptami 

Please note that the `array.push()` returns the length of the array and in above code we are supplying the number ( array length ) as options. 
- This is expected to fail. 

REF - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push

Druthi:
> let l = {label: 'c', value: 'c'}
> setTimeout(() => {newWidgets.setOptions(this.myVar1.push(l))}, 4000)

Heya, so here when you use `this.myVar1.push(l)` this expression returns the length of the array and not the array itself. 

There in the above case it would be `newWidgets.setOptions(1)` if `this.myVar1` has 1 element

Rishabh-Rathod:
> @Rishabh-Rathod I do understand that lint error can not be provided for non-boolean value and should be given during run time but what about if we provide string value example - setVisible('true')/(11) we clearly see that its string and number.

- Though these values are string or numeric literal, we currently don't have a linting infra in place to infer the value type and throw an error for the method if there is a mismatch. 
- This would be similar to throwing error for `showAlert([1,2,3])` where showAlert only accepts string as parameter. 


> Also is it possible to provide .then and .catch autocomplete? it could suggest user that function is thenable

Yes, this was a miss and we will add it. 
cc: @AnandiKulkarni

AnandiKulkarni:
DocumentViewer1.setURL("https://www.orimi.com/pdf-test.pdf") is not working
Screenshot 2023-06-12 at 1 00 42 PM

Druthi:
> DocumentViewer1.setURL("https://www.orimi.com/pdf-test.pdf") is not working Screenshot 2023-06-12 at 1 00 42 PM

Working for me

AnandiKulkarni:
@Druthi @Rishabh-Rathod 

- [ ] this doesnt work when table has multi select option as true	Table1.setSelectedRowIndex(2).then(s=>showAlert('success'))

https://github.com/appsmithorg/appsmith/assets/80756091/c3f4b832-9947-4def-95b1-4d50fe732740

- [ ] Table1.setSelectedRowIndex(1,2).then(s=>showAlert('success')) , only valid value for setselectedrow is number ....with above option functions runs scuuessfully and then block is executed with no error

ramsaptami:
- [ ] Table data does not set on page load
```
async onLoad () {
		// await Api1.run()
		await Table1.setData(Api1.data.results)
			.then(() => showAlert('data set on page load') )
	}
```
image

Druthi:
> * [ ]  Table data does not set on page load
@ramsaptami Checked this out in the app, the data that is being passed in setData is not the right format as it has nested objected within the Api results data. 

	```async onLoad () {
		// await Api1.run()
		const data = Api1.data.results.map(({cell, email, gender}) => { return {cell, email, gender}});
		await Table1.setData(data)
			.then(() => showAlert('data set on page load') )
	}```
	
	Instead the above code would work fine

Druthi:
> * [ ]  no validation for non-URL data for setImage method then functions contain then/catch blocks - mostly applicable to all methods and hence needs to be verified for all

This seems to be the behavior to not show the error alert when being called from the JSobject. This is the behavior even for showAlert. We'll have to a make a wider change if this is the expected behavior

Rishabh-Rathod:
> DocumentViewer1.setURL("https://www.orimi.com/pdf-test.pdf")

@AnandiKulkarni I tried this and it worked for me on the latest code.



> > * [ ]   no validation for non-URL data for setImage method then functions contain then/catch blocks - mostly applicable to all methods and hence needs to be verified for all
> 
> This seems to be the behavior to not show the error alert when being called from the JSobject. This is the behavior even for showAlert. We'll have to a make a wider change if this is the expected behavior

Adding to above point, This issue can be tracked here https://github.com/appsmithorg/appsmith/issues/24609

AnandiKulkarni:
@Rishabh-Rathod @Druthi - getting error that failed to validate - 
`myfunc2: ()=>{

	let var1 = Api1.data.products.map((s)=>{
		return{
		"brand" : s.brand,
			"name" : s.title
		}
	})
		console.log(var1)
JSONForm1.setData(var1[0])	
}`
Screenshot 2023-06-19 at 1 54 31 PM

ramsaptami:
> > * [ ]   Table data does not set on page load
> >   @ramsaptami Checked this out in the app, the data that is being passed in setData is not the right format as it has nested objected within the Api results data. Instead the below code would work fine
> 
> ```
> async onLoad () {
> 	// await Api1.run()
> 	const data = Api1.data.results.map(({cell, email, gender}) => { return {cell, email, gender}});
> 	await Table1.setData(data)
> 		.then(() => showAlert('data set on page load') )
> }
> ```

On the original snippet, `Table1.setData(Api1.data.results)` the data `Api1.data.results` is accepted on the table and displays information as expected. But why is there an expectation to transform the data during `set` action? Also, what if the user wants to display all of the information within a query or if the query contains a large number of columns?

Druthi:
> 

Got your point. I mis understood and didn't check cases other than on load. Will check this out.

ramsaptami:
- [ ] setColour does not work for switch widget. Is this applicable to switch group too?
- [ ] setData does ont work for rating widget
- [ ] setValue does not work for any of the defined widgets

Rishabh-Rathod:
/build-deploy-preview skip-tests=true recreate=true env=release

ramsaptami:
- [ ] there's no validation for boolean, numeric datatypes for JSON form setter

AnandiKulkarni:
- [ ] Button - setlable doesnt work for below code when button is disabled 

`async myFun2 () {
    Button2.isDisabled? Button1.setLabel("disabled"):Button2.setLabel("enabled")
	},`

https://github.com/appsmithorg/appsmith/assets/80756091/ab4bbf04-6611-490c-90f5-836290a1cbae


- [ ] setimage doesnt work when image in property pane has value but if the property pane image = empty setimage works well

https://github.com/appsmithorg/appsmith/assets/80756091/4bef33b8-e505-4dd7-9b78-1541c680cfe3


- [ ] setdata give error as "Invalid object at index 0" for Table6.setData(["Banana", "Orange", "Apple", "Mango"]) this is not providing any information on what array format is required 

- [ ] video is setup to play on pageload, >>  onplay another event should play but its not playing anything

Druthi:
> * Button - setlable doesnt work for below code when button is disabled

@AnandiKulkarni This one seems to be working correctly. Can you check your code again. 
- When Button2 is disabled -> Button1's label needs to change, 
- however you have two function that are running on page load that sets Button1s label, hence you see `var1` being set

AnandiKulkarni:
@Rishabh-Rathod currency input for setvalue(1234) fails and works when you add same value from property pane

https://github.com/appsmithorg/appsmith/assets/80756091/569cf6da-c348-41a6-a067-2d2cb97875cd

AnandiKulkarni:
@Rishabh-Rathod -
DatePicker1.setValue(null) is getting accepted , can we error out. Also can we checkboxgroup - remove setvalue

Rishabh-Rathod:
> @Rishabh-Rathod -

> * [ ]  DatePicker1.setValue(null) is getting accepted ,

Currently, the widget does accept null as a value to reset. If we expect an error, we will need to fix this on widget level. Could be raised as a platform-level issue.

Rishabh-Rathod:
@Aishwarya-U-R Could you please review the automation test for this PR?

app/client/cypress/e2e/Regression/ClientSide/Autocomplete/Autocomplete_setters_spec.ts
```diff
@@ -0,0 +1,71 @@
...
+      completeReplace: true,
+      toRun: false,
+      shouldCreateNewJSObj: true,
+      prettify: false,
+    });
+
+    _.agHelper.GetNClick(_.jsEditor._lineinJsEditor(5));
+    _.agHelper.TypeText(_.locators._codeMirrorTextArea, "Button1");
+
+    _.agHelper.GetNAssertElementText(
```
Aishwarya-U-R, line 30: Can `agHelper.GetElementsNAssertTextPresence` be used? I have seen this particular hints validation fail.

AnandiKulkarni:
- [ ] Rating1.setValue(-1) is not getting error out, no value is set but not getting error out

app/client/cypress/e2e/Regression/ClientSide/Widgets/WidgetPropertySetters_spec.ts
```diff
@@ -0,0 +1,185 @@
...
+  },
+  {
+    name: "setData",
+    property: "tableData",
+    widget: WIDGET.TABLE,
+    actionBinding: "{{Table1.setData([{name: 'test'}])}}",
+    valueBinding: "{{JSON.stringify(Table1.tableData)}}",
+    expectedValue: '[{"name":"test"}]',
+  },
+];
```
Rishabh-Rathod, line 99: @Aishwarya-U-R These are the setter methods list, we have covered for the basic functionality test.

app/client/cypress/e2e/Regression/ClientSide/Widgets/WidgetPropertySetters_spec.ts
```diff
@@ -0,0 +1,185 @@
...
+describe("Linting warning for setter methods", function () {
+  it("Lint error when setter is used in a data field", function () {
+    _.EntityExplorer.DragDropWidgetNVerify(WIDGET.BUTTON, 200, 200);
+    _.AggregateHelper.GetNClick(getWidgetSelector(WIDGET.BUTTON));
+    _.PropertyPane.TypeTextIntoField("Label", "{{Button1.setLabel('Hello')}}");
+
+    //Mouse hover to exact warning message
+    _.AggregateHelper.GetElement(_.CommonLocators._lintErrorElement).trigger(
+      "mouseover",
+    );
```
Aishwarya-U-R, line 158: Can we use `agHelper.HoverElement()`?

app/client/cypress/e2e/Regression/ClientSide/Widgets/WidgetPropertySetters_spec.ts
Aishwarya-U-R: Got it!

app/client/cypress/e2e/Regression/ClientSide/Widgets/WidgetPropertySetters_spec.ts
Rishabh-Rathod: cc: @Druthi

app/client/cypress/e2e/Regression/ClientSide/Autocomplete/Autocomplete_setters_spec.ts
```diff
@@ -0,0 +1,71 @@
+import * as _ from "../../../../support/Objects/ObjectsCore";
```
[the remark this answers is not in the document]
Rishabh-Rathod: @Aishwarya-U-R  The [namespace import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#namespace_import) pattern was taken from another currently present cypress test. What is the reasoning behind the suggestion for using [named import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#named_import) rather than namespace one? 

This would help us understand the reasoning behind it. So, we only follow the pattern according to our reasoning and make it a coding guideline for other PRs too.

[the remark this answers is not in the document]
Aishwarya-U-R: @Rishabh-Rathod - You might see a few cypress spec which imports all (we are in the process of updating them), but we realised ObjectsCore is growing day by day. Why to import it entirely when spec is going to use only 20% of objects from it. Hence to minimise the memory leak (if there may be any, as we are unsure of how exactly Cypress bundles the specs when running them) we recommend importing only needed objects from ObjectsCore.

app/client/cypress/e2e/Regression/ClientSide/Autocomplete/Autocomplete_setters_spec.ts
Rishabh-Rathod: In that case, shouldn't we fix the `GetNAssertElementText` instead of replacing it with `GetElementsNAssertTextPresence`? 

GetNAssertElementText is used 214 times whereas GetElementsNAssertTextPresence is recently introduced and only used 4 times.

Aishwarya-U-R: No, issue is not with `GetNAssertElementText`. Its because of the index passed to this method. The auto suggest text passed to be verified is not always present at that index & hence I have seen failures & replaced with `GetElementsNAssertTextPresence` - which checks if passed text is present in hints (any index). If you are sure it will always come in same index & case will not fail, please go ahead using `GetNAssertElementText` :)

app/client/cypress/e2e/Regression/ClientSide/Widgets/WidgetPropertySetters_spec.ts
Rishabh-Rathod: Is there a guideline that the developer could follow for which helper method to use? 
Mostly, we take inspiration from other cypress test cases and we ideally want to be in a stage where we could use the appropriate helper method while writing the test for 1st time.

app/client/cypress/e2e/Regression/ClientSide/Autocomplete/Autocomplete_setters_spec.ts
Rishabh-Rathod: Got it makes more sense now. Thanks for explaining.

Rishabh-Rathod: Made the changes suggested above.

app/client/cypress/e2e/Regression/ClientSide/Widgets/WidgetPropertySetters_spec.ts
Rishabh-Rathod: Made suggested changes.

app/client/cypress/e2e/Regression/ClientSide/Autocomplete/Autocomplete_setters_spec.ts
```diff
@@ -0,0 +1,56 @@
...
+      "Button1.setColor()",
+    );
+
+    //For table widget
+    entityExplorer.DragDropWidgetNVerify(draggableWidgets.TABLE, 500, 300);
+    entityExplorer.SelectEntityByName("JSObject1");
+    agHelper.GetNClick(jsEditor._lineinJsEditor(5));
+    agHelper.RemoveCharsNType(locators._codeMirrorTextArea, 7, "Table1.set");
+
+    agHelper.GetElementsNAssertTextPresence(locators._hints, "setData()");
```
Aishwarya-U-R, line 46: Is only `setData` available for Table? Can we validate all set auto suggest here?

app/client/cypress/e2e/Regression/ClientSide/Autocomplete/Autocomplete_setters_spec.ts
```diff
@@ -0,0 +1,56 @@
...
+  });
+
+  it("2. Check if setters are present in autocomplete for widgets in property Pane", () => {
+    entityExplorer.DragDropWidgetNVerify(draggableWidgets.INPUT_V2, 200, 600);
+
+    entityExplorer.SelectEntityByName("Button1");
+    propPane.EnterJSContext("onClick", "{{Input1.set", true, false);
+    agHelper.GetElementsNAssertTextPresence(locators._hints, "setDisabled()");
```
Aishwarya-U-R, line 54: Same query for this also? is only `setDisabled()` present for Input?

app/client/cypress/e2e/Regression/ClientSide/Widgets/WidgetPropertySetters_spec.ts
```diff
@@ -0,0 +1,187 @@
...
+
+      it("2. Verify the updated value", () => {
+        agHelper.GetNClick(getWidgetSelector(WIDGET.BUTTON));
+
+        agHelper.GetText(getWidgetSelector(WIDGET.TEXT)).then(($label) => {
+          expect($label).to.eq(expectedValue);
+        });
+      });
+
+      it("3. Delete all the widgets on canvas", () => {
```
Aishwarya-U-R, line 138: Can this be added as part of `afterEach`? Reason is to reduce the number of `it` blocks.

app/client/cypress/e2e/Regression/ClientSide/Widgets/WidgetPropertySetters_spec.ts
```diff
@@ -0,0 +1,187 @@
...
+          valueBinding,
+        );
+      });
+
+      it("2. Verify the updated value", () => {
+        agHelper.GetNClick(getWidgetSelector(WIDGET.BUTTON));
+
+        agHelper.GetText(getWidgetSelector(WIDGET.TEXT)).then(($label) => {
+          expect($label).to.eq(expectedValue);
+        });
```
Aishwarya-U-R, line 135: Can we add this to above `it` block itself? Test seems clear, it should be ok to combine without causing issue to readibility.

app/client/cypress/e2e/Regression/ClientSide/Autocomplete/Autocomplete_setters_spec.ts
Rishabh-Rathod: We have multiple setters for Input and Table but the logic for each setter is similar. It is more important to test different setter methods ( irrespective of the widget ) than covering setter method for each widget. 

Hence, the test was return to cover most of the important setter method present.

app/client/cypress/e2e/Regression/ClientSide/Autocomplete/Autocomplete_setters_spec.ts
Rishabh-Rathod: same as https://github.com/appsmithorg/appsmith/pull/23441#discussion_r1255622821

app/client/cypress/e2e/Regression/ClientSide/Widgets/WidgetPropertySetters_spec.ts
```diff
@@ -0,0 +1,185 @@
+import {
```
Aishwarya-U-R, line 1: This test is beautiful @Rishabh-Rathod 🎉

app/client/cypress/e2e/Regression/ClientSide/Widgets/WidgetPropertySetters_spec.ts
```diff
@@ -0,0 +1,185 @@
...
+
+        agHelper.GetNClick(getWidgetSelector(WIDGET.TEXT)).click();
+        agHelper.GetNClick(propPane._deleteWidget);
+      });
+    });
+  },
+);
+
+describe("Linting warning for setter methods", function () {
+  it("Lint error when setter is used in a data field", function () {
```
Aishwarya-U-R, line 152: Please index `it`s as it will help to debug a failing case from Cypress dashboard.

app/client/cypress/e2e/Regression/ClientSide/Autocomplete/Autocomplete_setters_spec.ts
Aishwarya-U-R: @Rishabh-Rathod - No, question was to add checks to validate presence of all newly introduced setter methods in hints for Table & Input - thus checking the discoverability is intact. 
I understand the reason for checking functionality for most import setters for most commonly used widgets included in this PR.

app/client/cypress/e2e/Regression/ClientSide/Autocomplete/Autocomplete_setters_spec.ts
Rishabh-Rathod: Will add a test for other setter methods in a separate PR.

Aishwarya-U-R (commented):
Approving Cypress changes with agreement that Autocomplete pending review comments will be addressed in next PR.

arunvjn (approved):

PR #23539, opened 2023-05-19
Apply env replacements to view.html and edit.html

sharat87:
WIP

### DIFF
--- a/deploy/docker/scripts/run-nginx.sh
+++ b/deploy/docker/scripts/run-nginx.sh
@@ -70,19 +70,24 @@ fi
 
 bash "$APP_TEMPLATE" "${APPSMITH_CUSTOM_DOMAIN-}" > /etc/nginx/sites-available/default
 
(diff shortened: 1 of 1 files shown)


SatishGandham (approved):

SatishGandham:
@sharat87 dont we have to make this change on gzipped versions of these files?

sharat87:
@SatishGandham, yes, indeed, but not a blocker actually.

We should be doing that for `index.html.gz` as well, but the current situation is, we aren't _using_ those gz files at all. NGINX does the gzipping in-memory, from what I know, which is why we haven't seen this issue so far, even though we've only been doing the replacements in `index.html`. 🙂

We'll pick that up as a separate PR.
Sound okay?

SatishGandham:
I created an issue about gzip not working on prod long back. 

I think this is working because of that issue. We should solve the two together, this PR is fine for now.

PR #23546, opened 2023-05-19
fix: Re-gzip html files after env variables substitution

sharat87:
We replace env variables in `index.html`, `view.html` and `edit.html`, just before start NGINX. But the `.gz` versions of these files don't have these changes applied to them. This PR gzips those HTML files again, after the substitutions are applied.

From https://github.com/appsmithorg/appsmith/pull/23539#issuecomment-1554509537.

### DIFF
--- a/deploy/docker/scripts/run-nginx.sh
+++ b/deploy/docker/scripts/run-nginx.sh
@@ -84,6 +84,9 @@ apply-env-vars() {
   )
   fs.writeFileSync("'"$served"'", content)
   '
+  pushd "$(dirname "$served")"
+  gzip --keep --force "$(basename "$served")"
+  popd
 }
 
 apply-env-vars /opt/appsmith/index.html.original /opt/appsmith/editor/index.html


SatishGandham:
@sharat87 , won't serving already generated gzipped and brotli compressed files be more efficient? 

Does it hinder our CDN efforts?

Does nginx cache the generated compressed files?

sharat87:
@SatishGandham, sorry, can you elaborate:

> won't serving already generated gzipped and brotli compressed files be more efficient?

The _already generated_ files don't have the env variable substitutions, right? That's why we need to gzip the HTML files again, _after_ we apply the substitutions, no?

> Does it hinder our CDN efforts?

I don't think so. Since, the CDN will just get the files that NGINX serves, right? Gzipped, or not?

> Does nginx cache the generated compressed files?

I'm not sure about this. I think yes, but I also think it'll notice if the file has changed on disk. But this should also be not very relevant, since we don't even start NGINX before all the replacements and gzipping is done. In fact, we start NGINX at the very end of this script.

SatishGandham:
Let me rephrase. Will it be more efficient if we insert the variable and generate the zip files instead of making nginx do it at run time. 

We are making nginx disregard all the gzip files we generate during build to help with our changes to just three html files.

SatishGandham (approved):
Thank you :)

PR #23569, opened 2023-05-21
chore: Added utility file to check for query compatibility

nidhi-nair:
## Description
This PR introduces a utility to check for whether certain migration queries can be run in an optimized way for a particular installation or not.

#### Type of change

No testing required, this utility is not being used yet, meant for syncing with EE.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/utils/CompatibilityUtils.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/utils/CompatibilityUtils.java
@@ -1,0 +1,39 @@
+package com.appsmith.server.migrations.utils;
+
+import com.appsmith.external.models.BaseDomain;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.data.mongodb.UncategorizedMongoDbException;
+import org.springframework.data.mongodb.core.MongoTemplate;
+import org.springframework.data.mongodb.core.query.Query;
+
+@Slf4j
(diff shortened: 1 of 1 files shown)


nidhi-nair:
Skipping ok to test because this util is not being consumed.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/utils/CompatibilityUtils.java
```diff
@@ -0,0 +1,39 @@
...
+     * @param clazz         The collection this query will be run on
+     * @param <T>           The BaseDomain type this collection maps to
+     * @return A query that is guaranteed to run on this instance, with or without optimization
+     */
+    public static <T extends BaseDomain> Query optimizeQueryForNoCursorTimeout
+    (MongoTemplate mongoTemplate, Query originalQuery, Class<T> clazz) {
+        try {
+            log.debug("Check if performant query can be used.");
+            Query queryBatchedPerformant = Query.of(originalQuery).noCursorTimeout();
+            Query queryBatchedPerformantLimit1 = Query.of(queryBatchedPerformant).limit(1);
```
sondermanish, line 27: QQ: this means that we are performing query on just one record, right?
So our query parameter be such that, next time the record modified is not included?

nidhi-nair: Yes that's correct

sondermanish (approved):

PR #23606, opened 2023-05-22
chore: add units tests for wds components

jsartisan:


app/client/packages/design-system/headless/src/components/Field/Field.tsx
```diff
@@ -27,7 +26,6 @@ export const Field = forwardRef((props: FieldProps, ref: FieldRef) => {
     validationState,
     wrapperClassName,
     wrapperProps = {},
-    ...otherProps
```
jsartisan, line 30: Note: this was a miss from my end. This prop is not required to be passed to field.

KelvinOm (approved):

PR #23619, opened 2023-05-23
fix: Fix crash on passing invalid mark option to category slider widget

### LINKED ISSUE #23228: [Bug]: Error while accessing Marks on NumberSlider WidgetTypeError: Cannot read properties of undefined (reading 'length')
Sentry Issue: [APPSMITH-67T](https://appsmith.sentry.io/issues/4104664722/?referrer=github_integration)

```
TypeError: Cannot read properties of undefined (reading 'length')
  at items (widgets/NumberSliderWidget/component/Marks.tsx:120:57)
  at Array.map (<anonymous>)
  at Marks (widgets/NumberSliderWidget/component/Marks.tsx:92:25)
  at next (RouteParamsMiddleware.ts:126:12)
...
(45 additional frame(s) were not displayed)
```

rajatagrawal:
Fixes #23228

Steps to Repro:

1. Dnd Category Slider Widget
2. Toggle JS Mode in options field
3. Enter `[[]]`
4. Observe the crash

Expected behavior : The widget shouldn't crash

### DIFF
--- a/app/client/cypress/e2e/Regression/ClientSide/Widgets/Sliders/CategroySlider_spec.ts
+++ b/app/client/cypress/e2e/Regression/ClientSide/Widgets/Sliders/CategroySlider_spec.ts
@@ -72,4 +72,12 @@ describe("Category Slider spec", () => {
       expect($label).to.eq("sm");
     });
   });
+
+  it("does not crash if an invalid mark option is passed", function () {
+    cy.get(".t--property-control-options .t--js-toggle").first().click();
+    cy.updateCodeInput(".t--property-control-options", "[[]]");
+    cy.get(".t--widget-categorysliderwidget")
+      .contains("Oops, Something went wrong.")
+      .should("not.exist");
+  });
 });
--- a/app/client/src/widgets/CategorySliderWidget/validations.ts
+++ b/app/client/src/widgets/CategorySliderWidget/validations.ts
@@ -12,7 +12,7 @@ export function optionsCustomValidation(
     if (options.length < 2) {
       return {
         isValid: false,
-        parsed: options,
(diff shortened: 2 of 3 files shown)


app/client/src/widgets/CategorySliderWidget/validations.ts
```diff
@@ -12,7 +12,7 @@ export function optionsCustomValidation(
     if (options.length < 2) {
       return {
         isValid: false,
-        parsed: options,
+        parsed: [],
```
keyurparalkar, line 15: Why are we changing the parsed value here? Wouldn't this change the existing behaviour were if options are less than 2 we still show the options in the evaluated popup.

rajatagrawal: The bug happens when we pass these options. If the options are passed, they may contain invalid value which is getting processed by the widget and crashing it.

All the other validation cases below are passing empty options array if any of the option is invalid

keyurparalkar (approved):

chandannkumar:
Tested this PR and LGTM

app/client/cypress/e2e/Regression/ClientSide/Widgets/Sliders/CategroySlider_spec.ts
```diff
@@ -72,4 +72,12 @@ describe("Category Slider spec", () => {
       expect($label).to.eq("sm");
     });
   });
+
+  it("does not crash if an invalid mark option is passed", function () {
```
Aishwarya-U-R, line 76: Index the test case & start writing new tests in TS format please.

PR #23634, opened 2023-05-23
test: updated location to drop into list widget

NandanAnantharamu:
- Trying to fix by changing the location that we have been dropping into list widget


NandanAnantharamu:
closing as this is not flaky anymore

PR #23666, opened 2023-05-23
chore: Datasource configuration migration to storage scaffold

nidhi-nair:
This PR is a skeleton for DS migration to storage

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration012TransferToDatasourceStorage.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration012TransferToDatasourceStorage.java
@@ -21,9 +21,8 @@ import static org.springframework.data.mongodb.core.query.Criteria.where;
 import static org.springframework.data.mongodb.core.query.Query.query;
 
 /**
- * In this migration, we are simply moving the configurations to the new collection.
- * We will follow this up with another migration that deletes the existing configurations
- * for datasources, simply to allow for a rollback mechanism for this migration.
+ * In this migration, we are moving configurations from each valid datasource
+ * to a new collection. The datasource will cease to have configurations after this point
  */
 @Slf4j
 //@ChangeUnit(order = "012", id = "migrate-configurations-to-data-storage", author = " ")
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration012TransferToDatasourceStorage.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration012TransferToDatasourceStorage.java
@@ -4,7 +4,9 @@ import com.appsmith.external.models.Datasource;
 import com.appsmith.external.models.DatasourceStorage;
 import com.appsmith.external.models.QDatasource;
 import com.appsmith.server.constants.FieldName;
+import com.appsmith.server.domains.Workspace;
 import com.appsmith.server.migrations.solutions.DatasourceStorageMigrationSolution;
+import com.appsmith.server.migrations.utils.CompatibilityUtils;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.dao.DuplicateKeyException;
 import org.springframework.data.mongodb.core.MongoTemplate;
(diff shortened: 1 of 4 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration012TransferToDatasourceStorage.java
```diff
@@ -0,0 +1,153 @@
...
+import static com.appsmith.server.repositories.ce.BaseAppsmithRepositoryCEImpl.fieldName;
+import static org.springframework.data.mongodb.core.query.Criteria.where;
+import static org.springframework.data.mongodb.core.query.Query.query;
+
+/**
+ * In this migration, we are moving configurations from each valid datasource
+ * to a new collection. The datasource will cease to have configurations after this point
+ */
+@Slf4j
+//@ChangeUnit(order = "012", id = "migrate-configurations-to-data-storage", author = " ")
```
sondermanish, line 30: QQ: So the Idea is to ship this commented out migration and then make changes in EE for solution and then uncomment this line in CE?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration012TransferToDatasourceStorage.java
```diff
@@ -0,0 +1,153 @@
...
+                        mongoTemplate.insert(datasourceStorage);
+                    } catch (DuplicateKeyException e) {
+                        log.warn("Looks like the datasource storage already exists for datasource id: {}",
+                                datasource.getId());
+                        log.warn("We will attempt to reset the datasource again.");
+                    }
+
+                    // Once the datasource storage exists, delete the older config inside datasource
+                    // And set `hasDatasourceStorage` value to true
+                    mongoTemplate.updateFirst(
```
sondermanish, line 118: I was wondering whether we should use a separate migration in next release cycle altogether to get rid of datasource configuration? Gives a fallback in case we miss dsStorage writes?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/solutions/ce/DatasourceStorageMigrationSolutionCE.java
```diff
@@ -0,0 +1,18 @@
...
+
+public class DatasourceStorageMigrationSolutionCE {
+
+    public Map<String, String> getDefaultEnvironmentsMap(MongoTemplate mongoTemplate) {
+        return Map.of();
+    }
+
+    public String getEnvironmentIdForDatasource(Map<String, String> wsIdToEnvIdMap,
+                                                String workspaceId) {
+        return FieldName.UNUSED_ENVIRONMENT_ID;
```
sondermanish, line 16: QQ: for EE the Id would be default environmentId?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration012TransferToDatasourceStorage.java
nidhi-nair: Yes, that's correct

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration012TransferToDatasourceStorage.java
nidhi-nair: I did consider this. But this is the last step, so in case there is anything that goes wrong before this step, we anyways would not have reset the flag or gotten rid of the configuration, the migration is repeatable n number of times.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/solutions/ce/DatasourceStorageMigrationSolutionCE.java
nidhi-nair: Yes, we'll be returning the value from the map

sondermanish (approved):
This Migration is commented out, this requires an implementation in EE and then the migration could be made live

sondermanish:
Ci is not required for this

PR #23717, opened 2023-05-25
chore: Make tooltip more accessiblity

### LINKED ISSUE #23587: Improve a11y for tooltip
When a tooltip is used with disabled buttons, we have to make sure that the text of the tooltip is accessible via screen reader. It is necessary to analyse the possibility of using attributes `aria-label` and `aria-labelledby`.

jsartisan:
Fixes #23587


app/client/packages/design-system/headless/src/components/Button/Button.tsx
```diff
@@ -14,22 +14,51 @@ export type ButtonRef = React.Ref<HTMLButtonElement>;
 type ButtonRefObject = React.RefObject<HTMLButtonElement>;
 
 export const Button = forwardRef((props: ButtonProps, ref: ButtonRef) => {
-  const { autoFocus, children, className, isDisabled } = props;
-  const { buttonProps, isPressed } = useButton(props, ref as ButtonRefObject);
+  const { autoFocus, children, className, isDisabled, visuallyDisabled } =
```
KelvinOm, line 17: Could you add comment in interface  for `visuallyDisabled` for what we are using this prop. This comment also will available in argsTable in storybook.

app/client/packages/design-system/headless/src/components/Button/Button.tsx
```diff
@@ -14,22 +14,51 @@ export type ButtonRef = React.Ref<HTMLButtonElement>;
...
+
+  if (props.visuallyDisabled) {
+    computedProps = {
+      ...props,
+      isDisabled: false,
+      // disabling click/press events
+      onPress: undefined,
+      onPressStart: undefined,
+      onPressEnd: undefined,
+      onPressChange: undefined,
```
KelvinOm, line 59: What about these events?

```
onKeyDown
onKeyUp
onPressUp
```

app/client/packages/design-system/theming/src/utils/TokensAccessor/LightModeTheme.ts
```diff
@@ -46,7 +46,7 @@ export class LightModeTheme implements ColorModeTheme {
       bgAccentSubtleActive: this.bgAccentSubtleActive.toString({
         format: "hex",
       }),
-      bgAssistive: this.bgAssistive.toString({ format: "hex" }),
+      bgAssistive: this.bgAssistive.toString(),
```
KelvinOm, line 49: Why?

jsartisan: I just did it because you were doing it at other places :P

KelvinOm: Where? :flushed: 

We have two types of colors here that we return from the function:
1. Hard coded hex(without toString)
2. Color object (with toString)

KelvinOm (approved):

PR #23726, opened 2023-05-25
chore: Add missing autocomplete definition declarations

rajatagrawal:
There were some autocomplete definitions declaration I missed last time I refactored the code. Adding them now.

### DIFF
--- a/app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.test.ts
+++ b/app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.test.ts
@@ -755,7 +755,8 @@ describe("#generate", () => {
           },
           Canvas1: {},
           Canvas2: {
-            entityDefinition: "",
+            entityDefinition:
+              "backgroundColor: Canvas2.backgroundColor,isVisible: Canvas2.isVisible",
             rowIndex: 0,
             metaWidgetId: "qpgtpiw3cu",
             metaWidgetName: "Canvas2",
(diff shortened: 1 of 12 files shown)


sbalaji1192 (approved):

app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.test.ts
```diff
@@ -1010,7 +1011,7 @@ describe("#generate", () => {
...
+      "{{\n      {\n        \n          Image1: { image: Image1.image,isVisible: Image1.isVisible }\n        ,\n          Text1: { isVisible: Text1.isVisible,text: Text1.text }\n        ,\n          Text2: { isVisible: Text2.isVisible,text: Text2.text }\n        ,\n          List6: { backgroundColor: List6.backgroundColor,isVisible: List6.isVisible,itemSpacing: List6.itemSpacing,selectedItem: List6.selectedItem,selectedItemView: List6.selectedItemView,triggeredItem: List6.triggeredItem,triggeredItemView: List6.triggeredItemView,listData: List6.listData,pageNo: List6.pageNo,pageSize: List6.pageSize,currentItemsView: List6.currentItemsView }\n        ,\n          Canvas2: { backgroundColor: Canvas2.backgroundColor,isVisible: Canvas2.isVisible }\n        \n      }\n    }}";
```
ashit-rath, line 1014: Why is this added? Was this test failing?

rajatagrawal: yes, in my changes, I have started to return default autocomplete definitions inherited from base widget for canvas widget as well. Hence this test started failing.

ashit-rath: Ideally in that binding the Canvas shouldn't appear. 

I see that you have added autocomplete for Canvas but Canvas widgets are internal only and shouldn't have any autocomplete. I think because of that it is showing up here.
As the CanvasWidget extends from Container, all the autocomplete of Container are inherited to the Canvas.

Is there any requirement to add autocomplete for Canvas widget?

rajatagrawal: Theoretically, it should have autocomplete properties of container widget since it inherits from it. But if it is not required, we can override the autocomplete definitions in canvas widget to be an empty object. Please suggest the best approach.

ashit-rath: I agree, we should just return an empty object or `undefined`.

PR #23768, opened 2023-05-26
feat: disabling the toast messages on view mode without debug flag

### LINKED ISSUE #23603: [Bug]: Remove toast errors on the view mode of application

### Description

Screenshot 2023-05-19 at 9 01 36 AM

### Steps To Reproduce

1. Create an application with multiple apis having run on load enabled
2. Open the view mode of application
3. Observe too many error toasts

### Public Sample App

_No response_

### Environment

Production

### Issue video log

_No response_

### Version

Both

### LINKED ISSUE #23605: [Bug]: Enable debug mode for displaying error toasts for debug flag on view mode

### Description

Show error toasts when there is `debug=true` in the url in the view mode else silently display them on console

### Steps To Reproduce

1. Create an application with multiple apis having run on load enabled
2. Open the view mode of application
3. Observe too many error toasts

### Public Sample App

_No response_

### Environment

Production

### Issue video log

_No response_

### Version

Both

AmanAgarwal041:
## Description
When the application is in published mode, this Pr offers the ability to disable the error toast. Only the toasts that the user initiated are displayed; all others are disabled and messages are logged in the console.
When the view mode url has `debug=true` as one of the query params, then all the error toasts are visible.

Fixes #23605
Fixes #23603 

>
>
>
- [x] Manual
- [ ] Jest
- [ ] Cypress
>
>
#### Dev activity

#### QA activity:


app/client/src/sagas/ToastSagas.ts
```diff
@@ -0,0 +1,37 @@
...
+export default function* showToast(
+  message: string,
+  options?: ToastProps,
+  extraOtions?: ExtraOptions,
+) {
+  const appMode: APP_MODE | undefined = yield select(getAppMode);
+  const urlObject = new URL(window?.location?.href);
+  const debugFlag = urlObject?.searchParams?.get("debug");
+  const debug = debugFlag === "true" || debugFlag;
+  if (appMode === APP_MODE.PUBLISHED && !debug && !extraOtions?.forceDisplay) {
```
Irongade, line 29: why are these conditions logical AND? For example, extraOptions.forceDisplay should force the toast to be shown irrespective of what mode the app is in.

AmanAgarwal041: In any case if the forceDisplay is `true` then it will make the if condition false. 
`if (!extraOtions?.forceDisplay) { if (appMode === APP_MODE.PUBLISHED && !debug) { return }}` will be same as above

Irongade (dismissed):

app/client/src/sagas/ToastSagas.ts
```diff
@@ -0,0 +1,34 @@
...
+
+type ExtraOptions = {
+  // This enables showing of toast no matter the conditions
+  forceDisplay?: boolean;
+};
+
+/**
+ * Shows toast
+ * @param message
+ * @param [options] These are toast props that toast from design-sytem(react-toastify) library takes
```
tanvibhakta, line 15: In which cases should a developer be using showToast and which case should they be using toast from design-system directly?

tanvibhakta (approved):

app/client/src/sagas/ToastSagas.ts
AmanAgarwal041: So this Saga was introduced to add some conditions before showing toasts to the users. You can directly call this functionality by passing `forceDisplay=true` in the options to bypass the conditions for showing the toast. We can even replace all the toast being used so that there is single file to use toast, if there are any future changes for the toast, those can be changed at one place only.

AmanAgarwal041: Can we resolve this conversation? Let me know your thoughts if you have a different opinion @tanvibhakta

tanvibhakta: If this is the recommended way of using toast, then yes, it should all come from a single file. Might it make sense to have these changes to the api within the design-system itself? That way you will be saved the effort of updating the calling function of toast everywhere.

AmanAgarwal041: - I don't know if this is a recommended way but seeing the changes that ads cascaded. I think we can lower down the future efforts if there is a big change like ads. 
- Addition to this, I wanted to add a check for displaying toast in debug mode at one place, so figured out this solution.

rohitagarwal88: @AmanAgarwal041: It seems that you are uncertain about whether this is the correct approach or not. I suggest that you, @tanvibhakta, and @albinAppsmith should carefully determine the right approach to prevent any future modifications and avoid rushing into promotion. Significant changes like ADS occur only once or twice in a company's history, and that too after years. Therefore, let's avoid making decisions solely based on the current situation.
cc @rohan-arthur

AmanAgarwal041: @rohitagarwal88 I am sure about implementing this solution to restrict debug mode toast appear, as this implementation actually restrains the repetition of the logic. This implementation is good for debug mode but its not compulsory to use it wherever we are using toasts.

Irongade (approved):

Issue #23522, opened 2023-05-19
[Feature]: Google CloudRun support for running Appsmith


sharat87:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

Appsmith supports running on ECS with/without Fargate, but we don't have official instructions for running on GCP's CloudRun offering. That's what this issue is focused on.

### Why should this be worked on?

Users who want an instance with minimal management overhead, and don't already have a Kubernetes cluster, would prefer a container service like ECS or CloudRun. For folks invested in GCP cloud, this would be CloudRun.

PR #23829, opened 2023-05-29
chore: update entrypoint to deploy appsmith on cloudrun

pratapaprasanna:
Fixes: [23522](https://github.com/appsmithorg/appsmith/issues/23522)

### DIFF
--- a/deploy/docker/entrypoint.sh
+++ b/deploy/docker/entrypoint.sh
@@ -1,20 +1,6 @@
 #!/usr/bin/env bash
+
 set -e
-
-if [[ -n "$FILESTORE_IP_ADDRESS" ]]; then
-
-        ## Trim APPSMITH_FILESTORE_IP and FILE_SHARE_NAME
-        FILESTORE_IP_ADDRESS="$(echo "$FILESTORE_IP_ADDRESS" | xargs)"
-        FILE_SHARE_NAME="$(echo "$FILE_SHARE_NAME" | xargs)"
-
-        echo "Running appsmith for cloudRun"
-        echo "Mounting File Sytem"
-        mount -t nfs -o nolock "$FILESTORE_IP_ADDRESS:/$FILE_SHARE_NAME" /appsmith-stacks
-        echo "Mounted File Sytem"
-        echo "Setting HOSTNAME for Cloudrun"
-        export HOSTNAME="cloudrun"
-fi
-
 # ip is a reserved keyword for tracking events in Mixpanel. Instead of showing the ip as is Mixpanel provides derived properties.
 # As we want derived props alongwith the ip address we are sharing the ip address in separate keys
 # https://help.mixpanel.com/hc/en-us/articles/360001355266-Event-Properties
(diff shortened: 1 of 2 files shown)


sharat87:
Unmarking myself for review, as discussed, since this is still WIP.

deploy/docker/entrypoint.sh
sharat87, line 1: 1. Let's please keep `set -e` and the Segment call as the very first things that run in the entrypoint. Can we move this to below that please?
2. Let's also change the indentation to 2-spaces to keep it consistent with the rest of this script please.

deploy/docker/entrypoint.sh
```diff
@@ -1,5 +1,19 @@
 #!/usr/bin/env bash
 
+if [ $FILESTORE_IP_ADDRESS ];
+then
```
sharat87, line 4: The `if` check is incorrect. I think we need `[[ -n "$FILESTORE_IP_ADDRESS" ]]` to check if it's non-empty.

```suggestion
if [[ -n "$FILESTORE_IP_ADDRESS" ]]; then
```

deploy/docker/entrypoint.sh
```diff
@@ -1,5 +1,19 @@
...
+        ## Trim APPSMITH_FILESTORE_IP and FILE_SHARE_NAME
+        FILESTORE_IP_ADDRESS=$(echo "$FILESTORE_IP_ADDRESS" | xargs)
+        FILE_SHARE_NAME=$(echo "$FILE_SHARE_NAME" | xargs)
```
sharat87, line 7: Quotes please. 🙂

```suggestion
        ## Trim APPSMITH_FILESTORE_IP and FILE_SHARE_NAME
        FILESTORE_IP_ADDRESS="$(echo "$FILESTORE_IP_ADDRESS" | xargs)"
        FILE_SHARE_NAME="$(echo "$FILE_SHARE_NAME" | xargs)"
```

deploy/docker/entrypoint.sh
```diff
@@ -1,5 +1,19 @@
...
+
+        echo "Installing appsmith for cloudRun"
```
sharat87, line 9: Not really _installing_, right? We're just running, no? 🤔

```suggestion
        echo "Running Appsmith for CloudRun"
```

sharat87 (changes requested):

deploy/docker/entrypoint.sh
pratapaprasanna: May be deploying appsmith for cloudrun ?

sharat87 (changes requested):
@pratapaprasanna, I see that the thread is marked as resolved, but I don't see the requested changes made. The CloudRun check is still above the Segment call, and the indentation is still not 2 spaces. Can we fix them and then mark them resolved please?

Also, noticed another thing, should we `export` the changed values to `FILESTORE_IP_ADDRESS` and `FILE_SHARE_NAME`, or is that okay?

shot-2023-06-05-12-38-20

pratapaprasanna:
> @pratapaprasanna, I see that the thread is marked as resolved, but I don't see the requested changes made. The CloudRun check is still above the Segment call, and the indentation is still not 2 spaces. Can we fix them and then mark them resolved please?
> 
> Also, noticed another thing, should we `export` the changed values to `FILESTORE_IP_ADDRESS` and `FILE_SHARE_NAME`, or is that okay?
> 
> shot-2023-06-05-12-38-20
Sorry I missed it. Doing them

deploy/docker/entrypoint.sh
```diff
@@ -21,6 +21,20 @@ if [[ -n ${APPSMITH_SEGMENT_CE_KEY-} ]]; then
     || true
 fi
 
+if [[ -n "$FILESTORE_IP_ADDRESS" ]]; then
```
sharat87, line 24: Can we try running this container once? I think this will fail to start the container. Since we now have a `set -e` above this, if this variable isn't defined, the entrypoint script will immediately break.

```suggestion
if [[ -n "${FILESTORE_IP_ADDRESS-}" ]]; then
```

Let's also run this once, on non-CloudRun systems to ensure we aren't missing anything else here please.

sharat87 (changes requested):

sharat87 (approved):

Issue #23840, opened 2023-05-29
[Bug]: Issues related to GS Datasource 


prapullc:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

1) Edit datasource name is not functional
2) Edit and add query is not functional

### Steps To Reproduce

Steps to reproduce:
1) Authorise the GSheet datsource 
2) Click on Edit button 
and observe the button is not functional
3) The Add Query button is inactive and not functional
4) edit data source name is not displayed

### Environment

Release

### Version

cloud

AmanAgarwal041:
RCA : When we authorize gsheet ds, then there is a request `/api/v1/saas/64785f2b14130308b062d88a/token?appsmithToken=${appsmithToken}` which is returning datasource object with userPermissions key as empty array after the authorisation is complete, which is resulting in the disabled edit and add button. This is a server side change.

PR #23920, opened 2023-05-31
feat: updating datasource endpoints contract

sondermanish:
## Description

With introduction of new collection called datasourceStorage the api contracts need to change.
This Pr addresses the same.

Fixes: #https://github.com/appsmithorg/appsmith-ee/issues/1348

The following flows are likely to be affected, where we would require testing:

- comprehensive testing for CRUD is required for all Datasource. i.e. (DB, SAAS, API, OAuth, AuthenticatedAPI, RESTApiDatasource)
- Import export, git flow,  (import from older version of appsmith, export to older version from this version)
- templates and forks without config & with config, mark forkable, e.t.c
- Reconnect datasources after import.
- Query's, JSObject and there data bindings, onpageload actions.
- DB schema
- Database name update.
- Generate page flows.
- Welcome tour
- Preview mode and data bindings.

## Test Plan:
Testsmith [Link](https://github.com/appsmithorg/TestSmith/issues/2413)


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/OAuth2ResponseDTO.java
```diff
@@ -10,7 +10,7 @@
 @ToString
 @NoArgsConstructor
 public class OAuth2ResponseDTO {
-    DatasourceDTO datasource;
+    Datasource datasource;
```
nidhi-nair, line 13: Why is this not storage instead?

app/server/appsmith-plugins/amazons3Plugin/src/main/resources/form.json
```diff
@@ -6,14 +6,14 @@
       "children": [
         {
           "label": "S3 service provider key",
-          "configProperty": "datasourceConfiguration.properties[1].key",
+          "configProperty": "datasourceStorages.active_env.datasourceConfiguration.properties[1].key",
```
nidhi-nair, line 9: Can you please explain this change?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/OAuth2ResponseDTO.java
sondermanish: This requires permission from datasources, which the client side uses to enable certain GAC flows.

app/server/appsmith-plugins/amazons3Plugin/src/main/resources/form.json
sondermanish: @ayushpahwa has more context on this

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCEImpl.java
```diff
@@ -135,30 +131,33 @@ public Mono<Datasource> createWithoutPermissions(Datasource datasource) {
...
-        if (datasource.getId() != null) {
+        if (hasText(datasource.getId())) {
             return Mono.error(new AppsmithException(AppsmithError.INVALID_PARAMETER, FieldName.ID));
         }
-        if (datasource.getPluginId() == null) {
+        if (!hasText(datasource.getPluginId())) {
             return Mono.error(new AppsmithException(AppsmithError.INVALID_PARAMETER, FieldName.PLUGIN_ID));
         }
-        if (!StringUtils.hasLength(datasource.getGitSyncId())) {
+        if (!hasText(datasource.getGitSyncId())) {
```
nidhi-nair, line 143: QQ what's the difference between hasLength and hasText?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCEImpl.java
```diff
@@ -135,30 +131,33 @@ public Mono<Datasource> createWithoutPermissions(Datasource datasource) {
...
             datasource.setGitSyncId(datasource.getWorkspaceId() + "_" + new ObjectId());
         }
-        if (datasource.getDatasourceStorages() == null || datasource.getDatasourceStorages().isEmpty()) {
+
+        if (isNullOrEmpty(datasource.getDatasourceStorages())) {
             return Mono.error(new AppsmithException(AppsmithError.INVALID_PARAMETER, FieldName.DATASOURCE));
         }
 
+        // Set this to null, datasource configuration has moved to datasourceStorage, we will stop storing this.
+        datasource.nullifyStorageReplicaFields();
```
nidhi-nair, line 152: Do we expect to be getting values here or is this just an additional check?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCEImpl.java
```diff
@@ -167,12 +166,6 @@ private Mono<Datasource> createEx(@NotNull Datasource datasource, Optional<AclPe
                         });
             }
             datasourceMono = datasourceMono
-                    .map(datasource1 -> {
-                        // Everything we create needs to use configs from storage
-                        datasource1.setDatasourceConfiguration(null);
-                        datasource1.setHasDatasourceStorage(true);
```
nidhi-nair, line 173: QQ, why are we getting rid of this flag?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCEImpl.java
```diff
@@ -302,14 +246,43 @@ private Mono<Datasource> updateByEnvironmentId(String id,
...
+                    Boolean userInvokedUpdate = Boolean.TRUE.equals(isUserRefreshedUpdate) ? Boolean.TRUE: Boolean.FALSE;
+                    analyticsProperties.put(FieldName.IS_DATASOURCE_UPDATE_USER_INVOKED_KEY, userInvokedUpdate);
                     return analyticsService.sendUpdateEvent(savedDatasource, analyticsProperties);
                 });
+    }
 
+    @Override
+    public Mono<Datasource> updateDatasourceStorage(@NotNull DatasourceStorageDTO datasourceStorageDTO,
+                                                    String activeEnvironmentId,
+                                                    Boolean IsUserRefreshedUpdate) {
```
nidhi-nair, line 258: Nit: variable name does not conform to standards

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCEImpl.java
```diff
@@ -302,14 +246,43 @@ private Mono<Datasource> updateByEnvironmentId(String id,
...
+        }
+
+        // querying for each of the datasource
+        Mono<Datasource> datasourceMono = repository.findById(datasourceId, datasourcePermission.getEditPermission())
+                .switchIfEmpty(Mono.error(new AppsmithException(AppsmithError.NO_RESOURCE_FOUND, FieldName.DATASOURCE, datasourceId)));
+
+        return datasourceMono
+                .flatMap(datasource -> {
+                            datasource.getDatasourceStorages().put(environmentId, datasourceStorageDTO);
+                            return datasourceStorageService.updateByDatasourceAndEnvironmentId(datasource, environmentId, Boolean.TRUE)
```
nidhi-nair, line 278: We were using this method because of the need for transfer service. Now that the transfer is done, let's default to updating by datasourceStorage itself?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCEImpl.java
```diff
@@ -538,8 +528,10 @@ public Flux<Datasource> getAllByWorkspaceIdWithoutStorages(String workspaceId, O
     public Flux<Datasource> getAllByWorkspaceIdWithStorages(String workspaceId, Optional<AclPermission> permission) {
 
         return repository.findAllByWorkspaceId(workspaceId, permission)
+                .publishOn(Schedulers.boundedElastic())
```
nidhi-nair, line 531: Why are we publishing on a different scheduler here?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCEImpl.java
sondermanish: Since hasDatasourceStorage was only used for migration, should we not remove this?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCEImpl.java
sondermanish: We don't expect, but server side check is placed separately

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCEImpl.java
sondermanish: `hasLength` will give you positive for whitespace characters e.g `"     "`, `hasText` would not

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCEImpl.java
nidhi-nair: I think we should maintain this at least till we stabilize ME so that we have a way to easily identify datasources that may or may not be in corrupted state. For example, the failures we are seeing right now in last migration.

sondermanish: reverting this

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCEImpl.java
```diff
@@ -244,6 +244,12 @@ public Mono<Datasource> updateDatasource(String id, Datasource datasource,
                     return dbDatasource;
                 })
                 .flatMap(this::validateAndSaveDatasourceToRepository)
+                .map(savedDatasource -> {
+                    //not required by client side in order to avoid updating it to a null storage,
```
nidhi-nair, line 248: Sorry, I do not understand this change?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCEImpl.java
sondermanish: Since we are able to use different threads to run subsequent db calls , this gives performance boost.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/OAuth2ResponseDTO.java
nidhi-nair: Sorry, could you please provide a few more specifics here? I am trying to understand why sending information as a map of datasourcestorages is warranted here.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceServiceCEImpl.java
sondermanish: Since we are only changing metadata of datasource i.e. name, we will be sending the datasource object without the datasourceStorages object, the client side does something similar to server side's `copyNestedNonNullProperties` . Hence we are setting it to null here.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/OAuth2ResponseDTO.java
sondermanish: The fix [issue](https://github.com/appsmithorg/appsmith/issues/23840) will give more context on the same

app/server/appsmith-plugins/amazons3Plugin/src/main/resources/form.json
ChandanBalajiBP: Assuming a case where a user is editing the production environment and switches to the staging environment, how can we have forms for both environments?

For example, let's say the production environment's ID is "prod_env" and the staging environment's ID is also "prod_env". So we need two forms with the config property `datasourceStorages.prod_env.datasourceConfiguration.properties[1].key` and `datasourceStorages.prod_env.datasourceConfiguration.properties[1].key`.

This can be achieved in multiple ways:

1. Server maintains forms for each environment - worst possible solution
2. Server sends forms for all environments and the client selects the form based on the user selection - Here we are storing the same form twice just because of a change in the config property.
3. We consider `active_env` as the current editing environment and move the required `datasourceStorage` to `active_env` based on selection.

Advantages of the `active_env` method:

- This will also enable us to revert functionality in the future if required.
- The server maintains a single form for any number of environments.
- Much easier implementation.

nidhi-nair: Doesn't this also mean that a form.json configuration is actually operating at a datasourceStorage level now? Is it possible to consider using that as the reference object for these property paths instead?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/OAuth2ResponseDTO.java
nidhi-nair: Okay, but it looks like we're conflating two concepts here. Returning a Datasource object at this point means that we're also operating at multiple environment level, which is not true. Should we not be dealing with a single `DatasourceStorage` here? This goes back to the same question of how client is dealing with datasource storage level permissions.

app/client/src/sagas/DatasourcesSagas.ts
```diff
@@ -651,11 +658,13 @@ function* testDatasourceSaga(actionPayload: ReduxAction<Datasource>) {
   }
 
   try {
+    //Chandan
```
ayushpahwa, line 661: Fix comment please

app/client/src/sagas/DatasourcesSagas.ts
```diff
@@ -1026,15 +1044,23 @@ function* storeAsDatasourceSaga() {
     (d) => !(d.key === "" && d.key === ""),
   );
 
-  set(datasource, "datasourceConfiguration.headers", filteredDatasourceHeaders);
```
ayushpahwa, line 1029: Why are we removing this?

ayushpahwa (changes requested):

app/client/src/api/DatasourcesApi.ts
```diff
@@ -47,10 +43,19 @@ class DatasourcesApi extends API {
     return API.post(DatasourcesApi.url, datasourceConfig);
   }
 
-  static testDatasource(datasourceConfig: Partial<Datasource>): Promise<any> {
-    return API.post(`${DatasourcesApi.url}/test`, datasourceConfig, undefined, {
-      timeout: DEFAULT_TEST_DATA_SOURCE_TIMEOUT_MS,
-    });
+  static testDatasource(
```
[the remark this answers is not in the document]
sondermanish: The environmentId is being sent inside the object itself, we can send the envId here as well, but As far as I remember, this envirionment Id header is for the active environment (what we select from dropdown). Also, we can test staging config while being on production environment, hence we are sending configuration along with required envId as field.
does this answer your question?

app/client/src/components/editorComponents/Debugger/index.tsx
```diff
@@ -27,14 +26,14 @@ function Debugger() {
 
 export function DebuggerTrigger() {
   const dispatch = useDispatch();
-  const showDebugger = useSelector(
-    (state: AppState) => state.ui.debugger.isOpen,
-  );
+  const showDebugger = useSelector(showDebuggerFlag);
   const selectedTab = useSelector(getDebuggerSelectedTab);
   const messageCounters = useSelector(getMessageCount);
-  const totalMessageCount = messageCounters.errors + messageCounters.warnings;
```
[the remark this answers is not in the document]
ChandanBalajiBP: It was leading to unnecessary rendering, hence updated

app/client/src/components/editorComponents/Debugger/index.tsx
```diff
@@ -27,14 +26,14 @@ function Debugger() {
...
   const hideDebuggerIcon = useSelector(hideDebuggerIconSelector);
-  dispatch(setErrorCount(totalMessageCount));
```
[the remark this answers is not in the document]
ChandanBalajiBP: Yes, came across and felt this is a small fix hence covered it here.

app/client/src/components/editorComponents/GlobalSearch/GlobalSearchHooks.test.ts
```diff
@@ -56,22 +56,34 @@ describe("getFilteredAndSortedFileOperations", () => {
 
   it("shows app datasources before other datasources", () => {
     const appDatasource: Datasource = {
-      datasourceConfiguration: {
-        url: "",
+      datasourceStorages: {
+        unused_env: {
```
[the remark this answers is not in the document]
ChandanBalajiBP: This is a CE test case, do we need to update still?

app/client/src/components/editorComponents/form/fields/EmbeddedDatasourcePathField.tsx
```diff
@@ -59,12 +63,14 @@ import { TEMP_DATASOURCE_ID } from "constants/Datasource";
 import LazyCodeEditor from "components/editorComponents/LazyCodeEditor";
 import { getCodeMirrorNamespaceFromEditor } from "utils/getCodeMirrorNamespace";
 import { isDynamicValue } from "utils/DynamicBindingUtils";
+import { getCurrentEnvironment } from "@appsmith/sagas/EnvironmentSagas";
 import { DEFAULT_DATASOURCE_NAME } from "constants/ApiEditorConstants/ApiEditorConstants";
 
 type ReduxStateProps = {
   workspaceId: string;
-  datasource: Datasource | EmbeddedRestDatasource;
+  datasource: DatasourceStorage | EmbeddedRestDatasource;
```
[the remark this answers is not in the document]
ChandanBalajiBP: @sondermanish can you pitch in here

app/client/src/sagas/DatasourcesSagas.ts
ChandanBalajiBP: Have moved the logic down to line 1052

sondermanish:
//build-deploy-preview skip-tests=true

app/client/src/components/editorComponents/form/fields/EmbeddedDatasourcePathField.tsx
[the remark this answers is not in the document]
sondermanish: @ayushpahwa we have kept it to the older format because of two major reasonsl

- if we were to change that, we would have to write migrations for this. on prod this is around 1.3M documents
- It would also change the git files, which would be seen as a diff by end users, need to align with git team first, hence we have delayed on these for later on

app/client/src/api/DatasourcesApi.ts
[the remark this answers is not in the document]
ayushpahwa: Nope, didn't get you. The header is the active env for the whole app while the test can be done for the env being currently edited in the datasource. Which object has the currently edited environment's id? @ChandanBalajiBP @sondermanish

app/client/src/ce/utils/Environments/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export const ENVIRONMENT_QUERY_KEY = "environment";
+
+// function to get the current environment from the URL
+export const getCurrentEnvironment = () => {
+  const queryParams = getQueryParams();
+  if (!!queryParams && queryParams.hasOwnProperty(ENVIRONMENT_QUERY_KEY)) {
+    return queryParams[ENVIRONMENT_QUERY_KEY].toLowerCase();
+  }
+  return "unused_env";
```
ayushpahwa, line 12: Can you please use a variable here? The one defined in APIUtils

app/client/src/components/editorComponents/Debugger/index.tsx
```diff
@@ -33,7 +33,7 @@ export function DebuggerTrigger() {
 
   useEffect(() => {
     dispatch(setErrorCount(messageCounters.errors));
-  }, []);
+  });
```
ayushpahwa, line 36: The dependency should be on `messageCounters.errors` right?

ayushpahwa (dismissed):

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/DatasourceStorageDTO.java
```diff
@@ -20,6 +23,9 @@ public class DatasourceStorageDTO implements Forkable<DatasourceStorageDTO> {
     Set<String> invalids;
     Set<String> messages;
 
+    String pluginId;
+    String workspaceId;
+
     public DatasourceStorageDTO(DatasourceStorage datasourceStorage) {
```
nidhi-nair, line 29: Hey we should be getting rid of all constructor kind of initialization, let's create a separate task for this and take it up in a different PR

ayushpahwa (commented):
removing since I made changes

app/client/src/pages/Editor/DataSourceEditor/index.tsx
```diff
@@ -192,7 +193,7 @@ class DatasourceEditorRouter extends React.Component<Props, State> {
       requiredFields: {},
       configDetails: {},
       filterParams: {
-        id: "",
+        id: DEFAULT_ENV_ID,
```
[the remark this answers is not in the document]
ayushpahwa: These filter params are for filtering the datasource based on the basis of the current env being edited. Hence by default, we are adding default env id. The main usage comes in EE

app/client/src/pages/Editor/SaaSEditor/DatasourceForm.tsx
```diff
@@ -460,8 +464,8 @@ class DatasourceSaaSEditor extends JSONtoForm<Props, State> {
                       pageId={pageId}
                     />
                   ) : null}
-                  {!_.isNil(sections)
-                    ? _.map(sections, this.renderMainSection)
+                  {!isNil(sections)
```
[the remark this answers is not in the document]
ayushpahwa: What's the difference here? We haven't introduced this, just removed the underscore import.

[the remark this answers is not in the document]
ayushpahwa: Also, the section can be undefined/null here also. Empty is not a case we know of.

app/client/src/pages/Editor/gitSync/ReconnectDatasourceModal.tsx
```diff
@@ -474,17 +475,25 @@ function ReconnectDatasourceModal() {
   useEffect(() => {
     if (isModalOpen && !isTesting) {
       const id = selectedDatasourceId;
-      const pending = datasources.filter((ds: Datasource) => !ds.isConfigured);
+      const pending = datasources.filter((ds: Datasource) =>
+        ds.datasourceStorages ? !isEnvironmentConfigured(ds) : true,
```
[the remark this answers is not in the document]
ayushpahwa: dsSotrages can't be empty in any case.

app/client/src/reducers/entityReducers/datasourceReducer.ts
```diff
@@ -300,6 +301,43 @@ const datasourceReducer = createReducer(initialState, {
...
+          return {
+            ...datasource,
+            datasourceStorages: {
+              [`${action.payload.environmentId}`]: action.payload,
+            },
+          };
+
+        return datasource;
+      }),
+      recentDatasources: [
```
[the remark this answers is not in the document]
ayushpahwa: I am guessing this is used in replay flow. I don't think we introduced this.

app/client/src/pages/Editor/SaaSEditor/DatasourceForm.tsx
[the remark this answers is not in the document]
Irongade: IsEmpty checks for null values or undefined as well. It is possible for the section to be an empty array or object. I've ran into this bug before as a Sentry issue.

app/client/src/pages/Editor/DataSourceEditor/index.tsx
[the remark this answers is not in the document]
Irongade: Can we rename the id to envId then to make it clearer.

btsgh:
As part of this second round, based on Ayush's suggestions in this thread, these are the test cases covered by QA:

- Creation and updation of ds one of all types - Done [Also covered Firestore, Snowflake, Redis, Mongo, SMTP, Oracle, Redshift]
- DB- Postgres/mongo - Done [Also covered Firestore, Snowflake, Redis, Mongo, SMTP, Oracle, Redshift]
- Gsheets - Done- identified issues - resolved.
- OAuth2.0 - Done - identified issues - resolved.
- Saas- airtable - Done [Also covered - hubspot, Twillio]
- Rest/graphql - Done [Also covered - cURL API, Auth rest API, Auth GraphQL] - identified issues - resolved.
- Creation, updation and execution of a query made from these datasources - Done
- Creation, updation and execution of an API without a datasource ( please try to navigating away and coming back to the query. Also check with  refresh after updating the query) - Done
- Git operations / git import - export - Done
- Import - Done - identified issues. - resolved.
- Fork from inside app [Make application forkable] - is working. Done

Also to reiterate - Fork issues still exist on release, fix has not been merged. So we have not tested fork on this DP. This PR looks good to go.

ayushpahwa (approved):
Re approving for CI merging

PR #23980, opened 2023-06-02
fix: Fix migrations failing with update error on DocumentDB

### LINKED ISSUE #1477: [Feature] The query section is too small for mongo queries.
## Summary

It is difficult to write long mongo queries in the query pane and bracket matching highlights would really help

## Motivation

The query field is optimized for SQL queries which are more horizontal in nature while mongo queries are more vertical in nature due to their JSON structure. This is true of the API pane as well and we should consider this in our new designs. For JSON, bracket matching is paramount

tkAppsmith:
Fixes https://github.com/appsmithorg/appsmith-ee/issues/1477

#### Type of change
>
>
>
- [x] Manual
- [ ] Jest
- [ ] Cypress
>
>

#### Dev activity

#### QA activity:


tkAppsmith:
Fix Documentation:
As per the aws document db reference, there is no support for aggregate update operation:
https://docs.aws.amazon.com/documentdb/latest/developerguide/mongo-apis.html#mongo-apis-aggregation-pipeline

The target of the failing migrations were to set a field of documents in various collections using the existing value of other field in the same document and for this kind of update an aggregation pipeline with stages in needed. Update queries and update definition wont work in this scenario because they usually assign a static value to the updated fields.

Following strategy is used  for respective migrations:

- migrateOrganizationIdToWorkspaceIdInDomainObjects:
Domain objects such as Datasource, ActionCollection etc are read to a stage by Projection operation and then the field to be updated is added to documents using the add field aggregation operation. Existing value of filed is overwritten in case of add field operation. Then finally the stage is written to the collection using out operation.
Caution: out operation replaces the whole collection by the last stage of aggregtation.

- migrateOrganizationIdToWorkspaceIdInNewActionDatasource:
Same as above

- defaultWorkspaceIdMigration:
The simple approach can’t be applied to this since this is using a match criteria to apply aggregation only to fields having ‘defaultWorkspaceId’ field. Hence we read the whole projection and branch out to 2 projections. One with documents which has the field ‘defaultWorkspaceId’ and the other without the field. Then we update the documents in first collection with the concerned field using the add field operation and then we out the stage to db clearing the collection and hence only the documents with the updated field is present now. Hence now we use the second projection that contains the documents that don’t have the concerned field written in bulk to the collection. This has to be done because the out aggregation operation overwrites the whole collection.

sharat87 (commented):
Thank you @tkAppsmith, for the fantastic detail. Super appreciate the amount you've taken to do this. 🙏

Going over this gives me an idea. Considering this migration, as is, works fine on MongoDB, and also considering there's no current Appsmith instances with DocumentDB, would it be simpler to just skip these migrations for new instances?

For example, in `migrateOrganizationIdToWorkspaceIdInDomainObjects`, we're making changes to datasources, applications, workspaces, among others. But for a fresh new instance, none of these will be there. So the update won't actually do anything in the database. And since there's no instances on DocumentDB today, _all_ DocumentDB instances will be _fresh new_ instances. So, if we add a check like this at the start of the migration function:

```java
if (mongoTemplate.findOne(new Query(), Datasource.class) == null) {
    System.out.println("No data to migrate.");
    return;
}
```

Then, for new instances, MongoDB or DocumentDB, this migration would just pass, which is both faster, and we won't see the aggregation error with DocumentDB. We can do similar checks for the other two migrations on `NewAction` and `PermissionGroup` respectively.

I've wrapped the `updateMulti` calls in these migrations in a `System.out.println`, ran it against a fresh new database, and all I see for _all_ these migrations is this:

```
AcknowledgedUpdateResult{matchedCount=0, modifiedCount=0, upsertedId=null}
```

So, clearly, the migrations aren't doing anything, right?

What do you think?

tkAppsmith:
Hi @sharat87 . You have absolutely right. For a new instance, almost all of the collections coming under these migrations are empty hence these migrations aren't doing anything. For permissionGroups, there are 2 records initially and they get filtered out in the match clause hence no updates are made. The only collection that gets effected by these migration is the user collection. It has one anonymous user which gets affected by these migrations.

sarojsarab (approved):
Backend is now starting with documentDb without any issue.

PR #23990, opened 2023-06-02
fix: Remove api callout

### LINKED ISSUE #23955: [Task]: Remove information Banner from API Editor page

### SubTasks
In the API editor, this banner seems to be confusing. Remove it..

Screenshot 2023-06-01 at 20 45 14

_No response_

Irongade:
Users were getting confused and mislead by the Callout on the API page. This PR gets rid of that.

For more context: https://www.notion.so/appsmith/CRUD-quick-wins-c6a53e027c2f4ebd89b158eccec66a66

Fixes #23955 
>
>

- [x] Manual
- [ ] Jest
- [ ] Cypress
>
>
#### Dev activity

#### QA activity:


sneha122 (approved):

app/client/cypress/e2e/Regression/ServerSide/ApiTests/API_ContextMenu_spec.js
```diff
@@ -1,8 +1,10 @@
 const testdata = require("../../../../fixtures/testdata.json");
 const apiwidget = require("../../../../locators/apiWidgetslocator.json");
 
-import { ObjectsRegistry } from "../../../../support/Objects/Registry";
-let ee = ObjectsRegistry.EntityExplorer;
+import {
```
Aishwarya-U-R, line 4: I meant `import * as _ from ../ObjectsCore` & use as :
` _.entityExplorer.ExpandCollapseEntity("Queries/JS");`
Basically it was created to have one Import & use any objects. Since objects are created once & going to be re-used in all spec files - it costs from creating new objects.

Irongade: This is generally not a good practice @Aishwarya-U-R. You should not be importing objects you don't need in your spec files. In other words, **you should only import objects you need in your spec files**. If this is a common pattern in the cypress test files, they need to be revoked.

Aishwarya-U-R: By Importing from ObjectsCore - we are making all the objects available for the spec - which is created only once.
When you invoke _.agHelper or _.apiPage - anyways only these object & its associated methods/locators are called.
This way, we have a single shared value that can be used across multiple files without duplicating the code or creating separate instances in each file.

Aishwarya-U-R: And modern bundlers anyways optimize the bundle size by eliminating unused code, so impact of importing unused objects is generally minimal which I understand is your concern. While it can increase developer writing experience & increases code readability & consistency.

Irongade: Hey @Aishwarya-U-R, I disagree with you here, To the best of my knowledge, each cypress spec file is ran and tested in isolation, so how can an import in one file be used in other multiple files, also I think if this were true, we would only import all the objects once at the root level and it will automatically be available in all spec files.

Also, just because modern bundlers can optimise bundle size does not mean we should write unoptimisable code. I can think of no reason why we would import 100% of objects just to use 20% of them in the file.

Lastly, that way of writing code is unclear compared to its alternative, `_.apiPage.Assert() `is less clear than `apiPage.Assert()`, especially when you have multiple of them in the file. However this last point is my personal opinion.

I'll be happy to see any documentation or article links where this is a good practice for cypress tests.

Aishwarya-U-R:
@Irongade - Merge latest release & trigger new oktotest. Cancelling the re-run since its past 12 hours from initial attempt.

AmanAgarwal041 (approved):

PR #24010, opened 2023-06-03
fix(eslint): fix direct `remixicon` imports in `packages/design-system/*`

iamakulov:
## Description

This PR:
- moves the icon-related ESLint rules from to `eslintrc.base.json` – to make them apply to all packages
- fixes direct `remixicon` imports surfaced by the change

#### Type of change

- [ ] Manual – smoke-tested that the dashboard and the editor open correctly
- [ ] Jest
- [ ] Cypress

#### Dev activity

#### QA activity:


app/client/packages/design-system/headless/package.json
```diff
@@ -9,6 +9,7 @@
     "prettier": "yarn g:prettier"
   },
   "dependencies": {
+    "@design-system/widgets-old": "workspace:^",
```
iamakulov, line 12: Note that to use `importRemixIcon`, I had to add a dependency onto `@design-system/widgets-old` (in `headless` and `widgets`). This doesn’t feel super clean, so let me know if you’d prefer this to be different! (Eg should `importRemixIcon` live somewhere else?)

KelvinOm: Hey @iamakulov, thank you!

We absolutely do not want to have a dependency on the `design-system/widgets-old`. We discussed this issue in the team and decided to use the SVG code of icons directly in components without importing. We also found that the RemixIcon code is not optimized and before using it in our components, it needs to be run through [SVGOMG](https://jakearchibald.github.io/svgomg/).

Could you help with this?

cc @jsartisan

somangshu: cc @SatishGandham

iamakulov: > We discussed this issue in the team and decided to use the SVG code of icons directly in components without importing. 

Hmm could you share more about this? Curious about the context of the decision

KelvinOm: We don't plan to use a large number of icons in WDS(shouldn't be more than there is now), so we don't need to implement something more complex. We will use icons in our components, but they will be supplied by a third-party library and will be passed them as prop.

iamakulov: Deal! In that case, I’ll keep the ESLint rule on for the `headless` package and will just inline the (optimized) svg 👍

iamakulov: https://github.com/appsmithorg/appsmith/pull/24010/commits/7c37dab2c0ef849a9d8ab89b5eedb6f237a32855 (tested in Storybook)

app/client/packages/design-system/widgets/src/custom.d.ts
```diff
@@ -0,0 +1,11 @@
+declare module "*.svg" {
```
KelvinOm, line 1: Do we really need this?

app/client/packages/design-system/widgets/src/components/Button/Button.test.tsx
```diff
@@ -2,10 +2,14 @@ import React from "react";
 import "@testing-library/jest-dom";
 import { Icon } from "@design-system/headless";
 import { render, screen } from "@testing-library/react";
-import EmotionHappyLineIcon from "remixicon-react/EmotionHappyLineIcon";
+import { importRemixIcon } from "@design-system/widgets-old";
```
KelvinOm, line 5: I'm sure that there should be the same approach as in headless components. Also with this approach, we will be able to avoid changes in `package.json` and `jest.config.js`

app/client/packages/design-system/headless/src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,18 +1,33 @@
...
 import React, { forwardRef, useContext, useRef } from "react";
 import { useVisuallyHidden } from "@react-aria/visually-hidden";
 import type { SpectrumCheckboxProps } from "@react-types/checkbox";
 import type { FocusableRef, StyleProps } from "@react-types/shared";
 import { useCheckbox, useCheckboxGroupItem } from "@react-aria/checkbox";
-
 import { CheckboxGroupContext } from "./context";
 
+// Adapted from remixicon-react/CheckLineIcon (https://github.com/Remix-Design/RemixIcon/blob/f88a51b6402562c6c2465f61a3e845115992e4c6/icons/System/check-line.svg)
+const CheckIcon = ({ size }: { size: number }) => {
```
KelvinOm, line 14: cc @jsartisan

iamakulov: Yup, lmk if you’d prefer a different API/structure of these components! I went with a simpler one.

My only worry is that the easier we make it to inline new icons to `design-system/headless` and `design-system/widgets`, the higher’s the chance someone will do that, which will lead to bundling more and more icons. This was why I didn’t create a reusable wrapper like this one:

```js
const IconWrapper = ({ size = 24, color = "currentColor", children ...props }) =>
  <svg viewbox="0 0 24 24" width={size} height={size} fill={color}>{children}</svg>
```

– having a wrapper like that would “normalize” inlining icons, make it a regular pattern instead of an exceptional one.

KelvinOm: I share your concerns but I suppose we can go without a wrapper and once we realize we need a more complex approach we'll rewrite all this. We also have an icon component and I think `IconWrapper` will lead to some confusion.

app/client/packages/design-system/widgets/src/components/Button/Button.test.tsx
iamakulov: https://github.com/appsmithorg/appsmith/pull/24010/commits/eba0d8af6dc503f08bbe043200a51b5691f113c2

app/client/packages/design-system/widgets/src/custom.d.ts
iamakulov: ++, I added it just for that! Removed it now.

iamakulov:
Failed build output:

```
[appsmith]: Process started
[appsmith]: 
[appsmith]: /home/runner/work/appsmith/appsmith/app/client/src/workers/Evaluation/__tests__/evaluate.test.ts
[appsmith]:   275:1  error  Reason: Dangling *.only tests skip other tests in the file and reduce test coverage  no-restricted-syntax
[appsmith]: 
[appsmith]: ✖ 1 problem (1 error, 0 warnings)
[appsmith]: 
[appsmith]: Process exited (exit code 1), completed in 1m 25s
```

Uhhh should I do anything with this? Code in question:

https://github.com/appsmithorg/appsmith/blob/4a90bb54eb0a5ef73b999284119d0ca657246443/app/client/src/workers/Evaluation/__tests__/evaluate.test.ts#L275-L312

app/client/packages/design-system/widgets/src/components/Spinner/index.styled.tsx
```diff
@@ -1,5 +1,35 @@
...
+  color = "currentColor",
+  size = 24,
+  ...props
+}: {
+  children?: React.ReactNode;
+  color?: string;
+  size?: number;
+  [key: string]: any;
+}) {
+  const className = "remixicon-icon " + (props.className || "");
```
KelvinOm, line 18: Why is it all here? Why not like in other files? This is also a regular icon that we animate through keyframes.

KelvinOm (changes requested):

app/client/packages/design-system/widgets/src/components/Spinner/index.styled.tsx
iamakulov: Oh okay, that’s a good question. I went with that implementation because, unlike all other icons, this one accepts arbitrary props:

https://github.com/appsmithorg/appsmith/blob/4a90bb54eb0a5ef73b999284119d0ca657246443/app/client/packages/design-system/widgets/src/components/Spinner/index.tsx#L5-L11

– so I wanted to ensure the maximum compatibility with existing code that might expect a Remix icon there. (There _is_ the `IconProps` type that only allows `className`, but I’ve seen a pretty common usage of `any` across the codebase, so I wasn’t sure if I can rely on that.)

But I just searched, and it looks like Spinner isn’t actually used anywhere except the `<Button />` component, lol. Let me fix that!

iamakulov: https://github.com/appsmithorg/appsmith/pull/24010/commits/cd82a701d74b7fec132eb9d5a53c5026f9835595

KelvinOm: Got it. We are in the process of developing new components for widgets, so it is only used here. We will start implementing them in the near future. 

[Any](https://github.com/appsmithorg/appsmith/blob/release/app/client/.eslintrc.base.json#L30) is my sadness.  I think we'll do something about it in the future.

app/client/package.json
```diff
@@ -305,6 +305,7 @@
     "jest-canvas-mock": "^2.3.1",
     "jest-environment-jsdom": "^27.4.1",
     "jest-styled-components": "^7.0.8",
+    "json5": "^2.2.3",
```
KelvinOm, line 308: Why do we need this package?

KelvinOm (approved):
LGTM

KelvinOm:
@iamakulov Thank you for your effort.

app/client/package.json
iamakulov: See this:

https://github.com/appsmithorg/appsmith/blob/cd82a701d74b7fec132eb9d5a53c5026f9835595/app/client/.eslintrc.js#L7-L9

In `.eslintrc.base.json`, we have a `@typescript-eslint/no-restricted-imports` rule banning some imports. In `.eslintrc.js`, we want to add _more_ imports into that rule. ESLint offers no way to do that (https://github.com/eslint/eslint/issues/9192) – you can only override a rule completely. So instead, we require the base config, take the banned imports, extend them with new ones, and _then_ override the rule.

We need to use `json5` because the base config uses comments. Perhaps we can just switch the base config to `.js` as well, dunno :)

app/client/packages/design-system/widgets/src/components/Spinner/index.styled.tsx
iamakulov: FWIW one client of mine implemented a very nice process to gradually enforce some TypeScript or ESLint flag (like this `any` one):

1. Keep the ESLint rule disabled in the codebase
2. Set up a separate CI job that runs ESLint with the rule _enabled_. (ESLint will obviously fail, but that’s okay)
3. After ESLint runs, count the number of errors in the codebase (eg by parsing the output), compare it with `release`, and fail a CI check if it gets larger
   1. You don’t need a separate storage to compare the numbers – you can just run the same job on `release`, save the number of errors into CI caches, and then pull these caches in every other branch
4. With time, as developers rewrite existing code, they won’t add any new `any`s – but will remove existing ones, so the number of errors will decrease

I saw that working pretty well over a course of a year!

Although, in that client’s case, a big part why this was successful was also the incentives. The number of `any`s was lightly gamified – you’d see it [+ the difference with `release`] shown in status checks in every PR:

> ✅ **\`any\` usage** — No change (137 ± 0)

or 

> ✅ **\`any\` usage** — Great job! (130, ↓ 7 vs release)

or

> ❌ **\`any\` usage** — Failed: found new usages of \`any\` (138, ↑ 1 vs release)

and there were a few developers who really got into getting that number lower.

KelvinOm: Interesting approach, thanks for sharing this 👍.

I believe that question is actually not only in `any` but also in other rules and checks as well. Previously, I have used the approach when we enabled a rule and then suppressed errors using this [package](https://github.com/amanda-mitchell/suppress-eslint-errors). It worked well, but it added a lot of annoying comments to the code, which worsened readability.

I'm also thinking about moving the reused parts of the code into packages and in package we can use as strict rules as possible.

PR #24032, opened 2023-06-05
chore: Revert "Updated Template"

nidhi-nair:
This reverts commit 42a1ebb4b90544ccb7468741b7cfe02526fb7edf.

Seems like the template file was cleared unknowingly.


nidhi-nair:
Force merging since this is a revert commit only.

PR #24035, opened 2023-06-05
chore: disable cleanup for dps

pratapaprasanna:


.github/workflows/cleanup-dp.yml
```diff
@@ -1,8 +1,9 @@
 name: Cleanup Deploy Previews
 
 on:
-  schedule:
-    - cron: 0 0 * * *
+  repository_dispatch:
```
sharat87, line 4: This is for manually running this, right? Then you need `workflow_dispatch` here. 🙂

```suggestion
  workflow_dispatch:
```

.github/workflows/cleanup-dp.yml
```diff
@@ -1,8 +1,9 @@
...
+#  schedule:
```
sharat87, line 5: Minor

```suggestion
#  schedule:
```

sharat87 (changes requested):

sharat87 (approved):

Issue #24013, opened 2023-06-05
[Bug]: Issues in starting appsmith with proxy


pratapaprasanna:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

```
version: "3"

services:
  appsmith:
    image: appsmith/appsmith-ee:nightly
    container_name: appsmith
    ports:
      - "80:80"
      - "443:443"
      - "9991:9991"
    volumes:
      - ./stacks:/appsmith-stacks
    restart: unless-stopped
    environment:
      HTTPS_PROXY: "1.2.3.4"
      HTTP_PROXY: "1.2.3.4"
```

### Steps To Reproduce

docker-compose up 
and the logs are as follows

```
 % Total  % Received % Xferd Average Speed Time  Time  Time Current
                 Dload Upload Total Spent  Left Speed
100 187 100  21 100 166  23  183 --:--:-- --:--:-- --:--:-- 206
{
 "success": true
nginx: [emerg] socket() [::]:80 failed (97: Address family not supported by protocol)
} % Total  % Received % Xferd Average Speed Time  Time  Time Current
                 Dload Upload Total Spent  Left Speed
100 187 100  21 100 166  22  179 --:--:-- --:--:-- --:--:-- 201
{
 "success": true
nginx: [emerg] socket() [::]:80 failed (97: Address family not supported by protocol)
```

### Environment

Production

### Version

v1.9.20.4

PR #24041, opened 2023-06-05
fix: ipv6 nginx startup error

pratapaprasanna:
Fixes: [24013](https://github.com/appsmithorg/appsmith/issues/24013)
The issue started with the introduction of adding of startup-events. Because of the ipv6 header
```

server {
	listen 80 default_server;
        listen [::]:80 default_server;
}

```
 present in the 
`/etc/nginx/sites-enabled/default` and calling nginx [here](https://github.com/appsmithorg/appsmith/blob/release/deploy/docker/entrypoint.sh#L392). 

With this change we are removing ```listen [::]:80 default_server;``` from the file before starting nginx.

### DIFF
--- a/deploy/docker/entrypoint.sh
+++ b/deploy/docker/entrypoint.sh
@@ -383,6 +383,9 @@ init_postgres || runEmbeddedPostgres=0
 }
 
 init_loading_pages(){
+  # The default NGINX configuration includes an IPv6 listen directive. But not all
+  # servers support it, and we don't need it. So we remove it here before starting
+  # NGINX.
   sed -i '/\[::\]:80 default_server;/d' /etc/nginx/sites-enabled/default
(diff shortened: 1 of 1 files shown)


deploy/docker/entrypoint.sh
```diff
@@ -383,6 +383,7 @@ init_postgres || runEmbeddedPostgres=0
 }
 
 init_loading_pages(){
+  sed -i '/\[::\]:80 default_server;/d' /etc/nginx/sites-enabled/default
```
sharat87, line 386: Let's add a commend describing what this is please?

```suggestion
  # The default NGINX configuration includes an IPv6 listen directive. But not all
  # servers support it, and we don't need it. So we remove it here before starting
  # NGINX.
  sed -i '/\[::\]:80 default_server;/d' /etc/nginx/sites-enabled/default
```

sharat87 (approved):

PR #24175, opened 2023-06-07
Spotless check with Palantir format

nidhi-nair:
Retains most lambda formatting.


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/constants/AnalyticsEvents.java
```diff
@@ -84,8 +85,7 @@ public enum AnalyticsEvents {
 
     DS_TEST_EVENT("Test_Datasource_Clicked"),
     DS_TEST_EVENT_SUCCESS("Test_Datasource_Success"),
-    DS_TEST_EVENT_FAILED("Test_Datasource_Failed")
-    ;
+    DS_TEST_EVENT_FAILED("Test_Datasource_Failed");
```
sharat87, line 88: Can we instead mandate the trailing comma and have the `;` in the next line? The diffs for adding a new enum look much better that way.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/annotations/documenttype/DocumentTypeMapper.java
```diff
@@ -1,5 +1,12 @@
+/* Copyright 2019-2023 Appsmith */
 package com.appsmith.external.annotations.documenttype;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
```
sharat87, line 9: Can we change the import ordering rules to have standard lib as a separate section like we currently do? We should be able to reduce this diff as well.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/constants/GitConstants.java
```diff
@@ -12,12 +13,13 @@ public class GitConstants {
     public static final String DEFAULT_COMMIT_MESSAGE = "System generated commit, ";
     public static final String EMPTY_COMMIT_ERROR_MESSAGE = "On current branch nothing to commit, working tree clean";
     public static final String MERGE_CONFLICT_BRANCH_NAME = "_mergeConflict";
-    public static final String CONFLICTED_SUCCESS_MESSAGE = "branch has been created from conflicted state. Please " +
-            "resolve merge conflicts in remote and pull again";
+    public static final String CONFLICTED_SUCCESS_MESSAGE = "branch has been created from conflicted state. Please "
+            + "resolve merge conflicts in remote and pull again";
```
sharat87, line 17: We've been consistent with having operators at the end of prev. line, instead of start of next line, if I recall right. Can we get that to continue as well?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/constants/AnalyticsEvents.java
nidhi-nair: Let's tackle this separately? We'll either need to find a rule for this in an existing formatter to import it or write a custom regex / ast based transformation logic, both of which I'd rather do as a separate effort.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/constants/GitConstants.java
nidhi-nair: Actually we don't have a rule for this at all in the default IDE specs, we've been un-opinionated about this so far. and I don't see a lot of strings getting affected by this. I prefer the change, looks cleaner. Would you like for us to write a custom rule to not have this instead?

nidhi-nair:
@sharat87 I think we've gotten to the Pareto maxima here, lmk if you think otherwise.

nidhi-nair:
@sharat87 please let me know whether this PR is good to merge as is or whether I should invest in this some more.

sharat87 (approved):
Looks good for a starter I think. I see that we're adding a copyright line like this:

```
/* Copyright 2019-2023 Appsmith */
```

Into _every_ Java file. So, come 2024, will we change this in _every_ Java file? That means the last modified date of any Java file won't ever be more than a year, right? It'll start showing misleading numbers I'm thinking. What say?

nidhi-nair:
Makes sense, there's no need for this line as of today, we'll add it if we ever find need of it.

nidhi-nair:
Closing this in favour of #25173 to discard merge conflicts

PR #24187, opened 2023-06-07
fix: NPE when TLS option not set at all

sharat87:
When TLS option is not set at all on the backend's env variables, this value ends up being `null` in the frontend. This means the initial value for this will be `null` instead of `false`. But since the backend is expecting `true` or `false` here, it breaks when it sees a `null`.

We fix it by ensuring we always sent a concrete value, `false`, if we encounter a `null` or other such `false`-y values.


ankitakinger (approved):
LGTM

PR #24317, opened 2023-06-09
chore: add codeowners file

KelvinOm:
## Description
Add codeowners file

Some pictures with examples

#### Type of change

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,23 @@
+# This is a comment.
+# Each line is a file pattern followed by one or more owners.
+# https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
+
+# WDS team
+app/client/packages/design-system/* @appsmithorg/wds
+app/client/packages/storybook/* @appsmithorg/wds
+
+app/client/.husky @KelvinOm
+app/client/.yarn @KelvinOm
+app/client/.editorconfig @KelvinOm
+app/client/.eslintrc.base.json @KelvinOm
+app/client/.eslintrc.js @KelvinOm
+app/client/.gitignore @KelvinOm
+app/client/.lintstagedrc @KelvinOm
+app/client/.nvmrc @KelvinOm
+app/client/.prettierignore @KelvinOm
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -0,0 +1,23 @@
+# This is a comment.
+# Each line is a file pattern followed by one or more owners.
+# https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
+
+# WDS team
+app/client/packages/design-system/* @appsmithorg/wds
+app/client/packages/storybook/* @appsmithorg/wds
```
KelvinOm, line 7: @somangshu I created a [WDS team](https://github.com/orgs/appsmithorg/teams/wds) with write permissions.

CODEOWNERS
```diff
@@ -0,0 +1,23 @@
...
+
+app/client/.husky @KelvinOm
```
KelvinOm, line 9: @somangshu I believe there should be a team here, not just me, and this falls on my plans to improve processes. I will prepare documentation on this a bit later.

KelvinOm:
@somangshu It will be necessary to see how it works in real life, sense only 1 review is mandatory for us and the approval are reset after the update. Maybe we need to think about how to improve everything here and possibly made mandatory approval from code owners.

CODEOWNERS
somangshu: yes we can create a team for this. 
Feels like most of the frontend team can be assigned

cc @riteshkew what do you say?

somangshu (approved):

CODEOWNERS
riteshkew: Creating teams would be better, so that multiple owners can take care of the requests rather than a single person. We can start with individuals and gradually update to teams

riteshkew (approved):
Looks good. I will take it up with the other teams and get the code paths added.

CODEOWNERS
riteshkew: Are we making the code owners review mandatory? @KelvinOm

KelvinOm: I believe that code owners review should be mandatory, yes

PR #24416, opened 2023-06-13
chore: Breakdown semantic options for button

### LINKED ISSUE #24311: Breakdown semantic options for button
Replace `"primary" | "secondary" | "tertiary"` variants with `filled | outlined | ghost` respectively
Add type prop with options `accent | neutral | positive | negative | warning`

jsartisan:
Fixes #24311

### DIFF
--- a/app/client/packages/design-system/widgets/src/components/Button/Button.stories.mdx
+++ b/app/client/packages/design-system/widgets/src/components/Button/Button.stories.mdx
@@ -30,27 +30,79 @@ There are 3 variants of the button component.
 
 <Canvas>
   <Story
-    name="Primary"
+    name="Fileld"
     args={{
-      children: "Primary",
+      children: "Filled",
     }}
   >
     {Template.bind({})}
   </Story>
   <Story
-    name="Secondary"
+    name="Outlined"
     args={{
-      variant: "secondary",
-      children: "Secondary",
+      variant: "outlined",
+      children: "Outlined",
     }}
   >
     {Template.bind({})}
   </Story>
   <Story
-    name="Tertiary"
+    name="Ghost"
     args={{
-      variant: "tertiary",
-      children: "Tertiary",
+      variant: "ghost",
+      children: "Ghost",
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+</Canvas>
+
+# Semantic Variants
+
+There are 4 variants of the button component.
+
+<Canvas>
+  <Story
+    name="Accent"
+    args={{
+      color: "accent",
+      children: "Accent",
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+  <Story
+    name="Positive"
+    args={{
+      color: "positive",
+      children: "Positive",
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+  <Story
+    name="Negative"
+    args={{
+      color: "negative",
+      children: "Negative",
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+  <Story
+    name="Warning"
+    args={{
+      color: "warning",
+      children: "Warning",
+    }}
+  >
+    {Template.bind({})}
+  </Story>
+  <Story
+    name="Neutral"
+    args={{
+      color: "neutral",
+      children: "Neutral",
     }}
   >
     {Template.bind({})}
--- a/app/client/packages/design-system/widgets/src/components/Button/Button.tsx
+++ b/app/client/packages/design-system/widgets/src/components/Button/Button.tsx
@@ -1,27 +1,39 @@
-import React, { forwardRef } from "react";
-import { useVisuallyHidden } from "@react-aria/visually-hidden";
-
-import { Text } from "../Text";
 import type {
   ButtonProps as HeadlessButtonProps,
   ButtonRef as HeadlessButtonRef,
 } from "@design-system/headless";
-import { Spinner } from "../Spinner";
-import { StyledButton } from "./index.styled";
+import React, { forwardRef } from "react";
+import { useVisuallyHidden } from "@react-aria/visually-hidden";
 import { Icon as HeadlessIcon } from "@design-system/headless";
 
-export type ButtonVariants = "primary" | "secondary" | "tertiary";
+import { Text } from "../Text";
+import { Spinner } from "../Spinner";
+import { DragContainer, StyledButton } from "./index.styled";
+import type { ButtonColor, ButtonVariant } from "./index.styled";
 
 export interface ButtonProps extends Omit<HeadlessButtonProps, "className"> {
-  /**
-   *  @default primary
+  /** variant of the button
+   *
+   * @default "filled"
    */
-  variant?: ButtonVariants;
+  variant?: ButtonVariant;
+
+  /** color tone of the button */
+  color?: ButtonColor;
+
+  /** when true, makes the button occupy all the space available */
   isFitContainer?: boolean;
-  isFocused?: boolean;
+
+  /** indicates the loading state of the button */
   isLoading?: boolean;
+
+  /** icon to be used in the button of the button */
   icon?: React.ReactNode;
+
+  /** Indicates the position of icon of the button */
   iconPosition?: "start" | "end";
+
+  /** makes the button visually and functionaly disabled but focusable */
   visuallyDisabled?: boolean;
 }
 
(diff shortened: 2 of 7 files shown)


app/client/packages/design-system/theming/src/color/DarkModeTheme.ts
```diff
@@ -45,23 +45,49 @@ export class DarkModeTheme implements ColorModeTheme {
...
       bgAssistive: this.bgAssistive.toString(),
+      bgPositive: this.bgPositive.toString(),
+      bgPositiveHover: this.bgPositiveHover.toString(),
+      bgPositiveActive: this.bgPositiveActive.toString(),
+      bgPositiveSubtleHover: this.bgPositiveSubtleHover.toString(),
+      bgPositiveSubtleActive: this.bgPositiveSubtleActive.toString(),
       bgNegative: this.bgNegative.toString(),
+      bgNegativeHover: this.bgNegativeHover.toString(),
+      bgNegativeActive: this.bgNegativeActive.toString(),
+      bgNegativeSubtleHover: this.bgNegativeSubtleHover.toString(),
```
ichik, line 56: This (and other `subtle`) variants. We don't have a use for them yet, let's not add tokens that we don't use.

app/client/packages/design-system/theming/src/color/DarkModeTheme.ts
```diff
@@ -45,23 +45,49 @@ export class DarkModeTheme implements ColorModeTheme {
...
+      bgNeutralSubtleHover: this.bgNeutralSubtleHover.toString(),
+      bgNeutralSubtleActive: this.bgNeutralSubtleActive.toString(),
       // fg
       fg: this.fg.toString(),
       fgAccent: this.fgAccent.toString(),
-      fgNegative: this.fgNegative.toString(),
+      fgPositive: this.fgPositive.toString(),
       fgNeutral: this.fgNeutral.toString(),
+      fgWarning: this.fgWarning.toString(),
+      fgNegative: this.fgNegative.toString(),
```
ichik, line 75: Alphabetization broke.

app/client/packages/design-system/theming/src/color/DarkModeTheme.ts
```diff
@@ -45,23 +45,49 @@ export class DarkModeTheme implements ColorModeTheme {
       bgAccentSubtleHover: this.bgAccentSubtleHover.toString(),
       bgAccentSubtleActive: this.bgAccentSubtleActive.toString(),
       bgAssistive: this.bgAssistive.toString(),
+      bgPositive: this.bgPositive.toString(),
```
KelvinOm, line 48: @ichik Please check this

app/client/packages/design-system/theming/src/color/DarkModeTheme.ts
ichik, line 56: `*Subtle`s are gonna be a pain to make meaningfully different from red/green/yellow seed.

ichik (approved):

app/client/packages/design-system/theming/src/color/DarkModeTheme.ts
ichik: This line looks fine? What should I be looking at?

app/client/packages/design-system/widgets/src/components/Button/Button.tsx
```diff
@@ -49,6 +56,7 @@ export const Button = forwardRef(
             <HeadlessIcon>
               <Spinner />
             </HeadlessIcon>
+            {/* TODO(pawan): How to make sure "Loading..." text is internationalized? */}
```
KelvinOm, line 59: Could you create ticket for this it our backlog?

app/client/packages/design-system/theming/src/color/DarkModeTheme.ts
KelvinOm: First of all, the names of the tokens.

app/client/packages/design-system/widgets/src/components/Button/index.styled.tsx
```diff
@@ -1,91 +1,152 @@
-import styled from "styled-components";
+import styled, { css } from "styled-components";
 import { Button as HeadlessButton } from "@design-system/headless";
+
 import type { ButtonProps } from "./Button";
 
+export const BUTTON_ICON_POSITIONS = ["start", "end"] as const;
```
KelvinOm, line 6: I think it's better to move types to Button.txt and not leave it in the styling file as it is done in other components. Perhaps we should consider putting the types in a separate file.

app/client/packages/design-system/widgets/src/components/Button/index.styled.tsx
```diff
@@ -1,91 +1,152 @@
...
+
+export type ButtonIconPosition = (typeof BUTTON_ICON_POSITIONS)[number];
+
+export const BUTTON_VARIANT = {
```
KelvinOm, line 10: Let's try to use `Union types` insted of `as const`. With such a record, [the types will be in a bundle](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBAYwgOwM7wEIFUAqODyAcgPoBqAggEoCS5hOcAvHAN4CwAUHHAGYCWAGwHAAJgC44AIn5DRkgDSduEAK4wBfZKImTV6zXMVc4AcwAWEdDvOWYCzgF84AQ1SIU6ANydOoSLDgYAE8wYDgAIzUYFFJnKD5nZHgmKWEeOzgAHykEYCTgKEksqXjzO08gA), and we do not need it. I would suggest using this construction only when we need to iterate by values, and this is definitely not our case. 

Also, if we use union types directly in the prop interface, they will be displayed normally in the storybook(compare how the type and variant are displayed in storybook). Automatic documentation is wonderful.

app/client/packages/design-system/widgets/src/components/Button/index.styled.tsx
```diff
@@ -1,91 +1,152 @@
...
+  }
+`;
+
+/**
+ * We have this Bug in Firefox where we are unable to drag
+ * buttons - https://bugzilla.mozilla.org/show_bug.cgi?id=568313
+ *
+ * We found a solution here - https://stackoverflow.com/a/43888410
+ */
+export const DragContainer = styled.div`
```
KelvinOm, line 143: Let's try to use this like a wrapper on widget level.

app/client/packages/design-system/theming/src/color/types.ts
```diff
@@ -4,27 +4,55 @@ export type ColorMode = "light" | "dark";
...
   getColors: () => {
+    // bg
     bg: string;
     bgAccent: string;
     bgAccentHover: string;
     bgAccentActive: string;
     bgAccentSubtleHover: string;
     bgAccentSubtleActive: string;
-    bgAssistive: string;
+    bgPositive: string;
```
KelvinOm, line 14: Please create tikets for this. Combine tasks into groups by neutral, positive, negative, and warning.

app/client/packages/design-system/widgets/src/components/Button/index.styled.tsx
```diff
@@ -1,91 +1,152 @@
...
   }
 
   &[data-loading] {
     cursor: default;
-  }
-
-  & [data-icon] {
-    height: calc(var(--root-unit) * 5);
-    width: calc(var(--root-unit) * 5);
+    opacity: 1;
```
KelvinOm, line 120: Why do we need this?

jsartisan: For aria-disabled = true , we are setting the opacity to 0.4 and aria-disabled is true for loading too but we don't want to lower the opacity. This is for that case.

app/client/packages/design-system/widgets/src/components/Button/index.styled.tsx
jsartisan: Hey we can't use it as a wrapper. The click events and hover events are not working properly when i do so. This is the reason button v1 was using click event on the drag container manually instead of button.

app/client/packages/design-system/widgets/src/components/Button/index.styled.tsx
KelvinOm: Please leave a comment about this in the code.

app/client/packages/design-system/widgets/src/components/Button/index.styled.tsx
KelvinOm: Ok. Let's leave it this way.

app/client/packages/design-system/widgets/src/components/Button/Button.stories.mdx
```diff
@@ -2,6 +2,7 @@ import { Canvas, Meta, Story, ArgsTable } from "@storybook/addon-docs";
 
 import { Button } from "./";
 import { Icon } from "@design-system/headless";
+import { BUTTON_COLOR, BUTTON_VARIANT } from "./index.styled";
```
KelvinOm, line 5: It is no longer necessary

KelvinOm (dismissed):

KelvinOm (approved):

Issue #14535, opened 2022-06-14
[Feature]: Increasing timeout on internal nginx


DanPhonovation:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary
I have appsmith deployed via helm.

I see that there is an internal nginx proxying api calls to localhost:8080. I need to increase the timeout on this call past the nginx default of 60 seconds, but can't find an existing entrypoint to modify this.

Any way this can be part of a ConfigMap or recommendations on injecting something like: 

```
proxy-connect-timeout: "360"
proxy-send-timeout: "360"
proxy-read-timeout: "360"
```

into sites-available/default in the /api block (which proxies to localhost:8080)?

### Why should this be worked on?

Custom edge cases. For long running API tasks >60 secs this will generate a 504 Gateway timeout otherwise.

<img src="https://front.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_5wrpo)

DanPhonovation:
For more clarity, I have a basic button-press UI that is calling a custom API performing a long running task. We have verified that indeed, it is appsmith's nginx issuing Gateway Timeout, and not our custom API. We did this by dropping into the appsmith pod and viewing nginx logs.

ajwillia:
I'm also experiencing the same issue; have a REST API that is long running (90-200+ seconds).  I ended up implementing a temporary work-around by adding the following lines to nginx.conf and restarting nginx
        proxy_send_timeout 300s;
        proxy_read_timeout 300s;
        fastcgi_send_timeout 300s;
        fastcgi_read_timeout 300s;

DanPhonovation:
This sed will do the job if anyone wants to include it in their own Dockerfile:
```
sed '/location \/api {/a \    proxy_read_timeout 300;\n    proxy_connect_timeout 300;\n    proxy_send_timeout 300;\n    fastcgi_send_timeout 300;\n    fastcgi_read_timeout 300;' appsmith.nginx.conf
```

heedfull:
Here is a screen shot of the issue, when running appsmith using the provided docker image
![image](https://user-images.githubusercontent.com/2331676/174295727-1a8dfb68-50ba-4fea-ad13-b40c6e313a0f.png)

jackdmoloney:
[...]

PR #24471, opened 2023-06-14
feat: Configurable api server timeout value on Nginx

sum35h:
## Description
- The default request timeout on nginx is set to 60s. This feat allows users to configure custom nginx timeout value specified via the Environment variable `APPSMITH_SERVER_TIMEOUT`
- Solves https://github.com/appsmithorg/appsmith/issues/14535
- [ ] Manual
- [ ] Jest
- [ ] Cypress

### DIFF
--- a/deploy/docker/templates/nginx/nginx-app-http.conf.template.sh
+++ b/deploy/docker/templates/nginx/nginx-app-http.conf.template.sh
@@ -93,6 +93,10 @@ server {
   }
 
   location /api {
+    proxy_read_timeout ${APPSMITH_SERVER_TIMEOUT:-60};
+    proxy_connect_timeout ${APPSMITH_SERVER_TIMEOUT:-60};
+    proxy_send_timeout ${APPSMITH_SERVER_TIMEOUT:-60};
+  
     proxy_pass http://localhost:8080;
   }
 
--- a/deploy/docker/templates/nginx/nginx-app-https.conf.template.sh
+++ b/deploy/docker/templates/nginx/nginx-app-https.conf.template.sh
@@ -110,6 +110,10 @@ server {
   }
 
   location /api {
+    proxy_read_timeout ${APPSMITH_SERVER_TIMEOUT:-60};
+    proxy_connect_timeout ${APPSMITH_SERVER_TIMEOUT:-60};
+    proxy_send_timeout ${APPSMITH_SERVER_TIMEOUT:-60};
+    
     proxy_pass http://localhost:8080;
   }
 


sharat87 (commented):
I wanted to confirm if a long-running API call can be made in the DP, but the DP seems to be down. Can you check please? Can you also review the below and share you thoughts?

From http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_read_timeout

> The timeout is set only between two successive read operations, not for the transmission of the whole response.

So, the effective timeout from setting this can be much longer. Is this what we're expecting here?

From http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_connect_timeout,

> [...] timeout for establishing a connection with a proxied server.

This is the time for just establishing the connection, and shouldn't affect timeouts from long-running queries, right?

From http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_send_timeout,

> [...] timeout for transmitting a request to the proxied server.

This is timeout between adjacent bytes of data sent to the backend. This is also not what we want here, right?

Also, the default value for all three from docs is shown as `60s`, not `60`. Does it work without the `s` as well?

sum35h:
@sharat87  The configs work without specifying seconds.

- `proxy_connect_timeout` -  is not required to increase the query time out.
-  `proxy_read_timeout` - does actually increase the query timeout despite the description, so this will be required.
-  `proxy_write_timeout` - did not try this but I think it will increase timeout for post/put, so it would alos be required?

deploy/docker/templates/nginx/nginx-app-http.conf.template.sh
```diff
@@ -93,6 +93,10 @@ server {
   }
 
   location /api {
+    proxy_read_timeout ${APPSMITH_SERVER_TIMEOUT:-60};
+    proxy_connect_timeout ${APPSMITH_SERVER_TIMEOUT:-60};
```
sum35h, line 97:
```suggestion
```

deploy/docker/templates/nginx/nginx-app-https.conf.template.sh
```diff
@@ -110,6 +110,10 @@ server {
   }
 
   location /api {
+    proxy_read_timeout ${APPSMITH_SERVER_TIMEOUT:-60};
+    proxy_connect_timeout ${APPSMITH_SERVER_TIMEOUT:-60};
```
sum35h, line 114:
```suggestion
```

deploy/docker/templates/nginx/nginx-app-http.conf.template.sh
```diff
@@ -93,6 +93,9 @@ server {
   }
 
   location /api {
+    proxy_read_timeout ${APPSMITH_SERVER_TIMEOUT:-60};
+    proxy_send_timeout ${APPSMITH_SERVER_TIMEOUT:-60};
+
```
sharat87, line 98:
```suggestion

```

deploy/docker/templates/nginx/nginx-app-https.conf.template.sh
```diff
@@ -110,6 +110,9 @@ server {
   }
 
   location /api {
+    proxy_read_timeout ${APPSMITH_SERVER_TIMEOUT:-60};
+    proxy_send_timeout ${APPSMITH_SERVER_TIMEOUT:-60};
+
```
sharat87, line 115:
```suggestion

```

sharat87 (changes requested):
Tested and works. Made suggestions for some whitespace cleanup, but otherwise ready to merge.

sharat87 (approved):

shalomy-cyera:
@sum35h Thanks for this fix!
Will it be available also for helm install?

sum35h:
@shalomy-cyera For the helm deployment, we'll need to apply the same settings to the ingress controller. We will update you once we publish the docs to apply the setting to helm installations.

shalomy-cyera:
> @shalomy-cyera For the helm deployment, we'll need to apply the same settings to the ingress controller. We will update you once we publish the docs to apply the setting to helm installations.

Thanks a lot @sum35h !

tlawrence-TAG:
You need to add configuration settings to the UI

Issue #24358, opened 2023-06-12
Bug: Embedded postgres fails to start due to file permissions on K8s statefulsets


sum35h:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Postgres daemon fails to start due to insufficient file permission. 
```
2023-06-09 09:12:29.452 UTC [1558] LOG:  database system is shut down
supervisor: couldn't chdir to /appsmith-stacks/data/postgres/main: EACCES
```

### Steps To Reproduce

1. Helm install appsmith ce
2. Exec into the pod
3.  Run supervisorctl status to observe postgres process is in FATAL state
4.  Run supervisorctl tail postgres to get error logs
5. Run psql to access postgres shell

### Environment

Release

### Version

1.9.21

sum35h:
```
root@d6f22da388da:/appsmith-stacks/data/postgres# ls -al
total 12
drwxr-xr-x  3 root    root     4096 May 10 08:38 .
drwxr-xr-x  8 root    root     4096 May 10 08:38 ..
drwx------ 19 mongodb postgres 4096 Jun  9 07:10 main
```

PR #24601, opened 2023-06-19
fix: Embedded postgres file owner fix

sum35h:
Fixes https://github.com/appsmithorg/appsmith/issues/24358 where postgres system files owner is updated with mongodb user on updating Appsmith.

Tested manually by updating v1.9.22 to nightly.

### DIFF
--- a/deploy/docker/entrypoint.sh
+++ b/deploy/docker/entrypoint.sh
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
 
 set -e
+
+stacks_path=/appsmith-stacks
+
 # ip is a reserved keyword for tracking events in Mixpanel. Instead of showing the ip as is Mixpanel provides derived properties.
 # As we want derived props alongwith the ip address we are sharing the ip address in separate keys
 # https://help.mixpanel.com/hc/en-us/articles/360001355266-Event-Properties
@@ -35,7 +38,6 @@ if [[ -n "${FILESTORE_IP_ADDRESS-}" ]]; then
   export HOSTNAME="cloudrun"
 fi
 
-stacks_path=/appsmith-stacks
 
 function get_maximum_heap() {
     resource=$(ulimit -u)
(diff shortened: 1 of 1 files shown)


deploy/docker/entrypoint.sh
```diff
@@ -349,15 +347,16 @@ init_postgres() {
...
-      mkdir -p $POSTGRES_DB_PATH
+      mkdir -p "$POSTGRES_DB_PATH"
 
       # Postgres does not allow it's server to be run with super user access, we use user postgres and the file system owner also needs to be the same user postgres
-      chown postgres:postgres $POSTGRES_DB_PATH
+      chown postgres:postgres "$POSTGRES_DB_PATH"
 
       # Initialize the postgres db file system
-      su -m postgres -c "/usr/lib/postgresql/13/bin/initdb -D $POSTGRES_DB_PATH"
+      su postgres -c "/usr/lib/postgresql/13/bin/initdb -D $POSTGRES_DB_PATH"
```
sharat87, line 359: Can you detail why we're making this change please?

deploy/docker/entrypoint.sh
```diff
@@ -31,8 +31,6 @@ if [[ -n "${FILESTORE_IP_ADDRESS-}" ]]; then
   FILE_SHARE_NAME="$(echo "$FILE_SHARE_NAME" | xargs)"
 
   echo "Running appsmith for cloudRun"
-  echo "creating mount point"
-  mkdir -p "$stacks_path"
```
sharat87, line 35: Can you detail why this isn't needed anymore for CloudRun please?

sharat87 (commented):
There's a couple of changes that aren't just setting the owner of Postgres files, can you confirm they're related to the same problem, if yes, share details, and if not, can we revert and keep the hotfix small and focused please?

deploy/docker/entrypoint.sh
sum35h: Missed that, uas on stale release.

PR #24731, opened 2023-06-22
chore: print workflow information

rajatagrawal:
This PR prints github event when ok-to-test command is run. This is needed for debugging optimization changes to skip server tests if there are no server side changes.


sarojsarab (approved):
LGTM

PR #24760, opened 2023-06-22
chore: Pass errors to widgets

### LINKED ISSUE #24663: [Feature]: Pass evaluation errors to widget when there are evaluation errors

### Summary

Widget should have access to evaluation errors because that allows it to make a decision on how to handle errors in property pane and give users a more visual cue to take a corrective action for fixing errors in the property pane.

The current pop up isn't too conspicuous to notice and also doesn't appear when the user deselects the property pane and reselects it.

1. This aligns with Widget Development API. Accessing the evaluation errors from __evaluations__ isn't recommended because it is an internal property. An internal property is not advisable to be used given our roadmap for community widgets.
2. Showing default data instead may not be the right way to handle errors in every use case.

Loom Video describing the requirement : https://www.loom.com/share/04c228b06bb34b97894e345b46cf0abe?sid=f855ac90-74fa-4ee3-b882-ec70b1b2d9bf

rajatagrawal:
Fixes #24663

**Summary**

Widget should have access to evaluation errors because that allows it to make a decision on how to handle errors in property pane and give users a more visual cue to take a corrective action for fixing errors in the property pane.

The current pop up isn't too conspicuous to notice and also doesn't appear when the user deselects the property pane and reselects it.

**Why should this be worked on?**

This aligns with Widget Development API. Accessing the evaluation errors from evaluations isn't recommended because it is an internal property. An internal property is not advisable to be used given our roadmap for community widgets.
Showing default data instead may not be the right way to handle errors in every use case.
Loom Video describing the requirement : https://www.loom.com/share/04c228b06bb34b97894e345b46cf0abe?sid=f855ac90-74fa-4ee3-b882-ec70b1b2d9bf


app/client/src/selectors/dataTreeSelectors.ts
```diff
@@ -89,7 +91,23 @@ export const getConfigTree = (): any => {
...
+    for (const propertyPath in evaluationErrors) {
+      if (evaluationErrors[propertyPath].length > 0) {
+        for (const evalError of evaluationErrors[propertyPath]) {
+          allErrors.push({ ...evalError });
+        }
+      }
+    }
+
+    widgetEntity.errors = allErrors;
+    return widgetEntity;
```
vsvamsi1, line 109: This selector will get triggered when dataTree changes, Can you make the getDataTree dependency to select the widgetEntity state directly. Something like this 

```
export const getWidgetEvalValues = createSelector(
  [(state, widgetName )=>getDataTree(state)?. [widgetName], (_state: AppState, widgetName: string) => widgetName],
  (widgetEntity: WidgetEntity, widgetName: string) => {
   if(!widgetName || !widgetEntity) return null;
    const allErrors: WidgetError[] = [];

    const evaluationErrors = get(widgetEntity, EVAL_ERROR_PATH, {});

    for (const propertyPath in evaluationErrors) {
      if (evaluationErrors[propertyPath].length > 0) {
        for (const evalError of evaluationErrors[propertyPath]) {
          allErrors.push({ ...evalError });
        }
      }
    }


    return {...widgetEntity, errors: allErrors};


```

vsvamsi1: Also if you have several widget instances reselect's caching may not work because they will keep invalidating each other with different widgetNames. Perhaps you may have lazily create the selector to ensure each widget has a unique instance.

```
export const makeGetWidgetEvalValues =()=> createSelector(
  [(state, widgetName )=>getDataTree(state)?.[widgetName], (_state: AppState, widgetName: string) => widgetName],
  (widgetEntity: WidgetEntity, widgetName: string) => {
   if(!widgetName || !widgetEntity) return null;
    const allErrors: WidgetError[] = [];

    const evaluationErrors = get(widgetEntity, EVAL_ERROR_PATH, {});

    for (const propertyPath in evaluationErrors) {
      if (evaluationErrors[propertyPath].length > 0) {
        for (const evalError of evaluationErrors[propertyPath]) {
          allErrors.push({ ...evalError });
        }
      }
    }


    return {...widgetEntity, errors: allErrors};

  },
);

--- jsx code--

const getWidgetEvalValues = useMemo(()=> selectors.makeGetWidgetEvalValues(),[])
 const evaluatedWidget = useSelector((state: AppState) =>
      getWidgetEvalValues(state, widgetName),
    );
```

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -863,6 +863,14 @@ export const WIDGET_DISPLAY_PROPS = {
   isDisabled: true,
   backgroundColor: true,
 };
+export interface WidgetError extends Error {
+  type: "data" | "widgetconfig" | "platform";
```
sbalaji1192, line 867: let's define it as "property" | "configuration" | "other"

app/client/src/utils/widgetRenderUtils.tsx
```diff
@@ -40,13 +45,58 @@ export const createCanvasWidget = (
...
+      name: "Unknown Error",
+      message: "Unavailable",
+      stack: "Unavailable",
+      type: "property",
+    };
+
+    const errorObject: Record<string, unknown> | undefined =
+      evalError.errorMessage as Record<string, unknown>;
+    if (errorObject) {
+      if (errorObject.name && (errorObject.name as string).length > 0) {
```
riodeuno, line 83: If the `errorObject` exists, why would the `name` and `message` property not be filled? Could you share an example? If this is the case, we may want to check with the @appsmithorg/fe-coders-pod to see if we can put this in their backlog.

For a stable system, we should ideally be able to use the `EvaluationError` type, to specifically mention `name` and `message` properties.

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -863,6 +863,12 @@ export const WIDGET_DISPLAY_PROPS = {
   isDisabled: true,
   backgroundColor: true,
 };
+export interface WidgetError extends Error {
```
riodeuno, line 866: There are changes coming to the `BaseWidget` and I was hoping that we can move these type definitions to a separate file in the `widgets` directory. Thoughts?

riodeuno (commented):
The code looks fine to me. A few areas have a concerned comment. Please look into them.

app/client/src/utils/widgetRenderUtils.tsx
rajatagrawal: @arunvjn Can you please confirm if evaluation errors will always have `errorMessage` and `raw` properties in them ? Further if `errorMessage` will have `name` and `message` fields inside it ?

app/client/src/widgets/BaseWidget.tsx
rajatagrawal: We can have a folder called `BaseWidgets` and have two files in it, `index.ts` and `constants.ts`. `index.ts` can have the definition of `BaseWidget` and `constants.ts` can have the types and interfaces declarations.

rajatagrawal: @riodeuno 
I can make this change if this suggestion seems reasonable.

riodeuno: We are looking at changing what goes into the BaseWidget. The general idea is that we'll have "pluggable" and modular layout mechanisms (Fixed mode, Auto Layout mode and the new Auto Layout mode). This will change how we organise the components in the editor. The mobile pod will share the engg tech spec for your review. In the meantime, I will recommend targeting the fact that we've the types mixed up with the BaseWidget definition, making it difficult to restructure.  As far as the way we do it, please do it as you see fit.

rajatagrawal: @riodeuno 

To manage the scope of this github issue, I have created another issue to refactor the organization of BaseWidget definitions [here](https://github.com/appsmithorg/appsmith/issues/25147).

We can merge this pull request and address organization changes in the new issue above.

If everything else looks good in this PR, please let me know!

riodeuno: @rajatagrawal Thanks for heeding my suggestions. I've taken a look at the code. Looks good to me.

rajatagrawal:
> Please update the PR description. This description is going to be useful for the following scenarios:
> 
> 1. During git blame, we asynchronously know exactly why these changes were made
> 2. During release promotion, the PR description gives the team an idea on how to describe these changes to our users.
> 
> Finally, I would also ask that we create an issue to update the widget development API, letting developers know about the `errors` API now available in widgets.

Issue for adding widgetErrors to Widget Development API is [here](https://github.com/appsmithorg/appsmith/issues/25164)

Also updated the PR description. QQ : Can the details of the PR be picked from the issue instead ? The detailed summary is described in the issue first and these details are simply duplicated in the PR description.

PR #24766, opened 2023-06-22
feat: Add a method to verify signature for cloud services response

abhvsn:
## Description
As of now the CS API does not have signature verification which can lead to data tampering for CS API response. This PR adds the method to add signature verification for CS API responses. 

Corresponding PRs:
CS: https://github.com/appsmithorg/cloud-services/pull/1023

Fixes https://github.com/appsmithorg/cloud-services/issues/1037

#### Type of change

- [ ] Manual

#### Dev activity

#### QA activity:


app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/SignatureVerifier.java
```diff
@@ -0,0 +1,78 @@
...
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.Base64;
+import java.util.Objects;
+
+import static com.appsmith.server.constants.ApiConstants.APPSMITH_SIGNATURE;
+import static com.appsmith.server.constants.ApiConstants.DATE;
+
+public class SignatureVerifier {
+    private static final String publicVerificationKey = "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAICNwJ+zx2opXjjOga/YyzRxb2czvNgQ/twA+miCKDIX3 appsmith";
```
sharat87, line 22: Can we rename this constant to `PUBLIC_VERIFICATION_KEY` please?

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/SignatureVerifier.java
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * Method to verify the API signature from CS.
+     * @param headers   Response headers from CS
+     * @return          If the signature is valid
+     */
+    public static boolean isSignatureValid(HttpHeaders headers) {
+        if (CollectionUtils.isEmpty(headers.get(APPSMITH_SIGNATURE))) {
+            return false;
+        }
+        String signature = Objects.requireNonNull(headers.get(APPSMITH_SIGNATURE)).get(0);
```
sharat87, line 37: The `isEmpty` above will do null-check too, so we can remove the `requireNonNull` call here.

```suggestion
        String signature = headers.get(APPSMITH_SIGNATURE).get(0);
```

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/SignatureVerifier.java
```diff
@@ -0,0 +1,78 @@
...
+        String date = Objects.requireNonNull(headers.get(DATE)).get(0);
```
sharat87, line 38: I think we want an explicit null-check here, instead of `requireNonNull`.

If I understand right, if the `Date` header is missing, we want to return `false`, instead of throwing an exception, right? Because that's what `requireNonNull` will do. If the parameter is `null`, it'll throw an NPE, which would've been thrown even if that `requireNonNull` call isn't there on this line. 🤷

app/server/appsmith-server/src/main/java/com/appsmith/server/constants/ApiConstants.java
```diff
@@ -0,0 +1,7 @@
+package com.appsmith.server.constants;
+
+public class ApiConstants {
+    public static final String INSTANCE_VERSION = "Appsmith-Version";
+    public static final String DATE = "Date";
+    public static final String APPSMITH_SIGNATURE = "Appsmith-Signature";
```
sharat87, line 6: This is a HTTP header, right? If yes, let's use `X-Appsmith-Signature`. Custom headers are recommended to start with `X-`.

That said, this particular header, `X-Appsmith-Signature` is something we're already using for a very different purpose. You can search the code base for this. It's used to add a signature to API requests made from Appsmith as part of API actions, so that the target endpoint can verify that the request really came from Appsmith.

If this is about to cause confusion, consider using a different name for the header here. 🙂

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/SignatureVerifier.java
```diff
@@ -0,0 +1,78 @@
...
+        return isSignatureValid(signature, date);
+    }
+
+    private static boolean isSignatureValid(String signature, String dateHeader) {
+
+        if (signature.split("\\.", 2).length != 2) {
+            return false;
+        }
+        String signingData = signature.split("\\.", 2)[0];
+        String encodedSignature = signature.split("\\.", 2)[1];
```
sharat87, line 51: Minor refactoring.

```suggestion
        String[] signatureParts = signature.split("\\.", 2);
        if (signatureParts.length != 2) {
            return false;
        }
        String signingData = signatureParts[0];
        String encodedSignature = signatureParts[1];
```

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/SignatureVerifier.java
sharat87: Also, looks like we only need the key part of this string. So, how about:

```suggestion
    // Public key for verifying signatures with ED25519 scheme.
    private static final String PUBLIC_VERIFICATION_KEY = "AAAAC3NzaC1lZDI1NTE5AAAAICNwJ+zx2opXjjOga/YyzRxb2czvNgQ/twA+miCKDIX3";
```

Then we don't have to do the `.split` to get the key every time.

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/SignatureVerifier.java
```diff
@@ -0,0 +1,78 @@
...
+
+        // Decode base64 signature and signing data to byte arrays
+        byte[] signatureBytes = Base64.getUrlDecoder().decode(encodedSignature);
+        byte[] signingDataBytes = signingData.getBytes(StandardCharsets.UTF_8);
+
+        String publicKeyBody = publicVerificationKey.split(" ")[1];
+        AsymmetricKeyParameter publicKeyParameters = OpenSSHPublicKeyUtil.parsePublicKey(Base64.getDecoder().decode(publicKeyBody));
```
sharat87, line 58: Can we do this parsing once, when the class is loaded, instead of on every verification method call?

sharat87 (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/SignatureVerifier.java
abhvsn: Will remove this I was of the same opinion but IntelliJ started throwing warning hence kept it.

app/server/appsmith-server/src/main/java/com/appsmith/server/constants/ApiConstants.java
abhvsn: Oh thanks this is new we are using `X-APPSMITH-SIGNATURE` but this can get easily confusing, replacing this with `X-Appsmith-CS-Signature`.

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/SignatureVerifier.java
abhvsn: Done

app/server/appsmith-server/src/main/java/com/appsmith/server/constants/ApiConstants.java
sharat87: Headers are case-insensitive. So `X-Appsmith-Signature` and `X-APPSMITH-SIGNATURE`, are the same thing for HTTP.

abhvsn: Hi @sharat87 we have already replaced the name with `X-Appsmith-CS-Signature` can you take a look at it if anything needs to be changed.

sharat87:
Unmarking myself for review, since this PR is in draft status. Please mark me again once it's ready for review. Thanks. 🙏

app/server/appsmith-server/src/test/java/com/appsmith/server/helpers/SignatureVerifierTest.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    @Test
+    public void invalidSignature_verifySignature_returnFalse() {
+
+        HttpHeaders headers = new HttpHeaders();
+        headers.set(DATE, Instant.now().toString());
+        headers.set(CLOUD_SERVICES_SIGNATURE, UUID.randomUUID().toString());
+        Assertions.assertFalse(SignatureVerifier.isSignatureValid(headers));
+    }
```
vishnu-gp, line 31: @abhvsn Should we have a valid signature too with a hardcoded date and signature?

vishnu-gp (approved):
Minor suggestion. LGTM

app/server/appsmith-server/src/test/java/com/appsmith/server/helpers/SignatureVerifierTest.java
abhvsn: Issue here is on Appsmith instance we don't have a way to generate the valid signature. Also if you check we have a protection against the replay attack where we say the signature is valid for only 24hrs.

PR #24822, opened 2023-06-26
chore: Apply nginx env vars on 404.html

hetunandu:
## Description

Apply nginx env variables of index.html to 404.html as well.

#### Type of change


sharat87:
Removing myself for review. I don't think this is ready for review just yet. Let me know if that's not correct.

PR #24846, opened 2023-06-26
chore: Added provision for specific basic auth creds for actuator endpoints

nidhi-nair:
## Description
Adds provision to use custom basic authentication creds for actuator endpoints. If none are provided, we would still need to use basic auth header with empty creds.

This PR requires us to make the relevant changes on cloud (add env vars and change Grafana settings) before it is merged into any of the branches.

Fixes https://github.com/appsmithorg/appsmith-ee/issues/1441

>
#### Type of change
>
>
- [x] Manual
>
>
#### Dev activity

#### QA activity:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
@@ -12,13 +12,16 @@ import com.appsmith.server.helpers.RedirectHelper;
 import com.appsmith.server.services.AnalyticsService;
 import com.appsmith.server.services.UserService;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.core.Ordered;
+import org.springframework.core.annotation.Order;
 import org.springframework.core.io.ClassPathResource;
 import org.springframework.http.HttpMethod;
-import org.springframework.security.authentication.ReactiveAuthenticationManager;
 import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
 import org.springframework.security.config.annotation.method.configuration.EnableReactiveMethodSecurity;
 import org.springframework.security.config.annotation.web.reactive.EnableWebFluxSecurity;
@@ -30,6 +33,7 @@ import org.springframework.security.web.server.ServerAuthenticationEntryPoint;
 import org.springframework.security.web.server.authentication.ServerAuthenticationEntryPointFailureHandler;
 import org.springframework.security.web.server.authentication.ServerAuthenticationFailureHandler;
 import org.springframework.security.web.server.authentication.ServerAuthenticationSuccessHandler;
+import org.springframework.security.web.server.util.matcher.PathPatternParserServerWebExchangeMatcher;
 import org.springframework.security.web.server.util.matcher.ServerWebExchangeMatchers;
 import org.springframework.web.reactive.function.server.RouterFunction;
 import org.springframework.web.reactive.function.server.RouterFunctions;
@@ -40,7 +44,6 @@ import org.springframework.web.server.session.WebSessionIdResolver;
 import reactor.core.publisher.Mono;
 
 import java.time.Duration;
-import java.util.ArrayList;
 import java.util.HashSet;
 
 import static com.appsmith.server.constants.Url.ACTION_COLLECTION_URL;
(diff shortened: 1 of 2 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
```diff
@@ -176,6 +192,19 @@ public SecurityWebFilterChain securityWebFilterChain(ServerHttpSecurity http) {
...
+    private ReactiveAuthenticationManager getInternalAuthenticationManager() {
+        return authentication -> {
+            String name = authentication.getName();
+            String password = authentication.getCredentials().toString();
+
+            if (INTERNAL_USER_NAME.equals(name) && INTERNAL_PASSWORD.equals(password)) {
+                return Mono.just(UsernamePasswordAuthenticationToken.authenticated(name, password, new ArrayList<>()));
+            } else {
+                return Mono.just(UsernamePasswordAuthenticationToken.unauthenticated(name, password));
+            }
```
sharat87, line 204: Can we change this to keep backwards-compatibility? If username and password aren't set, can we allow this API to be accessible as in the past? Or are we expecting sensitive data to now go out of this endpoint?

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
```diff
@@ -85,6 +91,12 @@ public class SecurityConfig {
     @Autowired
     private RedirectHelper redirectHelper;
 
+    @Value("${appsmith.internal.username}")
+    private String INTERNAL_USER_NAME;
```
sharat87, line 95: Minor, but this'll be uniform with the env variable.

```suggestion
    private String INTERNAL_USERNAME;
```

sharat87 (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
nidhi-nair: @sharat87 Shall I set another environment variable for this that determines whether auth needs to be used? Probably more foolproof than checking for blank fields.

sharat87: Oh nah that's too much for this I think. In fact, I was going to suggest removing username, and just have password. Basic auth can work with just password. One configuration is more that enough for this.

Checking for blank is fine enough, I think. If blank, allow them, if not, require authentication.

nidhi-nair: Okay so Spring does not have an `isBlank` check on conditional properties. I've set up a separate prop for the config right now. Do you think we should make a custom annotation for the check instead? I'm a little divided between maintainability and configurability.

app/server/appsmith-server/src/main/resources/application.properties
```diff
@@ -110,3 +110,7 @@ appsmith.admin.envfile=${APPSMITH_ENVFILE_PATH:/appsmith-stacks/configuration/do
 
 # RTS port
 appsmith.rts.port=${APPSMITH_RTS_PORT:8091}
+
+appsmith.internal.password=${APPSMITH_INTERNAL_PASSWORD:}
+appsmith.internal.auth.enabled=${APPSMITH_INTERNAL_AUTH_ENABLED:false}
+#random.prop=false
```
sharat87, line 116: Not random enough I think. :P

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
```diff
@@ -176,6 +210,14 @@ public SecurityWebFilterChain securityWebFilterChain(ServerHttpSecurity http) {
                 .build();
     }
 
+    private boolean isAuthorizedToAccessInternal(String password) {
+        if (StringUtils.isNotEmpty(INTERNAL_PASSWORD)) {
+            return INTERNAL_PASSWORD.equals(password);
+        } else {
+            return true;
+        }
```
sharat87, line 218: Feel free to not agree with this suggestion. 🙂

```suggestion
        // Either configured password is empty, or it's equal to what we received.
        return StringUtils.isEmpty(INTERNAL_PASSWORD) || INTERNAL_PASSWORD.equals(password);
```

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
```diff
@@ -109,6 +120,29 @@ public ForwardedHeaderTransformer forwardedHeaderTransformer() {
         return new ForwardedHeaderTransformer();
     }
 
+    @Order(Ordered.HIGHEST_PRECEDENCE)
+    @Bean
+    @ConditionalOnProperty(name = "appsmith.internal.auth.enabled")
```
sharat87, line 125: Can we use just the password for this? I see we have the implementation where, if password is empty, then auth is always accepted anyway, right? I think we can use `ConditionalOnExpression` to make this conditional on whether `password` is non-empty. One less configuration to care about. What say?

sharat87 (changes requested):

sharat87 (approved):

PR #24924, opened 2023-06-29
feat: Add support for connecting to Redis clusters

sharat87:
This PR adds support to Appsmith server to be able to connect to ElastiCache Redis with cluster mode turned on. When the `APPSMITH_REDIS_URL` is set to `redis://...`, the current default behaviour is preserved. But if it is set to `redis-cluster://...`, then we setup a pooled connection with cluster mod enabled.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/RedisConfig.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/RedisConfig.java
@@ -12,6 +12,11 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Primary;
 import org.springframework.data.redis.connection.ReactiveRedisConnectionFactory;
+import org.springframework.data.redis.connection.RedisClusterConfiguration;
+import org.springframework.data.redis.connection.RedisNode;
+import org.springframework.data.redis.connection.lettuce.LettuceClientConfiguration;
+import org.springframework.data.redis.connection.lettuce.LettuceConnectionFactory;
+import org.springframework.data.redis.connection.lettuce.LettucePoolingClientConfiguration;
 import org.springframework.data.redis.connection.lettuce.observability.MicrometerTracingAdapter;
 import org.springframework.data.redis.core.ReactiveRedisOperations;
 import org.springframework.data.redis.core.ReactiveRedisTemplate;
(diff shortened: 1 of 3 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/RedisConfig.java
```diff
@@ -49,6 +61,31 @@ ChannelTopic topic() {
...
+                return new LettuceConnectionFactory(redisUri.getHost(), redisUri.getPort());
+            }
+
+            case "redis-cluster" -> {
+                // For ElastiCache Redis with cluster mode enabled, with the configuration endpoint.
+                final LettuceClientConfiguration config = LettucePoolingClientConfiguration
+                    .builder()
+                    .build();
+                final RedisClusterConfiguration clusterConfig = new RedisClusterConfiguration();
+                clusterConfig.addClusterNode(new RedisNode(redisUri.getHost(), redisUri.getPort()));
```
nidhi-nair, line 81: Is there any capability to use the URI directly as a configuration here? Asking in case Redis urls support setting extra configs as params, like in Mongo.

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/RedisConfig.java
```diff
@@ -49,6 +61,31 @@ ChannelTopic topic() {
...
+                return new LettuceConnectionFactory(clusterConfig, config);
+            }
+
+            default -> throw new InvalidRedisURIException("Invalid redis scheme: " + scheme);
```
nidhi-nair, line 85: Nit: Do we not need to wrap this in an AppsmithException?

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/RedisConfig.java
sharat87: Doesn't look like. I wished Spring supported doing this with a custom scheme like we implemented here, but all the docs say is to use the `RedisClusterConfiguration` in code. But for us, in order to support both standalone Redis, and clustered Redis based on the same env variable, this appeared to be the best way.

I don't know if Redis supports extra parameters in the configuration, from the Lettuce code I've seen, doesn't look like. We'll have to take it up as a task/feature when needed.

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/RedisConfig.java
sharat87: I actually wanted this as a `RuntimeException` directly. But IntelliJ wasn't happy. I didn't make it an `AppsmithException` because this exception will be thrown before the server even starts, so not really an application exception, so, a status code and such detail didn't make sense.

Not a strong opinion though. Happy to change if that logic doesn't make sense. 🙂

nidhi-nair (approved):

PR #24953, opened 2023-06-30
chore: disable cra eslint

KelvinOm:
## Description
Disable CRA built-in ESLint checks since we have our own config and a separate step for this

#### Type of change

- [x] Manual
- [ ] Jest
- [ ] Cypress

#### Dev activity

### DIFF
--- a/app/client/.eslintrc.base.json
+++ b/app/client/.eslintrc.base.json
@@ -53,8 +53,8 @@
       "error",
       { "caseSensitive": false }
     ],
-    "no-console": "warn",
-    "no-debugger": "warn",
+    "no-console": "error",
+    "no-debugger": "error",
     "@typescript-eslint/no-restricted-imports": [
       "error",
       {
--- a/app/client/build.sh
+++ b/app/client/build.sh
@@ -18,6 +18,8 @@ fi
 # build cra app
 export REACT_APP_SENTRY_RELEASE=$GIT_SHA
 export REACT_APP_CLIENT_LOG_LEVEL=ERROR
+# Disable CRA built-in ESLint checks since we have our own config and a separate step for this
+export DISABLE_ESLINT_PLUGIN=true
 craco --max-old-space-size=7168 build --config craco.build.config.js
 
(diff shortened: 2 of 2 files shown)


rajatagrawal:
Tested the bug scenario on my local setup with this fix and it is working now. The lint was previously not working for this [commit](https://github.com/appsmithorg/appsmith/commit/ed55c72f93574b50316fe22f95032d384aa8bf04).

LGTM

rajatagrawal (approved):

app/client/.eslintrc.base.json
```diff
@@ -53,8 +53,8 @@
       "error",
       { "caseSensitive": false }
     ],
-    "no-console": "warn",
-    "no-debugger": "warn",
+    "no-console": "error",
+    "no-debugger": "error",
```
Rishabh-Rathod, line 57: @KelvinOm Could you please help me understand the reasoning behind making the above change from warning to error for the console and debugger?

This was intentionally kept as warning to make the development easier but also block console and debugger code from passing lint checks in post-commit hooks. 

Now with this change, every console or debugger line needs a eslint ignore comment which the code more probable to pass the lint check and getting committed.

rajatagrawal: @Rishabh-Rathod 

This was added because console logs were getting undetected in post commit hook but were failing checks in CI. This caused a lot of inconvenience because an expected build success would fail.

What are the use cases that you need eslint ignore comment for ? For temporary commits or refreshing server ?

Rishabh-Rathod: We can make the pre-commit check to fail for eslint warnings ( [example](https://stackoverflow.com/questions/51158815/running-eslint-in-precommit-does-not-stop-on-warnings) ). 

> What are the use cases that you need eslint ignore comments for? For temporary commits or refreshing the server?

While debugging using console.log or debugger as currently, it throws an error the CRA server doesn't refresh and until we fix the error i.e., either remove logs or add comments to ignore it, we are blocked by the error.

PR #24610, opened 2023-06-19
fix: change to lodash wrapper

keyurparalkar:
## Description
This issue is happening because we are getting lodash library as `undefined` when we pass it to` derived.value` function here. This is happening because of the following reason:
* [Bundle optimisation PR](https://github.com/appsmithorg/appsmith/pull/21667) introduces a new plugin called: `babel-plugin-lodash`. It helps in transforming the imports. 
* But this plugin doesn't work on the following pattern: `import _ from "lodash"`. Here it replaces `_` with `undefined`
[ This file](https://github.com/appsmithorg/appsmith/blob/3bc50d9632cf269935aed6657092ddb8b7e8c92f/app/client/src/workers/common/JSLibrary/lodash-wrapper.js#L1) explains the exact scenario that we are facing.
There is also [an open bug](https://github.com/lodash/babel-plugin-lodash/issues/235) for this issue on the `babel-plugin-lodash` repository.

This PR imports lodash from the lodash-wrapper instead of directly importing as follows: `import _ from "lodash"`
>
#### PR fixes following issue(s)
Fixes #23671

#### Type of change
- Bug fix (non-breaking change which fixes an issue)

#### How Has This Been Tested?

- [x] Manual
- [ ] Jest
- [x] Cypress
    - should check that value entered in currency field appears in the actual field
>
>
#### Dev activity
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] PR is being merged under a feature flag

#### QA activity:
- [ ] [Speedbreak features](https://github.com/appsmithorg/TestSmith/wiki/Guidelines-for-test-plans#speedbreakers-) have been covered
- [ ] Test plan covers all impacted features and [areas of interest](https://github.com/appsmithorg/TestSmith/wiki/Guidelines-for-test-plans#areas-of-interest-)
- [ ] Test plan has been peer reviewed by project stakeholders and other QA members
- [ ] Manually tested functionality on DP
- [ ] We had an implementation alignment call with stakeholders post QA Round 2
- [ ] Cypress test cases have been added and approved by SDET/manual QA
- [ ] Added `Test Plan Approved` label after Cypress tests were reviewed
[...]

PR #24969, opened 2023-06-30
perf: reduce the bundle size, vol. 2

iamakulov:
## Description

This PR takes the second (and last) round on reducing the bundle size. Specifically, it:
- reverts a regression that caused Lodash to be bundled in full
- code-splits away Uppy, `react-json-view`, and the welcome tour `app.json`
- removes `moment-timezone` (only from the main thread code) and unused code syntax highlighter definitions

This makes the bundle another 15% smaller (5.42 MB → 4.64 MB minified). In total, since [the first PR](https://github.com/appsmithorg/appsmith/pull/20964), we’ve shaved off 49% of the bundle (9.01 → 4.64 MB minified)!

Here’s the bundle before the change (with removed and code-split modules highlighted):

Screenshot 2023-06-30 at 13 10 25

Note: this PR relies on https://github.com/appsmithorg/design-system/pull/514 to fully code-split Uppy away.

#### Type of change

- [x] Manual: deep testing of the new File Picker, smoke testing of the guided tour and other changes
- [ ] Jest
- [ ] Cypress

#### Dev activity

#### QA activity:


iamakulov:
Flaky tests? Let’s try again:

/ok-to-test

app/client/src/WidgetQueryGenerators/PostgreSQL/index.ts
```diff
@@ -1,5 +1,5 @@
 import { BaseQueryGenerator } from "../BaseQueryGenerator";
-import { format } from "sql-formatter";
+import { formatDialect, postgresql } from "sql-formatter";
```
SatishGandham, line 2: @iamakulov , why is this change required?

app/client/src/workers/common/JSLibrary/lodash-wrapper.js
```diff
@@ -1,3 +1,53 @@
+// 🚧 NOTE: this file exists only for the worker thread, as the worker thread needs to pass
```
KelvinOm, line 1: To avoid a full lodash load can we use [this eslint rule](https://github.com/wix-incubator/eslint-plugin-lodash/blob/master/docs/rules/import-scope.md)?

app/client/src/WidgetQueryGenerators/PostgreSQL/index.ts
iamakulov: That’s because `sql-formatter` bundles a whole lot of dialect formatters, by default – and we only need one of them: https://github.com/sql-formatter-org/sql-formatter/issues/452

app/client/src/workers/common/JSLibrary/lodash-wrapper.js
iamakulov: Hmmm so right now [we’re relying on `babel-plugin-lodash`](https://github.com/appsmithorg/appsmith/blob/f80321aa6f868d3d8441a3bdc2bdb07415c1201f/app/client/craco.common.config.js#L18) to avoid full Lodash imports. Do you mean replacing `babel-plugin-lodash` with `eslint-plugin-lodash`, or using one in addition to another?

(If the former: I moderately prefer `babel-plugin-lodash` to ESLint rules. The former works everywhere, whereas the latter is one `// eslint-disable-next-line` away from bundling the full Lodash again. But if you feel this `lodash-wrapper.js` thingy is too much complexity, I’m happy to reconsider!)

KelvinOm: I would suggest using `eslint-plugin-lodash` in addition to `babel-plugin-lodash`.

If I understand correctly what is going on here, then when using the `eslint-plugin-lodash`, we will have to correct the imports and everything related to lodash-wrapper.js will become easier, right?

iamakulov: Hmmm. So, on a second thought, I think `eslint-plugin-lodash` are `babel-plugin-lodash` aren’t really compatible:

### `eslint-plugin-lodash` vs `babel-plugin-lodash`

`eslint-plugin-lodash` forces developers to do this change:

```js
// Before
import _ from "lodash"
_.map()

// After
import map from "lodash/map"
map()
```

But that’s pretty much what `babel-plugin-lodash` does automatically! We can either force the developers to change all imports themselves (with `eslint-plugin-lodash`), or we can use `babel-plugin-lodash` to do that automatically, but using both of them doesn’t make much sense.

### Why `lodash-wrapper.js` exists

The reason `lodash-wrapper.js` exists is this:

https://github.com/appsmithorg/appsmith/blob/ea4840b47a898aaebfeea8836e5a8ed99f24ebca/app/client/src/workers/common/JSLibrary/lodash-wrapper.js#L49-L61

This is a workaround for [a bug in `babel-plugin-lodash`](https://github.com/lodash/babel-plugin-lodash/issues/235). 

If we use `eslint-plugin-lodash` together with `babel-plugin-lodash`, `lodash-wrapper.js` will still be needed – as `babel-plugin-lodash` will still be enabled and will still be introducing the bug.

If we use only `eslint-plugin-lodash`, `lodash-wrapper.js` won’t be needed anymore – we’ll be able to just `// eslint-disable-next-line` in the worker where we need to import full Lodash.

### Tradeoffs

So, basically, here are the tradeoffs of each:

| `eslint-plugin-lodash` | `babel-plugin-lodash` |
| --- | --- |
| Requires the developer to change the import | Changes the import automatically |
| More explicit (= easier to debug) | More implicit (= harder to debug) |
| Easier to accidentally disable in some file | Harder to accidentally disable in some file |
| (Probably) bug-free | Has a non-obvious tricky bug with `_` becoming `undefined` which [we already stumbled upon](https://github.com/appsmithorg/appsmith/pull/24610) |

I’m pretty neutral about which one to pick – which tradeoffs, do you think, matter more here?

app/client/src/utils/importUppy.ts
```diff
@@ -0,0 +1,36 @@
+// We’re setting this flag to true when we know for sure the Uppy module was loaded and initialized.
+// When it’s `true`, the other modules will know that the importUppy function will resolve immediately
+// (in the next tick). They can use it to e.g. decide whether to show the loading spinner
+export let isUppyLoaded = false;
+
+export async function importUppy() {
+  const [Uppy, Dashboard, GoogleDrive, OneDrive, Url, Webcam] =
```
SatishGandham, line 7: We are not handling the case where the import fails. @iamakulov , can we handle this please.
One case where this will be a problem is, the instance upgrade while the use is on a different page and deleting all the previous JS files on server.

Now when the older js code requests these old files, they are no longer present on the server.

SatishGandham:
LGTM except the chunk loading error handling part.

We need QA on this for file picker.

### File picker
- See if there is any delay in opening the file picker
- There might be a slight delay, but does that hamper the UX

@sbalaji1192 , please review the PG query generator changes.

app/client/src/workers/common/JSLibrary/lodash-wrapper.js
KelvinOm: First of all thank you for such an answer. Excellent engineering approach! 👍 

I would suggest using `eslint-plugin-lodash` as it is simpler and more obvious. I'm in favor of keeping things as simple as possible. So any new developer will be able to quickly figure out what's going on and make changes. 

@riodeuno @SatishGandham need your comment here.

riodeuno: I second @KelvinOm 's opinion on this. `More explicit (= easier to debug)` had me sold.

What concerns me is this: `Easier to accidentally disable in some file`. How do we mitigate this? Any thoughts?

sbalaji1192 (approved):
Postgres SQL formatting looks good.

app/client/src/workers/common/JSLibrary/lodash-wrapper.js
KelvinOm: > One way to catch this is to set up bundle size monitoring that triggers an alert if the bundle suddenly increases by X% (where X is 5-10%+). I like using SpeedCurve for that. Happy to guide you through that!]

I'm sure we should set up bundle size monitoring anyway, not just because of lodash.

app/client/src/utils/importUppy.ts
iamakulov: Oh, great catch!

Thinking about this – I believe we can’t solve

> the instance upgrade while the user is on a different page and deleting all the previous JS files on server.

without reloading the page.

When we write

```js
import("./some-module")
```

webpack [compiles that down to](https://github.com/webpack/webpack/tree/main/examples/code-splitting-harmony)

```js
__webpack_require__.e(/*! import() */ 644)
```

where `__webpack_require__.e` is the function that loads the code-split chunk, and `644` is the module ID. The mapping between the module ID (`644`) and the actual file URL (`/static/644.[some-hash].js`) is hard-coded in the bundle (specifically, in the webpack runtime); e.g. here’s how it looks at `app.appsmith.com`:

Screenshot 2023-07-04 at 10 06 40

As the URLs are hard-coded in the bundle, they won’t change until the page reloads. So if an instance upgrade deletes `/static/644.[some-hash].js` from the server, retrying `import("./some-module")` will still resolve to `/static/644.[some-hash].js` and still fail! The user will need to reload the page to get the new bundle with the new URLs.

### Also: service worker

Keep in mind that we also have a preloading service worker. On a fast enough network, the moment the user visits _any_ page, we’ll preload _all_ code-split chunks within 10-30 seconds. So this error mode will actually be very rare.

### Also: not solvable on the frontend

There might be some tricky ways to make `import()` work even after an instance upgrade. But, in general, I believe, it’s a bad idea to try and re-fetch chunks after an upgrade. I.e., if
- a user visits Appsmith while the instance is at version `v111`
- then, the instance upgrades to `v112`
- then, the user tries to do `import("./some-module")` something

we _don’t_ want to serve the chunk that was generated by `v112`.

That’s because the code expectations might change between `v111` and `v112`. What if `v112` introduced some mandatory prop that `v111` doesn’t provide? Or what if `v112` starts handing some prop differently? That’ll lead to bugs or (in the worst case) data corruption.

Instead, the way I’ve seen clients typically solve this is by _keeping_ old (`v111`) files when deploying a new (`v112`) version. The clients I worked with typically achieve this with either
- a CDN – with a CDN, it’s easy (as long as you don’t do a purge, old files will stay in cache for some time)
- or S3 – when you store assets in S3, they just stay there even if the instance restarts

I haven’t seen clients solve this with just a bare server (most clients would have some CDN in front of the app), but I imagine it’s solvable with a bare server as well (eg we can keep some shared drive mount that retains the files between restarts?).

iamakulov: > the instance upgrade while the user is on a different page and deleting all the previous JS files on server.

So, to summarize, basically, this ↑ failure mode

- already exists (even [`retryPromise`](https://github.com/appsmithorg/appsmith/blob/ea4840b47a898aaebfeea8836e5a8ed99f24ebca/app/client/src/utils/AppsmithUtils.tsx#L334) doesn’t help with it)
- is rare (thanks to a service worker)
- and I’d solve it on the infra level, not on the frontend level. 

Are there any other cases where we might want to retry an `import()`? I imagine a bad network might be one, but a bad network will fuck up a lot of things (API calls, other chunks), so it’d need a fully thought-out solution, not just some import retries.

app/client/src/workers/common/JSLibrary/lodash-wrapper.js
iamakulov: Deal! Then I’ll PR `eslint-plugin-lodash` as a separate PR, to avoid blocking this one 👍

app/client/src/utils/importUppy.ts
iamakulov: > It's not rare, happened a lot of times until we put CDN.

Oh interesting! So the service worker wasn’t sufficient enough.

> I was thinking of informing the user to reload instead of crashing the UI. Any ideas to address it better?

Hmmm. If users do navigation events frequently enough (eg navigating from the editor to the view mode?), I guess we can check if the new version was released – and then force-reload the page on an SPA navigation?

But I’d only do that (and the “Please reload” toast) if the CDN is not available – as that’s a UX regression. (Maybe `areOldBuildResultsAvailable` can be a config flag?)

laveena-en:
@iamakulov The PR mentions `...smoke testing of guided tour and other changes`. Can you please mention the scope of "other changes" stated here for testing?
cc: @SatishGandham

iamakulov:
> Can you please mention the scope of "other changes" stated here for testing?

Sure – that was basically just verifying that the bundle changes applied + making sure the app loads / doesn’t throw any new errors. (I didn’t do deeper testing as the changes seemed safe enough + I was hoping Cypress will catch any regressions.)

laveena-en:
Tested and verified for the filepicker and guided tour. PR looks good to me!

rahulramesha (dismissed):
UI builder file changes looks good to me

tanvibhakta:
Sorry @rahulramesha , @sbalaji1192 need your reviews here again! I've just updated the design-system-old package so that the changes in this PR https://github.com/appsmithorg/design-system/pull/514 show up in this one.

iamakulov:
> Sorry @rahulramesha , @sbalaji1192 need your reviews here again!

FWIW the `design-system` change affects neither the file picker nor Postgres formatting – it’s fully separate. So probably these reviews could be skipped? (But ofc happy to defer to you if you believe they shouldn’t!)

app/client/src/WidgetQueryGenerators/PostgreSQL/index.ts
SatishGandham: @iamakulov , shouldn't we also change the format import in other places like
https://github.com/appsmithorg/appsmith/blob/release/app/client/src/WidgetQueryGenerators/MySQL/index.ts#L3
https://github.com/appsmithorg/appsmith/blob/release/app/client/src/WidgetQueryGenerators/Snowflake/index.ts#L2
https://github.com/appsmithorg/appsmith/blob/release/app/client/src/WidgetQueryGenerators/MSSQL/index.ts#L2

iamakulov: Yes! Looks like these were added recently during a recent merge. Thanks for the catch – I’ll update them and add a ESLint rule as well.

iamakulov:
↑ Very confused by all of these because
- the `FilePickerV2_Widget_Reskinning_spec.js` issue is transient – I reproduced it once – but then I stopped being able to repro it
- other failures are mostly about app rename suddenly not working?

Let me re-run.

iamakulov:
Still failing. I’m running a workshop on Thu so will investigate and fix on Fri 👍

iamakulov:
↑ The issue should be fixed now!

iamakulov:
Okay, so it looks like [my test run](https://github.com/appsmithorg/appsmith/pull/24969#issuecomment-1636721161) actually failed to run some tests (due to 429 Too Many Requests).

We then [re-ran the tests](https://github.com/appsmithorg/appsmith/pull/24969#issuecomment-1637005572), but re-ran them with the outdated code – thus still including the issue I fixed in https://github.com/appsmithorg/appsmith/pull/24969/commits/acdbc51692130aef3329742fa4f46c22c6722c24:

CleanShot 2023-07-16 at 16 24 15@2x

To make sure we’re running the latest code from this branch, I’m going to trigger a fresh test run.

SatishGandham (approved):

PR #25014, opened 2023-07-03
chore: Set reclaim policy to Retain

pratapaprasanna:


deploy/helm/templates/persistentVolume.yaml
```diff
@@ -12,7 +12,7 @@ spec:
   {{- range .Values.persistence.accessModes }}
     - {{ . | quote }}
   {{- end }}
-  persistentVolumeReclaimPolicy: Delete
+  persistentVolumeReclaimPolicy: {{ .Values.persistence.ReclaimPolicy }}
```
sharat87, line 15: Please consider the case of folks updating their current installation to the new Helm charts. Should we change this to `Retain` for them, or keep it to `Delete`? Depending on that, let's add the default here:

```
  persistentVolumeReclaimPolicy: {{ .Values.persistence.ReclaimPolicy | default "Retain" }}
```

sharat87 (changes requested):
Can we either link to an issue with details, or include the details in the PR description itself, on what problem we're solving here please?

sharat87:
@pratapaprasanna, do we still need this? Please ensure you follow up on PRs to take them forward. Thanks! 🙏

sharat87:
Changes in this PR are no-longer needed.

Issue #23713, opened 2023-05-25
[Bug]: Application Welcome page/Application page not loading randomly


Vijetha-Kaja:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

http://localhost/setup/welcome is not loading randomly in CI
![cypress-screenshot (30)](https://github.com/appsmithorg/appsmith/assets/119562824/d2c845af-c4d3-4fed-9b7e-c844ac9ca179)

### Steps To Reproduce

1. Navigate to http://localhost/setup/welcome

### Environment

Release

### Issue video log

https://github.com/appsmithorg/appsmith/assets/119562824/939d016e-871c-4501-826e-928f20e4c3b5

### Version

Self hosted

Vijetha-Kaja:
[Logs](https://github.com/appsmithorg/appsmith/files/11561481/1_ci-test.ci-test.0.txt) to help debug

Vijetha-Kaja:
Latest [run](https://cloud.cypress.io/projects/eyxvp8/runs/35154/test-results/2bcb39ce-0e71-4814-9259-d0b8c5c2ceb5) with above failure - [Logs](https://github.com/appsmithorg/appsmith/files/11592751/38_ci-test.ci-test.37.txt)

Vijetha-Kaja:
Latest [run](https://cloud.cypress.io/projects/eyxvp8/runs/35203/test-results) with above failures - [Logs](https://github.com/appsmithorg/appsmith/files/11599911/2_ci-test.ci-test.1.txt)

hetunandu:
There are two different scenarios in the failed runs. 
1. the welcome page does not come up at all (first step) 
2. the editor is loaded successfully (after login and create new application) but no app content is showing up. 

First seems flaky cause the server is not up. Second could be a concern but without logs not able to understand the problem. The error toast message in the video is empty as well.

Vijetha-Kaja:
Latest [run](https://cloud.cypress.io/projects/n14p1y/runs/7459/test-results) with above failure - [Logs](https://github.com/appsmithorg/appsmith/files/11704651/29_Run.the.cypress.test.txt)

Vijetha-Kaja:
[...]

PR #25037, opened 2023-07-03
chore: show stack trace with the failing action type for cypress

akash-codemonk:
Adds an additional toast to show on cypress when error message is an empty string or null to help with debugging. Showing the action name causing the error in the toast. Temporary will be removed after. 

Fixes https://github.com/appsmithorg/appsmith/issues/25182

Related to  https://github.com/appsmithorg/appsmith/issues/23713


app/client/src/sagas/ErrorSagas.tsx
```diff
@@ -226,6 +226,12 @@ export function* errorSaga(errorAction: ReduxAction<ErrorActionPayload>) {
       case ErrorEffectTypes.SHOW_ALERT: {
         // This is the toast that is rendered when any page load API fails.
         yield call(showToast, message, { kind: "error" });
+
+        if ((window as any).Cypress) {
+          if (message === "" || message === null) {
+            yield call(showToast, message + type, { kind: "error" });
```
hetunandu, line 232: If the message is empty string, how will it help to show it on a toast message?

Can we do a console log instead? We can probably use this
https://github.com/appsmithorg/appsmith/blob/release/app/client/cypress/plugins/index.js#L66

Another option is to use Cypress.log
https://docs.cypress.io/api/cypress-api/cypress-log

hetunandu (changes requested):

app/client/src/sagas/ErrorSagas.tsx
akash-codemonk: Yup `message` is not useful. The `type` will mention the action being dispatched. console.log or cypress.log is not useful since we only see a part of these logs in the cypress video recordings.  Have tried this before, these are completely missed. 

The toast with the action type will be easy to spot as we can already spot the empty toast now.

hetunandu: I still think just having the type would not be enough. That could be blank as well for all we know. 
We need a more solid way to identify where the error is originating from. We can consider crashing the page and printing the error in bold text on the screen.

akash-codemonk: Crash landing on cypress env with a stack trace. E.g screenshot
stack trace

hetunandu (approved):

PR #25104, opened 2023-07-05
chore: Stub /features call in tests

hetunandu:
This change will stub the call to `users/features` to get the feature flags. By default it will return an empty list and rely on the client default flag list to enable/disable features


akash-codemonk (approved):

Aishwarya-U-R (approved):
Approving Cypress changes!

Aishwarya-U-R:
@rahulbarwal - adding you as FYI - since you added the the fixtures that are removed now in this PR.

Issue #25127, opened 2023-07-05
[Bug]: S3 integration: implement fix for test datasource on access denied error for non Amazon service providers


sumitsum:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Ref: 

### Steps To Reproduce

.

### Environment

Production

### Version

cloud, self hosted

PR #25128, opened 2023-07-05
fix: fix Amazon S3 test datasource handling for access denied error

### LINKED ISSUE #3776: [Bug] S3: test datasource should return pass on permission error.
## Description
Currently, on bucket access permission error, s3 test datasource returns failure, it should return success as the datasource connection is valid.

24 May 2023:
Another case got reported recently: https://theappsmith.slack.com/archives/C0341RERY4R/p1684787842120239

sumitsum:
## Description
- Currently if an Amazon S3 user account does not have permission to list all buckets (but can access files in specific buckets), then the test datasource returns failure, because it relies on running `list buckets` command to test the status of connection (at least one command needs to be run to test status). In some cases, a user account may be valid and have access to files in specific buckets but may not have permission to list all buckets. In such cases currently, the test datasource feature returns failure which is not correct.
- This PR updates the test datasource code flow for Amazon S3 integration to check if the error returned by the remote server is specifically to deny list buckets action. If so, then test datasource would be assumed to be passed. 
- Please note that this change will only address the issue for Amazon S3 service and may not for other S3 service providers. However, at the moment it seems hard to implement the fix for other providers because (1) we don't have test accounts with other providers right now to simulate this behaviour (2) the usage for other service providers is extremely less in comparison and we have not see any such error reported for other service providers. Hence, right now, it does not seem like a good return on investment to pursue this fix for other service providers. I have created another [issue](https://github.com/appsmithorg/appsmith/issues/25127) to track it.  

Fixes #3776

#### Type of change

- [x] Manual
- [x] JUnit

## How to test ? 
- Get AWS S3 credentials from Sumit. 
- Check that for the above credentials test datasource fails on release but passes on this PR DP. 

#### Dev activity

#### QA activity:


app/server/appsmith-plugins/amazons3Plugin/src/main/java/com/external/plugins/AmazonS3Plugin.java
```diff
@@ -114,6 +115,7 @@ public class AmazonS3Plugin extends BasePlugin {
     private static final String OTHER_S3_SERVICE_PROVIDER = "other";
     private static final String AWS_S3_SERVICE_PROVIDER = "amazon-s3";
     public static String DEFAULT_FILE_NAME = "MyFile.txt";
+    public static final String ACCESS_DENIED_ERROR_CODE = "AccessDenied";
```
sondermanish, line 118: could we put this in FieldName in amazon s3 or rather in a Common FieldName

sumitsum: Hi @sondermanish thanks for the suggestion. This constant was not a good candidate for `FieldName` file because it contains path strings to actual fields in the query / datasource forms. As there seemed to be no other file that was a good fit I have now created a new file to hold such constants that don't seem to fit elsewhere :)

sondermanish (dismissed):

sumitsum:
Adding proxy approval because (0) PR was approved (1) approval was lost because of merge conflict resolution (2) Manish seems to be not available for a review right now.

sondermanish (approved):

PR #25162, opened 2023-07-06
chore: Add a check in chromatic workflow to run only on CE

jsartisan:
We don't want to run the chromatic build workflow on the EE repo. 
This PR adds a check in the workflow so that it runs only on the CE repo.


sum35h (approved):

jsartisan:
Closing this PR as not required anymore.

PR #25173, opened 2023-07-06
chore: Applied Spotless formatter

nidhi-nair:
Applied formatting rules based on Spotless


nidhi-nair:
Force merging based on review from #24175

PR #25211, opened 2023-07-07
ci: Add a note about Vercel DPs with env=release

sharat87:
Non-impacting change, only adds some more content to the DP URL comment.

More context around this at https://notion.so/031b87bce3404e3a95240d4c14c82e46.


nidhi-nair (approved):

nidhi-nair (approved):
Thanks for adding this!

PR #25228, opened 2023-07-10
ci: Added pre-commit hook to check for Spotless formatting

nidhi-nair:
Adds a check so that folks do not have to remember to have to build server when there are changes.

Also includes formatting for some files that were committed without merging from release.


mohanarpit:
@nidhi-nair There doesn't seem to be any changed files. Is this PR correct?

nidhi-nair:
@mohanarpit I was testing the hook, looks like we had a custom path that the commands needed to be moved to. Could you please review again?

mohanarpit (approved):

PR #25298, opened 2023-07-11
fix: removes race conditions in recent perf optimisation

hetunandu:
We missed a race condition in: #25104 which caused some test cases to fail. This would rarely happen to any new apps in production but to ensure backwards compatibility we are updating the code itself


arunvjn (approved):
Looks good

PR #25324, opened 2023-07-12
fix: pre-commit logic

KelvinOm:
## Description
Hot fix for pre-commit logic

### DIFF
--- a/app/client/.husky/check-staged-files.sh
+++ b/app/client/.husky/check-staged-files.sh
@@ -6,7 +6,12 @@ is_client_change=$(git diff --cached --name-only | grep -c "app/client")
 if [ "$is_server_change" -ge 1 ]; then
   echo "Running Spotless check ..."
     pushd app/server > /dev/null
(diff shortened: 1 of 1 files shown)


app/client/.husky/check-staged-files.sh
```diff
@@ -6,7 +6,12 @@ is_client_change=$(git diff --cached --name-only | grep -c "app/client")
 if [ "$is_server_change" -ge 1 ]; then
   echo "Running Spotless check ..."
     pushd app/server > /dev/null
-    (mvn spotless:check 1> /dev/null && popd > /dev/null) || (echo "Spotless check failed, please run mvn spotless:apply" && exit 1)
+    if (mvn spotless:check 1> /dev/null && popd > /dev/null) then
+      cd ../..
```
nidhi-nair, line 10: You can just use popd here to make use of the call stack above

dhruvikn (approved):

app/client/.husky/check-staged-files.sh
KelvinOm: fixed

dhruvikn (approved):

PR #25357, opened 2023-07-13
ci: Ad-hoc DP | Build Push Deploy from branch

sum35h:
Workflow and script to build from branch and deploy to uat-cluster.


sharat87:
@sum35h, do we intend to use this? My understanding was that triggering the workflow via API wasn't working very well, so we're not doing that and instead running it via the ops API? 🤔

sum35h:
> @sum35h, do we intend to use this? My understanding was that triggering the workflow via API wasn't working very well, so we're not doing that and instead running it via the ops API? 🤔

The ops API was only a workaround for the Hackathon demo, since building the image takes a lot of time (min 45 mins).
I'm planning to use this workflow  going forward, and will also use the Ops API to trigger this.

.github/workflows/ad-hoc-deploy-preview.yml
```diff
@@ -0,0 +1,603 @@
...
+        run: |
+          #  mongosh '' --eval 'db.appsmiht-ce-github-dp-runs.insert( { run_id: ${{ github.run_id }}, uid: ${{ github.event.inputs.uid }}, status: "Started" } )'
+          mongosh '' --eval 'db.appsmiht-ce-github-dp-runs.update(
+                                                                  { _id:  ${{ github.event.inputs.uid }} },
+                                                                  {
+                                                                    $set: {
+                                                                      status: "DEPLOYMENT IN PROGRESS"
+                                                                    }
+                                                                  }
+                                                                )'
```
sharat87, line 540: Did we get a chance to test this code? Doesn't look like it'll do what's intended.

.github/workflows/ad-hoc-deploy-preview.yml
sharat87, line 1: Can we reuse the `server-build.yml` and others instead of writing the workflow code for those steps here as well?

scripts/build_dp_from_branch.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
+# Params are in environment variables as PARAM_{SLUG}, e.g. PARAM_USER_ID
+
+# Configure the AWS & kubectl environment
+
+mkdir ~/.aws; touch ~/.aws/config
+
+echo "[default]
+aws_access_key_id = $AWS_ACCESS_KEY_ID
+aws_secret_access_key = $AWS_SECRET_ACCESS_KEY" > ~/.aws/credentials
```
sharat87, line 10: Can we avoid writing to default configuration files when possible please? We should be able to get the same affect even if we don't have this here.

```suggestion
```

sharat87 (changes requested):

.github/workflows/ad-hoc-deploy-preview.yml
sum35h: Due to the checkout step which uses` github.event.inputs.branch` unlike the `pr no` in workflows like `server-build.yml`, those will need to be edited to build from branch variable, if it is to be used in this workflow. Hence , went with writing these jobs again despite redundancy.

.github/workflows/ad-hoc-deploy-preview.yml
sum35h, line 540:
```suggestion
         # mongosh '' --eval 'db.appsmiht-ce-github-dp-runs.update(
                                                                 # { _id:  ${{ github.event.inputs.uid }} },
                                                                 # {
                                                                  #  $set: {
                                                                 #     status: "DEPLOYMENT IN PROGRESS"
                                                                #    }
                                                                #  }
                                                            #    )'
```

sum35h: This was supposed to be commented and tried later. I've commented it now.

.github/workflows/ad-hoc-deploy-preview.yml
sum35h: I'm Modifying dependent jobs to checkout branch.

sum35h: Done.

sharat87 (approved):

PR #25427, opened 2023-07-18
chore: Add Interaction tests for hover + focus state

### LINKED ISSUE #25296: Visual Testing for Interactions ( hover, active )
We need to add interactions tests for hover, active so that we can catch all changes being made to our components.

jsartisan:
Fixes #25296


app/client/packages/design-system/widgets/src/components/Button/index.styled.tsx
```diff
@@ -11,11 +11,13 @@ export const buttonStyles = css<ButtonProps>`
...
 
-        &[data-hovered]:not([aria-disabled]) {
+        &[data-hovered]:not([aria-disabled]),
+        &:hover:not([aria-disabled]) {
           background-color: var(--color-bg-${color}-hover);
         }
 
-        &[data-active]:not([aria-disabled]) {
+        &[data-active]:not([aria-disabled]),
+        &:active:not([aria-disabled]) {
```
KelvinOm, line 20: 1. I believe that we should not modify components just for testing purposes.  I think we can sacrifice the active state and test only [hover and focus states](https://www.chromatic.com/docs/hoverfocus).
2. Also, I would not want the stories for testing to leak into the docs one, they either have to be invisible to the docs consumer or we have to separate them somehow.

KelvinOm (changes requested):

app/client/packages/design-system/widgets/src/components/Button/index.styled.tsx
jsartisan: I would disagree here. If a small change in our code helps us to avoid any regression in UI, I would do it.
Anyway, I have incorporated the changes you suggested.
Removed the code I added for the active state
Used interaction tests with the 'play' function for hover and active
Hide the story in docs table for hover and active ( as they are for testing purposes )

KelvinOm (changes requested):
We discussed the approach to the tests in direct messages, it is necessary to make changes.

Issue #25237, opened 2023-07-10
Create flex component


KelvinOm:
Create flex component and add docs about it to storybook

PR #25449, opened 2023-07-18
feat: add flex layout component

KelvinOm:
## Description
Add flex component.

#25237

#### Type of change

- [x] Manual
- [ ] Jest
- [ ] Cypress

#### Dev activity


ichik (approved):

marks0351 (approved):

app/client/packages/design-system/widgets/src/components/Flex/types.ts
```diff
@@ -0,0 +1,220 @@
...
+  /** Sets the CSS [className](https://developer.mozilla.org/en-US/docs/Web/API/Element/className) for the element. Only use as a **last resort**. Use style props instead. */
+  className?: string;
+  /** Sets inline [style](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/style) for the element. Only use as a **last resort**. Use style props instead. */
+  style?: CSSProperties;
+}
+
+export type StyledFlexProps = PickRename<
+  FlexProps,
+  {
+    direction: "$direction";
```
jsartisan, line 187: This can be little annoying for developers to add all the keys like this  🫠

How about a Utility type like this. It works other way round, we tell it to modify all keys except the ones passed.

```ts
export type PrependSymbol<
  TObj extends object,
  TOmitKeys extends keyof TObj,
  TSymbol extends string = "$",
> = {
  [K in keyof TObj as K extends TOmitKeys
    ? never
    : `${TSymbol}${string & K}`]: TObj[K];
} & Omit<TObj, TOmitKeys>;

export type StyledFlexProps = PrependSymbol<
  FlexProps,
  "style" | "className",
  "$"
>;
```

app/client/packages/design-system/widgets/src/components/Flex/types.ts
```diff
@@ -0,0 +1,220 @@
...
+  flex?: Responsive<string | number | boolean>;
+  /** When used in a flex layout, specifies how the element will grow to fit the space available. See [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-grow). */
+  flexGrow?: Responsive<number>;
+  /** When used in a flex layout, specifies how the element will shrink to fit the space available. See [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-shrink). */
+  flexShrink?: Responsive<number>;
+  /** When used in a flex layout, specifies the initial main size of the element. See [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-basis). */
+  flexBasis?: Responsive<SizingDimension>;
+  /** Specifies how the element is justified inside a flex or grid container. See [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-self). */
+  justifySelf?: Responsive<
+    | "auto"
```
jsartisan, line 98: instead of defining values here by ourselves, can we leverage CSSProperties type?

```ts
justifySelf?: Responsive<CSSProperties["justifySelf"]>;
```

app/client/packages/design-system/widgets/src/components/Flex/types.ts
```diff
@@ -0,0 +1,220 @@
...
+   * Whether to wrap items onto multiple lines. See [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-wrap).
+   * @default false
+   */
+  wrap?: Responsive<boolean | "wrap" | "nowrap" | "wrap-reverse">;
+  /**
+   * The distribution of space around items along the main axis. See [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content).
+   * @default 'stretch'
+   */
+  justifyContent?: Responsive<
+    | "start"
```
jsartisan, line 32: Why are we going with "start" instead of "flex-start"?🤔

app/client/packages/design-system/widgets/src/components/Flex/dimensions.ts
```diff
@@ -0,0 +1,117 @@
...
+  | "spacing-3"
+  | "spacing-4"
+  | "spacing-5"
+  | "spacing-6"
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  | (string & {})
+  | number;
+
+export type SizingDimension =
+  | "sizing-0"
```
jsartisan, line 14: I can imagine the frustration ( if this is done manually )  you had for typing this. I would something like this

```ts
type Range<
  TStart extends number,
  TEnd extends number,
  TAcc extends number[] = [],
> = TAcc["length"] extends TEnd
  ? TAcc
  : Range<TStart, TEnd, [...TAcc, TAcc["length"]]>;
  
type PrependSymbol<
  TString extends string | number,
  TSymbol extends string = "",
> = `${TSymbol}${TString}`;

export type SizingDimension =
  | PrependSymbolToArryItems<Range<0, 101>[number], "sizing-">
  | (string & {})
  | number;
```

Range - utitlity will create array of lenght based on Tstart and Tend
PrependSymbol - will append a symbol

app/client/packages/design-system/widgets/src/components/Flex/types.ts
KelvinOm: We don't want to support all CSS options, only the most useful ones. Look at what happens if you make such an interface for `flex-direction`.

Снимок экрана 2023-07-21 в 12 55 40

app/client/packages/design-system/widgets/src/components/Flex/types.ts
KelvinOm: It for consistency with `align-items`. Do you think it's not worth it?

app/client/packages/design-system/widgets/src/components/Flex/dimensions.ts
KelvinOm: This was generated based on tokens, there is quite a bit of manual work. I tried to keep it as simple as possible since
I'm not sure yet that these are the final values. Firstly, they are linear, and secondly, halves may appear here. For example, I see a 2px in the [switch PR](https://github.com/appsmithorg/appsmith/pull/25526/files#diff-39856913ad28d191cb690f6be8a7190fa3adda7c3484f7327574481cf268b64aR11). I think this should be covered by sizing too.

Although we can write it like this `calc(var(--sizing-2) / 2)`

cc @ichik

KelvinOm: @jsartisan I tried this type, my autocompilation stopped working in Webstorm. Left it as it is.

app/client/packages/design-system/widgets/src/components/Flex/types.ts
KelvinOm: Added and applied type `OmitRename`.

jsartisan (approved):

jsartisan (approved):

PR #25478, opened 2023-07-19
chore: Detect other NGINX and offer to kill

sharat87:
In the `start-https.sh` script, detect if some other process is listening on the ports we need, and if yes, inform of this, and offer to kill and proceed.

### DIFF
--- a/app/client/start-https.sh
+++ b/app/client/start-https.sh
@@ -245,8 +245,7 @@ $(if [[ $use_https == 1 ]]; then echo "
 
     server {
 $(if [[ $use_https == 1 ]]; then echo "
-        listen $https_listen_port ssl default_server;
-        http2 on;
+        listen $https_listen_port ssl http2 default_server;
         server_name $domain;
         ssl_certificate '$cert_file';
         ssl_certificate_key '$key_file';


hetunandu:
I see this error when running the code
```
nginx: [emerg] unknown directive "http2" in /Users/hetunandu/code/appsmith/app/client/nginx/nginx.dev.conf:41
````

**Comment by sharat87** [comment]
Thanks @hetunandu, I've removed that change. Looks like I'll need to setup an NGINX version block for that change. Can you try now please?

**Comment by hetunandu** [comment]
I did not see any prompt to kill older processes but the script worked in switching between two endpoints and the client is working without any errors

```
> ./start-https.sh                                       => worked for localhost
> ./start-https.sh https://release.app.appsmith.com      => worked for release cloud
> ./start-https.sh                                       => Switched back to localhost
```


Let me know if there is another way to test this

sharat87:
Thanks @hetunandu. There's already a built-in method to kill the running NGINX started by the same `start-https.sh`. That's why you don't see the prompt.

For example, if you run the `~/ce/start-https.sh`, it'll start NGINX, and if you run `~/ce/start-https.sh` again, it'll kill that previously started NGINX, and start anew. But if you run `~/ee/start-https.sh`, it won't know about that other NGINX, and attempt to start. This is where you should see the prompt, and the effect of this PR. Am I making sense? 🤔

hetunandu:
Tested the above mentioned scenario and it worked as expected.

PR #25479, opened 2023-07-19
chore: code splitting for multiple env feature

ayushpahwa:
code split for EE feature


app/client/src/ee/actions/environmentAction.ts
```diff
@@ -0,0 +1,2 @@
+// Redux action to show the environment info modal before deploy
+export const showEnvironmentDeployInfoModal = () => ({});
```
ChandanBalajiBP, line 2: IMO we don't need this in ee folder

app/client/cypress/locators/MultipleEnvLocators.js
```diff
@@ -0,0 +1,8 @@
+export default {
+  env_switcher: '[data-cy="t--switch-env"]',
+  env_switcher_dropdown_opt_prod: '[data-cy="t--switch-env-dropdown-option-production"]',
+  env_switcher_dropdown_opt_stage: '[data-cy="t--switch-env-dropdown-option-staging"]',
+  ds_data_filter_disabled: '[data-cy="t--filter-disabled"]',
+  env_switcher_dropdown_opt: (envName) => `[data-cy="t--switch-env-dropdown-option-${envName}"]`,
+  ds_editor_env_filter: (envName) => `[data-cy="t--ds-data-filter-${envName}"]`,
+}
```
Parthvi12, line 8: @ayushpahwa Can we move these locators to EE under pages folder, where you have added common methods for ME?

app/client/cypress/support/Pages/DataSources.ts
```diff
@@ -403,96 +416,121 @@ export class DataSources {
   }
 
   public FillPostgresDSForm(
+    environment = this.tedTestConfig.defaultEnviorment,
```
Parthvi12, line 419: @ayushpahwa Can we add this variable to the end since its defaulting to production, so that we don't have to change this method in CE tests

app/client/cypress/support/Pages/DataSources.ts
```diff
@@ -403,96 +416,121 @@ export class DataSources {
...
+      .type(
+        password == ""
+          ? this.tedTestConfig.dsValues[environment].postgres_password
+          : password,
+      );
     this.ValidateNSelectDropdown("SSL mode", "Default");
   }
 
   public FillOracleDSForm(
+    environment = this.tedTestConfig.defaultEnviorment,
```
Parthvi12, line 454: @ayushpahwa Same as above, please update at all places

app/client/src/ce/selectors/featureFlagsSelectors.ts
```diff
@@ -15,3 +16,10 @@ export const selectFeatureFlagCheck = (
   }
   return false;
 };
+
+export const datasourceEnvEnabled = createSelector(
```
AmanAgarwal041, line 20: There is already a selector function `selectFeatureFlagCheck`. Any reason we created another specific to a flag ?

AmanAgarwal041 (dismissed):

app/client/cypress/locators/MultipleEnvLocators.js
ayushpahwa: Removed

app/client/src/ce/selectors/featureFlagsSelectors.ts
ayushpahwa: We created this for ease of testing. We had instances when we had to remove feature flag due to flagsmith. Also, since this is phasing out in less than a month, would be easier to remove if we use this function as base.

AmanAgarwal041: But in any case wherever you will be using this function, you will have to update those components as well. So why not use the already existing function. Each feature flag will be removed after some time. 

> We had instances when we had to remove feature flag due to flagsmith.
Didn't understand this

app/client/src/sagas/DatasourcesSagas.ts
```diff
@@ -202,6 +210,19 @@ function* fetchDatasourceStructureOnLoad() {
   } catch (error) {}
 }
 
+export function* fetchDatasourceStructureOnSoftRefresh() {
+  try {
+    // get datasources of all actions used in the the application
+    const datasourcesUsedInApplication: Datasource[] = yield select(
+      getDatasources,
+    );
+    for (const datasource of datasourcesUsedInApplication) {
```
Irongade, line 219: This should be forked!! This is a potential performance problem. Also, getDatasources only fetches all the datasources in the entire workspace!

Irongade: Also, why not simply trigger handleFetchDatasourceStructureOnLoad when SOFT_REFRESH_ACTION is dispatched

app/client/cypress/support/Pages/DataSources.ts
```diff
@@ -403,96 +403,121 @@ export class DataSources {
...
     cy.get(this._databaseName).clear().type(databaseName);
-    cy.get(this._username).type(
-      username == "" ? this.tedTestConfig.postgres_username : username,
-    );
-    cy.get(this._password).type(
-      password == "" ? this.tedTestConfig.postgres_username : password,
-    );
+    cy.get(this._username)
+      .clear()
+      .type(
```
Parthvi12, line 425: @ayushpahwa Why are we using .clear() here?

Parthvi12 (dismissed):

app/client/cypress/support/Pages/DataSources.ts
ayushpahwa: Hey, seems like it got slipped in. Fixed now

Parthvi12 (dismissed):

Parthvi12:
Adding test plan approved label for cypress changes.

AmanAgarwal041 (approved):

PR #25515, opened 2023-07-20
ci: Added common CI check for server and client formatting

nidhi-nair:
Combined both server and client side checks into one workflow, conditionally triggered based on path of files changed in the PR. This is so we can have one compulsory check for all kinds of formatting in the repo.


.github/workflows/client-prettier.yml
```diff
@@ -1,66 +1,33 @@
-name: Client Prettier
+name: Appsmith Client Prettier Check Workflow
```
sharat87, line 1: Can we please not prefix everything with `Appsmith`? It's already bad enough. 😬

shot-2023-07-20-13-07-19

.github/workflows/server-build.yml
```diff
@@ -110,23 +110,6 @@ jobs:
...
-        if: steps.run_result.outputs.run_result != 'success'
-        id: vars
-        run: |
-          # Since this is an unreleased build, we set the version to incremented version number with a
-          # `-SNAPSHOT` suffix.
-          latest_released_version="$(git tag --list 'v*' --sort=-version:refname | head -1)"
-          echo "latest_released_version = $latest_released_version"
-          next_version="$(echo "$latest_released_version" | awk -F. -v OFS=. '{ $NF++; print }')"
-          echo "next_version = $next_version"
-          echo version=$next_version-SNAPSHOT >> $GITHUB_OUTPUT
```
sharat87, line 127: This variable is used in the next step. Are we sure we want to remove this?

.github/workflows/server-spotless.yml
```diff
@@ -0,0 +1,36 @@
+# This workflow is responsible for running Spotless check on server code base
+name: Appsmith Server Spotless Check Workflow
```
sharat87, line 2: Minor.

```suggestion
name: Server Spotless Check
```

.github/workflows/formatting-checks.yml
```diff
@@ -0,0 +1,64 @@
+name: Perform formatting checks
```
sharat87, line 1:
```suggestion
name: Formatting checks
```

.github/workflows/client-prettier.yml
nidhi-nair: Agreed, felt awkward writing it as well :P

.github/workflows/server-build.yml
nidhi-nair: Oh that's true, I only checked for tag. Thanks for catching this, I'll revert the entire block

.github/workflows/formatting-checks.yml
```diff
@@ -0,0 +1,64 @@
+name: Formatting checks
+
+on:
+ pull_request:
```
sharat87, line 4:
```suggestion
  pull_request:
```

sharat87 (approved):

PR #25554, opened 2023-07-21
DONOT-MERGE: widget registration dependencies

sbalaji1192:
> Pull Request Template
>
>
Fixes # (issue number)
>
>
>
>
>
- [ ] Manual
- [ ] Jest
- [ ] Cypress
>
>
#### Dev activity

#### QA activity:

### DIFF
--- a/app/client/src/utils/WidgetRegisterHelpers.tsx
+++ b/app/client/src/utils/WidgetRegisterHelpers.tsx
@@ -7,8 +7,11 @@ import type BaseWidget from "widgets/BaseWidget";
 import WidgetFactory, { NonSerialisableWidgetConfigs } from "./WidgetFactory";
 
 import { ReduxActionTypes } from "@appsmith/constants/ReduxActionConstants";
-import { memoize } from "lodash";
-import type { WidgetConfiguration } from "widgets/constants";
+import { isArray, isFunction, memoize } from "lodash";
+import type {
+  WidgetConfiguration,
+  WidgetConfigurationGetter,
+} from "widgets/constants";
 import withMeta from "widgets/MetaHOC";
 import withWidgetProps from "widgets/withWidgetProps";
 import { generateReactKey } from "./generators";
@@ -40,8 +43,26 @@ const generateWidget = memoize(function getWidgetComponent(
 
 export const registerWidget = (
   Widget: typeof BaseWidget,
-  config: WidgetConfiguration,
+  config: WidgetConfiguration | WidgetConfigurationGetter,
 ) => {
+  if (isArray(config)) {
+    const [configGetter, dependencies] = config;
+
+    config = configGetter.apply(
+      null,
+      dependencies.map((dependency) => {
+        switch (dependency) {
+          case WidgetRegistrationDependency.PLATFORM_CONFIGS:
+            return { platform: true };
+          case WidgetRegistrationDependency.FEATURE_FLAGS:
+            return { features: true };
+          default:
+            return {};
+        }
+      }),
+    );
+  }
+
   const ProfiledWidget = generateWidget(
     Widget,
     !!config.needsMeta,
@@ -128,3 +149,8 @@ export const configureWidget = (config: WidgetConfiguration) => {
     payload: _config,
   });
 };
+
+export enum WidgetRegistrationDependency {
+  PLATFORM_CONFIGS = "PLATFORM_CONFIGS",
+  FEATURE_FLAGS = "FEATURE_FLAGS",
+}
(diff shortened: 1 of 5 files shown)


app/client/src/utils/WidgetRegisterHelpers.tsx
```diff
@@ -40,8 +43,26 @@ const generateWidget = memoize(function getWidgetComponent(
...
+  if (isArray(config)) {
+    const [configGetter, dependencies] = config;
+
+    config = configGetter.apply(
+      null,
+      dependencies.map((dependency) => {
+        switch (dependency) {
+          case WidgetRegistrationDependency.PLATFORM_CONFIGS:
+            return { platform: true };
+          case WidgetRegistrationDependency.FEATURE_FLAGS:
```
riodeuno, line 57: @sbalaji1192 Two concerns here:

1. We have the `features` property we use for `AutoHeight` is widget configs. Will this conflict?

2. In general should these platform configs be available to all Widgets? If I would like to create a new widget, should I have to add the `WidgetRegistrationDependency.PLATFORM_CONFIGS`, or should I be able to use them as we can with platform provided constants and generic components?

sbalaji1192: 1. It shouldn't have any conflicts with the feature from the widget development API. We will only inject the feature flags that are available today from the Fagsmith API.  
2. It will be only available to widgets that are explicitly asking for it. Just to simply the configuration object for the widgets that don't need it. Ideally, we should be injecting the constants from the platform in this way. Widgets should not import anything from outside the widgets folder and refer to any global values directly.

app/client/src/utils/WidgetRegistry.tsx
```diff
@@ -238,7 +244,7 @@ export const ALL_WIDGETS_AND_CONFIG: [any, WidgetConfiguration][] = [
 export const registerWidgets = () => {
   const start = performance.now();
   for (const widget of ALL_WIDGETS_AND_CONFIG) {
-    registerWidget(widget[0], widget[1] as WidgetConfiguration);
+    registerWidget(widget[0], widget[1]);
```
rajatagrawal, line 247: Alternate approach:

1. Fetch server configuration for a widget here and override the default widget config fetched from widget/index.ts file in this method.
2. Might also be helpful in this scenario below:
  - If someone develops a third party widget that we want to use in Appsmith for example, we can use this approach to modify its config. The third party widget code could be defined in a separate github repository. We can use registerWidget's caller function to override the default widget config supplied by the third party widget developer with our required configurations and use the widget.

3. Pros:
  - Keeps the widget interface clean and simple.
  - Helps to keep the widget unit testable with any widget configuration and simple JSON Objects.
  - Any platform where a widget is used (Appsmith, Jest Test, or some other third party canvas app in futre) can choose to override a default widget configuration if it needs to by simply override particular keys of the config. Making the widget config switchable between an object and function makes it feel clunky to use.

rajatagrawal: @sbalaji1192 @riodeuno

sbalaji1192: @rajatagrawal  
The approach you suggested is not scalable. We will be bloating the registerWidget method with different configuration overrides required for different widgets. Let's say `max file size` config variable is used by 10 widgets, we would have to know exactly how these widgets use this variable to know how to override it in the register widget method.
Regarding the third-party widget, As of today, we cannot use widgets that are not built for Appsmith inside Appsmith, since the widgets need to adhere to the widget development API. 

> The widget interface shouldn't define how these configurations are computed in different platforms or the configuration computation shouldn't be tightly coupled to a particular platform.

I don't agree with this statement fully. Any platform that supports extensions out there, dictates how to write an extension, they have a certain API and only through that you can define the extensions. Likewise, Appsmith already dictates how a widget needs to be defined and we are calling it Widget development API. What we should strive for is to keep the API as simple as possible and make things optional. ex. A widget can be created without property pane configuration.

>Keeps the widget interface clean and simple.

Can you please elaborate on why do you think implementing the registration dependency the way we did in this PR, makes things complex and messy?

>Helps to keep the widget unit testable with any widget configuration and simple JSON Objects.

Please elaborate on this one as well.

Pros of current implementation:
1. Optional, WDI doesn't enforce this new configuration. it's a sophisticated registration mode where widgets can ask for things from the platform to make it more extensible.
2. maintainable, We can keep adding registration dependencies without affecting the existing configurations.
3. Abstraction, Platform doesn't need to peak into widgets to make platform-specific changes.

PR #25555, opened 2023-07-21
chore: update tests to include ts definitions

ayushpahwa:
Made tests ts specific


app/client/cypress/support/Pages/HomePage.ts
```diff
@@ -331,6 +336,28 @@ export class HomePage {
...
+        cy.get(this.dropdownOption).click();
+        cy.get(this.useCaseDropdown).click();
+        cy.get(this.dropdownOption).click();
+        cy.get(this.roleUsecaseSubmit).click({ force: true });
+      }
+    });
+    cy.wait("@getMe");
+    this.agHelper.Sleep(3000);
+  }
+
```
Parthvi12, line 360: @ayushpahwa Can you use aggregate helper methods here, instead of .click() .type()

Parthvi12 (changes requested):

Parthvi12 (dismissed):

PR #25582, opened 2023-07-21
test: Cypress - Helpers update

sharanya-appsmith:
Added helper method: 

1.  to validate tooltip text
2. to assert if the class name contains "disabled"
3. dragNDrop to public 
4. to search and open an application


app/client/cypress/support/Pages/AggregateHelper.ts
```diff
@@ -287,6 +287,22 @@ export class AggregateHelper extends ReusableHelper {
     });
   }
 
+  public ValiadateToolTipText(textToValidate: string) {
```
Aishwarya-U-R, line 290: Can this be - In AssertHelper file as below?
```
public AssertTooltip(textToValidate: string){
    agHelper.GetNAssertContains('.rc-tooltip-inner' //create a locator for this, textToValidate)
}
```

Aishwarya-U-R: Why is if condition needed here?

app/client/cypress/support/Pages/HomePage.ts
```diff
@@ -336,6 +336,16 @@ export class HomePage {
     cy.xpath(this.locator._spanButton("Share")).first().should("be.visible");
   }
 
+  /**
+   * Searches for given app name and clicks edit icon
+   * @param appName
+   */
+  public searchAndOpenApplication(appName: string) {
```
Aishwarya-U-R, line 343: `SearchAndOpenApp`

app/client/cypress/support/Pages/AggregateHelper.ts
sharanya-appsmith: 1. other tooltip methods are in aggregate helper, hence added here. I can move this. 
2. sometimes after hover I couldn't see the tooltip immediately. I have added other asserts in the corresponding test. hence added if, just incase, didn't want the case to file because of this. Will remove it based on the suggested practice.

app/client/cypress/support/Pages/AssertHelper.ts
```diff
@@ -1,6 +1,7 @@
 import "cypress-wait-until";
 import { ObjectsRegistry } from "../Objects/Registry";
 import { ReusableHelper } from "../Objects/ReusableHelper";
+import { agHelper } from "../../support/Objects/ObjectsCore";
```
Aishwarya-U-R, line 4: This cant be imported here. AgHelper is already using AssertHelper. This will end in loop.

Aishwarya-U-R (approved):
Approving changes. Can merge after a CI run - just to be sure.

PR #25602, opened 2023-07-21
fix: Added embedded check for true env

nidhi-nair:
Changes to start to consider embedded datasources as oos in env


sondermanish (approved):
LGTM

sondermanish (approved):

PR #25667, opened 2023-07-25
ci: Moved QC around, removed unit tests running with ok to test

nidhi-nair:
QC now has all server and client side formatting and unit tests check. This would be a compulsory check for merge.
ok-to-test hence does not needs unit tests to run. This should reduce run times by half an hour.
TBD also now does not need Cypress to run at all before package. Since package never required tests to pass on release, this change simply reduces the time between pushing to release and the changes reflecting on release.app.appsmith.com 

A few other clean ups across the board.

### DIFF
--- a/.github/workflows/client-build.yml
+++ b/.github/workflows/client-build.yml
@@ -24,7 +24,7 @@ defaults:
     working-directory: app/client
 
 jobs:
-  build:
+  client-build:
     runs-on: ubuntu-latest-8-cores
     # Only run this workflow for internally triggered events
     if: |
--- a/.github/workflows/client-build.yml
+++ b/.github/workflows/client-build.yml
@@ -1,8 +1,6 @@
 name: Client Build
 
 on:
-  # This line enables manual triggering of this workflow.
-  workflow_dispatch:
   workflow_call:
     inputs:
       pr:
@@ -19,12 +17,6 @@ on:
         required: false
         type: string
         default: "false"
-
-  pull_request:
-    branches: [release, master]
-    paths:
-      - "app/client/**"
-      - "!app/client/cypress/manual_TestSuite/**"
 
 # Change the working directory for all the jobs in this workflow
 defaults:
--- a/.github/workflows/client-unit-tests.yml
+++ b/.github/workflows/client-unit-tests.yml
@@ -1,8 +1,6 @@
 name: Client Unit Tests
 
 on:
-  # This line enables manual triggering of this workflow.
-  workflow_dispatch:
   workflow_call:
     inputs:
       pr:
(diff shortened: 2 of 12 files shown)


nidhi-nair:
Cancelled ok-to-test cuz it looks like that one needs the workflow to be pushed into release first.

.github/workflows/client-build.yml
```diff
@@ -20,12 +18,6 @@ on:
         type: string
         default: "false"
 
-  pull_request:
```
KelvinOm, line 23: Why is it removed? We need to run the client build to check the correctness of the build.

.github/workflows/client-unit-tests.yml
```diff
@@ -15,26 +13,14 @@ on:
...
-      - "!app/client/cypress/manual_TestSuite/**"
-
 # Change the working directory for all the jobs in this workflow
 defaults:
   run:
     working-directory: app/client
 
 jobs:
-  build:
+  client-build-and-tests:
```
KelvinOm, line 22: Here we are not making a client build, we are just running tests on the code base. Let's call it just `client-unit-tests` or `client-jest-tests`

.github/workflows/quality-checks.yml
```diff
@@ -35,6 +35,15 @@ jobs:
...
+    name: server-unit-tests
+    needs: path-filter
+    if: needs.path-filter.outputs.server == 'true'
+    uses: ./.github/workflows/server-build.yml
+    secrets: inherit
+    with:
+      pr: ${{ github.event.pull_request.number }}
+
   client-formatting:
     name: client-formatting
```
KelvinOm, line 48: (nitpic) I would call it `client-prettier`, so it's more obvious what's going on inside. The same is true for server(`server-spotless`).

.github/workflows/quality-checks.yml
```diff
@@ -44,9 +53,27 @@ jobs:
...
+    if: needs.path-filter.outputs.client == 'true'
+    uses: ./.github/workflows/client-unit-tests.yml
+    secrets: inherit
+    with:
+      pr: ${{ github.event.pull_request.number }}
+
+  client-lint:
+    name: client-lint
+    needs: path-filter
+    if: needs.path-filter.outputs.client == 'true'
```
KelvinOm, line 68: Does it match to this paths? Where can I find it?

```
- "app/client/**"	
- "!app/client/cypress/manual_TestSuite/**"
```

.github/workflows/client-build.yml
nidhi-nair: This gets run as part of client-unit-tests, correct?

KelvinOm (approved):

PR #25698, opened 2023-07-25
ci: Modified cypress check

nidhi-nair:
Modified the check to allow for Cypress paths to be excluded properly


.github/workflows/quality-checks.yml
```diff
@@ -25,8 +25,8 @@ jobs:
               - 'app/server/**'
             client:
               - 'app/client/**'
-            cypress:
-              - 'app/client/cypress/manual_TestSuite/**'
+            not-cypress:
```
KelvinOm, line 28: Lets call it `not-cypress-manual`, because it's only for one specific folder.

KelvinOm (approved):

PR #25711, opened 2023-07-26
chore: Add property path to WidgetErrors

### LINKED ISSUE #25712: [Task]: WidgetErrors should also include the property path in each each error

### SubTasks

The widget may need this information to process it when it receives widget errors in the component. One use case is that the widget may need to discard certain widget errors if the property path is not matching certain criteria.

One use case I have encountered in ChartWidget is as below:

1. Introduce a syntax error in custom fusion chart.
2. Change chart type to Line chart
3. The widget should not show errors because the line chart configuration doesn't have error. The syntax errors are present in "customFusionChartConfig" property. Line chart configuration is invalid when there are errors in "chartData" property not in "customFusionChartConfig" property.
4. The widget should be able to filter out these errors using the property path.

rajatagrawal:
Fixes #25712

### DIFF
--- a/app/client/src/utils/widgetRenderUtils.tsx
+++ b/app/client/src/utils/widgetRenderUtils.tsx
@@ -65,19 +65,24 @@ function widgetErrorsFromStaticProps(props: Record<string, unknown>) {
     string,
     DataTreeError[]
   >;
-  const evaluationErrors: DataTreeError[] =
-    Object.values(evaluationErrorMap).flat();
   const widgetErrors: WidgetError[] = [];
-  for (const evalError of evaluationErrors) {
-    const widgetError: WidgetError = {
-      name: evalError.errorMessage.name,
-      message: evalError.errorMessage.message,
-      stack: evalError.raw,
-      type: "property",
-    };
 
-    widgetErrors.push(widgetError);
+  for (const propertyPath in evaluationErrorMap) {
+    const propertyErrors = evaluationErrorMap[propertyPath];
+
+    for (const evalError of propertyErrors) {
+      const widgetError: WidgetError = {
+        name: evalError.errorMessage.name,
+        message: evalError.errorMessage.message,
+        stack: evalError.raw,
+        type: "property",
+        propertyPath: propertyPath,
+      };
+
+      widgetErrors.push(widgetError);
+    }
   }
+
   return widgetErrors;
 }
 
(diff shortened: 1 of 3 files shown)


app/client/src/utils/widgetRenderUtils.tsx
```diff
@@ -65,19 +65,24 @@ function widgetErrorsFromStaticProps(props: Record<string, unknown>) {
...
-    const widgetError: WidgetError = {
-      name: evalError.errorMessage.name,
-      message: evalError.errorMessage.message,
-      stack: evalError.raw,
-      type: "property",
-    };
-
-    widgetErrors.push(widgetError);
+
+  for (const propertyPath in evaluationErrorMap) {
```
sbalaji1192, line 70: nit: if we are not breaking anywhere, I would prefer object entries and forEach. More readable.

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -871,6 +871,7 @@ export const WIDGET_DISPLAY_PROPS = {
 };
 export interface WidgetError extends Error {
   type: "property" | "configuration" | "other";
+  propertyPath?: string;
```
sbalaji1192, line 874: Can we just call this path, since type can be configuration and other. `propertyPath` seems specific to the property type.

rajatagrawal: Done

app/client/src/utils/widgetRenderUtils.tsx
rajatagrawal: Done

app/client/src/utils/widgetRenderUtils.tsx
```diff
@@ -65,19 +65,24 @@ function widgetErrorsFromStaticProps(props: Record<string, unknown>) {
...
-      message: evalError.errorMessage.message,
-      stack: evalError.raw,
-      type: "property",
-    };
-
-    widgetErrors.push(widgetError);
-  }
+
+  Object.keys(evaluationErrorMap).forEach((propertyPath) => {
+    const propertyErrors = evaluationErrorMap[propertyPath];
```
sbalaji1192, line 71: nit: Object entries is more suitable than object keys.

sbalaji1192 (approved):

app/client/src/utils/widgetRenderUtils.tsx
rajatagrawal: Yes, I can make this change. I just noticed that propertyPath is required inside the loop.

sbalaji1192: Object entries will give you both key and value

PR #25729, opened 2023-07-26
chore: Added check for non-existent storage while importing and exporting

### LINKED ISSUE #25715: [Bug]: Committing to app fails when there is only one environment  configured and query is created on different env

### Description

When an app is connected to git, and there is only staging configured. On creating query on prod and comitting fails with following error : 
Screenshot 2023-07-26 at 10 11 54 AM

### Steps To Reproduce

1. Connect an app to git
2. Create only staging env for any DS e.g. MySQL
3. Create new query on prod for this datasource
4. Switch to staging and create one more query and run
5. Refresh the app and observe internal server error is thrown
6. Try committing changes it gives error

### Public Sample App

_No response_

### Environment

Deploy Preview

### Issue video log

_No response_

### Version

Deploy preview

nidhi-nair:
Fixes #25715


sondermanish (approved):

AnaghHegde:
@nidhi-nair Can you add a test case for this flow?

PR #25785, opened 2023-07-27
fix: config checks for generate page CRUD

### LINKED ISSUE #25640: [Bug]: Generate Page With Data -> GSheets -> Unable to fetch DS structure when on Staging

### Description

When I Generate Page With Data -> a GSheets DS, while I am in Staging environment, I get an error - Unable to fetch DS structure.

### Steps To Reproduce

1. Create a Gsheets DS. Save and Authorize.
2. Ensure the environment dropdown is Production
3. Now choose Pages -> Generate Page with Data
4. Using above DS, try to generate a page with data, with some sheet selected. It goes through fine
5. Switch between prod and staging environments, and note that the Data is refreshed, but same for both, as for OOS plugins, Prod details are copied over to Staging
6. Now Keep the environment at Staging
7. Now choose Pages -> Generate Page with Data
8. Note that Spreadsheet list does not get populated and fails with an error "Failed Fetching DS structure, Please check your DS config"
The user should be able to select the sheet, and there should be no error as Prod details are ideally copied to Staging

https://github.com/appsmithorg/appsmith/assets/101863839/bfe7abb0-0796-4871-a81d-353d3c45e07e

### Public Sample App

_No response_

### Environment

Release

### Issue video log

_No response_

### Version

Cloud

ayushpahwa:
## Description
Generate page crud was checking for the config from DS storage even for non env enabled datasources. This PR fixes that.

Fixes #25640

### DIFF
--- a/app/client/src/ce/utils/Environments/index.tsx
+++ b/app/client/src/ce/utils/Environments/index.tsx
@@ -53,7 +53,7 @@ export const isEnvironmentConfigured = (
   return !!isConfigured ? isConfigured : false;
 };
 
-// function to check if the datasource is configured for the current environment
+// function to check if the datasource is configured for any environment
 export const doesAnyDsConfigExist = (
   datasource: Datasource | null,
   environment?: string,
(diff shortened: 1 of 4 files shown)


ChandanBalajiBP (approved):

ayushpahwa:
Addint skip-testPlan since the change was earlier tested on EE (didn't go in due to code split). Also, this section only refers to changes in EE and hence not testable here.

PR #25829, opened 2023-07-31
ci: Use cache from release image to build faster

sharat87:
This should enable using unchanged layers from the `release` image, as a cache, and build PR images faster. We only do this for images built for PRs and not for direct `release` or `master` branches.


sarojsarab (dismissed):
LGTM

Aishwarya-U-R (approved):
👏🏻

sarojsarab (approved):

PR #25838, opened 2023-07-31
fix: Fix missing cloud services URL when env variable is empty

sharat87:
When `APPSMITH_CLOUD_SERVICES_BASE_URL` is an empty string, Spring's interpolation in `application.properties` doesn't take the default value. This PR sets up the `entrypoint.sh` script so that it'll `unset` this env variable, when it's empty.


sharat87:
Merged by force, since the release image is broken without this. Tested locally that the fix works.

Issue #25307, opened 2023-07-11
[Task]: Update the widget_resize_end event to contain the details of the table's scrollbar and pagesize


keyurparalkar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

Issue #17857, opened 2022-10-25
[Bug]: Table shows a scrollbar when trying to fit all items in table window


dilippitchika:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

When i try to adjust the height of table to make sure no more rows are present, the table still ends up showing an extra row all the time.

### Steps To Reproduce

https://www.loom.com/share/6141092d38ef4be09ce24ed85fd4d8f6

### Version

Release

keyurparalkar:
@dilippitchika @sbalaji1192 I think rather than solving this issue via JS, we should solve it via another approach. 

I have analysed this issue and found out that this issue happens whenever the table's content exceeds the height of the parent container. 

To solve this we should:
1. When we remove the `ROW_VIRTUAL_OFFSET` from the size parameter of `FixedSizeList`
2. Set the table's header to be 30px 
3. Set the parent container's height with an offset. With this offset in place the height of the parent container remains greater than the table's content thus helping to keep table with no vertical scroll bars when rows are completely fit.

dilippitchika:
Internal comment

Requirement - @dilippitchika to rephrase

Context of table - Table row height is default, short or tall we will stick to this logic
1. 10 rows that perfectly fit without scrollbar and you have only 20% space remaining to show 11th row, then we only show 10 rows
2. 10 rows that perfectly fit without scrollbar and you have only 35% space remaining to show 11th row, then we only show 11 rows. WE WILL CONTINUE TO SEE SCROLLBAR
3. When 30% of space remains - 11th row is shown. WE WILL CONTINUE TO SEE SCROLLBAR. 

This doesn't apply to rows with images and wrapping, we will only follow the page size variable to determine the logic

Suggestions
1. Change this to task or feature
2. Hypothesis - Our current hypothesis is just to fit for scenarios where it's expected to not exceed.
3. New hypothesis - We will show blank space and let the 11th row only come when 100% space is available for 11th row.

NOT DOING
5. New new hypothesis - We have a max row height and min row height and the height of a row is variable. Virtualisation has to be updated

style is default - 40px
min row height - 36px
max row height - 44 px

10 rows fit and 11th row fits 50% - 20px
20px/10 = 2px per row - each row will have 42px now

Impact - Event - table resize exceed -> if there's a scrollbar on table and page size
Laveena is using this feature
1. 11 rows coming but space is only there for 10 - scrollbar true, pagesize 10
[...]

PR #25840, opened 2023-07-31
chore:  create a new table_vertical_scroll event to contain the details of the table's scrollbar and pagesize

gitstart-appsmith:
## Description
It updates the `TABLE_VERTICAL_SCROLL ` event to contain the details of the table's scrollbar and pagesize

### Step To Reproduce
1. Drop a Table widget
2. Resize the widget
3. Peek the console to check for the `TABLE_VERTICAL_SCROLL` log and the properties

Fixes https://github.com/appsmithorg/appsmith/issues/25307

## Type of change
* New feature (non-breaking change which adds functionality)

## How Has This Been Tested?
https://www.loom.com/share/0ecb378095ce41caa990832307b89d40?sid=893a752c-f6ef-4c7f-aa66-808abb98921b

## Checklist

- [x]  I have commented my code, particularly in hard-to-understand areas

___
 
 This code was written and reviewed by GitStart Community. Growing future engineers, one PR at a time.


app/client/src/widgets/TableWidgetV2/component/Table.tsx
```diff
@@ -163,6 +166,9 @@ export type HeaderComponentProps = {
 const emptyArr: any = [];
 
 export function Table(props: TableProps) {
+  const isTableResizing = useSelector(
```
keyurparalkar, line 169: @gitstart-appsmith usage of redux state directly inside the widget is an abstraction leak. All the props for a widget are provided by the platform via BaseWidget.

Can you please remove this line?

app/client/src/widgets/TableWidgetV2/component/Table.tsx
```diff
@@ -244,6 +260,9 @@ export function Table(props: TableProps) {
       }, 0);
     }
   }
+  if (!isTableResizing) {
```
keyurparalkar, line 263: Instead of `if` block, execute the `logEvent` function of `AnalyticsUtils` inside a `useEffect` here with dependency array containing `bottomRow`, `topRow`, `leftColumn` and `rightColumn` props.

The props mentioned above are available as a prop to the widget. 

This approach will make sure that the event gets fired whenever the widget is resized

app/client/src/widgets/TableWidgetV2/widget/index.tsx
```diff
@@ -1335,6 +1336,12 @@ class TableWidgetV2 extends BaseWidget<TableWidgetProps, WidgetState> {
     }
   };
 
+  handleVerticalScrollVisibility = (isVisible: boolean) => {
```
keyurparalkar, line 1339: This is not required since we are calling the analytics function inside the Table component

app/client/src/widgets/TableWidgetV2/component/index.tsx
```diff
@@ -291,6 +294,8 @@ export default React.memo(ReactTableComponent, (prev, next) => {
     prev.filters === next.filters &&
     prev.handleReorderColumn === next.handleReorderColumn &&
     prev.handleResizeColumn === next.handleResizeColumn &&
+    prev.handleVerticalScrollVisibility ===
+      next.handleVerticalScrollVisibility &&
```
keyurparalkar, line 298: This is not required since we are calling the analytics function inside the Table component

keyurparalkar (commented):
@gitstart-appsmith to avoid firing WIDGET_RESIZE_END event twice, can you please execute the `logEvent` function mentioned in the `useEffect` below with a new event name: `TABLE_VERTICAL_SCROLL`.

app/client/src/widgets/TableWidgetV2/component/Table.tsx
gitstart-appsmith: @keyurparalkar, the `TableProps` doesn't extend the `WidgetProps` which makes `bottomRow, topRow, leftColumn and rightColumn` not available. Should we make it to extend it?

keyurparalkar: @gitstart-appsmith yes, let us consume these props via widget.

app/client/src/widgets/TableWidgetV2/component/Table.tsx
```diff
@@ -315,7 +328,20 @@ export function Table(props: TableProps) {
         }
       });
     }
-  }, [props.isAddRowInProgress]);
+
+    AnalyticsUtil.logEvent("TABLE_VERTICAL_SCROLL", {
```
keyurparalkar, line 332: @gitstart-appsmith Can you please move this logic into another useEffect along with its dependencies?
This effect should be below the existing useeffect.

LuisChiej: Done

gitstart-appsmith: Done

keyurparalkar (dismissed):
Can you please run the `ok-to-test` command in the comment of this PR to start the CI checks and tests?

gitstart-appsmith:
ok-to-test

keyurparalkar:
@chandannkumar can you please test the conditions mentioned in this PR for `TABLE_VERTICAL_SCROLL` event rather than `WIDGET_RESIZE_END`?

keyurparalkar:
@gitstart-appsmith can you please update the description of the PR to include TABLE_VERTICAL_SCROLL instead of the mentioned event?

Aishwarya-U-R:
@keyurparalkar - Merge latest release & trigger new oktotest. Cancelling the re-run since its past 12 hours from initial attempt.

chandannkumar:
@gitstart-appsmith Observations:
1. The logic of showing `pageSize` & `yScrollVisibility` is not correct as we can observe wrong results most of the time.
video: https://www.loom.com/share/e9aeaac1c9504a24845c96803367504d?sid=22ea1ef5-4fda-421f-993f-510530754dc4

2. When table is re-sized to 0 (only table header are visible), the `pageSize` on the event is still 1
image

3. Also observed that, fixing the size of Table without a scroll is not smooth. As we cannot fix the table page size with exact pixel manually.
Refer comment - https://github.com/appsmithorg/appsmith/issues/17857#issuecomment-1630762915

app/client/src/widgets/TableWidgetV2/component/Table.tsx
```diff
@@ -307,6 +312,14 @@ export function Table(props: TableProps) {
       (column) => !!column.columnProperties.allowCellWrapping,
     );
 
+  const isVerticalScrollVisible = () => {
```
keyurparalkar, line 315: Please memoize this callback

app/client/src/widgets/TableWidgetV2/component/Table.tsx
```diff
@@ -312,13 +312,13 @@ export function Table(props: TableProps) {
       (column) => !!column.columnProperties.allowCellWrapping,
     );
 
-  const isVerticalScrollVisible = () => {
+  const isVerticalScrollVisible = useMemo(() => {
```
keyurparalkar, line 315: @gitstart-appsmith let us move this block inside the useEffect of AnalyticsUtil. In this way, we won't need the `isVerticalScrollVisible` value to be dependency for the below useEffect's.

keyurparalkar (changes requested):

app/client/src/widgets/TableWidgetV2/component/Table.tsx
```diff
@@ -331,18 +323,25 @@ export function Table(props: TableProps) {
   }, [props.isAddRowInProgress]);
 
   useEffect(() => {
+    const isVerticalScrollVisible = () => {
```
keyurparalkar, line 326: @gitstart-appsmith instead of wrapping this logic inside a function can we just store result of this logic in a variable like below:
```js
let yScrollVisibility = false;
if(condition...)
``` 
and then use it in the below event

keyurparalkar (changes requested):

gitstart-appsmith:
> @gitstart-appsmith avoid merging main directly to the branch. Instead merge the release branch

@keyurparalkar, it's just a comment. We didn't merge main into the branch

keyurparalkar (approved):

keyurparalkar:
@chandannkumar can you please re-test this PR and verify the previous callouts.

gitstart-appsmith:
Hello @chandannkumar 

> Event is not getting fired when if widget is resized in the fixed pageSize. This effects the scrollVisibility value as it results incorrect.

In our current setup, we trigger the event based on changes in the data and props.pageSize properties. But, after trying a few different approaches, we found a fix by adding props.bottomRow and props.topRow as new dependencies.
However, this tweak led to multiple event triggers, in cases where data is refreshed and the pageSize changes

> When table is re-sized to 0 (only table header are visible), the pageSize on the event is still 1

We noticed this inconsistency, during our testing however we also noticed that this behavior is by design. Based on the snippet below, if the table header is visible, the pageSize cannot be less than 1.
Would you like us to generally override this behavior?.
Would recommend that it’s done in a separate PR so that the potential side effects can be carefully analyzed.

https://github.com/appsmithorg/appsmith/blob/release/app/client/src/widgets/TableWidgetV2/widget/index.tsx#L1231-L1233
```
index.tsx
    pageSize={
        isVisibleHeaderOptions ? Math.max(1, pageSize) : pageSize + 1
    }
```

keyurparalkar:
> In our current setup, we trigger the event based on changes in the data and props.pageSize properties. But, after trying a few different approaches, we found a fix by adding props.bottomRow and props.topRow as new dependencies.
> However, this tweak led to multiple event triggers, in cases where data is refreshed and the pageSize changes

@gitstart-appsmith you can go with the fix of adding bottomRow and topRow dependencies. CC: @chandannkumar 

> We noticed this inconsistency, during our testing however we also noticed that this behavior is by design. Based on the snippet below, if the table header is visible, the pageSize cannot be less than 1.
> Would you like us to generally override this behavior?.
> Would recommend that it’s done in a separate PR so that the potential side effects can be carefully analyzed.
> 

@chandannkumar @gitstart-appsmith we can ignore this since this is an edge case.

app/client/src/widgets/TableWidgetV2/component/Table.tsx
```diff
@@ -327,12 +332,16 @@ export function Table(props: TableProps) {
 
   useEffect(() => {
     if (scrollBarRef.current) {
-      setPageHeight(scrollBarRef.current.getScrollElement().scrollHeight || 0);
+      setPageHeight({
+        scrollHeight: scrollBarRef.current.getScrollElement().scrollHeight || 0,
+        clientHeight: scrollBarRef.current.getScrollElement().clientHeight || 0,
```
keyurparalkar, line 337: @gitstart-appsmith why are we passing object in `pageHeight` instead of number? Since `clientHeight` is not being used anywhere, is there a reason we added it?

gitstart-appsmith: Hi @keyurparalkar,

In our test, we were able to reproduce edge cases where the scrollHeight remains thesame on a slight resize.
Including the scrollHeight and clientHeight makes it more stable and consistent.

[Here's a quick demo to reproduce the described case.](https://www.loom.com/share/34cb5ed1c0f046b8852135d57432ccb8?sid=2dfe38b6-e41d-40c2-a816-20e001342219)

keyurparalkar (approved):

keyurparalkar:
@chandannkumar @gitstart-appsmith has fixed the changes. Can you please re-test?

gitstart-appsmith:
Hello, @chandannkumar @keyurparalkar Thank you for highlighting this issue.
In the previous logic, we excluded certain logs to decrease the frequency of event logs since we have four dependent properties.
However, as you pointed out in your last comment, this approach can lead to unintended side effects.
With the current update, we’ve made adjustments to guarantee that the event log will be triggered whenever any of the dependencies undergo changes.
This means that in specific cases, you may notice multiple logs generated due to changes in various dependent properties, with the last log reflecting the latest state as shown in the demo below.

https://www.loom.com/share/6c8732fdfa3048a9a9f0bf57fe291d65?sid=169cc1f4-c0d6-4b4b-b21f-0f0ff19e20e4

keyurparalkar:
@chandannkumar can you please test the callouts once the DP is generated?

chandannkumar:
@gitstart-appsmith 
- We can observe that multiple events got fired for 1 time resizing of the Table widget(Page size). Although the last fired event's result are correct but multiple events could create inaccuracy in upcoming tasks
Video: https://www.loom.com/share/704280cbe8764ead812caf6ebd73a738?sid=fdac4924-5245-4487-aaf2-0213d2732df1

cc: @sbalaji1192

gitstart-appsmith:
Hi @chandannkumar, @keyurparalkar,

We've taken a close look at the issue you highlighted in your video, and we were able to replicate the multiple event logs.

It's essential for us to maintain all the current dependencies set up in the `useEffect` to ensure that we continue to meet the requirements you've raised in your previous reviews.

One possible solution we're considering is introducing a delay in the event logs. This approach would ensure that only the latest event log within a specified timeframe, say 1 second, gets logged. We believe this could help address the issue.

Before proceeding with this solution, we'd love to hear your thoughts and any suggestions you might have.

PR #25846, opened 2023-07-31
ci: Deploy preview from branch

sum35h:
Updated ad-hoc dp workflow to update the db with progress.
- This workflow should only be called via internal ops api.
- This workflow uses the internal ops api to patch data with the updates on the workflow statuses.


scripts/build_dp_from_branch.sh
```diff
@@ -5,6 +5,9 @@
 
 mkdir ~/.aws; touch ~/.aws/config
 
+echo "[default]
```
sum35h, line 8: Could not figure out setting creds for eks, so I've used this, since it's also being used by the pr dp script https://github.com/appsmithorg/appsmith/blob/release/scripts/deploy_preview.sh#L8

.github/workflows/server-build.yml
```diff
@@ -18,6 +18,11 @@ on:
         required: false
         type: string
 
+  pull_request:
```
sum35h, line 21: Not sure why this got removed, this is required by other workflows that use server build.

.github/workflows/ad-hoc-deploy-preview.yml
```diff
@@ -12,38 +12,27 @@ on:
...
-  #         #TODO Add mongo URI as secret
-
+  update-job-details-to-db:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Update workflow details
+        run: |
+          curl --location --request PATCH 'https://ops.appsmith.com/appsmith_deploy_workflow/${{github.event.inputs.uid}}' \
+          --header 'Content-Type: application/json' \
+          --header 'Authorization: Bearer ${{secrets.INTENRAL_OPS_API_TOKEN}}' \
```
sharat87, line 29: We can use `--oauth2-bearer` for this:

```suggestion
          --oauth2-bearer '${{secrets.INTENRAL_OPS_API_TOKEN}}' \
```

.github/workflows/ad-hoc-deploy-preview.yml
```diff
@@ -12,38 +12,27 @@ on:
...
+          --data ' {
+                  "status": "IN_PROGRESS",
+                  "githubWorkflowId": ${{github.workflow.id}},
+                  "githubWorkflowHtmlUrl": ${{github.workflow.githubWorkflowHtmlUrl}},
```
sharat87, line 33: Won't this be unquoted, and so result in invalid JSON?

```suggestion
                  "githubWorkflowHtmlUrl": "${{github.workflow.githubWorkflowHtmlUrl}}",
```

.github/workflows/ad-hoc-deploy-preview.yml
```diff
@@ -12,38 +12,27 @@ on:
...
-  #                                                                 }
-  #                                                               )'
-  #         #TODO Add mongo URI as secret
-
+  update-job-details-to-db:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Update workflow details
+        run: |
+          curl --location --request PATCH 'https://ops.appsmith.com/appsmith_deploy_workflow/${{github.event.inputs.uid}}' \
```
sharat87, line 27: Do we need `--location`? Are we expecting redirects here? 🤔

.github/workflows/ad-hoc-deploy-preview.yml
```diff
@@ -166,30 +138,12 @@ jobs:
...
-         # mongosh '' --eval 'db.appsmiht-ce-github-dp-runs.update(
-                                                                 # { _id:  ${{ github.event.inputs.uid }} },
-                                                                 # {
-                                                                  #  $set: {
-                                                                 #     status: "DEPLOYMENT IN PROGRESS"
-                                                                #    }
-                                                                #  }
-                                                            #    )'
+      - name: Set up Depot CLI
+        uses: depot/setup-action@v1
```
sharat87, line 142: Are we using Depot anywhere? 🤔

.github/workflows/ad-hoc-deploy-preview.yml
```diff
@@ -72,31 +61,14 @@ jobs:
...
-      #     mongosh '' --eval 'db.appsmiht-ce-github-dp-runs.update(
-      #                                                             { _id:  ${{ github.event.inputs.uid }} },
-      #                                                             {
-      #                                                               $set: {
-      #                                                                 status: "BUILD SUCCESSFUL"
-      #                                                               }
-      #                                                             }
-      #                                                           )'
+      - name: Set up Depot CLI
+        uses: depot/setup-action@v1
```
sharat87, line 65: Are we using Depot anywhere? 🤔

.github/workflows/server-build.yml
sharat87: I think this was intentional. It's being triggered as part of a different workflow now.

sharat87 (changes requested):

.github/workflows/ad-hoc-deploy-preview.yml
sum35h: There are no redirects, removing -L

.github/workflows/ad-hoc-deploy-preview.yml
sum35h: Noted.

.github/workflows/ad-hoc-deploy-preview.yml
sum35h: Actually, we're not using it. I went ahead with copying it from the other dp workflow yaml file and since it was working I did not touch it. I'm removing it here. Will remove it from the other dp workflow as well later.

.github/workflows/server-build.yml
```diff
@@ -18,6 +18,11 @@ on:
...
+    branches: [release, master]
+    paths:
+      - "app/server/**"
+
```
sum35h, line 25:
```suggestion
```

.github/workflows/server-build.yml
```diff
@@ -18,6 +18,10 @@ on:
         required: false
         type: string
 
+  pull_request:
+    branches: [release, master]
+    paths:
+      - "app/server/**"
```
sum35h, line 24:
```suggestion
```

sharat87 (approved):

PR #25856, opened 2023-07-31
ci: Default to release CS unless building for prod

sharat87:
When running Appsmith release image, it's currently using production CS, which is causing a lot of incompatibility related errors since `release` moves much faster than prod.

This changes this. Note that for `:latest` and `:nightly` images, we do NOT change this default. There's no change for Docker images with those tags. We _only_ want change `:release` image, DP images, and images built for Cypress runs.


sarojsarab (approved):
LGTM

PR #25870, opened 2023-08-01
ci: Use vars.EDITION instead of hard-coded edition to avoid conflicts with EE sync

sharat87:
Note: Please review/approve only, do not merge. This PR is expected to cause conflicts and so I'd like to be around when this is merged.


.github/workflows/fat-migration.yml
sharat87, line 1: @Aishwarya-U-R, are we still using this workflow file today? If not, can we delete it please?

Aishwarya-U-R: Not used 🤔

sharat87: Deleting the file then. Thanks for confirming.

Aishwarya-U-R: Have deleted it & PR is merged to release @sharat87 🙏🏻

sarojsarab (approved):
LGTM

PR #25878, opened 2023-08-01
fix: Fix failing product alert API for anonymous user

sharat87:
Fix critical problem where public apps are redirecting to login page, because of product-alerts API not accepting anonymous users.


tkAppsmith (dismissed):
Approving as this is needed to fix a critical issue which is not allowing anonymous users to use the app by returning 401 status

nosun10005:
The same thing is happening again

PR #25881, opened 2023-08-01
fix: Update product alert stub to check for 200 status

hetunandu:
Product alert api was mocked in the cypress tests which missed cases where the api would return a 401 error. Updating the mock to now make the call, verify the response status code to be 200 and then stub the response.


akash-codemonk (approved):

ankitakinger (dismissed):

Aishwarya-U-R (approved):

app/client/cypress/support/commands.js
```diff
@@ -1133,15 +1133,26 @@ Cypress.Commands.add("startServerAndRoutes", () => {
...
-    responseMeta: {
-      status: 200,
-      success: true,
-    },
-    data: {},
-    errorDisplay: "",
-  });
+
+  cy.intercept("GET", "/api/v1/product-alert/alert", (req) => {
+    try {
```
hetunandu, line 1138: Try catch is needed when intercepting calls and using req.continue because it starts a promise and out test case could navigate away from a page before the promise is resolved.

ankitakinger (approved):

app/client/cypress/support/commands.js
```diff
@@ -1374,6 +1385,7 @@ Cypress.Commands.add("createSuperUser", () => {
       expect(interception.request.body).contains("signupForNewsletter=true");
     });
   }
+  cy.wait("@getWorkspace");
```
hetunandu, line 1388: After super user creation, we make the user reach onboarding. If we logout before the page load, it will fail many api calls causing random errors

Aishwarya-U-R (commented):
Approving Cypress changes.

PR #25921, opened 2023-08-02
chore: Removing isAsync consumption on the backend

### LINKED ISSUE #25831: [Task]: Server side changes to remove isAsync field from JS functions

### SubTasks

- [x] Remove isAsync from JS actions
- [x] Stop resetting configuration fields for JS actions on the basis of isAsync
- [x] Treat all functions as async in the on page load calculations. This means that all functions will need to be parsed through for dependencies, and none of them should show up in on page load, unless they have had a `JSObject1.myFunc1.data` kind of usage

nidhi-nair:
Fixes #25831 

The test case has a DSL with the following bindings:
```
{
    "widgetName": "testWidget",
    "k1": "{{ aGetAction.data }}",
    "k2": "{{ aPostAction.data.users.name }}",
    "k3": "{{ anotherPostAction.run() }}",
    "k4": "{{ Collection.anAsyncCollectionActionWithoutCall.data }}",
    "k5": "{{ Collection.aSyncCollectionActionWithoutCall.data }}",
    "k6": "{{ Collection.anAsyncCollectionActionWithCall() }}",
    "k7": "{{ Collection.aSyncCollectionActionWithCall() }}",
    "k8": "{{ Collection.data() }}",
    "k9": "{{ Collection2.data() }}"
}
```

And the following function definitions:
```
async Collection.anAsyncCollectionActionWithoutCall() {
    return hiddenAction1.data
}

Collection.aSyncCollectionActionWithoutCall() {
    return hiddenAction2.data   
}

async Collection.anAsyncCollectionActionWithCall() {
    return hiddenAction3.data
}

Collection.aSyncCollectionActionWithCall() {
    return hiddenAction4.data
}

Collection.data() {
    return hiddenAction5.data
}

async Collection2.data() {
    return hiddenAction6.data
}
```

Here is the expected on page load matrix:
```
[
  ["aGetAction", "hiddenAction1", "hiddenAction2", "hiddenAction3", "hiddenAction4", "hiddenAction5", "hiddenAction6"],
  ["aPostAction"],
  ["Collection.anAsyncCollectionActionWithoutCall", "Collection.aSyncCollectionActionWithoutCall"]
]
```


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ActionConfiguration.java
```diff
@@ -83,7 +83,6 @@ public class ActionConfiguration implements AppsmithDomain {
     // Body, the raw class data, is shared with API type actions
     // Represents the values that need to be
     List<JSValue> jsArguments;
-    Boolean isAsync;
```
subrata71, line 86: Even if the client sends isAsync attribute in the request payload, the server will not capture this. Is that correct? @nidhi-nair

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/PageLoadActionsUtilCEImpl.java
```diff
@@ -1132,10 +1124,6 @@ private Set<String> actionCandidatesForPageLoadFromBinding(
 
                 // If the reference is to a sync JS function, discard it from the scheduling order
                 ActionDTO actionDTO = actionNameToActionMap.get(entity);
-                if (PluginType.JS.equals(actionDTO.getPluginType())
-                        && FALSE.equals(actionDTO.getActionConfiguration().getIsAsync())) {
-                    isCandidateForPageLoad = FALSE;
-                }
```
subrata71, line 1127: The bindings that exist in the page DSL for sync JS function, will now be treated as onPageLoad action after this PR goes in. And we are okay with that. Is that correct? @nidhi-nair

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ActionConfiguration.java
nidhi-nair: Yes, this field has no value on either client or server anymore. The upstream to this branch has the changes for client side.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/PageLoadActionsUtilCEImpl.java
nidhi-nair: Yes, all functions in Appsmith are now essentially async, and everything needs to follow the same rules as async functions on the backend moving forward.

subrata71: So, users will start seeing these actions being executed on onPageLoad event. This kind of surprise to the users is acceptable, right?

nidhi-nair: The final result is actually not affected here. The filtering step removes all references that are direct function calls, that rule applied to both sync and async functions

subrata71: I am not talking about the function calls but the following
`JSObject1.previousSyncFunction.data`. For this kind of reference the user didn't see it treated as on page load action but now they will see that coming as on page load action assuming such reference exists in the page DSL.

nidhi-nair: Actually in this case they did used to see the function as on page load

nidhi-nair: Sorry, in short, end user won't see a difference in functionality for any of their bindings :)

subrata71: > Actually in this case they did used to see the function as on page load

But not for the sync JS function. @nidhi-nair

subrata71: > Sorry, in short, end user won't see a difference in functionality for any of their bindings :)

This is true once the PR goes in prod.

nidhi-nair: A sync function used in that way would never have been functional

subrata71 (approved):
LGTM

PR #25927, opened 2023-08-02
chore: Add support for custom CS URL in local_testing.sh script

sharat87:
Slack thread with details: https://theappsmith.slack.com/archives/CGBPVEJ5C/p1690917878540839?thread_ts=1690866395.690119&cid=CGBPVEJ5C

### DIFF
--- a/scripts/local_testing.sh
+++ b/scripts/local_testing.sh
@@ -10,7 +10,7 @@ display_help()
   echo "If --local or -l is passed, it will build with local changes"
   echo "---------------------------------------------------------------------------------------"
   echo
-  echo "Syntax: $0 [-h] [-l] [-r [remote_url]] [branch_name]"
+  echo "Syntax: $0 [-h] [-l] [-r [remote_url]] [branch_name] [cs_url]"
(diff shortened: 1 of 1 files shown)


scripts/local_testing.sh
```diff
@@ -81,7 +83,8 @@ popd
 pushd app/client/packages/rts/ > /dev/null && ./build.sh > /dev/null && pretty_print "RTS build successful. Starting Docker build ..."
 
 popd
-docker build -t appsmith/appsmith-ce:local-testing . > /dev/null && pretty_print "Docker image build successful. Triggering run now ..."
+docker build -t appsmith/appsmith-ce:local-testing --build-arg APPSMITH_CLOUD_SERVICES_BASE_URL="${cs_url:-https://release-cs.appsmith.com}" . > /dev/null
```
ankitakinger, line 86: Just trying to understand this, so if we don't pass the CS URL in the command, it will be default take release-cs correct?

ankitakinger (approved):
LGTM

PR #25943, opened 2023-08-02
fix: Update dependencies to bring fixes for CVEs

sharat87:
1. CVE-2023-3635 at https://github.com/appsmithorg/appsmith/security/dependabot/251.
2. CVE-2023-2976 at https://github.com/appsmithorg/appsmith/security/dependabot/235.

### DIFF
--- a/app/server/appsmith-interfaces/pom.xml
+++ b/app/server/appsmith-interfaces/pom.xml
@@ -220,12 +220,6 @@
             <version>${okhttp3.version}</version>
         </dependency>
         <dependency>
-            <groupId>com.squareup.okio</groupId>
-            <artifactId>okio</artifactId>
-            <version>1.17.2</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
             <version>${mockito.version}</version>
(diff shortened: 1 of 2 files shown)


app/server/appsmith-server/pom.xml
```diff
@@ -166,11 +166,6 @@
             <version>4.3.2</version>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-            <version>30.0-jre</version>
-        </dependency>
```
sharat87, line 173: Doesn't look like we're using this anywhere.

app/server/appsmith-interfaces/pom.xml
```diff
@@ -219,12 +219,6 @@
             <artifactId>okhttp</artifactId>
             <version>${okhttp3.version}</version>
         </dependency>
-        <dependency>
-            <groupId>com.squareup.okio</groupId>
-            <artifactId>okio</artifactId>
-            <version>1.17.2</version>
-            <scope>test</scope>
-        </dependency>
```
sharat87, line 227: An updated version of this comes in as a transitive dependency of okhttp, defined just above. So we don't need this.

nidhi-nair (approved):

PR #26046, opened 2023-08-04
fix: Fix leaking NGINX version

sharat87:
When Appsmith is deployed with HTTPS, all HTTP requests are redirected to HTTP. But this redirection response, is leaking the NGINX version in the `Server` header. This PR fixes that.


mohanarpit (approved):

sharat87:
The test passes locally. 😐

shot-2023-08-07-00-58-00

PR #26066, opened 2023-08-06
chore: Refactor NGINX config templates, merge them, to reduce duplicate code

sharat87:
Majority of the NGINX config is the same, for both HTTP and HTTPS. Having two separate templates for them is making configuration changes error-prone, where we often risk forgetting making the same change in the other file.

This PR merges the two files into one, so the above risk isn't there. It also makes it easier to experiment with the file while developing, since we have to make every single change twice during development.


deploy/docker/templates/nginx/nginx-app.conf.sh
```diff
@@ -35,8 +44,30 @@ access_log /dev/stdout;
 server_tokens off;
 
 server {
+
+$(
+if [[ $use_https == 1 ]]; then
+  echo "
+  listen 80;
```
nidhi-nair, line 51: Should this be listening at the configurable port like in L#66?

deploy/docker/templates/nginx/nginx-app.conf.sh
```diff
@@ -35,8 +44,30 @@ access_log /dev/stdout;
...
+  server_name $custom_domain;
+  return 301 https://\$host\$request_uri;
+}
+
+server {
+  listen 443 ssl http2;
+  server_name _;
```
nidhi-nair, line 58: Why is this one not using custom domain?

sharat87: Also, keeping status-quo. It actually doesn't really matter. We can just change everything to just be `_` and move on. This is a wildcard, like accept any domain. This becomes important, if in a single NGINX configuration, we have multiple domains. But this NGINX configuration is Appsmith's own, and lives inside Appsmith container, so it doesn't really matter here. May be we switch everything to `_` later sometime. 😃

deploy/docker/templates/nginx/nginx-app.conf.sh
nidhi-nair: Bold of you to claim you have grown paranoid only just now. 😝

nidhi-nair (approved):

PR #26104, opened 2023-08-08
chore(deps): Update yaml dependency to get fix for CVE-2023-2251

sharat87:
Should resolve https://github.com/appsmithorg/appsmith/security/dependabot/224.


app/client/yarn.lock
```diff
@@ -30089,9 +30089,9 @@ __metadata:
   linkType: hard
 
 "yaml@npm:^2.1.1, yaml@npm:^2.2.1":
-  version: 2.2.1
-  resolution: "yaml@npm:2.2.1"
-  checksum: 84f68cbe462d5da4e7ded4a8bded949ffa912bc264472e5a684c3d45b22d8f73a3019963a32164023bdf3d83cfb6f5b58ff7b2b10ef5b717c630f40bd6369a23
+  version: 2.2.2
```
KelvinOm, line 30092: @sharat87 How do you made this changes?

sharat87: Same way I've done all the others.

1. I do a `yarn add yaml@2.2.2`.
2. Then edit `package.json` and remove the `"yaml": "^2.2.2",` line.
3. Then do `yarn` again.
4. Check the diff that only `yarn.lock` changed, and only `yaml`'s version changed, and only to `2.2.2`.

I hate this method as it feels like a workaround, but I searched yarn's docs for a simpler method, but nothing was simpler or more reliable than this. 🤷

KelvinOm (approved):

app/client/yarn.lock
sharat87: Interesting. The reason I was focused on having changes only in `yarn.lock`, unless it's a dependency we've listed in our `"dependencies"` or `"devDependencies"` already, is because that was how Dependabot's auto-created PRs were doing it. 🙂

Thanks for the review!

PR #26128, opened 2023-08-08
chore: Add `X-Content-Type-Options` to all downstream responses

sharat87:
So far, only calls that go to the Java backend, had the `X-Content-Type-Options` header in the responses. This PR adds them to all responses by

1. adding it to NGINX configuration.
2. removing it from Spring security's configuration, so we don't end up with _two_ `X-Content-Type-Options` headers in the response.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
@@ -152,7 +152,7 @@ public class SecurityConfig {
                 .disable()
                 .addFilterAt(new CSRFFilter(), SecurityWebFiltersOrder.CSRF)
                 // Default security headers configuration from
-                // https://docs.spring.io/spring-security/site/docs/5.0.x/reference/html/headers.html.
+                // https://docs.spring.io/spring-security/site/docs/5.0.x/reference/html/headers.html
                 .headers()
                 .contentTypeOptions()
                 .disable()
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
@@ -151,6 +151,14 @@ public class SecurityConfig {
                 .csrf()
                 .disable()
                 .addFilterAt(new CSRFFilter(), SecurityWebFiltersOrder.CSRF)
+                // Default security headers configuration from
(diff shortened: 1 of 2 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
```diff
@@ -151,6 +151,14 @@ public SecurityWebFilterChain securityWebFilterChain(ServerHttpSecurity http) {
                 .csrf()
                 .disable()
                 .addFilterAt(new CSRFFilter(), SecurityWebFiltersOrder.CSRF)
+                // Default security headers configuration from
+                // https://docs.spring.io/spring-security/site/docs/5.0.x/reference/html/headers.html.
```
nidhi-nair, line 155:
```suggestion
                // https://docs.spring.io/spring-security/site/docs/5.0.x/reference/html/headers.html
```

Nit: The period at the end does not make this URL functional

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
```diff
@@ -151,6 +151,14 @@ public SecurityWebFilterChain securityWebFilterChain(ServerHttpSecurity http) {
...
+                .headers()
+                .contentTypeOptions()
+                .disable()
+                .frameOptions()
```
nidhi-nair, line 159: Why are we disabling this?

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
sharat87: Wow, it works fine in IntelliJ! Guess I can't expect that quality everywhere. Thanks for noticing!

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
sharat87: Two reasons.

1. It only applies to `/api` requests, and other requests that end up with the backend server, and not all.
2. Our NGINX sets a CSP with a `frame-ancestors`, which is solving the same problem as this header, and is the recommended way forward. That CSP is even configurable for users, and it's value adapts to configuration. This one, doesn't.

Ref: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options.

nidhi-nair (dismissed):

PR #26132, opened 2023-08-08
fix: fixed failing queries using aggregation pipeline

### LINKED ISSUE #26090: [Bug]: Deploying applications when using DocumentDB as the DB fails.

### SubTasks

The code changes adding via [chore: Improve time to publish application](https://github.com/appsmithorg/appsmith/pull/25516) are not supported by DocumentDB.

Impact of the code change: When creating a new application, it errors out with an Internal Server Error toast message, and also when deploying an application, it errors out with an Internal Server Error toast message.

This is happening because, in order to to improve the publishing time for an application, we set the values of unpublished documents (actions & pages) to the published documents. It uses aggregation update and as mentioned in the [Notion Document](https://www.notion.so/appsmith/Document-DB-aggregates-521b6c3f5cb448079557753900696b05#2686bd1d5e564248b4a67f1134493376), these are not supported by DocumentDB.

tkAppsmith:
Fixes #26090
Fixes https://github.com/appsmithorg/appsmith-ee/issues/1659

#### Type of change
- [x] Manual
- [ ] Jest
- [ ] Cypress
>
>
#### Dev activity

#### QA activity:


app/server/appsmith-server/src/main/java/com/appsmith/server/ServerApplication.java
```diff
@@ -4,12 +4,16 @@
...
 import org.springframework.scheduling.annotation.EnableScheduling;
+import reactor.core.publisher.Hooks;
 
 @SpringBootApplication
 @EnableScheduling
 public class ServerApplication {
 
     public static void main(String[] args) {
+
+        Hooks.onOperatorDebug();
```
nidhi-nair, line 15: Please don't leave this on, there is an IDE configuration that allows you to do this without inserting it in code.

tkAppsmith: This pr is for evaluation of correctness. Hence a lot of things will be changed as per code style and unit tests. But it seems this will not work as the method used here is getting duplicateKeyException. @nidhi-nair

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewPageRepositoryCEImpl.java
```diff
@@ -251,14 +267,55 @@ public Mono<NewPage> findByGitSyncIdAndDefaultApplicationId(
...
+                    mongoOperations.getConverter().write(newPage, document);
+                    document.remove("_id");
+                    return (WriteModel<Document>) new UpdateOneModel<Document>(
+                            new Document("_id", new ObjectId(newPage.getId())), new Document("$set", document));
+                })
+                .collect(Collectors.toList());
+
+        return mongoOperations
+                .getCollection(mongoOperations.getCollectionName(NewPage.class))
+                .flatMapMany(documentMongoCollection -> documentMongoCollection.bulkWrite(dbObjects))
```
nidhi-nair, line 318: Can you please help me understand why the result of a bulk write is split into a list of different `BulkWriteResult` objects?

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ce/NewActionServiceUnitTest.java
```diff
@@ -199,9 +200,8 @@ public void testMissingPluginIdAndTypeFixForJSPluginType() {
     @Test
     public void testPublishActionArchivesAndPublishesActions() {
         String applicationId = "dummy-application-id";
-        UpdateResult updateResult = Mockito.mock(UpdateResult.class);
-        Mockito.when(updateResult.getModifiedCount()).thenReturn(10L);
-        Mockito.when(updateResult.getMatchedCount()).thenReturn(5L);
```
nidhi-nair, line 204: My previous question is because we are losing this insight as a result of this change now. Or is there a way to retain a sense of matched versus modified count?

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewActionRepositoryCEImpl.java
```diff
@@ -570,16 +577,33 @@ public Flux<NewAction> findByDefaultApplicationId(String defaultApplicationId, O
...
         Criteria applicationIdCriteria =
                 where(fieldName(QNewAction.newAction.applicationId)).is(applicationId);
-        // using aggregation update instead of regular update here
-        // it's required to set a field to a value of another field from the same domain
-        AggregationUpdate aggregationUpdate = AggregationUpdate.update()
-                .set(fieldName(QNewAction.newAction.publishedAction))
-                .toValue("$" + fieldName(QNewAction.newAction.unpublishedAction));
 
-        return updateByCriteria(List.of(applicationIdCriteria), aggregationUpdate, permission);
+        Mono<Set<String>> permissionGroupsMono =
```
nidhi-nair, line 584: Are we not routing these queries based on type of DB?

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/BaseRepositoryImpl.java
```diff
@@ -161,7 +161,9 @@ public Flux<T> findAll() {
                 .flatMapMany(principal -> {
                     Query query = new Query(notDeleted());
                     return mongoOperations.find(
-                            query, entityInformation.getJavaType(), entityInformation.getCollectionName());
+                            query.cursorBatchSize(10000),
```
nidhi-nair, line 164: Does it make sense to either rename this method, or overload it at least?

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewActionRepositoryCEImpl.java
tkAppsmith: Detection of db type is not yet reliable as discussed in last engineering meet.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ce/NewActionServiceUnitTest.java
tkAppsmith: Yes. I agree. I need to think how to incorporate this. I guess we cant get the matched count in the new approach.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewPageRepositoryCEImpl.java
tkAppsmith: I used an existing method hence Im not sure what the author's use case was. We can modify both the implementations as to retain the existing method interfaces so that the unit test is not affected.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/BaseRepositoryImpl.java
tkAppsmith: Ok. We can accept a cursor size and then pass that argument and in that manner we will be able to override this method.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewPageRepositoryCEImpl.java
tkAppsmith: I think the bulk write operation return a flux which then requires us to use subscription in caller as well as in unit tests hence it will still need code change. With this list setup I guess the results are getting updated and the above used matched and updated count is irrelevant in this case as all records that we are passing will be overwritten by matching _id.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ce/NewActionServiceUnitTest.java
nidhi-nair: I do see that the new type you are returning has a method defined in its interface for both. This is a minor point, I'm just curious as to why that is not possible here.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewActionRepositoryCEImpl.java
nidhi-nair: I thought we were going the environment variable route till we found a solution? This would affect larger instances, especially our own cloud instance right?

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewPageRepositoryCEImpl.java
nidhi-nair: Hmm .. Is that because with the aggregation pipeline it is possible to have multiple sets of updates? 🤔 Even in that case, I suppose we'd be interested only in the first entry in this list.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewActionRepositoryCEImpl.java
tkAppsmith: You mean an env variable set by user?? we thought that approach will be unreliable as users might configure the instance wrongly then things will fail at some point later on in the usage life cycle of app.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewPageRepositoryCEImpl.java
tkAppsmith: This bulkUpdate method is not run using an aggregation pipeline. I think the usage of reactive mongo interface to do bulk write dictates the return of flux of bulkWriteResult which basically might mean that this stream of objects each might correspond to batches of updates being done to collection.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewActionRepositoryCEImpl.java
nidhi-nair: We should at least have infrastructure created in code to distinguish between two types of database calls. That is why the environment variable is useful. As of today, the only instance that suffers really badly as a result of these performance problems is cloud. We could choose to not publish information about the environment variable, but we should at least have the provision to use this variable in our own production instance and switch between code paths.

This then also becomes an exercise in making sure that our code structure is designed to support alternate queries as and when required.

nidhi-nair: The idea here is that whenever we find the reliable identifier for figuring out if an instance is hosting DocDB or Mongo, we can immediately implement that conditional in our code, and switch out of the environment variable way of doing things. One criteria that definitely cannot be overlooked for this effort is that the cloud instance should be able to continue to use the most performant queries all the time.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewPageRepositoryCEImpl.java
nidhi-nair: Can you please confirm this so we can resolve this thread?

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewActionRepositoryCEImpl.java
tkAppsmith: You mean we should have stubs that read and populate the environment variable and for now they shall be defaulting to mongo only. Im asking this since we are not yet really certain how to detect document db instance with 100% accuracy. The host name in server build confg command can also be proxied.

nidhi-nair: Let's do it the other way around, unless mentioned in the environment variable, we can use DocDb. And then in our own cloud instances we can set the variable up.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewPageRepositoryCEImpl.java
```diff
@@ -251,14 +267,55 @@ public Mono<NewPage> findByGitSyncIdAndDefaultApplicationId(
...
-        AggregationUpdate aggregationUpdate = AggregationUpdate.update()
-                .set(fieldName(QNewPage.newPage.publishedPage))
-                .toValue("$" + fieldName(QNewPage.newPage.unpublishedPage));
 
-        return updateByCriteria(List.of(applicationIdCriteria), aggregationUpdate, permission);
+        Mono<Set<String>> permissionGroupsMono =
+                getCurrentUserPermissionGroupsIfRequired(Optional.ofNullable(permission));
+
+        return permissionGroupsMono.flatMap(permissionGroups -> {
+            AggregationOperation matchAggregationWithPermission = null;
```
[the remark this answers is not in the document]
tkAppsmith: I have done the necessary instrumentation for the changed queries on local. First I compared the application creation flow between release and this branch changes. I have selected application creation flow because it was the flow that invokes these queries. There is x10 difference in comparison in some apis (get pages and get actions) ie this branch is x10 slower. But workspace create application api call (applications?workspaceId={workspace}) is faster on this branch (x5 - x10). Hence I deduced that there are recent changes on release that have bought these optimisations on the unrelated get api calls. I then proceeded to revert all these changes in the current branch and retried to application creation flow and the numbers still stand at x10 for the get pages and get actions calls which means the faster api calls are due to optimisations made on the release branch and hence we can safely merge in these changes and they won’t have performance implications as such.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewPageRepositoryCEImpl.java
tkAppsmith: I ran some logs on the current approach. It seems the flux is returned because of batch based updates, where each element has its own matched and updated count. The current implemetation seems to be blocking because it converts the flux to list and returns a mono of that. We can change that to make it return a flux of write result but it will need to change both the methods and then corresponsing changes in service caller layer. We can take this up as refactoring work afterwards since this will need some amount of code changes and re review and performance metrics.

tkAppsmith: And for the total number of matched and updated counts, we will have to finally convert the flux to list and then sum up the counts of each element to get the total counts due to async nature of this operation.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewActionRepositoryCEImpl.java
tkAppsmith: Can we take this up as a separate initiate, We have some bugs still left to fix for doc db and we can club this initiative with them in next sprint. Right now the performance instrumentation shows no great effect of these query changes on api call response times.

tkAppsmith:
Open items issues:
Async bulk update: https://github.com/appsmithorg/appsmith/issues/27138
Overload or different name method with cursorbatchsize: https://github.com/appsmithorg/appsmith/issues/27139
Code toggle/routing for document db connected instances: https://github.com/appsmithorg/appsmith/issues/27140

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/BaseRepositoryImpl.java
tkAppsmith: https://github.com/appsmithorg/appsmith/issues/27139

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewActionRepositoryCEImpl.java
tkAppsmith: https://github.com/appsmithorg/appsmith/issues/27140

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewPageRepositoryCEImpl.java
tkAppsmith: https://github.com/appsmithorg/appsmith/issues/27138

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ce/NewActionServiceUnitTest.java
tkAppsmith: https://github.com/appsmithorg/appsmith/issues/27138

nidhi-nair (approved):
LGTM, is there anything from @nayan-rafiq 's comments that needs changing @tkAppsmith ?

tkAppsmith:
Hey nidhi. I have done as requested by @nayan-rafiq . He also has a comment if we can scope out moving bulk update to the base repository. I can add this in one of the issues we have taken up as part of next iteration of enhancements on this part of work.

btsgh:
Since, this same code change was regression tested on EE DP raised for Doc DB, and the issues for which this one was targeted are working, adding TestPlanApproved label. cc: @tkAppsmith

Issue #26019, opened 2023-08-04
chore: POC on CSS Modules 


jsartisan:
Currently, for every new CSS feature, we have to wait for its support in styled-components. We want to try CSS modules with post-css plugins and check if we can work with CSS modules so that we don't have to wait for every CSS feature in the styled component.

KelvinOm:
Duplicate of https://github.com/appsmithorg/appsmith/issues/25536

KelvinOm:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Widget-Design-System-6434e3402be2c1002ea740be/issues/appsmithorg/appsmith/26019?planning-poker) with Zenhub @ichik @jsartisan @KelvinOm

PR #26162, opened 2023-08-08
feat: enable CSS Modules fro WDS 

jsartisan:
Fixes: #26019


app/client/packages/design-system/widgets/src/components/Button/styles.module.css
```diff
@@ -81,14 +15,56 @@ export const StyledButton = styled(HeadlessButton)<StyledButtonProps>`
   min-inline-size: var(--sizing-8);
   border-radius: var(--border-radius-1);
 
-  // Note: adding important here as ADS is overriding the color of blueprint icon globally
-  // TODO(pawan): Remove this once ADS team removes the global override
+  @each $color in accent, neutral, negative, warning, positive {
```
KelvinOm, line 18: Could we create global variable for this and import it here? Something like this:

```
:root {
    --colors: accent, neutral, negative, warning, positive;
}

@each $color in var(--colors) {
    @import "$val.css";
}
```

KelvinOm (changes requested):

app/client/packages/design-system/widgets/src/components/Button/Button.tsx
```diff
@@ -1,18 +1,22 @@
...
+
 import type {
   BUTTON_COLORS,
   BUTTON_VARIANTS,
   BUTTON_ICON_POSITIONS,
 } from "./types";
-import { DragContainer, StyledButton } from "./index.styled";
+import { Text } from "../Text";
+import { Spinner } from "../Spinner";
+import styles from "./styles.module.css";
```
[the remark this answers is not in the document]
jsartisan: I tried this. Didn't work :|

app/client/packages/design-system/widgets/src/components/Button/styles.module.css
jsartisan: Where should we store it and how to load it?
Let's say I store it like` global.css` in `styles` folder. Will i load it like

`@import "../styles/global.css"`
If yes, what will happen if another component also needs it? As far as I know, only a few components will need semantics like buttons and icon buttons. I feel extracting these values to an variable won't help much as it will be mostly used in button.

app/client/packages/storybook/.storybook/main.ts
```diff
@@ -15,6 +15,27 @@ async function webpackConfig(config) {
...
+
+  config.module.rules.push({
+    test: /\.css$/,
+    use: [
+      {
+        loader: "postcss-loader",
+        options: {
+          postcssOptions: {
+            plugins: [
+              require("postcss-at-rules-variables"),
```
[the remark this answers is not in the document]
jsartisan: I just saw you used it like this only. Should I check without require too?

app/client/packages/storybook/package.json
```diff
@@ -19,6 +19,7 @@
     "@storybook/addon-docs": "^7.1.1",
     "@storybook/addon-measure": "^7.1.1",
     "@storybook/addon-outline": "^7.1.1",
+    "@storybook/addon-styling": "^1.3.5",
```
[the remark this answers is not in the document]
jsartisan: Removed.

app/client/packages/design-system/widgets/src/components/Button/Button.tsx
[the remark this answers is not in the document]
KelvinOm: This is because of our current build configuration. Let's skip this for now.

app/client/packages/design-system/theming/src/theme/src/ProviderStyleSheet.ts
```diff
@@ -0,0 +1,105 @@
+import { StyleSheet } from "@emotion/sheet";
+import kebabCase from "lodash/kebabCase";
+import { createTypographyStringMap } from "../../typography";
+import { cssRule } from "../../utils/cssRule";
+import { SheetTypes } from "./types";
+
+import type { RootUnit, ThemeToken } from "../../token";
+import type { FontFamily, Typography } from "../../typography";
+
+export class ProviderStyleSheet {
```
riodeuno, line 10: What are sheets and what purpose do they serve?
Also, could we add jsdoc comments to the methods in this file?

KelvinOm: @riodeuno Added comments. Let me know if I should do more or something is unclear.

riodeuno: Thanks for adding the comment. It is a lot clearer now. Let's set this thread to resolved.

PR #26208, opened 2023-08-09
test: cypress - switch group cases - 1

sharanya-appsmith:
Switch Group widget test cases:

1. Check for empty, duplicate values and other error texts in options
2. Check default selected value function and error texts
3. Setting selectedValues to undefined does not crash the app // old case -> moved to ts
4. Set Label, Tooltip, Inline and check switch group
5. Check visible, disabled, Height
6. Check events - on selection change


NandanAnantharamu (approved):
LGTM

PR #26319, opened 2023-08-14
fix: Add migration to show/hide data point labels on charts

### LINKED ISSUE #26321: [Bug]: The user should be able to toggle label values displayed on a chart

### Description

If the chart size is too small, the label values overlap with each other and make the chart unreadable. The users should have an option to choose to display/hide labels on chart series.

It used to look like this in Fusion charts [like this](https://theappsmith.slack.com/archives/C0341RERY4R/p1691494048886619?thread_ts=1691488244.791999&cid=C0341RERY4R) and with ECharts it looks [like this](https://theappsmith.slack.com/archives/C0341RERY4R/p1691494068510919?thread_ts=1691488244.791999&cid=C0341RERY4R)

We need to add a property that allows a user to do this.

For data migration,

1. We need to set this property to true if allow scroll is true (because fusion charts showed labels only when allow scroll was true)
2. The property needs to be set to false if allow scroll is false.

### Steps To Reproduce

1. Dnd a Chart Widget.
2. Select Line chart
3. Make the chart size too small.
4. Notice that the labels overlap with each other.
5. The labels shouldn't overlap with each other.

### Public Sample App

_No response_

### Environment

Production

### Issue video log

_No response_

### Version

Cloud

rajatagrawal:
Fixes #26321

### DIFF
--- a/app/client/src/constants/WidgetConstants.tsx
+++ b/app/client/src/constants/WidgetConstants.tsx
@@ -71,7 +71,7 @@ export const layoutConfigurations: LayoutConfigurations = {
   FLUID: { minWidth: -1, maxWidth: -1 },
 };
 
-export const LATEST_PAGE_VERSION = 83;
+export const LATEST_PAGE_VERSION = 84;
 
 export const GridDefaults = {
   DEFAULT_CELL_SIZE: 1,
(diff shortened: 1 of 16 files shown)


sbalaji1192 (approved):

laveena-en:
Verified the below for all basic charts:

- Data migration of `Show labels` wrt existing apps that have `Allow scroll` turned on/off
- State of `Show labels` on new widgets
- Sanity check for charts
- 
PR looks good to me.

PR #26320, opened 2023-08-14
chore: Remove unauthenticated access to actuator APIs

### LINKED ISSUE #26310: [Task]: Remove unauthenticated access to actuator API

### SubTasks

- Allow only basic authentication
- Check that both release and prod work properly with the change
- Make sure to add a note in release for the change

nidhi-nair:
## Description
Only allows authenticated access to actuator APIs moving forward.

Fixes #26310 

#### Type of change

- [x] Manual

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
@@ -11,7 +11,6 @@ import com.appsmith.server.helpers.RedirectHelper;
 import com.appsmith.server.services.AnalyticsService;
 import com.appsmith.server.services.UserService;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import org.apache.commons.lang3.StringUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Bean;
@@ -193,7 +192,7 @@ public class SecurityConfig {
                         ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, CUSTOM_JS_LIB_URL + "/*/view"),
                         ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, PRODUCT_ALERT + "/alert"))
                 .permitAll()
-                .pathMatchers("/public/**", "/oauth2/**", "/actuator/**")
+                .pathMatchers("/public/**", "/oauth2/**")
                 .permitAll()
                 .anyExchange()
                 .authenticated()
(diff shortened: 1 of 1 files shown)


nidhi-nair:
DP can't be used since endpoint is not exposed on CE. Have tested on local, works with auth, does not work without.

sharat87 (commented):
Should we not remove this endpoint from this line as well?
https://github.com/appsmithorg/appsmith/pull/26320/files#diff-5f89474836b0bbb2d36395f0103de005c86480bbba0dfcfe364a9e6a78622095R196

nidhi-nair:
Yes you're right, made the change @sharat87

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
```diff
@@ -121,14 +119,12 @@ public ForwardedHeaderTransformer forwardedHeaderTransformer() {
...
     @Bean
-    @ConditionalOnExpression(value = "'${appsmith.internal.password}'.length() > 0")
     public SecurityWebFilterChain internalWebFilterChain(ServerHttpSecurity http) {
         return http.securityMatcher(new PathPatternParserServerWebExchangeMatcher("/actuator/**"))
                 .authorizeExchange(authorizeExchangeSpec ->
                         authorizeExchangeSpec.anyExchange().authenticated())
                 .httpBasic(httpBasicSpec -> httpBasicSpec.authenticationManager(authentication -> {
-                    if (isAuthorizedToAccessInternal(
-                            authentication.getCredentials().toString())) {
+                    if (INTERNAL_PASSWORD.equals(authentication.getCredentials().toString())) {
```
sharat87, line 127: We used to have `null` protection as part of `StringUtils.isEmpty` on `INTERNAL_PASSWORD`, but we won't have that anymore. Just confirming that that's okay. Otherwise we're good here.

sharat87 (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
nidhi-nair: Empty setting here basically means empty password is expected in the basic auth request as `Basic base64(":")`. We could make that an unacceptable value and stop access to the endpoint if the password is not configured, if that still counts as a +ve on vulnerability. 🤔

sharat87: Nah that's okay, I was only concerned if there's a possibility of an NPE here.

I think that empty password's fine for now. I don't think it'll be flagged, but I do want to come back to it after a couple of versions. Just not to change too much too fast.

Let's go ahead with this.

PR #26348, opened 2023-08-14
poc:emotion

KelvinOm:
Add emotion POC


ichik (approved):

app/client/packages/design-system/theming/package.json
```diff
@@ -16,7 +16,7 @@
   "dependencies": {
     "@capsizecss/core": "^3.1.1",
     "@capsizecss/metrics": "^1.2.0",
-    "lodash": "*",
-    "styled-components": "6.0.2"
+    "@emotion/sheet": "^1.2.2",
```
KelvinOm, line 19: So here we use a [tiny package](https://github.com/emotion-js/emotion/tree/main/packages/sheet) ([just 1.5kB](https://bundlephobia.com/package/@emotion/sheet@1.2.2)). This package also uses a lot of CS in JS libraries under the hood.  

We use it because we need an [improved version](https://github.com/emotion-js/emotion/blob/main/packages/sheet/src/index.js#L113) of the [insertRule](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/insertRule) method that is used by this package.

We use this package only to create styles for `ThemeProvider` and `FLex` on the fly.

KelvinOm: https://github.com/appsmithorg/appsmith/assets/11555074/fce3426d-edc0-4197-a934-32cc81ba7b14

app/client/packages/design-system/widgets/src/components/Flex/src/flexCssRule.ts
```diff
@@ -0,0 +1,177 @@
...
+  switch (true) {
+    case cssProp === "wrap":
+      return `flex-wrap: ${flexWrapValue(value)};`;
+    case cssProp === "isHidden":
+      return `display: ${hiddenValue(value)};`;
+    case cssProp === "flexAlignValue":
+      return `align-items: ${flexAlignValue(value)};`;
+    case cssProp === "direction":
+      return `flex-direction: ${value};`;
+    case cssProp === "gap" ||
```
jsartisan, line 132: This feels odd. This is nothing but styled-system. Maybe we can just reuse panda? checkout this example - 
https://panda-css.com/docs/concepts/style-props.

jsartisan (approved):
LGTM. left some minor comments.

app/client/packages/design-system/theming/src/theme/src/ThemeProvider.tsx
```diff
@@ -1,35 +1,109 @@
...
+    color,
+    fontFamily,
+    opacity,
+    rootUnit,
+    sizing,
+    spacing,
+    typography,
+    zIndex,
+  } = theme;
+  const providerSheetKey = useRef(uniqueId("wds-provider-"));
```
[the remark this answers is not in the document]
KelvinOm: Here we create a unique id that does not change every rerender. `Ref` is not changed every rendering.

app/client/packages/design-system/theming/src/typography/src/typography.ts
```diff
@@ -14,6 +14,7 @@ export const getTypographyClassName = (key: keyof typeof TypographyVariant) => {
 
 export const createTypographyStringMap = (
   typography: Typography,
+  containerCLassName: string,
```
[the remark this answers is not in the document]
KelvinOm: Yes, thank you! 

Made a note to add https://eslint.org/docs/latest/rules/camelcase

app/client/packages/design-system/widgets/src/components/Flex/src/flexCssRule.ts
```diff
@@ -0,0 +1,177 @@
...
+        containerProps[key],
+      )),
+  );
+
+  return styles;
+};
+
+// the value can be of any type in accordance with flex component props
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const simpleStyles = (cssProp: string, value: any) => {
```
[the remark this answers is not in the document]
KelvinOm: There are two types of styles here, one for the container and the other simple ones not in the form of an object.

`direction={{ base: "column", "520px": "row" }}`
and 
`direction="row"`

app/client/packages/design-system/widgets/src/components/Flex/src/flexCssRule.ts
KelvinOm: Panda does not suit us because there is no possibility to create styles in runtime, and we need this for the provider and flex component. Check this docs about [dynamic styles](https://panda-css.com/docs/guides/dynamic-styling).

PR #26350, opened 2023-08-15
fix: moved jest to devdependencies for appsmithctl and updated dockerfile to exclude it during install

sum35h:
Update:
- Mobed jest to` devDependenies` of package manifest.
- Updated Dockerfile to install utils with `--only=prod` arg for npm install cmds.

Tested manually

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -84,7 +84,7 @@ COPY ./deploy/docker/entrypoint.sh ./deploy/docker/scripts/* info.*json ./
 
 # Add util tools
 COPY ./deploy/docker/utils ./utils
-RUN cd ./utils && npm install && npm install -g .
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -84,7 +84,7 @@ COPY ./deploy/docker/entrypoint.sh ./deploy/docker/scripts/* info.*json ./
 
 # Add util tools
 COPY ./deploy/docker/utils ./utils
-RUN cd ./utils && npm install && npm install -g .
+RUN cd ./utils && npm install --only=prod && npm install --only=prod -g .
```
sharat87, line 87: Let's also change this to `npm ci`, so it uses the pinned versions from `package-lock.json`.

sharat87 (changes requested):
Should we test this with the build-deploy-preview once, just in case?

PR #26358, opened 2023-08-15
chore:update node version to v18

KelvinOm:
## Description
Update node version


app/client/.yarnrc.yml
```diff
@@ -5,5 +5,7 @@ plugins:
     spec: "https://raw.githubusercontent.com/ambar/yarn-plugin-dedupe-on-install/main/index.js"
   - path: .yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs
     spec: "@yarnpkg/plugin-workspace-tools"
+  - path: .yarn/plugins/@yarnpkg/plugin-engines.cjs
```
KelvinOm, line 8: Yarn berry needs a separate plugin to check the engines versions.

https://github.com/yarnpkg/berry/issues/1177

app/client/cypress/package.json
```diff
@@ -3,8 +3,8 @@
   "version": "0.1.0",
   "private": true,
   "engines": {
-    "node": "^16.14.0",
-    "npm": "^8.5.5"
+    "node": "^18.17.1",
```
KelvinOm, line 6: Latest LTS version for this moment.

KelvinOm:
The build with the tests is [here](https://github.com/appsmithorg/appsmith/actions/runs/5866259516).

sharat87:
@KelvinOm @riodeuno, in addition to this, should we also consider changing the `node-version: a.b.c` lines to `node-version-file: app/client/package.json` instead? This will give us two things:

1. The version number is not duplicated a billion times all across the source code.
2. The NodeJS version is taken from the current branch, instead of the `release` branch for `ok-to-test`.

This is supported by [code here](https://github.com/actions/setup-node/blob/bea5baf987ba7aa777a8a0b4ace377a21c45c381/src/util.ts#L10).

I've tested in a separate test repo and it works.

shot-2023-08-15-12-19-13

What do you think?

riodeuno (dismissed):

riodeuno:
@sharat87 Yes, I agree. @KelvinOm do you see any challenges in using this?

KelvinOm:
> @sharat87 Yes, I agree. @KelvinOm do you see any challenges in using this?

I think it's great! Checking...

sharat87:
Oh `ok-to-test` will still use `v16`, since the workflow changes to use `package.json` need to be in the `release` branch. 🙂

KelvinOm:
The build with the tests is [here](https://github.com/appsmithorg/appsmith/actions/runs/5867172488). I think we can not wait for this. Since the last run was green, and here we just fixed the scripts for the

Снимок экрана 2023-08-15 в 15 44 11
 node version that works.

sharat87 (approved):

PR #26366, opened 2023-08-16
feat: Readonly root filesystem support in the Appsmith container

### LINKED ISSUE #26392: [Feature]: Support running with readonly root filesystem

### Summary

Quoting from pentest report:

> The container's root filesystem should be treated as a 'golden image' by using Docker run's --read-only option. This prevents any writes to the container's root filesystem at container runtime and enforces the principle of immutable infrastructure.

Also quoting,

> Enabling this option forces containers at runtime to explicitly define their data writing strategy to persist or not persist their data.
>
> This also reduces security attack vectors since the container instance's filesystem cannot be tampered with or written to unless it has explicit read-write permissions on its filesystem folder and directories.

sharat87:
Fixes #26392.


sharat87:
reminder: get Keycloak to also use the new truststore, when there's custom CA certs to trust.
verify: test SAML, when Appsmith is exposed with HTTPS, with cert signed by custom CA.

pratapaprasanna:
Duplicate of #28290

PR #26429, opened 2023-08-17
ci: Include info.json for all workflows

sharat87:


nidhi-nair (approved):
LGTM, just pointing out that EE will need airgap workflow changes as well! :)

PR #26431, opened 2023-08-17
chore: Remove unused, misleading response to PUT /env APIs

sharat87:
This response object to `PUT /env` API appears to indicate whether restart is required or not, because it includes the field `isRestartRequired`, but this is actually not used at all on the client. We might've used it in the past, but we don't anymore. This is a misleading response DTO, and it's already wasted my time thrice! 😭


app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/UserSignupCEImpl.java
```diff
@@ -279,17 +278,20 @@ public Mono<User> signupAndLoginSuper(UserSignupRequestDTO userFromRequest, Serv
...
                                     APPSMITH_DISABLE_TELEMETRY.name(),
                                     String.valueOf(!userFromRequest.isAllowCollectingAnonymousData()),
                                     APPSMITH_ADMIN_EMAILS.name(),
                                     user.getEmail()))
+                            // We need a non-empty value for `.elapsed` to work.
+                            .thenReturn(true)
                             .elapsed()
                             .map(pair -> {
                                 log.debug("UserSignupCEImpl::Time taken to apply env changes: {} ms", pair.getT1());
                                 return pair.getT2();
```
nidhi-nair, line 292: Unrelated, but why are we logging this?

nidhi-nair (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/UserSignupCEImpl.java
sharat87: No idea. I wanted to ask and remove that too, but not to change much in a small chore PR... I didn't. 🤔

PR #26433, opened 2023-08-17
chore: Fix warnings from server build's shade plugin

sharat87:
Reduces warnings by:

```
build-this-pr | grep -F '[WARNING] ' | wc -l
    86

build-release | grep -F '[WARNING] ' | wc -l
    442
```

These warnings are irrelevant, and are flooding the build output, causing us to miss on those warnings that could have significance.

Here's an example warning of this kind:

```
[WARNING] HikariCP-5.0.1.jar, mockito-inline-4.4.0.jar, ojdbc8-21.9.0.0.jar, oraclePlugin-1.0-SNAPSHOT.jar define 1 overlapping resource: 
[WARNING]   - META-INF/MANIFEST.MF
[WARNING] maven-shade-plugin has detected that some class files are
[WARNING] present in two or more JARs. When this happens, only one
[WARNING] single version of the class is copied to the uber jar.
[WARNING] Usually this is not harmful and you can skip these warnings,
[WARNING] otherwise try to manually exclude artifacts based on
[WARNING] mvn dependency:tree -Ddetail=true and the above output.
[WARNING] See https://maven.apache.org/plugins/maven-shade-plugin/
```


sharat87:
Based on above failures, reverted the fix for MSSQL, Redshift and Firestore.

PR #26490, opened 2023-08-18
fix: Update refresh time for tenant level feature flags to avoid race condition

abhvsn:
## Description
The tenant level feature flags are pulled from Flagsmith from a cron job scheduled to run after every couple of hours. Also the refresh rate is set to exact same time as that of cron execution interval i.e. 120mins. This is creating a race condition because of which the latest feature flag may not get refreshed after each 2hrs but may take upto 4hrs i.e. next execution cycle. With current PR we are making sure the `cron execution interval > refresh rate`.


NilanshBansal (approved):
LGTM

PR #26494, opened 2023-08-21
chore: Disallow authenticated users from accessing internally controlled endpoints

nidhi-nair:
This change makes sure that an authenticated user is also not allowed to access internal endpoints.


sharat87 (approved):
Works. Just want to point out that when logged in, we don't get the basic-auth request at all. No harm done, just wanted to document that difference here, for the future.

shot-2023-08-21-12-55-45

PR #26527, opened 2023-08-21
ci: Update ce push workflow to use dime defenders

sarojsarab:
## Description
- Update ce push workflow to use dime defenders
#### Type of change
- Workflow changes
- [ ] Workflow run
>
>


yatinappsmith:
> @sarojsarab - Do we need to update CE push workflows also? For CE we have some more time right? cc: @yatinappsmith



> @sarojsarab - Do we need to update CE push workflows also? For CE we have some more time right? cc: @yatinappsmith

@Aishwarya-U-R We will move to push workflow. This will help us test the new solution

PR #26614, opened 2023-08-24
update: disable postgres by default in helm

pratapaprasanna:


deploy/helm/templates/statefulset.yaml
```diff
@@ -103,6 +103,9 @@ spec:
           volumeMounts:
             - name: data
               mountPath: /appsmith-stacks
+          env:
+            - name: APPSMITH_ENABLE_EMBEDDED_DB
+              value: {{ .Values.applicationConfig.APPSMITH_ENABLE_EMBEDDED_DB | default "0" | quote }}
```
sharat87, line 108: This means that if `APPSMITH_ENABLE_EMBEDDED_DB` is set in `applicationConfig`, then embedded Postgres will run, right? Do we want that? Can we just hard code this value to just `"0"`?

```suggestion
              value: "0"
```

PR #26616, opened 2023-08-24
ci: cypress and perf tests on documentDB

sum35h:
Runs cypress and perf tests on an Appsmith instance with an existing DocumentDB backend instead of MongoDB.

Comment to trigger the run on a PR:
`/ok-to-test-with-documentdb`


.github/workflows/ci-test-with-documentdb.yml
```diff
@@ -0,0 +1,564 @@
...
+    # Service containers to run with this job. Required for running tests
+    services:
+      # Label used to access the service container
+      redis:
+        # Docker Hub image for Redis
+        image: redis
+        ports:
+          # Opens tcp port 6379 on the host and service container
+          - 6379:6379
+      mongo:
```
nidhi-nair, line 38: We don't need this right?

sum35h: Yes.
And it looks like neither the redis or mongo service is used by the other run, right? The appsmith still used the embedded redis and mongo since there is no env configs passed to it.

.github/workflows/ci-test-with-documentdb.yml
```diff
@@ -0,0 +1,564 @@
...
+  workflow_dispatch:
+  workflow_call:
+    inputs:
+      pr:
+        description: "This is the PR number in case the workflow is being called in a pull request"
+        required: false
+        type: number
+
+jobs:
+  ci-test:
```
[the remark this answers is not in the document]
sum35h: The db is not getting cleaned before, I'm using a fresh db.
No shapshots during init as well.

.github/workflows/ci-test-with-documentdb.yml
sum35h: The services is only applicable on the host level right, so appsmith containers are not aware of them, so it uses the embeded dbs.

nidhi-nair (dismissed):
LGTM, kinda difficult to parse through as it is, but the mongo part looks good.

PR #26623, opened 2023-08-24
feat: Add isSuggested value to templates for all plugins

### LINKED ISSUE #26624: [Feature]: Add suggestions to templates

### Summary

This PR adds suggestions to templates allowing us to provide users with guidance on what template they can quickly get started with.

Screenshot 2023-08-24 at 10 07 08

This presents a better experience for users.

Irongade:
This PR adds an isSuggested value to templates for all plugins that support DatasourceStructure

Fixes #26624 

>
>
>
- [ ] Manual
- [x] JUnit
- [ ] Jest
- [ ] Cypress
>
>
#### Dev activity

#### QA activity:


nidhi-nair:
@Irongade could we please attach the issue that we are fixing with this change? It helps for anyone outside of the immediate team working on this to gain context of what's up even when they look back at this PR in the future.

Irongade:
> @Irongade could we please attach the issue that we are fixing with this change? It helps for anyone outside of the immediate team working on this to gain context of what's up even when they look back at this PR in the future.

Sure, I just created one and I've attached it.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/DatasourceStructure.java
```diff
@@ -114,37 +114,50 @@ public static class Template {
...
+        boolean isSuggested;
 
         // To create templates for plugins which store the configurations
         // in List<Property> format
-        public Template(String title, String body, List<Property> configuration) {
+        public Template(String title, String body, List<Property> configuration, boolean isSuggested) {
             this.title = title;
             this.body = body;
             this.configuration = configuration;
+            if (isSuggested) {
```
sumitsum, line 125: Hi Ade, QQ: could we remove the `if` condition here ?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/DatasourceStructure.java
```diff
@@ -114,37 +114,50 @@ public static class Template {
...
         }
 
         // To create templates for plugins with UQI framework which store the configurations
         // as a map
-        public Template(String title, String body, Map<String, ?> configuration) {
+        public Template(String title, String body, Map<String, ?> configuration, boolean isSuggested) {
             this.title = title;
             this.body = body;
             this.configuration = configuration;
+            if (isSuggested) {
```
sumitsum, line 136: Could we remove the `if` condition here ? 
Same for all the other constructors.

sumitsum (changes requested):

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/DatasourceStructure.java
```diff
@@ -114,37 +114,50 @@ public static class Template {
         String title;
         String body;
         Object configuration;
+        boolean isSuggested;
```
sumitsum, line 117: I think we also save query templates in Mongo DB. For all those templates this field does not exist, so when they get loaded in the API server memory and sent to the client this field should show up as null - is the client side designed to handle null value ? 
Have you been able to test this feature with a datasource which had structure fetched beforehand and then the instance got udated with this feature ?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/DatasourceStructure.java
Irongade: The reason why I added an if condition here is because, this field should only be set for suggested templates only. For other templates, it should not exist.

sumitsum (approved):

PR #26630, opened 2023-08-24
fix: Check if chart data in props is valid or not

### LINKED ISSUE #26325: TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
Sentry Issue: [APPSMITH-6SE](https://appsmith.sentry.io/issues/4382323470/?referrer=github_integration)

```
TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
  at forEach (widgets/ChartWidget/component/EChartsDatasetBuilder.ts:26:42)
  at Array.forEach (<anonymous>)
  at <object>.value (widgets/ChartWidget/component/EChartsDatasetBuilder.ts:22:32)
  at _this.getEChartsOptions (widgets/ChartWidget/component/index.tsx:136:34)
  at _this.renderECharts (widgets/ChartWidget/component/index.tsx:195:35)
...
(15 additional frame(s) were not displayed)
```

### LINKED ISSUE #26548: TypeError: a.replace is not a function
Sentry Issue: [APPSMITH-6W3](https://appsmith.sentry.io/issues/4408929755/?referrer=github_integration)

```
TypeError: a.replace is not a function
  at m (../node_modules/fusioncharts/fusioncharts.js:13:176102)
  at a.draw (../node_modules/fusioncharts/fusioncharts.widgets.js:1:64053)
  at t.syncDraw (../node_modules/fusioncharts/fusioncharts.js:13:202559)
  at s (../node_modules/fusioncharts/fusioncharts.js:13:198841)
  at f (../node_modules/fusioncharts/fusioncharts.js:13:199127)
...
(7 additional frame(s) were not displayed)
```

rajatagrawal:
Fixes #26325
Fixes #26548

### DIFF
--- a/app/client/src/widgets/ChartWidget/widget/index.tsx
+++ b/app/client/src/widgets/ChartWidget/widget/index.tsx
@@ -32,10 +32,7 @@ export const isChartDataValid = (props: ChartWidgetProps): boolean => {
       for (const seriesID in props.chartData) {
         const seriesData = props.chartData[seriesID];
 
-        if (!seriesData) {
-          return false;
-        }
-        if (!seriesData.data) {
+        if (!seriesData || !seriesData.data) {
           return false;
         }
       }
--- a/app/client/src/widgets/ChartWidget/widget/index.tsx
+++ b/app/client/src/widgets/ChartWidget/widget/index.tsx
@@ -17,10 +17,32 @@ import type { AutocompletionDefinitions } from "widgets/constants";
 import { ChartErrorComponent } from "../component/ChartErrorComponent";
 import { syntaxErrorsFromProps } from "./SyntaxErrorsEvaluation";
 import { EmptyChartData } from "../component/EmptyChartData";
+import * as Sentry from "@sentry/react";
 
 const ChartComponent = lazy(() =>
   retryPromise(() => import(/* webpackChunkName: "charts" */ "../component")),
 );
+
+export const isChartDataValid = (props: ChartWidgetProps): boolean => {
+  switch (props.chartType) {
+    case "CUSTOM_FUSION_CHART": {
+      return !!props.customFusionChartConfig;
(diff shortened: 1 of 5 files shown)


app/client/src/widgets/ChartWidget/widget/index.tsx
```diff
@@ -17,11 +17,33 @@ import type { AutocompletionDefinitions } from "widgets/constants";
...
+export const isChartDataValid = (props: ChartWidgetProps): boolean => {
+  switch (props.chartType) {
+    case "CUSTOM_FUSION_CHART": {
+      return !!props.customFusionChartConfig;
+    }
+    default: {
+      for (const seriesID in props.chartData) {
+        const seriesData = props.chartData[seriesID];
+
+        if (!seriesData) {
```
sbalaji1192, line 35: combine these two if condition

sbalaji1192:
Why do we need to perform this check on the widget, Can we do this in the validation itself?

rajatagrawal:
> Why do we need to perform this check on the widget, Can we do this in the validation itself?

@sbalaji1192 

We have a check in the validation. But I don't why we are still getting undefined data. Last time this issue was fixed, I added checks in the validation only. The issue has regressed even then.

app/client/src/widgets/ChartWidget/widget/index.tsx
rajatagrawal: Done

sbalaji1192:
@rajatagrawal Let's figure out the root cause and fix this issue on the validation itself. From a widget perspective, it should always receive valid values for the properties defined in the property pane. We can ensure this by writing proper validation functions.
Ideally, the widget should not be firing sentry errors. We need to modify log.error to report the error to sentry as part of diff PR.

PR #26657, opened 2023-08-25
chore: Add support for 3D Custom ECharts

### LINKED ISSUE #26742: [Feature]: Add support for 3D Custom ECharts

### Summary

Custom EChart users should be able to render 3D ECharts as well and toggle between 2D and 3D charts seamlessly by changing the chart configuration.

Dev notes: If it is a 3D chart, we will use a `canvas` renderer for ECharts else a `svg` renderer for ECharts. 3D charts only work with canvas renderer.

Users will expect that all examples mentioned on ECharts example website would work in Appsmith.

rajatagrawal:
Fixes #26742


laveena-en:
@rajatagrawal When the browser is idle for a while, the chart data is disappearing 
image

Also, it looks like the chart is getting re-rendered every time there is any change in the canvas. Notice that the zoom state changes and there is data points seem to be reloaded https://www.loom.com/share/94ee9ebfe4e54ad8a92044d7b1dd5431?sid=67171e5b-66e5-4b44-a67d-90d34bb446cb

app/client/src/widgets/ChartWidget/component/helpers.ts
```diff
@@ -60,3 +60,55 @@ export const parseOnDataPointClickForBasicCharts = (
...
+    "globe",
+    "geo3D",
+    "mapbox3D",
+    "grid3D",
+    "xAxis3D",
+    "yAxis3D",
+    "zAxis3D",
+  ];
+
+  const keys = Object.keys(chartConfig);
```
sbalaji1192, line 75: nit: Can you name this variable as chartConfigKeys. It will be more readeable

app/client/src/widgets/ChartWidget/component/helpers.ts
```diff
@@ -60,3 +60,55 @@ export const parseOnDataPointClickForBasicCharts = (
...
+
+  for (const key of keys) {
```
sbalaji1192, line 77: nit: 
```suggestion
  if (keys.some(key => threeDConfigs.includes(key))) {
    return true;
  }
```

app/client/src/widgets/ChartWidget/component/helpers.ts
```diff
@@ -60,3 +60,55 @@ export const parseOnDataPointClickForBasicCharts = (
...
+    if (threeDConfigs.includes(key)) {
+      return true;
+    }
+  }
+
+  const seriesConfig = chartConfig.series;
+  if (Array.isArray(seriesConfig)) {
+    const filtered3DSeriesTypes = seriesConfig.filter((series) => {
```
sbalaji1192, line 85:
```suggestion
    return seriesConfig.some(config => isSeriesConfig3D(config))
```

app/client/src/widgets/ChartWidget/component/helpers.ts
```diff
@@ -60,3 +60,55 @@ export const parseOnDataPointClickForBasicCharts = (
     seriesTitle: seriesName,
   } as ChartSelectedDataPoint;
 };
+
+export const is3DChart = (chartConfig: Record<string, unknown>) => {
+  const threeDConfigs = [
```
sbalaji1192, line 65: can you move this to constant file?

app/client/src/widgets/ChartWidget/component/helpers.ts
```diff
@@ -60,3 +60,55 @@ export const parseOnDataPointClickForBasicCharts = (
...
+      return isSeriesConfig3D(series);
+    });
+    return filtered3DSeriesTypes.length > 0;
+  } else {
+    return isSeriesConfig3D(seriesConfig);
+  }
+};
+
+const isSeriesConfig3D = (seriesConfig: unknown) => {
+  const threeDSeriesTypes = [
```
sbalaji1192, line 95: can you move this to constant file?

app/client/src/widgets/ChartWidget/component/helpers.ts
```diff
@@ -60,3 +60,55 @@ export const parseOnDataPointClickForBasicCharts = (
...
+    "lines3D",
+    "map3D",
+    "surface",
+    "polygons3D",
+    "scatterGL",
+    "graphGL",
+    "flowGL",
+  ];
+
+  if (typeof seriesConfig == "object") {
```
sbalaji1192, line 108: This condition is true when seriesConfig is null
Use lodash isPlainObject function or conditional look up.

app/client/src/widgets/ChartWidget/component/helpers.ts
```diff
@@ -60,3 +60,55 @@ export const parseOnDataPointClickForBasicCharts = (
...
+    const seriesType = (seriesConfig as Record<string, unknown>).type as string;
```
sbalaji1192, line 109:
```suggestion
    seriesConfig?.type && threeDSeriesTypes.includes( seriesConfig.type)
```

app/client/src/widgets/ChartWidget/component/index.tsx
```diff
@@ -1,6 +1,7 @@
 import React from "react";
 import styled from "styled-components";
 import * as echarts from "echarts";
+import "echarts-gl";
```
sbalaji1192, line 4: Are we loading this for all users? Can we import this only if user is using 3d charts?

rajatagrawal:
> Also, it looks like the chart is getting re-rendered every time there is any change in the canvas. Notice that the zoom state changes and there is data points seem to be reloaded

@marks0351 @riodeuno is this something that can be addressed from canvas layer ? Expected behavior is that the chart widget should not receive an extra props update if there are no changes to this widget.

I also noticed that the chart widget doesn't receive an extra prop update if only the horizontal position of the other widget is changed instead of both horizontal and vertical position.

Relevant PR comment: https://github.com/appsmithorg/appsmith/pull/26657#issuecomment-1711057313

app/client/src/widgets/ChartWidget/component/index.tsx
rajatagrawal: The loading time of this library is very less. For code simplicity, I thought of loading the library for all users since the loading time of this library is very less.

If we do a conditional loading of the library, then we will also need to make all chart functions async because dynamic library import function is async.

app/client/src/widgets/ChartWidget/component/helpers.ts
rajatagrawal: Done

app/client/src/widgets/ChartWidget/component/helpers.ts
rajatagrawal: Without the `Record<string, unknown>` typecast, the compiler gives an error that "Property type doesn't exist on type 'object'". It is happening because seriesConfig is of type unknown in the function arguments.

I also checked that threeDSeriesTypes.includes(null/undefined) returns false and doesn't throw an exception

app/client/src/widgets/ChartWidget/component/helpers.ts
rajatagrawal: Done

app/client/src/widgets/ChartWidget/component/helpers.ts
rajatagrawal: Done

app/client/src/widgets/ChartWidget/component/helpers.ts
rajatagrawal: Added a unit test case for null `seriesConfig.type` condition

app/client/src/widgets/ChartWidget/component/helpers.ts
rajatagrawal: Done

app/client/src/widgets/ChartWidget/component/helpers.ts
rajatagrawal: Done

app/client/src/widgets/ChartWidget/component/helpers.ts
rajatagrawal: Done

rajatagrawal:
@laveena-en 

The bug with respect to chart disappearing after sometime is a bug in the library that we are using for rendering 3D ECharts. This issue has also been noticed in the 3D Chart example on ECharts website also. Can you please confirm if you can repro chart disappearing issue on the ECharts website ?

laveena-en:
@rajatagrawal The widget isn’t reflecting any changes when we switch from Custom 3D E-Chart to any other type. Please do take a look: https://www.loom.com/share/375f2b08f93e4951a332d4a49dc89074?sid=b8118cae-277b-46c9-aa11-8a0b2e57e79d

app/client/src/widgets/ChartWidget/component/index.tsx
sbalaji1192: Let's consider async loading even if the loading time is less. async loading small packages like this, here and there, will add up and save us some loading time eventually.

sbalaji1192 (commented):
Can you look into this comment - https://github.com/appsmithorg/appsmith/pull/26657#discussion_r1322303884

app/client/src/widgets/ChartWidget/component/index.tsx
rajatagrawal: @sbalaji1192 

Yes, will work on this.

laveena-en:
@rajatagrawal The issue with 3D charts not being rendered on switching pages/charts is still persisting: https://www.loom.com/share/a2e5f6ef7e5d4baaaf9a402c9fb3231a?sid=1e83f8ca-0c9f-478b-aaea-08b74be9000e

Can you please take a look?

app/client/src/widgets/ChartWidget/component/index.tsx
rajatagrawal: @sbalaji1192 

I noticed that the unit test suite becomes unstable and we are getting many flaky tests if the chart widget componentDidUpdate is made async.

I would need to revert the changes because the 

1. The tests are running fine individually but when the test suite is run as a whole, it is giving various multithreading issues.

laveena-en:
Tested and verified the changes.

Known issues:

[Chart being re-rendered](https://github.com/appsmithorg/appsmith/issues/27591)
[Eval issue discussion with FE coders](https://theappsmith.slack.com/archives/C02K0SZQ7V3/p1695368024762839)
Chart data disappears when browser is idle. Issue raised with e-charts GL [here](https://github.com/ecomfe/echarts-gl/issues/506)

app/client/src/widgets/ChartWidget/component/index.tsx
sbalaji1192: @rajatagrawal we should look into fixing the unit test suite issues, rather than just adding this library to the bundle? If this is not doable with component unit testing, we can leverage cypress to test those cases.

rajatagrawal: @sbalaji1192 

The whole unit test suite becomes unstable because the test suite runs in a multi threaded environment. Even if we do cypress tests, the existing unit test become flaky that would need to be fixed anyway because of async functions in chart component.

The library is already included in the bundle and we have to decide to load it or not. I had done a benchmark for the loading times of GL library and noticed that it only takes 0.1 miliseconds to load the GL library. [Relevant Slack thread](https://theappsmith.slack.com/archives/C05M865KG5A/p1693473428738249)

Given that the time take to load this library is very less, it is not adding a very significant overload to the performance.

sbalaji1192: Lets create an issue and add a todo for this , we can look into this later

sbalaji1192 (approved):

PR #26673, opened 2023-08-28
test: Cypress - button group cases 

sharanya-appsmith:
- Add, Delete more buttons to the group and verify
-  Add, Delete menu item to a button group and verify
- Assert button group options - single and menu visibility
- Assert button group options - single and menu disability
- Assert On click button group


NandanAnantharamu (approved):
LGTM see if you can add a check for preview mode in one of the files

PR #26752, opened 2023-08-29
chore: skip signposting for other tests

akash-codemonk:


app/client/cypress/support/Pages/HomePage.ts
```diff
@@ -252,7 +252,7 @@ export class HomePage {
       this.agHelper.AssertElementVisibility(
         this.entityExplorer._entityExplorer,
       );
-      this.onboarding.closeIntroModal();
+      this.onboarding.skipSignposting();
```
Aishwarya-U-R, line 255: Do we need to skip in all cases?

akash-codemonk: Yup. For all tests it will skip.
For the onboarding feature itself I signup separately https://github.com/appsmithorg/appsmith/blob/release/app/client/cypress/e2e/Regression/ClientSide/Onboarding/FirstTimeUserOnboarding_spec.js#L17 and do the assertions.

Aishwarya-U-R: Ok, Got it.

Aishwarya-U-R (approved):

PR #26795, opened 2023-08-30
chore: Drag and drop logic separation

rahulramesha:
## Description
This PR is to separate the Drag n drop logic for auto layout and fixed layout. So that both logics can exist independently


rahulramesha:
Closing this in favour of https://github.com/appsmithorg/appsmith/pull/26838

PR #26809, opened 2023-08-30
chore(deps): Update certbot from v0.40 to v2.6

sharat87:
⚠️ **This will cause conflicts.**


sharat87:
Manually tested automatic cert provisioning, as well as renewal. Both worked fine.

mohanarpit (approved):

PR #26843, opened 2023-08-31
chore: PoC for skipping Flaky tests on CI

aswathkk:
Trying out a way to skip flaky test run on CI

- Added `app/client/cypress/mockFlakyTestListingAPI.json` as a way to mock an API response
  - We should replace it with a proper API before merging this PR
  - Ideally this API endpoint should be provided from env
- Modified cypress-split.ts to ignore the flaky tests mentioned in the API response


app/client/cypress/mockFlakyTestListingAPI.json
```diff
@@ -0,0 +1,7 @@
+{
+  "flakyTests": [
+    "cypress/e2e/Regression/ClientSide/Binding/ChartText_spec.js",
```
nidhi-nair, line 3: Asking from a place of naivete, do we usually have to skip entire test specs, or is it particular tests that are failing?

app/client/cypress/cypress-split.ts
```diff
@@ -97,7 +97,21 @@ async function getSpecsToRun(
         totalRunnersCount,
         currentRunner,
       );
-      return specsToRun.join(",");
+
+      // TODO(aswathkk): Replace this with an API call to get the list of flaky tests
+      const flakyTestsResponse = await fetch(
+        "https://raw.githubusercontent.com/appsmithorg/appsmith/d683514c5d564ecb8928e01c37f26c6a83391912/app/client/cypress/mockFlakyTestListingAPI.json",
```
nidhi-nair, line 103: Do you think we can use the flaky tests app to get to this list?

aswathkk: I checked that. It is fetching the data from a Google sheet.
I think we should move it to a DB instead and expose it via an API which can be used in the App as well as in CI

app/client/cypress/mockFlakyTestListingAPI.json
aswathkk: Most of our tests are written in a way that a test is dependent on the previous test with in a spec. So, if a test in a spec fails mostly the next set of tests fail too.
Unless we write tests within a spec independent of each other, we should skip the entire spec.

app/client/cypress/cypress-split.ts
nidhi-nair: How about a local file? That would allow us to skip tests independently in different branches, based on where the fix is. A global list might put QA folks in trouble when they're actually trying to fix the test itself. This asks for an extra commit, but isolates tests by branch.

aswathkk: A local file doesn't add any value IMO. Cypress itself provides a functionality by adding `.skip` to test or spec itself.
The whole point of this exercise is to centralise skipping flaky tests.

Let me think about this case. I will respond later

aswathkk: In the case of fixing the flaky test itself, team can make use of `ci-test-limit` command which will help them to execute only specified specs.

Issue #21993, opened 2023-04-02
[Feature]: Add options for labels in select columns in a table widget


GreenFlux:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

The table widget now supports the `Select` column type, which allows the column to contain both a label and a value. This could be useful for currency fields, foreign keys, or any other case where you want to display a different version of the column value. However, there is a problem with sorting. 

With Select columns in a table widget, the `label` is shown to the user while selecting an option, but then the `value` is shown in the table widget, after the option has been selected. This is different from the behavior of a Select widget outside a table, _which continues to show the label_ after an option has been selected. 

**Feature Request:**
To address this inconsistent behavior, and add support for sorting currency values, we should allow the user to choose:
1. **Display as?** `[label | value ]`
2. **Sort by?** `[label | value ]`

### Why should this be worked on?

Multiple users have asked for a way to support a currency column in the table widget that both sorts and displays accurately. For now, users have to choose between displaying as a number (that sorts property) or as text with currency symbol (which sorts incorrectly). 

https://github.com/appsmithorg/appsmith/issues/5632
https://discord.com/channels/725602949748752515/1076320585002532976
https://discord.com/channels/725602949748752515/1074651219588558868

<img src="https://front.com/assets/img/favicons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_a2ioc)

AS-Laguna:
It would be useful if it would still use the `value` when interacting with table data programmatically, even if DisplayAs is set to `label`.

somangshu:
@keyurparalkar @momcilo-appsmith should this be prioritized?

benjamindonze:
More generally the select widget is not consistent regarding label / value. 
Let say I have a DB table car and a DB table color. Car has a FK to table color.
I want to create a table to edit cars. My table widget is populated with query:
`SELECT id, brand, model, color_id FROM car;`
with inline editing configured as https://app.appsmith.com/app/editable-table/cell-save-62d8f8d0e1c2ed505a0557cd. 
In the color column I set the widget type to select. 
I let value of properties **Computed Value** to `{{currentRow.color_id}}` and set the **Select properties options** with data of this query:
`SELECT id as value, name as label FROM color;` 
then here is current behavior:
- When displaying the table the column color shows the FK id. **NOK** Would expect to show the label of the color.
[...]

PR #26855, opened 2023-09-01
fix: Add Display as property to the select column in Table Widget

### LINKED ISSUE #26188: [Bug]: Select widget should display label instead of value while editing a table cell

### Description

Whenever a select column type cell is editable and when user actually selects the value from drop down for the cell, then select option's value property is updated in cell's data. It is expected that select option's label property should get updated in cell data.

### Steps To Reproduce

1. Drag and drop a table widget and bind a query to this widget.
2. Change one of it's column to select type and add select options to it.
3. Make this column editable
4. Now try to edit this cell
5. Observe that selecting an option from the select dropdown sets the cell's data to the `value` property of select option instead of `label`

### Public Sample App

_No response_

### Environment

Production

### Issue video log

_No response_

### Version

Cloud
```[tasklist]
### Tasks
- [ ] https://github.com/appsmithorg/appsmith/issues/27061
- [ ] https://github.com/appsmithorg/appsmith/issues/27093
```

Front logo [Front conversations](https://app.frontapp.com/open/top_3kid3g)

keyurparalkar:
## Description
This PR introduces a new property called `Display As` for the select column type in table widget so that it fixes the bug mentioned below. This change has the following expectations:
* Whenever a column is set to select, by default we show labels and not values.
* For existing tables, since we have created this problem we ensure we don't break their apps. We will give an option called "Display As" as suggested in the original issue and for new select columns, it will be Label and for existing columns, it will be Value.
* We make it clear that the computed value here refers to the default value/display value. Displaying this as the helper text below the computed value property only for select column type.
* Added a helper text below the computed value field that indicates that the computed value here is a default value/display value
* Added select option field validation. The validation is as such: When computed value is not present in the select option then we throw an error in select option. This is irrespective of displayAs property.
* For new row select options, there is no validation of computed value. Since are adding new data into the table and it has to adhere to the select options that we provide.
* For displaying value in cell,
    - When display as is set to label, we show empty values for missing labels in the widget
    - When display as is set to value, we show the computed value as it is even if labels are missing
* When computed value is not present in the select options, we still make the select options available. We do this because its not the select options that has issue but its computed value.
* Sorting, filtering and searching works on the data that is present in the table.  For select column type, we store select option's value in the table. Hence all these operations happens on select option's value property.

Fixes #26188

## Testing
- [ ] Manual
    - For new added row, there are no validations of computed value
    - Check if the sorting, filtering and searching is happening based on the select cell's value property.
- [ ] JUnit
- [ ] Jest
    - Added unit test that checks if `selectDisplayAs = value` for migrated select table columns 
- [ ] Cypress
    - Check if the property displayAs is available whenever column type is select.
    - Check if the computed value helperText is present.
    - Check that while editing a cell in displayAs = label, the select's value is set in the table's data rather than the actual label
    - Check that while editing a cell in displayAs = value, the select's value appears in the table's data always.
    - Check that select options value are being present on table when displayAs = value
    - Check that select options label are being present for that respective value when displayAs = label
    - When computed value is not present in the select options then we show the error in the select option field with the row number where the error has occured
    - When computed value is not present in the select options and displayAs = label, then table cell should show empty cell.
    - When computed value is not present in the select options and displayAs = value, then table cell should have that select' value property.
    - When display as is set to value, we show the computed value as it is even if labels are missing
    - When computed value is not select options, we show error in computed value but still the parsed=select options
>
>
#### Dev activity

#### QA activity:


app/client/src/widgets/TableWidgetV2/widget/propertyConfig/PanelConfig/Data.ts
```diff
@@ -146,6 +149,13 @@ export default {
     {
       helpText:
         "The value computed & shown in each cell. Use {{currentRow}} to reference each row in the table. This property is not accessible outside the column settings.",
+      helperText: (props: WidgetProps, propertName: string) => {
```
sbalaji1192, line 152: change the name to propertyPath

app/client/src/widgets/TableWidgetV2/component/cellComponents/SelectCell.tsx
```diff
@@ -189,6 +194,20 @@ export const SelectCell = (props: SelectProps) => {
     .map((d: DropdownOption) => d.value)
     .indexOf(value);
 
+  let baseCellValueFromSelectOptions: string | number | undefined = value;
+
+  if (displayAs === SelectOptionAccessor.LABEL) {
+    const filteredOptions = options.filter(
```
sbalaji1192, line 200: use indexOf instead of filter

app/client/src/widgets/TableWidgetV2/component/cellComponents/SelectCell.tsx
```diff
@@ -189,6 +194,20 @@ export const SelectCell = (props: SelectProps) => {
...
+      (item) => item[SelectOptionAccessor.VALUE] === value,
+    );
+    if (filteredOptions.length > 0) {
+      baseCellValueFromSelectOptions =
+        filteredOptions[0][SelectOptionAccessor.LABEL];
+    } else {
+      baseCellValueFromSelectOptions = "";
```
sbalaji1192, line 207: we should show value instead of empty string

app/client/src/widgets/TableWidgetV2/component/cellComponents/SelectCell.tsx
```diff
@@ -146,16 +150,17 @@ export const SelectCell = (props: SelectProps) => {
...
       onItemSelect(
-        option.value || "",
+        option?.value || "",
         rowIndex,
         alias,
         onOptionSelectActionString,
       );
     },
-    [onItemSelect, rowIndex, alias, onOptionSelectActionString],
+    [displayAs, onItemSelect, rowIndex, alias, onOptionSelectActionString],
```
sbalaji1192, line 163: we dont need displayAs

app/client/src/widgets/TableWidgetV2/component/cellComponents/SelectCell.tsx
```diff
@@ -146,16 +150,17 @@ export const SelectCell = (props: SelectProps) => {
     verticalAlignment,
     width,
   } = props;
+  const displayAs = selectDisplayAs?.toLowerCase();
```
sbalaji1192, line 153: why do we need to do this?

app/client/src/widgets/TableWidgetV2/widget/propertyConfig/PanelConfig/Select.ts
```diff
@@ -95,6 +97,31 @@ export default {
         }
       },
     },
+    {
+      propertyName: "selectDisplayAs",
```
sbalaji1192, line 101: need to write an updateHook on column type property to set default value for this

app/client/src/widgets/TableWidgetV2/component/cellComponents/SelectCell.tsx
keyurparalkar: As discussed this is an expected behaviour.
CC: @dilippitchika @sbalaji1192

sapusi:
Hey @keyurparalkar and @sbalaji1192 , 

is this PR ready for productive deployment and can it be merged into release? 

Can you please launch that process to close #21993 and #26188 ?

Thank you !

PR #26888, opened 2023-09-01
feat: login rate limit

shubham7saxena7:
Changes

add rate limit on login and signup APIs
add annotations to support rate limit on controllers which can be configured per API.
refactor SecurityConfig
implementation details

uses bucket4j for rate limiting
uses redis as a backend for distributed rate limiting

fixes: https://github.com/appsmithorg/appsmith-ee/issues/234


app/server/appsmith-server/src/main/java/com/appsmith/server/ratelimiting/RateLimitConfig.java
sharat87, line 1: I'm not very familiar with Bucket4j, so sorry if this question is inaccurate, but it looks like we're using sync APIs here. Instead of staying in reactive async code. Is this intentional? Can we switch to using reactive/async APIs for rate-limiting as well please?

sharat87 (commented):
I'll let @trishaanand put a final approval on the implementation, but I've put in a couple of questions I wanted you folks' thoughts on. Thanks!

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/RedisConfig.java
```diff
@@ -100,6 +107,41 @@ public ReactiveRedisConnectionFactory reactiveRedisConnectionFactory() {
...
+            redisClusterClient.setOptions(ClusterClientOptions.builder()
+                    .timeoutOptions(TimeoutOptions.builder()
+                            .timeoutCommands(true)
+                            .fixedTimeout(Duration.ofMillis(2000))
+                            .build())
+                    .build());
+            return redisClusterClient;
+        }
+
+        RedisClient redisClient = RedisClient.create(redisURL);
```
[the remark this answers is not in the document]
shubham7saxena7: we are not establishing a new connection - we are initialising a client which can spin up new connections in the bucket4j library on demand

We're doing this in reactiveRedisConnectionFactory method already, right? true to an extent but that's the connection factory which can give you a new connection on demand. 

the problem here is that libraries which deal with a lot of connections such as bucket4j want to manage the connections and not be given from outside a single connection. had it accepted a connection factory, it would have made sense. 

If we add a support for a new scheme tomorrow, like redis-sentinel:, we have to remember to do it here as well, right? not necessarily, only the redis-cluster scheme is not supported by lettuce core, for everything else the code needs no special handling

app/server/appsmith-server/src/main/java/com/appsmith/server/ratelimiting/RateLimitConfig.java
shubham7saxena7: oh you mean to say, instead of using Domain objects directly, we wrap them into reactive paradigms ? but this is a config object like redis beans of which we intialised at startup time so we do not need to plan for asynchronous management of these. if the comment is about rate limit service, how we call the underlying library, I am not sure what you mean and how to do it.

trishaanand (dismissed):

shubham7saxena7:
build-deploy-preview skip-tests=true

PR #26895, opened 2023-09-03
chore: airtable spec skip

ayushpahwa:
Skipping airtable spec


app/client/cypress_ci_custom.config.ts
```diff
@@ -54,6 +54,7 @@ export default defineConfig({
       "cypress/e2e/**/spec_utility.ts",
       "cypress/e2e/GSheet/**/**/*",
       "cypress/e2e/Sanity/Datasources/Airtable_Basic_Spec.ts",
+      "cypress/e2e/ME_airtable_spec.ts",
```
Aishwarya-U-R, line 57: Path is wrong for the spec, please check. And it should be raised in EE know?

ayushpahwa: Fixed the path
Added here since this is a common file for CE and EE. Also, it wouldn't matter if the file is not present in CE. right?

PR #26943, opened 2023-09-04
removed undefined for all widget properties(test)

vsvamsi1:
removed undefined for all widget properties

### DIFF
--- a/app/client/src/ce/workers/Evaluation/dataTreeUtils.ts
+++ b/app/client/src/ce/workers/Evaluation/dataTreeUtils.ts
@@ -1,7 +1,6 @@
 import type { DataTree } from "entities/DataTree/dataTreeFactory";
 import { get, set, unset } from "lodash";
 import type { EvalProps } from "workers/common/DataTreeEvaluator";
-import { removeFunctionsAndSerialzeBigInt } from "@appsmith/workers/Evaluation/evaluationUtils";
 
 /**
  * This method loops through each entity object of dataTree and sets the entity config from prototype as object properties.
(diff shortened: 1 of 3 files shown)


app/client/src/ce/workers/Evaluation/dataTreeUtils.ts
```diff
@@ -22,12 +21,17 @@ export function makeEntityConfigsAsObjProperties(
...
+        acc[k] = entity[k];
+        return acc;
+      }, {});
+    newDataTree[entityName] = { ...updated };
   }
-  const dataTreeToReturn = sanitizeDataTree
-    ? removeFunctionsAndSerialzeBigInt(newDataTree)
-    : newDataTree;
+
+  const dataTreeToReturn = sanitizeDataTree ? newDataTree : newDataTree;
```
ankitakinger, line 34: why do we need a conditional statement here? aren't both the scenarios returning same data to the variable?

app/client/src/ce/workers/Evaluation/dataTreeUtils.ts
```diff
@@ -22,12 +21,17 @@ export function makeEntityConfigsAsObjProperties(
   } = option;
   const newDataTree: DataTree = {};
   for (const entityName of Object.keys(dataTree)) {
-    const entity = dataTree[entityName];
-    newDataTree[entityName] = Object.assign({}, entity);
+    const entity = dataTree[entityName] as any;
+    const updated = Object.keys(entity as any)
```
ankitakinger, line 25: Can we give a specific type instead of any in the above 2 lines?

PR #26966, opened 2023-09-05
chore: Delete unused app/util/plugin-generation directory

sharat87:
These plop templates are no longer used, and are very likely stale.


riodeuno (approved):
As the chore is to remove the hbs/plop based generation of plugin files -- the change looks good to me.

Issue #26018, opened 2023-08-04
Cover color algo with unit tests


KelvinOm:
- [x] bg
- [x] bgAccent
- [x] bgAccentHover
- [x] bgAccentActive
- [x] bgAccentSubtle
- [x] bgAccentSubtleHover
- [x] bgAccentSubtleActive
- [x] bgAssistive
- [x] bgNeutral
- [x] bgNeutralHover
- [x] bgNeutralActive
- [x] bgNeutralSubtle
- [x] bgNeutralSubtleHover
- [x] bgNeutralSubtleActive
- [x] bgPositive
- [x] bgPositiveHover
- [x] bgPositiveActive
- [x] bgPositiveSubtleHover
- [x] bgPositiveSubtleActive
- [x] bgNegative
- [x] bgNegativeHover
- [x] bgNegativeActive
- [x] bgNegativeSubtleHover
- [x] bgNegativeSubtleActive
- [x] bgWarning
- [x] bgWarningHover
- [x] bgWarningActive
- [x] bgWarningSubtleHover
- [x] bgWarningSubtleActive
- [x] fg
- [x] fgAccent
- [x] fgNeutral
- [x] fgPositive
- [x] fgNegative
- [x] fgWarning
- [x] fgOnAccent
- [x] fgOnAssistive
- [x] fgOnNeutral
- [x] fgOnPositive
- [x] fgOnNegative
- [x] fgOnWarning
- [x] bdAccent
- [x] bdFocus
- [x] bdNeutral
- [x] bdNeutralHover
- [x] bdPositive
- [x] bdPositiveHover
- [x] bdNegative
- [x] bdNegativeHover
- [x] bdWarning
- [x] bdWarningHover
- [x] bdOnAccent
- [x] bdOnNeutral
- [x] bdOnPositive
- [x] bdOnNegative
- [x] bdOnWarning

KelvinOm:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Widget-Design-System-6434e3402be2c1002ea740be/issues/appsmithorg/appsmith/26018?planning-poker) with Zenhub @ichik @jsartisan @KelvinOm

PR #27021, opened 2023-09-06
chore: add unit test for bgAssitive + bgAccentSubtleActive + bgAccentSubtleHover

jsartisan:
Related #26018

### DIFF
--- a/app/client/packages/design-system/theming/src/color/tests/DarkModeTheme.test.ts
+++ b/app/client/packages/design-system/theming/src/color/tests/DarkModeTheme.test.ts
@@ -1,136 +1,187 @@
 import { DarkModeTheme } from "../src/DarkModeTheme";
 
-describe("@design-system/theming/color/DarkModeTheme", () => {
-  it("checks bg color", () => {
-    // chroma < 0.04
+describe("bg color", () => {
+  it("should return correct color when chroma is less than 0.04", () => {
     const { bg: bg1 } = new DarkModeTheme("oklch(0.92 0.02 110)").getColors();
     expect(bg1).toBe("rgb(4.3484% 4.3484% 4.3484%)");
+  });
 
-    // chroma > 0.04
+  it("should return correct color when chroma is greater than 0.04", () => {
     const { bg: bg2 } = new DarkModeTheme("oklch(0.92 0.05 110)").getColors();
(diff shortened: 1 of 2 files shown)


app/client/packages/design-system/theming/src/color/tests/DarkModeTheme.test.ts
```diff
@@ -88,4 +88,52 @@ describe("@design-system/theming/color/DarkModeTheme", () => {
...
+    expect(bgAccentSubtle3).toBe("rgb(0% 14.556% 23.9%)");
+
+    // seedChroma < 0.04
+    const { bgAccentSubtle: bgAccentSubtle4 } = new DarkModeTheme(
+      "oklch(0.30 0.03 231)",
+    ).getColors();
+    expect(bgAccentSubtle4).toBe("rgb(13.15% 13.15% 13.15%)");
+  });
+
+  it("it checks bgAccentSubtleHover color", () => {
```
KelvinOm, line 118: let's wrap each token into separate `describe` function.

ichik (dismissed):

app/client/packages/design-system/theming/src/color/tests/DarkModeTheme.test.ts
jsartisan: What benefit it would give us? 🤔

KelvinOm: When test falling, it will be easier for us to understand what token is a problem. It will also improve the readability of tests.

jsartisan: can you give me an example of a token how you are expecting it? I'll do it that way.

KelvinOm:
```
describe("bgAccentActive", () => {
  it("should return correct color if seedLightness < 0.4", () => {
    const { bgAccentActive: bgAccentActive1 } = new DarkModeTheme(
      "oklch(0.2 0.09 231)",
    ).getColors();
    expect(bgAccentActive1).toBe("rgb(0% 17.836% 27.428%)");
  });
});
```

KelvinOm (approved):

PR #27059, opened 2023-09-07
chore: Adding migrations to remove unassign permission from workspace developer and workspace app viewer roles

trishaanand:


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration022RemoveUnassignPermissionFromUnnecessaryRoles.java
```diff
@@ -0,0 +1,62 @@
...
+    public void rollbackExecution() {}
+
+    @Execution
+    public void executeMigration() {
+
+        // Fetch all default workspace roles except administrators
+        Criteria workspaceDeveloperAndAppViewerRolesCriteria = new Criteria()
+                .andOperator(
+                        Criteria.where("defaultDomainType").is(Workspace.class.getSimpleName()),
+                        Criteria.where("name").not().regex("^Administrator - .*"));
```
nsarupr, line 37: Nit: Should we fetch the fields using something like 
`fieldName(QPermissionGroup.permissionGroup.defaultDomainType)`

nsarupr, line 37: Can we use the constant from the FieldName?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration022RemoveUnassignPermissionFromUnnecessaryRoles.java
```diff
@@ -0,0 +1,62 @@
...
+                            .findFirst();
+
+                    if (!optionalUnassignPolicy.isPresent()) {
+                        return;
+                    }
+
+                    Policy unAssignPolicy = optionalUnassignPolicy.get();
+                    unAssignPolicy.getPermissionGroups().remove(permissionGroup.getId());
+
+                    mongoTemplate.save(permissionGroup);
```
nsarupr, line 59: Should we only get the selected fields and do an update operation instead of doing a save operation.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration022RemoveUnassignPermissionFromUnnecessaryRoles.java
```diff
@@ -0,0 +1,62 @@
...
+                        Criteria.where("name").not().regex("^Administrator - .*"));
+
+        Query optimizedQueryForInterestingPermissionGroups = optimizeQueryForNoCursorTimeout(
+                mongoTemplate, new Query(workspaceDeveloperAndAppViewerRolesCriteria), PermissionGroup.class);
+
+        mongoTemplate.stream(optimizedQueryForInterestingPermissionGroups, PermissionGroup.class)
+                .forEach(permissionGroup -> {
+                    if (CollectionUtils.isNullOrEmpty(permissionGroup.getPolicies())) {
+                        return;
+                    }
```
nsarupr, line 46: If we are not doing an update, in that case, we should add these filters to the query itself.

nsarupr (commented):
Suggestions added.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration022RemoveUnassignPermissionFromUnnecessaryRoles.java
trishaanand: Used an update instead of save.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration022RemoveUnassignPermissionFromUnnecessaryRoles.java
trishaanand: I dont like using variables in migrations to ensure the migrations are protected from field name changes which may happen.

nsarupr (approved):
Looks good to me.

nsarupr (approved):
The changes look good to me.

PR #27073, opened 2023-09-07
chore: Calculate initial widget size in DataTreeFactory

### LINKED ISSUE #26629: [Task]: Cleanup dataTreeFactory to avoid exposing position properties

### SubTasks

dataTreeFactory is adding `appPositioningType` and `isMobile` properties to the DataTree. This is used to derive height of the widgets within some widget derived properties.
This should be avoided and we should be directly providing height & width of the widget into the DataTree.

prsidhu:
## Description

- Remove ```appPositioningType``` and ```isMobile``` props from widget
DataTree.
- Include width and height as ```derivedHeight and derivedWidth``` in
widget DataTree. Prefer width and height if available, else calculate
using position props.
- Update TableWidgetV2 to use the ```derivedHeight``` prop in
derivedProps calculation to get pageSize.

Fixes #26629

list any relevant details for your test configuration.
> Delete anything that is not relevant
- [x] Manual
- [ ] JUnit
- [ ] Jest
- [ ] Cypress

1. Create an app.
2. Add a Table widget.
3. Connect the Table widget to a datasource.
4. Assert that the Table widget works as expected. The number of rows
per page are calculated accurately based on initial height of the
widget.

#### Dev activity
- [ ] I have added tests that prove my fix is effective or that my
feature works

#### QA activity:
- [ ] [Speedbreak
features](https://github.com/appsmithorg/TestSmith/wiki/Guidelines-for-test-plans#speedbreakers-)
have been covered
- [ ] Test plan covers all impacted features and [areas of
interest](https://github.com/appsmithorg/TestSmith/wiki/Guidelines-for-test-plans#areas-of-interest-)
- [ ] Test plan has been peer reviewed by project stakeholders and other
QA members
- [ ] We had an implementation alignment call with stakeholders post QA
Round 2


app/client/src/utils/ComponentSizeUtils.ts
```diff
@@ -0,0 +1,66 @@
+import type { BaseWidgetProps } from "widgets/BaseWidgetHOC/withBaseWidgetHOC";
```
marks0351, line 1: lets move this file to the common folder of layoutSystems

marks0351 (changes requested):
added just one minor nit.

marks0351 (dismissed):

aswathkk (changes requested):
There are some type issues due to which the client build fails. Can you check it @prsidhu?

app/client/src/layoutSystems/common/utils/ComponentSizeUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+  rightColumn,
+  topRow,
+}: BaseWidgetProps) => {
+  return {
+    componentWidth: (rightColumn - leftColumn) * parentColumnSpace,
+    componentHeight: (bottomRow - topRow) * parentRowSpace,
+  };
+};
+
+export const getComponentDimensions = (
```
aswathkk, line 61: `isAutoLayout` and `isMobile` seems like a property of Auto Layout system.

Can we make `getComponentDimensions` to accept only `widgetProps` and `appPositioningType`?
Based on this prop, we can check what kind of layoutComponentDimensions method to be called.

aswathkk: Also, can we memoize this function?
Seems like we would be calling it multiple times from DataTreeFactory

aswathkk (changes requested):

app/client/src/layoutSystems/common/utils/ComponentSizeUtils.ts
```diff
@@ -0,0 +1,72 @@
+import type { BaseWidgetProps } from "widgets/BaseWidgetHOC/withBaseWidgetHOC";
+
+export const getAutoLayoutComponentDimensions = ({
```
marks0351, line 3: pls add the jsdocs from the actual functions, it was updated on the final version of the base widget pr.

app/client/src/layoutSystems/common/utils/ComponentSizeUtils.ts
prsidhu: ```isMobile``` dependency can not be removed yet. Auto Layout system needs it from the MainCanvas. It will be removed with the layout system.

app/client/src/utils/ComponentSizeUtils.ts
aswathkk: @prsidhu Can we do it?

app/client/src/entities/DataTree/dataTreeFactory.ts
```diff
@@ -144,16 +144,12 @@ export class DataTreeFactory {
       const { configEntity, unEvalEntity } = generateDataTreeWidget(
         widget,
         widgetsMeta[widget.metaWidgetId || widget.widgetId],
+        appPositioningType,
```
riodeuno, line 147: If I understand this correctly:
Instead of passing the `appPositioningType` and `isMobile` directly to the `dataTree` object, we're trying to derive the appropriate `componentHeight` and `componentWidth` to be included in the `dataTree` object.

app/client/src/entities/DataTree/dataTreeWidget.ts
```diff
@@ -377,8 +381,19 @@ export const generateDataTreeWidget = (
 
   dataTreeWidget["meta"] = meta;
 
+  const { componentHeight, componentWidth } = getComponentDimensions(
+    dataTreeWidget,
+    appPositioningType,
+    isMobile,
```
riodeuno, line 387: We still have `isMobile` for backwards compatibility with Auto Layout, and we will not be relying on this in Anvil?
Is this assumption correct?

app/client/src/selectors/dataTreeSelectors.ts
```diff
@@ -24,6 +24,13 @@ import type { EvaluationError } from "utils/DynamicBindingUtils";
 import { getEvalErrorPath } from "utils/DynamicBindingUtils";
 import ConfigTreeActions from "utils/configTree";
 import { DATATREE_INTERNAL_KEYWORDS } from "constants/WidgetValidation";
+import { AppPositioningTypes } from "reducers/entityReducers/pageListReducer";
+
+const getAppPositioningType = (state: AppState) =>
+  AppPositioningTypes[
+    state.ui.applications.currentApplication?.applicationDetail?.appPositioning
```
riodeuno, line 31: I'm assuming `appPositioning?.type` instead of `appPositioning.type` is for legacy applications that may not have the `appPositioning` object?

riodeuno: Let's add a comment to this, so that readers are aware of this fact and choice in the future.

app/client/src/utils/ComponentSizeUtils.ts
```diff
@@ -0,0 +1,85 @@
+import { memoize } from "lodash";
+import { AppPositioningTypes } from "reducers/entityReducers/pageListReducer";
```
riodeuno, line 2: I am wondering if the reducer is the right place for this type definition. Wouldn't a separate types or constants file in the layout system be a better spot?

app/client/src/utils/ComponentSizeUtils.ts
```diff
@@ -0,0 +1,85 @@
...
+import type { BaseWidgetProps } from "widgets/BaseWidgetHOC/withBaseWidgetHOC";
+
+/**
+ * getAutoLayoutComponentDimensions
+ *
+ * utility function to compute a widgets dimensions in Auto layout system
```
riodeuno, line 8: The Auto-Layout Layout system uses the mobile breakpoints for layouts. For this reason, the rows and columns occupied in the mobile breakpoint, and the rows and columns occupied in other viewport widths can be different. As a result, we check if we're in the mobile breakpoint using the `isMobile`, and use the appropriate `leftColumn`, `rightColumn`, `topRow` and `bottomRow` (with or without the `mobile` prefix depending on the `isMobile` value) to compute the component dimensions. The components here are the widgets.

app/client/src/utils/ComponentSizeUtils.ts
```diff
@@ -0,0 +1,85 @@
...
+  (
+    props: BaseWidgetProps,
+    appPositioningType: AppPositioningTypes,
+    isMobile: boolean,
+  ): {
+    componentHeight: number;
+    componentWidth: number;
+  } => {
+    switch (appPositioningType) {
+      case AppPositioningTypes.AUTO:
```
riodeuno, line 79: I am assuming Anvil will not be relying on this, and will be using the absolute values from the `positionsReducer`?

app/client/src/utils/ComponentSizeUtils.ts
```diff
@@ -0,0 +1,85 @@
...
+  rightColumn,
+  topRow,
+}: BaseWidgetProps) => {
+  return {
+    componentWidth: (rightColumn - leftColumn) * parentColumnSpace,
+    componentHeight: (bottomRow - topRow) * parentRowSpace,
+  };
+};
+
+export const getComponentDimensions = memoize(
```
riodeuno, line 69: Based on the Layout system we're using (`appPositioningType`) we use the widget properties to generate the component dimensions in pixels.
We memoize this function because?

app/client/src/utils/ComponentSizeUtils.ts
prsidhu: @riodeuno Anvil will need initial position calculation when dropping a new widget. Position observer will execute and take over afterwards.

app/client/src/entities/DataTree/dataTreeFactory.ts
prsidhu: yes that's correct.

app/client/src/entities/DataTree/dataTreeWidget.ts
prsidhu: yes.

app/client/src/utils/ComponentSizeUtils.ts
prsidhu: @aswathkk This is what is causing the circular dependency.

app/client/src/utils/ComponentSizeUtils.ts
prsidhu: It can be moved to the ```autoLayoutTypes.ts```. This is done in the Anvil layout structure PR that follows.

app/client/src/utils/ComponentSizeUtils.ts
prsidhu: To avoid re-computation if props for the widget haven't changed. What do you suggest?

PR #27107, opened 2023-09-08
fix: fix incorrect parsing of email ID in login rate limit

shubham7saxena7:
Previously, we had incorrect parsing logic wherein we were parsing '+' character for space because of URLDecoding into UTF-8 charset. 

we have removed the decoding logic now

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/filters/PreAuth.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/filters/PreAuth.java
@@ -12,7 +12,6 @@ import org.springframework.web.server.WebFilterChain;
 import reactor.core.publisher.Mono;
 
 import java.net.URI;
-import java.net.URLDecoder;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 
@@ -51,7 +50,7 @@ public class PreAuth implements WebFilter {
         return exchange.getFormData().flatMap(formData -> {
             String username = formData.getFirst(FieldName.USERNAME.toString());
             if (username != null && !username.isEmpty()) {
-                return Mono.just(URLDecoder.decode(username, StandardCharsets.UTF_8));
+                return Mono.just(username);
             }
 
             return Mono.just("");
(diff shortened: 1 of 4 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/filters/PreAuth.java
```diff
@@ -51,7 +50,7 @@ private Mono<String> getUsername(ServerWebExchange exchange) {
         return exchange.getFormData().flatMap(formData -> {
```
nsarupr, line 50: Let's use map here, instead of using `flatMap` and then doing a Mono.just(<some_string>);
Also, do you want to throw an error, if there is no username in the `formData`?

app/server/appsmith-server/src/main/java/com/appsmith/server/filters/PreAuth.java
```diff
@@ -33,7 +32,7 @@ public Mono<Void> filter(ServerWebExchange exchange, WebFilterChain chain) {
                 return rateLimitService
                         .tryIncreaseCounter(RateLimitConstants.BUCKET_KEY_FOR_LOGIN_API, username)
                         .flatMap(counterIncreaseAttemptSuccessful -> {
-                            if (!counterIncreaseAttemptSuccessful) {
+                            if (Boolean.FALSE.equals(counterIncreaseAttemptSuccessful)) {
```
nsarupr, line 35: Can we assert with the `TRUE` instead of `FALSE`? In case the returned response is `null`, it will still not throw any error but rather pass it forward.

shubham7saxena7: the whole point is that counterIncreaseAttemptSuccessful should be false

nsarupr (approved):
Changes look good to me.
Let's have a DP testing for this one.

app/server/appsmith-server/src/main/java/com/appsmith/server/filters/PreAuth.java
```diff
@@ -51,7 +50,7 @@ private Mono<String> getUsername(ServerWebExchange exchange) {
         return exchange.getFormData().flatMap(formData -> {
             String username = formData.getFirst(FieldName.USERNAME.toString());
             if (username != null && !username.isEmpty()) {
-                return Mono.just(URLDecoder.decode(username, StandardCharsets.UTF_8));
```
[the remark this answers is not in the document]
shubham7saxena7: because we are getting it from form data - it was a reasonable assumption to decode the username to a standard charset. had it not been json data, we wouldn't have.

RakshaKShetty:
@nsarupr , @shubham7saxena7  i tried with the following characters " !$%^&*+_-?~ ." as a part of the email and this seems to working fine.
Tried sign-in scenarios with a valid username and password and  invalid username password both works.
Please highlight if i have missed anything

PR #27127, opened 2023-09-09
chore: Use single COPY command in Dockerfile for constant/static files

sharat87:
Move the files that are copied into the Docker image, into an `fs` folder, that reflects the folder structure of that in the image. This means two things right away:

1. A single `COPY` instruction in `Dockerfile` is enough to copy all the files to their places.
2. The structure of files in the repo reflects that in the Docker image. This makes working with the files/folders and troubleshooting with them much easier.

❗ Note: **There's actually only 3 files changed, rest are just moved.**


Dockerfile
```diff
@@ -34,8 +34,7 @@ RUN curl --silent --show-error --location https://www.mongodb.org/static/pgp/ser
   && curl --silent --show-error --location https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - \
   && apt update \
   && apt-get install --no-install-recommends --yes mongodb-org nodejs redis build-essential postgresql-13 \
-  && apt-get clean \
-  && rm -rf /var/lib/apt/lists/*
```
mohanarpit, line 38: Any reason we removed the `rm -rf` command?

mohanarpit (approved):

Dockerfile
sharat87: It's a duplicate. We're deleting the same thing a couple instructions down. 🙂

PR #27129, opened 2023-09-09
fix: Leave a role should be supported without permission as long as the role is directly assigned to the user

trishaanand:
Fixes https://github.com/appsmithorg/appsmith-ee/issues/2136


app/server/appsmith-server/src/test/java/com/appsmith/server/services/PermissionGroupServiceTest.java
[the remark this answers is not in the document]
trishaanand: Yes. This is in CE code repo and tests for common flow for leave role functionality. The EE file tests for EE only functionalities in EE repo

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserWorkspaceServiceCEImpl.java
```diff
@@ -106,6 +109,25 @@ public Mono<User> leaveWorkspace(String workspaceId) {
...
+                 * workspace. This means that we can filter the roles to find the single role assigned to the user
+                 * which the user intends to leave.
+                 */
+                .filter(tuple -> {
+                    PermissionGroup permissionGroup = tuple.getT1();
+                    User currentUser = tuple.getT2();
+
+                    Set<String> assignedToUserIds = permissionGroup.getAssignedToUserIds();
+
+                    if (assignedToUserIds.contains(currentUser.getId())) {
```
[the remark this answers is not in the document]
trishaanand: Not a big fan of ternary operator due to reduced code clarity for junior developers. If its not a big deal, I would like to leave this change as is.

shubham7saxena7 (dismissed):

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/PermissionGroupServiceCEImpl.java
```diff
@@ -373,4 +370,45 @@ public Mono<PermissionGroup> bulkUnAssignFromUserAndSendEvent(PermissionGroup pe
...
+                        return Mono.error(new AppsmithException(
+                                AppsmithError.USER_NOT_ASSIGNED_TO_ROLE,
+                                currentUser.getUsername(),
+                                permissionGroup.getName()));
+                    }
+
+                    assignedToUserIds.remove(userId);
+
+                    Update updateObj = new Update();
+                    String path = fieldName(QPermissionGroup.permissionGroup.assignedToUserIds);
```
nsarupr, line 404: Could this lead to a race condition, if two users try to leave the same Role simultaneously?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserWorkspaceServiceCEImpl.java
```diff
@@ -106,6 +109,25 @@ public Mono<User> leaveWorkspace(String workspaceId) {
...
+                        return TRUE;
+                    }
+
+                    return FALSE;
+                })
+                .map(tuple -> tuple.getT1())
                 .single()
```
nsarupr, line 131: If the flux is empty, `single()` ends up throwing an error. Earlier, we were ensuring the same by throwing an error (Line#110-111).  Can we please take care of that?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/PermissionGroupServiceCEImpl.java
```diff
@@ -373,4 +370,45 @@ public Mono<PermissionGroup> bulkUnAssignFromUserAndSendEvent(PermissionGroup pe
         return bulkUnassignFromUsers(permissionGroup, users)
                 .flatMap(permissionGroup1 -> sendEventUserRemovedFromRole(permissionGroup, usernames));
     }
+
+    @Override
+    public Mono<Boolean> leaveExplicitlyAssignedSelfRole(String permissionGroupId) {
```
nsarupr, line 375: Can you check if we need to send an event here?

nsarupr (changes requested):
Added comments. Please check.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/PermissionGroupServiceCEImpl.java
trishaanand: No event needed till we actually use this for other functionalities.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserWorkspaceServiceCEImpl.java
trishaanand: Its already throwing an error if we don't find the user explicitly assigned to one of the default roles `getAllByAssignedToUserAndDefaultWorkspace`.  In fact I should remove this entire filtering logic since we would only get one default role from the query. I am removing the filtering mechanism right now.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/PermissionGroupServiceCEImpl.java
trishaanand: Yes. Unfortunately this exists for all set operations today.

nsarupr (approved):
Changes look good to me .

PR #27156, opened 2023-09-11
chore: Move cleanup logs from cron to backend server

sharat87:
This is moving the cleanup script to delete old log files, from being a cron job, to being a scheduled task in the backend server. Why?

1. We want to support running Appsmith with readonly root, which is a request from security teams at enterprise companies.
2. Cron doesn't play nice. It wants to save a `.pid` file under `/var/run`, so it fails to start when running with readonly root. This is not configurable.
3. Since our use of cron is minimal only, we're moving away from it, especially since the backend server is already capable of running scheduled tasks.
4. This moves one job, there's still another. Based on experience from this, we'll work on the other.

Another advantage to doing this is that since this job is now running in the backend server, which has access to MongoDB and Redis, we can coordinate when running as a cluster, that only _one_ backend is running the clean up job. This is for much later though.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/cron/CleanUpOldLogs.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/cron/CleanUpOldLogs.java
@@ -14,7 +14,7 @@ import java.util.stream.Stream;
 @Component
 public class CleanUpOldLogs {
 
-    private static final int CUTOFF_AGE = 7 * 24 * 60 * 60 * 1000;
+    private static final int CUTOFF_AGE = 7 * 24 * 60 * 60 * 1000; // 7 days
 
     @Scheduled(cron = "0 0 0 * * SAT")
     public void cleanUpOldLogs() throws IOException {
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/cron/CleanUpOldLogs.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/cron/CleanUpOldLogs.java
@@ -1,0 +1,45 @@
+package com.appsmith.server.cron;
+
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.scheduling.annotation.Scheduled;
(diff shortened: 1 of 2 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/cron/CleanUpOldLogs.java
```diff
@@ -0,0 +1,45 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.stream.Stream;
+
+@Slf4j
+@Component
+public class CleanUpOldLogs {
+
+    private static final int CUTOFF_AGE = 7 * 24 * 60 * 60 * 1000;
```
nidhi-nair, line 17:
```suggestion
    private static final int CUTOFF_AGE = 7 * 24 * 60 * 60 * 1000; // 7 days
```

app/server/appsmith-server/src/main/java/com/appsmith/server/cron/CleanUpOldLogs.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Scheduled(cron = "0 0 0 * * SAT")
+    public void cleanUpOldLogs() throws IOException {
```
nidhi-nair, line 20: Minor, do you want to raise Sentry errors for the failures in here? I'm not sure if we have exposed Sentry configuration set up steps to other folks.

nidhi-nair (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/cron/CleanUpOldLogs.java
sharat87: I don't think we get Sentry alerts for self-hosted setups, so, doesn't make much sense. On our own cloud, your question gives me an idea, I'll set reminders for the next few weeks to see that this is indeed cleaning up correctly. Thanks! 🙂

PR #27208, opened 2023-09-12
chore: add unit tests for bgNeutral and bgNeutralHover

jsartisan:
Related #26018

### DIFF
--- a/app/client/packages/design-system/theming/src/color/tests/DarkModeTheme.test.ts
+++ b/app/client/packages/design-system/theming/src/color/tests/DarkModeTheme.test.ts
@@ -188,3 +188,63 @@ describe("bgAssistive color", () => {
     expect(bgAssistive1).toBe("rgb(92.148% 92.148% 92.148%)");
   });
 });
+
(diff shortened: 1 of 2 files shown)


ichik (approved):

app/client/packages/design-system/theming/src/color/tests/DarkModeTheme.test.ts
```diff
@@ -188,3 +188,63 @@ describe("bgAssistive color", () => {
     expect(bgAssistive1).toBe("rgb(92.148% 92.148% 92.148%)");
   });
 });
+
+describe("bgNeutral color", () => {
+  it("should return correct color when lightness is less than 0.5", () => {
+    const { bgNeutral: bgNeutral1 } = new DarkModeTheme(
```
KelvinOm, line 194: In the current implementation of the tests, we no longer need to rename tokens. We can just write `const { bgNeutral} = new DarkModeTheme()`

KelvinOm (approved):

PR #27241, opened 2023-09-13
ci: Cypress Push Slack notification fixes

Aishwarya-U-R:
## Description
- Since ci-hosted run has no build failures - that is removed
- For push flow, step id is added for correct status capture


sarojsarab (approved):
LGTM

PR #27281, opened 2023-09-13
feat: add property hideNoneIcon to IconSelectControl

### LINKED ISSUE #26713: [Bug]: (+) icon symbol is not displayed by default on selecting none for menu and input widget

### Description

Icon property behavior is not consistent for icon widget and other widget  having icon as their property

Note : Issue exists for the widgets which has icon property 
ex :

- menu widget
- Input widget

Attached video : https://www.loom.com/share/60e053095e424356867e067cea5426da?sid=5bae4489-5f2f-4fa9-b6ca-0c161b33835c

### Steps To Reproduce

1. Drag and drop menu widget/input widget
2. Go to styling panel
3. Select any icon (lock icon) in icon property
4. now search for number in icon property such as 123, it will give `none` as search result
5. Select none and observe no icon displays 
6. Repeat same steps for icon widget and observe  (+) icon will be displayed by default on selecting none

### Public Sample App

_No response_

### Environment

Production

### Issue video log

_No response_

### Version

Appsmith v1.9.35-SNAPSHOT

dhruvikn:
## Description

Add a new property `hideNoneIcon ` to IconSelectControl. When this property is `true`, the icon selector will not show `(none)` in the list and neither will search for it.

Fixes #26713

#### Type of change

- [x] Manual
- [ ] JUnit
- [ ] Jest
- [ ] Cypress
>
>
#### Dev activity

#### QA activity:
- [x] Test plan covers all impacted features and [areas of interest](https://github.com/appsmithorg/TestSmith/wiki/Guidelines-for-test-plans#areas-of-interest-)


app/client/src/components/propertyControls/IconSelectControl.tsx
```diff
@@ -131,13 +131,28 @@ class IconSelectControl extends BaseControl<
...
+
   constructor(props: IconSelectControlProps) {
     super(props);
     this.iconSelectTargetRef = React.createRef();
     this.virtuosoRef = React.createRef();
     this.searchInput = React.createRef();
     this.initialItemIndex = 0;
     this.filteredItems = [];
+
+    const noneIndex = ICON_NAMES.indexOf(NONE);
```
aswathkk, line 146: Do we really need to check the index of `NONE` since we are unshifting it and it would always be at 0th position?
We can just use the icon set to check if `NONE` is there or not.
How about the following change?

```suggestion
    const noneIconExists = icons.has(NONE);

    if (props.showNoneIcon && !noneIconExists) {
      ICON_NAMES.unshift(NONE);
      icons.add(NONE);
    } else if (!props.showNoneIcon && noneIconExists) {
      ICON_NAMES.shift();
      icons.delete(NONE);
    }
```

aswathkk: Multiple instances of the `IconSelectControl` class may be created, and each instance modifies the ICON_NAMES array and the icons set. Without this logic, the NONE icon may be added or removed multiple times, leading to unexpected behaviour. That's why you are adding and removing the NONE icon. Right?
Can you add a comment explaining the same?

aswathkk (changes requested):
Added minor comment which I think would make the code a bit more readable.

Also, isn't it better to name this property `hideNoneIcon` rather than `showNoneIcon`?
Using that, we can also avoid making the default value as `true`

app/client/src/components/propertyControls/IconSelectControl.tsx
```diff
@@ -138,6 +138,23 @@ class IconSelectControl extends BaseControl<
...
+
+    /**
+     * Multiple instances of the IconSelectControl class may be created,
+     * and each instance modifies the ICON_NAMES array and the icons set.
+     * Without the below logic, the NONE icon may be added or removed
+     * multiple times, leading to unexpected behaviour.
+     */
+    const noneIconExists = icons.has(NONE);
+
+    if (!!!props.hideNoneIcon && !noneIconExists) {
```
aswathkk, line 150: Why do we need `!!!`?
Won't `!` be fine?

app/client/src/components/propertyControls/IconSelectControl.tsx
```diff
@@ -138,6 +138,23 @@ class IconSelectControl extends BaseControl<
...
+      ICON_NAMES.unshift(NONE);
+      icons.add(NONE);
+    } else if (!!props.hideNoneIcon && noneIconExists) {
```
aswathkk, line 153: Similarly, do we need `!!`?

aswathkk (changes requested):

app/client/src/components/propertyControls/IconSelectControl.tsx
dhruvikn: Yes, you're right. Force typecasting here would be an overkill.

app/client/src/components/propertyControls/IconSelectControl.tsx
dhruvikn: Same as above. 🙏

laveena-en:
Tested & verified, changes look good to me!

PR #27310, opened 2023-09-14
ci: Use esbuild to build RTS

sharat87:
This fixes RTS build to use `esbuild`.

1. This means the whole `node_modules` won't need to be copied over to the Docker image. There's unused insignifant _test_ files in there, that don't add any value, but are causing irrelevant CVEs to be reported on our Docker image. See example at https://github.com/appsmithorg/appsmith-ee/pull/2349.

2. Much faster. Not that RTS build is our slow point, but still. Perhaps we can move client to `esbuild` too. 🙂

## Why are we doing this?

The current method of loading RTS into the Docker image means that _all_ contents of _all_ dependencies are copied over. The whole `node_modules`. But several of these packages include _test_ files too, that aren't needed at runtime at all. One of such test files is creating a false alert for a CVE on our Docker image. Has absolutely no relevance and impact, but it's there.

To fix that, I [had to `rm -rf /opt/appsmith/rts/node_modules/*/test` in the Docker image](https://github.com/appsmithorg/appsmith-ee/pull/2349/files). This felt very hacky, and very dirty. It felt like we're introducing more debt and more duct tape around the current build process.

So, `esbuild`.

## Where is `esbuild` coming from?

We're using `esbuild` v0.18.20 only, while the latest is v0.19.3. We need to update `design-system`'s storybook dependency, I think, to get a more recent version of `esbuild`. I'm yet to figure this out and can use some help. 🙂


app/client/packages/rts/src/server.ts
```diff
@@ -35,16 +34,10 @@ const io = new Server(server, {
...
-// Initializing Sockets
 initializeSockets(io);
 
 // parse incoming json requests
 app.use(express.json({ limit: "5mb" }));
-// Initializing Routes
-app.use(express.static(path.join(__dirname, "static")));
-app.get("/", (_, res) => {
-  res.redirect("/index.html");
-});
```
sharat87, line 47: Unused. NGINX isn't configured to send _any_ traffic with these paths to RTS.

mohanarpit (approved):

KelvinOm (commented):
I'm really glad to see this PR 👍.

>Much faster. Not that RTS build is our slow point, but still. Perhaps we can move client to esbuild too. 🙂

We consider VIte as a [building tool](https://www.notion.so/appsmith/FE-improvements-cbb2225fc17a47508138d65e9d9a8a24?p=51971d78e293496d9c6c3c960284a36c&pm=s), but it's not simple. It takes time to try it and test it.

>We're using esbuild v0.18.20 only, while the latest is v0.19.3. We need to update design-system's storybook dependency, I think, to get a more recent version of esbuild. I'm yet to figure this out and can use some help. 🙂

Esbuild comes to us as dependency for Vite. Since we have the latest version of Vite we cannot get the latest version Esbuild through this mechanism. Anyway, it is better to explicitly indicate that we use Esbuild in `rts` as a dependency.


It also seems that we can now not do a build for `ast` and `dsl` [here](https://github.com/appsmithorg/appsmith/blob/release/app/client/package.json#L20). Also we can delete the `postinstall` scripts in these packages. Before we had issues with build because of copying node_modules in `rts`.

sharat87:
@KelvinOm, sorry, saw your comment a little late.

> [...] it is better to explicitly indicate that we use Esbuild in rts as a dependency.

Yeah that's what I suspected too. I'll raise a separate PR for this. Thanks.

> It also seems that we can now not do a build for ast and dsl [here](https://github.com/appsmithorg/appsmith/blob/release/app/client/package.json?rgh-link-date=2023-10-02T13%3A17%3A42Z#L20).

Are you saying this build command will be failing after this PR is merged? 🤔 But then the Cypress tests should've failed, right? I'm not sure I follow.

KelvinOm:
> Are you saying this build command will be failing after this PR is merged? 🤔 But then the Cypress tests should've failed, right? I'm not sure I follow.

Nope, I'm talking about the fact that this is no longer necessary and we can speed up and simplify scripts a bit. We can work with the code directly, we don't need to build it before.

PR #27457, opened 2023-09-19
feat: added an intermediary step to enable new create new app flow

### LINKED ISSUE #27236: [Task]: Change of onboarding point to start creating application behind feature flag

### SubTasks

- Change the creation of application flow : Choose between template or scratch
- Fix the test cases if failing or support if for A/B test `false` value.
Notion doc : https://www.notion.so/appsmith/By-improving-the-templates-discoverability-we-ll-increase-the-of-users-who-start-from-a-template--5f219f72c5d74466a31662b9d928b01d

Designs: https://www.figma.com/file/kbU9xPv44neCfv1FFo9Ndu/User-Activation?type=design&node-id=2338-29522&mode=design&t=yFOm2Xh625BLfr41-0

AmanAgarwal041:
## Description
Added an intermediary step when a user registers and before a default application gets created.
Previously when a user used to register then a default workspace got created and then default application got created. But now if the flag `ab_create_new_apps_enabled` has a value `true` for the new user registering then in that case: 
- It is applicable to both cloud & self-hosted instances to non-admin users
- default workspace gets created 
- we redirect from client from Signup success page with a query param `startCreateNewApp=true` 
- Once the page is redirected to `/applications` with query param `startCreateNewApp=true`  we start another flow for creating a new app where user can select between `Start from scratch` or `Start from template`.

Since the flag check is done from server side when the new user registers for enabling the new create app flow, hence we added a way to stop the checking of flag value for cypress test cases. So, added a cookie check where we by default set the cookie `Skip-Server-Side-Flag-Check` to `true` for all cypress test cases to skip the flag check on server side. We can enable the check for a test case by setting the value to `false`.

Fixes #27236 
Notion Fix: https://www.notion.so/appsmith/Test-Party-Onboarding-Create-New-App-4321a350a91447ed950580a9efa793e6?p=9437501105a8458489640065dcb41a51&pm=s

>
>
>
- [x] Manual
- [ ] JUnit
- [ ] Jest
- [ ] Cypress
>
>
#### Dev activity

#### QA activity:


app/client/src/ce/pages/Applications/CreateNewAppsOption.tsx
```diff
@@ -0,0 +1,214 @@
...
+  `}
+`;
+
+const BackWrapper = styled.div`
+  position: sticky;
+  background: ;
+  ${(props) => `
+    top: ${props.theme.homePage.header}px;
+    `}
+  background: ${Colors.WHITE};
```
ankitakinger, line 50: can we use ADS colors here?

app/client/src/ce/pages/Applications/CreateNewAppsOption.tsx
```diff
@@ -0,0 +1,214 @@
...
+  display: flex;
+  gap: 16px;
+  margin-top: 48px;
+`;
+
+const CardContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  padding: 48px;
+  border: 1px solid #cdd5df;
```
ankitakinger, line 94: can we use ADS colors here?

app/client/src/ce/pages/Applications/index.tsx
```diff
@@ -801,30 +892,10 @@ export class Applications<
 
   componentWillUnmount() {
     this.props.setHeaderMetaData(false, false);
-    this.props.searchApplications("");
```
ankitakinger, line 804: why did we change the logic for `searchApplications` here exactly?

ankitakinger:
the back button looks misplaced:
image

app/client/src/ce/pages/Applications/index.tsx
```diff
@@ -98,6 +98,11 @@ import ApplicationCardList from "@appsmith/pages/Applications/ApplicationCardLis
 import { usePackage } from "@appsmith/pages/Applications/helpers";
 import PackageCardList from "@appsmith/pages/Applications/PackageCardList";
 import WorkspaceAction from "@appsmith/pages/Applications/WorkspaceAction";
+import { CreateNewAppsOption } from "./CreateNewAppsOption";
```
ankitakinger, line 101:
```suggestion
import { CreateNewAppsOption } from "@appsmith/pages/Applications/CreateNewAppsOption";
```

app/client/src/ce/pages/Applications/index.tsx
AmanAgarwal041: Because we moved the component to another child wrapper, so we moved the search application logic as it was not required in the parent component. ref : https://github.com/appsmithorg/appsmith/pull/27457/files#diff-d447b4855fe8828a835b4701a61fc7d678c6cbe192348d428c8b92bfafcbf953R829

app/client/src/ce/pages/Applications/index.tsx
AmanAgarwal041: Done

app/client/src/ce/pages/Applications/CreateNewAppsOption.tsx
```diff
@@ -0,0 +1,217 @@
...
+  `}
+`;
+
+const BackWrapper = styled.div`
+  position: sticky;
+  background: ;
+  ${(props) => `
+    top: ${props.theme.homePage.header}px;
+    `}
+  background: var(--ads-color-background);
```
ankitakinger, line 49: ADS v2 colors please. I think what you are looking for is `--ads-v2-color-bg`.

Aishwarya-U-R (commented):
Approving Cypress changes.

btsgh:
- fix the copy for let's on the Starting point page
- back button should not be shown
- the tabs  Apps | Templates should not be shown.

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBcnpuIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--dae3d6edd644c64fd35c4ecc9e360865b501a10b/Screen%20Recording%202023-09-28%20at%2015.01.14.mov)

- When we select Start with Use Case, It goes to the list of Templates we can choose from, here also we can see the Tabs and Apps is  highlighted in this case. As per figma, only Back button should be present here
- Invoke the instance first time, and answer the questions on Almost there page, register here with aparna@appsmith.com, and complete this flow, I am taken to the sign in page. [I was expecting the Starting point screens in this PR to appear here itself]. Sign in with the earlier entered creds, it throws an error - invalid creds. Usual flow as on today is, if we enter the credentials at the Almost there, user is directly taken to the login page, without having to re-enter credentials or sign up again.

app/client/src/assets/images/start-from-scratch.svg
sneha122, line 1: @AmanAgarwal041 Should we store assets locally or get it from assets.appsmith.com

AmanAgarwal041: We can store this on assets.appsmith.com. Will update this in another pr.

app/client/src/ee/pages/Applications/CreateNewAppsOption.tsx
```diff
@@ -0,0 +1,3 @@
+export * from "ce/pages/Applications/CreateNewAppsOption";
+import { default as CE_CreateNewAppsOption } from "ce/pages/Applications/CreateNewAppsOption";
```
sneha122, line 2: @AmanAgarwal041 Curious, why did we create a separate component in ee? I thought we do separation only when there is different logic for CE and EE, in this case also in EE we are going to have different logic is it?

sneha122 (approved):

app/client/src/ee/pages/Applications/CreateNewAppsOption.tsx
AmanAgarwal041: For now, there won't be a different logic but since we have couple of different things on ee like multiple env etc. Does it make sense to keep it out of ce or ee ? 
Updated the code as per change requested earlier by Ankita https://github.com/appsmithorg/appsmith/pull/27457#discussion_r1333059904 in `Applications/index.tsx` to require `CreateNewAppsOption.tsx` file.

sneha122: Cool makes sense, given that we have environments associated with it, its good to have separate components

sneha122 (approved):

PR #27502, opened 2023-09-20
fix restarting dp malicous dockerenv

pratapaprasanna:


scripts/deploy_preview.sh
```diff
@@ -98,6 +98,10 @@ helm upgrade -i $CHARTNAME appsmith-ee/$HELMCHART -n $NAMESPACE --create-namespa
   --set persistence.efs.volumeHandle=$DP_EFS_ID:/ce/ce$PULL_REQUEST_NUMBER \
   --set resources.requests.cpu="1m" \
   --set resources.requests.memory="1Mi" \
+  --set applicationConfig.APPSMITH_API_BASE_URL="http://localhost:8080/api/v1" \
+  --set applicationConfig.APPSMITH_REDIS_URL="redis://127.0.0.1:6379" \
+  --set applicationConfig.APPSMITH_ENCRYPTION_PASSWORD="obYwdTiwlFbs6" \
+  --set applicationConfig.APPSMITH_ENCRYPTION_SALT="tqcnUJBCXQ9CB" \
```
sharat87, line 104: Any reason we're hard-coding these? Like, all DPs will have this from now on, right? That means all DPs will fail with decryption error, unless they are recreated, right?

scripts/deploy_preview.sh
```diff
@@ -98,6 +98,10 @@ helm upgrade -i $CHARTNAME appsmith-ee/$HELMCHART -n $NAMESPACE --create-namespa
   --set persistence.efs.volumeHandle=$DP_EFS_ID:/ce/ce$PULL_REQUEST_NUMBER \
   --set resources.requests.cpu="1m" \
   --set resources.requests.memory="1Mi" \
+  --set applicationConfig.APPSMITH_API_BASE_URL="http://localhost:8080/api/v1" \
+  --set applicationConfig.APPSMITH_REDIS_URL="redis://127.0.0.1:6379" \
```
sharat87, line 102: Any reason we need to set this here? 🤔

pratapaprasanna: The main reason is docker.env is getting corrupted and sweeping away all the values . we are setting values which help in getting dp up and running despite a corrupt docker.env

scripts/deploy_preview.sh
pratapaprasanna: Yes, just to avoid a dp crash even if docker.env is cleared

sharat87: Sure, but this takes precedence over the values in `docker.env`, _if_ `docker.env` is *not* cleared, right? In those cases, we'll be breaking the DP, although there was no problem to begin with, no? 🤔

sharat87 (commented):
Responded to threads above.

PR #27540, opened 2023-09-21
chore: fix airgap explorer test

akash-codemonk:
In the test we were looking for the add js libraries button, but we cannot add js libraries on airgap. So skipping add libraries from the assertion list.

### DIFF
--- a/app/client/cypress/e2e/Regression/ClientSide/ExplorerTests/Pin_spec.js
+++ b/app/client/cypress/e2e/Regression/ClientSide/ExplorerTests/Pin_spec.js
@@ -109,7 +109,10 @@ describe("Entity explorer tests related to pinning and unpinning", function () {
     function () {
       agHelper.AssertElementVisibility(entityExplorer._entityExplorer);
       entityExplorer.PinUnpinEntityExplorer(true);
-      const menu = Object.keys(ExplorerMenu);
+      // We cannot add libraries on airgap
+      const menu = Object.keys(ExplorerMenu).filter(
+        (menu) => menu !== ExplorerMenu.ADD_LIBRARY,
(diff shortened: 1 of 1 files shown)


app/client/cypress/e2e/Regression/ClientSide/ExplorerTests/Pin_spec.js
```diff
@@ -109,7 +109,10 @@ describe("Entity explorer tests related to pinning and unpinning", function () {
     function () {
       agHelper.AssertElementVisibility(entityExplorer._entityExplorer);
       entityExplorer.PinUnpinEntityExplorer(true);
-      const menu = Object.keys(ExplorerMenu);
+      // We cannot add libraries on airgap
+      const menu = Object.keys(ExplorerMenu).filter(
+        (menu) => menu !== ExplorerMenu.ADD_LIBRARY,
+      );
```
Aishwarya-U-R, line 115: How does it skip this for Airgap run alone?

akash-codemonk: This runs on airgap only right https://github.com/appsmithorg/appsmith/blob/c1bae998ac6d2ed726356fa43531b4a5fc1fc1dc/app/client/cypress/e2e/Regression/ClientSide/ExplorerTests/Pin_spec.js#L107

Aishwarya-U-R: Oh got it.

Aishwarya-U-R:
@akash-codemonk - Merging this now. Lets check the Airgap run.

PR #25939, opened 2023-08-02
chore: update app typography

jsartisan:
Fixes #21426

PR #27549, opened 2023-09-21
fix: Cypress tests that were introduced in improved App typography PR

jsartisan:
This PR fixes a cypress test failure which was introduced in #25939

### DIFF
--- a/app/client/cypress/e2e/Regression/ClientSide/Widgets/ListV2/ListV2_NestedList_spec.ts
+++ b/app/client/cypress/e2e/Regression/ClientSide/Widgets/ListV2/ListV2_NestedList_spec.ts
@@ -123,14 +123,6 @@ describe("Nested List widget V2 ", () => {
   });
 
   it("5. Verify Theme change", () => {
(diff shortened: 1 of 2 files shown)


app/client/cypress/e2e/Regression/ClientSide/Widgets/ListV2/ListV2_NestedList_spec.ts
```diff
@@ -123,14 +123,6 @@ describe("Nested List widget V2 ", () => {
   });
 
   it("5. Verify Theme change", () => {
-    [0, 1, 2].forEach((index) => {
```
Aishwarya-U-R, line 126: Why do we remove this now?

jsartisan: It's because now the current font is system font which means text will just inherit font from body. So the previous check won't work anymore.

Aishwarya-U-R (approved):

PR #27558, opened 2023-09-22
chore: fix the jgit version updates in other places

AnaghHegde:
Should resolve the following CVE reports:

1. https://github.com/appsmithorg/appsmith/security/dependabot/263
1. https://github.com/appsmithorg/appsmith/security/dependabot/264

### DIFF
--- a/app/server/appsmith-git/pom.xml
+++ b/app/server/appsmith-git/pom.xml
@@ -29,15 +29,7 @@
         <dependency>
             <groupId>org.eclipse.jgit</groupId>
             <artifactId>org.eclipse.jgit.ssh.jsch</artifactId>
-            <version>5.13.0.202109080827-r</version>
-        </dependency>
-
-        <!-- https://mvnrepository.com/artifact/org.eclipse.jgit/org.eclipse.jgit.junit.ssh -->
-        <dependency>
-            <groupId>org.eclipse.jgit</groupId>
-            <artifactId>org.eclipse.jgit.junit.ssh</artifactId>
-            <version>6.4.0.202211300538-r</version>
-            <scope>test</scope>
+            <version>6.6.1.202309021850-r</version>
         </dependency>
 
         <dependency>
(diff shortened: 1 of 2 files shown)


nayan-rafiq (approved):

app/server/appsmith-git/pom.xml
```diff
@@ -29,15 +29,7 @@
...
-            <version>5.13.0.202109080827-r</version>
-        </dependency>
-
-        <!-- https://mvnrepository.com/artifact/org.eclipse.jgit/org.eclipse.jgit.junit.ssh -->
-        <dependency>
-            <groupId>org.eclipse.jgit</groupId>
-            <artifactId>org.eclipse.jgit.junit.ssh</artifactId>
-            <version>6.4.0.202211300538-r</version>
-            <scope>test</scope>
+            <version>6.6.1.202309021850-r</version>
```
sharat87, line 32: Can we avoid duplicating this version number please? Let's define a property like

```
<jgit.version>6.6.1.202309021850-r</jgit.version>
```

in `app/server/pom.xml`, then in `app/server/appsmith-interfaces/pom.xml`, as well as in this file here, we can replace the hard-coded version with `${jgit.version}`.

What say?

AnaghHegde: We have just one artifact of jgit in this pom file. Do you still think we need a variable here?

sharat87: It's just one here, but it's duplicated in `app/server/appsmith-interfaces/pom.xml`. Otherwise next time also we'll upgrade just `jgit` version in that interfaces module's `pom.xml`, and this will be left out. If use a variable in both of these places we can avoid this. 🙂

AnaghHegde: I get what you are saying. But Its not duplicated in `app/server/appsmith-interfaces/pom.xml`, i just checked. It was duplicated in `app/server/appsmith-git/pom.xml` and i removed the artifact that are not being used.

sharat87: We're duplicating the version in these places:

https://github.com/appsmithorg/appsmith/blob/7d4ecdb7a6b316958350d2a1320cf3278b62e65c/app/server/appsmith-git/pom.xml#L32

https://github.com/appsmithorg/appsmith/blob/7d4ecdb7a6b316958350d2a1320cf3278b62e65c/app/server/appsmith-interfaces/pom.xml#L151

Next time we upgrade jgit again, we have to remember to change the JSCH version too in that other place.

AnaghHegde: They are different modules but i don't think we should be worrying about that.

PR #27585, opened 2023-09-25
chore: Refactor Text Input with css modules

jsartisan:


app/client/packages/design-system/widgets/src/components/Spinner/src/styles.module.css
```diff
@@ -0,0 +1,19 @@
+.spinner {
+  animation: spin 1s linear infinite;
+  height: 1.5em;
+  width: 1.5em;
+
+  & path {
```
KelvinOm, line 6: We can remove this since we have `fill="currentColor"` on SVG itself.

KelvinOm (commented):
Could you redo the last two places and delete `styled-components` library?

Снимок экрана 2023-09-25 в 09 58 31

app/client/packages/design-system/widgets/src/components/Flex/stories/styles.module.css
```diff
@@ -0,0 +1,32 @@
+.demo-block {
+  min-width: var(--sizing-12);
+  min-height: var(--sizing-12);
+  flex: 1;
+
+  &[data-variant="accent"] {
```
KelvinOm, line 6: It should be a `data-color` instead of an variant. Also you can use a loop here.

KelvinOm (approved):

KelvinOm (approved):

PR #27669, opened 2023-09-27
Simpler focus logic

### LINKED ISSUE #26852: [Task]: Improve focus color logic

### SubTasks

The current logic is that the focus color becomes reddish if the seed color is blueish. The red focus color makes it seem as if the component is in an error state. We will need to imrove the logic for this (monochromatic or follow seed).

ichik:
Closes #26852


app/client/packages/design-system/theming/src/color/tests/LightModeTheme.test.ts
```diff
@@ -769,8 +769,8 @@ describe("bdAccent color", () => {
 
 describe("bdFocus color", () => {
   it("should return correct color when lightness < 0.6", () => {
-    const { bd } = new LightModeTheme("oklch(0.45 0.4 60)").getColors();
-    expect(bd).toEqual("rgb(80.718% 72.709% 66.526%)");
+    const { bdFocus } = new LightModeTheme("oklch(0.45 0.4 60)").getColors();
+    expect(bdFocus).toEqual("rgb(56.074% 13.73% 0%)");
```
ichik, line 773: @jsartisan @KelvinOm Somehow this was enough to make tests pass, which is strange. We may be testing wrong tokens from what I can see from some fails that I've been getting.
image
image

KelvinOm: In this test and several others, we use the wrong token(`bd` instead of `bdfocus`). @palashkaria  could you check this? 

It also seems that the coverage does not work quite correctly if we run all the tests at once. Since getColor runs all the color getters. In general, it is necessary to check all `describe` blocks separately and check the coverage again.

jsartisan:
@ichik Fixing the issue of testing wrong tokens here - https://github.com/appsmithorg/appsmith/pull/27684/files

app/client/packages/design-system/theming/src/color/src/DarkModeTheme.ts
```diff
@@ -841,30 +841,14 @@ export class DarkModeTheme implements ColorModeTheme {
   }
 
   private get bdFocus() {
```
KelvinOm, line 843: In DarkMode for black color focus looks like pink, shouldn't it be a more close to white in this case?
Снимок экрана 2023-09-28 в 10 24 25

KelvinOm:
> @ichik Fixing the issue of testing wrong tokens here - https://github.com/appsmithorg/appsmith/pull/27684/files

Fix is merged. @ichik Could you update your branch?

app/client/packages/design-system/theming/src/color/tests/LightModeTheme.test.ts
jsartisan: @ichik This is fixed. Please take the pull from the latest release.

app/client/packages/design-system/theming/src/color/src/DarkModeTheme.ts
ichik: @KelvinOm See comment in the code: https://github.com/appsmithorg/appsmith/blob/baea458802b996344e144d5420eed601247df74a/app/client/packages/design-system/theming/src/color/src/DarkModeTheme.ts#L847

I think it's beneficial for accessibility.

ichik:
@KelvinOm @jsartisan I've rebased and seeing on bunch of errors with tests.

ichik:
@KelvinOm updated.

KelvinOm (approved):

PR #27693, opened 2023-09-28
chore: Integrate Checkbox component 

### LINKED ISSUE #27689: Integrate WDS Checkbox component

jsartisan:
Fixes #27689


app/client/src/widgets/CheckboxWidgetV2/widget/index.tsx
```diff
@@ -0,0 +1,111 @@
...
+import type { SetterConfig } from "entities/AppTheming";
+import type { DerivedPropertiesMap } from "WidgetProvider/factory";
+import { EventType } from "constants/AppsmithActionConstants/ActionConstants";
+
+import * as config from "./../config";
+import BaseWidget from "../../BaseWidget";
+import type { CheckboxWidgetProps } from "./types";
+import type { WidgetState } from "../../BaseWidget";
+
+class CheckboxWidget extends BaseWidget<CheckboxWidgetProps, WidgetState> {
```
ashit-rath, line 12: Any reason for not using the name `CheckboxWidgetV2`? The reason for asking this is if I were to do a global search in my local editor for `CheckboxWidget`; I would see both V1 and V2 results and may lead to some level of confusion (checking the path of the file would only help in this case).

dhruvikn: I think we are doing this because all WDS widgets are behind a feature flag. When the flag's value is true, we only show WDS widgets and nothing else. So the searching should not be an issue. 

Please add if I'm missing anything. @jsartisan

dhruvikn: Ah okay, you were talking about search in your local editor. Apologies, please ignore my previous comment.

app/client/src/widgets/CheckboxWidgetV2/config/propertyPaneConfig/contentConfig.ts
```diff
@@ -0,0 +1,109 @@
...
+        controlType: "SWITCH",
+        helpText: "Disables input to this widget",
+        isJSConvertible: true,
+        isBindProperty: true,
+        isTriggerProperty: false,
+        validation: { type: ValidationTypes.BOOLEAN },
+      },
+      {
+        propertyName: "animateLoading",
+        label: "Animate loading",
```
KelvinOm, line 84: Do we need this and how this should work?

KelvinOm (commented):
We don't want to control the height of the checkbox through the property pane, right?

Снимок экрана 2023-10-02 в 18 04 37

app/client/src/widgets/CheckboxWidgetV2/config/autoLayoutConfig.ts
```diff
@@ -0,0 +1,19 @@
+export const autoLayoutConfig = {
```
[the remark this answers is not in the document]
marks0351: @riodeuno @KelvinOm In our previous discussion this is one of the reason why i mentioned should we think about having layout system specific widgets coz clearly we are building wds widgets only to work on anvil layout system as oppose to previous set of widgets which were written to work on both layout systems fixed and auto.
This means configurations of widgets often might only make sense in the anvil layout system paradigm. previous such configs Auto Height which is a common config seen in all widgets but only makes sense in Fixed paradigm, fill hug which only makes sense in auto and anvil, auto dimensions which also make sense only in auto and anvil.
same is to be said with Blue print operations written in some of the widgets.

We either provide widgets suites specific to the layout system or provide appropriate configs for all layoutsystems for every widget promising that one day we will make the widget work in all layout systems.

app/client/src/widgets/CheckboxWidgetV2/config/defaultsConfig.ts
```diff
@@ -0,0 +1,17 @@
+import { FILL_WIDGET_MIN_WIDTH } from "constants/minWidthConstants";
+import { ResponsiveBehavior } from "layoutSystems/autolayout/utils/constants";
+
+export const defaultsConfig = {
+  rows: 4,
```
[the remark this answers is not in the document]
marks0351: @KelvinOm yes this config will change, we would have something like width(pixels) and height(pixels).

app/client/src/widgets/CheckboxWidgetV2/config/autoLayoutConfig.ts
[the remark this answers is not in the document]
marks0351: @KelvinOm having minWidth and minHeight inside the component i.e, the view component in wds package restricts layout system components like DnD to know the default values and show a preview of the widget while dragging. for the current set of use cases I cannot make a compelling argument to do so, but at some point we might need them as widget specific configs on the layout system components.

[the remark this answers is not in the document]
KelvinOm: @riodeuno @marks0351 
1. In general, my vote for having a anvil specific config. What do we need to do to achieve this?
2. If we don't know why we need minWidth and minHeight values now(component itself knows this), then it's better not to add them. No need to write code for the future. We can always add them if we need them.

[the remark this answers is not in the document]
marks0351: @prsidhu do you see any usecase where having widget specific values like minwidth, minheight available to the layout system?

app/client/src/widgets/CheckboxWidgetV2/widget/index.tsx
jsartisan: >Any reason for not using the name CheckboxWidgetV2?
No particular reason as such. 

> The reason for asking this is if I were to do a global search in my local editor for CheckboxWidget
The file is still in a folder called `CheckboxWidgetV2`, so it should be easy to tell which one is v2 and v1

app/client/src/widgets/CheckboxWidgetV2/config/propertyPaneConfig/contentConfig.ts
jsartisan: We can remove this. I am not even sure what is the use of this prop in v1.

jsartisan: @riodeuno can you tell us do we need this prop or not? 🤔 I never understood the use of this prop.

app/client/src/widgets/CheckboxWidgetV2/config/autoLayoutConfig.ts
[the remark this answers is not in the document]
riodeuno: Ideally widgets should be agnostic of the layout system in which they're going to be used. However, the fact that widgets are an interface that connects the platform to the underlying components, and the way the current layout system architecture works, we will need to provide a few specific configurations that apply in specific layout systems.

Now, as WDS is being developed specifically for Anvil, and the fact that we don't need them to work with Fixed or Auto Layout, we can provide layout system specific configurations in a property called `anvil`, or as top level properties. @KelvinOm I'd recommend that you make a choice and more forward with this, changing the structure of configurations in the future is not a technical hurdle.

`minWidth` and `minHeight` are properties that may be necessary to provide [CSS based reflow](https://github.com/appsmithorg/appsmith/blob/14f4bae198851dc8694d9bf547ab11be38aec243/app/client/src/layoutSystems/anvil/common/AnvilFlexComponent.tsx#L124) in the anvil layout system. 

So, far our running theory for `minWidth` is as follows (please note that this is subject to change)
- Widgets like Input Widget will take up 100% of the provided space. We can place Input widget side by side in a row. The Layout system will not wrap any of widgets as 100% can be 2px, accommodating 10 input widgets in a box of width ~200px. Input widgets are unlikely to be use-able at this dimension. 

One alternative is for the Input widget component to have a `min-width`. As the layout system places widgets using flexbox, the widgets should ideally wrap.

Another alternative is for the input widget to let the layout system know of its `min-width` by using the widget configurations. This will allow the layout system to apply the same `min-width` to the bounding box in which the widget is placed, and the layout system only uses flexbox at the "bounding box" level. Widgets will wrap properly.

- Another type of widget is the one that simply uses the content for its width. Like the button widget. In this scenario, the layout system simply tries to fit the content (widget) using flexbox. If it doesn't fit, a wrap occurs. In this case as well, we can have both alternatives of `min-width` in component or `minWidth` as a configuration to be applied to the bounding box.

- Another type of widget is a widget that uses container queries. For example, a widget that changes their layout based on the available width in the "bounding box". Here, the widget component can still have the `min-width` defined, and the layout system simply tries to place a widget in a manner that it sees fit. These can be called "large" widgets.

- Another type of widget is a widget that can be manually or semantically resized. For example, the image widget. In this scenario, the control of the width of the widget is with the widget properties. However, the layout system is the one that defines how to enable resizing (in case of manual) for the widget. This means that the widget must specify constraints for resizing. Is it resizing while maintaining aspect ratio? What is the minimum width it can be resized to, what is the minimum height it can be resized to? etc. So, these properties will now have to be communicated via the widget interface as a part of the widget configurations. Widgets will simply fill the boxes that are the result of resizing.

Based on the above, we're likely to need the following widget configurations for the layout system to function appropriately:
1. widget type - 100% width | fit-contents | resizable | large  
2. min width - to add constraints on the bounding boxes 
3. min height - possibly necessary only for manually resizable widgets.

We also need to understand the implications of vertical alignment. This is likely to be w.r.t zones and may have nothing to do with widgets themselves.

We need to develop this theory and discuss to finalise the properties, but I will recommend that all new WDS widgets have placeholders for the above two properties in their configuration. We will conclude on this in the next 2 weeks.

cc @prsidhu

app/client/src/widgets/CheckboxWidgetV2/config/propertyPaneConfig/contentConfig.ts
riodeuno: Good question. I am not a 100% sure. I reckon the idea behind this is to understand if we need to show a "skeleton component" instead of the widget while any of the widgets properties are being evaluated. There must be a rationale behind making this user configurable. @dilippitchika may have more context.

marks0351: yea this prop is to deliberately allow the widget to be used while its waiting on some api to be loaded. by default if bound to an api the widget is in loading state/skeleton widget and is not intractable until the api is returned.

app/client/src/widgets/CheckboxWidgetV2/config/autoLayoutConfig.ts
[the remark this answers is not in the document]
KelvinOm: @riodeuno Thank you for the detailed answer. I agree with everything you say, but I don't understand yet how we can better make a config for Anvil. We will solve this as a separate task.

app/client/src/components/wds/constants.ts
```diff
@@ -1,4 +1,5 @@
 export const WDS_V2_WIDGET_MAP = {
-  BUTTON_WIDGET: "BUTTON_WIDGET_V2",
-  INPUT_WIDGET_V2: "INPUT_WIDGET_V3",
+  BUTTON_WIDGET: "WDS_BUTTON_WIDGET",
+  INPUT_WIDGET_V2: "WDS_INPUT_WIDGET",
```
KelvinOm, line 3: Just `INPUT_WIDGET`?

KelvinOm (approved):
The code looks good to me. Still, I want to understand what we will do with these properties.
Снимок экрана 2023-10-05 в 11 32 23

сс @ichik @riodeuno

riodeuno:
> The code looks good to me. Still, I want to understand what we will do with these properties. Снимок экрана 2023-10-05 в 11 32 23
> 
> сс @ichik @riodeuno

@KelvinOm This property will be useless in Anvil. All widgets will be either Auto, or resized (manually or semantically). Let's remove them.

app/client/src/components/wds/constants.ts
ichik: Isn't this only relevant for the previous version of Auto Layout that is going to be deprecated?

jsartisan: This is just a map that stores the list of wds widgets.

dhruvikn (dismissed):

shadabbuchh:
**Bugs in this PR:**

1. Long label name doe not get displayed properly. (This usually happens when there is no space in between the label)
Screenshot 2023-10-06 at 11 58 04 AM

2. Focus state visible upon checking/unchecking the box.
Screenshot 2023-10-06 at 12 03 21 PM

3. When `Required` is enabled & the checkbox is not selected, we do not see an error/red highlight around the widget.
Screenshot 2023-10-06 at 3 34 13 PM

4. When `Visible` is turned off and then we deploy the app, we get to see the widget for a fraction of a second and then it vanishes from the screen.

https://github.com/appsmithorg/appsmith/assets/39921438/f25a374d-50e5-4572-b45b-b407ab9a5c50

app/client/packages/design-system/widgets/src/components/Checkbox/src/Checkbox.tsx
```diff
@@ -25,7 +25,7 @@ const _Checkbox = (props: CheckboxProps, ref: HeadlessCheckboxRef) => {
         inlineLabelStyles["inline-label"],
       )}
     >
-      {Boolean(children) && <Text>{children}</Text>}
+      {Boolean(children) && <Text wordBreak="break-all">{children}</Text>}
```
KelvinOm, line 28: Do we really need an additional prop for this? Let's just add a CSS and that's it. It's also better to use `overflow-wrap: break-word` instead of `word-break: break-word`

KelvinOm (changes requested):

app/client/packages/design-system/widgets/src/components/Text/src/styles.module.css
```diff
@@ -4,6 +4,7 @@
   /* using auto instead of 100% so that text doesnot push everything to the edge */
   width: auto;
   color: inherit;
+  overflow-wrap: anywhere;
```
KelvinOm, line 7: Why `anywhere`?

app/client/packages/design-system/widgets/src/components/Text/src/types.ts
```diff
@@ -36,4 +36,6 @@ export interface TextProps {
   className?: string;
   /** The children of the component. */
   children: ReactNode;
+  /** sets word break */
+  wordBreak?: "normal" | "break-all" | "keep-all" | "break-word";
```
KelvinOm, line 40: This is not needed anymore

KelvinOm (changes requested):

app/client/packages/design-system/widgets/src/components/Text/src/styles.module.css
jsartisan: wanted to break by each character. In the QA review, shadab tested the label with single big word like LabelLabelLabelLabelLablLabelLabe, so overflow-wrap: break-word won't break it since it's a single word only.

shadabbuchh:
1. Long label issues is fixed
Screenshot 2023-10-12 at 10 40 53 AM

2. Created a separated bug for the focus state - https://github.com/appsmithorg/appsmith/issues/27921
3. Required state issues is resolved.
Screenshot 2023-10-12 at 10 42 30 AM


> 4. When Visible is turned off and then we deploy the app, we get to see the widget for a fraction of a second and then it vanishes from the screen.

As discussed this happens due to Skeleton loading & is expected
Cc: @jsartisan

app/client/packages/design-system/widgets/src/components/Text/src/types.ts
jsartisan: removed.

KelvinOm (approved):

PR #11001, opened 2022-02-08
fix: added new library and reduced function calls for global data

ApekshaBhosale:
## Description
Performance issue - App was taking a lot of time to create the first data Tree
solution - 
1. As part of Promises, we were creating global data to provide context for each evaluation which clones data tree each time. In this, we are limiting the number of times the data tree will be cloned by only adding platform functions for triggered paths. 
2. We were using cloneDeep from loadash which is not the faster one. Now we will be using rfdc https://www.npmjs.com/package/rfdc. 
  BenchMarking - 
  benchDeepCopy*100: 457.568ms
  benchLodashCloneDeep*100: 1230.773ms
  benchCloneDeep*100: 655.208ms
  benchFastCopy*100: 747.017ms
  benchRfdc*100: 281.018ms
  benchRfdcProto*100: 277.265ms
  benchRfdcCircles*100: 328.148ms
  benchRfdcCirclesProto*100: 323.004ms

Fixes #

## Type of change
- Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

> Please describe the tests that you ran to verify your changes. Provide instructions, so we can reproduce.
> Please also list any relevant details for your test configuration.

- Test A
- Test B

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

## Test coverage results :test_tube:
<details><summary>:green_circle: Total coverage has increased</summary>

    // Code coverage diff between base branch:release and head branch: object-create-perf 
Status | File | % Stmts | % Branch | % Funcs | % Lines 
 -----|-----|---------|----------|---------|------ 
 :green_circle: | total | 55.06 **(0)** | 36.39 **(0.01)** | 34.71 **(0.01)** | 55.44 **(0.01)**
 :green_circle: | app/client/src/utils/autocomplete/TernServer.ts | 52.94 **(0.23)** | 41.67 **(0.84)** | 36.21 **(0)** | 56.99 **(0.25)**
 :green_circle: | app/client/src/workers/Actions.ts | 91.67 **(0.12)** | 68.85 **(0)** | 80 **(0)** | 90.63 **(0.15)**
[...]

PR #27722, opened 2023-09-29
chore: prevent mutation during evaluation and execution

### LINKED ISSUE #19221: [Bug]-[80]:app goes into bad state when we use Assignment operator instead of equality comparison operator in text widget

### Description

https://www.loom.com/share/b62e126011a24ed394502dded9171ba1

App goes into bad state when we use "=" instead of "==" in the bindings with text widget and the value is an integer

### Steps To Reproduce

1. DnD a text widget
2. Dnd any other widget and use the binding anywhere "{{Text1.text=1}}"
3. app crashes

### Public Sample App

_No response_

### Issue video log

_No response_

### Version

Production

arunvjn:
## Description
Fixes the data mutation bug introduced back in [#11001](https://github.com/appsmithorg/appsmith/pull/11001). 
We now create 2 data tree in evaluation flow, one that is set against the execution context (prone to mutation) and another one which only gets the values of the path that is being evaluated. 

Accidental mutation often crashes the app. To replicate  the issue
1. D&D a text widget and an Input widget
2. In the input widget default text field, write `{{ Text1.text = 123 }}`

Context - https://theappsmith.slack.com/archives/C02K0SZQ7V3/p1695975703880489

Fixes #19221
>
>
#### Type of change
>
>
- [x] Manual
- [x] Jest
>
>
#### Test Plan
- [x] Test evaluations across widgets, framework functions, setter methods
- [x] Existing application sanity check
- [x] Template apps and CRUD
>
>
#### Issues raised during DP testing
none
>
>
>
#### Dev activity

#### QA activity:
- [x] [Speedbreak features](https://github.com/appsmithorg/TestSmith/wiki/Guidelines-for-test-plans#speedbreakers-) have been covered
- [x] Test plan covers all impacted features and [areas of interest](https://github.com/appsmithorg/TestSmith/wiki/Guidelines-for-test-plans#areas-of-interest-)


ohansFavour (dismissed):
The changes look good

vsvamsi1:
Hi @arunvjn  can you check the MISC_OPEN_SIMPLE_MODAL, the eval_scripting increased by 800ms.

app/client/src/workers/common/DataTreeEvaluator/index.ts
```diff
@@ -936,11 +937,15 @@ export default class DataTreeEvaluator {
...
     staleMetaIds: string[];
   } {
-    const tree = klona(oldUnevalTree);
-    updateTreeWithData(tree, klonaJSON(DataStore.getDataStore()));
+    const safeTree = klona(unEvalTree);
+    const dataStore = DataStore.getDataStore();
+    const dataStoreClone = klonaJSON(dataStore);
 
+    updateTreeWithData(safeTree, dataStoreClone);
+    updateTreeWithData(unEvalTree, dataStore);
```
Rishabh-Rathod, line 945: Shall we move the cloning of tree and updateTreeWithData outside of `evaluateTree`?

Rishabh-Rathod (approved):
LGTM

app/client/src/workers/common/DataTreeEvaluator/index.ts
```diff
@@ -949,248 +954,256 @@ export default class DataTreeEvaluator {
...
-              return set(currentTree, fullPropertyPath, evalPropertyValue);
+            continue;
           }
-        },
-        tree,
-      );
+          default:
+            set(currentTree, fullPropertyPath, evalPropertyValue);
+            set(safeTree, fullPropertyPath, klona(evalPropertyValue));
+            continue;
```
vsvamsi1, line 1189: I was checking under what case it would it enter here, it appears when we are loading the app then we are cloning the widget state, can we consider that as the worse case time complexity.

app/client/src/workers/common/DataTreeEvaluator/index.ts
```diff
@@ -949,248 +954,256 @@ export default class DataTreeEvaluator {
...
 
-              staleMetaIds = staleMetaIds.concat(
-                getStaleMetaStateIds({
-                  entity: widgetEntity,
-                  entityConfig: entityConfig as WidgetEntityConfig,
-                  propertyPath,
-                  isNewWidget,
-                  metaWidgets,
+            const evalPath = getEvalValuePath(fullPropertyPath);
+            const valueForSafeTree = klona(evalPropertyValue);
```
vsvamsi1, line 1119: In terms of time complexity this statement makes it worser than earlier.

app/client/src/workers/Evaluation/handlers/evalExpression.ts
```diff
@@ -8,5 +9,5 @@ export default function (request: EvalWorkerASyncRequest) {
   const evalTree = dataTreeEvaluator?.evalTree;
   const configTree = dataTreeEvaluator?.configTree;
   if (!evalTree || !configTree) return {};
-  return evaluateAsync(expression, evalTree, configTree, {}, undefined);
+  return evaluateAsync(expression, klona(evalTree), configTree, {}, undefined);
```
vsvamsi1, line 12: When does this get executed and what is the difference between evalTrigger and evalExpression? Is there anyway way we can avoid this klona?

app/client/src/workers/Evaluation/handlers/evalTrigger.ts
```diff
@@ -34,7 +35,7 @@ export default async function (request: EvalWorkerASyncRequest) {
 
   return dataTreeEvaluator.evaluateTriggers(
     dynamicTrigger,
-    evalTree,
+    klona(evalTree),
```
vsvamsi1, line 38: Can you check if this necessary since in the previous statement "evalAndValidateSubTree" does perform a klona on the evalTree and sets it to itself.

app/client/src/workers/common/DataTreeEvaluator/index.ts
```diff
@@ -338,9 +338,10 @@ export default class DataTreeEvaluator {
 
     const evaluationOrder = this.sortedDependencies;
 
+    const unEvalTreeClone = klona(this.oldUnEvalTree);
```
vsvamsi1, line 341: Can you check if this is necessary since we are performing the deepClone in the evaluateTree ( line 940)

app/client/src/workers/common/DataTreeEvaluator/index.ts
arunvjn: That's right. We didn't clone evaluated values of actions before but we need 2 copies to prevent mutation

app/client/src/workers/common/DataTreeEvaluator/index.ts
arunvjn: Yes this is necessary.

app/client/src/workers/Evaluation/handlers/evalTrigger.ts
arunvjn: Yes this is necessary.

app/client/src/workers/Evaluation/handlers/evalExpression.ts
arunvjn: This is a relatively infrequent operation compared to `evalTrigger`. Gets called for evaluating sub fields of action selector today

app/client/src/workers/common/DataTreeEvaluator/index.ts
vsvamsi1: My bad i noticed that line 946 may mutate it 
``` 
    const currentTree = unEvalTree;

```
In that case can do a deepClone of unEvalTree in line 946 and not safeTree in line 940. We can make regular updates to safeTree either through immer or immutable setIn method https://github.com/lodash/lodash/issues/1696 .

app/client/src/workers/common/DataTreeEvaluator/index.ts
```diff
@@ -949,248 +954,256 @@ export default class DataTreeEvaluator {
...
-            evalPropertyValue,
-            fullPropertyPath,
-          );
+        const entityType = entityConfig.ENTITY_TYPE;
 
-          const entityType = entityConfig.ENTITY_TYPE;
+        if (!propertyPath) {
+          set(currentTree, fullPropertyPath, evalPropertyValue);
+          set(safeTree, fullPropertyPath, klona(evalPropertyValue));
+          continue;
```
vsvamsi1, line 1039: I was checking under what case it would it enter here, it appears when we are loading the app then we are cloning the widget state.

app/client/src/workers/common/DataTreeEvaluator/index.ts
```diff
@@ -1714,7 +1730,7 @@ export default class DataTreeEvaluator {
       );
       return this.getDynamicValue(
         `{{${replacedBinding}}}`,
-        this.evalTree,
+        klona(this.evalTree),
```
vsvamsi1, line 1733: When does evalActionBindings get called? , if there are several bindings we are cloning the entire data tree for each binding. Is there something we can do to avoid this?

app/client/src/workers/common/DataTreeEvaluator/index.ts
arunvjn: `evaluateTree` gets called in both first tree and update tree flow, deep clone at 946 will make things worse

app/client/src/workers/common/DataTreeEvaluator/index.ts
arunvjn: Can you please elaborate on how this would help ?

app/client/src/workers/common/DataTreeEvaluator/index.ts
arunvjn: This is a nice catch, will need fixing

app/client/src/workers/common/DataTreeEvaluator/index.ts
arunvjn: I have removed these lines, we don't need them

app/client/src/workers/common/DataTreeEvaluator/index.ts
Rishabh-Rathod: It is not expected that the evaluateTree method would end up creating `safeTree` or updateUnEvalTreeWithData from its name. It could be better we move those operation outside the `evaluateTree`

app/client/src/workers/common/DataTreeEvaluator/index.ts
```diff
@@ -338,9 +341,10 @@ export default class DataTreeEvaluator {
 
     const evaluationOrder = this.sortedDependencies;
 
+    const unEvalTreeClone = klonaJSON(this.oldUnEvalTree);
```
ohansFavour, line 344: Is there a reason you've decided to use klonaJSON as opposed to klona/full here?. My perception is that we use klonaJSON when we are sure that the input will always be a valid JSON. We can't say that for the unEvalTree.

app/client/src/workers/common/DataTreeEvaluator/index.ts
```diff
@@ -949,248 +968,270 @@ export default class DataTreeEvaluator {
...
+            set(contextTree, fullPropertyPath, evalPropertyValue);
+            set(safeTree, fullPropertyPath, klona(evalPropertyValue));
+            break;
+          }
+          case ENTITY_TYPE_VALUE.JSACTION: {
+            const variableList =
+              (entityConfig as JSActionEntityConfig).variables || [];
+            if (variableList.indexOf(propertyPath) === -1) break;
+
+            const prevEvaluatedValue = get(safeTree, fullPropertyPath);
```
ohansFavour, line 1149: Need some clarity here.

While generating the evaluationOrder (see getEvaluationOrder), we usually set the value of paths to be evaluated back to their uneval value. So, I expect the safeTree at this point to contain the uneval value for the fullPropertyPath.

From the name of this variable "prevEvaluatedValue", what we want here is an "evaluated" value and I don't think this change will give us that.

Correct me if I am wrong.

app/client/src/workers/common/DataTreeEvaluator/index.ts
arunvjn: Can the first tree contain anything else apart from primitives ? Can you help me with a case here ?

app/client/src/workers/common/DataTreeEvaluator/index.ts
arunvjn: Yes, you are right. Nice catch

app/client/src/workers/common/DataTreeEvaluator/index.ts
```diff
@@ -949,248 +968,270 @@ export default class DataTreeEvaluator {
...
+            const evalPath = getEvalValuePath(fullPropertyPath, {
+              isPopulated: true,
+              //what is the purpose of this argument
+              fullPath: true,
+            });
+
+            /** Variables defined in a JS object are not reactive.
+             * Their evaluated values need to be reset only when the variable is modified by the user.
+             * When uneval value of a js variable hasn't changed, it means that the previously evaluated values are in both trees already  */
+            if (skipVariableValueAssignment) {
```
ohansFavour, line 1182: If variables are not reactive, then they should not even be part of the evaluationOrder unless a user modifies their unevalValue.

Today, in the dependencyMap, variables depend on other properties and that is why they are showing up in the evaluationOrder when not necessary. Could we consider making this improvement in the dependencyMap rather than here (after we've needlessly evaluated it).

arunvjn: I agree, we should take this up in another PR since this an existing behaviour

app/client/src/workers/common/DataTreeEvaluator/index.ts
ohansFavour: True. I can't think of any case for now.

app/client/src/workers/common/DataTreeEvaluator/index.ts
ohansFavour: All right

ohansFavour (approved):

ramsaptami:
LGTM
cc: @arunvjn

ramsaptami (approved):

PR #27728, opened 2023-09-30
test: Cypress| Fix omnibar test

Parthvi12:
## Description
Doc link got updated which was causing test to fail, fixed it.

locally

### DIFF
--- a/app/client/cypress/e2e/Regression/ClientSide/OtherUIFeatures/Omnibar_spec.js
+++ b/app/client/cypress/e2e/Regression/ClientSide/OtherUIFeatures/Omnibar_spec.js
@@ -20,7 +20,7 @@ describe("Omnibar functionality test cases", () => {
     agHelper.Sleep(2000);
     deployMode.StubWindowNAssert(
       '//span[text()="Learn more"]',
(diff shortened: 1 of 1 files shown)


Parthvi12:
@Aishwarya-U-R Can we force merge it to unblock CI, tests are passing locally.

Screenshot 2023-09-30 at 7 24 15 PM

app/client/cypress/e2e/Regression/ClientSide/OtherUIFeatures/Omnibar_spec.js
```diff
@@ -20,7 +20,7 @@ describe("Omnibar functionality test cases", () => {
     agHelper.Sleep(2000);
     deployMode.StubWindowNAssert(
       '//span[text()="Learn more"]',
-      "connect-datasource",
+      "connect-to-a-database",
```
mohanarpit, line 23: @Parthvi12 I am not sure if this is a good fix. In this scenario, we are relying on documentation team to not change links. Also, this test will fail whenever the docs team chooses to reorder the pages. I'd suggest putting in a better fix here such as checking for 200 OK.

Issue #27523, opened 2023-09-20
[Bug]: Helm chart uses images with vulnerabilities


felixthekraut:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

When scanning using Amazon's ECR service, a number of Critical and High vulnerabilities are flagged in third party images used in the Appsmith helm chart.

`helm pull appsmith/appsmith --untar --namespace appsmith`

The following images need to be updated:
1 year old: `bitnami/redis-cluster:6.2.7-debian-10-r23 -> bitnami/redis-cluster:7.0.13-debian-11-r10`

4 months old: `bitnami/mongodb:5.0.17-debian-11-r5 -> bitnami/mongodb:5.0.21-debian-11-r5`

1 year old: `bitnami/mongodb-sharded:5.0.9-debian-10-r0 -> bitnami/mongodb-sharded:5.0.21-debian-11-r5`

Here are the scan outputs:

[appsmith-ce-latest.json.txt](https://github.com/appsmithorg/appsmith/files/12677663/appsmith-ce-latest.json.txt)
[redis-6.2.7-debian-10-r23.json.txt](https://github.com/appsmithorg/appsmith/files/12677664/redis-6.2.7-debian-10-r23.json.txt)
[redis-6.2-alpine.json.txt](https://github.com/appsmithorg/appsmith/files/12677665/redis-6.2-alpine.json.txt)
[mongodb-5.0.17-debian-11-r5.json.txt](https://github.com/appsmithorg/appsmith/files/12677666/mongodb-5.0.17-debian-11-r5.json.txt)
[mongodb-5.0.9-debian-10-r0.json.txt](https://github.com/appsmithorg/appsmith/files/12677667/mongodb-5.0.9-debian-10-r0.json.txt)

### Steps To Reproduce

Scan Appsmith helm chart in Amazon ECR

### Environment

Production

### Version

Appsmith Community v1.9.33

Mershab99:
Hi, this is also a blocker for ARM based deployments, as this, and the 'Chart.lock' file use versions of Redis and MongoDB Bitnami chart dependencies, that themselves do not link to ARM images.

PR #27733, opened 2023-10-02
Fix images with vulnerabilities

pratapaprasanna:
Fixes: [27523](https://github.com/appsmithorg/appsmith/issues/27523)

### DIFF
--- a/deploy/helm/values.yaml
+++ b/deploy/helm/values.yaml
@@ -8,6 +8,8 @@ redis:
   replica:
     replicaCount: 1
     nodeSelector: {}
+  image:
+    tag: 7.0.13-debian-11-r10
 
 mongodb:
   enabled: true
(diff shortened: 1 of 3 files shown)


deploy/helm/values.yaml
sharat87, line 1: I see we're adding the `image` fields here. For existing users, their `values.yml` won't have this, right? Will the new Helm chart work fine for that, or are these required?

felixthekraut: I don't think that this will be a problem. Helm should add the new image tags to existing setups by updating the values.yaml automatically, then proceed to template rendering, deployment, rolling update, versioning, etc. The versioning allows rollback as well.

pratapaprasanna: There will be no problem when they update the helm chart the new images will be pulled and used to deploy.

sharat87 (approved):

PR #27749, opened 2023-10-02
chore: Add CSP for eval protection

sharat87:
WIP Experiment.


rajatagrawal:
I think that our frontend is a single page app and most of the APIs are application/json. Add a CSP header to APIs will not have any effect because the HTML isn't served from API response. All the HTML is served from frontend and API response, which is JSON format, is used to only fill the data.

PR #27827, opened 2023-10-04
chore: replace emotion/sheet to emotion/css

### LINKED ISSUE #27750: Improve Flex component performance

KelvinOm:
## Description
Replace emotion/sheet to emotion/css. 

**Motivation**
During performance testing, it was revealed that creating a separate class for each component is slow.

Rendering 10k elements
Снимок экрана 2023-10-05 в 13 34 12

As part of the task, an approach using styles through attributes was also tested, it works well, but we still need to create a separate class in runtime if we work with container queries. Because of this, I think it's best to use a ready-made solution from emotion/css, since it creates only one class out of the box for elements with the same styles, which allows to render elements with good performance.

Rendering 80k elements
Снимок экрана 2023-10-05 в 13 28 33

Fixes #27750 

#### Type of change

- [x] Manual
- [ ] JUnit
- [x] Jest
- [ ] Cypress

#### Dev activity


app/client/packages/design-system/widgets/src/components/Flex/src/dimensions.ts
```diff
@@ -115,5 +114,4 @@ export type SizingDimension =
   | "sizing-99"
   | "sizing-100"
   // eslint-disable-next-line @typescript-eslint/ban-types
-  | (string & {})
-  | number;
+  | (string & {});
```
KelvinOm, line 117: I deleted the number here, since in fact we do not support numbers, only custom strings. Because of this, I was forced to fix the places where the flex component used in Anvil.

cc @prsidhu

app/client/packages/design-system/widgets/src/components/Flex/src/flexCss.ts
```diff
@@ -0,0 +1,145 @@
...
+  if (value == null) return;
+
+  if (typeof value === "object" && !Array.isArray(value)) {
+    return Object.keys(value).reduce((prev, current) => {
+      if (current !== "base") {
+        return (
+          prev +
+          `@container (min-width: ${current}) {& {
+          ${cssProp}: ${
+            //@ts-expect-error: type mismatch
```
KelvinOm, line 96: @jsartisan If you know how to specify the types better here, please help me.

jsartisan (approved):

PR #27841, opened 2023-10-05
Stabilise server suite tests

nsarupr:
stabilise server suite tests


app/server/appsmith-server/src/test/java/com/appsmith/server/services/ActionCollectionServiceTest.java
```diff
@@ -199,9 +202,12 @@ public void setup() {
     @AfterEach
     @WithUserDetails(value = "api_user")
     public void cleanup() {
```
trishaanand, line 204: Clean up can be skipped no? Each user would create a new workspace. Unless we are counting workspaces/applications somewhere in the test.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ApplicationPageServiceTest.java
```diff
@@ -52,26 +56,47 @@ public class ApplicationPageServiceTest {
...
+
+    @BeforeEach
+    public void setup() {
+        Workspace unsavedWorkspace = new Workspace();
+        unsavedWorkspace.setName("ApplicationPageServiceTest Workspace");
+        workspace = workspaceService.create(unsavedWorkspace).block();
+    }
+
+    @AfterEach
+    public void cleanup() {
```
trishaanand, line 75: Same. If we are creating a new workspace per test, we can actually skip this right?

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ApplicationSnapshotServiceTest.java
```diff
@@ -40,21 +44,53 @@ public class ApplicationSnapshotServiceTest {
...
+
+    @Autowired
+    SessionUserService sessionUserService;
+
+    Workspace workspace;
+
+    @BeforeEach
+    public void setup() {
+        User user = sessionUserService.getCurrentUser().block();
+        if (null == user) {
```
trishaanand, line 61: Why was this required? Are there test cases which are running without a user? Do we need to do this for other files as well?

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ce/ActionServiceCE_Test.java
```diff
@@ -183,84 +195,84 @@ public class ActionServiceCE_Test {
...
-                            workspaceId, tuple.getT2(), tuple.getT1().getId(), gitData.getBranchName()))
-                    .block();
-
-            gitConnectedPage = newPageService
-                    .findPageById(gitConnectedApp.getPages().get(0).getId(), READ_PAGES, false)
-                    .block();
-
-            branchName = gitConnectedApp.getGitApplicationMetadata().getBranchName();
-        }
+        //        if (workspaceId == null) {
```
trishaanand, line 206: Please remove commented code.

trishaanand (commented):
Partial review. Added comments which should be valid for rest of the files as well.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ce/ActionServiceCE_Test.java
nsarupr: Removed commented code.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ActionCollectionServiceTest.java
nsarupr: Not skipping the cleanup post discussion.
Discussion:
```In the past, I have seen some test cases, where Test1 is failing a count assertion, only because Test2 created some resources. Just for that sake, I am trying to add a cleanup for applications and workspaces. This will be doing 2 things at once, (a) remove the unnecessary resources and (b) clean cache for the existing users, because deleting workspace will eventually delete the roles.```

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ApplicationSnapshotServiceTest.java
abhvsn: @nsarupr do we need this to be a de-facto standard for other test classes as well, considering devs may add the testcases which does not require user context just like the ones we are seeing here?

abhvsn (commented):
Very minor comments

abhvsn (approved):

trishaanand (approved):

PR #27852, opened 2023-10-05
chore: add strict-boolean-expressions rule

KelvinOm:
## Description
Add `@typescript-eslint/strict-boolean-expressions` eslint rule for packages.


app/client/package.json
```diff
@@ -34,7 +34,7 @@
     "generate:widget": "plop --plopfile generators/index.js",
     "postinstall": "node cypress/apply-patches.js && yarn init-husky",
     "storybook": "yarn workspace @design-system/storybook storybook",
-    "lint": "eslint --cache --ext .js,.ts,.tsx ./src && eslint --cache --ext .js,.ts ./cypress",
+    "lint": "eslint --cache ./src && cd ./cypress && eslint -c .eslintrc.json --cache ./",
```
KelvinOm, line 37: Removed the extension indication as it is covered by default. I also had to change the script for cypress checks, as I added a path to the `tsconfig` and because of this there were conflicts in the eslint configurations.

cc @Aishwarya-U-R

app/client/packages/rts/.eslintrc.json
```diff
@@ -1,7 +1,8 @@
 {
   "extends": ["../../.eslintrc.base.json"],
-  "ignorePatterns": ["dist"],
+  "ignorePatterns": ["dist", "server.test.ts"],
```
KelvinOm, line 3: `server.test.ts` is excluded in `tsconfig` so we have to exclude it here as well.

Aishwarya-U-R:
> https://github.com/appsmithorg/appsmith/actions/runs/6429347315

@KelvinOm - Please always merge with latest release & then start run. AutoHeight_Form_spec.ts is fixed & merged now.

app/client/cypress/.eslintrc.json
```diff
@@ -1,9 +1,17 @@
 {
   "extends": ["../.eslintrc.base.json"],
+  "ignorePatterns": [
+    "locators",
+    "fixtures",
+    "AudioRecorder_spec.js",
+    "audio_spec.js",
+    "Mongo_Spec.js"
+  ],
```
dhruvikn, line 9: Just wanted to understand why are we ignoring these files?

dhruvikn (changes requested):

KelvinOm:
> > /ok-to-test
> 
> @KelvinOm - Please always merge with latest release & then start run. AutoHeight_Form_spec.ts is fixed & merged now.

@Aishwarya-U-R Yeah, I'm practicing it. The previous launch was yesterday.

app/client/cypress/.eslintrc.json
KelvinOm: @dhruvikn Thanks for the question. I completely forgot about it.

@Aishwarya-U-R Lint can't handle the situation when we have .ts and .js files with same name. Can we remove the JS implementation?

Снимок экрана 2023-10-06 в 12 07 01

app/client/package.json
KelvinOm: I forgot to say that this is another reason to move the cypress to packages. There would be no such problem there.

app/client/cypress/.eslintrc.json
Aishwarya-U-R: @KelvinOm - Can you rename .ts to `AudioRecorder_Spec.ts` - does this case account as different file? Else please rename Js to `AudioRecorder1_spec.js` & proceed.

Aishwarya-U-R: Same for Mongo_spec also.

dhruvikn (approved):

jsartisan (approved):

PR #27868, opened 2023-10-06
fix: remove redundant package builds

KelvinOm:
## Description
Remove redundant package builds

### DIFF
--- a/app/client/packages/dsl/package.json
+++ b/app/client/packages/dsl/package.json
@@ -10,9 +10,7 @@
     "test:jest": "jest --watch",
     "lint": "yarn g:lint",
     "prettier": "yarn g:prettier",
-    "build": "rollup -c",
-    "start": "rollup -c",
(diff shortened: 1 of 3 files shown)


app/client/packages/dsl/package.json
```diff
@@ -5,23 +5,13 @@
   "description": "",
   "author": "",
   "license": "ISC",
```
sharat87, line 7: Let's remove this License? 🙂

sharat87 (approved):
Minor thing, looks good.

brayn003:
Since we are removing package 'build' on 'postinstall', wouldn't this create issues when building rts?

KelvinOm:
> Since we are removing package 'build' on 'postinstall', wouldn't this create issues when building rts?
I checked it locally, I also don't see any problems with the `rts` build when we start the `ok-to-test` process.

sharat87 (approved):

PR #27892, opened 2023-10-09
chore: skip signposting discovery test

akash-codemonk:
Fixes https://github.com/appsmithorg/appsmith/issues/28034

The feature is behind feature flag and it has been turned off for all users. The feature is no longer shown to users, so we skip the test altogether. 

The test file is deleted along with the code for the feature.


Aishwarya-U-R:
@akash-codemonk - the test is to enable it via featureflag - '`ab_gif_signposting_enabled`' & test it right? And when will complete spec be deleted? - `The test file can deleted along with the code for the feature.` - Dint get this. 
also please add the skipping reason on top of spec in comments.

akash-codemonk:
> @akash-codemonk - the test is to enable it via featureflag - '`ab_gif_signposting_enabled`' & test it right? And when will complete spec be deleted? - `The test file can deleted along with the code for the feature.` - Dint get this. also please add the skipping reason on top of spec in comments.

Yes it is enabled by featureflag.

I think we can delete the spec now itself, but I think it is better to remove the feature from appsmith and then remove the test. Today the implementation of the feature exists but it is not shown to the user as we plan to remove it. I just though it will be faster to skip the test now and remove the implementation and test separately. 
@Aishwarya-U-R

Aishwarya-U-R:
@akash-codemonk - What you mean by remove feature? No more signpost walk-thru for user then?

akash-codemonk:
> @akash-codemonk - What you mean by remove feature? No more signpost walk-thru for user then?

Yes we have stopped showing this to users already.

Aishwarya-U-R:
Got it. In that case please decide to remove this spec also now only. Having it under skipped, will increase our skipped test cases count. 
Else move the complete spec to cypress/manual_TestSuite/CommentedScriptFiles - if you think you need time to decide if this can be removed or not.

hetunandu (approved):

PR #27903, opened 2023-10-09
fix: Making login rate limit service reactive

trishaanand:
Refactor : While debugging the production downtime, took a deeper look into the code path of rate limiting. Blocking calls to redis were being made on the main threads which can impact the overall response times of the appsmith server (since the number of main threads equals no of CPUs on the machine). No blocking calls should be made in the main thread. Moving executions to bounded elastic threadpool and making the reset flow reactive. Also the code was not split according to the code split guidelines of the server. Taking care of that as well.


nsarupr (approved):
Looks good to me.
@shubham7saxena7 's approval on [slack thread](https://theappsmith.slack.com/archives/C02K2MZERSL/p1696867470614649?thread_ts=1696856374.549379&cid=C02K2MZERSL)

shubham7saxena7 (approved):

PR #27962, opened 2023-10-11
update github actions for dp

pratapaprasanna:


.github/workflows/on-demand-build-docker-image-deploy-preview.yml
```diff
@@ -22,6 +22,17 @@ jobs:
             PR: ${{ github.event.client_payload.pull_request.number }}.
             recreate: ${{ github.event.client_payload.slash_command.args.named.recreate }}.
 
+      - name: Git checkout code
+        uses: actions/checkout@v3
+
```
sharat87, line 27: I don't think we need this, right?

.github/workflows/on-demand-build-docker-image-deploy-preview.yml
```diff
@@ -22,6 +22,17 @@ jobs:
...
+      - name: Fail dp incase of merge conflicts
+        run: |
+          status="$(gh pr view ${{ github.event.client_payload.pull_request.number }} --json mergeable --jq .mergeable)";
+          if [ $status = "CONFLICTING" ]; then
```
sharat87, line 31: Please prefer double square brackets.

```suggestion
          if [[ $status == "CONFLICTING" ]]; then
```

.github/workflows/on-demand-build-docker-image-deploy-preview.yml
```diff
@@ -22,6 +22,17 @@ jobs:
...
+             echo "Fail building dp as there are merge conflicts"
+             exit
```
sharat87, line 33: This won't _fail_, right? Should we do `exit 1` to get GitHub to stop running the following steps?

sharat87 (changes requested):

PR #27992, opened 2023-10-12
fix:lint staged

KelvinOm:
## Description

1. Optimise `lint-staged` scripts so that they are not executed twice. For some patterns, I removed the `prettier` checks since they are executed by`eslint` checks as well.
2. Add caching for all scripts.
3. Optimisation of the `git secrets`, now we check only staged files. Earlier we checked all the files that `git ls-files` returns. [Here more about this.](https://github.com/awslabs/git-secrets#options-for-scan).


ankitakinger (approved):
LGTM

rajatagrawal (approved):

PR #28144, opened 2023-10-17
chore: Added ce compatible layer for git

nidhi-nair:
Contains new layer for ce compatible to prevent merge conflicts on EE


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce_compatible/GitServiceCECompatible.java
```diff
@@ -0,0 +1,3 @@
+package com.appsmith.server.services.ce_compatible;
+
+public interface GitServiceCECompatible {}
```
abhvsn, line 3:
```suggestion
public interface GitServiceCECompatible extends GitServiceCE {}
```

abhvsn (changes requested):
Very tiny comment, otherwise LGTM

abhvsn (approved):

PR #28197, opened 2023-10-18
chore: migration changes

sondermanish:
## Description

- Added flags to resume migration in case of server restart during migration

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration032CE01AddMigrationFlagsForXmlParser.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration032CE01AddMigrationFlagsForXmlParser.java
@@ -1,0 +1,61 @@
+package com.appsmith.server.migrations.db.ce;
+
+import com.appsmith.server.constants.ApplicationConstants;
+import com.appsmith.server.domains.Application;
+import com.appsmith.server.domains.CustomJSLib;
+import com.appsmith.server.migrations.MigrationHelperMethods;
+import io.mongock.api.annotations.ChangeUnit;
+import io.mongock.api.annotations.Execution;
+import io.mongock.api.annotations.RollbackExecution;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.dao.DuplicateKeyException;
+import org.springframework.data.mongodb.core.MongoTemplate;
+import org.springframework.data.mongodb.core.query.Query;
+import org.springframework.data.mongodb.core.query.Update;
+
+/**
+ *  Appsmith provides xmlParser v 3.17.5 and few other customJSLibraries by default, xmlParser has been
+ *  flagged because it has some vulnerabilities. Appsmith is stopping natively providing support for xmlParser.
+ *  This however, would break existing applications which are using xmlParser. In order to prevent this,
+ *  applications require to have xmlParser as added library.
+ *
+ *  This migration takes care of adding a migration flag and adding a document in customJsLib for xml-parser
+ */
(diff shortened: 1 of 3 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration032CE03RemovingMigrationFlagPostJsLibAddition.java
```diff
@@ -0,0 +1,37 @@
...
+import org.springframework.data.mongodb.core.MongoTemplate;
+import org.springframework.data.mongodb.core.query.Query;
+import org.springframework.data.mongodb.core.query.Update;
+
+/**
+ * This migration removes the migration flag from application class, after other two migrations of same series have run
+ */
+@Slf4j
+@ChangeUnit(order = "032-03", id = "remove-xml-parser-flag-from-application", author = " ")
+public class Migration032CE03RemovingMigrationFlagPostJsLibAddition {
```
nidhi-nair, line 18: Can we combine unsetting with the previous migration? This is because we should make sure unsetting happens only for applications that the update has passed for.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration032CE01AddMigrationFlagsForXmlParser.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @RollbackExecution
+    public void rollbackExecution() {}
+
+    @Execution
+    public void addMigrationFlagToApplicationCollection() {
+        // Temporary addition of migration flag in Application Collection
+        mongoTemplate.updateMulti(
+                new Query().addCriteria(MigrationHelperMethods.notDeleted()),
```
nidhi-nair, line 42: From your comment it seems that we did not used to have xml parser as a default library before. Is it not possible to only find applications that were using xml parser library already?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration032CE03RemovingMigrationFlagPostJsLibAddition.java
sondermanish: got it, doing that

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration032CE01AddMigrationFlagsForXmlParser.java
sondermanish: This is required because now we are going ahead an deprecating native support for xmlParser,
The users could be using xmlParser in JsObjects , which would require string searches in another collection. hence adding it for all the applications

nidhi-nair: Sorry I don't think I understood. Do you mean it was possible for folks to use xmlParser earlier without including it in the customjslibs list?

sondermanish: yes, because xmlParser is shipped as a static file, which is being deprecated now.
image

nidhi-nair (approved):

nidhi-nair:
Please make changes to the exception handling before you merge.

PR #28198, opened 2023-10-18
chore: Use Docker base images to build

sharat87:
This will use the Appsmith base image which contains all the downloads needed, so the image build in daily CI should be much more reliable, and quite likely much faster.

All workflows have already been updated to use the BASE build argument to set the base image.

Once this is merged, building the Docker image will fail if `BASE` build argument is not passed. This is that the base is set explicitly everywhere to not cause any confusion.

The deleted content has been moved to https://github.com/appsmithorg/appsmith/blob/8d34a2ac280e2263851fd24e720d26811545b625/deploy/docker/base.dockerfile, which builds the `base-*:*` images.


trishaanand (approved):

Aishwarya-U-R:
@sharat87 - can you please help point me to this change?

`All workflows have already been updated to use the BASE build argument to set the base image.`

sharat87:
@Aishwarya-U-R:

- https://github.com/appsmithorg/appsmith/commit/988f60e7ff31537b99d6a849fc6c4db2001a9dd7
- https://github.com/appsmithorg/appsmith/commit/be63e1b404f192e2d57c55bce259ce143085ffd2

PR #28238, opened 2023-10-19
[DO NOT MERGE] - Experimental Anvil

riodeuno:
- This is a PR that is used to generate a user testable deploy preview for Anvil (and WDS).
- No code should be pushed directly to this branch.
- The only way to update this branch is to pull in the `release` branch and update the experimental tag version
- The experimental tag version is supposed to help us identify the state of the code using screenshots of the app from users.
- The experimental tag is the only piece of code specific to this branch and must never be merged into release.

Again, DONOT MERGE this PR.

### DIFF
--- a/app/client/public/index.html
+++ b/app/client/public/index.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+<!doctype html>
 <html lang="en">
   <head>
     <meta charset="utf-8" />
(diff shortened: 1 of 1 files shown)


app/client/src/layoutSystems/common/WidgetNamesCanvas/index.tsx
```diff
@@ -147,10 +147,13 @@ const OverlayCanvasContainer = (props: { canvasWidth: number }) => {
     );
 
     scrollParent.addEventListener("mousemove", mouseMoveHandler);
+    wrapperRef.current.addEventListener("mousemove", mouseMoveHandler);
```
riodeuno, line 150: @prsidhu @marks0351 I shouldn't be doing this. But, I've added a patch fix for the widget name drag. I'll have a proper fix for this in `release` tomorrow. Adding a comment to highlight this.

PR #28245, opened 2023-10-20
chore: Refactored import service to use type parameterized generic interface impls

nidhi-nair:
Contains changes to make sure that import service also has domain specific logic separated out into respective services


app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/ce/MappedImportableResourcesCE_DTO.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@NoArgsConstructor
+@Data
+public class MappedImportableResourcesCE_DTO {
+
+    Map<String, String> pluginMap = new HashMap<>();
```
nayan-rafiq, line 16: Usually, we don't use inline initialization. Is that initialization really required?

app/server/appsmith-server/src/test/resources/test_assets/ImportExportServiceTest/valid-application.json
nayan-rafiq, line 1: What's this change for? Will it impact user's applications if they have the older names in the application JSON?

nayan-rafiq (commented):
Looks good. Left few comments.

app/server/appsmith-server/src/test/resources/test_assets/ImportExportServiceTest/valid-application.json
nidhi-nair: I have modified the get plugins logic to only get valid plugins in the workspace, instead of getting all available plugins in the DB. That is the correct behaviour to use because the user trying to import an app might have been trying to import something into an invalid workspace. The tests needed to reflect this behaviour though, so the json has been modified toonly include the test plugins we have seeded the workspace with, that is, `installed-plugin` and `installed-js-plugin`.

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/ce/MappedImportableResourcesCE_DTO.java
nidhi-nair: It's to avoid NPEs around the import export flow. The original sequential flow was also initializing this map in the stack memory of the function call. I've tried to keep changes minimal to avoid breaking stuff I don't understand. :) Did you mean that I should init this in the default constructor manually?

nayan-rafiq (approved):

Issue #28259, opened 2023-10-20
[Task]: Handle connection rate limiting for given plugins


sneha122:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SubTasks

Whenever query is executed for following plugins and connection is not successful, we need to add rate limiting:
- MySQL
- SMTP
- Elastic Search
- ArangoDB
- Redis

Issue #27742, opened 2023-10-02
[Task]: Junit Test cases


sneha122:
### Is there an existing issue for this?

- [X] I have searched the existing issues

TanyaAgarwal28:
Hi @sneha122 , Can you please provide more insights on this issue? Will like to pick it up.

sneha122:
Hi @TanyaAgarwal28 This task is to write JUnit test cases for an ongoing feature of [Rate limit](https://github.com/appsmithorg/appsmith/issues/27745), This may a little complex to pick it up as feature is not complete yet, meanwhile please feel free to check out the [Good first issues](https://github.com/appsmithorg/appsmith/issues?q=is%3Aissue+is%3Aopen+label%3A%22Good+First+Issue%22)

TanyaAgarwal28:
Oh okay thanks

Issue #27740, opened 2023-10-02
[Task]: At the beginning of each connection initiation during query execution, check for key in cache and block execution if required


sneha122:
### Is there an existing issue for this?

- [X] I have searched the existing issues

PR #28260, opened 2023-10-20
feat: rate limiting added for connection creation during query execution

### LINKED ISSUE #27738: [Task]: Consume tokens on connection failure during query execution and on token exhaustion set key in cache

### SubTasks

_No response_

sneha122:
## Description

This PR adds rate limiting on connection creations. Connections with datasources are created in following situations:
- Whenever a query is executed, we create connection to datasource
- Whenever datasource structure is fetched
- Whenever trigger API is called in case of generate CRUD

In all above cases, if connection is successful, its context is saved in cache, but when the connection is unsuccessful, connection context is not saved in cache and connection creation is initiated every time. In such case, we need to apply rate limiting on connection creation, so that no brute force attacks can happen.

The code in this PR is written with the assumption that every time connection creation fails, each plugin throws an error, whenever we get this error, we consume token from bucket, if we exhaust this bucket, we block this IP for next 5 minutes. This error is thrown in following plugins:
- Postgres
- Oracle
- Redshift
- MsSQL

Other plugins like MySQL, SMTP, Elastic Search, ArangoDB, Redis, No such error is thrown, which can be caught and rate limiting applied, hence creating a different task to handle for these plugins. check #28259

Junit test cases will be covered in #27742 

Fixes #27738 , #27740 
>
>
#### Type of change
>
>
>
- [x] Manual
- [ ] JUnit
- [ ] Jest
- [ ] Cypress
>
>
#### Dev activity

#### QA activity:

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/datasources/base/DatasourceServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/datasources/base/DatasourceServiceCEImpl.java
@@ -615,7 +615,8 @@ public class DatasourceServiceCEImpl implements DatasourceServiceCE {
      * and will also be used for creating connections during query execution.
      * For more details: https://github.com/appsmithorg/appsmith/issues/22868
      */
-    protected Mono<String> getRateLimitIdentifier(DatasourceStorage datasourceStorage) {
+    @Override
+    public Mono<String> getRateLimitIdentifier(DatasourceStorage datasourceStorage) {
         Mono<PluginExecutor> pluginExecutorMono = pluginExecutorHelper
                 .getPluginExecutor(pluginService.findById(datasourceStorage.getPluginId()))
                 .switchIfEmpty(Mono.error(new AppsmithException(
@@ -625,7 +626,11 @@ public class DatasourceServiceCEImpl implements DatasourceServiceCE {
                 .getEndpointIdentifierForRateLimit(datasourceStorage.getDatasourceConfiguration()));
     }
 
-    protected Mono<Boolean> isEndpointBlockedForConnectionRequest(DatasourceStorage datasourceStorage) {
+    /*
+     * This method checks if the given endpoint is blocked for connection request or not.
+     */
+    @Override
+    public Mono<Boolean> isEndpointBlockedForConnectionRequest(DatasourceStorage datasourceStorage) {
         Mono<String> rateLimitIdentifierMono = this.getRateLimitIdentifier(datasourceStorage);
         return featureFlagService
                 .check(FeatureFlagEnum.rollout_datasource_test_rate_limit_enabled)
@@ -645,7 +650,11 @@ public class DatasourceServiceCEImpl implements DatasourceServiceCE {
                 });
     }
 
-    protected Mono<Boolean> consumeTokenIfAvailable(DatasourceStorage datasourceStorage) {
+    /*
+     * This method consumes a token from bucket if available, otherwise returns false.
+     */
+    @Override
+    public Mono<Boolean> consumeTokenIfAvailable(DatasourceStorage datasourceStorage) {
         Mono<String> rateLimitIdentifierMono = this.getRateLimitIdentifier(datasourceStorage);
         return featureFlagService
                 .check(FeatureFlagEnum.rollout_datasource_test_rate_limit_enabled)
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/datasources/base/DatasourceServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/datasources/base/DatasourceServiceCEImpl.java
@@ -549,8 +549,19 @@ public class DatasourceServiceCEImpl implements DatasourceServiceCE {
                                                     // This will block the test API for next 5 minutes, as bucket has
                                                     // been exhausted, and return too many requests response
                                                     return this.blockEndpointForConnectionRequest(datasourceStorage)
-                                                            .flatMap(isAdded -> Mono.just(tooManyRequests))
-                                                            .onErrorResume(error -> Mono.just(tooManyRequests));
+                                                            .flatMap(isAdded -> {
+                                                                if (isAdded) {
+                                                                    return Mono.just(tooManyRequests);
+                                                                } else {
+                                                                    return Mono.just(new DatasourceTestResult(
+                                                                            new AppsmithException(
+                                                                                            AppsmithError
+                                                                                                    .DATASOURCE_CONNECTION_RATE_LIMIT_BLOCKING_FAILED)
+                                                                                    .getMessage()));
+                                                                }
+                                                            })
+                                                            .onErrorResume(error -> Mono.just(
+                                                                    new DatasourceTestResult(error.getMessage())));
                                                 }
                                                 return Mono.just(datasourceTestResult);
                                             })
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/datasources/base/DatasourceServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/datasources/base/DatasourceServiceCEImpl.java
@@ -94,6 +94,8 @@ public class DatasourceServiceCEImpl implements DatasourceServiceCE {
     // This will block the creation of datasource connection for 5 minutes, in case of more than 3 failed connection
     // attempts
     private final Integer BLOCK_TEST_API_DURATION = 5;
+    private final AppsmithException tooManyRequestsException =
+            new AppsmithException(AppsmithError.TOO_MANY_FAILED_DATASOURCE_CONNECTION_REQUESTS);
 
     @Autowired
     public DatasourceServiceCEImpl(
@@ -514,9 +516,7 @@ public class DatasourceServiceCEImpl implements DatasourceServiceCE {
                         .flatMap(datasourceStorageService::checkEnvironment)
                         .flatMap(this::verifyDatasourceAndTest);
             } else {
-                AppsmithException exception =
-                        new AppsmithException(AppsmithError.TOO_MANY_FAILED_DATASOURCE_CONNECTION_REQUESTS);
-                return Mono.just(new DatasourceTestResult(exception.getMessage()));
+                return Mono.just(new DatasourceTestResult(tooManyRequestsException.getMessage()));
             }
         });
     }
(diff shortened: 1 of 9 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
```diff
@@ -252,20 +266,67 @@ public boolean isValidDatasourceContextAvailable(
...
-            if (isValidDatasourceContextAvailable(datasourceStorage, datasourceContextIdentifier)) {
-                log.debug("Resource context exists. Returning the same.");
-                return Mono.just(datasourceContextMap.get(datasourceContextIdentifier));
-            }
-        }
-        return createNewDatasourceContext(datasourceStorage, datasourceContextIdentifier);
+        Mono<String> rateLimitIdentifierMono = datasourceService.getRateLimitIdentifier(datasourceStorage);
+        return datasourceService
+                .isEndpointBlockedForConnectionRequest(datasourceStorage)
+                .flatMap(isBlocked -> {
```
sumitsum, line 272: Hi Sneha, should this check be put after the following if else clause: 
```
if (datasourceId == null) {
            log.debug(
                    "This is a dry run or an embedded datasourceStorage. The datasourceStorage context would not exist in this "
                            + "scenario");
        } else {
            if (isValidDatasourceContextAvailable(datasourceStorage, datasourceContextIdentifier)) {
                log.debug("Resource context exists. Returning the same.");
                return Mono.just(datasourceContextMap.get(datasourceContextIdentifier));
            }
        }
```
This clause returns a valid connection if it already exists. If not, then the next line creates a new connection via `createNewDatasourceContext` - i.e. should the check be placed just before this line ?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
```diff
@@ -252,20 +266,67 @@ public boolean isValidDatasourceContextAvailable(
...
+                                log.debug("Resource context exists. Returning the same.");
+                                return Mono.just(datasourceContextMap.get(datasourceContextIdentifier));
+                            }
+                        }
+
+                        // For adding rate limit on create connection, the assumption is that whenever
+                        // connection creation fails error is thrown by respective plugin, following plugins throw error
+                        // in case of invalid connection
+                        // Postgres, Oracle, Mssql and Redshift
+                        return createNewDatasourceContext(datasourceStorage, datasourceContextIdentifier)
```
sumitsum, line 297: QQ: what happens for MySQL when connection fails ?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
```diff
@@ -252,20 +266,67 @@ public boolean isValidDatasourceContextAvailable(
...
+                                .flatMap(datasourceContext -> {
```
sumitsum, line 298: Is this line required ? 
`.flatMap(datasourceContext -> Mono.just(datasourceContext)` ?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
```diff
@@ -252,20 +266,67 @@ public boolean isValidDatasourceContextAvailable(
...
+                                    return Mono.just(datasourceContext);
+                                })
+                                .onErrorResume(AppsmithPluginException.class, error -> {
+                                    return datasourceService
+                                            .consumeTokenIfAvailable(datasourceStorage)
+                                            .zipWith(rateLimitIdentifierMono)
+                                            .flatMap(tuple -> {
+                                                Boolean isSuccessful = tuple.getT1();
```
sumitsum, line 306: Can you please rename this to `wasTokenAvailable` ?

sumitsum (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
```diff
@@ -252,20 +266,67 @@ public boolean isValidDatasourceContextAvailable(
...
+                                                String rateLimitIdentifier = tuple.getT2();
+                                                if (!isSuccessful) {
```
sumitsum, line 308: The code block seems duplicate with the test datsource flow ? Can we refactor it into a common method for both ?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
```diff
@@ -252,20 +266,67 @@ public boolean isValidDatasourceContextAvailable(
...
-        } else {
-            if (isValidDatasourceContextAvailable(datasourceStorage, datasourceContextIdentifier)) {
-                log.debug("Resource context exists. Returning the same.");
-                return Mono.just(datasourceContextMap.get(datasourceContextIdentifier));
-            }
-        }
-        return createNewDatasourceContext(datasourceStorage, datasourceContextIdentifier);
+        Mono<String> rateLimitIdentifierMono = datasourceService.getRateLimitIdentifier(datasourceStorage);
+        return datasourceService
+                .isEndpointBlockedForConnectionRequest(datasourceStorage)
```
sumitsum, line 271: Can you please add a comment here explaining (1)  the concept of rate limit (2) what plugins it would apply to right now and why it wouldn't work with other plugins (i.e. the connection pool issue).

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
sneha122: In case of mysql, connection pool is created. We should handle that case separately as discussed

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
sneha122: I have tried to refactor common code as much as possible in the recent commit, please check and let me know if further refactoring is required or not.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
```diff
@@ -252,20 +266,67 @@ public boolean isValidDatasourceContextAvailable(
...
+                .flatMap(isBlocked -> {
+                    if (isBlocked) {
```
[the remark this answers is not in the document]
sneha122: I have tried to refactor common code as much as possible in the recent commit, please check and let me know if further refactoring is required or not.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
```diff
@@ -265,7 +271,39 @@ public Mono<DatasourceContext<?>> getDatasourceContext(DatasourceStorage datasou
...
             }
         }
-        return createNewDatasourceContext(datasourceStorage, datasourceContextIdentifier);
+
+        // As per https://github.com/appsmithorg/appsmith/issues/27745, Rate Limiting needs to applied
+        // on connection failure, otherwise brute force attacks may happen and our IP may get blocked
+        // For connection creation, the assumption is that whenever connection creation fails error
+        // is thrown by respective plugin, following plugins throw error: Postgres, Oracle, Mssql and Redshift
+        // For other plugins like MySQL, SMTP, Elastic Search, ArangoDB, Redis, In createNewDatasourceContext,
+        // Instead of connection, connection pool is created, connection is created when we query execution happens
```
sumitsum, line 280: Minor correction here - for the plugins that return error, they may also be using connection pool - the difference being that in their case the connection pool is trying to establish connection right away vs in other cases where the connection pool attempts to create a connection only when the first query is run.

sumitsum (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/datasources/base/DatasourceServiceCEImpl.java
```diff
@@ -536,39 +536,21 @@ protected Mono<DatasourceTestResult> verifyDatasourceAndTest(DatasourceStorage d
...
-                                                                    Duration.ofMinutes(BLOCK_TEST_API_DURATION),
-                                                                    exception)
-                                                            .flatMap(isAdded -> {
-                                                                return Mono.just(tooManyRequests);
-                                                            })
-                                                            .onErrorResume(error -> {
-                                                                return Mono.just(tooManyRequests);
-                                                            });
+                                                    return this.blockEndpointForConnectionRequest(datasourceStorage)
+                                                            .flatMap(isAdded -> Mono.just(tooManyRequests))
```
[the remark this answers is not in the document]
sneha122: So this function essentially adds a key in the redis with timeout and returns a Mono boolean, two things can happen here:
- Key is already present -> in this case we are throwing an error with too many requests from RedisUtils itself
- Key is absent -> we add the key and true is returned

I dont see a case where isAdded would be false, maybe in a case where the redis set operation itself fails due to connection or any other issue, in that case also we would need to return too many requests right or should we return the actual error to user (I feel actual error would be useless from user's perspective) and sending it as too many requests makes sense. Let me know what you think

app/server/appsmith-server/src/main/java/com/appsmith/server/datasources/base/DatasourceServiceCEImpl.java
```diff
@@ -536,39 +536,21 @@ protected Mono<DatasourceTestResult> verifyDatasourceAndTest(DatasourceStorage d
...
+                                                            .onErrorResume(error -> Mono.just(tooManyRequests));
```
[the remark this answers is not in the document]
sneha122: But again sending that error to user, will it make sense for user to see such error? Error occurred due to our flow of rate limiting blockage?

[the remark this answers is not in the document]
sumitsum: @sneha122 I think it would be good to do so, because then the user may report that issue to us or it may get logged via analytics - both of which may get lost if we mask the error.

app/server/appsmith-server/src/main/java/com/appsmith/server/datasources/base/DatasourceServiceCEImpl.java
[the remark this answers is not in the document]
sumitsum: Regardign `in that case also we would need to return too many requests right or should we return the actual error to user (I feel actual error would be useless from user's perspective) ` - I think it would be good to return the actual error, because then the user may report that issue to us or it may get logged via analytics - both of which may get lost if we mask the error.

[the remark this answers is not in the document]
sneha122: Makes sense, will make those changes, in case of false, no error would be thrown, so should we create a custom error for this? something like BLOCKING_FAILED?

app/server/appsmith-server/src/main/java/com/appsmith/server/datasources/base/DatasourceServiceCEImpl.java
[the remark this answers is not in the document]
sneha122: Makes sense, will add it

app/server/appsmith-server/src/main/java/com/appsmith/server/exceptions/AppsmithError.java
```diff
@@ -1019,6 +1019,14 @@ public enum AppsmithError {
             "Too many requests",
             ErrorType.INTERNAL_ERROR,
             null),
+    DATASOURCE_CONNECTION_RATE_LIMIT_BLOCKING_FAILED(
+            200,
```
sumitsum, line 1023: QQ: why is the error code set to 200 ? Doesn't 200 stand for all ok ?

sumitsum (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/exceptions/AppsmithError.java
sneha122: In case test and connection both, even though code is different the API says 200 and in the list of invalids, we get the error, so thought of keeping 200 itself

sumitsum: Hi Sneha, can you please change the code ? This could be confusing to understand in future.

sumitsum (dismissed):

sumitsum (approved):

chandannkumar:
Tested this PR and LGTM
- Tested for following Datasources 
Postgres
Oracle
Redshift
MsSQL
- Tested with Read/write and other connection types

sumitsum (approved):

PR #28290, opened 2023-10-23
feat: Support running as a non-root user

sharat87:
Running an Appsmith as a non-root user:

```sh
docker run --name appsmith --user 70:70
```

The `70:70` figures are the UID and GID respectively. It can mostly be any number, safe to user figures are 70 to 79, or anything above 200 and below 65000. The important bit, is that it shouldn't change on restart or manual updates etc.

No product functionality should be affected when running as a non-root user.


sharat87:
@pratapaprasanna, looks good. Since I'm the author of the PR, I can't mark it as approved. Please merge and take it forward.

PR #28330, opened 2023-10-24
chore: Refactored import and export svc methods for modularity

nidhi-nair:
Breaks down all the exportable and importable methods into structured calls to parallelize where possible, and enable overriding.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/onpageload/internal/PageLoadExecutablesUtilCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/onpageload/internal/PageLoadExecutablesUtilCEImpl.java
@@ -34,7 +34,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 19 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/onpageload/internal/PageLoadExecutablesUtilCEImpl.java
```diff
@@ -101,7 +102,7 @@ public Mono<List<Set<DslExecutableDTO>>> findAllOnLoadExecutables(
 
         Set<String> onPageLoadExecutableSetRef = new HashSet<>();
         Set<String> explicitUserSetOnLoadExecutablesRef = new HashSet<>();
-        Set<String> bindingsFromExecutablesRef = new HashSet<>();
+        Set<String> bindingsFromExecutablesRef = ConcurrentHashMap.newKeySet();
```
nayan-rafiq, line 105: I already raised a PR for this. Is it required to go with this PR?

app/server/appsmith-server/src/main/java/com/appsmith/server/onpageload/internal/PageLoadExecutablesUtilCEImpl.java
```diff
@@ -855,9 +856,8 @@ private Mono<Set<ExecutableDependencyEdge>> recursivelyAddExecutablesAndTheirDep
         }
 
         // All executables found from possibleExecutableNames set would add their dependencies in the following set for
-        // further
-        // walk to find more executables recursively.
-        Set<String> newBindings = new HashSet<>();
+        // further walk to find more executables recursively.
+        Set<String> newBindings = ConcurrentHashMap.newKeySet();
```
nayan-rafiq, line 860: Same here. It's ok if we make these changes with this PR.

nayan-rafiq (commented):
Looks good to me.

app/server/appsmith-server/src/main/java/com/appsmith/server/onpageload/internal/PageLoadExecutablesUtilCEImpl.java
nidhi-nair: Whoops! Sorry, I think this just remained from when we were debugging. Ill remove this and force merge later. Thanks for catching this :D

PR #28513, opened 2023-10-31
feat: Add IFrame widget for Custom EChart and support for custom functions in property pane

### LINKED ISSUE #27095: Allow users to use custom functions for chart widget in an iFrame.

rajatagrawal:
Fixes #27095


rajatagrawal:
@arunvjn Can you please help review changes in worker/Evaluation folder ?

arunvjn (dismissed):
Custom Validation changes looks good!

app/client/src/widgets/ChartWidget/component/CustomEChartIFrameComponent.tsx
```diff
@@ -0,0 +1,265 @@
...
+            <script src="/libraries/echarts.min.js"></script>
+            <script src="/libraries/echarts-gl.min.js"></script>
+            <script src="/libraries/lodash.min.js"></script>
+        </head>
+
+        <body>
+            <div id="chartdiv" style="position:absolute; height: 100%; width: 100%;"></div>
+    
+            <script>
+                ${stringifyFns([
```
sbalaji1192, line 193: let's not do it this way. It would be hard to manage these functions as the widget grows in size. Let's create a script and define these functions on that file. We should read the content of the file as a string and include in the srcDoc.

app/client/src/widgets/ChartWidget/component/CustomEChartIFrameComponent.tsx
```diff
@@ -0,0 +1,265 @@
...
+  useEffect(() => {
+    window.addEventListener("message", onMessage);
+    return () => {
+      window.removeEventListener("message", onMessage);
+    };
+  });
+
+  useEffect(() => {
+    let shouldUpdateOptions = true;
+    const propsEqual = equal(
```
sbalaji1192, line 221: do we need to check the deep equal here. can we just add props as deps of this useEffect? If you think we need to check deep equality, you can use `useDeepEffect` from app/client/src/utils/hooks/useDeepEffect.ts

app/client/src/widgets/ChartWidget/component/CustomEChartIFrameComponent.tsx
```diff
@@ -0,0 +1,265 @@
...
+      }
+    } else {
+      if (propsEqual) {
+        shouldUpdateOptions = false;
+      }
+    }
+
+    postMessageFn({
+      props: props,
+      shouldResize: true,
```
sbalaji1192, line 240: instead of having these flags, we should have message types. It would be more descriptive.

app/client/src/widgets/ChartWidget/component/CustomEChartIFrameComponent.tsx
```diff
@@ -0,0 +1,265 @@
...
+  });
+
+  return (
+    <>
+      {errorMsg && (
+        <ChartErrorComponent message={errorMsg} stack={errorStack} />
+      )}
+      {!errorMsg && (
+        <>
+          <IframeContainer
```
sbalaji1192, line 253: let's make sure we are not rerendering the chart every time something changes.

app/client/src/widgets/ChartWidget/component/helpers.ts
```diff
@@ -93,102 +91,52 @@ export const getTextWidth = (text: string, font: string) => {
...
-  isCustomChart = () => {
-    return !this.isBasicChart;
-  };
-
-  isCustom2DChart = () => {
-    return !this.isCustom3DChart;
-  };
-}
+export const getCustomEChartOptions = (props: ChartComponentProps) => {
+  return props.customEChartConfig;
```
sbalaji1192, line 95: nit: do we need this helper. seems simple

app/client/src/widgets/ChartWidget/component/helpers.ts
```diff
@@ -93,102 +91,52 @@ export const getTextWidth = (text: string, font: string) => {
...
-  } else {
-    return isSeriesConfig3D(seriesConfig);
-  }
-};
+export const dataClickCallbackHelper = (
+  params: echarts.ECElementEvent,
+  props: ChartComponentProps,
+  chartType: ChartType,
+) => {
+  const dataPointClickParams = parseOnDataPointClickParams(params, chartType);
```
sbalaji1192, line 137: seems like this logic needs to be wrapped inside the component itself. Ideally we should hide the iframe implementation from the rest of the platform by making the component a blackbox.

app/client/src/widgets/ChartWidget/component/helpers.ts
```diff
@@ -209,3 +157,7 @@ export const isCustomFusionChart = (type: ChartType) => {
 export const isCustomEChart = (type: ChartType) => {
   return type == "CUSTOM_ECHART";
 };
+
+export const numberInLocale = (label: number | string) => {
+  return label.toLocaleString();
```
sbalaji1192, line 162: nit: can be called directly instead of the caller statement. If a function is simple we should replace the calling statement with the body, it would improve readability, since the reader won't have to jump around much to understand the flow.

app/client/src/widgets/ChartWidget/component/index.tsx
```diff
@@ -437,4 +379,14 @@ class ChartComponent extends React.Component<
...
 }
 
-export default ChartComponent;
+export const mapStateToProps = (
+  state: AppState,
+  ownProps: ChartComponentProps,
+) => {
+  return {
+    needsOverlay:
+      ownProps.widgetId !== getWidgetPropsForPropertyPane(state)?.widgetId,
```
sbalaji1192, line 388: need some help here to understand

app/client/src/widgets/ChartWidget/component/CustomEChartIFrameComponent.tsx
```diff
@@ -0,0 +1,265 @@
...
+        "*",
+      );
+    }
+
+    echartsInstance.off("click");
+
+    echartsInstance.on("click", (event: echarts.ECElementEvent) => {
+      const newEvent = _.omit(_.cloneDeep(event), "event");
+      parent.postMessage(
+        JSON.stringify({ type: "click-event", event: newEvent }),
```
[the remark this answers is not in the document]
rajatagrawal: For custom chart click events, we don't know how to parse x and y values for the click event. Some custom charts may not even have x and y values. Hence it was decided that for custom echart events the complete event will be sent to the user.

app/client/src/widgets/ChartWidget/component/helpers.ts
rajatagrawal: I added a function so that I can unit test it. It is an important piece of line that I want to ensure if always present, so added [unit tests](https://github.com/appsmithorg/appsmith/blob/f02eb1a84718cdf982b751d6b53117aebfe90d1d/app/client/src/widgets/ChartWidget/component/helpers.test.ts) for it.

It was getting difficult to add assertions for this line if a unit test was being added for its caller function.

app/client/src/widgets/ChartWidget/component/index.tsx
rajatagrawal: This code adds an overlay to the iframe widget when it is not selected so that a user can select it using a hand pointer. The overlay is removed when the widget gets selected.

A similar pattern is used by the iframe widget, code snippet [here](https://github.com/appsmithorg/appsmith/blob/5e9f44c2f81bec396025f68eed0b605c11a86914/app/client/src/widgets/IframeWidget/component/index.tsx#L151)

rajatagrawal: If the overlay is not added, the user is unable to select the widget by clicking on its content.

app/client/src/widgets/ChartWidget/component/CustomEChartIFrameComponent.tsx
rajatagrawal: I tried to use this approach because it allows us to write the functions in typescript. Having typescript definition helps improve readability and debugging of this code.

If we move the function definitions to a separate file, it may not be possible to write them in typescript.

The end goal is to have compile time checks for the code that goes into the script just like other react code.

If there is a way to do it with a separate file, I can use that approach.

app/client/src/widgets/ChartWidget/component/helpers.ts
rajatagrawal: Added it here because it is a common business logic function that determines how to parse click event params for basic charts, fusion charts and custom echarts. The nested functions add a bit of logic and do a bit of parsing before actually passing the event data to click event callbacks.

So unit tests were required to test the logic and the parsing logic. Added a unit tests to test for each of the inputs possible.

Moved it to a helper class because having it inside the component would need to write a lot of unrequired boiler plate code and make the tests less readable.

app/client/src/widgets/ChartWidget/component/CustomEChartIFrameComponent.tsx
```diff
@@ -0,0 +1,265 @@
...
+  props: ChartComponentConnectedProps,
+) {
+  const [errorMsg, setErrorMsg] = useState<string | undefined>(undefined);
+  const [errorStack, setErrorStack] = useState<string | undefined>(undefined);
+  const prevProps = usePrevious(props);
+  const iframeID = props.widgetId + "-chartiframe";
+
+  const postMessageFn = (data: MessageProps) => {
+    const iFrameWindow = (
+      document.getElementById(iframeID) as HTMLIFrameElement
```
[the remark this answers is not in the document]
rajatagrawal: Done

app/client/src/widgets/ChartWidget/component/CustomEChartIFrameComponent.tsx
```diff
@@ -0,0 +1,265 @@
...
+        shouldDispose: data.shouldDispose,
+        width: data.props.dimensions.componentWidth,
+        height: data.props.dimensions.componentHeight,
+      },
+      "*",
+    );
+  };
+
+  const onMessage = (event: MessageEvent) => {
+    const iFrameWindow = (
```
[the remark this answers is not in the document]
rajatagrawal: Done

app/client/src/widgets/ChartWidget/component/CustomEChartIFrameComponent.tsx
```diff
@@ -0,0 +1,265 @@
...
+      document.getElementById(iframeID) as HTMLIFrameElement
+    )?.contentWindow;
+
+    if (event.source != iFrameWindow) {
+      return;
+    }
+
+    const data = JSON.parse(event.data);
+
+    if (data.type == "click-event") {
```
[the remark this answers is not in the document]
rajatagrawal: Done

app/client/src/widgets/ChartWidget/component/CustomEChartIFrameComponent.tsx
```diff
@@ -0,0 +1,265 @@
...
+      const newEvent = _.omit(_.cloneDeep(event), "event");
+      parent.postMessage(
+        JSON.stringify({ type: "click-event", event: newEvent }),
+        "*",
+      );
+    });
+  }
+
+  function parseConfigurationForCallbackFns(chartConfig: string, _: any) {
+    const config = JSON.parse(JSON.stringify(chartConfig));
```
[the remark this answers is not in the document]
rajatagrawal: Done

app/client/src/widgets/ChartWidget/component/helpers.ts
rajatagrawal: Done

app/client/src/widgets/ChartWidget/component/CustomEChartIFrameComponent.tsx
rajatagrawal: There is a nested logic that uses props equality to determine the logic of use effect. It may not be possible to achieve the desired logic with deps.

app/client/src/widgets/ChartWidget/component/CustomEChartIFrameComponent.tsx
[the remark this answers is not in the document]
sbalaji1192: Got it.

app/client/src/widgets/ChartWidget/component/helpers.ts
sbalaji1192: We don't need to unit test a native function.

app/client/src/widgets/ChartWidget/component/index.tsx
sbalaji1192: oh got it. We might need to lift this state access to a platform component. We should refrain from accessing states directly on the widgets.But for now lets add a small todo and tag me,

app/client/src/widgets/ChartWidget/component/CustomEChartIFrameComponent.tsx
sbalaji1192: Sorry, I don't understand, we're using the equal function here to check deep equality, `useDeepEffect` also using the same function.

app/client/src/widgets/ChartWidget/component/helpers.ts
sbalaji1192: if this is not related to iframe component then lets leave it here.

app/client/src/widgets/ChartWidget/component/helpers.ts
rajatagrawal: That's true that native function don't need to be tested.

The purpose of adding the test is to ensure that this native function gets called otherwise it will break the chart functionality.

rajatagrawal: Incase some refactors the code to remove .toLocaleString() and forgets to add it back, the unit test will be there to check that .toLocaleString gets added

app/client/src/widgets/ChartWidget/component/index.tsx
rajatagrawal: I tried that, but it became a big task. We might need to do it in withWidgetProps.

app/client/src/widgets/ChartWidget/component/CustomEChartIFrameComponent.tsx
rajatagrawal: I mean that we need to run the use effect when the props are equal too, not just when the props are not equal.

app/client/src/widgets/ChartWidget/component/helpers.ts
sbalaji1192: > Incase some refactors the code to remove .toLocaleString() and forgets to add it back, the unit test will be there to check that .toLocaleString gets added

We need to add a comment in that case. Unit test is not the right option.

app/client/src/widgets/ChartWidget/component/index.tsx
sbalaji1192: Yes. withWidgetProps is the right place. Lets add a todo

app/client/src/widgets/ChartWidget/component/helpers.ts
rajatagrawal: Got it.

rajatagrawal: Will make that change.

rajatagrawal: Done

rajatagrawal:
> @rajatagrawal Reviewed the widget side of the code and provided a few callouts and questions. Can you please look into it?
> 
> 1. Lets include the version details in the public/libraries we added in this PR. It will be helpful in future when we want to upgrade.
> 2. The title of the PR doesn't cover all the changes this PR has. Can we rethink?
> 3. Can you add details of the contract of the message we pass between the iframe and the parent in a txt file in this folder.?

1. Added the library versions.
2. Discussed offline. Since we have two PRs, this PR title is specific to only custom echart changes.
3. Add MessageContractDocumentation.md file that contains this documentation.

app/client/src/widgets/ChartWidget/component/CustomEChartIFrameComponent.tsx
rajatagrawal: Done

app/client/src/widgets/ChartWidget/component/CustomEChartIFrameComponent.tsx
rajatagrawal: I rechecked by putting a console log statement inside in the script tag in the iframe srcdoc and verified that the console log gets printed only once even while updating chart options.

app/client/src/widgets/ChartWidget/component/CustomEChartIFrameComponent.tsx
rajatagrawal: In the interest of other priorities would it be okay if I add this as a todo item for myself and pick up on my personal time ? I will have to read up on writing a webpack plugin.

sbalaji1192:
>Discussed offline. Since we have two PRs, this PR title is specific to only custom echart changes.
Pr also has the iframe code right? We should include that as well.

sbalaji1192 (approved):

PR #28633, opened 2023-11-06
chore: Refactored on page load logic to be generic wrt creator context

nidhi-nair:
Make sure on load calculations are valid in any context.


app/server/appsmith-server/src/main/java/com/appsmith/server/onload/internal/OnLoadExecutablesUtilCE.java
```diff
@@ -35,7 +38,10 @@ Mono<List<Set<DslExecutableDTO>>> findAllOnLoadExecutables(
      */
     Mono<Boolean> updateExecutablesExecuteOnLoad(
             List<Executable> onLoadExecutables,
-            String pageId,
+            String creatorId,
             List<LayoutExecutableUpdateDTO> executableUpdatesRef,
-            List<String> messagesRef);
+            List<String> messagesRef,
+            CreatorContextType creatorType);
```
subrata71, line 44: QQ: Is the new field `creatorType` going to create uncommitted changes for git-connected app?

nidhi-nair: Not yet, this PR is only introducing conditional logic on the basis of the field, but we aren't really storing it anywhere. That change would come in during instantiation related changes I would assume.

subrata71: So, it's going to have an impact on pages. Is that correct? If yes, can we please give heads up to git-pod about this?

PR #28671, opened 2023-11-06
chore: add configuration for large and small widgets

prsidhu:
## Description

Use case - Widgets need to be distinguished as large or small in Anvil.

1. Add property ``isLargeWidget`` to ``AnvilConfig`` interface.
2. Update AnvilConfig for all widget. Especially, WDS widgets.

#### Type of change

## Testing

- [x] Manual
- [ ] JUnit
- [ ] Jest
- [ ] Cypress

#### Dev activity

#### QA activity:


app/client/src/WidgetProvider/constants.ts
```diff
@@ -458,3 +460,7 @@ export interface WidgetConfigProps {
   widgetName: string;
   enhancements?: Record<string, unknown>; // TODO(abhinav): SPECIFY TYPES
 }
+
+export const largeWidgets: Record<WidgetType, boolean> = {
```
KelvinOm, line 464: I really don't like the fact that we have everything stored in one place. Here we have  a FixedLayout and an AutoLayout, and now Anvil is beginning to spread. Let's think about how to separate Anvil from all this. Abstractions will start leak very soon and it will be quite difficult to work with the code.

cc @riodeuno

app/client/src/widgets/AudioWidget/widget/index.tsx
```diff
@@ -78,6 +79,7 @@ class AudioWidget extends BaseWidget<AudioWidgetProps, WidgetState> {
 
   static getAnvilConfig(): AnvilConfig | null {
     return {
+      isLargeWidget: !!largeWidgets[this.type],
```
KelvinOm, line 82: Why is it so difficult? Why can't I just set `true` or `false` here? `isLargeWidget: false`?

KelvinOm (changes requested):

app/client/src/WidgetProvider/constants.ts
riodeuno: @KelvinOm I'm not sure if I understand your point correctly. Could you suggest an alternative or point to an issue so that we can discuss this accordingly?

Are you suggesting that we should have one map for all widget configs regardless of the layout system?

I do agree with the fact that the knowledge of the current layout system shouldn't be present anywhere outside of the layout system's data flow.

Also, with Fixed and Auto Layout slated to be deprecated sometime in the future, references layout systems should also be deprecated.

app/client/src/widgets/AudioWidget/widget/index.tsx
riodeuno: @prsidhu Could you elaborate on the need for using an external map to describe this property?

app/client/src/WidgetProvider/constants.ts
riodeuno: If you're referring to the fact that a map like `largeWidgets` exists in this file, instead of having the widget define it statically. Let's wait for @prsidhu's response on the rationale.

KelvinOm: @riodeuno I'm not only talking about `largeWidgets` map, but also about the fact that everything related to Anvil engine should be separated. This will simplify the code and help with the deprivation of code in the future.

app/client/src/widgets/AudioWidget/widget/index.tsx
prsidhu: If we need to access all large widgets, we won't have to keep iterating over all widget types in real time.

KelvinOm: @prsidhu We don't need that map right now, right? If we need it, we can easily collect this information and make a map. Until then, I'd prefer to keep the code as simple as possible.

KelvinOm (approved):

marks0351 (dismissed):

marks0351 (approved):

PR #28686, opened 2023-11-07
feat: Show generate page on Datasource Page

hetunandu:
Will show the Generate page on Datasource detail page which is similar to the Active DS list.
- It will show up on the same datasources which were supported in the Active DS list
- It will *not* show up for datasources that have schema view enabled
- It will *not* show up for users not with the permissions to create queries or pages

This change will be useful when we remove the Active DS list when introducing the App Sidebar 

Screenshot 2023-11-07 at 4 21 20 PM


app/client/src/pages/Editor/DataSourceEditor/DSFormHeader.tsx
```diff
@@ -142,6 +153,23 @@ export const DSFormHeader = (props: DSFormHeaderProps) => {
     ];
   };
 
+  const routeToGeneratePage = () => {
+    if (!showGenerateButton) {
+      // disable button when it doesn't support page generation
+      return;
+    }
+    AnalyticsUtil.logEvent("DATASOURCE_CARD_GEN_CRUD_PAGE_ACTION");
```
sneha122, line 161: @hetunandu Should we log the source as well in event? Like from where this generate CRUD had been called? I dont remember if this was there on active DS or not

app/client/src/pages/Editor/DataSourceEditor/hooks.ts
```diff
@@ -83,3 +103,64 @@ export const useDatasourceQuery = ({
...
+
+  const generateCRUDSupportedPlugin: GenerateCRUDEnabledPluginMap = useSelector(
+    getGenerateCRUDEnabledPluginMap,
+  );
+
+  const canCreatePages = getHasCreatePagePermission(
+    isGACEnabled,
+    userAppPermissions,
+  );
+  const canCreateDatasourceActions = hasCreateDSActionPermissionInApp(
```
sneha122, line 153: @hetunandu Did these permissions not exist when generate CRUD was on active DS list? Asking because we have created a whole new definition here for when and when not show generate CRUD

app/client/src/pages/Editor/DataSourceEditor/DSFormHeader.tsx
hetunandu: There wont be another option for Generate CRUD since we are removing the Active DS page. I see a max of 2 weeks where both would be present. In that case, I dont see it necessary to track this particular data point.

app/client/src/pages/Editor/DataSourceEditor/hooks.ts
hetunandu: It would need more than what we have on Active DS because when the schema view is active, a Generate Page is added which takes into account the current table being viewed in the schema. This current button takes the user to a new page where the table/collection is selected first and then a page is created. 

Permission to create a page and actions was present in the Active DS button as well.
https://github.com/appsmithorg/appsmith/blob/release/app/client/src/pages/Editor/IntegrationEditor/DatasourceCard.tsx#L202-L204

sneha122 (approved):

app/client/src/pages/Editor/DataSourceEditor/hooks.ts
```diff
@@ -83,3 +103,64 @@ export const useDatasourceQuery = ({
     failedFetchingPreviewData,
   };
 };
+
+export const useShowPageGenerationOnHeader = (
```
AmanAgarwal041, line 107: Should we use, this function check wherever we are adding a generate page button in future ?

AmanAgarwal041 (approved):

app/client/src/pages/Editor/DataSourceEditor/hooks.ts
hetunandu: Yup. It will would need to be updated if the schema preview plugins are updated, but it should work for any new datasource plugin that can support schema

PR #28829, opened 2023-11-13
feat: wds modal improvements

### LINKED ISSUE #28814: Use any trigger for modal
- Add the ability to use any trigger for a modal
- Add the ability to add a wrapper for modal content

KelvinOm:
## Description
- Add the ability to use a dialog without a trigger component.
- Add the ability to pass a triggerRef to the component that helps to set the correct and aria attributes to trigger.

Fixes #28814 

#### Type of change

- [x] Manual
- [ ] JUnit
- [ ] Jest
- [ ] Cypress

#### Dev activity


app/client/packages/design-system/headless/src/components/Popover/src/PopoverContent.tsx
```diff
@@ -16,65 +15,56 @@ const _PopoverContent = (props: PopoverContentProps, ref: Ref<HTMLElement>) => {
...
-    );
-  } else {
-    return (
-      <FloatingPortal root={root}>
-        <FloatingFocusManager
-          closeOnFocusOut={closeOnFocusOut}
-          context={context}
-          modal={false}
+  return (
+    <FloatingPortal root={root}>
```
KelvinOm, line 45: Separated code for select/menu and model implementation

app/client/packages/design-system/widgets/src/components/Modal/stories/CustomModal.tsx
```diff
@@ -0,0 +1,26 @@
...
+export const CustomModal = () => {
+  const [isOpen, setIsOpen] = useState(false);
+  const ref = useRef(null);
+
+  return (
+    <>
+      <Button onPress={() => setIsOpen(!isOpen)} ref={ref}>
+        Modal trigger
+      </Button>
+      <Modal isOpen={isOpen} setOpen={setIsOpen} triggerRef={ref}>
```
KelvinOm, line 13: @marks0351 Pay attention that it is necessary to pass `triggerRef` that will add needed aria attributes and render Modal at the closest theme provider.  Without `triggerRef` modal will also work, but we will lose a11y.

app/client/packages/design-system/widgets/src/components/Modal/stories/ModalExamples.tsx
```diff
@@ -0,0 +1,133 @@
...
+      </Modal>
+
+      <ButtonGroupItem
+        onPress={() => setMediumOpen(!isMediumOpen)}
+        ref={mediumRef}
+      >
+        Medium
+      </ButtonGroupItem>
+      <Modal
+        initialFocus={2}
```
KelvinOm, line 61: @marks0351 Also note that it is better to use `initialFocus` instead of autofocus(how it was before) for better accessibility.

app/client/packages/design-system/widgets/src/components/Modal/stories/CustomModal.tsx
```diff
@@ -0,0 +1,26 @@
...
+        <div style={{ border: "4px solid rgb(255, 155, 78)" }}>
```
KelvinOm, line 14: @marks0351 This is an example of using wrapper fo modal component. Does it make sense us something like this in widget?

app/client/packages/design-system/widgets/src/components/Modal/stories/ModalExamples.tsx
marks0351: can you let me know a quick use case to differentiate using initial and auto focus?

KelvinOm: In general, we shouldn't use `autofocus` at all. Later we will add the appropriate [eslint rule](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-autofocus.md) for this. 

Both `initialFocus` and `autofocus` are used to set the focus for a better UX. So if we set the focus on the submit button, we will immediately be able to confirm the actions by pressing the space and enter without doing additional effort.

marks0351 (approved):

PR #28904, opened 2023-11-16
fix: Adding multiple certs in ca-certs folder

sharat87:
Fix issue with alias names clashing in `keytool -import` command, when there's more than one cert file in the `ca-certs` folder.

The fix is to explicitly set the alias for each `keytool -import` run, to the file itself, so clashes don't happen.

### DIFF
--- a/deploy/docker/fs/opt/appsmith/entrypoint.sh
+++ b/deploy/docker/fs/opt/appsmith/entrypoint.sh
@@ -318,7 +318,8 @@ setup-custom-ca-certificates() (
 
   # Add the custom CA certificates to the store.
   find "$stacks_ca_certs_path" -maxdepth 1 -type f -name '*.crt' \
-    -exec keytool -import -noprompt -keystore "$store" -file '{}' -storepass changeit ';'
+    -print \
+    -exec keytool -import -alias '{}' -noprompt -keystore "$store" -file '{}' -storepass changeit ';'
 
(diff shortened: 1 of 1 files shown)


deploy/docker/fs/opt/appsmith/entrypoint.sh
```diff
@@ -318,7 +318,8 @@ setup-custom-ca-certificates() (
 
   # Add the custom CA certificates to the store.
   find "$stacks_ca_certs_path" -maxdepth 1 -type f -name '*.crt' \
-    -exec keytool -import -noprompt -keystore "$store" -file '{}' -storepass changeit ';'
+    -print \
```
mohanarpit, line 321: Do we need to print this out?

sharat87: Don't need to, but it'll print out which CA cert files have been successfully added, and which failed. Without this `-print, if there's three files in `ca-certs`, we see output like this:

```
Certificate was added to keystore
Certificate was added to keystore
Certificate was added to keystore
```

But with `-print`, we see this:

```
one.crt
Certificate was added to keystore
two.crt
Certificate was added to keystore
three.crt
Certificate was added to keystore
```

Not a strong opinion, but I think this will help troubleshoot.

mohanarpit (approved):

PR #28907, opened 2023-11-16
chore: Moving references around for EE consumption

nidhi-nair:
Make a CE version for refactor entity dto. 
Allow for unhandled entity types during refactor.


app/server/appsmith-server/src/main/java/com/appsmith/server/refactors/applications/RefactoringSolutionCEImpl.java
```diff
@@ -201,6 +201,7 @@ protected EntityRefactoringService<?> getEntityRefactoringService(RefactorEntity
             case JS_ACTION -> jsActionEntityRefactoringService;
             case ACTION -> newActionEntityRefactoringService;
             case JS_OBJECT -> actionCollectionEntityRefactoringService;
+            default -> null;
```
subrata71, line 204: Does `null` fit best here? Or should we just through an error here?

nidhi-nair: Since the type can not have a value that is not recognized, I'm throwing null here.

subrata71: In any scenario can the execution reach the default case?

subrata71 (approved):

PR #28911, opened 2023-11-16
feat: added multipart app/json support

### LINKED ISSUE #28419: [Feature]: Support application/json as content type for multipart/form-data type request

### Summary

Multipart/form-data type request only support 3 type (Text, File and Array). I would like to have application/json as an content type option when building multipart form data request.

image

Backend service framework often have request handler support to transform request body to an Object and Json is always preferred data type in these cases.

For our case, if we set Text (text/plain) as content type, backend service treat it as string and won't transform it to object.

sondermanish:
### PR fixes following issue(s)
Fixes #28419 

#### Type of change
- Feature Improvement

- [x] Manual
- [x] JUnit


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/restApiUtils/helpers/DataUtils.java
```diff
@@ -194,6 +196,39 @@ public String parseFormData(List<Property> bodyFormData, Boolean encodeParamsTog
...
                         }
                         break;
+                    case JSON:
+                        if (property.getValue() == null) {
+                            // Should we throw out an error over here?, We don't expect this to be the case
+                            break;
+                        }
+
+                        if (!(property.getValue() instanceof String jsonString)) {
+                            bodyBuilder.part(key, property.getValue());
```
sondermanish, line 206: Not sure whether to add the application/json media-type over here.

sondermanish: same question with empty strings? do we stop execution or should we do a soft pass

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/restApiUtils/helpers/DataUtils.java
```diff
@@ -194,6 +196,39 @@ public String parseFormData(List<Property> bodyFormData, Boolean encodeParamsTog
                             bodyBuilder.part(key, property.getValue());
                         }
                         break;
+                    case JSON:
+                        if (property.getValue() == null) {
+                            // Should we throw out an error over here?, We don't expect this to be the case
+                            break;
+                        }
+
+                        if (!(property.getValue() instanceof String jsonString)) {
```
sumitsum, line 205: Do we know what other type can the value be apart from a String ? If no, could we instead do this - convert the value to string first i.e. `property.getValue().toString()` - and then try to read this value back as JSON - would allow us to remove this `if` block ?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/restApiUtils/helpers/DataUtils.java
```diff
@@ -194,6 +196,39 @@ public String parseFormData(List<Property> bodyFormData, Boolean encodeParamsTog
...
+                            objectFromJson = objectFromJson(jsonString);
+                        } catch (JsonSyntaxException | ParseException e) {
+                            throw new AppsmithPluginException(
+                                    AppsmithPluginError.PLUGIN_JSON_PARSE_ERROR,
+                                    jsonString,
+                                    "Malformed JSON: " + e.getMessage());
+                        }
+
+                        if (objectFromJson == null) {
+                            bodyBuilder.part(key, "");
```
sumitsum, line 227: Is it possible for parsed JSON object to be null for a non-empty string ? If not, then should we remove this if condition because `null` and empty string case has been handled before this ?

app/server/appsmith-plugins/restApiPlugin/src/test/java/com/external/plugins/RestApiPluginTest.java
```diff
@@ -1019,6 +1019,97 @@ public void testMultipartFormData() {
...
+        actionConfig.setHttpMethod(HttpMethod.POST);
+        final Property key1 = new Property("key1", "onlyValue");
+        final Property key2 = new Property(
+                "key2", "{\"name\":\"fileName\", \"type\":\"application/json\", \"data\":{\"key\":\"value\"}}");
+        final Property key3 = new Property(
+                "key3", "[{\"name\":\"fileName2\", \"type\":\"application/json\", \"data\":{\"key2\":\"value2\"}}]");
+        final Property key4 = new Property(null, "irrelevantValue");
+        key2.setType("FILE");
+        key3.setType("FILE");
+        String jsonString = " {\"id\": abcd,\"email\": \"abcd@xyz.mn\",\"userName\": \"Michael Law\"} ";
```
sumitsum, line 1042: Can you please add an array of JSON as a another use case to check if that works ?

app/server/appsmith-plugins/restApiPlugin/src/test/java/com/external/plugins/RestApiPluginTest.java
```diff
@@ -1019,6 +1019,97 @@ public void testMultipartFormData() {
                 .verifyComplete();
     }
 
+    @Test
+    public void testMultipartFormDataWithJSON() {
```
sumitsum, line 1023: Can you please add a test case with mustache binding in place of JSON value ?

sumitsum (changes requested):

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/restApiUtils/helpers/DataUtils.java
```diff
@@ -194,6 +196,39 @@ public String parseFormData(List<Property> bodyFormData, Boolean encodeParamsTog
                             bodyBuilder.part(key, property.getValue());
                         }
                         break;
+                    case JSON:
+                        if (property.getValue() == null) {
+                            // Should we throw out an error over here?, We don't expect this to be the case
```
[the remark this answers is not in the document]
sondermanish: null is not a valid value in Multipart, I've added a check to skip the null values

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/restApiUtils/helpers/DataUtils.java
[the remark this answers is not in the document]
sondermanish: I did not quite get the question, I've added it as a declaration for further use of the variable. Do you wan't to get rid of the declaration?

app/server/appsmith-plugins/restApiPlugin/src/test/java/com/external/plugins/RestApiPluginTest.java
sondermanish: added as suggested

app/server/appsmith-plugins/restApiPlugin/src/test/java/com/external/plugins/RestApiPluginTest.java
sondermanish: added as suggested

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/restApiUtils/helpers/DataUtils.java
sondermanish: I'm not sure if this case will ever arise, however the multipart can't intake null, should we error out over here, or just skip it?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/restApiUtils/helpers/DataUtils.java
sondermanish: Although anything coming from client side should be effectively a string,  If we save the action config via postman, we might have some values which are not string.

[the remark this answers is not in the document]
sumitsum: Hi @sondermanish yes, I was wonder if we can remove it because it is not used at the moment.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/restApiUtils/helpers/DataUtils.java
sumitsum: @sondermanish let us skip this change for now, because we don't know why this situation should arise. In case such a situation does come up, I think it is better to error out and find a proper fix for it then.

app/server/appsmith-plugins/restApiPlugin/src/test/java/com/external/plugins/RestApiPluginTest.java
```diff
@@ -1019,6 +1019,248 @@ public void testMultipartFormData() {
...
+
+        ActionConfiguration actionConfig = new ActionConfiguration();
+        actionConfig.setAutoGeneratedHeaders(List.of(new Property("content-type", "multipart/form-data")));
+        actionConfig.setHttpMethod(HttpMethod.POST);
+
+        String jsonString = " {\"id\": abcd,\"email\": \"abcd@xyz.mn\",\"userName\": \"Michael Law\"} ";
+        final Property jsonKey = new Property("dictionary", jsonString);
+        jsonKey.setType("JSON");
+
+        String listString = "[\"a\",\"b\",\"c\"]";
```
sumitsum, line 1128: Hi @sondermanish can you please add an array of JSON as value - e.g. [{...}, {...}] ?

sumitsum (changes requested):

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/restApiUtils/helpers/DataUtils.java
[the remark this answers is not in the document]
sumitsum: Sorry, looks like it is being used. My mistake :)

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/restApiUtils/helpers/DataUtils.java
```diff
@@ -194,6 +203,36 @@ public String parseFormData(List<Property> bodyFormData, Boolean encodeParamsTog
                             bodyBuilder.part(key, property.getValue());
                         }
                         break;
+                    case JSON:
+                        if (!(property.getValue() instanceof String jsonString)) {
```
sumitsum, line 207: Hi Manish, can you please add a comment here explaining why we need this check ?

app/server/appsmith-plugins/restApiPlugin/src/test/java/com/external/plugins/RestApiPluginTest.java
sondermanish: Added as suggested

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/restApiUtils/helpers/DataUtils.java
sondermanish: added as suggested

sumitsum (approved):

PR #28954, opened 2023-11-20
chore: Refactored is name allowed

nidhi-nair:
Refactored `isNameAllowed` to allow extending the number of checks that need to be applied for figuring out invalid names that already exist in context.

`isNameAllowed` should be a refactoringService method, moved it to be there. `updateLayout` now belongs to a service of its own. All other changes to accommodate dependencies among beans.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/actioncollections/refactors/ActionCollectionRefactoringServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/actioncollections/refactors/ActionCollectionRefactoringServiceCEImpl.java
@@ -2,8 +2,10 @@ package com.appsmith.server.actioncollections.refactors;
 
 import com.appsmith.external.constants.AnalyticsEvents;
 import com.appsmith.external.models.ActionDTO;
+import com.appsmith.external.models.CreatorContextType;
 import com.appsmith.external.models.MustacheBindingToken;
 import com.appsmith.server.actioncollections.base.ActionCollectionService;
+import com.appsmith.server.constants.FieldName;
 import com.appsmith.server.domains.ActionCollection;
 import com.appsmith.server.dtos.ActionCollectionDTO;
 import com.appsmith.server.dtos.EntityType;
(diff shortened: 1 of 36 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/actioncollections/refactors/ActionCollectionRefactoringServiceCEImpl.java
```diff
@@ -138,4 +142,15 @@ public Mono<Void> updateRefactoredEntity(RefactorEntityNameDTO refactorEntityNam
                 })
                 .then();
     }
+
+    @Override
+    public Flux<String> getExistingEntityNames(String contextId, CreatorContextType contextType, String layoutId) {
+        MultiValueMap<String, String> params = new LinkedMultiValueMap<>();
+        if (contextId != null) {
```
sondermanish, line 149: let's use `String.hasText()` over here?

app/server/appsmith-server/src/main/java/com/appsmith/server/refactors/applications/RefactoringSolutionCEImpl.java
```diff
@@ -226,4 +229,46 @@ private Mono<Void> sendRefactorAnalytics(
...
+        boolean isFQN = newName.contains(".");
+
+        Iterable<Flux<String>> existingEntityNamesFlux = getExistingEntityNamesFlux(pageId, layoutId, isFQN);
+
+        return Flux.merge(existingEntityNamesFlux)
+                .collect(Collectors.toSet())
+                .map(existingNames -> !existingNames.contains(newName));
+    }
+
+    protected Iterable<Flux<String>> getExistingEntityNamesFlux(String pageId, String layoutId, boolean isFQN) {
```
sondermanish, line 245: Do we see any use case where we would need these values separately? If not then merging these inside the method itself could be done as well. Although this would just be a matter of preference

nidhi-nair: No, but its been written this way so that the overriding implementation can add to the list and parallelize all calls in one go. Does that make sense?

sondermanish: This would be parallel because we are asking for the list of publisher containers later in logic execution steps?

nidhi-nair: That's correct.

sondermanish (approved):
refactors only! LGTM

Issue #28924, opened 2023-11-17
Templates forking experience in self hosted instances


rahulbarwal:
When I fork a template on a self hosted instance. All the queries of that template are broken but when I do the same on cloud, it seems to work. Why is this happening?

[Slack Message](https://theappsmith.slack.com/archives/C02RCR2PURZ/p1699380804525999?thread_ts=1699380804.525999&cid=C02RCR2PURZ)

PR #28998, opened 2023-11-21
chore: remove the redirection for mockdb to localhost

AnaghHegde:
## Description
This PR fixes the experience of Templates forking in self hosted instances. And also for to Set up a process to keep the embedded DB up to date with template db schemas.
We have removed the redirection of mockdb end point used in templates App when forked in self hosted instance from localhost/internal postgres db. 
This also has a migration which is to make sure that none of existing apps using the internal postgres does not break due to the removal of redirection. The migration will make sure that existing self hosted instances using the posgress db and has a datasource with mockdb end point will be replaces with localhost. 

Fixes https://github.com/appsmithorg/appsmith/issues/28924

#### Type of change

- [ ] Manual

#### Dev activity

#### QA activity:


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration035RemoveMockDbEndPointInDatasourceInSelfHostedInstance.java
```diff
@@ -0,0 +1,79 @@
...
+    public void removeMockDbEndpointInDatasource() {
+        // This migration is applicable for self-hosted instances only.
+        // The end point used in templates is redirected to internal postgres db
+        // We are now switching to RDS and hence want to remove this endpoint from the datasource and replace it with
+        // localhost, so that the existing queries do not break. And the data stored in the db by users is not lost
+
+        // APPSMITH_ENABLE_EMBEDDED_DB && isCloudHosting
+        Map<String, String> envVariable = getEnvVariable();
+        if (Boolean.FALSE.equals(commonConfig.isCloudHosting())
+                && envVariable.containsKey("APPSMITH_ENABLE_EMBEDDED_DB")) {
```
sharat87, line 54: Please use `System.getenv("APPSMITH_ENABLE_EMBEDDED_DB")` to get this instead of parsing the `docker.env` directly. Environment variables can be set form many places other than `docker.env`.

deploy/docker/fs/opt/appsmith/entrypoint.sh
```diff
@@ -355,7 +355,6 @@ configure_supervisord() {
     if [[ $runEmbeddedPostgres -eq 1 ]]; then
       cp "$supervisord_conf_source/postgres.conf" "$SUPERVISORD_CONF_TARGET"
       # Update hosts lookup to resolve to embedded postgres
```
sharat87, line 357: Should we remove this comment as well?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration035RemoveMockDbEndPointInDatasourceInSelfHostedInstance.java
```diff
@@ -0,0 +1,79 @@
...
+
+            Query query = query(where("structure").exists(true));
+
+            Update update = new Update().unset("structure");
```
sharat87, line 58: Looks like we aren't using these. Can we remove them?

sharat87 (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration035RemoveMockDbEndPointInDatasourceInSelfHostedInstance.java
```diff
@@ -0,0 +1,54 @@
...
+    @Execution
+    public void removeMockDbEndpointInDatasource() {
+        // This migration is applicable for self-hosted instances only.
+        // The end point used in templates is redirected to internal postgres db
+        // We are now switching to RDS and hence want to remove this endpoint from the datasource and replace it with
+        // localhost, so that the existing queries do not break. And the data stored in the db by users is not lost
+
+        // APPSMITH_ENABLE_EMBEDDED_DB && isCloudHosting
+        String envValue = System.getenv("APPSMITH_ENABLE_EMBEDDED_DB");
+        if (Boolean.FALSE.equals(commonConfig.isCloudHosting()) && StringUtils.isNotEmpty(envValue)) {
```
sharat87, line 46: Only the value `0` will disable the embedded DB. Any other non empty string will enable it, which is also the default.

```suggestion
        boolean isEmbeddedDbEnabled = !"0".equals(System.getenv("APPSMITH_ENABLE_EMBEDDED_DB"));
        if (isEmbeddedDbEnabled && Boolean.FALSE.equals(commonConfig.isCloudHosting())) {
```

AnaghHegde: If the env itself is not set then wouldn't that be empty?

sharat87: Sorry, I'm talking about when it's not empty. The current logic treats the value `0` and `1` the same. We're only checking for a non-empty value. That's not the case in the `entrypoint.sh` script. There we check if this value is exactly `0`, and only then the embedded DB is disabled.

sharat87 (changes requested):

sharat87 (approved):

PR #29007, opened 2023-11-21
chore: Added context type handling for isNameAllowed

nidhi-nair:
Added `contextType` as a param for `isNameAllowed` method so that the method can be reused for other contexts as well.


sondermanish (approved):
Minor change to further generalise. LGTM

PR #29012, opened 2023-11-21
refactor: Cypress SelectEntityByName

hetunandu:
Refactors SelectEntityByName to allow all types of selections but makes defining the Entity Type compulsory


app/client/cypress/e2e/Regression/Apps/CommunityIssues_Spec.ts
```diff
@@ -11,6 +11,9 @@ import {
   assertHelper,
   draggableWidgets,
 } from "../../../support/Objects/ObjectsCore";
+import EditorNavigation, {
+  EntityType,
+} from "../../../support/Pages/EditorNavigation";
```
Aishwarya-U-R, line 16: Can this please be added as const object in Objects core & imported from that - similar to other page objects we have today?

hetunandu: I do not think that is the right way to do it. Why is that a pattern today?

Aishwarya-U-R: Oh, why is it not the right way please?

hetunandu: The Object Registry is used to ensure a singleton is passed
```
static get CommunityTemplates(): CommunityTemplates {
    if (ObjectsRegistry.communityTemplates__ === undefined) {
      ObjectsRegistry.communityTemplates__ = new CommunityTemplates();
    }
    return ObjectsRegistry.communityTemplates__;
  }
```

That can be easily done in the export statement itself:
```
class EditorNavigation {
    ...
}

export default new EditorNavigation();
```


And the Object Core file is just to collect these exports to kind of improve import but that causes irregular patterns like
`_.entityExplorer.SelectEntityByName` or `entityExplorer.SelectEntityByName` or `this.entityExplorer.SelectEntityByName`


They cause unnecessary confusion on new users writing tests. It is better to always call the function via the class which will be same no matter where you import it.

```
import EditorNavigation from "...."

...



EditorNavigation.SelectEntityByName(...)

````

Aishwarya-U-R: Ok, got it.

Aishwarya-U-R (approved):

PR #29035, opened 2023-11-22
test: Cypress - added script to validate tags added in cypress test cases

sharanya-appsmith:


KelvinOm (approved):
LGTM! But I think it's worth asking someone else to watch yml and cypress, ones who specialises in this more.

PR #29108, opened 2023-11-26
chore: Use version from info.json directly

sharat87:
This will bring us one step closer to the `info.json` being the one source of truth for the version information. We shouldn't need the step to overwrite the contents of the `version.js` file during build in CI.


sharat87:
Tested working on DP.

mohanarpit (approved):

KelvinOm (approved):

PR #29175, opened 2023-11-28
add auto heal to cron

pratapaprasanna:
## Summary by CodeRabbit

- **New Features**
  - Introduced auto-healing functionality to automatically restart unresponsive backend services.
  - Added SSL configuration support for custom domains.

- **Chores**
  - Implemented periodic backend service status checks.
  - Enhanced startup scripts to support new auto-healing feature based on environment configuration.

### DIFF
--- a/deploy/docker/fs/opt/appsmith/auto_heal.sh
+++ b/deploy/docker/fs/opt/appsmith/auto_heal.sh
@@ -16,6 +16,7 @@ do
         then
                 echo "TIMESTAMP=`date "+%Y-%m-%d %H:%M:%S"` backend is unresponsive";
                 supervisorctl restart backend || echo true
+                sleep 60
         else
                 echo "TIMESTAMP=`date "+%Y-%m-%d %H:%M:%S"` backend is responsive";
         fi
--- a/deploy/docker/fs/opt/appsmith/auto_heal.sh
+++ b/deploy/docker/fs/opt/appsmith/auto_heal.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+
+set -e
+
+set -o xtrace
+
+sleep 120
+
+while true;
+do
+        ### Get backend state
+        echo "auto heal check"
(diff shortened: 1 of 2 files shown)


deploy/docker/fs/opt/appsmith/auto_heal.sh
```diff
@@ -0,0 +1,23 @@
...
+
+set -o xtrace
+
+sleep 120
+
+while true;
+do
+        ### Get backend state
+        echo "auto heal check"
+        appsmith_status="$(curl --write-out %{http_code} --silent --output /dev/null localhost:80/api/v1/tenants/current)"
```
rajatagrawal, line 13: Does this check for the case when auto heal is already restarting the backend server and it doesn't trigger another restart ?

pratapaprasanna: Yes. It will wait until the restart is complete

rajatagrawal (approved):

deploy/docker/fs/opt/appsmith/auto_heal.sh
```diff
@@ -0,0 +1,23 @@
...
+        echo "TIMESTAMP=`date "+%Y-%m-%d %H:%M:%S"` backend response "$appsmith_status
+        if [[ "$appsmith_status" -ne 200 ]];
+        then
+                echo "TIMESTAMP=`date "+%Y-%m-%d %H:%M:%S"` backend is unresponsive";
+                supervisorctl restart backend || echo true
```
sharat87, line 18: Shouldn't we add a `sleep 120` after restart as well?

sharat87 (approved):

deploy/docker/fs/opt/appsmith/auto_heal.sh
pratapaprasanna: I dont think it is required because `supervisorctl restart backend `will be on hold until backend reaches running state.
But, I think it is good to keep sleep 60 just so that we can be sure about it

PR #29264, opened 2023-12-01
fix: added changes for mustache binding

sondermanish:


nidhi-nair (approved):

nidhi-nair:
LGTM, there is usage on EE that needs to be handled as well

PR #29281, opened 2023-12-03
ci: Updating actions/checkout to v4 and defaulting to fetch-depth 1 instead of 0

mohanarpit:
## Summary by CodeRabbit

- **Chores**
  - Updated GitHub Actions workflows to use `actions/checkout@v4` for improved performance and reliability.
  - Removed `fetch-depth` parameter to simplify checkout steps across various workflows.
  - Standardized quote usage for consistency in workflow files.

- **Documentation**
  - Adjusted formatting and descriptions in workflow files for better clarity and readability.

- **Refactor**
  - Aligned multiple workflow files to follow a consistent structure and naming convention.

fetch-depth 0 causes the Github workflow to checkout the entire Git history. This is not required. We only need to check out the head of the commit. By default, actions/checkout has fetch-depth=1, hence removing it from the workflow completely for simplicity.

### DIFF
--- a/.github/workflows/perf-test-v2.yml
+++ b/.github/workflows/perf-test-v2.yml
@@ -25,10 +25,6 @@ jobs:
       github.event_name == 'push' ||
       github.event_name == 'workflow_dispatch' ||
       github.event_name == 'repository_dispatch'
-    defaults:
-      run:
-        working-directory: app/client
-        shell: bash
 
     strategy:
       fail-fast: false
--- a/.github/workflows/perf-test-v2.yml
+++ b/.github/workflows/perf-test-v2.yml
@@ -29,7 +29,7 @@ jobs:
       run:
         working-directory: app/client
         shell: bash
-    
+
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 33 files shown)


.github/workflows/perf-test-v2.yml
```diff
@@ -25,11 +25,7 @@ jobs:
       github.event_name == 'push' ||
       github.event_name == 'workflow_dispatch' ||
       github.event_name == 'repository_dispatch'
-    defaults:
```
sarojsarab, line 28: Is there any reason why we are removing this step? I am asking this because if we are removing it, then we should update the path in the below commands in the `Setting up the performance tests` step and other steps that have similar things as well
```
          chmod a+x ./perf/setup-perf-test.sh
          ./perf/setup-perf-test.sh
```

sarojsarab (changes requested):

.github/workflows/perf-test-v2.yml
mohanarpit: I removed it because these lines are a duplicate of [this section.](https://github.com/appsmithorg/appsmith/pull/29281/files#diff-4a059fc58d52b3ad8be3bc87721b75f9166f5f593b409415bb6208be85258b5fR14) This was redundant.

sarojsarab (approved):
LGTM

PR #29305, opened 2023-12-05
chore: Split custom js libs acc to context

nidhi-nair:
Contains structural changes to classes and split required for reusing custom js libs logic in other contexts.


nidhi-nair:
Merging since EE equivalent has been approved.

PR #29529, opened 2023-12-12
ci: Fix for the junit run-result cache saving issue

sarojsarab:
## Description
Currently, when we run the server build without tests, we still try to check the cache for run-result, but as the cache is not present during the first run, where we just build the server, and on the success of the workflow, actions/cache create the cache with the specified key,

Which is creating an issue while we try to save the actual run-result cache when we run the tests in server-unit-tests.

Solution : 
- Added a condition to check the cache present only while running tests

#### Type of change
- Workflow changes (server-build.yml)
## Testing
- Workflow run

## Summary by CodeRabbit

- **Refactor**
  - Updated the server build process to conditionally skip certain steps based on user input.

- **Chores**
  - Improved automation in the build workflow to enhance efficiency.


sarojsarab:
[Workflow run](https://github.com/appsmithorg/appsmith/actions/runs/7177575204)

Aishwarya-U-R (approved):

Issue #29624, opened 2023-12-14
[Task]: Introduce theme settings in Application object for theme v2


AnaghHegde:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SubTasks

- Add option to store the theme settings in ApplicationDetail for theme v2

PR #29625, opened 2023-12-14
feat: Add option to store the theme settings for application

AnaghHegde:
## Description
Currently, a theme is a separate entity that is tied to an application. With the new WDS and the new theming engine, we don't need the theme to be a separate entity. It can be just a few additional properties of an application ( just like how navigation settings work today.
More can be found here - https://www.notion.so/appsmith/New-Theme-Settings-c28bde5baae6424c8742384565d37089

Fixes https://github.com/appsmithorg/appsmith/issues/29624

#### Type of change

- [ ] Manual
- [ ] JUnit

#### Dev activity

#### QA activity:

## Summary by CodeRabbit

- **New Features**
  - Introduced theme customization settings, allowing users to personalize the appearance of their applications with options for accent color, border radius, sizing, and density.

- **Enhancements**
  - Updated application update process to include theme settings as part of the configuration.

- **Tests**
  - Added tests to ensure theme settings are correctly applied and persisted across application versions.


app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Application.java
```diff
@@ -413,4 +423,22 @@ public enum Type {
...
+    public static class ThemeSettings {
+
+        @JsonView(Views.Public.class)
+        private String accentColor;
+
+        @JsonView(Views.Public.class)
+        private String borderRadius;
+
+        @JsonView(Views.Public.class)
+        private int sizing;
```
nayan-rafiq, line 439: If we use primitive `int` type, it'll have default value set to zero. Is that ok with the implementation?

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Application.java
```diff
@@ -413,4 +423,22 @@ public enum Type {
...
+
+        @JsonView(Views.Public.class)
+        private int density;
```
nayan-rafiq, line 442: Same like the above. Just need to make sure we're handling zero as the default value.

nayan-rafiq (commented):
I suggest to add tests for import-export service as well. So, for example, if an application has no theme settings, then the ApplicationJson should have null set to that property and when there is a theme set, the application json should have the property set. Same for import services - if an import is done with a theme in application json, imported application should have these properties set.

NilanshBansal (changes requested):
@AnaghHegde you need to set the `themeSettings` in `setPropertiesToApplicationDetail()`  in `ImportExportUtils.java`.
Ref: https://github.com/appsmithorg/appsmith/pull/28307/files

NilanshBansal:
Also, please add tests for the following:
1. Git service. 
Ref: https://github.com/appsmithorg/appsmith/blob/8da859184e1ca1a66110b6abc4a7aedf29450109/app/server/appsmith-server/src/test/java/com/appsmith/server/services/GitServiceTest.java#L2230

2. Import Export.
Ref: https://github.com/appsmithorg/appsmith/pull/28307/files

app/server/appsmith-server/src/test/java/com/appsmith/server/solutions/ImportApplicationServiceTests.java
```diff
@@ -2404,9 +2404,9 @@ public void discardChange_removeNewActionCollection_removedActionCollectionResto
      */
     @Test
     @WithUserDetails(value = "api_user")
-    public void discardChange_addNavigationSettingAfterImport_addedNavigationSettingRemoved() {
-        Mono<ApplicationJson> applicationJsonMono =
-                createAppJson("test_assets/ImportExportServiceTest/valid-application-without-navigation-setting.json");
+    public void discardChange_addNavigationAndThemeSettingAfterImport_addedNavigationSettingRemoved() {
```
NilanshBansal, line 2407: @AnaghHegde can you write a separate test case for the theme setting with a separate json import. 
The scenario here is: 
If the theme settings are applied to an application initially and then git discard flow is triggered, the added theme settings should be removed.

NilanshBansal (changes requested):

NilanshBansal:
Rest LGTM!

app/server/appsmith-server/src/test/java/com/appsmith/server/solutions/ImportApplicationServiceTests.java
AnaghHegde: @NilanshBansal Why do you think this should be a separate test case? If you look at the assert statements, i have added the respective checks that looks at the `themeSettings`. And since these are part of the `ApplicationDetail` i prefer having these together. I don't see why this has to be a separate test case.

NilanshBansal: @AnaghHegde makes sense, in that case can we please rename this test case from `discardChange_addNavigationAndThemeSettingAfterImport_addedNavigationSettingRemoved()` to `discardChange_addNavigationAndThemeSettingAfterImport_addedNavigationAndThemeSettingRemoved()`.

AnaghHegde: Good catch, let me update the name

NilanshBansal (dismissed):

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Application.java
```diff
@@ -413,4 +423,37 @@ public enum Type {
             ANVIL
         }
     }
+
+    @Data
+    @AllArgsConstructor
```
nayan-rafiq, line 428: I don't see any use of AllArgsConstructor. If the constructor is not used, we can remove the annotation as well. The @Data annotation should be enough in this case.

app/server/appsmith-server/src/test/java/com/appsmith/server/solutions/ImportApplicationServiceTests.java
```diff
@@ -2419,6 +2463,16 @@ public void discardChange_addNavigationSettingAfterImport_addedNavigationSetting
                     Application.NavigationSetting navigationSetting = new Application.NavigationSetting();
                     navigationSetting.setOrientation("top");
                     applicationDetail.setNavigationSetting(navigationSetting);
+
+                    Application.ThemeSetting themeSettings = new Application.ThemeSetting();
```
nayan-rafiq, line 2467: Can we use a private method for the code in line 2467 to 2473? It's used in two places so it'll be helpful to use a private method here.

nayan-rafiq (changes requested):
LGTM. Minor change request.

nayan-rafiq (approved):

PR #29726, opened 2023-12-19
chore: Delete migrations older than `v1.9.2`, the last checkpoint version

sharat87:
1. We delete migrations that are ineffective on a fresh installation. For example, migrations that modify existing workspaces or applications etc.
2. We add a check to avoid reusing the same migration ids in the future.
3. Deleted all unused methods in the changelog classes, that were once used in the migrations that now no longer exist.

## Summary by CodeRabbit

- **Refactor**
	- Enhanced the database configuration for better migration management.
- **Chores**
	- Implemented checks to prevent the use of forbidden migration IDs.


sharat87:
That test failure doesn't make sense. Looks like it should've passed.
shot-2023-12-21-03-32-02

nidhi-nair (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/MongoConfig.java
```diff
@@ -56,6 +65,101 @@
         repositoryBaseClass = BaseRepositoryImpl.class)
 public class MongoConfig {
 
+    private static final Set<String> FORBIDDEN_IDS = Set.of(
+            // List generated during PR developmeng with the following command:
+            // git diff release HEAD | awk -F\" '/^-[[:space:]]+@ChangeSet/ {print "\"" $4 "\","}'
+            // Any deleted migration in the future, should go into this list.
```
AnaghHegde, line 71: QQ - Why are we maintaining this list? Also how do we ensure the order of deleted migration is correct?

AnaghHegde (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/MongoConfig.java
sharat87: Hey, thanks for the review. I've noted this in the PR description:

> We add a check to avoid reusing the same migration ids in the future.

To elaborate, today, if there's two migrations with the same `id` string, then Mongock throws an error on startup, saying there's an `id` reused. This is because Mongock remembers which migration has already been run and which needs to be run, by storing the `id` string in the `mongockChangeLog` collection.

Now since these migrations are deleted, it's possible that we may accidentally create a new migration in the future, with one of these ids. And Mongock won't complain then. But when run on a user's _old_ instance, it's possible that that user's `mongockChangeLog` already has an entry for the deleted migration. Seeing that, Mongock will assume this new migration was already run, and will silently say `PASSED OVER` for this new migration.

This list is to avoid exactly that. I'm hoping we add the ids of migrations that we delete in the future as well.

> how do we ensure the order of deleted migration is correct?

I'm not sure I understand. Deleted migrations don't need an order no? 🤔

sharat87: I'm going to merge this in since you've already approved. But do let me know if there's more questions on Slack. If we need more changes to make, I'll push them as a separate PR. 

Thanks!

PR #29795, opened 2023-12-21
chore: Split changes for on page load, refactor, publish, etc

nidhi-nair:
Split changes only

## Summary by CodeRabbit

- **New Features**
  - Enhanced pattern matching to include special characters in certain identifiers.
  - Expanded functionality to allow multiple executable names and user-friendly naming for configurations.
  - Introduced checks for permissible messages during the initial load of executables.

- **Enhancements**
  - Improved entity name retrieval logic to accommodate different view modes.
  - Refined object equality determination with additional entity reference type checks.

- **Bug Fixes**
  - Fixed value replacement in template keys to prevent potential issues with escaped characters.

- **Refactor**
  - Removed default actions to streamline action collection logic.
  - Adjusted method signatures to include view mode parameters, enhancing control flow based on user interface state.

- **Tests**
  - Updated test cases to align with the new method signatures and logic changes.

- **Documentation**
  - No visible changes to end-user documentation in this release.

- **Chores**
  - Updated package declarations to reflect the new structure.

- **Style**
  - No visible style changes to the end-user in this release.

- **Revert**
  - No reverts in this release.


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/MustacheHelper.java
```diff
@@ -41,7 +41,7 @@ public class MustacheHelper {
      * {{JSON.stringify(fetchUsers)}}
      * This pattern should return ["JSON.stringify", "fetchUsers"]
      */
-    private static final Pattern pattern = Pattern.compile("[a-zA-Z_][a-zA-Z0-9._]*");
+    private static final Pattern pattern = Pattern.compile("[a-zA-Z_][a-zA-Z0-9._$]*");
```
subrata71, line 44: Should we override this pattern in EE?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ce/ActionCE_DTO.java
```diff
@@ -281,6 +291,11 @@ public LayoutExecutableUpdateDTO createLayoutExecutableUpdateDTO() {
         return layoutExecutableUpdateDTO;
     }
 
+    @Override
+    public Boolean isOnLoadMessageAllowed() {
```
subrata71, line 295: Nit: Since this method returns a primitive type, can we change it to `boolean` instead?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/MustacheHelper.java
nidhi-nair: It's a static as it is. And I'd rather keep the pattern the same so that our logic in CE works on the same kind of nomenclature as EE. Upgrades will not be a problem in that way.

subrata71 (approved):

PR #29848, opened 2023-12-26
chore: New Theme Settings Integration

### LINKED ISSUE #29262: New theme settings Integration
After the backend team sets up the API for new theme setting, we will need to integrate it with new theme setting panel.

jsartisan:
Fixes #29262 

## Summary by CodeRabbit

- **Refactor**
	- Renamed theme configuration property for consistency and clarity within the app's theme settings feature.


app/client/src/ce/selectors/applicationSelectors.tsx
```diff
@@ -333,8 +333,9 @@ export const getCurrentPluginIdForCreateNewApp = (state: AppState) => {
 };
 
 export const getAppThemeSettings = (state: AppState) => {
-  return (
-    state.ui.applications.currentApplication?.applicationDetail?.themeSetting ||
-    defaultThemeSetting
-  );
+  return {
```
KelvinOm, line 336: Why do we need these changes?

KelvinOm (approved):

app/client/src/ce/selectors/applicationSelectors.tsx
jsartisan: Just making sure we spread out the theme setting that we get from the backend over the default settings. Previously it was like "If the backend has themeSettings, use that, otherwise use default Settings"

PR #29903, opened 2023-12-27
chore: Update the build process to refer the Appsmith version from the info.json in all build artifacts

### LINKED ISSUE #29623: [Task]: Update the build process to refer the Appsmith version from the info.json file in all build artifacts.

### SubTasks

- [client-build](https://github.com/appsmithorg/appsmith/blob/release/.github/workflows/client-build.yml)
- [server-build.yml](https://github.com/appsmithorg/appsmith/blob/release/.github/workflows/server-build.yml)

abhvsn:
Fixes #29623 

TBP run: https://github.com/appsmithorg/appsmith/actions/runs/7347634939

The build-time version has been incorrect for quite some time now. This PR will fix that as well.

shot-2024-01-12-12-26-05

## Summary by CodeRabbit

- **Refactor**
  - Streamlined build and versioning process for both client and server.
  - Enhanced logging of build version and commit details in the server application.

- **New Features**
  - Improved version tracking by having the server read version information from a generated JSON file.

- **Bug Fixes**
  - Resolved issue with incorrect version details being tagged to Docker images.

- **Documentation**
  - Updated build and deployment instructions to reflect the new workflow changes.


.github/workflows/client-build.yml
```diff
@@ -221,13 +221,19 @@ jobs:
...
+        run: |
+          if [[ -f ../../deploy/docker/fs/opt/appsmith/info.json ]]; then
+            echo "info.json already exists. Skipping generation."
+          else
+            echo "info.json does not exist. Generating..."
+            if [[ -f ../../scripts/generate_info_json.sh ]]; then
+              ../../scripts/generate_info_json.sh
+            fi
+          fi
+          echo version=$(jq -r .version ../../deploy/docker/fs/opt/appsmith/info.json) >> $GITHUB_OUTPUT
```
abhvsn, line 236: @sharat87 is there any better way to handle this relative path?

sharat87: This is actually not needed anymore. The `REACT_APP_VERSION_ID` and `REACT_APP_VERSION_RELEASE_DATE` should also be removed. The client gets the version at runtime today. Check out this code:

https://github.com/appsmithorg/appsmith/blob/7c6434b7789900be6054e8f92a608fd656a58a3c/deploy/docker/fs/opt/appsmith/run-caddy.sh#L18-L20

I've had this task of removing this from this workflow for quite some time now. I've removed it from rts build.

.github/workflows/server-build.yml
```diff
@@ -145,22 +145,19 @@ jobs:
...
-          echo tag=$(echo ${GITHUB_REF:11}) >> $GITHUB_OUTPUT
+          if [[ -f ../../deploy/docker/fs/opt/appsmith/info.json ]]; then
+            echo "info.json already exists. Skipping generation."
+          else
+            echo "info.json does not exist. Generating..."
+            if [[ -f ../../scripts/generate_info_json.sh ]]; then
+              ../../scripts/generate_info_json.sh
+            fi
+          fi
+          echo version=$(jq -r .version ../../deploy/docker/fs/opt/appsmith/info.json) >> $GITHUB_OUTPUT
```
sharat87, line 160: Even here, I'd prefer just not setting the version at build time at all, and removing these steps. Backend can read the `/opt/appsmith/info.json` at runtime and get the version from there, instead of reading it from the manifest file.

_That's_ what will make `info.json` a single source of truth for version information at runtime.

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/ProjectProperties.java
```diff
@@ -1,23 +1,43 @@
...
+
 @Configuration
-@PropertySource(value = "classpath:META-INF/maven/com.appsmith/server/pom.properties", ignoreResourceNotFound = true)
 @Getter
+@Slf4j
 public class ProjectProperties {
 
     public static final String EDITION = "CE";
+    private String version = "UNKNOWN";
+    private static final String INFO_JSON_PATH = "/opt/appsmith/info.json";
```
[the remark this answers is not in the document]
abhvsn: We are trying to avoid the extra configs hence have not added this to env.

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/ProjectProperties.java
```diff
@@ -1,23 +1,43 @@
...
+                    // Parse JSON content using the AppsmithInfo class
+                    ObjectMapper objectMapper = new ObjectMapper();
+                    BuildInfo appsmithInfo = objectMapper.readValue(jsonContent, BuildInfo.class);
+                    version = appsmithInfo.getVersion();
+                    return appsmithInfo.getVersion();
+                }
+            } catch (IOException e) {
+                // Ignore the exception and return "UNKNOWN" as the version
+                log.debug("Error reading version from info.json {}", e.getMessage());
+            }
```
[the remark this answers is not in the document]
abhvsn: We expect the file to be unavailable for local testing etc hence not logging the complete stacktrace.

sharat87 (approved):
Looks good. Can we also remove the `REACT_APP_VERSION_*` env variables and the `versions:set` command from the `github-release.yml` workflow as well? Or we can do that as a separate PR after we get confidence from this. Perfectly fine.

abhvsn:
> Looks good. Can we also remove the `REACT_APP_VERSION_*` env variables and the `versions:set` command from the `github-release.yml` workflow as well? Or we can do that as a separate PR after we get confidence from this. Perfectly fine.

Oh man! We have this version extraction spread across lot of files, thanks Shri for taking the initiative and making sure to refer to single source going forward. 

Let me take the changes related to `github-release.yml` in separate PR.

sharat87:
@abhvsn, sorry just realized, one other thing. Today, on server startup, we get a log line like this:

```
[2023-12-28 14:49:33,355]  - Starting ServerApplication vv1.9.57 using Java 17.0.9 with PID 1518 (/opt/appsmith/backend/server.jar started by root in /opt/appsmith/backend)
```

When users share logs with us, we use this line to figure out what version a log file belongs to. But I'm guessing Spring gets this version from the `pom.properties` file. Can we check if it's only printing `v1.0-SNAPSHOT` with the changes in this PR?

If yes, can we print the `version` and `commitSha` values from `info.json` on a line in the backend logs as well please?

I'm trying to see if we can set the version for Spring at runtime too, perhaps via a command line argument.

sharat87:
Doesn't look like. Spring is getting the version from the package's `MANIFEST.MF` file, and there doesn't seem to be a way to override that.

shot-2023-12-29-01-49-43

Let's log the version and commitSha in a line in the `ServerApplication` class ourselves instead.

abhvsn:
Screenshot 2023-12-29 at 1 56 24 PM

sharat87 (approved):

PR #29924, opened 2023-12-28
chore: Don't query on `deleted` field directly

sharat87:
Use the `notDeleted` query builder function instead.

## Summary by CodeRabbit

- **Refactor**
  - Improved the logic for querying applications by introducing more efficient criteria handling.
  - Enhanced the filtering process to more accurately exclude deleted records from aggregation results.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomApplicationRepositoryCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomApplicationRepositoryCEImpl.java
@@ -281,11 +281,6 @@ public class CustomApplicationRepositoryCEImpl extends BaseAppsmithRepositoryImp
                         + fieldName(QApplication.application.gitApplicationMetadata.defaultApplicationId))
                 .is(defaultApplicationId));
         query.addCriteria(notDeleted());
-        query.equals(where("this." + gitApplicationMetadata + "."
-                        + fieldName(QApplication.application.gitApplicationMetadata.branchName))
-                .equals("this." + gitApplicationMetadata + "."
-                        + fieldName(QApplication.application.gitApplicationMetadata.defaultBranchName)));
-
         return mongoOperations.findOne(query, Application.class);
     }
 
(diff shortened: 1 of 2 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomApplicationRepositoryCEImpl.java
```diff
@@ -282,12 +280,7 @@ public Mono<Application> getApplicationByDefaultApplicationIdAndDefaultBranch(St
         query.addCriteria(where(gitApplicationMetadata + "."
                         + fieldName(QApplication.application.gitApplicationMetadata.defaultApplicationId))
                 .is(defaultApplicationId));
-        query.addCriteria(where(fieldName(QApplication.application.deleted)).ne(true));
-        query.equals(where("this." + gitApplicationMetadata + "."
```
mohanarpit, line 286: Are we sure we want to remove this condition as well?

sharat87: This isn't _adding_ any condition here. It's checking `.equals` between a query object and a string, which is always false, and then not using that `false` result. This statement is effectively not doing anything.

mohanarpit: Oh ... I see it now. Interestingly, this function isn't really getting used anywhere as well. For now, we can delete the `query.equals()` clause, but we can also consider deleting the entire function.

mohanarpit (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomApplicationRepositoryCEImpl.java
sharat87: Yes, I noticed that as well. But I'm yet to check EE repo as well or there'll be build failures there. I intend to clean it up after I get rid of the boolean `deleted` field.

Thanks.

mohanarpit: I had checked the EE repo only. But I agree that we should do the cleanup in a separate PR.

PR #29943, opened 2023-12-29
chore(deps): Update commons-compress

sharat87:
Solves for these CVEs:

1. CVE-2021-36090
2. CVE-2021-35517
3. CVE-2021-35516
4. CVE-2021-35515


nidhi-nair (approved):
Approving on the basis of correctness. Please do include the change set on these dependencies in case you need a review on area of impact.

PR #23979, opened 2023-06-02
perf: Optimising DSL fetch api

NilanshBansal:
#### PR fixes following issue(s)
Fixes #23637 

- Chore (housekeeping or task changes that don't impact user perception)

#### How Has This Been Tested?

- [x] Manual
- [x] Jest
- [ ] Cypress

#### Dev activity
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] PR is being merged under a feature flag

#### QA activity:
- [ ] [Speedbreak features](https://github.com/appsmithorg/TestSmith/wiki/Test-plan-implementation#speedbreaker-features-to-consider-for-every-change) have been covered
- [ ] Test plan covers all impacted features and [areas of interest](https://github.com/appsmithorg/TestSmith/wiki/Guidelines-for-test-plans/_edit#areas-of-interest)
- [ ] Test plan has been peer reviewed by project stakeholders and other QA members
- [ ] Manually tested functionality on DP
- [ ] We had an implementation alignment call with stakeholders post QA Round 2
- [ ] Cypress test cases have been added and approved by SDET/manual QA
- [ ] Added `Test Plan Approved` label after Cypress tests were reviewed
- [ ] Added `Test Plan Approved` label after JUnit tests were reviewed

PR #29985, opened 2024-01-03
fix: fix error response returned from un-authorized page access

### LINKED ISSUE #29984: [Bug]: [Perf] Consolidated API : fix signin redirection on anonymous user access

### Description

- When using consolidated API if a non public page is accessed by anonymous user the redirection to sign-in page does not happen properly. 

### Steps To Reproduce

- Use consolidated API 
- Try to access a non-public deployed app using Chrome's incognito tab

### Public Sample App

_No response_

### Environment

Production

### Issue video log

_No response_

### Version

release

sumitsum:
## Description
- Currently, when using consolidated API, an anonymous user is not re-directed to sign-in page properly when accessing non-public application page. 
  - This happens because of a difference in the error code that the API server returns. The error code happened to be different because the API server used a different method of accessing the page than v1/pages API (although both seem to valid ways of accessing the page). 
  - This PR updates the consoldated API service to use the expected API to fetch page info. 

Fixes #29984

#### Type of change

## Testing
- [x] Manual

#### Dev activity

#### QA activity:

## Summary by CodeRabbit

- **Refactor**
  - Modified the method for retrieving the application ID during page load to enhance performance and reliability.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ConsolidatedAPIServiceImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ConsolidatedAPIServiceImpl.java
@@ -208,9 +208,8 @@ public class ConsolidatedAPIServiceImpl implements ConsolidatedAPIService {
         /* Fetch application id if not provided */
         Mono<String> applicationIdMonoCache;
         if (isBlank(applicationId)) {
-            applicationIdMonoCache = applicationPageService
-                    .getPage(defaultPageId, isViewMode)
-                    .map(PageDTO::getApplicationId)
+            applicationIdMonoCache = newPageService
+                    .findRootApplicationIdFromNewPage(branchName, defaultPageId)
                     .cache();
         } else {
             applicationIdMonoCache = Mono.just(applicationId).cache();


rajatagrawal (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ConsolidatedAPIServiceImpl.java
```diff
@@ -208,9 +208,8 @@ public Mono<ConsolidatedAPIResponseDTO> getConsolidatedInfoForPageLoad(
         /* Fetch application id if not provided */
         Mono<String> applicationIdMonoCache;
         if (isBlank(applicationId)) {
-            applicationIdMonoCache = applicationPageService
```
mohanarpit, line 211: @sumitsum If this is not the recommended way of fetching the page, should we replace this function across the board in the code base and remove it from the `applicationPageService` interface itself?

sumitsum: Hi @mohanarpit sorry, I am not aware if this API is not a recommended way. From the history info on GitHub it seems this API was introduced around 3 years ago and has also been updated recently to [improve the page fetch speed](https://github.com/appsmithorg/appsmith/pull/23979). For now my understanding is that over the past 3 years multiple methods have been introduced unknowingly to fetch page from the DB - each one being a valid way to fetch the page. However, I can see only 2 usage of this method across the CE code base and hence one could assume that this method is rarely used and we may want to deprecate it so that there is a fixed interface to access page. Having said that, in case we want to deprecate it we would also have to port the performance [fix](https://github.com/appsmithorg/appsmith/pull/23979) recently added to this method to other alternate method(s).

PR #30044, opened 2024-01-04
chore: consolidated API: update variable names, fix error propagation, handle defaults in case error

sumitsum:
## Description
- Update variable names to make them more understandable
- Fix an error handling clause wherein the error was not caught at the right place. 
- Replace with defaults:
  - when workspace Id is not found then don't supply workspace Id to downstream sub-APIs (effectively passing null value which is handled via validation check in the sub-APIs).
  - when feature flag API errors out, then assign false value to migrate DSL flag
- This PR is targeted to be merged in development branch `oneApiChanges` instead of `release`  

#### PR fixes following issue(s)
Fixes #29994 #29995

#### Type of change
- Bug fix (non-breaking change which fixes an issue)
- Chore (housekeeping or task changes that don't impact user perception)

- [x] Manual

>
>
#### Dev activity
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] PR is being merged under a feature flag

#### QA activity:
- [ ] [Speedbreak features](https://github.com/appsmithorg/TestSmith/wiki/Guidelines-for-test-plans#speedbreakers-) have been covered
- [ ] Test plan covers all impacted features and [areas of interest](https://github.com/appsmithorg/TestSmith/wiki/Guidelines-for-test-plans#areas-of-interest-)
- [ ] Test plan has been peer reviewed by project stakeholders and other QA members
- [ ] Manually tested functionality on DP
- [ ] We had an implementation alignment call with stakeholders post QA Round 2
- [ ] Cypress test cases have been added and approved by SDET/manual QA
- [ ] Added `Test Plan Approved` label after Cypress tests were reviewed
- [ ] Added `Test Plan Approved` label after JUnit tests were reviewed

Issue #29988, opened 2024-01-03
[Task]: fix JUnit TC for consolidated APIs


sumitsum:
- JUnit TCs for consolidated API throw mockito related error on EE but work fine on CE. Due to this the tests were disabled. Need to fix this issue and re-enable the tests.
- Also, add JUnit test cases for sub api authorization test using anonymous_user.
- [ref](https://github.com/appsmithorg/appsmith/pull/29972)

PR #30094, opened 2024-01-08
chore: consolidatedApi: split endpoint to view and edit mode

### LINKED ISSUE #30093: [Task]: consolidated API: split endpoint into view and edit mode
### SubTasks

- split consolidated API endpoint into view and edit mode.

sumitsum:
## Description
- Split consolidated API endpoint into two endpoints: view (`/view`) and edit mode (`/edit`)
- Carry over changes that were merged directly to Vamsi's dev branch: 
  -  #29985 
  - #30044

Fixes #30093 

#### Type of change

- [x] Manual
- Junit TCs are blocked due to TBP failure on EE. They will be picked up next via another PR. Letting this PR go in would allow Vamsi to test the FE changes while I work on the test cases. Tracking it [here](https://github.com/appsmithorg/appsmith/issues/29988). 

#### Dev activity

#### QA activity:

## Summary by CodeRabbit

- **New Features**
  - Introduced separate data loading methods for edit and view modes to enhance user experience.

- **Enhancements**
  - Improved clarity in the consolidated API response structure for better understanding of the data presented.

- **Security Updates**
  - Adjusted security settings to allow access to a new view-only API endpoint.

- **Bug Fixes**
  - Implemented error handling improvements in data migration processes.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
@@ -213,7 +213,7 @@ public class SecurityConfig {
                         ServerWebExchangeMatchers.pathMatchers(
                                 HttpMethod.POST, USER_URL + "/verifyEmailVerificationToken"),
                         ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, PRODUCT_ALERT + "/alert"),
-                        ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, CONSOLIDATED_API_URL))
+                        ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, CONSOLIDATED_API_URL + "/view"))
                 .permitAll()
                 .pathMatchers("/public/**", "/oauth2/**")
                 .permitAll()
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/ConsolidatedAPIResponseDTO.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/ConsolidatedAPIResponseDTO.java
@@ -21,59 +21,56 @@ import java.util.Map;
 @Setter
 public class ConsolidatedAPIResponseDTO {
     /* v1/users/me */
-    ResponseDTO<UserProfileDTO> v1UsersMeResp;
+    ResponseDTO<UserProfileDTO> userProfile;
 
     /* v1/users/features */
-    ResponseDTO<Map<String, Boolean>> v1UsersFeaturesResp;
+    ResponseDTO<Map<String, Boolean>> featureFlags;
 
     /* v1/tenants/current */
-    ResponseDTO<Tenant> v1TenantsCurrentResp;
+    ResponseDTO<Tenant> tenantConfig;
 
     /* v1/product-alert/alert */
-    ResponseDTO<ProductAlertResponseDTO> v1ProductAlertResp;
+    ResponseDTO<ProductAlertResponseDTO> productAlert;
 
     /* v1/pages */
-    ResponseDTO<ApplicationPagesDTO> v1PagesResp;
+    ResponseDTO<ApplicationPagesDTO> pages;
 
     /* v1/actions/view */
-    ResponseDTO<List<ActionViewDTO>> v1ActionsViewResp;
+    ResponseDTO<List<ActionViewDTO>> publishedActions;
 
     /* v1/actions */
-    ResponseDTO<List<ActionDTO>> v1ActionsResp;
+    ResponseDTO<List<ActionDTO>> unpublishedActions;
 
     /* v1/collections/actions/view */
-    ResponseDTO<List<ActionCollectionViewDTO>> v1CollectionsActionsViewResp;
+    ResponseDTO<List<ActionCollectionViewDTO>> publishedActionCollections;
 
     /* v1/collections/actions */
-    ResponseDTO<List<ActionCollectionDTO>> v1CollectionsActionsResp;
+    ResponseDTO<List<ActionCollectionDTO>> unpublishedActionCollections;
 
     /* v1/themes/applications/{applicationId}/current */
-    ResponseDTO<Theme> v1ThemesApplicationCurrentModeResp;
+    ResponseDTO<Theme> currentTheme;
 
     /* v1/themes/applications/{applicationId} */
-    ResponseDTO<List<Theme>> v1ThemesResp;
+    ResponseDTO<List<Theme>> themes;
 
     /* v1/pages/{pageId}/view */
-    ResponseDTO<PageDTO> v1PublishedPageResp;
-
-    /* v1/pages/{pageId} */
-    ResponseDTO<PageDTO> v1PageResp;
+    ResponseDTO<PageDTO> pageWithMigratedDsl;
 
     /* v1/pages/{pageId} - for all pages */
-    ResponseDTO<List<PageDTO>> v1PageDSLs;
+    ResponseDTO<List<PageDTO>> pagesWithMigratedDsl;
 
     /* v1/libraries/{applicationId}/view */
-    ResponseDTO<List<CustomJSLib>> v1LibrariesApplicationResp;
+    ResponseDTO<List<CustomJSLib>> customJSLibraries;
 
     /* v1/plugins */
-    ResponseDTO<List<Plugin>> v1PluginsResp;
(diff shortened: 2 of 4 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
```diff
@@ -213,7 +213,7 @@ public SecurityWebFilterChain securityWebFilterChain(ServerHttpSecurity http) {
                         ServerWebExchangeMatchers.pathMatchers(
                                 HttpMethod.POST, USER_URL + "/verifyEmailVerificationToken"),
                         ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, PRODUCT_ALERT + "/alert"),
-                        ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, CONSOLIDATED_API_URL))
+                        ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, CONSOLIDATED_API_URL + "/view"))
```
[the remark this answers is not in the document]
sumitsum: Hi @rajatagrawal , this is the config where APIs that are accessible to anonymous users is whitlelist. Because we don't want to whitelist the edit endpoint for anonymous users, it is not added here.

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/ConsolidatedAPIResponseDTO.java
```diff
@@ -21,59 +21,56 @@
 @Setter
 public class ConsolidatedAPIResponseDTO {
     /* v1/users/me */
-    ResponseDTO<UserProfileDTO> v1UsersMeResp;
+    ResponseDTO<UserProfileDTO> userProfile;
```
rajatagrawal, line 24: Have the endpoints that don't need to be part of the view mode removed from this API ?

rajatagrawal, line 24: Have the endpoints that don't need to be part of the view mode removed from this API ?

rajatagrawal, line 24: Have the endpoints that don't need to be part of the view mode removed from this API ?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ConsolidatedAPIServiceImpl.java
```diff
@@ -246,7 +246,14 @@ public Mono<ConsolidatedAPIResponseDTO> getConsolidatedInfoForPageLoad(
...
         Mono<Boolean> migrateDslMonoCache = featureFlagsForCurrentUserResponseDTOMonoCache
-                .map(ResponseDTO::getData)
+                .map(responseDTO -> {
+                    if (INTERNAL_SERVER_ERROR_STATUS
+                            == responseDTO.getResponseMeta().getStatus()) {
+                        return Map.of();
+                    }
+
+                    return responseDTO.getData();
+                })
```
mohanarpit, line 256: @sumitsum Any reason we are handling the internal server error specially here? Is this similar to the handling in our current scenario as well?

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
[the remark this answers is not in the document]
rajatagrawal: Thanks!

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/ConsolidatedAPIResponseDTO.java
mohanarpit: @rajatagrawal As part of the DTO, I'd recommend keeping it the exact same. This way, the client will receive the same DTO object back regardless of whether it's view mode or edit mode. It's just that some keys will be null/missing in either of them. This is the current behaviour as well.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ConsolidatedAPIServiceImpl.java
sumitsum: Hi @mohanarpit this is as per the expectation from the client. We had discussed this in the team and decided that in case the feature flag API fails for any reason we want to continue with the execution assuming that the `migrateDsl` flag value is false. cc: @vsvamsi1

mohanarpit: @sumitsum Is this what happens today as well?

sumitsum: Hi @mohanarpit I need to confirm this with @vsvamsi1 once and revert back : my understanding is that the client makes a separate API call to feature flags (in normal flow) and if that call were to fail the client would rely on a default value for the feature flag (which is currently set to false) for the `migrateDsl` i.e `release_server_dsl_migrations_enabled` feature flag in `FeatureFlag.ts` file.

vsvamsi1: Yes @sumitsum @mohanarpit, feature flag call is independent to the initialisation flow, if the call fails the initialisation will continue and it should be fine since it will switch to the default behaviour.

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/ConsolidatedAPIResponseDTO.java
rajatagrawal: got it.

Sorry I meant it to be a generic comment for the view mode api. Does the code need to exclude enpoint responses that are not needed in the view mode ? I am assuming all the end point responses are being returned as the edit mode in the view mode as well.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ConsolidatedAPIServiceImpl.java
mohanarpit: @sumitsum In that case, can we please remove this special handling? If there is an "Internal server error" when fetching feature flags, we return the response as is to the client. The assumption is that the client's default flow will kick in.

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/ConsolidatedAPIResponseDTO.java
sumitsum: HI @rajatagrawal no, only view mode responses are returned in the view mode. There is a if condition check that checks for the `mode` attribute that is passed from each controller endpoint as a parameter. It is hardcoded so respective values for each endpoint.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ConsolidatedAPIServiceImpl.java
sumitsum: Hi @mohanarpit I would have check on this with @vsvamsi1 . I will post regarding this on the Slack thread so that we can resolve it there.

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/ConsolidatedAPIResponseDTO.java
rajatagrawal: Thanks, I looked at the file in the PR again and my queries are clear now. Thanks!

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ConsolidatedAPIServiceImpl.java
sumitsum: HI @mohanarpit a small clarification, this feature flag Mono actually gets used when fetching the page DSLs, which means this error could potentially show up in the response for page DSLs.

vsvamsi1: Yeah what ever errored response the feature flags results in please relay to that client, but the server flow should continue to determine other sub api responses.
cc @mohanarpit @sumitsum

PR #30108, opened 2024-01-08
fix: modal body

KelvinOm:
## Summary by CodeRabbit

- **New Features**
  - Updated theming capabilities for a more consistent and customizable user interface.

- **Style**
  - Enhanced Modal widget styling for better visual integration within the app.

- **Documentation**
  - Improved Storybook previews with custom font settings for a more accurate representation of components.

### DIFF
--- a/app/client/packages/design-system/theming/src/hooks/src/useCssTokens.tsx
+++ b/app/client/packages/design-system/theming/src/hooks/src/useCssTokens.tsx
@@ -82,11 +82,9 @@ export function useCssTokens(props: UseCssTokensProps) {
   }, [typography]);
 
   useEffect(() => {
-    if (fontFamily != null) {
-      setFontFamilyClassName(css`
-        ${fontFamilyCss(fontFamily)}
-      `);
-    }
+    setFontFamilyClassName(css`
+      ${fontFamilyCss(fontFamily)}
+    `);
   }, [fontFamily]);
 
   useEffect(() => {
--- a/app/client/packages/design-system/widgets/src/components/Modal/src/styles.module.css
+++ b/app/client/packages/design-system/widgets/src/components/Modal/src/styles.module.css
@@ -17,6 +17,14 @@
   gap: var(--outer-spacing-4);
   padding-inline: var(--outer-spacing-4);
   padding-block: var(--outer-spacing-4);
+
+  .body {
+    overflow: auto;
+    max-height: 100%;
+    padding-inline: var(--outer-spacing-4);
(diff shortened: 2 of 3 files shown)


app/client/packages/design-system/widgets/src/components/Modal/src/styles.module.css
```diff
@@ -17,6 +17,14 @@
   gap: var(--outer-spacing-4);
   padding-inline: var(--outer-spacing-4);
   padding-block: var(--outer-spacing-4);
+
+  .body {
```
KelvinOm, line 21: Just increased the specificity

app/client/packages/design-system/theming/src/hooks/src/useCssTokens.tsx
```diff
@@ -82,11 +82,9 @@ export function useCssTokens(props: UseCssTokensProps) {
   }, [typography]);
 
   useEffect(() => {
-    if (fontFamily != null) {
-      setFontFamilyClassName(css`
-        ${fontFamilyCss(fontFamily)}
-      `);
-    }
+    setFontFamilyClassName(css`
```
KelvinOm, line 85: There is no need to check here, since the check is in `fontFamilyCss` function.

app/client/packages/storybook/.storybook/preview-head.html
```diff
@@ -31,4 +31,8 @@
   a {
     color: var(--color-fg-accent) !important;
   }
+
+  .css-qa4clq :where(div:not(.sb-anchor, .sb-unstyled, .sb-unstyled div)) {
```
jsartisan, line 35: This seems wrong. `css-qa4clq` is unique string that will be different on each build no?

KelvinOm: Yes, this class will change for the new version of Storybook. Do you think it's better to extend this to all divs?

jsartisan: Extending to all divs is not a good option :| 
Wish there was an id that we could use. If that's not available. Then this is fine.

jsartisan (approved):

app/client/packages/storybook/.storybook/preview-head.html
KelvinOm: There are classes like `sbdocs sbdocs-content`. This look more reliable, but in anyway, when updating the version, everything can break here. I'll leave it as it is now.

PR #30162, opened 2024-01-09
fix: Validation appears momentarily in input widget

### LINKED ISSUE #29649: [Bug]: Validation appears momentarily in input widget

### Description

[WDS Input widget]
When we freshly drop an input widget (or refresh the page), and then type a space character in the input widget (the input widget has validation enabled), the error message momentarily shows up shifting the layout. 

### Steps To Reproduce

1. Add an input widget and configure a validation error message, and set the validation to return `{{false}}`
2. Type a "space" character in the input widget.

https://github.com/appsmithorg/appsmith/assets/103687/7882c0a5-9434-47c0-b753-d1057ea57aa2

### Public Sample App

_No response_

### Environment

Production

### Issue video log

_No response_

### Version

Cloud / Self Hosted - 1.9.57

jsartisan:
Fixes #29649

### DIFF
--- a/app/client/src/widgets/InputWidgetV2/widget/index.tsx
+++ b/app/client/src/widgets/InputWidgetV2/widget/index.tsx
@@ -838,6 +838,13 @@ class InputWidget extends BaseInputWidget<InputWidgetProps, WidgetState> {
         ? { autoComplete: "off" }
         : {};
 
+    //inputText is a state value and value is a derived, if they're not same , we are rendering with stale value.
+    // so we can ignore the error state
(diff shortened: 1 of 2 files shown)


jsartisan:
@KelvinOm I talked to @rishabhrathod01 about the derived properties running 1 render behind due to which we see the flicker for a millisecond. There is no easy fix to that root cause. For now, we can just follow the fix you gave for the old input widget.

jsartisan:
Closing the PR as the fix does not actually fix the problem.

PR #30171, opened 2024-01-10
chore: restrict use of deleted boolean field

sharat87:
Instead of removing the `deleted` field, we restrict its use, so it doesn't grow again.

More details in comments in cod.


mohanarpit (dismissed):

nidhi-nair (dismissed):

sharat87:
Test recording that this doesn't create a large diff.


https://github.com/appsmithorg/appsmith/assets/120119/f84db508-1de7-46dc-a879-de17cf93d4e8

PR #30288, opened 2024-01-13
chore: remove unused APPSMITH_TNC_PP env variable

sharat87:
## Summary by CodeRabbit

- **Refactor**
	- Removed configuration and references associated with `enableTNCPP` across various files.
- **Chores**
	- Cleaned up legacy configuration elements from the application setup.


mohanarpit (approved):

mohanarpit:
I'm approving this PR right now. Not sure how the functionality got removed. In the past, this variable would determine if we showed the Appsmith t&c and privacy policy for self-hosted branded instances. But right now, I don't see this variable being used anywhere and the lack of this functionality is clearly not causing any problems. 

Hence, let's remove it from the code base. But I'm adding this comment as a note to our future selves about why the variable was added in the first place.

KelvinOm (approved):

PR #30292, opened 2024-01-15
chore: Remove unused services

sharat87:


sharat87:
All Cypress tests run on EE PR and passed.

mohanarpit (approved):

mohanarpit:
@sharat87 In the next PR, do we also want to write migrations to delete the unused collections? Or do you believe that's not important right now?

sharat87:
@mohanarpit, thanks. I don't think its too important, but I do intend to look at it. If we do it though, I'd prefer it be an async migration, in that the server being ready shouldn't depend on this migration completing, or even be successful. If we can do it in that async fashion, only then will we have that migration.

nidhi-nair (approved):
At long last! Thanks for this <3

KelvinOm (approved):

PR #30364, opened 2024-01-17
chore: Cypress wait improvements to code editor focus and cursor checks

hetunandu:
Improves the cypress test method for editor focus and cursor checks removing a wait on focus and asserting instead. This should help in reducing flakiness

## Summary by CodeRabbit

- **Refactor**
	- Improved cursor positioning logic in code input fields for a more responsive user experience.

### DIFF
--- a/app/client/cypress/support/widgetCommands.js
+++ b/app/client/cypress/support/widgetCommands.js
@@ -516,7 +516,7 @@ Cypress.Commands.add(
       codeMirrorInput.focus();
       cy.wait(200);
       codeMirrorInput.setCursor(cursor);
-      cy.wait(1000); //time for value to set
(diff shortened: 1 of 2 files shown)


app/client/cypress/support/widgetCommands.js
```diff
@@ -516,7 +516,10 @@ Cypress.Commands.add(
       codeMirrorInput.focus();
       cy.wait(200);
       codeMirrorInput.setCursor(cursor);
-      cy.wait(1000); //time for value to set
+      cy.wait(200);
```
mohanarpit, line 519: Can we please remove this wait completely? We want to remove static waits like this in favour of dynamically waiting on an event like API response, element visibility, etc.

PR #30376, opened 2024-01-17
chore: aforce issue for multipart file upload 

sondermanish:
Fixes # (issue number)

### DIFF
--- a/app/server/appsmith-server/src/main/resources/application.properties
+++ b/app/server/appsmith-server/src/main/resources/application.properties
@@ -20,6 +20,7 @@ spring.data.mongodb.auto-index-creation=false
 spring.data.mongodb.authentication-database=admin
 # Ensures that the size of the request object that we handle is controlled. By default it's 212KB.
 spring.codec.max-in-memory-size=150MB
+spring.webflux.multipart.max-in-memory-size=150MB
 appsmith.codec.max-in-memory-size=${APPSMITH_CODEC_SIZE:150}
 
 # Log properties


app/server/appsmith-server/src/main/resources/application.properties
```diff
@@ -20,6 +20,7 @@ spring.data.mongodb.auto-index-creation=false
 spring.data.mongodb.authentication-database=admin
 # Ensures that the size of the request object that we handle is controlled. By default it's 212KB.
 spring.codec.max-in-memory-size=150MB
+spring.webflux.multipart.max-in-memory-size=150MB
```
nidhi-nair, line 23: Do you want to make this configurable for users as well like how the codec size is?

sondermanish: my bad, should have started with that, thanks for pointing out!

nidhi-nair: Sorry to nitpick, but we want to always minimize configuration as much as possible. If both those configurations are essentially solving the same problem, would it be possible to use the same environment variable to configure them?

app/server/appsmith-server/src/main/resources/application.properties
```diff
@@ -20,7 +20,9 @@ spring.data.mongodb.auto-index-creation=false
 spring.data.mongodb.authentication-database=admin
 # Ensures that the size of the request object that we handle is controlled. By default it's 212KB.
 spring.codec.max-in-memory-size=150MB
-appsmith.codec.max-in-memory-size=${APPSMITH_CODEC_SIZE:150}
+# Ensures the max size of multipart files being sent as request objects
+spring.webflux.multipart.max-in-memory-size=${APPSMITH_WEBFLUX_MUTIPART_SIZE:150MB}
+appsmith.codec.max-in-memory-size=${APPSMITH_CODEC_SIZE:150MB}
```
nidhi-nair, line 25: Doesn't codec need the value to be an integer in code?

Also, please chcek the other comment around using the same variable name is the use case is the same.

nidhi-nair (approved):

PR #30418, opened 2024-01-18
feat: add icon style in application settings

jsartisan:


app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Application.java
```diff
@@ -423,13 +423,25 @@ public static class ThemeSetting {
...
+        IconStyle iconStyle;
+
         public ThemeSetting(Type colorMode) {
             this.colorMode = colorMode;
         }
 
+        public ThemeSetting(IconStyle iconStyle) {
+            this.iconStyle = iconStyle;
+        }
+
```
AnaghHegde, line 436: Can you remove this? This is not required

AnaghHegde (changes requested):
Minor change, rest looks good

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Application.java
jsartisan: Removed. But the same code exists for colorMode. Hope removing it won't affect anything. What does it do anyway?

jsartisan: I have tested on local. Working fine

AnaghHegde (approved):

PR #30505, opened 2024-01-22
chore: Update WDS default seed to be distinct from Appsmith branding

### LINKED ISSUE #29758: chore: change default WDS seed color
Currently we use Appsmith's brand purple, this makes editor UI and widgets to look a bit too similar. We should consider a different default seed, e.g. `#0080ff`.

For follow-ups: Ideal product decision would be picking a random seed and other theme parameters on app creation (from a pre-selected nice-looking combinations).

ichik:
Closes #29758

## Summary by CodeRabbit

- **Style**
	- Updated the app's accent color to improve visual consistency and enhance user interface aesthetics.

### DIFF
--- a/app/client/packages/design-system/theming/src/token/src/defaultTokens.json
+++ b/app/client/packages/design-system/theming/src/token/src/defaultTokens.json
@@ -1,5 +1,5 @@
 {
-  "seedColor": "#553de9",
+  "seedColor": "#0080ff",
   "fluid": {
     "minVw": 360,
     "maxVw": 1920,
--- a/app/client/src/constants/AppConstants.ts
+++ b/app/client/src/constants/AppConstants.ts
@@ -132,7 +132,7 @@ export const defaultNavigationSetting = {
 
 export const defaultThemeSetting: ThemeSetting = {
   fontFamily: "System Default",
-  accentColor: "#553DE9",
+  accentColor: "#0080ff",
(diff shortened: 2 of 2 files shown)


app/client/packages/design-system/theming/src/token/src/defaultTokens.json
```diff
@@ -1,5 +1,5 @@
 {
-  "seedColor": "#553de9",
+  "seedColor": "#0080ff",
```
KelvinOm, line 2: This will affect only Storybrook. @jsartisan, could you help with appropriate changes in main app?

jsartisan: sure. I'll check. we need the change new wds theme settings right?

KelvinOm: Yes.

jsartisan:
@ichik changes done for the main app.

ichik:
@jsartisan what about all the tests that we have for old color? Is this safe to just merge?

jsartisan:
@ichik All those tests should work fine because i mainly changed the value of new theme settings and we don't have any tests for those and these wds theme settings are separate from old theme settings. so we should be good. I have run the ok-to-test. Let's see if something fails.

KelvinOm (approved):

PR #30571, opened 2024-01-24
fix: Allow incoming TLS traffic without restricting the domain

sharat87:
This is a fix for a user's problem. They have custom domain set, a custom cert in the `stacks/ssl` folder, but because a different team operates a reverse-proxy, they aren't sure which _host_ is actually used by the reverse proxy. And the way we bind to port 443 requires that that puzzle be solved, for very little extra value.

This change makes it so that we accept any incoming TLS connections, if a custom domain is set, which should be much more convenient.

[Slack Thread](https://theappsmith.slack.com/archives/C0341RERY4R/p1705700120412079).

Already deployed on users' system, and they've confirmed its working.

## Summary by CodeRabbit

- **New Features**
	- Improved `https` URL routing to automatically adapt based on custom certificate availability.
- **Documentation**
	- Added explanatory comments for domain requirements in TLS configuration.
- **Refactor**
	- Enhanced server response header management for better security compliance.

### DIFF
--- a/deploy/docker/fs/opt/appsmith/caddy-reconfigure.mjs
+++ b/deploy/docker/fs/opt/appsmith/caddy-reconfigure.mjs
@@ -31,8 +31,6 @@ if (CUSTOM_DOMAIN !== "") {
 
 }
 
-const tlsConfig = certLocation == null ? "" : `tls ${certLocation}/fullchain.pem ${certLocation}/privkey.pem`
-
 const frameAncestorsPolicy = (process.env.APPSMITH_ALLOWED_FRAME_ANCESTORS || "'self'")
   .replace(/;.*$/, "")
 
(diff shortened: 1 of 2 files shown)


deploy/docker/fs/opt/appsmith/caddy-reconfigure.mjs
```diff
@@ -131,7 +131,7 @@ if (CUSTOM_DOMAIN !== "") {
   // If no custom domain, no extra routing needed.
   // We have to own the http-to-https redirect, since we need to remove the `Server` header from the response.
   parts.push(`
-  https://${CUSTOM_DOMAIN} {
+  https:// {
```
mohanarpit, line 134: Do we have to make the same change in the http:// block below? I'm asking this for my knowledge.

mohanarpit (dismissed):

deploy/docker/fs/opt/appsmith/caddy-reconfigure.mjs
sharat87: No actually, that'll likely break. Because we're binding to port `80` a little above on all hosts, if we remove custom domain in `http://` block, it'll be a duplicate configuration. The `:80` and `http://` syntaxes are equivalent to Caddy.

pratapaprasanna (approved):

PR #30576, opened 2024-01-24
fix: setup proxy variables in the start

pratapaprasanna:
## Summary by CodeRabbit

- **Refactor**
	- Enhanced proxy configuration handling in the initialization script for improved efficiency and reliability.


sharat87 (commented):
Let's add timeout on the curl requests as well and we should be ready to merge.

sharat87 (changes requested):
@pratapaprasanna, it looks like we're doing `setup_proxy_variables` **before** `init_env_file`, which means we're not loading `docker.env` before setting up proxy variables. So if the proxy is configured in `docker.env`, we won't be capturing it, right?

PR #30638, opened 2024-01-25
fix: fix modal states

KelvinOm:


app/client/packages/design-system/widgets/src/components/Modal/src/ModalHeader.tsx
```diff
@@ -20,17 +20,12 @@ export const ModalHeader = (props: ModalHeaderProps) => {
...
-    setOpen(false);
-  };
-
   return (
     <Flex alignItems="center" gap="spacing-4" justifyContent="space-between">
       <Text id={id} lineClamp={1} title={title} variant="caption">
         {title}
       </Text>
-      <IconButton icon="x" onPress={closeHandler} variant="ghost" />
+      <IconButton icon="x" onPress={() => setOpen(false)} variant="ghost" />
```
riodeuno, line 28: Does this make sure to run the `onClose`? `onClose` may have bindings, so they need to run.

KelvinOm: Yes, we called `onClose` [here](https://github.com/appsmithorg/appsmith/blob/master/app/client/packages/design-system/headless/src/components/Popover/src/PopoverContent.tsx#L34) immediately after modal is unmounted

PR #30669, opened 2024-01-27
chore: Fluent interface for queryAll method

sharat87:
🚨 **DO NOT MERGE. ONLY REVIEW/APPROVE.**

There is ~13 signatures of the `queryAll` method with parameters ranging from 2 to 7, different combinations of the same superset. This is extremely hard to grok and the resulting uses of `queryAll` aren't very readable either. This PR introduces a fluent interface for this method. If this works better for us, I'll look to move more such over-signature-ed methods to a fluent interface.

This sprang from a frustration in trying to understand this method's usages, to port to Postgres, and pulling my hair apart in the process.

This is an incomplete PR. I'll finish it, if we're aligned that this is a good idea, so I may have asked for a premature PR review. Thanks!

Here's an example of the fluent API compared to current API:
shot-2024-01-27-10-37-57@2x

## Summary by CodeRabbit

- **Refactor**
	- Enhanced query building and execution process for improved performance and maintainability across various data access methods.


trishaanand:
@sharat87 :  Aligned on the direction here.

sharat87:
Opening a new one.

PR #30765, opened 2024-01-31
wip

sharat87:
This is already there in AppsmithRepository.


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserDataServiceCEImpl.java
```diff
@@ -311,7 +311,7 @@ protected List<String> addIdToRecentList(List<String> srcIdList, String newId, i
         }
         // keeping the last maxSize ids, there may be a lot of ids which are not used anymore
         if (srcIdList.size() > maxSize) {
-            srcIdList = srcIdList.subList(0, maxSize);
+            srcIdList.subList(maxSize, srcIdList.size()).clear();
```
sharat87, line 314: Instead or getting the sub-list and returning that, which would be an instance of `ArrayList$SubList`, we instead remove the extra items in the original `ArrayList` and return that itself.

This is because the `SubList` objects are just _views_ on the underlying `ArrayList` object, and cannot be serialized by Hibernate directly.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/PermissionGroupRepositoryCE.java
```diff
@@ -14,8 +12,6 @@ public interface PermissionGroupRepositoryCE
 
     Flux<PermissionGroup> findAllById(Set<String> ids);
 
-    Mono<PermissionGroup> findById(String id, AclPermission permission);
```
sharat87, line 17: This is defined in the custom repo interface as well, so not needed here.

sharat87:
Test failure because of mocked mongo template, which needs to go away anyway in moving to Postgres.

PR #30778, opened 2024-01-31
chore: use default spring query instead of custom one

sharat87:
## Summary by CodeRabbit

- **Refactor**
	- Enhanced user lookup in the authentication flow to improve reliability and consistency by modifying the method for finding users by email. This includes a fallback mechanism to find users by case-insensitive email, prioritizing the most recently created account.
- **Tests**
	- Updated unit and repository tests to align with the new method of user lookup based on case-insensitive email.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/handlers/ce/CustomFormLoginServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/handlers/ce/CustomFormLoginServiceCEImpl.java
@@ -33,7 +33,7 @@ public class CustomFormLoginServiceCEImpl implements ReactiveUserDetailsService 
     public Mono<UserDetails> findByUsername(String username) {
         return repository
                 .findByEmail(username)
-                .switchIfEmpty(repository.findByCaseInsensitiveEmail(username))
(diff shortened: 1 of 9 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/handlers/ce/CustomFormLoginServiceCEImpl.java
```diff
@@ -33,7 +33,7 @@ public CustomFormLoginServiceCEImpl(UserRepository repository) {
     public Mono<UserDetails> findByUsername(String username) {
         return repository
                 .findByEmail(username)
-                .switchIfEmpty(repository.findByCaseInsensitiveEmail(username))
+                .switchIfEmpty(repository.findFirstByEmailIgnoreCaseOrderByCreatedAtDesc(username))
```
trishaanand, line 36: Lets get rid of order by considering we are looking for only one user?

sharat87: We're looking for one user only, which makes order important. What if there's two user documents in the collection, with different casing of the email. We want to consistently _always_ fetch the same user document, whichever that may be. Our tests actually assert this. If I change that order to `Asc`, the test fails.

trishaanand (approved):

Issue #30764, opened 2024-01-31
[Bug]: On creating a new query, the UI of the current query page gets displayed in a different manner


shadabbuchh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

When I click on `New query/API`, the UI of the existing query page is displayed in a different way & a lot of information is missing in the new layout.

### Steps To Reproduce

1. Go to https://app.appsmith.com/.
2. Click on `Queries` & then on `New query/API`.
3. Now again click on the `New query/API`.

**Expected:** The existing query page should not change & all the relevant information should be present.
**Actual:** The layout of the existing query page changes.

### Environment

Production

### Issue video log

https://www.loom.com/share/8d756dbd22c74f1d80ff042b9db244d3

### Version

Cloud

shadabbuchh:
Issues fixed. Verified on release.

https://github.com/appsmithorg/appsmith/assets/39921438/ee7e2369-d5a9-4864-9bfb-97b358925970

PR #30790, opened 2024-01-31
fix: improve query add url split

hetunandu:
improve upon the query add url fix by splitting it for extensibility with EE

We will now extend a utility function `getQueryAddUrl` on EE to add EE cases instead of rewriting the implementation as it does today

fixes: #30764

## Summary by CodeRabbit

- **Refactor**
	- Enhanced URL construction method for adding queries in the Editor Pane to improve navigation and consistency.

### DIFF
--- a/app/client/src/ce/pages/Editor/IDE/EditorPane/Query/hooks.ts
+++ b/app/client/src/ce/pages/Editor/IDE/EditorPane/Query/hooks.ts
@@ -3,8 +3,6 @@ import history from "utils/history";
 import { LIST_PATH } from "@appsmith/constants/routes/appRoutes";
 import { useLocation } from "react-router";
 import { FocusEntity, identifyEntityFromPath } from "navigation/FocusEntity";
-import { useCurrentEditorState } from "pages/Editor/IDE/hooks";
-import { EditorEntityTabState } from "@appsmith/entities/IDE/constants";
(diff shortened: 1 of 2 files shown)


albinAppsmith (approved):

app/client/src/ce/pages/Editor/IDE/EditorPane/Query/hooks.ts
```diff
@@ -20,35 +12,16 @@ import { getHasCreateActionPermission } from "@appsmith/utils/BusinessFeatures/p
...
-              pluginPackageName: currentEntityInfo.params.pluginPackageName,
-              add: true,
-            });
-          }
-      }
-    }
+    const url = getQueryAddUrl(currentEntityInfo);
     history.push(url);
-  }, [currentEntityInfo.id, location, segmentMode]);
+  }, [currentEntityInfo.id]);
```
sneha122, line 24: @hetunandu What if some other property in currentEntityInfo changes, the callback function will not get updated as it caches based on currentEntityInfo.id, will this be a problem?

hetunandu: No because the `id` is the most important aspect here. A different id is 99.9999 % scenarios will mean a different type of entity. Adding id instead of the object so that the hook caching is more efficient

PR #30804, opened 2024-01-31
chore: Use Spring projection for workspace user pics

sharat87:
We use Spring's native support for projections here, instead of setting the `.fields()` ourselves. The advantage is that this way translates directly to Postgres.

The cost is that, the field names in projection are duplicated. For example, if `profilePhotoAssetId` is renamed or otherwise changed, we won't see a compile error. This can be prevented with a test. The `findPhotoAssetsByUserIds_WhenPhotoAssetIdExist_ReturnsPhotoAssetId` test is able to catch this case perfectly well.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new interface for retrieving user profile photo information, enhancing the way profile photos are handled and displayed.
- **Refactor**
  - Removed and optimized various methods across the server codebase to improve the efficiency of profile photo data retrieval.
  - Simplified service dependencies, focusing on streamlined access to user profile photo assets.
- **Tests**
  - Updated tests to align with the new approach for fetching and asserting profile photo information.


trishaanand (dismissed):

app/server/appsmith-server/src/main/java/com/appsmith/server/projections/ce/UserDataProfilePhotoProjection.java
```diff
@@ -0,0 +1,7 @@
+package com.appsmith.server.projections.ce;
+
+public interface UserDataProfilePhotoProjection {
```
trishaanand, line 3: Since this file exists in CE, the interface name should be `UserDataProfilePhotoProjectionCE`. You would need to introduce `UserDataProfilePhotoProjection` which extends `UserDataProfilePhotoProjectionCE`.

trishaanand: In this case, the EE file (main file : UserDataProfilePhotoProjection) would be empty and only be extending the CE interface.

sharat87: Done.

trishaanand (approved):

Issue #7358, opened 2021-09-12
[Bug] Group Option Popup zindex (hides inside the Form Widget container)


rishabhrathod01:
## Description

[What happened]
<img width="423" alt="Screenshot 2021-09-10 at 9 40 24 PM" src="https://user-images.githubusercontent.com/23132741/133003381-49abf3c3-363a-4399-81c4-cece562fbe3f.png">

### Steps to reproduce the behaviour:

Add steps to reproduce this behaviour, include console / network logs & screenshots

1. Go to New Application and Go to any Empty Page
2. Add a form widget and decrease its size as shown in image.
3. Try to group the buttons

### Important Details

- Version: [Cloud / Self-Hosted vx.x]
- OS: [e.g.MacOSX]
- Environment [production, release, deploy preview]

Sarthakagarwal22:
I would like to pick up this issue

somangshu:
Greetings @Sarthakagarwal22, thanks for showing interest 🎉 , This is all your. Assigning this to you now. 

Please don't forget to read the [Contribution Guidelines](https://github.com/appsmithorg/appsmith/blob/master/CONTRIBUTING.md). Would appreciate if you can open a PR within the next 2 days.

Also, please don't forget to register with [Appsmith's Hacktoberfest Community](https://hacktoberfest.appsmith.com/) in order to be eligible for Appsmith swag.

Sarthakagarwal22:
<img width="531" alt="Screenshot 2021-10-05 at 9 56 41 AM" src="https://user-images.githubusercontent.com/16587695/135960003-076aab15-7aa5-423a-b386-9903234e2ee1.png">
Does this look like a valid solution

somangshu:
@jsartisan need you to check this out

jsartisan:
@Sarthakagarwal22 the valid solution for this is to use the portal to render the context menu. The reason why this is happening is that the context menu is rendering in the same container.

Sarthakagarwal22:
@jsartisan I would need some help here. I tried using the portal, but I couldn't understand the flow to be honest. A basic approach that I took was, added a css property, which made it work in form. Can we pair up somehow to solve this?

jsartisan:
Yes. Let's have a call tomorrow? Anytime between 10-7.

jsartisan:
@Sarthakagarwal22 Can you come on discord?

Sarthakagarwal22:
[...]

PR #30813, opened 2024-02-01
fix: display of group option popup inside Form Widget Container

Rhitottam:
## Description
This fix uses React portal to render the group option popup outside the component and position it based on the position of the grouped components. This allows the popup to be shown without any z-index issues. 
Fixes # (7358)https://github.com/appsmithorg/appsmith/issues/7358
#### Media
Screenshot 2024-02-01 at 9 44 27 AM
Screenshot 2024-02-01 at 9 44 41 AM

#### Type of change

- [ ] Manual
- [ ] JUnit
- [ ] Jest
- [ ] Cypress
>
>
#### Dev activity

#### QA activity:

## Summary by CodeRabbit

- **Refactor**
    - Enhanced state management for widget grouping.
    - Improved context menu rendering for performance and flexibility.

### DIFF
--- a/app/client/src/layoutSystems/fixedlayout/common/widgetGrouping/WidgetsMultiSelectBox.tsx
+++ b/app/client/src/layoutSystems/fixedlayout/common/widgetGrouping/WidgetsMultiSelectBox.tsx
@@ -403,13 +403,7 @@ function WidgetsMultiSelectBox(props: {
       <StyledSelectBoxHandleLeft />
       <StyledSelectBoxHandleRight />
       <StyledSelectBoxHandleBottom />
-      <StyledActionsContainer
-        // height={height}
-        // left={boundLeft ?? undefined}
-        ref={menuRef}
-        // top={boundTop ?? undefined}
-        // visibility={menuVisible}
-      >
+      <StyledActionsContainer ref={menuRef}>
         {/* copy widgets */}
         <Tooltip
           boundary="viewport"
--- a/app/client/src/layoutSystems/fixedlayout/common/widgetGrouping/WidgetsMultiSelectBox.tsx
+++ b/app/client/src/layoutSystems/fixedlayout/common/widgetGrouping/WidgetsMultiSelectBox.tsx
@@ -188,32 +188,42 @@ function WidgetsMultiSelectBox(props: {
   const isDragging = useSelector(
     (state: AppState) => state.ui.widgetDragResize.isDragging,
   );
+  /**
+   * Observer to track the position of the multi-selection bounding box
+   * and update the position of the context menu accordingly
+   */
+  const positionObserver = React.useRef<IntersectionObserver>(
+    new IntersectionObserver(
+      ([node]) => {
+        if (menuRef.current) {
+          requestAnimationFrame(() => {
+            if (menuRef.current) {
+              menuRef.current.style.top = `${node.intersectionRect.top}px`;
+              menuRef.current.style.left = `${node.intersectionRect.left}px`;
+              menuRef.current.style.visibility = node.isIntersecting
+                ? "visible"
+                : "hidden";
+              menuRef.current.style.flexDirection =
+                node.intersectionRect.height < 160 ? "row" : "column";
+            }
+          });
+        }
+      },
+      {
+        root: null,
+        threshold: [0, 0.125, 0.25, 0.5, 0.625, 0.75, 0.875, 1],
+      },
+    ),
+  );
+  /**
+   * Update the bounding rectangle to handle scroll, resize events
+   */
   const updateBoundingClientRect = () => {
     const node = draggableRef.current;
-    if (node) {
-      const visibilityObserver = new IntersectionObserver(
-        ([node]) => {
-          if (menuRef.current) {
-            requestAnimationFrame(() => {
-              if (menuRef.current) {
-                menuRef.current.style.top = `${node.intersectionRect.top}px`;
-                menuRef.current.style.left = `${node.intersectionRect.left}px`;
-                menuRef.current.style.visibility = node.isIntersecting
-                  ? "visible"
-                  : "hidden";
-                menuRef.current.style.flexDirection =
-                  node.intersectionRect.height < 160 ? "row" : "column";
-              }
-            });
-          }
-        },
-        {
-          root: null,
-          threshold: [0, 0.125, 0.25, 0.5, 0.625, 0.75, 0.875, 1],
-        },
-      );
-      visibilityObserver.disconnect();
-      visibilityObserver.observe(node);
+    const observer = positionObserver.current;
+    if (observer) {
+      observer.disconnect();
+      if (node) observer.observe(node);
     }
   };
   /**
@@ -298,7 +308,7 @@ function WidgetsMultiSelectBox(props: {
   ]);
 
   /**
-   * Update the component positions whenever the component re-renders
+   * Update the component positions whenever the component re-renders and check for updates at regular intervals
    */
   useEffect(() => {
     if (shouldRender) updateBoundingClientRect();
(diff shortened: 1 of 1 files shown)


Rhitottam:
Hey @somangshu  and @rishabhrathod01 , could you please take a moment to review this fix when you have some free time? I haven't addressed the related Cypress tests yet, but I'll get to them once this fix is confirmed. Thanks a lot!"

somangshu:
This is a much more involved issue.
@marks0351 WDYT?

marks0351:
@Rhitottam I think pop up approach is the right way for this issue.
However I see some issues in the dp, did not extensively test it.
- When I select only two widgets widget grouping actions should appear, it doesn't right now unless I move them explicitly creating a rerender.
- The actions popup is not sticking to the widgets so when I scroll I see them in the same position.
- In Modal widgets case, actions are not accessible.

Pls do check usecases with nested containers as well as  nested and scrollable containers.

Rhitottam:
Hi @marks0351 Thanks for the feedback! I will check it out!

Rhitottam:
Hi @marks0351, I've made some changes to address the initial render issue. To handle scrolling or window resizing issues, I've utilized setInterval to update the position of the options menu. Would you mind taking a look when you have a moment?   Thanks a lot! [Widgets Multi-Select Box actions menu_ Appsmith.webm](https://github.com/appsmithorg/appsmith/assets/4265002/6cc1c540-c09a-4951-911c-a0ff7ce9ab2f)

app/client/src/layoutSystems/fixedlayout/common/widgetGrouping/WidgetsMultiSelectBox.tsx
```diff
@@ -248,6 +275,33 @@ function WidgetsMultiSelectBox(props: {
     props.noContainerOffset,
   ]);
 
+  /**
+   * Update the component positions whenever the component re-renders
+   */
+  useEffect(() => {
+    const updateBoundingClientRect = () => {
```
marks0351, line 282: I think you can use IntersectionObserver instead to do this. you would have more info to make sure the options are always available on the viewable part of the widgets.
Also try using debounce/throttle to make sure the options position is updated in a performant way.
Additionally try using css transitions to make sure position changes are percieved smoother.

marks0351: This is great progress btw.

marks0351 (changes requested):

app/client/src/layoutSystems/fixedlayout/common/widgetGrouping/WidgetsMultiSelectBox.tsx
Rhitottam: Thanks for the feedback @marks0351 . Interesting! I hadn't used Observers for a while. It will be a great way to check how it works! Thanks a lot!!

app/client/src/layoutSystems/fixedlayout/common/widgetGrouping/WidgetsMultiSelectBox.tsx
```diff
@@ -248,6 +288,43 @@ function WidgetsMultiSelectBox(props: {
...
+
+    return () => {
+      if (intervalId) clearInterval(intervalId);
+    };
+  }, [shouldRender, menuVisible]);
+  /**
+   * Check if the multi-select box is shown within the viewport
+   */
+  useEffect(() => {
+    const visibilityObserver = new IntersectionObserver(
```
marks0351, line 312: @Rhitottam can you move this outside the useeffect and create a useref for it.
you can check `app/client/src/layoutSystems/common/canvasArenas/StickyCanvasArena.tsx` for this reference.

app/client/src/layoutSystems/fixedlayout/common/widgetGrouping/WidgetsMultiSelectBox.tsx
```diff
@@ -248,6 +288,43 @@ function WidgetsMultiSelectBox(props: {
...
 
+  /**
+   * Update the component positions whenever the component re-renders
+   */
+  useEffect(() => {
+    if (shouldRender && menuVisible) updateBoundingClientRect();
+    // Update the bounding rectangle to handle scroll, resize events
+    const intervalId =
+      shouldRender && menuVisible
+        ? setInterval(() => {
```
marks0351, line 299: I would suggest to remove this delay or make it less 50ms, create a ref for the group actions popup/modal and update its styles on the dom directly instead of using states. 
let me know what you think.

marks0351:
@Rhitottam this is already great, One nit is that the popup is sticky but is not usable if the top most widget is the group is only partialy visible. 
Screenshot 2024-02-08 at 7 16 58 PM

you can use the entries thrown by the IntersectionObserver callback to get the intersection positions to reposition the pop up.
This file `app/client/src/layoutSystems/common/canvasArenas/StickyCanvasArena.tsx` has such implementation for reference.

app/client/src/layoutSystems/fixedlayout/common/widgetGrouping/WidgetsMultiSelectBox.tsx
Rhitottam: Hi @marks0351 , I have kept the delay so as to avoid calculating the bound so often. It causes a reflow which might cause performance issues. Will check if the styles can be set directly!!

app/client/src/layoutSystems/fixedlayout/common/widgetGrouping/WidgetsMultiSelectBox.tsx
```diff
@@ -248,6 +307,22 @@ function WidgetsMultiSelectBox(props: {
     props.noContainerOffset,
   ]);
 
+  /**
+   * Update the component positions whenever the component re-renders and check for updates at regular intervals
+   */
+  useEffect(() => {
+    if (shouldRender) updateBoundingClientRect();
```
marks0351, line 314: @Rhitottam observing and unobserving based on shouldRender flag should suffice here right? or am I missing something?

app/client/src/layoutSystems/fixedlayout/common/widgetGrouping/WidgetsMultiSelectBox.tsx
```diff
@@ -168,6 +188,44 @@ function WidgetsMultiSelectBox(props: {
...
+   */
+  const positionObserver = React.useRef<IntersectionObserver>(
+    new IntersectionObserver(
+      ([node]) => {
+        if (menuRef.current) {
+          requestAnimationFrame(() => {
+            if (menuRef.current) {
+              menuRef.current.style.top = `${node.intersectionRect.top}px`;
+              menuRef.current.style.left = `${node.intersectionRect.left}px`;
+              menuRef.current.style.visibility = node.isIntersecting
```
marks0351, line 203: @Rhitottam can you do the isIntersecting check and go ahead with making position based css changes coz when the popup disappears there is a bad UX where it jumps to position 0,0 while getting hidden.

app/client/src/layoutSystems/fixedlayout/common/widgetGrouping/WidgetsMultiSelectBox.tsx
Rhitottam: HI @marks0351 , the observer has been included within the setInterval function to facilitate obtaining updated positions of the widget select box. This ensures that we capture changes whenever the window or the widget parent undergoes scrolling or resizing.  The intersection observer on the box might not function correctly when the select box is entirely within the viewport, and the intersection ratio remains unchanged.

app/client/src/layoutSystems/fixedlayout/common/widgetGrouping/WidgetsMultiSelectBox.tsx
Rhitottam: Ahh yes! i will only update the visibility style if its not intersecting

app/client/src/layoutSystems/fixedlayout/common/widgetGrouping/WidgetsMultiSelectBox.tsx
marks0351: ah understood.
I think one issue stems from the fact that the portal/popup is on the document.body, so if we can create the same on the StyledSelectionBox by using a ref that should fix the need for update when scrolling.
Now for the other one, scaling/resizing you can try creating an effect with selectedWidgets state as dependecny and run updateBoundingClientRect.

side note: can we rename updateBoundingClientRect to observeSelectionBox to make it more readable.

app/client/src/layoutSystems/fixedlayout/common/widgetGrouping/WidgetsMultiSelectBox.tsx
marks0351: had looked at the deploy preview, The UX now is spot on. Great progress @Rhitottam 🚀

app/client/src/layoutSystems/fixedlayout/common/widgetGrouping/WidgetsMultiSelectBox.tsx
Rhitottam: Adding the popup to the StyledSelectionBox would fix it. But won't that bring back the issue we are fixing, where the popup is clipped whenever the box is inside another container like the form widget. But maybe, we won't need to create a portal at all, since the popup is mostly inside the styled selection box. Let me check!! Thanks a lot!

marks0351: you are right it will be clipped when only a very small portion of the StyledSelectionBox is visibile, which is fine. I would suggest to add a check for intersection ratio or height so that actions are not visible when the viewable StyledSelectionBox height is just very low.

Rhitottam: Hi @marks0351 ,  I noticed this after moving the popup inside the StyledSelectionBox, in a scenario where the box is contained within a nested container with a height less than that of the box, there's a potential issue. Specifically, if the nested container is scrolled, the observer may not be triggered because there's no change in the intersection height. However it is visible again on scrolling back. 
Can take a look here:-

https://github.com/appsmithorg/appsmith/assets/4265002/9db7586f-ae0d-4ee1-a56b-89d3acc5c228

marks0351 (approved):

marks0351:
@Rhitottam can you pls pull the latest release branch, so that I can run tests on top of this change.

marks0351:
@Rhitottam can you pull release once more

PR #30877, opened 2024-02-02
chore: get rid of fluid tokens

KelvinOm:
Get rid of fluid tokens. I left the original scale function unchanged, I use the min value for tokens.

## Summary by CodeRabbit

- **Refactor**
    - Simplified fluid sizing calculations across the design system for consistency and performance.
    - Improved state management in ThemeProvider for more responsive theming.
- **Style**
    - Updated Modal component padding for better visual spacing.
- **Chores**
    - Updated various Storybook configurations and styling for better performance and usability.


ichik (approved):
Seems like we're skewing a bit too much towards “everything is super small” state of affairs. Widgets on default density/sizing are a bit too small and when setting sizing to smallest setting unacceptably small, but otherwise everything works. I should re-tweak the coefficients in a follow-up.

KelvinOm:
> Seems like we're skewing a bit too much towards “everything is super small” state of affairs. Widgets on default density/sizing are a bit too small and when setting sizing to smallest setting unacceptably small, but otherwise everything works. I should re-tweak the coefficients in a follow-up.

We can make config fixes with this PR or as a separate task. Whichever is more convenient for you.

ichik:
@KelvinOm I'll need some time anyway and if we keeps PR small perhaps it'll help us later, I'll just raise a follow-up.

jsartisan (dismissed):

jsartisan (approved):

jsartisan (approved):

PR #30888, opened 2024-02-05
ci: Add server failed tests as PR comment

sharat87:
And, don't show duplicates. Not sure why we're seeing duplicates though.

### DIFF
--- a/.github/workflows/server-build.yml
+++ b/.github/workflows/server-build.yml
@@ -189,6 +189,7 @@ jobs:
                 sed 's/^/- /' 'failed-server-tests.txt'
               )"
               echo "$content" >> "$GITHUB_STEP_SUMMARY"
+              # Add a comment to the PR with the list of failed tests.
               curl --header "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
                 --data "$(jq -n --arg body "$content" '$ARGS.named')" \
                 "https://api.github.com/repos/$GITHUB_REPOSITORY/issues/${{ inputs.pr }}/comments"
--- a/.github/workflows/server-build.yml
+++ b/.github/workflows/server-build.yml
@@ -183,13 +183,14 @@ jobs:
           args+=("-DtestResultFile=$PWD/failed-server-tests.txt")
           if ! mvn test "${args[@]}"; then
             if [[ -s failed-server-tests.txt ]]; then
-              {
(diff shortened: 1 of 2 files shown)


app/server/appsmith-server/src/test/java/com/appsmith/server/TestResultLoggerExtension.java
```diff
@@ -20,7 +20,7 @@
  */
 public class TestResultLoggerExtension implements TestWatcher, AfterAllCallback {
 
-    private final List<String> failedTests = new ArrayList<>();
+    private final Collection<String> failedTests = new TreeSet<>();
```
sharat87, line 23: Using `TreeSet` so the list is sorted.

.github/workflows/server-build.yml
```diff
@@ -183,11 +183,15 @@ jobs:
...
+              content="$(
+                echo "## Failed server tests"
                 echo
                 sed 's/^/- /' 'failed-server-tests.txt'
-              } >> "$GITHUB_STEP_SUMMARY"
+              )"
+              echo "$content" >> "$GITHUB_STEP_SUMMARY"
+              curl --header "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+                --data "$(jq -n --arg body "$content" '$ARGS.named')" \
+                "https://api.github.com/repos/$GITHUB_REPOSITORY/issues/${{ inputs.pr }}/comments"
```
mohanarpit, line 194: @sharat87 Can you help me understand this change? What is this curl command doing?

mohanarpit: I guessed the functionality. But I'd still appreciate a comment because the body of the curl command has `ARGS.named` and I was unable to picture what will get posted on the PR eventually. Not the highest of priorities. Can ignore also.

mohanarpit (approved):

.github/workflows/server-build.yml
sharat87: Oh that, that's a `jq` command to _build_ JSON. Constructing a JSON string is a little hard in shell scripts. This is similar to how we build `info.json` here: https://github.com/appsmithorg/appsmith/blob/1afd321e13a021583f6054285ba368c477e5f7b4/scripts/generate_info_json.sh#L25-L33

PR #30897, opened 2024-02-05
feat: add ratelimiting for caddy

pratapaprasanna:
Fixes: [31031](https://github.com/appsmithorg/appsmith/issues/31031)

## Summary by CodeRabbit

- **New Features**
	- Added rate-limiting functionality to enhance security and prevent abuse.

### DIFF
--- a/deploy/docker/base.dockerfile
+++ b/deploy/docker/base.dockerfile
@@ -1,7 +1,7 @@
 FROM caddy:builder-alpine AS caddybuilder
 
 RUN xcaddy build \
-		--with github.com/mholt/caddy-ratelimit
+  --with github.com/mholt/caddy-ratelimit
 
 FROM ubuntu:20.04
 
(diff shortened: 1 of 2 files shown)


sharat87 (changes requested):
Let's confirm if `xcaddy build` is building the latest tagged version of Caddy or the current `master` branch. We need the latest tagged version to be used here.

sharat87 (approved):

sharat87 (changes requested):
Actually, let's do this in two separate PRs. Let's only merge the changes to `base.dockerfile`, and that base image is built, _then_ we merge the changes to the configure script. Otherwise all PRs' CIs will be broken until the next release image is built.

sharat87:
@pratapaprasanna, one more thing, can we confirm if the rate limiting happens on remote-ip, even when there's a load balancer? Like, does it apply the limiting rule on the load balancer's IP address, or on the IP address from `X-Forwarded-For` header?

sharat87 (approved):

PR #30898, opened 2024-02-05
chore: Fluent API for `count()`

sharat87:
There's three signatures for `count()`, and the query building code is duplicated from the `query*` methods. This PR fixes both.

## Summary by CodeRabbit

- **Refactor**
	- Improved database query methods for better performance and maintainability.
	- Enhanced permission handling in database operations.
- **Bug Fixes**
	- Removed redundant and unused methods, enhancing the application's efficiency.
- **Tests**
	- Updated tests to align with the refactored codebase, ensuring reliability.


app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/BaseAppsmithRepositoryCEImpl.java
```diff
@@ -152,47 +144,20 @@ protected DBObject getDbObject(Object o) {
...
     }
 
+    /**
+     * @deprecated using `Optional` for function arguments is an anti-pattern.
+     */
+    @Deprecated
     public Mono<T> findById(String id, Optional<AclPermission> permission) {
-        return findById(id, null, permission);
+        return findById(id, permission.orElse(null));
     }
```
sharat87, line 161: IntelliJ even has a big detailed warning about this. We need to reduce our usage of this anti-pattern.

shot-2024-02-05-09-41-40

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/BaseAppsmithRepositoryCEImpl.java
```diff
@@ -152,47 +144,20 @@ protected DBObject getDbObject(Object o) {
...
-
-            return mongoOperations
-                    .query(this.genericDomain)
-                    .matching(query.cursorBatchSize(10000))
-                    .one()
-                    .flatMap(obj -> setUserPermissionsInObject(obj, permissionGroups));
-        });
+        return queryBuilder()
+                .byId(id)
+                .fields((List<String>) null)
```
mohanarpit, line 150: @sharat87 Is this required? Since it's a builder, can it be avoided?

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/BaseAppsmithRepositoryCEImpl.java
```diff
@@ -481,6 +415,15 @@ public Mono<T> queryFirstExecute(QueryAllParams<T> params) {
                         .flatMap(obj -> setUserPermissionsInObject(obj, permissionGroups1)));
     }
 
+    public Mono<Long> countExecute(QueryAllParams<T> params) {
```
nidhi-nair, line 418: We're conflating `countWithProvidedPermissions` with `countWithAllApplicablePermissions` here. Can we also refactor this so that `countWithAllApplicableStrictPermissions` is also accounted for? Let's also provide docs so it's easier to know what to expect at this point.

The strictPermissions bit might be applicable to the entire builder paradigm. The difference is that all permissions includes anon permissions as well for every user, while strict is just what the user has been assigned.

nidhi-nair (changes requested):
Thank you for also cleaning up our repo layer along the way! ❤️

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/BaseAppsmithRepositoryCEImpl.java
sharat87: Ah, I searched for and removed ones that had `List.of()`. Didn't notice this. Thanks.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/BaseAppsmithRepositoryCEImpl.java
sharat87: Thanks for the pointers @nidhi-nair. I'll pick this up in the next PR for this refactor.

nidhi-nair (approved):

PR #30994, opened 2024-02-08
test: Fix flakiness in fork app test

sarojsarab:
## Description
- Fixing flakiness in ForkApplication_spec.ts
-- Changed the sequence as the sequence is causing the issue
-- Removed the addDsl() from before and added to the test it is necessary.

#### Type of change
- Cypress
- [x] Cypress
>
>

## Summary by CodeRabbit

- **Tests**
	- Updated regression test cases related to forking applications.
	- Modified paths to test spec files for improved organization.


Aishwarya-U-R (dismissed):

app/client/cypress/e2e/Regression/ClientSide/Fork/ForkApplication_spec.ts
```diff
@@ -30,30 +28,62 @@ describe(
...
+      embedSettings.ToggleMarkForkable();
+      embedSettings.TogglePublicAccess();
+
+      inviteModal.OpenShareModal();
+      homePage.InviteUserToApplication(
+        fakerHelper.GetRandomText(5) + "@appsmith.com",
+        "App Viewer",
+      );
+
+      inviteModal.CloseModal();
```
mohanarpit, line 47: @sarojsarab In this test case, can you please help me understand why we are inviting a user to the application? Isn't this test just asserting that the application is forkable & that the fork button is visible?

sarojsarab: @mohanarpit I just changed the test sequence here and fixed the flakiness, so even I am not sure why the original creator of this test case has done this. And also the original purpose of this PR was just to fix the flakiness so did not dig much at that point of time.

mohanarpit: Got it. Thanks for the explanation. Why did we change the sequence of tests though?

sarojsarab: The previous test  was doing a partial import and for this test when it creates a new application in the same workspace it was asking for setting up the datasource which is irrelevant for this test, so changing sequence fixed the spec.

PR #31073, opened 2024-02-13
chore: Upgrade Mockito and surefire plugin

sharat87:
Upgrading Mockito ahead of upgrading Spring itself.

## Summary by CodeRabbit

- **Tests**
	- Updated argument types in test methods for enhanced accuracy.
	- Modified mock setups in tests for improved reliability.

### DIFF
--- a/app/server/appsmith-server/src/test/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImplTest.java
+++ b/app/server/appsmith-server/src/test/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImplTest.java
@@ -165,9 +165,7 @@ class ActionExecutionSolutionCEImplTest {
                 configService,
                 tenantService);
 
-        ObservationRegistry.ObservationConfig mockObservationConfig =
-                Mockito.mock(ObservationRegistry.ObservationConfig.class);
-        Mockito.when(observationRegistry.observationConfig()).thenReturn(mockObservationConfig);
+        Mockito.when(observationRegistry.isNoop()).thenReturn(true);
     }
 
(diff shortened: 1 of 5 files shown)


app/server/appsmith-server/src/test/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImplTest.java
```diff
@@ -165,9 +165,7 @@ public void beforeEach() {
                 configService,
                 tenantService);
 
-        ObservationRegistry.ObservationConfig mockObservationConfig =
-                Mockito.mock(ObservationRegistry.ObservationConfig.class);
-        Mockito.when(observationRegistry.observationConfig()).thenReturn(mockObservationConfig);
+        Mockito.when(observationRegistry.isNoop()).thenReturn(true);
```
nidhi-nair, line 168: Could you please help explain this change?

sharat87: Previously, we're mocking the observation config with a mock object. That means any code that calls `.observationConfig()` would get the mock object. But then, any code that calls `.observationConfig().someMethod()`, is throwing an NPE, since that mock object doesn't have any configured return values.

But since the whole reason we're mocking this is to just disable observation altogather, it's simpler to make `isNoop` return `true`. In which case, all operations in Observation infrastructure will be disabled and no NPEs should show up.

This is essentially a fix for an NPE that was showing up when I was trying to upgrade Spring.

nidhi-nair (approved):

PR #31102, opened 2024-02-13
fix: Add missing policies field in Appsmith AI Default datasource

vivonk:
## Description
Add missing fields in Appsmith AI Default datasource

#### Type of change

- [x] Manual
- [ ] JUnit
- [ ] Jest
- [ ] Cypress

#### Dev activity

#### QA activity:

## Summary by CodeRabbit

- **New Features**
	- Introduced tracking for creation and update times (`createdAt` and `updatedAt`) for Actions and Datasources to enhance data management and auditing capabilities.
	- Enhanced the default Appsmith AI Datasources with necessary fields and default permission groups to ensure seamless operation and access control.
- **Refactor**
	- Consolidated and refined migration scripts to add missing fields and configurations to Appsmith AI Datasources, improving their functionality and reliability.


trishaanand:
@vivonk : This PR seems to be updating an existing migration. Any instance (including prod) which has already run this migration would not run it again. And hence this would be an incomplete fix for all such instances. Only instances which haven't yet run this migration would benefit from this code change. The only way forward is for you to add a new migration. LMK if you disagree.

cc : @manish535

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration045AddDefaultAppsmithAiDatasourceForOrphanActions.java
```diff
@@ -11,6 +11,7 @@
 import org.springframework.data.mongodb.core.query.Criteria;
 import org.springframework.data.mongodb.core.query.Query;
 
+import java.time.Instant;
```
sondermanish, line 14: We would also need to run this migration again @vivonk. For doing this we just need to change the id of the migration. (just add a suffix `-rerun`)
Try to make sure that running this migration again doesn't move the system in bad state.  
that is to make this migration indempotent.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration045AddDefaultAppsmithAiDatasourceForOrphanActions.java
```diff
@@ -77,11 +79,14 @@ public void addDefaultAppsmithAiDatasourceForOrphanActions() {
                         datasource.setName(DEFAULT_APPSMITH_AI_DATASOURCE);
                         datasource.setPluginId(pluginId);
                         datasource.setWorkspaceId(workspaceId);
+                        datasource.setCreatedAt(Instant.now());
```
sondermanish, line 82: on line number 76 let's do `find()` instead of running `findOne()`, that is because users may have already created new `AiDatsources`. You can still store just the datasource in Map for which the name is `DEFAULT_APPSMITH_AI_DATASOURCE`.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration045AddDefaultAppsmithAiDatasourceForOrphanActions.java
```diff
@@ -77,11 +79,14 @@ public void addDefaultAppsmithAiDatasourceForOrphanActions() {
...
+                        datasource.setUpdatedAt(Instant.now());
```
sondermanish, line 83: if you follow the above comment then we would have to add `conditionals`  before setting the UpdatedAt and CreatedAt fields. That is because DS created by users would have these attributes present.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddCreatedAtUpdatedAtInDefaultAppsmithAiDatasource.java
```diff
@@ -0,0 +1,60 @@
...
+
+        Query datasourceQuery = new Query();
+        datasourceQuery.addCriteria(Criteria.where(FieldNameCE.PLUGIN_ID)
+                .is(pluginId)
+                .and(FieldNameCE.CREATED_AT)
+                .exists(false));
+
+        try (Stream<Datasource> datasourceStream = mongoTemplate.stream(datasourceQuery, Datasource.class)) {
+            datasourceStream.forEach(datasource -> {
+                datasource.setCreatedAt(Instant.now());
```
sondermanish, line 52: @vivonk, there are more fixes required in these datasource object
here is a sample object from `internal.db` 

```
_id: 65c9c3dba729be26d52cfa7f
name : "DEFAULT_APPSMITH_AI_DATASOURCE"
pluginId : "65a7454f43c9d03a6fad6784"
workspaceId : "655f4569815189411356483f"
deleted : false
policies: Array (empty)
```
We would require policies for the datasource to be ACL compliant, We would also require the datasource to have invalids

sondermanish (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddCreatedAtUpdatedAtInDefaultAppsmithAiDatasource.java
sondermanish: For policies, we would either need to obtain permission groups and create policy, or inherit it from the `workspace`. for that bit we would require to fetch workspaces in which these datasources exist.

vivonk: @sondermanish what if these fields are not present, won't GAC work on these datasources? Trying to understand in-case user won't face any issue if these fields are not present.

sondermanish: refer to `policyGenerator.getAllChildPolices` to understand how this works, I'm not sure whether this could be used here. however that is a good starting point.

The sure shot way for this is to  

- Find all the datasources via given criteria
- create a map  of `Map<String, Set<Datasource>>` where string would be `workspaceId` & datasource from the corresponding workspace.
- create a list of workspaceIds from the map.keys() -> find all such workspaces.
- for each workspaces, query permissionGroup repo to find `defaultPermissionGroups` from the ids of `workspace.getDefaultPermissionGroups()` 
- now we'll create all five policies for the datasources using these permission groups.
- save each of the datasource
done.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddMissingFieldsInDefaultAppsmithAiDatasource.java
```diff
@@ -0,0 +1,164 @@
...
+        pluginQuery.addCriteria(Criteria.where(PACKAGE_NAME).is(APPSMITH_AI_PLUGIN));
+        Plugin plugin = mongoTemplate.findOne(pluginQuery, Plugin.class);
+        if (plugin == null) {
+            log.error("Appsmith AI plugin not found");
+            return;
+        }
+        String pluginId = plugin.getId();
+
+        Query datasourceQuery = new Query();
+        datasourceQuery.addCriteria(Criteria.where(FieldNameCE.PLUGIN_ID)
```
sondermanish, line 66: let's create queries and criteria like the referred method?
com.appsmith.server.migrations.solutions.DatasourceStorageMigrationSolution.newerCheckForDeletedCriteria;

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddMissingFieldsInDefaultAppsmithAiDatasource.java
```diff
@@ -0,0 +1,164 @@
...
+                appViewerPGId = permissionGroup.getId();
+            } else if (permissionGroup.getName().contains(DEVELOPER)) {
+                devPGId = permissionGroup.getId();
+            } else if (permissionGroup.getName().contains(ADMINISTRATOR)) {
+                adminPGId = permissionGroup.getId();
+            }
+        }
+
+        return Set.of(
+                new Policy(AclPermission.READ_DATASOURCES.getValue(), Set.of(devPGId, adminPGId)),
```
sondermanish, line 143: Since we only have requirement of two types of set one for (dev & admin) another for (dev, admin, and viewer) let's create only two sets? let's only add values to sets which are `nonEmpty()`  it's not ideal to have empty values in policies

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddMissingFieldsInDefaultAppsmithAiDatasource.java
```diff
@@ -0,0 +1,164 @@
...
+
+    private Set<PermissionGroup> fetchDefaultPermissionGroups(String workspaceId) {
+        Workspace workspace =
+                mongoTemplate.findOne(new Query(Criteria.where(FieldNameCE.ID).is(workspaceId)), Workspace.class);
+        if (workspace == null) {
+            log.error("Workspace not found for id {}", workspaceId);
+            return new HashSet<>();
+        }
+        Query permissionGroupQuery = new Query();
+        permissionGroupQuery.addCriteria(Criteria.where(FieldNameCE.ID).in(workspace.getDefaultPermissionGroups()));
```
sondermanish, line 161: workspace.getDefaultPermissionGroups could be empty as well, let's place a check and skip if the default permission group is empty, the workspace won't be usable anyway over here?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddMissingFieldsInDefaultAppsmithAiDatasource.java
vivonk: IMO this is not really needed given that even if someone has archived the documents, we still want to bring the consistency in the document structures.

sondermanish: what i meant was to refer to aa criteria like how it has been used, it was just a reference

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddMissingFieldsInDefaultAppsmithAiDatasource.java
```diff
@@ -0,0 +1,178 @@
+package com.appsmith.server.migrations.db.ce;
+
+import com.appsmith.external.models.Datasource;
+import com.appsmith.external.models.DatasourceConfiguration;
+import com.appsmith.external.models.DatasourceStorage;
+import com.appsmith.external.models.Policy;
+import com.appsmith.external.models.Property;
+import com.appsmith.server.acl.AclPermission;
+import com.appsmith.server.constants.ce.FieldNameCE;
```
sondermanish, line 9: Access `FieldName` instead

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddMissingFieldsInDefaultAppsmithAiDatasource.java
```diff
@@ -0,0 +1,178 @@
...
+        datasourceConfiguration.setProperties(List.of(new Property("Files", new ArrayList<>())));
+        datasourceConfiguration.setUrl("https://ai.appsmith.com");
+        return datasourceConfiguration;
+    }
+
+    /**
+     * Create a datasource storage for the datasource
+     */
+    private void createDatasourceStorage(Datasource datasource) {
+        if (workspaceIdToEnvironmentId == null) {
```
sondermanish, line 109: The idea is to move this from here to main method, let's create the `workspaceIdToEnvironmentId` map in the main execution method only. we would be able to avoid if condition.
Another reason, how this could benefit is that when the streams are running in parallel and if execution order finds this null 2 or more times then the fetching may happen as well.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddMissingFieldsInDefaultAppsmithAiDatasource.java
```diff
@@ -0,0 +1,178 @@
...
+            workspaceIdToEnvironmentId = datasourceStorageMigrationSolution.getDefaultEnvironmentsMap(mongoTemplate);
+        }
+        String envId = datasourceStorageMigrationSolution.getEnvironmentIdForDatasource(
+                workspaceIdToEnvironmentId, datasource.getWorkspaceId());
+
+        DatasourceStorage datasourceStorage = new DatasourceStorage();
```
sondermanish, line 115: Apart from above comment this method looks good

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddMissingFieldsInDefaultAppsmithAiDatasource.java
```diff
@@ -0,0 +1,178 @@
...
+                devAdminPGIds.add(permissionGroupId);
+                allPGIds.add(permissionGroupId);
+            }
+        }
+
+        return Set.of(
+                new Policy(AclPermission.READ_DATASOURCES.getValue(), devAdminPGIds),
+                new Policy(AclPermission.EXECUTE_DATASOURCES.getValue(), allPGIds),
+                new Policy(AclPermission.DELETE_DATASOURCES.getValue(), devAdminPGIds),
+                new Policy(AclPermission.MANAGE_DATASOURCES.getValue(), devAdminPGIds));
```
sondermanish, line 154: there's one more policy that we need, CREATE_DATASOURCE_ACTIONS

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddMissingFieldsInDefaultAppsmithAiDatasource.java
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    private Set<PermissionGroup> fetchDefaultPermissionGroups(String workspaceId) {
+        Query workspaceQuery = new Query(Criteria.where(FieldNameCE.ID).is(workspaceId));
+        workspaceQuery.fields().include(FieldNameCE.DEFAULT_PERMISSION_GROUPS);
+
+        Workspace workspace = mongoTemplate.findOne(workspaceQuery, Workspace.class);
+        if (workspace == null
+                || workspace.getDefaultPermissionGroups() == null
+                || workspace.getDefaultPermissionGroups().isEmpty()) {
```
sondermanish, line 168: let's use `CollectionUtils.isEmpty()`

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddMissingFieldsInDefaultAppsmithAiDatasource.java
```diff
@@ -0,0 +1,178 @@
...
+            log.error("Workspace not found for id {} or no default permission groups", workspaceId);
+            return new HashSet<>();
+        }
+
+        Query permissionGroupQuery = new Query();
+        permissionGroupQuery.fields().include(FieldNameCE.ID, FieldNameCE.NAME);
```
sondermanish, line 174: Nit: let's use `FieldName` instead

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddMissingFieldsInDefaultAppsmithAiDatasource.java
```diff
@@ -0,0 +1,178 @@
...
+        permissionGroupQuery.addCriteria(Criteria.where(FieldNameCE.ID).in(workspace.getDefaultPermissionGroups()));
+        return new HashSet<>(mongoTemplate.find(permissionGroupQuery, PermissionGroup.class));
```
sondermanish, line 176: can `mongoTemplate` return null? Hashset requires a nonNull collection I guess, Please handle this.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddMissingFieldsInDefaultAppsmithAiDatasource.java
```diff
@@ -0,0 +1,178 @@
...
+
+        return Set.of(
+                new Policy(AclPermission.READ_DATASOURCES.getValue(), devAdminPGIds),
+                new Policy(AclPermission.EXECUTE_DATASOURCES.getValue(), allPGIds),
+                new Policy(AclPermission.DELETE_DATASOURCES.getValue(), devAdminPGIds),
+                new Policy(AclPermission.MANAGE_DATASOURCES.getValue(), devAdminPGIds));
+    }
+
+    private Set<PermissionGroup> getDefaultPermissionGroups(String workspaceId) {
+        return workspaceIdToDefaultPermissionGroups.computeIfAbsent(workspaceId, this::fetchDefaultPermissionGroups);
```
sondermanish, line 158: This function is good to go, memoization is cherry on top.  make sure that `fetchDefaultPermissionGroups` does not throw any unchecked exception otherwise whole operation will stop, let's just log as an error and move forward

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddMissingFieldsInDefaultAppsmithAiDatasource.java
vivonk: I checked DB and saw that no datasource save this ACLPermission and in code as well this permission is not being used. Am I missing something?

vivonk: If we have to add, that means all of PGs should be allowed right?

sondermanish: wait a minute, let me check some DBs

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddMissingFieldsInDefaultAppsmithAiDatasource.java
```diff
@@ -0,0 +1,182 @@
...
+                String workspaceId = datasource.getWorkspaceId();
+                String uniqueGitSyncId = workspaceId + "_" + new ObjectId();
+
+                Set<PermissionGroup> defaultPermissionGroups = getDefaultPermissionGroups(workspaceId);
+
+                datasource.setPolicies(createPolicies(defaultPermissionGroups));
+                datasource.setInvalids(new HashSet<>());
+                datasource.setCreatedAt(Instant.now());
+                datasource.setUpdatedAt(Instant.now());
+                datasource.setGitSyncId(uniqueGitSyncId);
```
sondermanish, line 83: Pretty sure that we would get empty gitSyncIds for each datasource since it was not there, do we add gitSyncId anytime apart from datasource creation? @nidhi-nair .
if yes then we should check for nulls otherwise this is good.

vivonk: I will anyway add a check for this

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddMissingFieldsInDefaultAppsmithAiDatasource.java
```diff
@@ -0,0 +1,182 @@
...
+        datasourceStorage.setDatasourceConfiguration(createEmptyDatasourceConfiguration());
+
+        mongoTemplate.insert(datasourceStorage);
+    }
+
+    /**
+     * Create policies for the datasource based on the default permission groups of the workspace
+     */
+    private Set<Policy> createPolicies(Set<PermissionGroup> permissionGroups) {
+        Set<String> devAdminPGIds = new HashSet<>();
```
nidhi-nair, line 133: You will also need to check if the action this datasource is being attached to had granted permissions to other groups and handle accordingly, to cover for [this](https://github.com/appsmithorg/appsmith/blob/275c7f5394d238dbe41bbbf220f678afea8b10e4/app/server/appsmith-server/src/main/java/com/appsmith/server/newactions/base/NewActionServiceCEImpl.java#L1526) piece of code.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddMissingFieldsInDefaultAppsmithAiDatasource.java
nidhi-nair: Any save operation adds gitsyncid if it is missing in datasources. A null check would be recommended. Good catch Manish.

nidhi-nair:
@vivonk quick note, running automation tests on a migration PR is not really helpful, since automation tests always run on a fresh DB that would likely not have any of the entires that would affect our tests. There's a chance that some of our seeded data could be tested but that doesn't seem like the case in this instance. Just something to be prudent about in the future. 😄

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddMissingFieldsInDefaultAppsmithAiDatasource.java
vivonk: @nidhi-nair this won't be the case since we have attached a fresh new data source to existing orphan actions. And now fixing the fields in that particular fresh datasource itself.

nidhi-nair: Yes I understand that part. The scenario I am speaking of is granting execution rights on this datasource for any action that may have been made available to `anonymousUser`

vivonk: Thanks for catching this @nidhi-nair , I have covered this case as well in the migration now, please review the method. - `addPublicNewActionPoliciesToExecuteDatasource`

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddMissingFieldsInDefaultAppsmithAiDatasource.java
```diff
@@ -0,0 +1,266 @@
...
+            });
+        } catch (Exception e) {
+            log.error("Error during processing migration add-created-at-updated-at-default-appsmith-datasource", e);
+        }
+    }
+
+    /**
+     * Fetch list of all application ids for the datasource and check if any policy contains public permission group id
+     */
+    public Pair<Boolean, Set<String>> getAllApplicationIdsOfDatasourceAndCheckIfAnyAppPublic(Datasource datasource) {
```
sondermanish, line 117: nit: this could be private

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddMissingFieldsInDefaultAppsmithAiDatasource.java
```diff
@@ -0,0 +1,266 @@
...
+        newActionsQuery
+                .fields()
+                .include(FieldName.ID, FieldName.APPLICATION_ID, fieldName(QNewAction.newAction.policies));
+
+        Set<String> allApplicationIds = new HashSet<>();
+        AtomicReference<Boolean> isPublic = new AtomicReference<>(false);
+        try (Stream<NewAction> newActionStream = mongoTemplate.stream(newActionsQuery, NewAction.class)) {
+            newActionStream.forEach(newAction -> {
+                if (!StringUtils.hasText(newAction.getApplicationId())) {
+                    return;
```
sondermanish, line 130: if we are returning here only then we might miss on `publicPermissionGroupId`

vivonk: My bad, I thought return continues rest of code further, that's not the case. Updating this piece of code

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddMissingFieldsInDefaultAppsmithAiDatasource.java
```diff
@@ -0,0 +1,266 @@
...
+    private Set<Policy> createPolicies(boolean isPublic, Set<PermissionGroup> permissionGroups) {
+        Set<String> devAdminPGIds = new HashSet<>();
+        Set<String> allPGIds = new HashSet<>();
+
+        for (PermissionGroup permissionGroup : permissionGroups) {
+            String permissionGroupId = permissionGroup.getId();
+            if (permissionGroup.getName().contains(FieldName.VIEWER)) {
+                allPGIds.add(permissionGroupId);
+            } else if (permissionGroup.getName().contains(FieldName.DEVELOPER)) {
+                devAdminPGIds.add(permissionGroupId);
```
sondermanish, line 207: this would include appDeveloper and workspace Developers right?
if we are adding this value in devAndAdminPGIds then we would be having appDeveloper permission in Manage and Delete Datasources as well.

vivonk: Yes, both of them

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration046AddMissingFieldsInDefaultAppsmithAiDatasource.java
```diff
@@ -0,0 +1,266 @@
...
+            log.error("Workspace not found for id {} or no default permission groups", workspaceId);
+            return new HashSet<>();
+        }
+
+        Query permissionGroupQuery = new Query();
+        permissionGroupQuery.fields().include(FieldName.ID, FieldName.NAME);
+        Criteria applicationSpecificPermissionGroupCriteria = new Criteria();
+        applicationSpecificPermissionGroupCriteria.andOperator(
+                Criteria.where(fieldName(QPermissionGroup.permissionGroup.defaultDomainType))
+                        .is("Application"),
```
sondermanish, line 253: let's create variable for this in the same file.

sondermanish (approved):

PR #31104, opened 2024-02-13
chore: Fluent API for update methods

sharat87:
## Summary by CodeRabbit

- **Refactor**
	- Improved update operations across various repositories for better performance and code clarity.
	- Simplified method calls and consolidated criteria and permission settings for more efficient query handling.
- **Chores**
	- Updated method usage to align with new update logic, enhancing code maintainability.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/params/QueryAllParams.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/params/QueryAllParams.java
@@ -4,9 +4,12 @@ import com.appsmith.external.models.BaseDomain;
 import com.appsmith.server.acl.AclPermission;
 import com.appsmith.server.constants.FieldName;
 import com.appsmith.server.repositories.ce.BaseAppsmithRepositoryCEImpl;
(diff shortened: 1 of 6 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/params/QueryAllParams.java
```diff
@@ -56,6 +61,16 @@ public Mono<Long> count() {
         return repo.countExecute(this);
     }
 
+    public Mono<UpdateResult> updateAll(@NonNull UpdateDefinition update) {
+        scope = Scope.ALL;
```
mohanarpit, line 65: Is this variable being used any where?

mohanarpit (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/params/QueryAllParams.java
sharat87: Yes, actually. It's used in the execute call to decide which of `updateMulti` or `updateFirst` to call on MongoDB API.

PR #31136, opened 2024-02-15
chore: Remove `mongoOperations` in `updateFieldByDefaultIdAndBranchName` method

sharat87:
It's doing exactly the same thing, but with a lot of the code duplicated.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/BaseAppsmithRepositoryCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/BaseAppsmithRepositoryCEImpl.java
@@ -216,26 +216,18 @@ public abstract class BaseAppsmithRepositoryCEImpl<T extends BaseDomain> {
             String branchName,
             String branchNamePath,
             AclPermission permission) {
-        return ReactiveSecurityContextHolder.getContext()
-                .map(ctx -> ctx.getAuthentication())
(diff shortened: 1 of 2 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/BaseAppsmithRepositoryCEImpl.java
```diff
@@ -458,31 +449,8 @@ protected Mono<Set<String>> getAnonymousUserPermissionGroups() {
...
-
-        andCriteria.andOperator(criteriaList.toArray(new Criteria[0]));
-
-        query.addCriteria(andCriteria);
-        if (sort != null) {
-            query.with(sort);
-        }
-
-        return mongoOperations.query(this.genericDomain).matching(query).all().map(obj -> obj);
+    public Flux<T> queryAllWithoutPermissions(List<Criteria> criterias, List<String> includeFields) {
```
mohanarpit, line 452: We removed the limit parameter here. Is that intentional? Do we not need it any more?

sharat87: Nope. There's only one place this method is getting called, and one other place in EE code. Both of them pass null and -1 for sort and limit. So clearly not needed.

mohanarpit (approved):

PR #31158, opened 2024-02-15
chore: Refine WDS dimension tokens

### LINKED ISSUE #30890: Fine tune sizing and spacing tokens
Follow-up for #30877

ichik:
Closes #30890 

image

This is not a full refinement, because we set height/block-size explicitly for some of our widgets with styles [such as these](https://github.com/appsmithorg/appsmith/blob/0d16dc0382b8fdd46dba8b5dc65c7cfc7c496c03/app/client/packages/design-system/widgets/src/components/Button/src/styles.module.css#L171). As such, changes to e.g. inner spacing only affect the horizontal/inline size and not the vertical/block one since the block padding doesn't affect an element that has block size set explicitly.
Ideally, we should switch to implicit sizing to fully embrace multi-dimensional sizing and density theming control over the widget dimensions (i.e. control it with padding that consumes the tokens of the inner spacing scale). If that is not feasible I'd like to know why and how we can closely tie together the sizing scale with inner spacing.

### DIFF
--- a/app/client/packages/design-system/theming/src/token/src/tokensConfigs.json
+++ b/app/client/packages/design-system/theming/src/token/src/tokensConfigs.json
@@ -18,7 +18,7 @@
     "userDensityRatio": 1.5
   },
   "innerSpacing": {
-    "V": 1.75,
+    "V": 1.5,
     "R": 2.5,
     "N": 2,
     "stepsUp": 8,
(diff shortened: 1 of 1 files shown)


KelvinOm (commented):
We are waiting for improvements for typography.

ichik:
More things could be improved (i.e., title variants), but I suggest we merge this as a stop-gap since we also don't have a conclusion on widgets' block sizing. I'd rather move on it iteratively.

image
image

KelvinOm (approved):

PR #31165, opened 2024-02-16
chore: Add `updateFirstAndFind` to Fluent repo API

sharat87:
## Summary by CodeRabbit

- **Refactor**
	- Improved efficiency and clarity in server repository methods, including updates and permission group handling.
- **New Features**
	- Enhanced update methods to ensure better handling of different scopes and permissions.


app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/BaseAppsmithRepositoryCEImpl.java
```diff
@@ -286,61 +264,84 @@ public Flux<T> queryAllExecute(QueryAllParams<T> params) {
...
+        } else {
+            return Mono.error(new AppsmithException(AppsmithError.INVALID_PARAMETER, "scope"));
+        }
     }
 
-    private Mono<Set<String>> tryGetPermissionGroups(QueryAllParams<T> params) {
-        return Mono.justOrEmpty(params.getPermissionGroups())
-                .switchIfEmpty(Mono.defer(() -> getCurrentUserPermissionGroupsIfRequired(
-                        Optional.ofNullable(params.getPermission()), params.isIncludeAnonymousUserPermissions())));
+    private Mono<Void> ensurePermissionGroups(QueryAllParams<T> params) {
```
mohanarpit, line 335: Why are we no longer returning the Set of permissionGroups back? Wasn't the return value being used in the calling function?

mohanarpit (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/BaseAppsmithRepositoryCEImpl.java
sharat87: Hey, thanks for asking. We were using the return value, but that's been refactored too. Instead of returning the permission groups, we set it in he `params` object itself. This was needed because we want to use the permission groups to populate the object we get from `find` operation, done just after the `update` operation. This is the `updateFirstAndFind` API this PR is introducing.

mohanarpit: Thanks. Can we add a function comment stating that the params object will be modified in this function? Will ensure clarity for future developers using the function that the input parameter will be modified.

sharat87: I'll rename it to `ensurePermissionGroupsInParams` instead. Might as well. Thanks! 🙂

trishaanand (approved):

PR #31167, opened 2024-02-16
chore: bug: adjust WDS caption text style

### LINKED ISSUE #30896: bug: adjust WDS caption text style

jsartisan:
Fixes #30896

## Summary by CodeRabbit

- **Refactor**
	- Reordered typography variants for improved consistency in design.
- **New Features**
	- Enhanced feature flag checks to directly return `true` for specific flags, improving feature toggling.
- **Style**
	- Adjusted the order of style options in the widget property pane for better usability.

### DIFF
--- a/app/client/packages/design-system/theming/src/token/src/types.ts
+++ b/app/client/packages/design-system/theming/src/token/src/types.ts
@@ -77,8 +77,8 @@ export const FONT_METRICS = {
 // we use "as const" here because we need to iterate by variants
 export const TYPOGRAPHY_VARIANTS = {
   footnote: "footnote",
-  body: "body",
   caption: "caption",
+  body: "body",
   subtitle: "subtitle",
   title: "title",
   heading: "heading",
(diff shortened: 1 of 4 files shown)


jsartisan:
@vadim Note: In the PR, I have just reverted the order caption and body. The reason is in the code, we don't specifically create sizes for body,caption,footnote etc. We just create some sizes and assign them by index.

```ts
return Object.keys(TYPOGRAPHY_VARIANTS).reduce((prev, current, index) => {
    return {
      ...prev,
      [current]: styles[index],
    };
  }, {} as Typography);
```

ichik:
@jsartisan this seems to change the effective body size (and as a result all the widgets having text of that size), can we avoid that?

jsartisan:
Few solutions i can think of:
1. Change default value of fontsize to "caption" instead of "body"
2. Rename "caption" altogether maybe?

@vadim

KelvinOm (approved):

ichik:
@jsartisan
1. Makes our semantic naming meaningless
2. I don't understand how changing the name would change the value.

KelvinOm:
> @jsartisan this seems to change the effective body size (and as a result all the widgets having text of that size), can we avoid that?

@ichik You're right, but everything seems to be correct here. If we want to change the values, we can tweak the config and make the growth slower or something.

jsartisan:
@vadim
> I don't understand how changing the name would change the value.
I was just wondering why people are getting confused with caption is because of its name? maybe they expect "caption" to be smaller than "body". so i thought just changing the name would have cleared their confusion. just a random thought.

PR #31233, opened 2024-02-22
Revert "fix: settings pane overflowing when number of pages increase"

albinAppsmith:
Reverts appsmithorg/appsmith#31180


albinAppsmith:
closing this since it. is not the cause of failures

PR #31283, opened 2024-02-26
chore: Remove mongoTemplate in BaseService

sharat87:
The `BaseService` keeps references to a `ReactiveMongoTemplate`, a `MongoConverter` and a `Scheduler`.

For the first two, they shouldn't be directly used at all in Service classes, that's an abstraction leak. That's what repositories are for.

The third one isn't really used anywhere in `BaseService` at all.

This PR removes all three of these from `BaseService`.

## Summary by CodeRabbit

- **Refactor**
	- Simplified constructor parameters across various service classes by removing dependencies on `Scheduler`, `MongoConverter`, and `ReactiveMongoTemplate`, streamlining initialization.
- **New Features**
	- Added a new method in `Bridge` class to build criteria for MongoDB queries.
- **Bug Fixes**
	- Adjusted `updateExecute` method in `BaseAppsmithRepositoryCEImpl` to handle policy updates correctly and ensure accurate matched count mapping.
- **Tests**
	- Updated unit tests to reflect changes in service class constructors and the removal of certain parameters.
- **Documentation**
	- No visible changes to end-users.


app/server/appsmith-server/src/main/java/com/appsmith/server/services/TenantServiceImpl.java
```diff
@@ -17,22 +15,11 @@ public class TenantServiceImpl extends TenantServiceCEImpl implements TenantServ
     public TenantServiceImpl(
             Scheduler scheduler,
```
sharat87, line 16: We're not deleting `Scheduler` here since it's used in EE and it'll break the build on sync. Ideally, it should be removed here and added in the EE repo, but that doesn't seem to be worth it for the price of breaking build on sync, perhaps later.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ActionCollectionServiceImplTest.java
```diff
@@ -137,7 +134,7 @@ public void setUp() {
                 scheduler,
                 validator,
                 mongoConverter,
-                reactiveMongoTemplate,
+                null,
```
AnaghHegde, line 137: Why are you passing the `null` here?

sharat87: Hey, good question actually! 😅

That's the `mongoTemplate` object that the service doesn't use anymore. I set it to `null` temporarily to check the build and intended to remove the parameter altogather after. Missed that. Fixed in [a new commit](https://github.com/appsmithorg/appsmith/pull/31283/commits/67b8cc60a4504605196758509b320e463bbeb41c). Thanks for spotting!

mohanarpit (approved):

PR #31369, opened 2024-02-28
chore: Remove `mongoTemplate` from custom action repo impl

sharat87:
1. We remove the `mongoTemplate` and replace it with `mongoOperations`, which we get from the base impl class anyway. One less thing to port.

2. We also move the actual _publish_ logic to a separate protected method so it can be reused in the EE repo impl class, which has this logic duplicated.

## Summary by CodeRabbit

- **Refactor**
	- Updated constructors in specific repository classes for streamlined dependency management.
	- Enhanced the publish actions process with improved aggregation operations and a new method for action state management.
	- Transitioned from `mongoTemplate` to `mongoOperations` for certain database operations, optimizing performance.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/CustomNewActionRepositoryImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/CustomNewActionRepositoryImpl.java
@@ -17,6 +17,6 @@ public class CustomNewActionRepositoryImpl extends CustomNewActionRepositoryCEIm
             MongoConverter mongoConverter,
             CacheableRepositoryHelper cacheableRepositoryHelper,
             MongoTemplate mongoTemplate) {
-        super(mongoOperations, mongoConverter, cacheableRepositoryHelper, mongoTemplate);
+        super(mongoOperations, mongoConverter, cacheableRepositoryHelper);
     }
 }
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewActionRepositoryCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewActionRepositoryCEImpl.java
@@ -12,7 +12,6 @@ import com.appsmith.server.repositories.CacheableRepositoryHelper;
 import lombok.extern.slf4j.Slf4j;
 import org.bson.types.ObjectId;
 import org.springframework.data.domain.Sort;
-import org.springframework.data.mongodb.core.MongoTemplate;
 import org.springframework.data.mongodb.core.ReactiveMongoOperations;
 import org.springframework.data.mongodb.core.aggregation.Aggregation;
 import org.springframework.data.mongodb.core.aggregation.AggregationOperation;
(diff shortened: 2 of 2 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewActionRepositoryCEImpl.java
```diff
@@ -515,37 +509,34 @@ public Flux<NewAction> findByDefaultApplicationId(String defaultApplicationId, O
...
-                                        .withValueOf(Fields.field(NewAction.Fields.unpublishedAction))
-                                        .build();
-                                Aggregation combinedAggregation = Aggregation.newAggregation(
-                                        matchAggregation,
-                                        matchAggregationWithPermission,
-                                        wholeProjection,
-                                        addFieldsOperation);
-                                return mongoTemplate.aggregate(combinedAggregation, NewAction.class, NewAction.class);
-                            })
-                            .subscribeOn(Schedulers.boundedElastic());
```
subrata71, line 546: Can the removal of `subscribeOn` have any implication? Can we test this please?

sharat87: Hey @subrata71, great question, thanks for spotting. The `mongoTemplate` object here, is an instance of `MongoTemplate`, as in, non-reactive API. So we had to wrap it up in a `Mono.fromCallable` and subscribe it to a separate thread pool, so that it doesn't block our HTTP request threads.

But the `mongoOperations` object, is an instance of `ReactiveMongoOperations`, which is the reactive API, and so won't be blocking our request threads. So we don't need to subscribe it to a separate thread pool.

Let me know if I'm not making sense, I'll try to rephrase. Thanks!

subrata71: It makes absolute sense now. Thanks for the detailed answer. 👍

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/CustomNewActionRepositoryImpl.java
```diff
@@ -17,6 +17,6 @@ public CustomNewActionRepositoryImpl(
             MongoConverter mongoConverter,
             CacheableRepositoryHelper cacheableRepositoryHelper,
             MongoTemplate mongoTemplate) {
-        super(mongoOperations, mongoConverter, cacheableRepositoryHelper, mongoTemplate);
+        super(mongoOperations, mongoConverter, cacheableRepositoryHelper);
```
subrata71, line 20: Removal of `mongoTemplate` might break the code in the EE repo. Do we have the corresponding EE PR ready already?

sharat87: Indeed, removing `mongoTemplate` _will_ break the build on EE repo. That's why we're not removing it here. The `mongoTemplate` field and constructor param are left alone. We're only removing it from the call to the super class, which is the CE Impl class, where we don't need it anymore.

The use of `mongoTemplate` in this class in EE repo, is duplicated code doing the same thing as in the CE class. That's why the new function `copyUnpublishedActionToPublishedAction` is marked as `protected` and I'll raise another PR on EE repo once this goes there, to remove that duplicated function and just call this one instead.

subrata71 (approved):

PR #31380, opened 2024-02-29
test: Add AssertJ for plugins, for better assertions

sharat87:
PR adds assertj to the plugin modules as well so we can write better assertions there. We're changing just one test, but to illustrate as an example.

With the previous assertion used here, failure messages looked like this:

```
Expected <true>, but got <false>.
```

But with the better assertions introduced in this PR, we should see something like:

```
Expected `Random unexpected string` to end with `Host not allowed.`
```

Which should help us much better in troubleshooting.

## Summary by CodeRabbit

- **Tests**
	- Improved the reliability of a test method related to denying AWS instance metadata access.

### DIFF
--- a/app/server/appsmith-plugins/pom.xml
+++ b/app/server/appsmith-plugins/pom.xml
@@ -5,136 +5,79 @@
     <modelVersion>4.0.0</modelVersion>
 
     <parent>
-
         <groupId>com.appsmith</groupId>
-
         <artifactId>integrated</artifactId>
-
         <version>1.0-SNAPSHOT</version>
-
     </parent>
 
     <artifactId>appsmith-plugins</artifactId>
-
     <version>1.0-SNAPSHOT</version>
-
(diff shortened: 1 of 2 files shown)


app/server/appsmith-plugins/pom.xml
```diff
@@ -5,270 +5,166 @@
...
                 <exclusion>
-
                     <groupId>junit</groupId>
-
                     <artifactId>junit</artifactId>
-
                 </exclusion>
-
             </exclusions>
-
```
nidhi-nair, line 271: Oh did this just disappear automatically on your run of Spotless?

sharat87: No actually. I cleaned it up, and Spotless didn't complain, so I'm counting that as a win. 🙂

nidhi-nair (approved):

Issue #31411, opened 2024-03-01
[Bug]: The filepicker widgets selects multiple files and then removes one of them, causing the final content of all file.data to be abnormal.


gzlinzihong:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

1. When using filepicker widgets to select multiple files and then remove one of them, all file.data contents are abnormal in the end.
![image](https://github.com/appsmithorg/appsmith/assets/56068188/4d85f693-8ce1-4b7f-b786-94e4535f22ea)
![image](https://github.com/appsmithorg/appsmith/assets/56068188/fdb6d995-5440-44a9-8aba-97e2d7abc444)

### Steps To Reproduce

1. Select multiple files
2. Delete one of the files
3. View filePicker1.data

### Environment

Production

### Version

Cloud And self hosted,latest version

Nikhil-Nandagopal:
@gzlinzihong this is a duplicate of #29605 so I'll close it in favor of that

gzlinzihong:
@Nikhil-Nandagopal  And the second point。The MULTIPART_FORM_DATA type in the body of the restapi plugin, when passing text type parameters, including Chinese, Japanese, etc., is garbled

Based on the above two questions, I have solved them on self hosted. I'm not sure if it's possible to raise a PR to see if it works

PR #12240, opened 2022-03-24
fix: Using blobs for multipart data

nidhi-nair:
## Description
From [this reference](https://developer.mozilla.org/en-US/docs/Web/API/FormData/Using_FormData_Objects) for `FormData`, it seems that some sort of conversions are possible when passing values into form data as simply strings. This PR modifies those into blobs to retain native raw format.

We're also handling the proper encoding for data when an unknown content-type is used.

Fixes #12237 

## Type of change
- Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?
- Manually tested
- DP available at test.appsmith.com

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

## Test coverage results :test_tube:
<details><summary>:white_circle: Total coverage has not changed</summary>

    // Code coverage diff between base branch:release and head branch: fix/bin-data-as-blobs 
Status | File | % Stmts | % Branch | % Funcs | % Lines 
 -----|-----|---------|----------|---------|------ 
 :red_circle: | app/client/src/sagas/ActionExecution/PluginActionSaga.ts | 17.96 **(-0.06)** | 2.31 **(0)** | 9.52 **(0)** | 20.58 **(-0.17)**
 :green_circle: | app/client/src/selectors/commentsSelectors.ts | 85.25 **(1.64)** | 64.71 **(2.95)** | 73.33 **(0)** | 90.59 **(2.35)**
 :red_circle: | app/client/src/utils/autocomplete/TernServer.ts | 52.71 **(-0.23)** | 40.83 **(-0.84)** | 36.21 **(0)** | 56.74 **(-0.25)**</details>

PR #31422, opened 2024-03-01
fix: file picker widgets removing files causing abnormal content, resulting in text garbled characters of type restapiplugin MULTIPART-FORM-DATA

gzlinzihong:
Fixes # (issue number)
> [31411](https://github.com/appsmithorg/appsmith/issues/31411)
#### Type of change
- Bug fix 
>
- JUnit
- Cypress
>
>
#### Dev activity
- My code follows the style guidelines of this project
- I have performed a self-review of my own code
- I have commented my code, particularly in hard-to-understand areas
- My changes generate no new warnings
- New and existing unit tests pass locally with my changes

## Summary by CodeRabbit

- **New Features**
	- Enhanced the FilePicker widget to support better file management.

- **Bug Fixes**
	- Improved handling of character encoding in form data to support UTF-8.

- **Tests**
	- Added tests for new FilePicker functionality and data parsing improvements.


Nikhil-Nandagopal:
@gzlinzihong thanks for the PR! We'll get someone to review it

sneha122:
@Parthvi12 Can you please review the cypress tests?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/restApiUtils/helpers/DataUtils.java
```diff
@@ -190,7 +190,7 @@ public String parseFormData(List<Property> bodyFormData, Boolean encodeParamsTog
                         byte[] valueBytesArray = new byte[0];
                         if (StringUtils.hasLength(String.valueOf(property.getValue()))) {
                             valueBytesArray =
-                                    String.valueOf(property.getValue()).getBytes(StandardCharsets.ISO_8859_1);
+                                    String.valueOf(property.getValue()).getBytes(StandardCharsets.UTF_8);
```
sneha122, line 193: @nidhi-nair @sumitsum Do you guys see any problems in extending the character set for multipart form data text type? Looks ok to me. Please let me know if anything else needs to be changed in order to accommodate for UTF-8

sneha122:
Hi @gzlinzihong Thanks a lot for your contribution! 

There are a couple of linting failures on server side. Could you please execute following command in server directory?
`mvn spotless:apply`

On client side as well there is a linting failure in file `cypress/e2e/Regression/ClientSide/Widgets/Filepicker/FilePickerV2_spec.js`, can you please format this file using prettier formatter?

These should fix all of your linting errors and we can merge the PR as quickly as we can

gzlinzihong:
@sneha122 thank you for your feedback.

I have made the necessary amendments to the code on both the server side and the client side as per your suggestion. I've run the `mvn spotless:apply` command on the server side, and I've also formatted the `cypress/e2e/Regression/ClientSide/Widgets/Filepicker/FilePickerV2_spec.js` file with Prettier on the client side. 

All the linting errors should now be corrected. Feel free to let me know if there is anything else that needs my attention. 

Thank you again for your help and patience.

gzlinzihong:
Hi @sneha122 I'm sorry for the confusion, I've realized that I made a mistake in the naming of my branch. If necessary, we can close this pull request and I will rename the branch and create a new pull request. Let me know if you would like me to proceed in this manner

sneha122:
> @sneha122 thank you for your feedback.
> 
> I have made the necessary amendments to the code on both the server side and the client side as per your suggestion. I've run the `mvn spotless:apply` command on the server side, and I've also formatted the `cypress/e2e/Regression/ClientSide/Widgets/Filepicker/FilePickerV2_spec.js` file with Prettier on the client side.
> 
> All the linting errors should now be corrected. Feel free to let me know if there is anything else that needs my attention.
> 
> Thank you again for your help and patience.

Thanks a lot for these changes, the respective checks are running on the PR, once those are done we can check if we are good to go!!

sneha122:
> Hi @sneha122 I'm sorry for the confusion, I've realized that I made a mistake in the naming of my branch. If necessary, we can close this pull request and I will rename the branch and create a new pull request. Let me know if you would like me to proceed in this manner

Hi @gzlinzihong, No worries we can keep this PR request and merge this once all checks pass.

sneha122:
Hi @gzlinzihong, 

Could you please help me understand second part of the issue: `The MULTIPART_FORM_DATA type in the body of the restapi plugin, when passing text type parameters, including Chinese, Japanese, etc., is garbled`? Can you please give me exact steps to reproduce the issue so that it will be easier for me to test the deploy preview?

app/client/cypress/e2e/Regression/ClientSide/Widgets/Filepicker/FilePickerV2_spec.js
```diff
@@ -80,5 +80,61 @@ describe(
...
+        .selectFile("cypress/fixtures/testRemoveFile1.json", {
+          force: true,
+        });
+      cy.get(commonlocators.filePickerUploadButton).click();
+      //eslint-disable-next-line cypress/no-unnecessary-waiting
+      cy.wait(500);
+
+      // Upload a new file
+      cy.get(widgetsPage.filepickerwidgetv2).click();
+      cy.get(".uppy-DashboardContent-addMore").click();
```
Parthvi12, line 113: @gzlinzihong Could you please use locator i.e.` _fileUploadAddMore`
from CommonLocators here?

app/client/cypress/e2e/Regression/ClientSide/Widgets/Filepicker/FilePickerV2_spec.js
```diff
@@ -80,5 +80,61 @@ describe(
...
+      cy.wait(500);
+
+      // Check file data
+      cy.get(".t--widget-textwidget").should(
+        "contain",
+        "data:application/json;base64",
+      );
+
+      cy.get(widgetsPage.filepickerwidgetv2).click();
+      cy.get(".uppy-Dashboard-Item-action--remove").first().click();
```
Parthvi12, line 130: Please avoid using direct locator here as well, instead use it from commonlocatos file.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/restApiUtils/helpers/DataUtils.java
sumitsum: Hi @sneha122 looks like the data format was changed from UTF_8 to ISO_8859_1 explicitly via [this](https://github.com/appsmithorg/appsmith/pull/12240/files#diff-6e6b1b2bdab919a30fa6185105ecb590c8d7ea5361429461dc5535972212f6f6) PR. @nidhi-nair @AmanAgarwal041 can you please help here ?

sumitsum: Hi @sneha122 the relvant change seems to have been added as part of [this](https://github.com/appsmithorg/appsmith/pull/9473) PR. From the PR it seems like adding ISO_8859_1 instead of UTF_8 could be an explicit decision. @nidhi-nair can you please check out the PR once and confirm if it is safe to replace with UTF_8 encoding ? 
Also, in case we go ahead with the replacement, there are two other places in the same file where ISO_8859_1 encoding is used - we should check if they need replacement as well.

sneha122:
@gzlinzihong Thanks a lot for the detailed steps, Tested the PR looks fixes look good to me!

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/restApiUtils/helpers/DataUtils.java
gzlinzihong: Hi @nidhi-nair Could you help confirm this?

gzlinzihong:
Hi @sneha122 I would like to ask if anyone else will review it in the future?

sneha122:
Hi @gzlinzihong Let me get confirmation from Nidhi on the change of ISO to UTF-8, post that we are good to go ahead!

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/restApiUtils/helpers/DataUtils.java
nidhi-nair: Hey folks, apologies for the delay in response. That's a great catch, Sneha! Yes, the change is intentional. The two types of encoding differ from each other in that UTF-8 is a variable length encoding and ISO_8859_1 (Latin) is a fixed length encoding (of 8 bits). As a result UTF-8 has a different representation for special characters outside the first 256 code points. Latin encoding treats each byte as a separate character instead.

The reason our request parsing uses Latin encoding over UTF-8 is because the multipart form request that we use for action execution sends each character across in a single byte encoded format. I have lost a little bit of context on why that is the case, but it is likely because there is no special handling to let the client side library know otherwise. If you notice in [L#470](https://github.com/appsmithorg/appsmith/blob/b0e960818962fda088a65cd1481a71462cdbce77/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImpl.java#L470) at the request parsing stage, we have interpretted the incoming byte stream as Latin encoded, and created a String using that explicit type of encoding. As a result, when we are trying to convert the interpretted String into a bytw stream again inside `DataUtils`, we are using the same format, so we end up at the same kind of byte representation.

If we are making this change, we need to make sure that we use the same format of encoding across the board. Let me know if I can provide more details here. Perhaps what we need to do is to introduce the ability to even clarify the encoding for end users' endpoints.

nidhi-nair: Additionally, could I also request that we add Cypress tests to test for files as well as text bindings with non-ASCII character sets, so we know our solution here will be permanent, and no future change disregards the conclusion we get to at the end of this thread? Please make sure to cover Chinese, Japanese, and Arabic characters at a minimum.

gzlinzihong: > Hey folks, apologies for the delay in response. That's a great catch, Sneha! Yes, the change is intentional. The two types of encoding differ from each other in that UTF-8 is a variable length encoding and ISO_8859_1 (Latin) is a fixed length encoding (of 8 bits). As a result UTF-8 has a different representation for special characters outside the first 256 code points. Latin encoding treats each byte as a separate character instead.
> 
> The reason our request parsing uses Latin encoding over UTF-8 is because the multipart form request that we use for action execution sends each character across in a single byte encoded format. I have lost a little bit of context on why that is the case, but it is likely because there is no special handling to let the client side library know otherwise. If you notice in [L#470](https://github.com/appsmithorg/appsmith/blob/b0e960818962fda088a65cd1481a71462cdbce77/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImpl.java#L470) at the request parsing stage, we have interpretted the incoming byte stream as Latin encoded, and created a String using that explicit type of encoding. As a result, when we are trying to convert the interpretted String into a bytw stream again inside `DataUtils`, we are using the same format, so we end up at the same kind of byte representation.
> 
> If we are making this change, we need to make sure that we use the same format of encoding across the board. Let me know if I can provide more details here. Perhaps what we need to do is to introduce the ability to even clarify the encoding for end users' endpoints.

Hi @nidhi-nair It is a text type rather than a blob type. You can check line [L#453](https://github.com/appsmithorg/appsmith/blob/b0e960818962fda088a65cd1481a71462cdbce77/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImpl.java#L453) where the bytes array has been converted to utf-8. However, when restapi is used later, the text type is converted to ISO_8859_1 again, resulting in garbled characters.

For specific reproduction steps, please check the chat between me and @sneha122 below this PR.

gzlinzihong: In addition, I checked ActionApi.tsx. The body is of FormData type. When appending a string value to FormData, the binary converted by the browser should be utf-8 encoded by default.

nidhi-nair: Whoops, missed that piece of context in the conversation entirely, sorry about that! Sneha or @gzlinzihong can you please help to check if the line you pointed to can be moved to Latin encoding instead? This is to check if we can use a uniform type of encoding throughout. I see that there is actually some interspersing happening between these two encodings already at [L#553](https://github.com/appsmithorg/appsmith/blob/b0e960818962fda088a65cd1481a71462cdbce77/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImpl.java#L553), and it makes me wonder what kind of behaviour we would see if a user was expecting to embed the value from a file into their binding. Please feel free to invalidate the use case itself if that is what your findings show. This is a core part of our execution flow, which is where my paranoia is coming from. :)

gzlinzihong: Hi @nidhi-nair Sorry, I don't understand how this is related to [L#553](https://github.com/appsmithorg/appsmith/blob/b0e960818962fda088a65cd1481a71462cdbce77/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImpl.java#L553) because the blob case is not actually executed.
I don't think changing the UTF-8 encoding to ISO_8859_1 in [L#453](https://github.com/appsmithorg/appsmith/blob/b0e960818962fda088a65cd1481a71462cdbce77/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImpl.java#L453) is a better way. Because this will introduce more problems, the test parameter in the example below is directly a constant, defined in actionConfig. 
image
image

If the UTF-8 encoding is modified to ISO_8859_1 in [L#453](https://github.com/appsmithorg/appsmith/blob/b0e960818962fda088a65cd1481a71462cdbce77/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImpl.java#L453),although the form of the binding variable can be solved smoothly, the garbled problem will not be solved if the variable is a constant string, and because the encoding is modified before the plugin is executed, the impact The scope will design the execution of all plugins

sneha122:
Hi @gzlinzihong, As I can see in your PR, you have resolved following two issues:
1. When using filepicker widgets to select multiple files and then remove one of them, all file.data contents are abnormal in the end.
2. The MULTIPART_FORM_DATA type in the body of the restapi plugin, when passing text type parameters, including Chinese, Japanese, etc., is garbled.

First one is completely resolved with relevant cypress tests and I have tested the changes as well. As for second one, we are still discussing the change from ISO to UTF, While we are still discussing and coming to conclusion on that, it will be great to get at least first fix merged, and we cannot do that with this PR as it contains changes for second one as well. Is it possible for you to create a separate PRs for both issues? That way we can merge changes for first one very quickly and second one we can discuss and take it forward separately. Please let me know what you think about this

gzlinzihong:
> Hi @gzlinzihong, As I can see in your PR, you have resolved following two issues:
> 
> 1. When using filepicker widgets to select multiple files and then remove one of them, all file.data contents are abnormal in the end.
> 2. The MULTIPART_FORM_DATA type in the body of the restapi plugin, when passing text type parameters, including Chinese, Japanese, etc., is garbled.
> 
> First one is completely resolved with relevant cypress tests and I have tested the changes as well. As for second one, we are still discussing the change from ISO to UTF, While we are still discussing and coming to conclusion on that, it will be great to get at least first fix merged, and we cannot do that with this PR as it contains changes for second one as well. Is it possible for you to create a separate PRs for both issues? That way we can merge changes for first one very quickly and second one we can discuss and take it forward separately. Please let me know what you think about this

OK, I'll do it

gzlinzihong:
Hi @sneha122 
I have split it into 2 pr.
thank you for your patience!
[filePickerPR](https://github.com/appsmithorg/appsmith/pull/31646)
[garbled characters](https://github.com/appsmithorg/appsmith/pull/31647)

sneha122:
Thanks a lot @gzlinzihong This makes it super easy for us to merge the fix which is good to go, now we can focus on utf-8 separately, I will check the file picker PR and run necessary checks on it.

sneha122:
Closing this PR as we have created two separate ones, out of which [one](https://github.com/appsmithorg/appsmith-ee/pull/3699) is merged, other one is pending and can be tracked [here](https://github.com/appsmithorg/appsmith/pull/31647)

Issue #31081, opened 2024-02-13
[Task]: Migrate the map chart widget to use the echarts.


sbalaji1192:
### SubTasks

- [x] Rework the UI layer of the map chart widget to use the Echarts
- [x] Make sure that  User experience is unchanged

sbalaji1192:
@somangshu @rajatagrawal CC

PR #31482, opened 2024-03-05
chore: [Map chart widget] Replace fusion charts with Echarts as chart provider

sbalaji1192:
# Description
This pr replaces fusion chart lib with echarts in the map chart widget. It also checkins the necessary maps.

Fixes https://github.com/appsmithorg/appsmith/issues/31081

#### Type of change

- [x] Manual
- [ ] JUnit
- [ ] Jest
- [ ] Cypress
>
>
#### Dev activity

#### QA activity:

## Summary by CodeRabbit

- **New Features**
    - Updated the Widget interface to make link properties optional, enhancing flexibility.
    - Added a `retryPromise` function for improved error handling and retry logic in asynchronous operations.
    - Introduced new mapping data and utilities for the MapChartWidget, enabling detailed country/region information and dynamic map types.

- **Enhancements**
    - Improved error handling in widget callouts to gracefully handle missing links.
    - Enhanced the MapChartWidget with new functionalities including dynamic map data loading, chart resizing, and skeleton UI for loading state.

- **Refactor**
    - Refactored chart configurations and event handling in MapChartWidget for better performance and readability.


somangshu:
@sbalaji1192 QQ on the code, did we have JSON files for fusion chart maps as well, if yes, are we deleting them?

app/client/src/widgets/MapChartWidget/component/index.tsx
```diff
@@ -1,27 +1,11 @@
...
+import React, { useEffect, useMemo, useRef, useState } from "react";
 import styled from "styled-components";
-// Include the react-fusioncharts component
-import ReactFC from "react-fusioncharts";
-// Include the fusioncharts library
-import type { ChartObject } from "fusioncharts";
-import FusionCharts from "fusioncharts";
-
-// Import FusionMaps
-import FusionMaps from "fusioncharts/fusioncharts.maps";
```
somangshu, line 10: This is what I was talking about [here](https://github.com/appsmithorg/appsmith/pull/31482#issuecomment-1978909309), this was a cause of our build size increasing, would be amazing to reduce the size removing any unused assets.

app/client/src/widgets/MapChartWidget/component/index.tsx
```diff
@@ -40,217 +24,132 @@ const MapChartContainer = styled.div<{
...
-  shortLabel: string;
-  value: number;
-}
-
-function MapChartComponent(props: MapChartComponentProps) {
   const { caption, colorRange, data, onDataPointClick, showLabels, type } =
     props;
 
   const fontFamily =
-    props.fontFamily === "System Default" ? "inherit" : props.fontFamily;
```
somangshu, line 76: Whats the point of this condition?

somangshu (approved):
No critical comments on the code. LGTM

ichik:
@somangshu 

> right now the hover state is indistinguishable

I can clearly distinguish resting and hover fills on the video. The similarity between hovers of one fill to resting of another is to an extent unavoidable when sticking to single-hue color harmony.

https://colorbrewer2.org could be used to pick solid defaults.

rajatagrawal:
@sbalaji1192 

Can we consider removing `fusionmaps` package from package.json file since it won't be used anymore ?

rajatagrawal:
Few QA callouts that I had addressed during chart widget migration:

1. Dnd map chart widget.
2. Deselect the legend item with orange box and notice that the orange part of the map gets unhiglighted.
3. Resize the widget.
4. Notice that the map gets highlighted again.
5. Expected behavior: The map should remain unhighlighted.

Nit: The font size for captions in the on the maps is smaller than the previous one.

rajatagrawal:
Nit:

1. The legends are getting clipped when the map chart widget widget is less. They don't get clipped in fusion map chart. Chart Widget makes them scrollable. This solution can be considered or anything that replicates fusion map chart behavior.
2. Antartica shows in yellow color in new version. It as orange in previous version.
3. The map (World with Antartica) is not getting resized when the widget width is very small. It is getting clipped. Fusion chart versions gets resized at all widths.

app/client/src/widgets/MapChartWidget/component/index.tsx
```diff
@@ -40,217 +24,132 @@ const MapChartContainer = styled.div<{
...
-      ...chartConfigs,
-    };
-    newChartConfigs["dataSource"]["colorRange"]["color"] = colorRange;
-    chart.setChartData(newChartConfigs.dataSource, "json");
-  }, [JSON.stringify(colorRange)]);
+    loadMap(type).then(() => {
+      setIsLoading(false);
+      // we are using Math.random to force the chart to set options again
+      // to avoid the race conditions while loading maps
+      setKey(Math.random());
```
rajatagrawal, line 102: Can this useEffect also set options once the map chart data is loaded ? It may help in simplifying the re-rendering logic achieved with `setKey` at the moment.

rajatagrawal: I am not able to fully understand the corner case that is addressed by `setKey`. Can you please explain the corner case ?

rajatagrawal (changes requested):
Requesting some changes based on the comments shared.

rajatagrawal:
Nit:

1. The gray color for Venezuela in South America map isn't matching previous fusion map chart widget.
2. The oceania map isn't matching previous verison. The smaller countries are not visible in the map, though their legends are visible.

sbalaji1192:
@somangshu 

> the onDataPointClick does not let me refer to the selectedDatePoint prop. This was also the case before, but why?
Not Sure why we didn't allow this. Allowed it now.

sbalaji1192:
@rajatagrawal 
Regarding the retaining the unselection when user resizes the widget. Its not possible to do this with echarts. We are using the height and width to come with position values for the map. as a result we need to call setOptions when the widget or height changes, so we will loose the unselection.

>The oceania map isn't matching previous verison. The smaller countries are not visible in the map, though their legends are visible.

The map used by fusion is not accurate , The one we are using is the actual representation of the oceania.

>The legends are getting clipped when the map chart widget widget is less. They don't get clipped in fusion map chart. Chart Widget makes them scrollable. This solution can be considered or anything that replicates fusion map chart behavior.

I discussed this with somangshu, I was not able to find a way to wrap the visual map compoenet a.k.a legends. We decided to leave it for now.

> Antartica shows in yellow color in new version. It as orange in previous version.

This is because of the difference in how fusion and echarts handle the color ranges when there is overlapping value. For example consider the following range

`[{min: 0, max: 1, color: #f00}, {min: 1, max: 2, color: #0f0}]`

as you can see there is an overlap in the range. Now for a region with value 1 fusion paints it in "#0f0" while echarts paints it in "#f00"

sbalaji1192:
@somangshu @rajatagrawal Removed the unused packages.

rajatagrawal:
1. To address the unselection bug, I had compared previous and current chart options. If they were equal, setOption wasn't called, only resize was called. Relevant snippets are [here](https://github.com/appsmithorg/appsmith/blob/e84e27318c79ec7316d96f47bda36358c57a4309/app/client/src/widgets/ChartWidget/component/index.tsx#L199C16-L199C32) and [here](https://github.com/appsmithorg/appsmith/blob/e84e27318c79ec7316d96f47bda36358c57a4309/app/client/src/widgets/ChartWidget/component/index.tsx#L208).

2. Legend can be made scrollable using [this property](https://echarts.apache.org/en/option.html#legend.type). Reference usage in Chart widget [here](https://github.com/appsmithorg/appsmith/blob/e84e27318c79ec7316d96f47bda36358c57a4309/app/client/src/widgets/ChartWidget/component/EChartsConfigurationBuilder.ts#L200).

sbalaji1192:
@rajatagrawal 
1. In map chart , we need to call set options to update the position related values. check [here](https://github.com/appsmithorg/appsmith/pull/31482/files#diff-b44079808b6c3d5218cf7b1ffa48ae278d82b1f1b0848230b39baaa79609147eR126)
2.  We are using the visual map component to represent the visual range , it doesn't have the scroll option , check [here](https://echarts.apache.org/en/option.html#visualMap-piecewise)

somangshu (approved):

jnikhila:
> > @jnikhila @harshilp24 can you check the copy?

> Screenshot 2024-03-05 at 7 57 01 PM

@somangshu - here's the updated copy. However, I have a QQ - What happens if the charts are not properly rendered? Shouldn't we tell them to do some action?

` The Map Widget now uses the ECharts library to plot charts.  Please verify if your charts are properly rendered with the new library.`

In case the space is limited - the shorter version could be:
`The Map Widget now uses the ECharts library to plot charts. Please verify chart rendering.`

sbalaji1192:
@jnikhila for the action part , shall we append
`The Map Chart Widget now uses the ECharts library to plot charts. Please verify chart rendering. Contact support if you there is any issue`

jnikhila:
> @jnikhila for the action part , shall we append `The Map Chart Widget now uses the ECharts library to plot charts. Please verify chart rendering. Contact support if you there is any issue`
 updated:
 
 `The Map Chart Widget now uses the ECharts library to plot charts. Please verify chart rendering and contact support if you face any issues. `
 
@sbalaji1192 - Sure, if we want to raise it with Support but the basic idea is that they have to redo it (similar to the Chart widget). Right? In which case, we should link it to the docs.
@harshilp24 - I don't think we have a guide for this. Also, we will have to update the [Map Chart](https://docs.appsmith.com/reference/widgets/map-chart) docs to reflect this change.

rajatagrawal (approved):

somangshu (approved):
LGTM

PR #31486, opened 2024-03-05
chore: update node version and appropriate git workflows

KelvinOm:
## Description
- update node version and appropriate git workflow
- added the path to webpack cache folder, this should speed up bundle creation about a minute

[Test, build and push Docker Image](https://github.com/appsmithorg/appsmith/actions/runs/8421752151)
[Build Client, Server & Run only Cypress](https://github.com/appsmithorg/appsmith/actions/runs/8421752151)

## Summary by CodeRabbit

- **Chores**
	- Updated actions/cache and actions/setup-node to v4 across various workflows for improved caching and Node.js setup.
	- Modified the `yarn install` command to use `--immutable` flag, enhancing dependency management.
- **Documentation**
	- Updated comments within workflows to include cautionary and important notes, ensuring better clarity.
- **Refactor**
	- Adjusted caching paths and keys for more efficient caching behavior.
	- Changed Node.js installation to version 20.11.1 in Dockerfile, aligning with the latest version for better performance and security.
- **Tests**
	- Modified assertion in `getCurrentLocationSaga` test to check for the presence of a property, improving test accuracy.

### DIFF
--- a/.github/workflows/appsmithctl.yml
+++ b/.github/workflows/appsmithctl.yml
@@ -56,7 +56,7 @@ jobs:
 
       # In case this is second attempt try restoring status of the prior attempt from cache
       - name: Restore the previous run result
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
           path: |
             ~/appsmithctl_run_result
@@ -73,14 +73,14 @@ jobs:
 
       - name: Use Node.js
         if: steps.appsmithctl_run_result.outputs.appsmithctl_run_result != 'success'
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
           node-version-file: app/client/package.json
 
       # Install all the dependencies
       - name: Install dependencies
         if: steps.appsmithctl_run_result.outputs.appsmithctl_run_result != 'success'
-        run: yarn install --frozen-lockfile
+        run: yarn install --immutable
 
       # Run the Jest tests only if the workflow has been invoked in a PR
       - name: Run the jest tests
--- a/.github/workflows/ci-sanity.yml
+++ b/.github/workflows/ci-sanity.yml
@@ -160,7 +160,7 @@ jobs:
 
       - name: Use Node.js
         if: steps.run_result.outputs.run_result != 'success'
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
           node-version-file: app/client/package.json
 
--- a/.github/workflows/ci-test-custom-script.yml
+++ b/.github/workflows/ci-test-custom-script.yml
@@ -85,7 +85,7 @@ jobs:
       # In case this is second attempt try restoring status of the prior attempt from cache
       - name: Restore the previous run result
         id: cache-appsmith
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
           path: |
             ~/run_result
@@ -120,7 +120,7 @@ jobs:
         run: echo ${{ steps.run_result.outputs.run_result }}
 
       - name: Restore the docker image cache
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
           path: cicontainer.tar.gz
           key: docker-image-${{github.run_id}}
@@ -193,18 +193,20 @@ jobs:
 
       - name: Use Node.js
         if: steps.run_result.outputs.run_result != 'success'
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
           node-version-file: app/client/package.json
 
-      # actions/setup-node@v3 doesn’t work properly with Yarn 3
+      # actions/setup-node@v4 doesn’t work properly with Yarn 3
       # when the project lives in a subdirectory: https://github.com/actions/setup-node/issues/488
       # Restoring the cache manually instead
       - name: Restore Yarn cache
         if: steps.run_result.outputs.run_result != 'success'
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
-          path: app/client/.yarn/cache
+          path: |
+            app/client/.yarn/cache
+            app/client/node_modules/.cache/webpack/
           key: v1-yarn3-${{ hashFiles('app/client/yarn.lock') }}
 
       # Install all the dependencies
@@ -408,7 +410,7 @@ jobs:
       # Force store previous run result to cache
       - name: Store the previous run result
         if: failure()
-        uses: actions/cache/save@v3
+        uses: actions/cache/save@v4
         with:
           path: |
             ~/run_result
--- a/.github/workflows/client-lint.yml
+++ b/.github/workflows/client-lint.yml
@@ -75,7 +75,9 @@ jobs:
       # Install all the dependencies
       - name: Install dependencies
         if: steps.run_result.outputs.run_result != 'success'
-        run: yarn install --immutable
+        run: |
+          corepack enable
+          yarn install --immutable
 
         # Run the Lint for client and packages
       - name: Run lint
--- a/.github/workflows/client-lint.yml
+++ b/.github/workflows/client-lint.yml
@@ -39,7 +39,7 @@ jobs:
 
       # In case this is second attempt try restoring status of the prior attempt from cache
       - name: Restore the previous run result
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
           path: |
             ~/run_result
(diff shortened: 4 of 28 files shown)


KelvinOm:
@AmanAgarwal041 Could you please take a look at [this test](https://github.com/appsmithorg/appsmith/actions/runs/8155100090/job/22291053667?pr=31486#step:11:874)? For some reason, it depends on the node version, which in general should not be the case. I'm not sure that we can somehow mock the ErrorMessage. Do we even need this test?

.github/workflows/appsmithctl.yml
```diff
@@ -56,7 +56,7 @@ jobs:
 
       # In case this is second attempt try restoring status of the prior attempt from cache
       - name: Restore the previous run result
-        uses: actions/cache@v3
```
KelvinOm, line 59: Do we really need all these workflows? 🤔

AmanAgarwal041:
> @AmanAgarwal041 Could you please take a look at [this test](https://github.com/appsmithorg/appsmith/actions/runs/8155100090/job/22291053667?pr=31486#step:11:874)? For some reason, it depends on the node version, which in general should not be the case. I'm not sure that we can somehow mock the ErrorMessage. Do we even need this test?

I am not sure on removing the test case. Maybe @albinAppsmith can pitch in to add more details. It should ideally should not be dependent on specific node version though.

albinAppsmith:
> > @AmanAgarwal041 Could you please take a look at [this test](https://github.com/appsmithorg/appsmith/actions/runs/8155100090/job/22291053667?pr=31486#step:11:874)? For some reason, it depends on the node version, which in general should not be the case. I'm not sure that we can somehow mock the ErrorMessage. Do we even need this test?
> 
> I am not sure on removing the test case. Maybe @albinAppsmith can pitch in to add more details. It should ideally should not be dependent on specific node version though.

Hey @AmanAgarwal041 @KelvinOm, I haven't done any changes here. Only changes was done are, after ADS 2.0 migration some tests were failing, I fixed those failures. So I actually cannot comment on the dependency of node version here.

KelvinOm:
> > @AmanAgarwal041 Could you please take a look at [this test](https://github.com/appsmithorg/appsmith/actions/runs/8155100090/job/22291053667?pr=31486#step:11:874)? For some reason, it depends on the node version, which in general should not be the case. I'm not sure that we can somehow mock the ErrorMessage. Do we even need this test?
> 
> I am not sure on removing the test case. Maybe @albinAppsmith can pitch in to add more details. It should ideally should not be dependent on specific node version though.

@AmanAgarwal041 The problem lies only in this [piece of code](https://github.com/appsmithorg/appsmith/pull/23768/files#diff-3f03ee4a85052611b399b3a1ec48f6d75848cab9fb6010e976e9edc2bc8c9822R85-R91). So with node version update, we begin to receive a slightly different response than before. I don't quite understand the purpose of this test, and since it was added by you, I ask for your help.

.github/workflows/appsmithctl.yml
sharat87: Unfortunately, for this workflow, yes. I'm hoping to move `appsmithctl`'s code into RTS or a separate package in `client/packages` and port it to TypeScript, at which point we won't need this, but that's still some ways away. Until then, we'll need something like this. :|

app/client/package.json
```diff
@@ -6,7 +6,7 @@
     "./lib/*": "./lib/*.js"
   },
   "engines": {
-    "node": "^18.17.1",
+    "node": "^20.11.1",
```
sharat87, line 9: This is changing the runtime NodeJS version of RTS as well, right? If yes, can we change the installed NodeJS version in the Docker image to v20 as well please? Or we can do that in a separate PR after this is merged.

This is located at: https://github.com/appsmithorg/appsmith/blob/d7d0eb381c0fada1132588e5a691872cf90ab61d/deploy/docker/base.dockerfile#L47-L48

sharat87 (dismissed):

.github/workflows/appsmithctl.yml
KelvinOm: Thanks for the information. I'm more interested not only in this workflow, but in all the other as well. I know what only some of them do. Maybe we need to document this somehow?

app/client/package.json
KelvinOm: Updated, please take a look

KelvinOm:
[Build Client, Server & Run only Cypress](https://github.com/appsmithorg/appsmith/actions/runs/8169488174)
[Test, build and push Docker Image](https://github.com/appsmithorg/appsmith/actions/runs/8169485716)

.github/workflows/ci-sanity.yml
```diff
@@ -160,7 +160,7 @@ jobs:
 
       - name: Use Node.js
         if: steps.run_result.outputs.run_result != 'success'
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
           node-version-file: app/client/package.json
```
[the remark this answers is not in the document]
riodeuno: @KelvinOm @sharat87 To make sure it is an easier change in the future, shall we do as Rabbit AI suggests?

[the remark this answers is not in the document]
KelvinOm: So we have three such workflows.
Снимок экрана 2024-03-07 в 12 15 34

@sarojsarab can we use node version from package.json here?

cc @riodeuno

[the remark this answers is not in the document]
sarojsarab: @KelvinOm Yes we can use.

deploy/docker/base.dockerfile
```diff
@@ -44,8 +44,8 @@ RUN set -o xtrace \
 # Install NodeJS
 RUN set -o xtrace \
   && mkdir -p /opt/node \
-  && file="$(curl -sS 'https://nodejs.org/dist/latest-v18.x/' | awk -F\" '$2 ~ /linux-'"$(uname -m | sed 's/x86_64/x64/; s/aarch64/arm64/')"'.tar.gz/ {print $2}')" \
-  && curl "https://nodejs.org/dist/latest-v18.x/$file" | tar -xz -C /opt/node --strip-components 1
+  && file="$(curl -sS 'https://nodejs.org/dist/v20.11.1/' | awk -F\" '$2 ~ /linux-'"$(uname -m | sed 's/x86_64/x64/; s/aarch64/arm64/')"'.tar.gz/ {print $2}')" \
+  && curl "https://nodejs.org/dist/v20.11.1/$file" | tar -xz -C /opt/node --strip-components 1
```
sharat87, line 48: Can we use `latest-v20.x` instead of pinning the minor and patch version numbers here please? That aligns with the fact that we're setting NodeJS version in `package.json` to `^20.11.1` and not `20.11.1`? [Docs on the difference](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#dependencies).

.github/workflows/ci-test-custom-script.yml
```diff
@@ -193,18 +193,20 @@ jobs:
 
       - name: Use Node.js
         if: steps.run_result.outputs.run_result != 'success'
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
           node-version-file: app/client/package.json
 
-      # actions/setup-node@v3 doesn’t work properly with Yarn 3
+      # actions/setup-node@v4 doesn’t work properly with Yarn 3
```
sharat87, line 200: May be remove this comment? Or is it still relevant? 🤔

.github/workflows/client-lint.yml
```diff
@@ -75,7 +75,10 @@ jobs:
       # Install all the dependencies
       - name: Install dependencies
         if: steps.run_result.outputs.run_result != 'success'
-        run: yarn install --immutable
+        run: |
+          corepack enable
+          yarn -v
```
sharat87, line 80: Do we need these in the other places we're doing `yarn install --immutable`?

.github/workflows/rts-build.yml
```diff
@@ -114,7 +114,8 @@ jobs:
       - name: Install dependencies
         if: steps.run_result.outputs.run_result != 'success'
         run: |
-          npm install -g yarn
```
sharat87, line 117: Can we keep this please? This is the RTS build workflow which runs on our self-hosted runners, not on GitHub's standard runners. Unfortunately, we have to install `yarn` there manually, we haven't figured out how to get it there automatically yet.

.github/workflows/appsmithctl.yml
sharat87: Ideally, yes, we've attempted this a few times in the past. The thing with documenting is it's very hard to maintain / update it, when we change things. It inevitably becomes stale. Perhaps a comment at the start of the workflow files itself would serve better, since it's right there reminding people to update. 🤔 🤷

sharat87 (changes requested):
Mostly looks good, just one point on rts-build and we should be ready to merge this in. Thanks for taking the pain to do this! ❤️

.github/workflows/ci-test-custom-script.yml
KelvinOm: Relevant :unamused: The whole issue is that we don't have yarn pre-installed.

.github/workflows/rts-build.yml
KelvinOm: I can return it. I used `corepack` here sense this is [recommended way to install yarn](https://yarnpkg.com/getting-started/install). `corepack` should also be available in node [starting from version 14.19](https://nodejs.org/api/corepack.html)

@sharat87 Do you think it won't work? How can I check it?

.github/workflows/client-lint.yml
KelvinOm: Nope. I was just checking that we are using the correct version of yarn when installing dependencies. I'll delete it.

Снимок экрана 2024-03-26 в 13 31 42

.github/workflows/rts-build.yml
sharat87: Oh okay, then we should be good. I didn't know what `corepack` even was. I'll read up. Thanks. That looks good then.

sharat87 (dismissed):

PR #31499, opened 2024-03-05
chore: WDS color refinement

### LINKED ISSUE #31284: chore: WDS color refinements
- decrease `bg` chroma (more noticeable in light mode with warm seeds)
- check if other `bg*` and `bd*` tokens need decrease too
- decrease text and heading lightness (light mode)

ichik:
Closes #31284

1. Decreased overall chroma everywhere except for actual `bgAccent`.
2. Couldn't reproduce the too-light heading text issue in Storybook, `fg*` colors are all appropriately dark. Perhaps, a wrong token is applied in the editor?

Before / After comparisons for warm and cold seeds:
image
image

### DIFF
--- a/app/client/packages/design-system/theming/src/color/src/DarkModeTheme.ts
+++ b/app/client/packages/design-system/theming/src/color/src/DarkModeTheme.ts
@@ -145,10 +145,13 @@ export class DarkModeTheme implements ColorModeTheme {
       color.oklch.c = 0;
     }
 
-    if (!this.seedIsAchromatic) {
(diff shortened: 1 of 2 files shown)


KelvinOm (dismissed):
Subjectively, but It seems that there are still more Chroma than necessary in the dark mode.

ichik:
> Subjectively, but It seems that there are still more Chroma than necessary in the dark mode.

Decreased some more.

KelvinOm (approved):

PR #31521, opened 2024-03-06
chore: Remove unnecessary call to get default tenant ID

sharat87:
## Summary by CodeRabbit

- **Refactor**
	- Enhanced application data retrieval by integrating tenant-specific services, improving user experience in multi-tenant environments.
	- Modified data retrieval method to directly access current user information for improved efficiency.


nidhi-nair (commented):
Repo layer dependency on service layer smells funny, could you please help justify? Feels like this info should be coming from the calling function. I'm afraid this will set the wrong precedent.

sharat87:
Oh freak. Thanks for noticing. I refactored and redid this so many times, I completely missed that. Let me get back. It was supposed to be `tenantRepository`. 🤦

nidhi-nair (approved):

PR #31544, opened 2024-03-06
chore: Remove BaseDomain on Layout

sharat87:
Migration takes ~24s on release, and ~210s (3.5 mins) on prod.

That makes this a slightly long migration, so I'd like reviewers' opinion on if this PR's changes are worth dealing with that migration time.

This change is needed in the Postgres branch. Some of the fields on this, especially the fact that the `id` field has a persistence-specific `@Id` annotation, is throwing Hibernate off in weird ways that I didn't spend much time diving into. Removing `BaseDomain` on `Layout`, was a much simpler/smaller change than that, and it got Hibernate to behave again.

## Summary by CodeRabbit

- **Refactor**
	- Updated the `Layout` class for improved data handling and compatibility.
- **Chores**
	- Removed unused `userPermissions` field from `LayoutDTO`.
	- Adjusted `UpdateLayoutService` to reflect changes in data structure.
	- Executed database migration to clean up legacy fields in layouts.


app/server/appsmith-server/src/test/resources/test_assets/ActionCollectionServiceTest/mockObjects.json
```diff
@@ -20,8 +20,7 @@
             "MainContainer"
           ],
           "allOnPageLoadActionNames": [],
-          "actionsUsedInDynamicBindings": [],
-          "policies": []
```
mohanarpit, line 24: Will this lead to a diff for Git connected apps?

mohanarpit (dismissed):

app/server/appsmith-server/src/test/resources/test_assets/ActionCollectionServiceTest/mockObjects.json
sharat87: Oh man! Yeah it'll likely cause a diff on git repos. 😞

Let me see if I can set fake read-only fields on `Layout` and yet removing `BaseDomain` to get this to behave. Then I can keep `Layout` the same in MongoDB and Postgres for the most part.

mohanarpit (approved):

Issue #31525, opened 2024-03-06
[Task]: Enforcing strong password based on tenant config


deepika-appsmith:
When a user signs up using the form login feature, it is essential to enforce strong password requirements if the tenant configuration for strong passwords is enabled. This includes ensuring that passwords meet specific criteria, such as being at least 8 characters long and containing at least one uppercase letter, one lowercase letter, one number, and one symbol.

Steps:
1. Check the tenant configuration settings to determine if strong password requirements are enabled.
2. If strong password requirements are enabled:
a. Validate the user's password input to ensure it meets the following criteria:
Minimum length of 8 characters.
Contains at least one uppercase letter.
Contains at least one lowercase letter.
Contains at least one number.
Contains at least one symbol (e.g., !@#$%^&*).
b. If the password meets all the criteria, proceed with user sign-up as usual.
c. If strong password requirements are not enabled, throw appropriate error messages.

PR #31605, opened 2024-03-08
feat: Add functionality to enforce strong password for form signups

abhvsn:
**PR to selectively enable strong password requirements.**

Fixes: https://github.com/appsmithorg/appsmith/issues/31525

## Summary by CodeRabbit

- **New Features**
	- Introduced a strong password policy feature that can be enabled or disabled per tenant.
- **Bug Fixes**
	- Added a new error type and code `INSUFFICIENT_PASSWORD_STRENGTH` for handling cases of insufficient password strength, improving user feedback on password requirements.
- **Refactor**
	- Updated password validation logic to enforce stronger password criteria when the strong password policy is enabled.
	- Refactored user creation and password reset flows to consider tenant-specific password policies, enhancing security measures.


app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ValidationUtils.java
```diff
@@ -7,22 +7,52 @@
 import java.util.regex.Pattern;
 
 public final class ValidationUtils {
-    public static final int LOGIN_PASSWORD_MIN_LENGTH = 6;
+    public static final int LOGIN_PASSWORD_MIN_LENGTH = 8;
```
sharat87, line 10: It looks like we're changing behavior for existing users here. Like someone was using a password with 6-chars, they'd suddenly be locked out and can't login unless they do a forgot-password with a new 8+ character password, right? Can we avoid changing the current situation and use 8 as minimum _only_ if strong password enforcement is enabled please?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserServiceCEImpl.java
```diff
@@ -77,9 +78,12 @@
 import java.util.regex.Pattern;
 
 import static com.appsmith.server.acl.AclPermission.MANAGE_USERS;
+import static com.appsmith.server.constants.ce.FieldNameCE.DEFAULT;
+import static com.appsmith.server.constants.ce.FieldNameCE.TENANT;
```
sharat87, line 82: Let's import from the `FieldName` class instead of `FieldNameCE` please.

sharat87 (commented):
Let's also get a DP to test the behavior directly once. Thanks for the quick TAT on this. 🍰

app/server/appsmith-server/src/main/java/com/appsmith/server/exceptions/AppsmithError.java
```diff
@@ -1020,6 +1020,14 @@ public enum AppsmithError {
             "Trigger parameters empty.",
             ErrorType.INTERNAL_ERROR,
             null),
+    INSUFFICIENT_PASSWORD_STRENGTH(
+            400,
+            AppsmithErrorCode.INSUFFICIENT_PASSWORD_STRENGTH.getCode(),
+            "Your password must meet the following criteria: Must contain at least one digit, one lower case letter, one upper case letter, one special character, and no whitespace",
```
deepika-appsmith, line 1026: "Password must be at least 8 characters long and contain at least one uppercase letter, one lowercase letter, one number, one symbol, and no whitespaces."

abhvsn: @deepika-appsmith do we want to include the max length as well in that case?

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ValidationUtils.java
sharat87: I see. Thanks. So it changes the current behaviour for signup/forgot-password, right? Can we mandate 8 _only_ for folks who have enabled strong, and keep the current 6 for the rest please? I'm trying to avoid any change in behaviour as that will just lead to questions to our support asking to make this configurable because some people don't want to change. 🙏

abhvsn: @sharat87 is it really that significant, because if we keep both 6 and 8 as minimum length based on the strong password requirement then this needs to be handled in the client side check as well even before we make a server call. I'm sensing we are overthinking here but let me know your thoughts, I'll update the implementation if needed 🙂

sharat87: But, later when we get to client-side changes, we'll need the client to get access to this setting for the right error message anyway, right? 

Either way, not a strong enough opinion beyond this though. Approving.

sharat87 (approved):

PR #31629, opened 2024-03-09
chore: Get optimization for adding plugin details to actions

sharat87:
The optimization done in adding missing plugin details to a list of actions, was set for only one signature of this function. This PR takes it to the other signature as well.

## Summary by CodeRabbit

- **Refactor**
	- Improved the process for enriching actions with missing plugin details for enhanced performance and reliability.


sharat87:
Converting to draft. This _may be_ makes some tests fail on EE. Verify that before merging.

PR #31640, opened 2024-03-11
chore: Add separate request/response views

sharat87:
This PR gets finer control into what fields are allowed in request-body-only, vs what's allowed in response-body-only. This leaves the fields to separately controlled regarding what can go into the database and what can't.

[Slack thread](https://theappsmith.slack.com/archives/CPQNLFHTN/p1710125307810949).

## Summary by CodeRabbit

- **Refactor**
	- Enhanced data handling in action controllers for improved performance and maintainability.
- **New Features**
	- Introduced interfaces for better management of data views in action responses.
- **Documentation**
	- Added documentation for new interfaces and their application.


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/views/ResponseOnly.java
```diff
@@ -0,0 +1,3 @@
+package com.appsmith.external.views;
+
+public interface ResponseOnly extends Views.Public {}
```
abhvsn, line 3: Let's add a comment why do we need this view even if we have `Views.Public` for API responses.

abhvsn (commented):
@sharat87 I can see other domains e.g. BaseDomain, LayoutDTO, DatasourceDTO, Application etc and some are present in EE repo as well. Are we going to add the RequestOnly annotation in other controller classes as well?

sharat87:
@abhvsn, not as a singular task right now, no. I'll move the APIs that are blocking my efforts on Postgres, but other than that, I'd rather not touch what's not broken. Eventually though, I'd like to completely remove `Public` and just use `RequestOnly` and `ResponseOnly`, but that's not pressing enough to happen as a single big bang. Does that sound okay?

abhvsn (approved):

abhvsn:
> @abhvsn, not as a singular task right now, no. I'll move the APIs that are blocking my efforts on Postgres, but other than that, I'd rather not touch what's not broken. Eventually though, I'd like to completely remove Public and just use RequestOnly and ResponseOnly, but that's not pressing enough to happen as a single big bang. Does that sound okay?

@sharat87 sounds good, let's create a ticket to track this request though. Approved!

mohanarpit (approved):

PR #31653, opened 2024-03-11
Fix: Deselect all widgets on canvas click

riodeuno:
## Description
Clicking on the canvas in Anvil layouts did not deselect all widgets.
This PR fixes it. Now all widgets

## Automation

/ok-to-test tags="@tag.Anvil"

> [!CAUTION]  

### DIFF
--- a/app/client/src/layoutSystems/anvil/canvas/AnvilMainCanvas.tsx
+++ b/app/client/src/layoutSystems/anvil/canvas/AnvilMainCanvas.tsx
@@ -19,8 +19,20 @@ export const AnvilMainCanvas = (props: BaseWidgetProps) => {
   const clickToClearSelections = useClickToClearSelections(props.widgetId);
 
   const handleOnClickCapture = useCallback(
+    // We need to make sure to call this only if we're clicking on the main canvas
+    // To do this, we'll inspect the event.target to make sure it has the className
+    // we would expect only the main canvas to have.
+    // Note: This className also exists in the view mode, but this file is rendered only in edit mode.
     (event) => {
-      if (event.eventPhase === 2) clickToClearSelections(event);
+      // Get the main canvas identifier (layoutId)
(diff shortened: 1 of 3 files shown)


app/client/src/layoutSystems/anvil/canvas/AnvilMainCanvas.tsx
```diff
@@ -19,8 +19,20 @@ export const AnvilMainCanvas = (props: BaseWidgetProps) => {
   const clickToClearSelections = useClickToClearSelections(props.widgetId);
 
   const handleOnClickCapture = useCallback(
+    // We need to make sure to call this only if we're clicking on the main canvas
+    // To do this, we'll inspect the event.target to make sure it has the className
+    // we would expect only the main canvas to have.
+    // Note: This className also exists in the view mode, but this file is rendered only in edit mode.
     (event) => {
-      if (event.eventPhase === 2) clickToClearSelections(event);
```
marks0351, line 23: I think this might be needed, pls check the checkbox and switch widgets usecases. there is an interaction when when we select these widgets and if the toggles were also clicked in the same interaction then their state has to change.

marks0351 (commented):
gentle reminder to also add tests for this interaction coz it has regressed so many times.

app/client/src/layoutSystems/anvil/canvas/AnvilMainCanvas.tsx
riodeuno: The change here makes it even more specific. We call the `clickToClearSelections` only if we've clicked on the MainCanvas. 
But, something is definitely wrong as the tests failed. I'll check it out.

riodeuno: @marks0351 I've resolved the issue with cypress test and added a cypress test. Please review again.

PR #31655, opened 2024-03-11
fix: Zones and Sections elevation token update

riodeuno:
## Description
This PR makes the following changes:
- Adds elevation token for `border-color` zones and sections, based on their elevations
- Updates the `Background` property pane description for zones and sections to make them relevant to what they mean semantically for the user
- Adds a padding to the containers within zones and sections, when they have their backgrounds visible.
- The above padding is not applied when backgrounds are not visible, as described in the hypothesis for asymmetric padding.

### Before:
Screenshot 2024-03-11 at 2 55 47 PM

### After:
Screenshot 2024-03-11 at 2 54 02 PM

## Automation

/ok-to-test tags="@tag.Anvil"

> [!CAUTION]  

## Summary by CodeRabbit

- **New Features**
	- Updated `ContainerComponent` to support conditional styling for enhanced visual separation based on `elevatedBackground` property.
- **Refactor**
	- Renamed configuration sections from "Background" to "General" and updated labels and help texts to better describe the purpose of the `elevatedBackground` property in both Section and Zone widgets.

### DIFF
--- a/app/client/src/widgets/anvil/Container.tsx
+++ b/app/client/src/widgets/anvil/Container.tsx
@@ -27,7 +27,7 @@ const StyledContainerComponent = styled.div<
       background-color: var(--color-bg-elevation-${props.elevation});
       border-radius: var(--border-radius-elevation-${props.elevation})};
       border-color: var(--color-bd-elevation-${props.elevation});
-      border-width: 1px;
+      border-width: var(--border-width-1);
       border-style: solid;
 
       /* Add padding to the container to maintain the visual spacing rhythm */
--- a/app/client/src/widgets/anvil/Container.tsx
+++ b/app/client/src/widgets/anvil/Container.tsx
@@ -19,16 +19,24 @@ const StyledContainerComponent = styled.div<
   outline: none;
   border: none;
   position: relative;
-  background-color: ${(props) =>
-    props.elevatedBackground
-      ? `var(--color-bg-elevation-${props.elevation})`
-      : "inherit"};
-  border-radius: ${(props) =>
-    `var(--border-radius-elevation-${props.elevation})`};
-  padding-block: var(--outer-spacing-0);
-  padding-inline: var(--outer-spacing-0);
+  border-radius: var(--border-radius-1);
+  /* If the elevatedBackground is true, then apply the elevation styles */
+  ${(props) => {
+    if (props.elevatedBackground) {
+      return `
+      background-color: var(--color-bg-elevation-${props.elevation});
+      border-radius: var(--border-radius-elevation-${props.elevation})};
+      border-color: var(--color-bd-elevation-${props.elevation});
+      border-width: 1px;
+      border-style: solid;
 
-  border-width: var(--border-width-1);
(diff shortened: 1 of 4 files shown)


app/client/src/widgets/anvil/Container.tsx
```diff
@@ -19,16 +19,24 @@ const StyledContainerComponent = styled.div<
...
-  padding-inline: var(--outer-spacing-0);
+  border-radius: var(--border-radius-1);
+  /* If the elevatedBackground is true, then apply the elevation styles */
+  ${(props) => {
+    if (props.elevatedBackground) {
+      return `
+      background-color: var(--color-bg-elevation-${props.elevation});
+      border-radius: var(--border-radius-elevation-${props.elevation})};
+      border-color: var(--color-bd-elevation-${props.elevation});
+      border-width: 1px;
```
riodeuno, line 30: @ichik @KelvinOm I saw this in the storybook. Should I use a token here, or use 1px?
I don't see any themeing options that change the border width. So, I was wondering if this will ever be relevant.

KelvinOm: We have to use a `--border-width-1` token here. We need to fix this in Storybook as well.

@ichik Could it be that we also need tokens like `--border-width-elevation-1`?

app/client/src/widgets/anvil/Container.tsx
```diff
@@ -19,16 +19,24 @@ const StyledContainerComponent = styled.div<
...
+      border-style: solid;
 
-  border-width: var(--border-width-1);
+      /* Add padding to the container to maintain the visual spacing rhythm */
+      /* This is based on the hypothesis of asymmetric padding */
+      padding-block: var(--inner-spacing-1);
```
KelvinOm, line 35: In my future PR, I will be changing the behaviuor here.

app/client/src/widgets/anvil/Container.tsx
ichik: Currently, only one width is planned to be used for indicating elevation on zones and sections, although we have two tokens  (`--border-width-1` and `--border-width-2`).

KelvinOm (approved):

app/client/src/widgets/anvil/Container.tsx
riodeuno: @KelvinOm I see that you've approved the PR. Would you like for me to use `--border-width-1` here, or skip it in this PR?

marks0351: Should we get this as a UI component exported from wds package?

ichik: @KelvinOm I don't think border width should change with elevation, so seems unneeded.

KelvinOm: @riodeuno I can fix this. 

@marks0351 Yes, it could be WDS components. We'll leave it as it is for now, I'll make it a component later.

KelvinOm (approved):

PR #31659, opened 2024-03-11
chore: WDS TextInput style

### LINKED ISSUE #31021: chore: refine WDS TextInput style
From:

To (roughly):

Key visual detail for the iteration: background instead of border in resting state.

🔗 [Concept visual style in Figma](https://www.figma.com/file/ySJn0fQnPKvTLK83DUDjj2/TextInput-Refinement?type=design&node-id=1%3A23&mode=design&t=fBuECg7AVe3KCoA4-1). Tokens breakdown doesn't match ours and states breakdown doesn't match web platform. Using existing `subtle` tokens may be sufficient (perhaps with small adjustments to token visuals). Addition of new tokens may be needed too.

ichik:
Closes #31021

Before
https://github.com/appsmithorg/appsmith/assets/80973/3a75dc1d-c340-428c-ab7a-53820471c006

After
https://github.com/appsmithorg/appsmith/assets/80973/70249593-2705-4d67-8408-44ce301cb8ae

## Summary of changes

- **New Features**
	- Updated text input styles to use background colors, improving user interaction cues for hover, focus, and invalid states.

### DIFF
--- a/app/client/packages/design-system/widgets/src/styles/src/text-input.module.css
+++ b/app/client/packages/design-system/widgets/src/styles/src/text-input.module.css
@@ -62,7 +62,7 @@
   */
 
   & [data-field-input][data-focused] {
-    background-color: var(--color-bg-neutral-subtle-active);
+    background-color: transparent;
     box-shadow: none;
   }
 
--- a/app/client/packages/design-system/widgets/src/styles/src/text-input.module.css
+++ b/app/client/packages/design-system/widgets/src/styles/src/text-input.module.css
@@ -46,6 +46,7 @@
   *-----------------------------------------------------------------------------
   */
   & [data-readonly][data-field-input] {
+    background-color: transparent;
     box-shadow: none;
     padding-inline: 0;
   }
--- a/app/client/packages/design-system/widgets/src/styles/src/text-input.module.css
+++ b/app/client/packages/design-system/widgets/src/styles/src/text-input.module.css
@@ -6,7 +6,7 @@
     align-items: center;
     gap: var(--inner-spacing-1);
     border-radius: var(--border-radius-elevation-3);
-    box-shadow: inset 0 0 0 1px var(--color-bd-neutral);
+    background-color: var(--color-bg-neutral-subtle);
     padding-inline: var(--inner-spacing-1);
     flex: 1;
     max-inline-size: 100%;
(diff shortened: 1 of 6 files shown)


app/client/packages/design-system/widgets/src/styles/src/text-input.module.css
```diff
@@ -80,6 +91,10 @@
     box-shadow: 0 0 0 1px var(--color-bd-negative);
   }
 
+  & [data-field-input][data-invalid][data-hovered] {
+    box-shadow: 0 0 0 1px var(--color-bd-negative-hover);
+  }
+
   & [data-field-input][data-invalid][data-focused] {
     box-shadow: 0 0 0 2px var(--color-bd-negative);
```
ichik, line 99: @jsartisan this doesn't seem to work (as in the main focus color supersedes the error, I'm not sure if that's even needed. Having a negative border in resting and hover only should be sufficient in most cases, the only exception I can think of would be runtime validation that runs without losing focus.
Let's maybe just remove this?

cc @tbrizitsky

ichik, line 99: @jsartisan this doesn't seem to work (as in the main focus color supersedes the error, I'm not sure if that's even needed. Having a negative border in resting and hover only should be sufficient in most cases, the only exception I can think of would be runtime validation that runs without losing focus.
Let's maybe just remove this?

cc @tbrizitsky

app/client/packages/design-system/widgets/src/styles/src/text-input.module.css
```diff
@@ -80,6 +91,10 @@
...
   }
```
ichik, line 100: @jsartisan this doesn't seem to work (as in the main focus color supersedes the error, I'm not sure if that's even needed. Having a negative border in resting and hover only should be sufficient in most cases, the only exception I can think of would be runtime validation that runs without losing focus.
Let's maybe just remove this?

cc @tbrizitsky

jsartisan (dismissed):

KelvinOm (changes requested):
Vadim, you're cool! :) 

Few comments:
1. Subjectively, in the dark mode, input border become really contrast(bright).
Снимок экрана 2024-03-15 в 16 42 40

2. It is necessary to remove the background for the readonly mode
Снимок экрана 2024-03-15 в 16 46 16

KelvinOm (approved):

PR #31678, opened 2024-03-12
chore: Remove non-static `or` API in Bridge

sharat87:
Removing the `BridgeQuery.or` API since

1. it reads confusing. Are the passed-in items applied with an `or`, or the current queries applied with the parameters with an `or`. Unlike `and` method, this difference with `or` can change the meaning of the query drastically.
2. it doesn't translate very well into Postgres world. Multiple hoops to manage the same API, just not worth it.

The static `Bridge.or` looks, reads, and works much better. That's the suggested alternative for this.

## Summary by CodeRabbit

- **Refactor**
	- Updated logical operations in query handling: deprecated the use of `or` in favor of more robust handling with `and`.
- **Bug Fixes**
	- Improved filtering logic for unpublished actions by name and page ID, ensuring more accurate search results.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ce/bridge/BridgeQuery.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ce/bridge/BridgeQuery.java
@@ -88,13 +88,16 @@ public final class BridgeQuery<T extends BaseDomain> extends Criteria {
         return this;
     }
 
-    public BridgeQuery<T> or(BridgeQuery... items) {
-        checks.add(new Criteria().orOperator(items));
-        return this;
+    /**
+     * Please use {@code Bridge.or} instead. This API looks and reads very confusing and unintuitive, so is explicitly
+     * disabled.
+     */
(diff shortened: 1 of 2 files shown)


mohanarpit (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ce/bridge/BridgeQuery.java
```diff
@@ -88,13 +88,16 @@ public BridgeQuery<T> isFalse(@NonNull String key) {
...
     }
 
-    public BridgeQuery<T> or(BridgeQuery... items) {
-        checks.add(new Criteria().orOperator(items));
-        return this;
+    /**
+     * Please use {@code Bridge.or} instead. This API looks and reads very confusing and unintuitive, so is explicitly
+     * disabled.
+     */
+    public BridgeQuery<T> or(BridgeQuery<T> ignoredUnused) {
```
mohanarpit, line 95: Instead of deprecating, since you are the only user of this function, should we just remove it from the code base?

PR #31694, opened 2024-03-12
fix: added error handling for updateLayout failure while importing

sondermanish:
## Description
Added error handling for update layout failures while importing the application.

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags="@tag.Git, @tag.ImportExport"

> [!IMPORTANT]  
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8253349451>
> Commit: `6c1a23afd96dba4ddcdcf1b0ca0f9791392784a2`
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8253349451&attempt=6" target="_blank">Click here!</a>

## Summary by CodeRabbit

- **New Features**
	- Improved the application import process to handle layout update errors more gracefully, ensuring the import proceeds without interruption.
- **Tests**
	- Added tests to validate the robustness of the application import feature, specifically when encountering layout update failures.


app/server/appsmith-server/src/main/java/com/appsmith/server/applications/imports/ApplicationImportServiceCEImpl.java
```diff
@@ -485,6 +486,14 @@ public Mono<Application> updateImportableArtifact(Artifact importableArtifact) {
                     return Flux.fromIterable(application.getPages())
                             .map(ApplicationPage::getId)
                             .flatMap(updateLayoutService::updatePageLayoutsByPageId)
+                            .onErrorResume(throwable -> {
```
nidhi-nair, line 489: Let's move this handler to inside the updateLayout call itself so that this does not end up becoming a catch all. There might be legitimate reasons to fail imports as well.

nidhi-nair (changes requested):
Please add a test case, especially so for a critical fix, we do not want to revisit this ever again.

app/server/appsmith-server/src/main/java/com/appsmith/server/applications/imports/ApplicationImportServiceCEImpl.java
sondermanish: apart from NPEs, this block would only fail if we have the issue in the `updatePageLayoutByPageId`.
there might be other places which would want to receive the exception and resolve it by its own logic?

nidhi-nair: Let's find those out as a result of failure and fix. We stand a bigger risk of corrupting users' apps if we don't fail for unknown reasons and data isn't handled well in the product.

sondermanish: Okay, I might have been talking about different thing all together (adding an error handling in updateLayout method itself)! Did you mean that we should handle the error inside that flatMap chain itself?

app/server/appsmith-server/src/test/java/com/appsmith/server/imports/internal/ImportServiceTests.java
```diff
@@ -5191,4 +5191,129 @@ public void exportApplicationByWhen_WhenGitConnectedAndDatasourceRenamed_Queries
                 })
                 .verifyComplete();
     }
+
+    @Test
+    @WithUserDetails(value = "api_user")
+    public void importApplication_withupdateLayoutFailures_Success() {
+        FilePart filePart = createFilePart("test_assets/ImportExportServiceTest/faulty-dsl.json");
```
nidhi-nair, line 5198: Nit: Can we move this file to the same package in the resources directory? Helps organization and loading by local path.

app/server/appsmith-server/src/test/java/com/appsmith/server/imports/internal/ImportServiceTests.java
```diff
@@ -5191,4 +5191,129 @@ public void exportApplicationByWhen_WhenGitConnectedAndDatasourceRenamed_Queries
                 })
                 .verifyComplete();
     }
+
+    @Test
+    @WithUserDetails(value = "api_user")
+    public void importApplication_withupdateLayoutFailures_Success() {
```
nidhi-nair, line 5197: Super tiny nit on camel case usage for method name

nidhi-nair (commented):
Minor points, LGTM otherwise.

nidhi-nair (approved):

PR #31742, opened 2024-03-13
chore: Inject AuthenticationPrincipal in controller method

sharat87:
Instead of using `sessionUserService.getCurrentUser()`, or the `ReactiveSecurityContextHolder.getContext` directly (which we are doing in several places), this injection will let us get the principal directly at controller-level.

Yes, it produces the anonymous user, when there's no session.

Why? Less code. More relying on letting Spring do the right thing for us. :stuck_out_tongue:

Why aren't we making this change across the board everywhere? Sure, eventually. Small PR like this helps me get consensus, be less daunting to review, and most important of all, easy to revert if we notice something going wrong. In a week or two, if we want to do this, we can start rolling it out to more places in code.

## Summary by CodeRabbit

- **Refactor**
	- Improved user profile retrieval process for enhanced efficiency.

> [!IMPORTANT]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8446283232>
> Commit: `8eef3717b3eb9d3fbf5b28c4c1a8eb491f088483`
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8446283232&attempt=2" target="_blank">Click here!</a>

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/UserControllerCE.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/UserControllerCE.java
@@ -24,6 +24,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 import org.springframework.http.codec.multipart.Part;
+import org.springframework.security.core.annotation.AuthenticationPrincipal;
 import org.springframework.web.bind.annotation.DeleteMapping;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
@@ -161,10 +162,8 @@ public class UserControllerCE extends BaseController<UserService, User, String> 
 
     @JsonView(Views.Public.class)
     @GetMapping("/me")
-    public Mono<ResponseDTO<UserProfileDTO>> getUserProfile() {
-        return sessionUserService
-                .getCurrentUser()
-                .flatMap(service::buildUserProfileDTO)
+    public Mono<ResponseDTO<UserProfileDTO>> getUserProfile(@AuthenticationPrincipal User user) {
+        return service.buildUserProfileDTO(user)
                 .map(profile -> new ResponseDTO<>(HttpStatus.OK.value(), profile, null));
     }
 


app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/UserControllerCE.java
```diff
@@ -161,10 +162,8 @@ public Mono<ResponseDTO<Boolean>> resetPasswordAfterForgotPassword(
 
     @JsonView(Views.Public.class)
     @GetMapping("/me")
-    public Mono<ResponseDTO<UserProfileDTO>> getUserProfile() {
-        return sessionUserService
-                .getCurrentUser()
-                .flatMap(service::buildUserProfileDTO)
+    public Mono<ResponseDTO<UserProfileDTO>> getUserProfile(@AuthenticationPrincipal User user) {
```
mohanarpit, line 165: @sharat87 Thanks for introducing this change. But doesn't this mean that we will have to cascade the User object from the controller to all layers instead of fetching the user dynamically where a particular function may require it?

sharat87: One, I won't go ahead and declare this as an improvement, but it's definitely what I'd prefer. Functions accepting things as parameters, as opposed to implicitly figuring out via context, has always made more sense and felt more self-contained. It may make such functions easier to test as well, but that's just hypothesis. For that reason, I don't want us to sit down and change all the places we get current user to do it this way.

Two, this is similar to getting the current `exchange` object, to get values of any headers etc. In a lot of places, we get the `exchange` injected in the controller method, and pass it along in the chain until we need something out of it. But in one/two places, we don't get it from the controller, and instead infer it from the state, like how we commonly do for getting the current user. Here as well, I prefer the explicit nature of controller injection, but sure, whatever gets us there faster, not a strong preference.

Three, in this particular case, we're getting the current user in the controller method itself, just to pass it on to the service method. This is a prime candidate for this refactor, if there's any. 🙂

mohanarpit: Understood. Thanks for the detailed explanation. I agree with you philosophically that explicit parameter passing in functions is better than implicit inference. 

While I understand that we aren't making a big refactor here, we should set this as a principle so that we avoid having 2 ways of doing things in our code base. If we notice the sessionService being used in a function that we interact with, we should refactor it to look like this. Otherwise, we will have confusing coding principles 6 months down the line.

mohanarpit (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/UserControllerCE.java
mohanarpit: Lastly, have we tested this for the anonymousUser? I'm sure we have, but I'm just double-checking.

sharat87: Agreed. Thanks for the approval. I've already tested with anonymous user before raising the PR, and included that note in the PR description as well. 🙂

PR #31766, opened 2024-03-13
chore: Tiny NodeJS script as a linter for field name constants

sharat87:
The point is to prevent unfortunate field name problems like this: https://github.com/appsmithorg/appsmith/pull/31760/files.

This NodeJS script does a very rudimentary analysis on all Java files, with a few regular expressions, and finds anomalies. As such, since it's not very smart, it's quite strict. I intend to make it a little more strict in the coming days, but it's a start.

It's not hooked into any processes/CI yet, but that will also come in next. Since it's not very smart, it actually runs quite fast (.8s on EE).

The script also doesn't exit with a non-zero exit code when it finds a problem. Also will be solved as part of integrating it into CI.

## Summary by CodeRabbit

- **Refactor**
	- Improved string concatenation practices across various domain models for better readability and maintainability.
- **New Features**
	- Introduced a utility method for standardized string concatenation using a dot separator.
- **Chores**
	- Added a script to validate field constants, ensuring consistency and adherence to naming conventions.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Application.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Application.java
@@ -484,14 +484,14 @@ public class Application extends BaseDomain implements Artifact {
 
     public static class Fields extends BaseDomain.Fields {
         public static final String gitApplicationMetadata_gitAuth =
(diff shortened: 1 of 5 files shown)


nidhi-nair (approved):
Thanks for getting to this as fast as you have! ❤️  Infinitely helps with the queasy feeling around this solution! LGTM considering our current level of involvement from the team in this layer.

PR #31855, opened 2024-03-17
chore: Cleanup unused, unneeded code

sharat87:
## Summary by CodeRabbit

- **New Features**
	- Introduced a more flexible search functionality.
- **Refactor**
	- Simplified and clarified archive operations.
- **Documentation**
	- Marked specific methods as deprecated in favor of new implementations.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/BaseRepositoryImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/BaseRepositoryImpl.java
@@ -127,6 +127,8 @@ public class BaseRepositoryImpl<T extends BaseDomain, ID extends Serializable>
 
     @Override
     public Mono<Boolean> archiveById(ID id) {
+        Assert.notNull(id, "The given id must not be null!");
+
         Query query = new Query(getIdCriteria(id));
         query.addCriteria(notDeleted());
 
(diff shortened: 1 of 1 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/BaseRepositoryImpl.java
```diff
@@ -148,19 +127,12 @@ public Mono<T> archive(T entity) {
...
     public Mono<Boolean> archiveById(ID id) {
-        Assert.notNull(id, "The given id must not be null!");
-
-        return ReactiveSecurityContextHolder.getContext()
-                .map(ctx -> ctx.getAuthentication())
-                .map(auth -> auth.getPrincipal())
-                .flatMap(principal -> {
-                    Query query = new Query(getIdCriteria(id));
-                    query.addCriteria(notDeleted());
+        Query query = new Query(getIdCriteria(id));
```
mohanarpit, line 130: Do we want to port the `Assert.notNull` as well? Or is that happening underneath the hood?

sharat87: Good catch. I added Lombok's `@NonNull` annotation on the `id` parameter when I removed it, but then figured let's do fewer things at a time and reverted that. Missed restoring this null-check back when I did that. Thanks! Fixed now.

mohanarpit (approved):

Issue #31877, opened 2024-03-18
[Feature]: Write logs into a custom log-directory


pratapaprasanna:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

Currently, we see appsmith write logs to /appsmith-stacks/* . If there is a flexibility to write logs into a custom log directory.
It will help teams who use log-collectors and log-aggregators.

### Why should this be worked on?

If there is a flexibility to write logs into a custom log directory.
It will help teams who use log-collectors and log-aggregators to write them on custom log directory which can further be exported onto splunk or logstash which makes it easy for teams in monitoring

PR #31926, opened 2024-03-19
feat: add ability to change application logs location

pratapaprasanna:
Fixes [31877](https://github.com/appsmithorg/appsmith/issues/31877) 

```
sudo docker run -it -d --name ap -e APPSMITH_LOG_DIR=/logs --pull always appsmith/appsmith-dp:ce-31926


(base) ➜  appsmith-ee git:(release) ✗ docker exec -it -u root ap bash
root@3d17510be0d7:/opt/appsmith# ls /logs/*
/logs/backend:
backend-3d17510be0d7.log

/logs/editor:
access-3d17510be0d7.log  error-3d17510be0d7.log

/logs/rts:
rts-3d17510be0d7.log
root@3d17510be0d7:/opt/appsmith#

```

## Summary by CodeRabbit

- **New Features**
    - Enhanced flexibility in log management by introducing the ability to configure custom log directories through an environment variable.


sharat87:
@pratapaprasanna, I made a few changes directly myself in the interest of time, since we want this one moving urgently.

1. Renamed env variable to just `APPSMITH_LOG_DIR`. Setting it means it's _custom_ so we can keep it short by not having "custom" in it. 🙂
2. Use the same env variable in the supervisord config files. Otherwise, next month we'll frustrated with "what the hell is the difference between `APPSMITH_CUSTOM_LOG_DIR` and `CUSTOM_LOG_DIR`?".
3. At the end of the `entrypoint.sh`, we create a bunch of subdirectories in the log path. We were doing that in the default path. So I changed that to also use the env variable to reduce duplication.

Let me know if this looks okay and we can merge it in.

PR #31980, opened 2024-03-21
chore: RMVF based changes for DnD, Modals, and Space distribution.

### LINKED ISSUE #31935: [Feature]: Update Visuals of DnD based on RMVF designs
[RMVF Drag and Drop designs](https://www.figma.com/file/k4mtF0nZGOAanbbb7v1FYN/On-canvas-UI-2.0-(RVMF)-draft?type=design&node-id=28-3298&mode=design&t=LJHAnFOvE9z7wgLE-0)

- Update Drop Indicator visuals
- Update cell highlight visuals
- Update Space redistribution handle visulas
- Modal empty canvas visuals to indicate Drag and Drop

marks0351:
Fixes #31935
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  


app/client/src/layoutSystems/anvil/sectionSpaceDistributor/constants.ts
```diff
@@ -1,7 +1,6 @@
 export const SpaceDistributorHandleDimensions = {
-  width: 8,
-  offsetTop: 10,
-  border: 1,
+  width: 6,
+  offsetTop: "var(--border-radius-elevation-2)",
   borderRadius: 4,
   padding: 2,
```
KelvinOm, line 5: Why border-radius token used for offsetTop?

app/client/src/layoutSystems/anvil/canvasArenas/styles.css
```diff
@@ -0,0 +1,37 @@
...
+  --empty-modal-drop-arena-active-bg: #ffe5fe;
+  --empty-modal-drop-arena-bg: #e6efff;
+  --empty-modal-drop-arena-outline-color: #3075ff;
+  --empty-modal-drop-arena-outline-active-color: #ce01d3;
+  --empty-modal-drop-arena-text-color: #001040;
+  --space-distribution-handle-active-bg: #0045c6;
+  --space-distribution-handle-bg: #3075ff;
+}
+
+.detached-widgets-drop-overlay {
```
KelvinOm, line 15: Why do we use global styles? To make an isolated css module, use the file naming format like `styles.module.css`

app/client/src/layoutSystems/anvil/canvasArenas/styles.css
```diff
@@ -0,0 +1,37 @@
...
+  background-color: var(--modal-overlay-color);
+}
+
+.detached-widgets-drop-overlay-content {
+  background-color: var(--modal-overlay-content-color) !important;
```
KelvinOm, line 20: Why we need `important` here?

app/client/src/layoutSystems/anvil/canvasArenas/EmptyModalDropArena.tsx
```diff
@@ -0,0 +1,76 @@
...
+  height: 100% !important;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  color: var(--empty-modal-drop-arena-text-color);
+  position: absolute;
+  inset: 0;
+  font-size: 14px;
+  font-weight: 510;
+  line-height: 19.6px;
```
KelvinOm, line 44: Seems like styles for font can be variable as well

app/client/src/layoutSystems/anvil/canvasArenas/DetachedWidgetsDropArena.tsx
```diff
@@ -0,0 +1,55 @@
+import { Modal, ModalContent } from "@design-system/widgets";
```
KelvinOm, line 1: `import { Popover, PopoverModalContent } from "@design-system/headless";`

You can use a headless popover, a modal component is made on the basis of this. You won't have to override styles through important then.

https://wds.appsmith.com/?path=/docs/design-system-headless-popover--docs

KelvinOm (changes requested):

app/client/src/layoutSystems/anvil/sectionSpaceDistributor/propertyPane/PropertyPaneSectionSpaceDistributor.tsx
```diff
@@ -14,6 +14,7 @@ import type { WidgetLayoutProps } from "layoutSystems/anvil/utils/anvilTypes";
 import { PropertyPaneSpaceDistributionHandle } from "./PropertyPaneSpaceDistributionHandle";
 
 const MockedSection = styled.div`
+  --outer-spacing-4: 4px;
```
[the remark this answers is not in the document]
marks0351: https://theappsmith.slack.com/archives/C025SE88KNE/p1711107498278139?thread_ts=1711025103.560039&cid=C025SE88KNE

app/client/src/layoutSystems/anvil/utils/constants.ts
```diff
@@ -39,3 +40,8 @@ export const widgetHierarchy: Record<string, number> = {
   [anvilWidgets.ZONE_WIDGET]: 3,
   OTHER: 4,
 };
+
+export const AnvilEditorColors = {
```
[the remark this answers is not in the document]
marks0351: makes sense. however there are two needs one where colors are needed in js and the mostly used one is being used in css. ones needed is js are being defined here.

app/client/src/layoutSystems/anvil/sectionSpaceDistributor/constants.ts
marks0351: thanks for pointing it out, this is typo as part of an experimental change and needs to be cleaned up.

app/client/src/layoutSystems/anvil/canvasArenas/styles.css
marks0351: il remove modal and use a popover implementation as u suggested, that should remove !important used in different places.

app/client/src/layoutSystems/anvil/canvasArenas/AnvilCanvasDraggingArena.tsx
```diff
@@ -24,6 +27,7 @@ interface AnvilCanvasDraggingArenaProps {
 export const AnvilCanvasDraggingArena = (
   props: AnvilCanvasDraggingArenaProps,
 ) => {
+  const isPreviewMode = useSelector(previewModeSelector);
```
riodeuno, line 30: I saw this `combinedPreviewMode` selector. Not sure if that is what is needed across the board. But, I've seen it used in some instances. Could you verify if that is what we need?

app/client/src/layoutSystems/anvil/canvasArenas/AnvilCanvasDraggingArena.tsx
```diff
@@ -42,13 +46,22 @@ export const AnvilCanvasDraggingArena = (
...
-    <AnvilHighlightingCanvas
-      anvilDragStates={anvilDragStates}
-      deriveAllHighlightsFn={deriveAllHighlightsFn}
-      layoutId={layoutId}
-      onDrop={onDrop}
-    />
-  );
+  const isMainCanvasDropArena =
+    anvilDragStates.mainCanvasLayoutId === props.layoutId;
+  return !isPreviewMode ? (
```
riodeuno, line 51: Can we stop this from rendering when we try to use the `AnvilCanvasDraggingArena` or even higher?
Having mode based control at a finer level is harder to debug when compared to when it is at a higher level. It is better for rendering performance as well.

app/client/src/layoutSystems/anvil/canvasArenas/AnvilHighlightingCanvas.tsx
```diff
@@ -49,6 +49,8 @@ export function AnvilHighlightingCanvas({
       canvasPadding={0}
       getRelativeScrollingParent={getNearestParentCanvas}
       ref={canvasRef}
+      // increases pixel density of the canvas
+      scaleFactor={2}
```
riodeuno, line 53: Could you elaborate on why we're doing this? I've also seen `transform:scale(1)` which causes issues when using fixed positioning for on canvas UI.

app/client/src/layoutSystems/anvil/canvasArenas/DetachedWidgetsDropArena.tsx
```diff
@@ -0,0 +1,53 @@
...
+    });
+  };
+  return props.anvilDragStates.activateOverlayWidgetDrop ? (
+    <DetachedWidgetsDropArenaWrapper onMouseUp={onMouseUp}>
+      <Popover isOpen modal>
+        <PopoverModalContent
+          contentClassName="detached-widgets-drop-overlay-content"
+          overlayClassName="detached-widgets-drop-overlay"
+        >
+          Drop the Modal here
```
riodeuno, line 48: Let's use a file to store these so that it becomes easier when translating in the future.

app/client/src/layoutSystems/anvil/canvasArenas/EmptyModalDropArena.tsx
```diff
@@ -0,0 +1,76 @@
+import React from "react";
+import { useSelector } from "react-redux";
+import { getDragDetails, getWidgetByID } from "sagas/selectors";
+import styled from "styled-components";
+import { WDSModalWidget } from "widgets/wds/WDSModalWidget";
+import type { DragDetails } from "reducers/uiReducers/dragResizeReducer";
+
+const StyledEmptyModalDropArenaWrapper = styled.div<{ isModalEmpty: boolean }>`
```
riodeuno, line 8: Is there a need for the `isModalEmpty`? From the name of the component it seems like we're dealing with an empty modal in the first place.
Please elaborate.

app/client/src/layoutSystems/anvil/canvasArenas/EmptyModalDropArena.tsx
```diff
@@ -0,0 +1,76 @@
...
+const StyledEmptyModalDropArena = styled.div<{
+  isActive: boolean;
+  isModalEmpty: boolean;
+}>`
+  visibility: ${(props) => (props.isModalEmpty ? "visible" : "hidden")};
+  background-color: ${(props) =>
+    props.isActive
+      ? "var(--empty-modal-drop-arena-active-bg)"
+      : "var(--empty-modal-drop-arena-bg)"};
+  border-radius: 6px;
```
riodeuno, line 26: This radius will have to be parametrised. Please check with the team, if necessary, let's create an issue to do this once confirmed.

app/client/src/layoutSystems/anvil/canvasArenas/EmptyModalDropArena.tsx
```diff
@@ -0,0 +1,76 @@
...
+  layoutId: string;
+}) => {
+  const dragDetails: DragDetails = useSelector(getDragDetails);
+
+  /**
+   * boolean to indicate if the widget is being dragged on this particular canvas.
+   */
+  const isCurrentDraggedCanvas =
+    dragDetails && dragDetails.draggedOn === layoutId;
+  const widget = useSelector(getWidgetByID(canvasId));
```
riodeuno, line 62: Is `canvasId` a widget id?

app/client/src/layoutSystems/anvil/canvasArenas/EmptyModalDropArena.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const isModal = widget?.type === WDSModalWidget.type;
```
riodeuno, line 63: The fact that `EmptyModalDropArena` is rendered should tell us that this is within a modal right? Do we need this check?

app/client/src/layoutSystems/anvil/canvasArenas/hooks/useCanvasDragging.ts
```diff
@@ -395,6 +432,6 @@ export const useCanvasDragging = (
   ]);
 
   return {
-    showCanvas: isDragging,
+    showCanvas: isDragging && !activateOverlayWidgetDrop,
```
riodeuno, line 435: The `activateOverlayWidgetDrop` is being used in quite a few places. How complex would it be if we wanted to use this flag once in the data flow?

app/client/src/layoutSystems/anvil/layoutComponents/ModalLayoutProvider.tsx
```diff
@@ -0,0 +1,23 @@
+import type { BaseWidgetProps } from "widgets/BaseWidgetHOC/withBaseWidgetHOC";
+import { LayoutProvider } from "./LayoutProvider";
+import React from "react";
+import { previewModeSelector } from "selectors/editorSelectors";
+import { useSelector } from "react-redux";
+import { EmptyModalDropArena } from "../canvasArenas/EmptyModalDropArena";
+
+export const ModalLayoutProvider = (props: BaseWidgetProps) => {
+  const isPreviewMode = useSelector(previewModeSelector);
+  const showEmptyModalDropArena =
```
riodeuno, line 10: We use this pattern often, I've created a selector for this in the widget selection prototype. If you'd like, you can create one and I'll re-use the one you create.

app/client/src/layoutSystems/anvil/sectionSpaceDistributor/SpaceDistributionHandle.tsx
```diff
@@ -16,38 +16,48 @@ interface SpaceDistributionNodeProps {
...
 ) => {
   if (ref.current && entries.length) {
     const target = entries[0].target as HTMLElement;
-    const updatedLeft =
-      target.offsetLeft - SpaceDistributorHandleDimensions.width * 0.5;
-    ref.current.style.left = updatedLeft + "px";
+    if (target && target.parentElement) {
+      // making this change to compute offset left coz offsetLeft of the dom api does not provide decimal values
+      // which is causing the handle to jump on clicking it the first time
+      const parentLeft = target.parentElement.getBoundingClientRect().left;
```
riodeuno, line 54: Is the `layoutPositionObserver`'s data not valid for this purpose?
Do you think we should wrap the `getBoundingClientRect` into `requestAnimationFrame`?

app/client/src/layoutSystems/anvil/utils/constants.ts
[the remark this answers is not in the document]
riodeuno: Agreed. The pattern to use here would be to generate css vars as we do in WDS.

app/client/src/selectors/layoutSystemSelectors.ts
```diff
@@ -6,16 +6,7 @@ import { LayoutSystemTypes } from "layoutSystems/types";
...
-    state.ui.applications?.currentApplication?.applicationDetail?.appPositioning
-      ?.type
-  ) {
-    return LayoutSystemTypes[
-      state.ui.applications.currentApplication?.applicationDetail
-        ?.appPositioning?.type
-    ];
-  }
-  return LayoutSystemTypes.FIXED;
+  return state ? LayoutSystemTypes.ANVIL : LayoutSystemTypes.ANVIL;
```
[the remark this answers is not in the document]
riodeuno: Seems to be an enforcement of the Anvil layout system for testing. 
@marks0351 I see that we've updated the `userSaga` as well, was this necessary?
If it was necessary, maybe we need to test this DP after reverting this change to prevent any issues in the release

riodeuno (changes requested):
I have a few points of discussion and request for changes.

riodeuno:
@marks0351 @KelvinOm How can we avoid the use of `!important` in our CSS?

1. Does it need to be avoided?
2. Will inspecting the DOM hierarchy holistically help the situation?

app/client/src/layoutSystems/anvil/canvasArenas/AnvilCanvasDraggingArena.tsx
marks0351: combinedPreviewMode is previewmode + protected mode(git).

app/client/src/layoutSystems/anvil/canvasArenas/AnvilHighlightingCanvas.tsx
marks0351: transform scale is css, the scaling above would increase pixels in the canvas that we are using to provide sharper edges for the shapes drawn on it.

app/client/src/layoutSystems/anvil/canvasArenas/EmptyModalDropArena.tsx
marks0351: Nope the component does not know if the modal is empty yet but I see the confusion, let me fix this to make it more readable.

app/client/src/layoutSystems/anvil/canvasArenas/EmptyModalDropArena.tsx
marks0351: you mean to get the value based on theme?

app/client/src/layoutSystems/anvil/canvasArenas/hooks/useCanvasDragging.ts
marks0351: it is computed once in canvas DnD states and is supplied to other components.

app/client/src/layoutSystems/anvil/sectionSpaceDistributor/SpaceDistributionHandle.tsx
marks0351: when distributing space we don't read positions coz its a rapid operation.
let me check if requestAnimationFrame is needed in this case.

app/client/src/layoutSystems/anvil/utils/constants.ts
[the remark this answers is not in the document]
marks0351: Guys colors are not only used for applying css, we also use them to paint on a canvas. I can move the cell highlighting color to a css variable but will still need the drop indicator color. However I prefer both of them stay in one place coz they are related.

app/client/src/selectors/layoutSystemSelectors.ts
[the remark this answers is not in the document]
marks0351: its just a way to enable anvil.

app/client/src/layoutSystems/anvil/canvasArenas/AnvilCanvasDraggingArena.tsx
marks0351: above this would be the layout components which is built as a class component that only receives renderMode. this is first layer where we can get preview mode state. I plan to move these layers to the widget level, then we will move the check a layer up.

marks0351: I mentioned this problem when Layout components were being implemented.

app/client/src/layoutSystems/anvil/canvasArenas/EmptyModalDropArena.tsx
marks0351: let me clean this up, 
layout components were rendered by canvas widgets in their first implementation so you see canvasId, then we removed using canvas widgets in anvil but did not clean up these props.

app/client/src/layoutSystems/anvil/sectionSpaceDistributor/SpaceDistributionHandle.tsx
```diff
@@ -72,7 +83,7 @@ export const SpaceDistributionHandle = ({
 
   // Calculate the left position of the distribution handle
   const leftPositionOfHandle =
-    left - SpaceDistributorHandleDimensions.width * 0.5;
+    left - (SpaceDistributorHandleDimensions.width + zoneGap) * 0.5;
```
KelvinOm, line 86: What is 0.5?

app/client/src/layoutSystems/anvil/sectionSpaceDistributor/propertyPane/PropertyPaneSectionSpaceDistributor.tsx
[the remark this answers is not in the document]
KelvinOm: I don't understand how the link to the thread should help me understand why we are overriding the token value here. Please help me understand.

app/client/src/layoutSystems/anvil/utils/constants.ts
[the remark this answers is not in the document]
KelvinOm: We can use something like `getPropertyValue` to get value from CSS variable. I still prefer to have one source of truth.

app/client/src/layoutSystems/anvil/canvasArenas/EmptyModalDropArena.tsx
riodeuno: Actually this may be based on the widget radius and an offset. https://www.notion.so/appsmith/On-canvas-UI-polishing-00918bdcdb5b4b37b0b194aeba2a376c?pvs=4#89a73d2a965f4d8a80985f34ad9f2e93

app/client/src/layoutSystems/anvil/editor/canvas/styles/anvilEditorDetachedWidgetsOverlay.css
```diff
@@ -0,0 +1,25 @@
+.detached-widgets-drop-overlay {
+  background-color: var(--modal-overlay-color);
+  position: fixed;
+  overflow: hidden;
```
KelvinOm, line 4: I think you don't need these styles 

```
  overflow: hidden;
  inset: 0px;
  ```

File: app/client/src/layoutSystems/anvil/editor/canvas/styles/anvilEditor.css [hunk 22]
```diff
@@ -0,0 +1,7 @@
+/* 
+This file is used to declare styles of the Anvil Editor only Components.
+Please note that these styles will not be applied/available for the published app(AnvilViewerCanvas). 
+*/
+@import "./anvilEditorInteractions.css";
```
KelvinOm, line 5: I still don't understand why we use global styles. Use CSS modules or styled components.

KelvinOm (changes requested):

app/client/src/layoutSystems/anvil/sectionSpaceDistributor/SpaceDistributionHandle.tsx
marks0351: it denotes that we are dividing it by 2.

app/client/src/layoutSystems/anvil/sectionSpaceDistributor/propertyPane/PropertyPaneSectionSpaceDistributor.tsx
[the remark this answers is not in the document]
marks0351: I was trying to fix the issue that was introduced by the alignement pr, where --outer-spacing-4 token is not available for the property pane.

app/client/src/layoutSystems/anvil/sectionSpaceDistributor/SpaceDistributionHandle.tsx
KelvinOm: Why two? Can we create a variable with a semantic name?

app/client/src/layoutSystems/anvil/sectionSpaceDistributor/propertyPane/PropertyPaneSectionSpaceDistributor.tsx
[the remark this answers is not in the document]
KelvinOm: We can just tweak `convertFlexGrowToFlexBasis` function or do another one. We don't need all this for property pane, right?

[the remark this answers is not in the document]
marks0351: yes that is correct, those are the two alteranatives and I think this way the implemenation is singular and easy to debug. is there any reason not to declare the token?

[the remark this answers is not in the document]
KelvinOm: We don't use tokens in property pane they are simply not there and it's confusing. Now you are just trying to make the function suitable for the needs of the property pane, this is fundamentally wrong. The golden principle of single responsibility tells us that we should do two different functions. This is important because in the future, if we want to make changes to the calculation of space redistribution, this will not lead to regression.

Does it make sense for you?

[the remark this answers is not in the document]
marks0351: got it. makes sense.

app/client/src/layoutSystems/anvil/index.ts
```diff
@@ -13,7 +13,12 @@ import { AnvilCanvas } from "./canvas/AnvilCanvas";
  *
  */
 const getAnvilSystemPropsEnhancer = (props: BaseWidgetProps) => {
-  return props;
+  return {
+    ...props,
+    allowWidgetInteraction: !(
```
marks0351, line 18: @KelvinOm this prop will be available for all anvil wds widgets to implement interaction.

KelvinOm (dismissed):

riodeuno (approved):

PR #32073, opened 2024-03-26
fix: logic for fetching onPageLoadActions for blocks import

AnaghHegde:
## Description
Fetch the actions from database and populate the ids for onPageLoad actions for the building block. 

## Automation

/ok-to-test tags="tags.ImportExport"

> [!CAUTION]  

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/imports/internal/partial/PartialImportServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/imports/internal/partial/PartialImportServiceCEImpl.java
@@ -4,6 +4,7 @@ import com.appsmith.external.constants.AnalyticsEvents;
 import com.appsmith.external.models.CreatorContextType;
 import com.appsmith.external.models.Datasource;
 import com.appsmith.server.acl.AclPermission;
+import com.appsmith.server.actioncollections.base.ActionCollectionService;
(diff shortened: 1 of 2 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/imports/internal/partial/PartialImportServiceCEImpl.java
```diff
@@ -426,41 +436,61 @@ public Mono<BuildingBlockResponseDTO> importBuildingBlock(BuildingBlockDTO build
...
-                                                        // Filter the onPageLoadActions based on the json file
-                                                        buildingBlockResponseDTO
-                                                                .getOnPageLoadActions()
-                                                                .addAll(dslExecutableDTOS.stream()
-                                                                        .filter(dslExecutableDTO ->
-                                                                                newOnPageLoadActionNames.contains(
-                                                                                        dslExecutableDTO.getName()))
-                                                                        .toList());
+                        // Fetch all actions and action collections and update the onPageLoadActions with correct ids
+                        return actionCollectionService
```
sondermanish, line 447: Implementation is functionally fine.
Nit: if we collect map for action and action collection and for `buildingBlock's pageLoadAction` then we can loop on `newOnPageLoadAction` itself only once. you may choose to do this later as well

sondermanish (approved):
Approving with the nit, we would tackle the changes in a separate PR

PR #32089, opened 2024-03-27
ci: Allow special chars in PR body

sharat87:
Currently, we get the value of the PR body into a variable, by injecting it into a variable definition. This has two problems:

1. If the PR body has a `'` character in it, then the string definition will terminate there, and cause an syntax error in the rest of the script.
2. This is prone to [script injections as documented by GitHub](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections).

## Summary by CodeRabbit

- **Chores**
	- Enhanced PR automation workflow with better tag extraction, improved messaging for tag usage, and refined notifications.


nidhi-nair (approved):
Thanks for picking this up!

PR #32104, opened 2024-03-27
[docs]: add NET_BIND_SERVICE to security context for OpenShift instal…

pratapaprasanna:


sharat87 (changes requested):
Minor change, otherwise it looks like `capabilities` may be a child of `runAsUser`, which is confusing.

PR #32155, opened 2024-03-28
fix: Limit email length as per RFC 5321

sharat87:
Adds a length check on validating email addresses, as per [RFC 5321](https://datatracker.ietf.org/doc/html/rfc5321#section-4.5.3.1), quoting here:

> [4.5.3.1.1](https://datatracker.ietf.org/doc/html/rfc5321#section-4.5.3.1.1).  Local-part
> 
>    The maximum total length of a user name or other local-part is 64
>    octets.
> 
> [4.5.3.1.2](https://datatracker.ietf.org/doc/html/rfc5321#section-4.5.3.1.2).  Domain
> 
>    The maximum total length of a domain name or number is 255 octets.

## Summary by CodeRabbit

- **New Features**
	- Enhanced email validation based on RFC 5321 standards, ensuring email addresses meet specific length and format requirements before processing.

- **Improvements**
	- Simplified email validation logic by removing redundant null checks.

- **Tests**
	- Added tests for inviting users with invalid email addresses, focusing on excessively long local or domain parts.

> [!IMPORTANT]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8464010263>
> Commit: `8af5c9e5e2a5a9862afd8b3293a0eea3106fd8b9`
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8464010263&attempt=1" target="_blank">Click here!</a>

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ValidationUtils.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ValidationUtils.java
@@ -22,6 +22,16 @@ public final class ValidationUtils {
             SPECIAL_CHARACTERS, LOGIN_PASSWORD_MIN_LENGTH, LOGIN_PASSWORD_MAX_LENGTH);
 
     public static boolean validateEmail(String emailStr) {
+        // Limits defined by RFC 5321 at https://datatracker.ietf.org/doc/html/rfc5321#section-4.5.3.1.
+        if (!StringUtils.hasLength(emailStr)) {
+            return false;
+        }
+
+        final String[] parts = emailStr.split("@");
+        if (parts.length != 2 || parts[0].length() > 64 || parts[1].length() > 255) {
+            return false;
+        }
+
         return EmailValidator.getInstance().isValid(emailStr);
     }
 
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/UserAndAccessManagementServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/UserAndAccessManagementServiceCEImpl.java
@@ -9,6 +9,7 @@ import com.appsmith.server.domains.Workspace;
 import com.appsmith.server.dtos.InviteUsersDTO;
 import com.appsmith.server.exceptions.AppsmithError;
 import com.appsmith.server.exceptions.AppsmithException;
+import com.appsmith.server.helpers.ValidationUtils;
 import com.appsmith.server.repositories.UserRepository;
 import com.appsmith.server.services.AnalyticsService;
 import com.appsmith.server.services.CaptchaService;
@@ -116,6 +117,9 @@ public class UserAndAccessManagementServiceCEImpl implements UserAndAccessManage
 
         Set<String> usernames = new HashSet<>();
         for (String username : originalUsernames) {
+            if (!ValidationUtils.validateEmail(username)) {
+                return Mono.error(new AppsmithException(AppsmithError.INVALID_PARAMETER, FieldName.USERNAMES));
+            }
             usernames.add(username.toLowerCase());
         }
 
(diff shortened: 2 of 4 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ValidationUtils.java
```diff
@@ -22,6 +22,16 @@ public final class ValidationUtils {
...
     public static boolean validateEmail(String emailStr) {
+        // Limits defined by RFC 5321 at https://datatracker.ietf.org/doc/html/rfc5321#section-4.5.3.1.
+        if (!StringUtils.hasLength(emailStr)) {
+            return false;
+        }
+
+        final String[] parts = emailStr.split("@");
+        if (parts.length != 2 || parts[0].length() > 64 || parts[1].length() > 255) {
+            return false;
+        }
```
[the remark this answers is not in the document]
sharat87: 1. The hard-coded numbers are taken directly from the linked RFC.
2. The numbers aren't re-used anywhere other than just this.
3. Hard-coded numbers are actually readable here, since the `>` conditions are immediately apparent.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/UserAndAccessManagementServiceCEImpl.java
```diff
@@ -116,6 +117,9 @@ public Mono<List<User>> inviteUsers(InviteUsersDTO inviteUsersDTO, String origin
 
         Set<String> usernames = new HashSet<>();
         for (String username : originalUsernames) {
+            if (!ValidationUtils.validateEmail(username)) {
```
mohanarpit, line 120: What would happen here for GitHub usernames? Any impact that we should be aware of?

mohanarpit (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/UserAndAccessManagementServiceCEImpl.java
sharat87: Invalid emails aren't accepted in the invite popup for a long time now. Inviting can _only_ be done with emails.

shot-2024-03-28-07-39-52

trishaanand:
@sharat87 : This looks ok from CE perspective. But there are a couple of flows (UserGroup invite, Role invite, and any new flows coming in in the future). Would it make more sense to move the email validation to `UserService.userCreate` flow instead? That ways we don't have to handle it at invite level but one common change would handle it all?

sharat87:
@trishaanand, thanks for sharing that. I'd rather add the validation separately there _also_, so that for the invite flow, we fail fast and fail especially.

For example, in the `inviteUsers` method, there [quite a bit of code happening](https://github.com/appsmithorg/appsmith/blob/8af5c9e5e2a5a9862afd8b3293a0eea3106fd8b9/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/UserAndAccessManagementServiceCEImpl.java#L176) before we call `.userCreate`.

But I agree in principle though. I'll add this validation call into `userCreate` as well. I'd rather we over-validate than under. Does that sound okay?

trishaanand:
> @trishaanand, thanks for sharing that. I'd rather add the validation separately there _also_, so that for the invite flow, we fail fast and fail especially.
> 
> For example, in the `inviteUsers` method, there [quite a bit of code happening](https://github.com/appsmithorg/appsmith/blob/8af5c9e5e2a5a9862afd8b3293a0eea3106fd8b9/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/UserAndAccessManagementServiceCEImpl.java#L176) before we call `.userCreate`.
> 
> But I agree in principle though. I'll add this validation call into `userCreate` as well. I'd rather we over-validate than under. Does that sound okay?

Sounds good to me!

PR #32208, opened 2024-03-28
chore: Remove varargs and List signatures for `.criteria()`

sharat87:
These signatures don't translate very to the Postgres branch.

Server and Cypress tests verified on EE.

## Summary by CodeRabbit

- **Refactor**
    - Improved code readability and maintainability by adding static final string declarations for `defaultResources` fields.
    - Enhanced the `ActionCollectionCE` class by referencing `DefaultResources` for `defaultResources_collectionId`.


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/BranchAwareDomain.java
```diff
@@ -21,5 +23,12 @@ public void sanitiseToExportDBObject() {
         super.sanitiseToExportDBObject();
     }
 
-    public static class Fields extends BaseDomain.Fields {}
+    public static class Fields extends BaseDomain.Fields {
+        public static final String defaultResources_applicationId =
+                dotted(defaultResources, DefaultResources.Fields.applicationId);
+        public static final String defaultResources_collectionId =
+                dotted(defaultResources, DefaultResources.Fields.collectionId);
```
nidhi-nair, line 30: CollectionId is not supposed to be part of the branch aware domain, it's specific to collections only.

sharat87: @nidhi-nair, sorry, don't understand. `BranchAwareDomain` has a `defaultResources` field, of type `DefaultResources`. And `DefaultResources` has a field `collectionId` defined in `DefaultResourcesCE`, as a `String`.

So `BranchAwareDomain` can have a field constant for `defaultResources_collectionId`, right? 🤔

nidhi-nair: That's something of a debt for us. The same type has been used to define every possible path that could exist in a default resources object no matter where it exists.

At the domain level all such objects are only expected to have appId, branchName and it's own id (actionId for actions, collectionId for collections and so on). At the DTO level, the default resources are expected to have pageId / moduleId and collectionId.

There might be a few other fields I'm missing in this explanation here but that's the general gist.

sharat87: Sure... don't disagree, but this is just a constant for a field that exists. This field name was used in a query already, and I just moved the constant over here. Can we not include that debt as part of the scope for this PR?

nidhi-nair: That reference seems to belong in the category of "id of a domain" so this constant does not belong in BranchAwareDomain still. Can you pull it down to ActionCollection please? I'm asking for this because our code base is already riddled with usage today where the wrong property is filled up in the wrong spot just because it was available to fill.

nidhi-nair (approved):

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/BranchAwareDomain.java
sharat87: Sure, moving it over, but I don't think this is a good idea.

1. It breaks consistency in how these field constants are defined today. That is, only fields declared in the outer class, get to show up in the inner `Fields` class. The `defaultResources` isn't defined in `ActionCollectionCE`, but the field name constant is still there.
2. If the `collectionId` field shouldn't be there, why isn't it marked deprecated? The field is the problem, no point adjusting the field name constant.
3. If every class extending `BranchAwareDomain` gets a `defaultResources.collectionId`, then every such class should also have a `Fields.defaultResources_collectionId`.
4. This is just a constant variable, folks can still use `BranchAwareDomain.Fields.defaultResources` and `DefaultResources.Fields.collectionId` directly and this doesn't stop that.

Either way, I've moved it into `ActionCollectionCE` now to get the rest of the changes in this PR going. If we get to repay that debt, perhaps this will go away. Thanks!

nidhi-nair: Thanks for accommodating. The third point is not true is why I requested for the change. Agree that the constant itself does not do much, just trying to add as many deterrents as possible to incorrect usage, appreciate you making the change!

PR #32227, opened 2024-03-28
fix: Make sure order of vertices does not make opl calculations flaky

nidhi-nair:
## Description
 
Fixes the check for valid edges in the graph by making sure to fail fast as soon as an invalid vertex is identified.

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Summary by CodeRabbit

- **Refactor**
	- Improved the logic and clarity in the execution flow of certain internal operations, enhancing efficiency and readability.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/onload/internal/OnLoadExecutablesUtilCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/onload/internal/OnLoadExecutablesUtilCEImpl.java
@@ -702,14 +702,11 @@ public class OnLoadExecutablesUtilCEImpl implements OnLoadExecutablesUtilCE {
 
                     Set<String> vertices = Set.of(source, target);
 
-                    boolean isValidVertex = true;
+                    boolean isValidEdge = true;
 
                     // Assert that the vertices which are entire property paths have a possible parent which is either
                     // an executable or a widget or a static variable provided by appsmith at page/application level.
                     for (String vertex : vertices) {
-                        if (!isValidVertex) {
-                            break;
-                        }
                         Optional<String> validEntity = getPossibleParents(vertex).stream()
                                 .filter(parent -> {
                                     if (!executableNames.contains(parent)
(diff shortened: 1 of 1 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/onload/internal/OnLoadExecutablesUtilCEImpl.java
```diff
@@ -703,11 +702,14 @@ private DirectedAcyclicGraph<String, DefaultEdge> constructDAG(
 
                     Set<String> vertices = Set.of(source, target);
 
-                    AtomicReference<Boolean> isValidVertex = new AtomicReference<>(true);
+                    boolean isValidVertex = true;
 
                     // Assert that the vertices which are entire property paths have a possible parent which is either
                     // an executable or a widget or a static variable provided by appsmith at page/application level.
-                    vertices.stream().forEach(vertex -> {
+                    for (String vertex : vertices) {
```
sharat87, line 709: This loop can be further simplified I think. Like, no need for `isValidVertex` at all.

```
                    for (String vertex : vertices) {
                        Optional<String> validEntity = getPossibleParents(vertex).stream()
                                .filter(parent -> {
                                    if (!executableNames.contains(parent)
                                            && !widgetNames.contains(parent)
                                            && !APPSMITH_GLOBAL_VARIABLES.contains(parent)) {
                                        return false;
                                    }
                                    return true;
                                })
                                .findFirst();
                        // If any of the generated entity names from the path are valid appsmith entity name,
                        // the vertex is considered valid
                        if (!validEntity.isPresent()) {
                            return false;
                        }
                    }

                    return true;
```

Should to the same thing I think. 🤔

Other than that I don't know enough to comment/review on this. 😄

nidhi-nair: I retained the variable for readability mostly. You think this reads well for other folks? 🤔 I'll change it then!

sharat87: Less on the variable, but the point where the loop breaks becomes more obvious. Either way, not a strong opinion, very minor. 🙂

sharat87 (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/onload/internal/OnLoadExecutablesUtilCEImpl.java
nidhi-nair: Gotcha, moved the break around to accommodate for this. I'm going to force merge since this is a minor change.

PR #32294, opened 2024-04-01
chore: Autowire MongoDB beans with annotation

sharat87:
The heart change of this PR is in [`BaseAppsmithRepositoryCEImpl`](https://github.com/appsmithorg/appsmith/pull/32294/files#diff-5b9270d513f9654dd47da330dca88f1426fdbedf35dde0d5823ac29795b1bebb).

By using `@Autowire` annotation to inject the `mongoOperations` and `mongoConverter` beans, we remove MongoDB imports and injections in all the `Impl` classes. This is a whole lot of diff that shows up on Postgres, that holds no semantic value.

EE Server and Sanity tests pass.
Do not merge when PR author isn't available. There's quite a few conflicts from this.

## Summary by CodeRabbit

- **Refactor**
  - Simplified the constructor implementation across various repository classes to enhance maintainability and readability. This includes removing specific dependencies from constructors and adopting a more streamlined approach for dependency injection, improving the overall code structure without altering the functionality visible to end-users.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/CustomApplicationRepositoryImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/CustomApplicationRepositoryImpl.java
@@ -2,11 +2,8 @@ package com.appsmith.server.repositories;
 
 import com.appsmith.server.repositories.ce.CustomApplicationRepositoryCEImpl;
 import com.appsmith.server.solutions.ApplicationPermission;
-import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.data.mongodb.core.ReactiveMongoOperations;
-import org.springframework.data.mongodb.core.convert.MongoConverter;
 import org.springframework.stereotype.Component;
 
 @Component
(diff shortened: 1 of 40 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/CustomApplicationRepositoryImpl.java
```diff
@@ -16,10 +13,7 @@ public class CustomApplicationRepositoryImpl extends CustomApplicationRepository
 
     @Autowired
     public CustomApplicationRepositoryImpl(
-            @NonNull ReactiveMongoOperations mongoOperations,
-            @NonNull MongoConverter mongoConverter,
-            CacheableRepositoryHelper cacheableRepositoryHelper,
-            ApplicationPermission applicationPermission) {
-        super(mongoOperations, mongoConverter, cacheableRepositoryHelper, applicationPermission);
+            CacheableRepositoryHelper cacheableRepositoryHelper, ApplicationPermission applicationPermission) {
```
mohanarpit, line 16: Since the `CacheableRepositoryHelper` is also being autowired through field annotations, do we need it in the constructor?

sharat87: These two are being used in `CustomApplicationRepositoryCEImpl.java`. Previously, it was taking them from `BaseAppsmithRepositoryCEImpl`, but now I've made them `private` so other classes don't just assume it, and start using. If a class wants access to one of these, they have to explicitly inject it for their own class.

That's one part.

The other is, why is this still in a constructor param, and not a field annotation. I'm _only_ converting constructor params to field injects in the `BaseAppsmithRepository*Impl` classes, since that has the biggest impact and benefit by far. For the other domain-specific repo classes like this one, breaking the recommendation of not using constructor params is just not worth it.

Does that sound okay?

I'll go add a comment to that class describing why we're using field annotations instead of constructor params (after the PR perhaps) to point out why we're not following the recommendation there.

mohanarpit (approved):

PR #32326, opened 2024-04-02
fix: action selector create flow

### LINKED ISSUE #7745: [Bug]-[2000]:Query created from property pane, doesn't bind the query to the property it was added from
## Description
- [ ] when user adds a query from the property pane, it doesn't bind the query to the property it was added from
- [x] when user hits add query from the property pane it takes to the new datasources page but it should take to the active datasources when there are datasources already added

hetunandu:
## Description

Fixes the broken flow where, on creation of a query or JS object in via the ActionCreator UI, the widget property that initiated it, does not get bound to the newly created item.

We will now keep reference of the setter method that would set the binding value to the property and when the item is created, it will call the method with the correct bind value. 

This will wait till that particular action type is created.

Fixes #7745 

## Automation

/ok-to-test tags="@tag.JS"

> [!IMPORTANT]  
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8550162310>
> Commit: `57e69a0c47496d1a1ee97e04991f34ca79440181`
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8550162310&attempt=1" target="_blank">Click here!</a>

## Summary by CodeRabbit

- **New Features**
	- Introduced the ability to create new queries and JavaScript functions directly from action creators, enhancing the workflow for building applications.
	- Added support for binding newly created queries, APIs, and JavaScript objects to button onClick events, streamlining the process of connecting actions to UI elements.
- **Enhancements**
	- Improved the code modularity and readability in the Editor Navigation by refactoring the logic for closing announcement modals.
- **Tests**
	- Added regression tests to ensure the reliability of creating actions via an action selector and their proper binding to properties.


app/client/src/components/editorComponents/ActionCreator/helpers.tsx
```diff
@@ -440,10 +438,18 @@ function getApiAndQueryOptions(
     id: "create",
     icon: "plus",
     className: "t--create-datasources-query-btn",
-    onSelect: () => {
-      dispatch(setShowCreateNewModal(true));
-      if (isSideBySideEnabled) {
-        dispatch(setIdeEditorViewMode(EditorViewMode.SplitScreen));
```
albinAppsmith, line 446: Why did the auto switch to side by side removed?

ashit-rath (commented):
@hetunandu Can we have an EE DP? I would like to test out a few things there

ashit-rath (dismissed):

app/client/src/sagas/ActionSagas.ts
```diff
@@ -1110,6 +1117,31 @@ function* updateEntitySavingStatus() {
...
+    yield put(setIdeEditorViewMode(EditorViewMode.SplitScreen));
+  }
+
+  // Wait for a query to be created
+  const createdQuery: ReduxAction<BaseAction> = yield take(
+    ReduxActionTypes.CREATE_ACTION_SUCCESS,
+  );
+
+  // A delay is needed to ensure the callback function has reference to the latest created Query
+  yield delay(100);
```
rishabhrathod01, line 1139: Could you elaborate a little here, was there a reference issue with callback function?

hetunandu: During some stress testing, I saw that the reference to the latest query is not available to the component. Sometimes it is because the ActionSelector component has not even rendered. This was the reason I had asked if there was a more static method that can achieve this.

Adding the delay will help in some scenarios but it is not foolproof and can break at times.

hetunandu: As discussed on the same thread, we can unblock by this method right now while we figure out how to add this static method

rishabhrathod01: >  I saw that the reference to the latest query is not available to the component. Sometimes it is because the ActionSelector component has not even rendered. 

- `CREATE_NEW_QUERY_FROM_ACTION_CREATOR` will only be called from the action selector UI is rendered, correct? I couldn’t think of a scenario where action selector is not rendered and still the action is triggered. 
- I am trying to reproduce the issue to understand how reference to the query is lost, could you share the steps to reproduce.

hetunandu: If you go to New Datasource option and create a query from there, it is in the datasource screen and it will not render the property pane, and hence not the action selector as well

rishabhrathod01: Okay, as I tried the scenarios. I realised that once the action is created we close the property pane and hence the Action selector will be unmounted. Due to this, we might not be able to update the binding. 
- Though i was not able to reproduce such scenario but could we make sure we only open the side by side view once we call the method? This would avoid such issue correct?

hetunandu: > Though i was not able to reproduce such scenario but could we make sure we only open the side by side view once we call the method? This would avoid such issue correct?

Not possible as the action creation success is what navigates the user to the action url. We also wait for the same event before calling the setter method as that adds the newly created action in the store and then updates the Action Selector component.

rishabhrathod01: Got it.

rishabhrathod01 (approved):
LGTM

PR #32333, opened 2024-04-02
fix: MultiEnv switcher visible in view mode in cloud

### LINKED ISSUE #32315: [Bug]: Switch environment dropdown is visible in Preview and View mode also

### Description

When I deploy the application, I see that the Switch environment dropdown is visible in Preview and View [deployed] mode also

### Steps To Reproduce

Taking a look at [this](https://www.notion.so/appsmith/Engineering-Tech-Spec-Multiple-Environments-92ac94139bcb40378d59e4a0151ba139?pvs=4#1535aded3f32409fa85038fbba2d8cf8) workflow, I understand that for the view mode [as admin] as well as preview mode, the switch env workflow, and hence, ramps should not be seen.
This was the behavior until latest tagged.
However, on release, I see that on the view mode [as admin] and preview mode, the switch env workflow and ramps are seen.
cc:

### Public Sample App

_No response_

### Environment

Release

### Severity

### Issue video log

_No response_

### Version

Cloud

brayn003:
## Description
Issue with ME switcher showing up in cloud

Fixes #32315

## Automation

/ok-to-test tags="@tag.MultiEnv"

> [!IMPORTANT]  
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8520479330>
> Commit: `20ae88502963ec88f686ef237d4892886189f088`
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8520479330&attempt=1" target="_blank">Click here!</a>

## Summary by CodeRabbit

- **New Features**
	- Updated logic for displaying the environment switcher based on new feature flags and user roles.
- **Tests**
	- Added tests for the updated environment switcher visibility logic.

### DIFF
--- a/app/client/src/ce/hooks/useShowEnvSwitcher.ts
+++ b/app/client/src/ce/hooks/useShowEnvSwitcher.ts
@@ -20,13 +20,18 @@ const useShowEnvSwitcher = ({ viewMode }: { viewMode: boolean }) => {
     const isLoaded = areEnvironmentsFetched(state, workspace?.id);
     const list = getEnvironmentsWithPermission(state);
     const isDefault = list?.[0]?.isDefault;
-    return isLoaded && (list.length === 0 || (list.length === 1 && isDefault));
+    if (!isFeatureEnabled) {
+      return true;
+    } else {
+      return (
+        isLoaded && (list.length === 0 || (list.length === 1 && isDefault))
+      );
+    }
   });
 
   const isRampAllowed = useSelector((state) =>
     showProductRamps(RAMP_NAME.MULTIPLE_ENV, true)(state),
   );
-
   if (!isFeatureEnabled && !isRampAllowed) {
     return false;
   }
--- a/app/client/src/ce/hooks/useShowEnvSwitcher.ts
+++ b/app/client/src/ce/hooks/useShowEnvSwitcher.ts
@@ -30,6 +30,7 @@ const useShowEnvSwitcher = ({ viewMode }: { viewMode: boolean }) => {
   if (!isFeatureEnabled && !isRampAllowed) {
     return false;
   }
+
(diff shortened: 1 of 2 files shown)


app/client/src/ce/hooks/useShowEnvSwitcher.ts
```diff
@@ -20,16 +20,22 @@ const useShowEnvSwitcher = ({ viewMode }: { viewMode: boolean }) => {
...
+        isLoaded && (list.length === 0 || (list.length === 1 && isDefault))
+      );
+    }
   });
 
   const isRampAllowed = useSelector((state) =>
     showProductRamps(RAMP_NAME.MULTIPLE_ENV, true)(state),
   );
-
   if (!isFeatureEnabled && !isRampAllowed) {
```
ankitakinger, line 35: do we need this check here then?

brayn003: yes, this is another case

ankitakinger: But `!isFeatureEnabled` check will already pass on top so same check won't be needed here right? Also, if it makes sense, we can actually move the entire condition to the top newly added line.

brayn003: Umm, `isFeatureEnabled` in line number 23 is actually getting called inside the `isMultiEnvNotPresent` selector function. Both of them have different scopes

ankitakinger: oh alright. Didn't open the collapsed lines above. Thanks.

ankitakinger (approved):

PR #32343, opened 2024-04-02
fix: Scheduling synchronized block on bounded elastic threadpool instead of main event threadpool

trishaanand:
…

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  


sharat87 (approved):
Approving. That typecast is making me nervous. I'll raise another PR to make it go away. That okay?

PR #32347, opened 2024-04-02
chore: Minimal surface API for HealthCheckServiceCE

sharat87:
## Summary by CodeRabbit

- **Refactor**
	- Improved health check functionality for enhanced app performance and reliability.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/HealthCheckServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/HealthCheckServiceCEImpl.java
@@ -29,7 +29,7 @@ public class HealthCheckServiceCEImpl implements HealthCheckServiceCE {
 
     @Override
     public Mono<String> getHealth() {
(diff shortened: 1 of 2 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/HealthCheckServiceCEImpl.java
```diff
@@ -30,11 +29,10 @@ public HealthCheckServiceCEImpl(
 
     @Override
     public Mono<String> getHealth() {
-        return Mono.zip(getRedisHealth(), getMongoHealth()).map(tuple -> "All systems are Up");
+        return Mono.when(getRedisHealth(), getMongoHealth()).thenReturn("All systems are Up");
```
mohanarpit, line 32: Only because you are modifying this line already. 😄 

And thanks for this imrpovement!

```suggestion
        return Mono.when(getRedisHealth(), getMongoHealth()).thenReturn("All systems are up");
```

mohanarpit (dismissed):

PR #32352, opened 2024-04-03
chore: Get control over what stack traces show up in logs

sharat87:
This should give us control over what exceptions print a stack trace and what exceptions don't need to. Should help us get more control over our logs and reduce noice.

shot-2024-04-02-12-33-05

## Summary by CodeRabbit

- **New Features**
	- Enhanced error logging capabilities to selectively hide stack traces in logs for improved log readability and security.
- **Bug Fixes**
	- Improved error handling for datasource timeouts, minimizing unnecessary stack trace logging.

### DIFF
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/exceptions/BaseException.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/exceptions/BaseException.java
@@ -41,7 +41,7 @@ public abstract class BaseException extends RuntimeException {
 
     @SuppressWarnings("unchecked")
     public <T extends BaseException> T hideStackTraceInLogs() {
-        hideStackTraceInLogs = false;
+        hideStackTraceInLogs = true;
         return (T) this;
     }
 
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/exceptions/BaseException.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/exceptions/BaseException.java
@@ -11,6 +11,8 @@ public abstract class BaseException extends RuntimeException {
 
     private Map<String, String> contextMap;
 
+    private boolean hideStackTraceInLogs = false;
+
     public BaseException(String message) {
         super(message);
         contextMap = MDC.getCopyOfContextMap();
@@ -36,4 +38,14 @@ public abstract class BaseException extends RuntimeException {
     public abstract String getDownstreamErrorCode();
 
     public abstract String getErrorType();
+
+    @SuppressWarnings("unchecked")
+    public <T extends BaseException> T hideStackTraceInLogs() {
+        hideStackTraceInLogs = false;
+        return (T) this;
+    }
+
+    public boolean shouldHideStackTraceInLogs() {
+        return hideStackTraceInLogs;
+    }
 }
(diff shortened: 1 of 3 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/DatasourceStructureSolutionCEImpl.java
```diff
@@ -116,12 +116,11 @@ public Mono<DatasourceStructure> getStructure(DatasourceStorage datasourceStorag
...
                 .timeout(Duration.ofSeconds(GET_STRUCTURE_TIMEOUT_SECONDS))
-                .onErrorMap(
-                        TimeoutException.class,
-                        error -> new AppsmithPluginException(
+                .onErrorMap(TimeoutException.class, error -> new AppsmithPluginException(
                                 AppsmithPluginError.PLUGIN_GET_STRUCTURE_TIMEOUT_ERROR,
                                 "Appsmith server timed out when fetching structure. Please reach out to appsmith "
-                                        + "customer support to resolve this."))
+                                        + "customer support to resolve this.")
+                        .hideStackTraceInLogs())
```
mohanarpit, line 123: This is a fantastic start! Thank you so much for creating the draft PR. Do you think a more holistic way of achieving this would be to add this flag to the AppsmithException enum? This way, we can control the stack trace printing from a single location. Or do you believe that the same exception in different places needs to be logged differently?

sharat87: I'm guessing you meant `AppsmithError` enum, and replying with that correction.

Latter yes. Adding this to `AppsmithError` is not granular enough. It can never be granular enough. It's one of the reasons I don't believe that enum is serving us well anymore (an opinion I've silently held for >2yrs).

Putting it here, one,  gives us much needed granularity, two, avoids us getting in the pit of "oops, this error doesn't include stack trace in logs so now I don't have what I need to troubleshoot", and three, makes it explicitly clear at the source that this is not going to print the stack trace and make bloody sure we're okay with that.

Implicit over-communication is okay, but under-communication has to be explicit in my opinion.

Does that sound okay? 🙂

mohanarpit: Sorry, yes. I meant `AppsmithError`. And yes, I am also in agreement with you about being explicit in code, to prevent miscommunications and human errors.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/exceptions/BaseException.java
```diff
@@ -36,4 +38,14 @@ public String getMessage() {
     public abstract String getDownstreamErrorCode();
 
     public abstract String getErrorType();
+
+    @SuppressWarnings("unchecked")
+    public <T extends BaseException> T hideStackTraceInLogs() {
+        hideStackTraceInLogs = false;
+        return (T) this;
+    }
```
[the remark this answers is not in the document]
sharat87: Thanks. Fixed.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/DatasourceStructureSolutionCEImpl.java
sharat87: @mohanarpit, could you approve if it looks okay please? 🙂

mohanarpit (approved):

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/exceptions/BaseException.java
```diff
@@ -36,4 +38,14 @@ public String getMessage() {
     public abstract String getDownstreamErrorCode();
 
     public abstract String getErrorType();
+
+    @SuppressWarnings("unchecked")
+    public <T extends BaseException> T hideStackTraceInLogs() {
+        hideStackTraceInLogs = true;
```
mohanarpit, line 44: QQ: why is the default value to hide? Shouldn't it be false?

PR #32399, opened 2024-04-04
chore: Added trace and span info to context

nidhi-nair:
## Description
Contains changes to make sure that any ongoing observation shows up in the logs as well as is retained in context in case we need to trigger a span manually.

Screenshot 2024-04-04 at 9 50 49 AM

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Summary by CodeRabbit

- **New Features**
	- Enhanced performance monitoring with micrometer tracing and observation handling.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/MDCConfig.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/MDCConfig.java
@@ -3,13 +3,11 @@ package com.appsmith.server.configurations;
 import com.appsmith.server.filters.MDCFilter;
 import com.appsmith.server.helpers.LogHelper;
 import io.micrometer.observation.Observation;
-import io.micrometer.tracing.Tracer;
 import io.micrometer.tracing.handler.TracingObservationHandler;
 import jakarta.annotation.PostConstruct;
 import jakarta.annotation.PreDestroy;
 import org.reactivestreams.Subscription;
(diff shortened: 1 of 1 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/MDCConfig.java
```diff
@@ -2,22 +2,33 @@
...
 @Configuration
 public class MDCConfig {
 
     private static final String MDC_CONTEXT_REACTOR_KEY = "MDCConfig";
+    private static final String OBSERVATION = "micrometer.observation";
+    private static final String TRACE_ID = "traceId";
+    private static final String SPAN_ID = "spanId";
+
+    @Autowired
+    Tracer tracer;
```
sharat87, line 31: Can we use constructor param injection instead? 🤔
Like make it `private final` and add an `@RequiredArgsConstructor` on the class?

sharat87 (dismissed):
Minor consideration, so approving.

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/MDCConfig.java
nidhi-nair: Whoops, we don't need that bean, I was just using it to see if the brave beans got initialized. :P Thanks for catching that!

sharat87 (approved):

PR #32403, opened 2024-04-04
chore: Remove more MongoDB API uses in tests

sharat87:
Removes a lot of API uses from MongoDB libraries, that don't strictly have to depend on MongoDB dependencies.

### DIFF
--- a/app/server/appsmith-server/src/test/java/com/appsmith/server/services/LayoutActionServiceTest.java
+++ b/app/server/appsmith-server/src/test/java/com/appsmith/server/services/LayoutActionServiceTest.java
@@ -76,7 +76,6 @@ import java.util.stream.Collectors;
 
 import static com.appsmith.server.acl.AclPermission.READ_PAGES;
 import static com.appsmith.server.constants.FieldName.DEFAULT_PAGE_LAYOUT;
-import static com.mongodb.assertions.Assertions.assertNull;
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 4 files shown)


app/server/appsmith-server/src/test/java/com/appsmith/server/services/LayoutActionServiceTest.java
```diff
@@ -1279,7 +1278,7 @@ public void introduceCyclicDependencyAndRemoveLater() {
         // by default there should be no error in the layout, hence no error should be sent to ActionDTO/ errorReports
         // will be null
         assertNotNull(createdAction);
-        assertNull(createdAction.getErrorReports());
+        assertThat(createdAction.getErrorReports()).isNull();
```
mohanarpit, line 1281: For my curiosity, what is the difference between the 2 styles of assert? Do we have a preference one way or another?

mohanarpit (approved):

app/server/appsmith-server/src/test/java/com/appsmith/server/services/LayoutActionServiceTest.java
sharat87: In general, we want to be using `asssertThat()...` since it has a much richer assertion vocabulary, and produces incredibly better error messages on assertion failures. But for simpler assertions like checking for `null`, it doesn't make much difference and is more a matter of consistency.

PR #32411, opened 2024-04-04
chore: Use update clause from Bridge API

abhvsn:
PR to update Field usage and adding a UpdateBridge API while saving the datasource structure.

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8642139106>
> Commit: 90a85e25480d9a58ecdc64c24dd88ca9ee3367fd
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8642139106&attempt=1" target="_blank">Click here!</a>

## Summary by CodeRabbit

- **New Features**
	- Enhanced the data model to support default resources for pages and actions, improving the management and organization of these elements.
- **Refactor**
	- Improved code quality and maintainability by utilizing static references and utility methods for field access in database queries, leading to more robust and error-resistant code.


app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/ce/ActionCollectionCE_DTO.java
```diff
@@ -175,5 +176,5 @@ public String calculateContextId() {
         return this.getPageId();
     }
 
-    public static class Fields {}
+    public static class Fields extends BranchAwareDomain.Fields {}
```
sharat87, line 179: We should only be doing this if the parent class also extends `BranchAwareDomain` right? Looks like `ActionCollectionCE_DTO` doesn't extend `BranchAwareDomain`. Any reason we're doing this? 🤔

sharat87 (changes requested):
Looks good mostly. I think we want to move the new field constants to a class that extends `BranchAwareDomain` and we should be good. I'm also working on a linter that will ensure this for us soon. Thanks!

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/ce/ActionCollectionCE_DTO.java
abhvsn: @sharat87 this was done to have the consistency in how we want to refer the nested fields in defaultResources like `defaultResources_pageId`, `defaultResources_actionId`. In case of `ActionCollectionCE_DTO` we have non-null values for `defaultResources.pageId` which is being referenced in DB calls.
Example:
```
String pageIdFieldPath = dotted(
                        ActionCollection.Fields.publishedCollection,
                        ActionCollectionDTO.Fields.defaultResources_pageId);
```

sharat87: But that consistency doesn't make sense here, right? The `defaultResources` in `ActionCollectionCE_DTO` isn't coming from extending `BranchAwareDomain`. It's coming from explicit definition in `ActionCollectionCE_DTO` class itself. These are two separate fields, `ActionCollectionCE_DTO.defaultResources` and `BranchAwareDomain.defaultResources`. And so the field name constant should also be a separate definition, and not be linked.

Ideally, the `Fields` class should extend another class's `Fields`, only if the parent class itself extends that other's class' `Fields`.

abhvsn: Sounds good will update the implementation.

sharat87 (approved):

PR #32446, opened 2024-04-05
fix: template name usage in analytics

### LINKED ISSUE #32437: [Bug]: An internal server error is being thrown on adding any template within an application.

### Description

An internal server error is being thrown on adding any template within an application.
image

### Steps To Reproduce

1. Create an application on https://release.app.appsmith.com/
2. Try to add a page from any template/building block
3. Notice that it fails and an internal server error is thrown

### Public Sample App

_No response_

### Environment

Release

### Severity

### Issue video log

https://www.loom.com/share/5aa9d034e40d4b80ac10125924585e8b?sid=64527492-3473-4b5b-ad2a-506dc2cf9b98

### Version

Release v1.19

AnaghHegde:
## Description
Fix NPE caused by the template name used for analytics PR. 

Fixes #32437 

## Automation

/ok-to-test tags="tag.Templates"

> [!IMPORTANT]  
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8567991310>
> Commit: `8e04b161dcacb89fb4c08ec8a32cbd4c2e7c4129`
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8567991310&attempt=1" target="_blank">Click here!</a>

## Summary by CodeRabbit

- **Refactor**
	- Improved how the application name is handled when merging templates with applications for enhanced reliability.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
@@ -281,6 +281,8 @@ public class ApplicationTemplateServiceCEImpl implements ApplicationTemplateServ
             List<String> pagesToImport) {
         Mono<ApplicationImportDTO> importedApplicationMono = getApplicationJsonFromTemplate(templateId)
                 .flatMap(applicationJson -> {
+                    String templateName =
+                            applicationJson.getExportedApplication().getName();
                     if (branchName != null) {
                         return applicationService
                                 .findByBranchNameAndDefaultApplicationId(
@@ -292,15 +294,13 @@ public class ApplicationTemplateServiceCEImpl implements ApplicationTemplateServ
                                         applicationJson,
                                         pagesToImport))
                                 .map(importableArtifact -> (Application) importableArtifact)
-                                .zipWith(Mono.just(
-                                        applicationJson.getExportedApplication().getName()));
+                                .zipWith(Mono.just(templateName));
                     }
                     return importService
                             .mergeArtifactExchangeJsonWithImportableArtifact(
                                     organizationId, applicationId, null, applicationJson, pagesToImport)
                             .map(importableArtifact -> (Application) importableArtifact)
-                            .zipWith(Mono.just(
-                                    applicationJson.getExportedApplication().getName()));
+                            .zipWith(Mono.just(templateName));
                 })
                 .flatMap(tuple -> {
                     Application application = tuple.getT1();
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
@@ -281,8 +281,9 @@ public class ApplicationTemplateServiceCEImpl implements ApplicationTemplateServ
             List<String> pagesToImport) {
         Mono<ApplicationImportDTO> importedApplicationMono = getApplicationJsonFromTemplate(templateId)
                 .flatMap(applicationJson -> {
-                    String templateName =
-                            applicationJson.getExportedApplication().getName();
+                    String templateName = applicationJson.getExportedApplication() != null
+                            ? applicationJson.getExportedApplication().getName()
+                            : "";
                     if (branchName != null) {
                         return applicationService
                                 .findByBranchNameAndDefaultApplicationId(
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
@@ -281,9 +281,11 @@ public class ApplicationTemplateServiceCEImpl implements ApplicationTemplateServ
             List<String> pagesToImport) {
         Mono<ApplicationImportDTO> importedApplicationMono = getApplicationJsonFromTemplate(templateId)
                 .flatMap(applicationJson -> {
-                    String templateName = applicationJson.getExportedApplication() != null
-                            ? applicationJson.getExportedApplication().getName()
-                            : "";
+                    String templateName = "";
+                    if (applicationJson.getExportedApplication() != null
+                            && applicationJson.getExportedApplication().getName() != null) {
+                        templateName = applicationJson.getExportedApplication().getName();
+                    }
                     if (branchName != null) {
                         return applicationService
                                 .findByBranchNameAndDefaultApplicationId(


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
```diff
@@ -281,6 +281,8 @@ public Mono<ApplicationImportDTO> mergeTemplateWithApplication(
             List<String> pagesToImport) {
         Mono<ApplicationImportDTO> importedApplicationMono = getApplicationJsonFromTemplate(templateId)
                 .flatMap(applicationJson -> {
+                    String templateName =
+                            applicationJson.getExportedApplication().getName();
```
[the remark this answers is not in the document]
trishaanand: @AnaghHegde : Can you please add the null check for exportedApplication object before accessing the internal property (name)? This is a good developer practice to follow.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
```diff
@@ -292,15 +294,13 @@ public Mono<ApplicationImportDTO> mergeTemplateWithApplication(
...
-                                        applicationJson.getExportedApplication().getName()));
+                                .zipWith(Mono.just(templateName));
                     }
                     return importService
                             .mergeArtifactExchangeJsonWithImportableArtifact(
                                     organizationId, applicationId, null, applicationJson, pagesToImport)
                             .map(importableArtifact -> (Application) importableArtifact)
-                            .zipWith(Mono.just(
-                                    applicationJson.getExportedApplication().getName()));
+                            .zipWith(Mono.just(templateName));
```
trishaanand, line 303: If templateName is null, the existing NPE error being seeing on release would still be seen. This code change doesn't fix the underlying issue at all.

AnaghHegde: @trishaanand the reason for the templateName is null is because in the method `mergeArtifactExchangeJsonWithImportableArtifact` we are setting the name of the application inside the json to be null to avoid renaming the application in the Add page from template flow. I hope that is clear.

trishaanand: @AnaghHegde : Mono.just can not hold null. Please refer to the manual around this : https://github.com/reactive-streams/reactive-streams-jvm#user-content-2.13

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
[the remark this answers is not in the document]
AnaghHegde: @trishaanand While i agree that this is a good practice, but in this scenario these json files are handled by Appsmith folks internally. Rest assured, the application object will not be empty because this is already handled while uploading the template to CS.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
AnaghHegde: @trishaanand I am saying that template name is null only because in the flow we are setting as null, else this cannot be null. Its done here - https://github.com/appsmithorg/appsmith/blob/release/app/server/appsmith-server/src/main/java/com/appsmith/server/applications/imports/ApplicationImportServiceCEImpl.java#L257
We do not need to check for the name to be null. This check is handled while uploading the template to CS.

trishaanand: I am not a big fan of relying on third parties to do the validation. If we are at risk of throwing an NPE, we do validations ourselves and add handling for this. 
In this case, the NPE is indeed being thrown because of Mono.just emitting a null value :
https://app.mezmo.com/8c1d4711f3/logs/view?t=timestamp%3A1712298228885&a=1712298228885.1729304112732663820&q=NullPointerException&tags=ha-release

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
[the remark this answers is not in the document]
AnaghHegde: @trishaanand I have made changes. But the null check is done in the code flow before we upload the json file to CS. If the name is null the upload will fail. I am not relying on anything else.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
AnaghHegde: I have made changes. But the null check is not done via the third party, its done in the code flow before we upload the json file to CS. If the name is null the upload will fail.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
```diff
@@ -281,8 +281,9 @@ public Mono<ApplicationImportDTO> mergeTemplateWithApplication(
             List<String> pagesToImport) {
         Mono<ApplicationImportDTO> importedApplicationMono = getApplicationJsonFromTemplate(templateId)
                 .flatMap(applicationJson -> {
-                    String templateName =
-                            applicationJson.getExportedApplication().getName();
+                    String templateName = applicationJson.getExportedApplication() != null
```
trishaanand, line 284: @AnaghHegde : Thanks for making this change. This gracefully handles a bad state where exported application is null. Unfortunately, this still doesn't assure us that templateName would not be null.

Could we default templateName to an empty string if you want to continue to use Mono.just(templateName)?

AnaghHegde: The default value is empty string, look at line number 286.

trishaanand: @AnaghHegde Apologies. May be I am not uderstanding this code change completely. Here's my understanding of the above code : 
templateName is set to empty string if exportedApplication is null. 
What happens if exportedApplication is not null, but exportedApplication.name is null.

trishaanand (approved):

PR #32448, opened 2024-04-05
chore: Add separate request/response views

sharat87:
This PR gets finer control into what fields are allowed in
request-body-only, vs what's allowed in response-body-only. This leaves
the fields to separately controlled regarding what can go into the
database and what can't.

[Slack
thread](https://theappsmith.slack.com/archives/CPQNLFHTN/p1710125307810949).

Passes `@tag.All` Cypress tests on both CE and EE.

## Sanity result:

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8644101097>
> Commit: 0ce8d68c5e2f3b8197f091309688ebef004ffae4
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8644101097&attempt=1" target="_blank">Click here!</a>

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8644921068>
> Commit: 0ce8d68c5e2f3b8197f091309688ebef004ffae4
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8644921068&attempt=2" target="_blank">Click here!</a>

## Summary by CodeRabbit

- **New Features**
	- Enhanced data privacy and control by implementing new views for HTTP request and response handling.

- **Refactor**
	- Simplified controller construction and dependency management in server components.

- **Bug Fixes**
	- Removed unnecessary properties from HTTP responses to streamline client-server communication.


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ce/ActionCE_DTO.java
```diff
@@ -40,89 +41,84 @@
...
      * This is a list of fields specified by the client to signify which fields have dynamic bindings in them.
      * TODO: The server can use this field to simplify our Mustache substitutions in the future
      */
-    @JsonView(Views.Public.class)
+    @JsonView({Views.Public.class, FromRequest.class})
     List<Property> dynamicBindingPathList;
 
-    @JsonProperty(access = JsonProperty.Access.READ_ONLY)
-    @JsonView(Views.Public.class)
+    @JsonView(ToResponse.class)
```
nidhi-nair, line 109: I understand that we're using `FromRequest` to differentiate views that are restricted for client to edit. Could you please help me understand what the difference between `ToResponse` and `Public` is in this respect?

subrata71 (commented):
**QQ:** 
How can we ensure that certain fields are included in both the request and response for `API_1`, while for `API_2`, these fields are allowed in the request payload only?

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ce/ActionCE_DTO.java
sharat87: Hey, thanks for asking. Good question. It likely will work fine if I keep it to `ToResponse`. I've done so many experiments I don't recall why I didn't revert this. Let me go check that and come back. Thanks!

sharat87:
@subrata71, thanks. Very good question. It was something I was thinking about when working on this.

Instead of giving you an answer, I'll share two points that should help you take a decision when you face something like that.

One, we can always introduce more view class like `FromRequst`, and use that in the controller method. Can make that view class' name descriptive, make it designed for reuse... or not, we have that freedom.

Two, the above can get hairy though. A simpler solution could be to use a separate DTO class, just for the request payload, and only have fields that we want in the payload. Just don't even have field definitions for anything we don't want to accept. This can make for very readable and easy-to-understand flow. The more controller methods use the same DTO class with different view classes, the harder it will be to understand and debug.

Let me know if that doesn't help. Thanks!

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ce/ActionCE_DTO.java
sharat87: Done. Removed `ToResponse`, and all Cypress tests pass on EE. Can you review/approve please? Thanks!

nidhi-nair (approved):

PR #32461, opened 2024-04-05
ci: Delete jobs that are not part of SDLC process anymore

nidhi-nair:
## Description

Deleting CypressAddKnownfailedtests because the job was meant to help with maintaining our flaky tests list, but we have an Appsmith app to manage this now.

Deleting label-checker because we do not follow the process for "Test plan approved" anymore.

## Automation

/ok-to-test tags=""

> [!CAUTION]  


mohanarpit:
@sharat87 @pratapaprasanna  Are you sure we want to remove the `packer-ami` CI job? I know we haven't built an AMI in a while, but do you believe it's good to have a CI job that does this?

nidhi-nair:
@mohanarpit Goutham mentioned that we have moved this job to the cloud-deployment repo and is hence okay to be removed from here.

Issue #9941, opened 2021-12-22
[Bug]-[2800]:CRUD happens even with Read-Only access in MYSQL


chandannkumar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Update, delete functionality are working even if the datasource is authorized only for Read access.

### Steps To Reproduce

1. Create a Datasource of MYSQL with  Read Only Access
2. Try to update or delete using mysql query
3. Observe the table got updated in Read-only access

### Environment

Production

### Version

Cloud

rohan-arthur:
same happens for postgres too
6m usage: 1000+2500 = 3500
applying 20% requirement: 700

## Stats
| Stat | Values |
| ------------- | ------------- |
| Reach     | 3500 |
| Effort (months)  | 0.5 |

ajinkyakulkarni:
## Stats
| Stat | Values |
| ------------- | ------------- |
| Reach     | 3500 |
| Effort (months)  | 0.25 |

Nikhil-Nandagopal:
@rohan-arthur @ajinkyakulkarni not everyone will need read only mode so I'm lowering the reach

sumitsum:
It seems the issue exists with MySQL but not Postgres. Attaching the usage in Postgres: 

https://user-images.githubusercontent.com/1757421/180231770-bc581357-8a7f-4256-8d4d-e7b7347d45de.mov

appsmithguru:
[Observations and RCA on this issue](https://www.notion.so/appsmith/DDL-DML-in-Read-Only-mode-in-DB-Datasource-038ffb9f48034170a3642e703b19d985)

rohan-arthur:
Easy way to set this up:

1. Create datasource from [our notion page](https://www.notion.so/appsmith/Servers-and-Credentials-665f00d9505745d59db4f3b4eda27c79?pvs=4#02d45a9ecb944962a121e2237ea50c2b), ensure that it is marked `read only`
2. run query to get the total count: `SELECT COUNT(*) FROM a;`
3. run query to add a new row: `INSERT INTO a (PersonID, LastName, attendence, salary, joiningdate, logindateandtime, gross)
VALUES (12, 'new person', true, 100, '2024-01-01', '2024-01-02T08:00:00Z', 2);`
4. repeat the step to run query to get the total count: `SELECT COUNT(*) FROM a;`

You will see that the count is incremented by 1

sneha122:
[...]

PR #32487, opened 2024-04-08
fix: connection mode hidden from mysql form config

sneha122:
## Description
MySQL datasource config has a field called `Connection mode`, which allows users to configure their datasource either in read write / read only mode. This is supposed to block the write queries if datasource is configured with read only mode, but with current implementation, it does not block, it lets us execute write query. 

On checking the codebase, it seems that this field is not at all being used in mySQL plugin to configure the datasource, further more mysql driver used may not even support configuring connection mode as of now, hence as preventive solution, hiding the connection mode form config for now.

Fixes https://github.com/appsmithorg/appsmith/issues/9941
> [!WARNING]  

## Automation

> [!IMPORTANT]  
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8612888064>
> Commit: `0a0d20e1fb40546e635ba018704b5f13ecc97ab1`
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8612888064&attempt=1" target="_blank">Click here!</a>


rishabhrathod01:
image

Changes look good on DP, also code looks good to me. Waiting for the CI to pass.

rishabhrathod01 (dismissed):
Change looks good. 
Need to fix the CI test failure and it will be ready to merge.

app/client/cypress/support/Pages/DataSources.ts
```diff
@@ -1008,7 +1008,9 @@ export class DataSources {
     dsName: "PostgreSQL" | "MySQL" | "MongoDB",
   ) {
     this.ReconnectModalValidation(dbName, dsName);
-    this.ValidateNSelectDropdown("Connection mode", "Read / Write");
+    if (dsName !== "MySQL") { // MySQL does not have a connection mode as it's not configured correctly in backend
+      this.ValidateNSelectDropdown("Connection mode", "Read / Write");
+    }
```
mohanarpit, line 1013: @sneha122 Instead of adding this if check, do you think it's better to have the relevant fields being filled and asserted inside the functions `this.FillPostgresDSForm()`, `this.FillMySqlDSForm()`, etc? This way, each datasource form can be modified individually and it also encapsulates the act of filling a form in a particular function.

sneha122: Yes, completely agree with you! This way whenever we decide to add support for mysql, we can modify its individual function to make this assertion, will update the code, Thanks!

mohanarpit (approved):

rishabhrathod01 (dismissed):
LGTM

rishabhrathod01 (approved):

Issue #29804, opened 2023-12-22
[Task]: Rotate private-public keypair used to generate and validate the signature from CS  


abhvsn:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SubTasks

- Make sure to keep the deprecated public key in Appsmith instance to avoid interruption.

Merge sequence:
- Appsmith PR which can support both the keys older and updated one
- CS PR to merge the keys with updated keypair
- Merge the Appsmith PR which removes the deprecated public key

deepika-appsmith:
marking this high priority as it has been 6 months since the keys were changed. For security aspect we should rotate the keys.

abhvsn:
Doc on how to rotate the keys: https://www.notion.so/appsmith/Procedure-to-rotate-keys-bb87c366f0f744839d3613398a12f385

daaliachhak17:
Is this issue closed ? or needs more work?

PR #32525, opened 2024-04-09
chore: Rotate sigining keys used in signature verification for CS API calls

abhvsn:
## Description
Doc on how to update signing keys: https://www.notion.so/appsmith/Procedure-to-rotate-keys-bb87c366f0f744839d3613398a12f385

CS PR: https://github.com/appsmithorg/cloud-services/pull/1630

Fixes https://github.com/appsmithorg/appsmith/issues/29804

## Automation

> [!IMPORTANT]  
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8615279532>
> Commit: `5020002c35349338b67848b7529c55aa90910ebc`
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8615279532&attempt=1" target="_blank">Click here!</a>

## Summary by CodeRabbit

- **New Features**
	- Enhanced security by supporting key rotation, allowing verification with multiple public keys.


app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/SignatureVerifier.java
```diff
@@ -5,33 +5,42 @@
...
 @Slf4j
 public class SignatureVerifier {
 
-    // Public key for verifying signatures with ED25519 scheme.
-    private static final String PUBLIC_VERIFICATION_KEY =
+    // Public key for verifying signatures with ED25519 scheme. We are using two keys to support key rotation. Once the
+    // keys are rotated on CS side, we can remove the previous key in the upcoming versions.
+    private static final String CURRENT_PUBLIC_VERIFICATION_KEY =
+            "AAAAC3NzaC1lZDI1NTE5AAAAIAt5shghJhvGR0JQHBoOZ73N8JjRo31MPpY1y67O1C2G";
+    private static final String PREVIOUS_PUBLIC_VERIFICATION_KEY =
```
sharat87, line 28: @abhvsn, is the old key going to be removed on CS? Does that mean old versions of Appsmith won't be able to verify the signature anymore?

abhvsn: @sharat87 nope, on CS we will introduce new keys but older keys will still be active to support the instances running on lower versions. This comment is talking about removing the `PREVIOUS_PUBLIC_VERIFICATION_KEY` from Appsmith repo itself.

sharat87: Then can we do it the other way around? Like, add the new private key to CS first, let that be deployed, then we can switch the key here in Appsmith, right? We won't need this temporary _current_ and _previous_ key dance here, right?

sharat87 (commented):
https://github.com/appsmithorg/appsmith/pull/32525#discussion_r1565352703

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/SignatureVerifier.java
abhvsn: @sharat87 we want to match the Appsmith's tagged version to be able to add the new private key in CS. Hence the requirement to add the key to instance first and then to CS.

sharat87: Understood. Thanks for the detail. I still think this is a very fragile way of solving this. We need a more robust way.

How about instead of tying the keys on CE to Appsmith version, we use a separate numbering system. Like, `key1`, `key2` etc. Appsmith sends this along as a query param to get CS to use the right key. We can handle the one version number we already have for backwards compatibility.

This will free us from having to do this current-key/previous-key dance _every time_ we have to change the key.

If we want to take up something like that separately after this is merged, I'm fine, but this cannot be a sustainable solution. We need something else, may be like the above, may be something else.

Let me know. Thanks again!

abhvsn: Ok understood, let me see if I can come up with better solution for this. Also as we don't have to hurry we can wait for the final solve instead of revisiting this again.

Issue #29248, opened 2023-12-01
[Task]: Sunset Confirm Before calling a function


satbir121:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SubTasks

Confirm before calling only works when called using run button and on page load. If you execute a function from a button click, confirm before calling does not work. We have no way to fix this.

We are going to close following issues:

[confirm before calling not available](https://github.com/appsmithorg/appsmith/issues/27251)
[Confirm before calling setting does not work when a JS function is trigged on a button click](https://github.com/appsmithorg/appsmith/issues/17500)

PR #30506, opened 2024-01-22
feat: Frontend changes for consolidated-api with EE test case support

vsvamsi1:
## Description
Our objective in this pr is to improve the page load time of our application by calling a consolidated-api which contains all the resources to load our pages. This PR contains all the client side changes to call the consolidated-api as well as feature flag related changes. This PR should be compatible with the EE repo.
Fixes #29650 & #29939
#### Type of change
- Chore (housekeeping or task changes that don't impact user perception)
>
>
>
- [ ] Manual
- [ ] JUnit
- [ ] Jest
- [ ] Cypress
>
>
#### Dev activity
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] PR is being merged under a feature flag

#### QA activity:
- [ ] [Speedbreak features](https://github.com/appsmithorg/TestSmith/wiki/Guidelines-for-test-plans#speedbreakers-) have been covered
- [ ] Test plan covers all impacted features and [areas of interest](https://github.com/appsmithorg/TestSmith/wiki/Guidelines-for-test-plans#areas-of-interest-)
- [ ] Test plan has been peer reviewed by project stakeholders and other QA members
- [ ] Manually tested functionality on DP
- [ ] We had an implementation alignment call with stakeholders post QA Round 2
- [ ] Cypress test cases have been added and approved by SDET/manual QA
- [ ] Added `Test Plan Approved` label after Cypress tests were reviewed
- [ ] Added `Test Plan Approved` label after JUnit tests were reviewed

## Summary by CodeRabbit

- **New Features**
  - Introduced a new API for consolidated page data fetching.
  - Added a new feature flag for enabling consolidated page load fetch.

- **Improvements**
  - Improved app loading performance by using consolidated data fetching.
  - Enhanced Cypress tests with updated wait commands and response handling.

- **Refactor**
  - Refactored `AppRouter` to use hooks and functional components.
  - Updated various reducers, actions, and sagas for new loading state management.

- **Tests**
  - Adjusted existing Cypress tests to align with the new consolidated data fetching process.

Issue #1910, opened 2020-11-25
[Task] Improve Computed Value


riodeuno:
- [x] Add bindings
- [x] Inform the user of the current row in the prompt
Use {{ currentRow.columnIdentifier }} to access any column in the table

PR #32527, opened 2024-04-09
fix: deprecate JS object confirm before calling

rishabhrathod01:
## Description

We deprecate confirm before calling for JSObjects
- For new JSObject, confirm before calling option won't be visible
- For already present JSObject, only for function where confirm before calling was enabled will be visible and once turned off will not be visible anymore.

Fixes https://github.com/appsmithorg/appsmith/issues/29248

## Test scenarios

- For an application where JSObject confirmed before calling was enabled, it should work same.
- For new JSObject, confirm before calling option shouldn't be visible. 
- Once user disable confirm before calling it should not be visible anymore. 
- Deprecation warning on info.

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8654538946>
> Commit: 6ac458bdec864a2963fa5eb59b5912607ddfa3a7
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8654538946&attempt=1" target="_blank">Click here!</a>


app/client/cypress/e2e/Regression/ServerSide/OnLoadTests/JSOnLoad1_Spec.ts
```diff
@@ -1,209 +0,0 @@
...
-      agHelper.AssertElementVisibility(
-        jsEditor._dialogBody((jsName as string) + ".getEmployee"),
-      );
-      jsEditor.ConfirmationClick("Yes");
-      agHelper.ValidateToastMessage("getEmployee ran successfully"); //Verify this toast comes in EDIT page only
-    });
-
-    //Skipping due to - "tableData":"ERROR: invalid input syntax for type smallint: "{}""
-    it("4. Tc 53 - Verify OnPage Load - Enabled & Disabling - Before Function calling for JSOBject", function () {
-      EditorNavigation.SelectEntityByName(
```
sneha122, line 130: @rishabhrathod01 test 4 and 5 are not modifying anything related to confirm before calling setting, can we please check once if we want to keep these tests or not?

sneha122: Same for test 6

app/client/cypress/e2e/Regression/ServerSide/OnLoadTests/JSOnLoad2_Spec.ts
```diff
@@ -71,119 +64,6 @@ describe(
...
-      }
-      deployMode.NavigateBacktoEditor();
-      for (let dialog = 1; dialog <= 5; dialog++) {
-        jsEditor.ConfirmationClick("Yes");
-        agHelper.Sleep(500);
-      }
-    });
-
-    it("7. Tc #1909 - Verify the sequence of of JS Object on page load", () => {
-      EditorNavigation.SelectEntityByName("JSObject1", EntityType.JSObject);
```
sneha122, line 172: Do we need to remove this test? Since it's not using anything in confirm before calling

app/client/cypress/e2e/Regression/ServerSide/OnLoadTests/JsOnLoad3_Spec.ts
```diff
@@ -185,249 +118,6 @@ describe(
...
-      agHelper.GetNClick(locators._widgetInDeployed("imagewidget"));
-      agHelper.AssertElementVisibility(jsEditor._dialogBody("CatFacts"));
-      jsEditor.ConfirmationClick("Yes");
-      cy.get(locators._toastMsg).contains(/Your cat fact|Oh No/g);
-      deployMode.NavigateBacktoEditor();
-      jsEditor.ConfirmationClick("No");
-    });
-
-    it("3. Tc #1646, 60 - Honouring the order of execution & Bug 13826 + Bug 13646", () => {
-      homePage.NavigateToHome();
```
sneha122, line 253: Please check this test as well and if we really need to remove it

app/client/cypress/support/Pages/JSEditor.ts
```diff
@@ -332,26 +328,13 @@ export class JSEditor {
...
-    funName: string,
-    onLoad = true,
-    bfrCalling = true,
-  ) {
+  public EnableDisableAsyncFuncSettings(funName: string, onLoad = true) {
     // Navigate to Settings tab
     this.agHelper.GetNClick(this._settingsTab);
     // Set onPageLoad
     this.agHelper.CheckUncheck(this._onPageLoadSwitch(funName), onLoad);
-    // Set confirmBeforeExecute
```
sneha122, line 350: @rishabhrathod01 Should we add a test for deprecation warning implementation which asserts backwards compatibility? I know we will eventually have to remove this test when we deprecate it, but I was wondering if we can add a test till we deprecate, what do you think?

app/client/cypress/e2e/Regression/ServerSide/OnLoadTests/JSOnLoad2_Spec.ts
rishabhrathod01: This test is dependent on the previous test where we checked for confirmation dialog. I could also try to merge both of them and only remove confirm before calling related checks.

app/client/cypress/e2e/Regression/ServerSide/OnLoadTests/JsOnLoad3_Spec.ts
rishabhrathod01: The line `apiPage.ToggleConfirmBeforeRunning(true);` will fail in this test as there won't be an option to enable confirm before calling and the same is happening later too in the same test case. `jsEditor.EnableDisableAsyncFuncSettings("getId", false, true);`

app/client/cypress/support/Pages/JSEditor.ts
rishabhrathod01: @sneha122 For backward compatibility we have a test in `JSonLoad2_spec` where imported application has confirm before calling enabled and we assert confirmation dialog. Do we need any other case to be covered for backward compatibility?

sneha122: No I think this should be enough, thanks @rishabhrathod01

app/client/cypress/e2e/Regression/ServerSide/OnLoadTests/JSOnLoad2_Spec.ts
sneha122: Yeah my only concern here was that, we are removing these tests but if there is something else that we are asserting will get removed too and will result in less coverage for js objects

rishabhrathod01:
@sneha122 

> I see this Log in to comment which looks irrelevant to this PR but I did not see this on release or prod but only on DP (Looks like it was only there for release-ee DP, we can prolly ignore it)

This is a vercel feature on deploy preview to add comments once we log in. For creating DP pointing to release we use vercel hence this popup is visible.

app/client/cypress/e2e/Regression/ServerSide/OnLoadTests/JSOnLoad2_Spec.ts
rishabhrathod01: Yeah, i will merge this cases and make sure `Verify the sequence of of JS Object on page load` is still asserted.

sneha122: Thank you!

sneha122:
> This is a vercel feature on deploy preview to add comments once we log in. For creating DP pointing to release we use vercel hence this popup is visible.

Got it



> I would like to keep this UI consistent until we remove this feature. Cases where even a single confirm before calling option is enabled we will see the column and position would move for both of the cases. IMO, it is better to keep the position consistent as few JSObject settings shouldn’t look different from others for users. Thoughts?

Agree with you, let's do that once we remove this feature completely, but in meantime can we please confirm the UI with someone from design team? IMO this is a small UI issue but still a quality issue, so let's confirm with design team once



> Any idea, Where shall we show the warning message?

Got it for now, we can probably show a warning toast message to just let the user know why it disappeared

sneha122 (approved):

app/client/cypress/e2e/Regression/ServerSide/Datasources/Oracle_Spec.ts
```diff
@@ -446,11 +445,7 @@ WHERE aircraft_type = 'Passenger Plane'`;
     propPane.EnterJSContext("onClick", `{{Query1.run()}}`);
     deployMode.DeployApp(locators._widgetInDeployed(draggableWidgets.TABLE));
     agHelper.ClickButton("Submit");
-    jsEditor.ConfirmationClick("No"); //Handling both No & Yes from confirmation dialog
```
trishaanand, line 449: @rishabhrathod01 @sneha122 : Aren't these lines asserting if the confirm before calling feature of the API/Query is working? This wasn't about JS Object's feature. Can you please confirm if this was the correct change?

app/client/cypress/e2e/Regression/ServerSide/JsFunctionExecution/JSFunctionExecution_spec.ts
```diff
@@ -411,12 +411,6 @@ describe("JS Function Execution", { tags: ["@tag.JS"] }, function () {
 
     agHelper.RefreshPage();
     agHelper.Sleep(2000); //for confirmatiom modal to appear before clicking on "Yes" button for CI runs
```
trishaanand, line 413: @rishabhrathod01 @sneha122 : Is there a reason to leave this sleep behind considering the static wait for just for confirm before execute dialog interaction?

app/client/cypress/e2e/Regression/ServerSide/JsFunctionExecution/JSFunctionExecution_spec.ts
```diff
@@ -461,12 +455,7 @@ describe("JS Function Execution", { tags: ["@tag.JS"] }, function () {
     agHelper.Sleep();
     agHelper.WaitUntilAllToastsDisappear();
     agHelper.Sleep();
```
trishaanand, line 457: @rishabhrathod01 @sneha122 : Same comment as above.

app/client/cypress/e2e/Regression/ServerSide/OnLoadTests/JSOnLoad2_Spec.ts
```diff
@@ -119,8 +118,8 @@ describe(
         },
       );
 
-      jsEditor.EnableDisableAsyncFuncSettings("astros", true, true);
-      jsEditor.EnableDisableAsyncFuncSettings("city", true, true);
+      jsEditor.EnableDisableAsyncFuncSettings("astros", true);
```
trishaanand, line 121: @rishabhrathod01 @sneha122 : Shouldn't we just remove this test case entirely considering the test case is `6. Tc #1910 - Verify the Number of confirmation models of JS Object on page load` and we have removed this feature itself here in the PR?

app/client/cypress/e2e/Regression/ServerSide/OnLoadTests/JSOnLoad2_Spec.ts
```diff
@@ -138,8 +137,8 @@ describe(
         },
       );
 
-      jsEditor.EnableDisableAsyncFuncSettings("cat", true, true);
-      jsEditor.EnableDisableAsyncFuncSettings("hogwartsstudents", true, true);
+      jsEditor.EnableDisableAsyncFuncSettings("cat", true);
```
trishaanand, line 140: @rishabhrathod01 @sneha122 : Same as above. Instead of updating the test case, it looks like the test case itself should be removed.

app/client/cypress/e2e/Regression/ServerSide/OnLoadTests/JSOnLoad4_Spec.ts
```diff
@@ -59,19 +65,16 @@ describe(
...
       cy.get(jsEditor._funcDropdown).click();
       cy.get(jsEditor._funcDropdownOptions).then(function ($ele) {
-        expect($ele.eq(0).text()).to.be.oneOf(syncFunctions);
-        expect($ele.eq(1).text()).to.be.oneOf(asyncFunctions);
-        expect($ele.eq(2).text()).to.be.oneOf(asyncFunctions);
-        expect($ele.eq(3).text()).to.be.oneOf(asyncFunctions);
-        expect($ele.eq(4).text()).to.be.oneOf(asyncFunctions);
-        expect($ele.eq(5).text()).to.be.oneOf(asyncFunctions);
-        expect($ele.eq(6).text()).to.be.oneOf(syncFunctions);
+        expect($ele.eq(0).text()).to.be.oneOf(functions);
```
trishaanand, line 71: @rishabhrathod01 @sneha122 : This functions seemed to be asserting whether a function belongs to async or sync list. We have removed that assertion and just are asserting for the presence. I am not sure of the relationship of this code change with removal of the confirm before execute feature removal. Can you shed some light on this? Thanks!

app/client/cypress/e2e/Regression/ServerSide/OnLoadTests/JsOnLoad3_Spec.ts
```diff
@@ -118,64 +102,13 @@ describe(
...
-        agHelper.AssertContains("cancelled"); //callTrump
-
-        // //Another for API called via JS callQuotes()
-        // agHelper.AssertElementVisibility(jsEditor._dialogBody("Quotes"));
-        // jsEditor.ConfirmationClick("No");
-        //agHelper.WaitUntilToastDisappear('The action "Quotes" has failed');No toast appears!
-
-        agHelper.AssertElementAbsence(jsEditor._dialogBody("WhatTrumpThinks")); //Since JS call is NO, dependent API confirmation should not appear
-
         agHelper.RefreshPage("getConsolidatedData");
```
trishaanand, line 107: @rishabhrathod01 @sneha122  : Any reason to refresh the page considering the last line is to deploy the app. We should already have the page open at this point right?

app/client/cypress/e2e/Regression/ServerSide/OnLoadTests/JsOnLoad3_Spec.ts
```diff
@@ -118,64 +102,13 @@ describe(
...
         // agHelper.AssertElementVisibility(
```
trishaanand, line 108: @rishabhrathod01 @sneha122  : Could we please remove this commented code?

app/client/cypress/e2e/Regression/ServerSide/OnLoadTests/JsOnLoad3_Spec.ts
```diff
@@ -118,64 +102,13 @@ describe(
...
 
-        //agHelper.AssertElementVisibility(jsEditor._dialogBody("Quotes"));
-        agHelper.AssertElementExist(jsEditor._dialogInDeployView);
-        jsEditor.ConfirmationClick("Yes"); //quotes - api
-
-        //agHelper.Sleep(2000);
-        //agHelper.AssertElementVisibility(jsEditor._dialogBody("Quotes"));
-        //agHelper.AssertElementExist(jsEditor._dialogInDeployView);
-        //agHelper.GetNClick(".ads-v2-button__content-children", 1, true);
         agHelper.Sleep(4000); //to let the api's call be finished & populate the text fields before validation!
```
trishaanand, line 112: @rishabhrathod01 @sneha122  : Any reason to leave a static wait behind considering the app has already been deployed and page loaded.

app/client/cypress/support/Pages/JSEditor.ts
```diff
@@ -319,11 +319,7 @@ export class JSEditor {
   //   this.agHelper.GetNClick(this._closeSettings)
   // }
 
-  public VerifyAsyncFuncSettings(
-    funName: string,
-    onLoad = true,
-    bfrCalling = true,
-  ) {
+  public VerifyAsyncFuncSettings(funName: string, onLoad = true) {
     // this.agHelper.AssertExistingToggleState(this._functionSetting(Cypress.env("MESSAGES").JS_SETTINGS_ONPAGELOAD()), onLoad)
```
trishaanand, line 323: @rishabhrathod01 @sneha122  : Could we please get rid of these commented lines. Thanks!

app/client/cypress/e2e/Regression/ServerSide/OnLoadTests/JSOnLoad4_Spec.ts
rishabhrathod01: Hey @trishaanand, 

This was a refactor. We have removed the async and sync determination of function 6-7 months back but this test was not modified with that change. 

I was updating all the related cypress test hence updated this too.

trishaanand: Understood. Thanks a lot for the clarification.

app/client/cypress/e2e/Regression/ServerSide/OnLoadTests/JsOnLoad3_Spec.ts
rishabhrathod01: @trishaanand as the comment mention's this wait time is for the onPageLoad Api to get executed so that the data populates in the widget.

app/client/cypress/support/Pages/JSEditor.ts
rishabhrathod01: Sure, I will remove the commented code in a new PR.

app/client/cypress/e2e/Regression/ServerSide/OnLoadTests/JsOnLoad3_Spec.ts
rishabhrathod01: Sure. I will create a PR for this refactor.

app/client/cypress/e2e/Regression/ServerSide/OnLoadTests/JsOnLoad3_Spec.ts
rishabhrathod01: @trishaanand I will need to dig deeper here, i am not sure why was this present earlier. Let me check and get back on this one.

app/client/cypress/e2e/Regression/ServerSide/Datasources/Oracle_Spec.ts
rishabhrathod01: @trishaanand you are correct. This code was misleading as it uses jsEditor class method to assert ApiEditor confirmation. I will undo this and will use generic or specific class name here to avoid this confuction in future.

app/client/cypress/e2e/Regression/ServerSide/JsFunctionExecution/JSFunctionExecution_spec.ts
rishabhrathod01: I left it just to make sure settings tab is loaded after refresh before to try to navigate to it `agHelper.GetNClick(jsEditor._settingsTab);` here. I didn't experiment if this could be removed or not.

app/client/cypress/e2e/Regression/ServerSide/OnLoadTests/JSOnLoad2_Spec.ts
rishabhrathod01: The assertion are removed here but the setup was required for next test case assertion, i should remove this case and move the setup logic to next test case itself. My miss on that side.

app/client/cypress/e2e/Regression/ServerSide/Datasources/Oracle_Spec.ts
sneha122: Thanks @trishaanand for catching these 🚀, Missed during the review because I mainly relied on the fact that since tests are passing in CI, we should be okay with the changes, I will be more diligent about test cases review too from now onwards.

app/client/cypress/e2e/Regression/ServerSide/OnLoadTests/JsOnLoad3_Spec.ts
rishabhrathod01: @vsvamsi1 could you please check if we could skip the refresh page that was added here.  https://github.com/appsmithorg/appsmith/pull/30506/files#diff-25b467f92a60af5a65ad19e00f03aca3e9aa2c9ee324b4744cf15d1e233473e7

app/client/cypress/e2e/Regression/ServerSide/Datasources/Oracle_Spec.ts
rishabhrathod01: Adding the check back in this PR https://github.com/appsmithorg/appsmith/pull/32957

PR #32572, opened 2024-04-10
feat: feature branch for server side autocommit

### LINKED ISSUE #32888: [Epic] Server side auto-commit

sondermanish:
## Description
- Feature branch for server side AutoCommit.
Fixes #32888 
## Automation

/ok-to-test tags="@tag.Git"

## Communication
- [ ] Yes
- [x] No

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9213172317>
> Commit: de1635016890f4a3f0e32130c307562fa7091f3f
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9213172317&attempt=1" target="_blank">Click here!</a>


app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ce/GitAutoCommitHelperCECompatibleImpl.java
```diff
@@ -0,0 +1,24 @@
...
+        return Mono.just(Boolean.FALSE);
+    }
+
+    @Override
+    public Mono<Boolean> autoCommitServerMigration(String defaultApplicationId, String branchName) {
+        return Mono.just(Boolean.FALSE);
+    }
+
+    @Override
+    public Mono<Boolean> isServerAutocommitRequired(String workspaceId, GitArtifactMetadata gitMetadata) {
```
nidhi-nair, line 21: Nit on nomenclature, we seem to be using different casing for "autoCommit" and "autocommit"

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationPageServiceCEImpl.java
```diff
@@ -352,28 +354,53 @@ private Mono<Boolean> migrateSchemasForGitConnectedApps(Application application,
 
         // since this method is only called when the app is in edit mode
         Mono<PageDTO> pageDTOMono = getPage(newPages.get(0), false);
-        return pageDTOMono
+
+        Mono<Boolean> isClientAutocommitRequiredMono = pageDTOMono
```
nidhi-nair, line 358: The flows here are merging then diverging for their own separate logic. This violates interface segregation, please consider making it so that "autocommitApplication" only does that, and the server and client specific methods manage logic that would be unique to them. You can break those down into parameterized beans of their own as well, as long as the responsibility for defining specific logic is taken away from the generic method.

sondermanish: Indeed there could be segregation in the interface. However IMO that should be consumed by ApplicationPageService and should be separate from the AutoCommitPart. Because as per flow the responsibility of triggering an autocommit process should be on the ApplicationPage itself after it validates whether their is a requirement, the validation can be deferred to interface in focus, this would also adhere to the single responsibility of methods as well.
Creating a task for this to as well.

nidhi-nair (approved):

PR #32581, opened 2024-04-10
test micro diff

vsvamsi1:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  


app/client/src/workers/common/DataTreeEvaluator/index.ts
```diff
@@ -501,20 +502,41 @@ export default class DataTreeEvaluator {
...
+        microDiff(oldUnEvalTreeJSCollections, localUnEvalTreeJSCollection) ||
+        [],
     );
+    const jsDifferences: Diff<
+      Record<string, JSActionEntity>,
+      Record<string, JSActionEntity>
+    >[] = microDiffDifferences.map((v: Record<string, any>) => {
+      const { oldValue, path, type, value } = v;
+      //convert microDiff format to deepDiff format
+      if (type === "CREATE") {
```
rajatagrawal, line 519: would it be possible to extract this into a helper function ?

rishabhrathod01: Agreed, we should extract this to a separate method.

rajatagrawal (approved):

PR #32592, opened 2024-04-11
fix: default port numbers for the datasources

### LINKED ISSUE #32136: Port in datasources form should be set by default
https://www.notion.so/Port-in-datasources-form-should-be-set-by-default-cf1c30e7fb0d4142b95eaa73206a4ff1

AmanAgarwal041:
## Description
This PR covers `arango`, `elasticSearch`, `mongo`, `mssql`, `mysql`, `oracle`, `postgres`, `redis`, `redshift`, `smtp` datasources for setting the default port values for each. This adds a test case that tests the default port values for all

Fixes #32136 
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8687578779>
> Commit: d4b873c51761269acd5a13670d6c7beb1c0f23b9
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8687578779&attempt=1" target="_blank">Click here!</a>

## Summary by CodeRabbit

- **New Features**
	- Set default port values for various database plugins to streamline datasource configurations.

- **Tests**
	- Enhanced Cypress test suite to verify default port settings across multiple database plugins.
	- Updated test scenarios to reflect correct validation messages and error handling in datasource configurations.


rohan-arthur:
@AmanAgarwal041 since we have hard-coded default ports now on the backend and frontend, as an improvement we should plan(in this or another issue) to centralize this hard-coding to the backend and retrieve it to the frontend. Better for future maintainability.

AmanAgarwal041:
> @AmanAgarwal041 since we have hard-coded default ports now on the backend and frontend, as an improvement we should plan(in this or another issue) to centralize this hard-coding to the backend and retrieve it to the frontend. Better for future maintainability.

@rohan-arthur I don't think we need another file to keep the ports, already `form.json` for each plugins have their own configs which has the default ports, not anywhere else.

app/client/cypress/e2e/Sanity/Datasources/Default_Port_Number_Spec.ts
```diff
@@ -0,0 +1,119 @@
+import { agHelper, dataSources } from "../../../support/Objects/ObjectsCore";
+
+describe(
```
mohanarpit, line 3: @AmanAgarwal041 Do you think this test should be a different spec? Or should we add these assertions to existing specs like `Regression/ServerSide/Datasources/Oracle_spec.ts`?

AmanAgarwal041: @mohanarpit I would prefer all the default ports to be at once place, as we don't single spec file for some plugins like `mysql` or `postgres` etc. Any thoughts if this is in the `default_Port_Number_Spec.ts` ?

mohanarpit: The question I find myself asking: Why is asserting the default port number different from us asserting other fields in the datasource form? 

Yes, we will have to add this assertion in different spec files `Oracle_spec.ts`, `MySQL_spec.ts`, etc. But it ensures long term readability and maintainability of the suite.

AmanAgarwal041: Makes sense. I will update the tests in the specific test files. Thanks for the suggestion @mohanarpit

AmanAgarwal041: Done

mohanarpit: Sorry, I just discovered `Port_Number_Placeholder_Spec.ts`. Is that spec still relevant with the changes you've made to the tests here?

AmanAgarwal041: Yeah the `Port_Number_Placeholder_Spec` makes sense to keep it for now as, if anybody adds more `host` and `port` values for any datasource, then for them the default values will not be present but the placeholder should be present. Should we move this as well to individual datasource specs ?

mohanarpit: If possible, yes please. It'll be great if we can move the assertions from that file to the individual specs so that we can delete the spec file `Port_Number_Placeholder_Spec` completely. Let's leave the code slightly better than we found it. 

To be clear, I'm not suggesting we add another `it` block to assert the placeholder value. We should do this assertion in the same `it` block as the default value. Makes sense?

app/client/cypress/e2e/Regression/ServerSide/Datasources/Oracle_Spec.ts
```diff
@@ -450,6 +450,13 @@ WHERE aircraft_type = 'Passenger Plane'`;
     deployMode.NavigateBacktoEditor();
   });
 
+  it("8. Verify the default port for the datasource", function () {
```
mohanarpit, line 453: The other question I had was, is it necessary for this to be a different `it` block or do you think it should be in the same `it` block as TC 1. I'm asking this because I feel that we don't need to add a whole new integration test to assert the default/placeholder value. 

In line 37, we are already creating the plugin and we can just do the assertion over there in a cheap & quick fashion rather than having to navigate to the DS page, creating a new datasource and then asserting this value.

AmanAgarwal041: Sorry, didn't notice that. Thanks for the details. And it makes sense to move them in the 1st it block, will check for other datasources as well.

AmanAgarwal041: Done. Not valid for other datasources specs. Updated for Oracle.

sumitsum (dismissed):

sumitsum (approved):

Issue #32554, opened 2024-04-10
Introduce query_type field in actions API call


rohan-arthur:
https://www.notion.so/Introduce-query_type-field-in-actions-API-call-2c193b76a98742bd8611fdc2a7be06c8

sneha122:
Reverted the code and [PR](https://github.com/appsmithorg/appsmith/pull/32711) has been merged

PR #32593, opened 2024-04-11
feat: suggest queries query type added in action DTO

sneha122:
## Description
This PR adds query_type to actionDTO, when fetching queries, query type indicates whether it is a fetch query or any other query, this information can be further used to suggest queries to user when they bind data to widgets. This way we can suggest relevant queries to users first and their chances of success with Appsmith are higher.

Note: This PR only introduces the field in actionDTO and returns UNKNOWN, in subsequent PRs we will be adding plugin specific implementation to return relevant query_type. I have not added any unit tests as of now, we can add those once plugin specific implementations are done.

Fixes #`Issue Number`  
_or_  
Fixes [`Issue URL`](https://github.com/appsmithorg/appsmith/issues/32554)
> [!WARNING]  

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8643950238>
> Commit: 08e28bca9e4054ae78d4f4188cfecfc708a62608
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8643950238&attempt=1" target="_blank">Click here!</a>

### DIFF
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ce/ActionCE_DTO.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ce/ActionCE_DTO.java
@@ -1,6 +1,7 @@
 package com.appsmith.external.models.ce;
 
 import com.appsmith.external.constants.ActionCreationSourceTypeEnum;
+import com.appsmith.external.constants.DatasourceQueryType;
 import com.appsmith.external.dtos.DslExecutableDTO;
 import com.appsmith.external.dtos.LayoutExecutableUpdateDTO;
 import com.appsmith.external.exceptions.ErrorDTO;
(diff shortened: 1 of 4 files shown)


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ce/ActionCE_DTO.java
```diff
@@ -178,6 +179,10 @@ public class ActionCE_DTO implements Identifiable, Executable {
     @JsonView(Views.Public.class)
     ActionCreationSourceTypeEnum source;
 
+    @Transient
```
sumitsum, line 182: Hi Sneha, @Transient annotation may be redundant here because ActionCE_DTO object does not get persisted in the DB.

sneha122: @sumitsum I believe it gets saved in DB because if you see ActionDTO, it extends ActionCE_DTO model and ActionDTO model is used inside NewAction model in properties unpublishedAction and publishedAction, and this NewAction eventually gets saved in DB

PR #32606, opened 2024-04-11
chore: flaky git test fix

sondermanish:
## Description
- Fix for flaky git test
- Note: no requirement for ci to run

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Summary by CodeRabbit

- **Refactor**
	- Enhanced the clarity and maintainability of the test methods by splitting comparison logic into separate methods and introducing new methods for JSON object comparisons.

### DIFF
--- a/app/server/appsmith-server/src/test/java/com/appsmith/server/git/ServerSchemaMigrationEnforcerTest.java
+++ b/app/server/appsmith-server/src/test/java/com/appsmith/server/git/ServerSchemaMigrationEnforcerTest.java
@@ -13,6 +13,7 @@ import com.appsmith.server.helpers.PluginExecutorHelper;
 import com.appsmith.server.imports.internal.ImportService;
 import com.appsmith.server.services.WorkspaceService;
 import com.google.gson.Gson;
+import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import lombok.extern.slf4j.Slf4j;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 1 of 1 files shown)


app/server/appsmith-server/src/test/java/com/appsmith/server/git/ServerSchemaMigrationEnforcerTest.java
```diff
@@ -105,7 +106,24 @@ public void setup() {
...
+    /**
+     * Each entity in the map is a separate file in git file-system, it's imperative that we compare them separately
+     * If the comparison fails then it would essentially mean that users would see the diff,
+     * hence it should not be ignored
+     * @param target
+     * @param source
+     */
+    public void verifyMapAssertions(JsonObject target, JsonObject source) {
+        for (String key : source.keySet()) {
+            assertThat(convertElementToString(source.get(key))).isEqualTo(convertElementToString(target.get(key)));
```
nidhi-nair, line 118: Is there any possibility of nested jsons here that might get missed?

sondermanish: All entities getting converted to string over here are individual files in themselves.  if we see a diff, that should be a legit concern

sondermanish: Not sure if this is what you meant by nested jsons?

nidhi-nair: @sondermanish I actually meant to ask what happens if the type after we "get" the value from the given JsonObject is also a JsonObject, which could experience the same kind of unordered representation. Is that a possibility here?

sondermanish: The logic for serialisation is the same as what we use to serialise the app json to git fs.  Since we are comparing strings, that should be equal in all cases

sondermanish:
Since we have only worked on a test case and it's passing locally we can probably force merge this.

nidhi-nair (approved):

PR #32626, opened 2024-04-11
fix: JSON form splits keys to multiple lines when it should not

### LINKED ISSUE #32230: [Bug]: JSON form splits keys to multiple lines when it should not

### Description

When having a JSON from with some keys, it splits it to to lines for no reason.  i.e. "For Sale" is split to two lines even though it is wide enough to fit "Manufacture".  There appears to be no way to control this.

Appsmith support requested I open this bug in Discord.

### Steps To Reproduce

1. Create a new test app
2. Add a JSON form widget
3. set the source data to `{{{Manufacture: false, "For Sale": true}}}`
4. Notice that for sales is split into two lines

### Public Sample App

_No response_

### Environment

Production

### Severity

Low (Inconsistent UI/UX)

### Issue video log

_No response_

### Version

Self Hosted - 1.13

akshat8605:
Fixes #32230 
_or_  
Fixes https://github.com/appsmithorg/appsmith/issues/32230
> [!WARNING]  

## Automation

> [!CAUTION]  

## Summary by CodeRabbit

- **Style Updates**
	- Improved label display in JSON Form Widget to prevent text wrapping.
	- Adjusted the width setting of the Switch Widget for better layout consistency.


app/client/src/widgets/JSONFormWidget/component/FieldLabel.tsx
```diff
@@ -82,6 +82,7 @@ const StyledRequiredMarker = styled.div`
 
 const StyledLabelText = styled.p<StyledLabelTextProps>`
   margin-bottom: 0;
+  white-space: nowrap;
```
ashit-rath, line 85: @akshat8605 While this might solve the line break issue, it does not fundamentally solve the problem. If you check the screenshot; you would notice that due to the `nowrap` the line just extended and pushed out the switch input and ideally it should break to a new line. 
The use case for a long label like this would be a terms and conditions text which are usually really long.

The inherent problem is that the switch component is trying to take `100%` of the available width so there isn't enough space for the label to grow. So if we make the switch width `auto` and the label to flex 1 then the CSS engine will auto break lines as per the width available

Screenshot 2024-04-12 at 12 26 59 PM

PR #32627, opened 2024-04-11
fix: Checkbox type column on a table misalignment

### LINKED ISSUE #21790: [Bug]-[50]:Checkbox type column on a table widget does not center icon even when alignment is centered

### Description

We see the checkbox icon right aligned when column type is a checkbox and alignment is centered. Applicable for other columns types like icon, image etc...

### Steps To Reproduce

image
image

### Public Sample App

_No response_

### Environment

Production

### Issue video log

_No response_

### Version

Cloud

akshat8605:
Fixes #21790   
_or_  
Fixes `Issue URL`
> [!WARNING]  

Media : https://www.loom.com/share/ed1527330fc74c29ac1befb648aeb34f?sid=070dc600-ca1a-40d3-86bd-60767de0987e

## Automation

/ok-to-test tags="@tag.Table"

> [!CAUTION]  


app/client/src/widgets/CheckboxWidget/component/index.tsx
```diff
@@ -24,7 +24,7 @@ const CheckboxContainer = styled.div<StyledCheckboxContainerProps>`
     display: flex;
     height: 100%;
     justify-content: start;
-    width: 100%;
+    width: auto;
```
ashit-rath, line 27: @akshat8605 Let's not modify this directly from `100%` to `auto` as it may be some ripple effect since this component is used in multiple places. What I would recommend is to introduce a new prop called `isFullWidth` for the `CheckboxComponent` and pass this prop to this styled component `CheckboxContainer`. The `isFullWidth` is a optional boolean property whole default value would be true.

```css
width: ${({ isFullWidth }) => isFullWidth ? "100%" : "auto" }
```
Now use this newly added prop in `src/widgets/TableWidgetV2/component/cellComponents/CheckboxCell.tsx` where this component is used.

Let me know if this makes sense

akshat8605: @ashit-rath Sure, This makes sense if we look for for the edge case this will be helpful, I am implementing this.

ashit-rath: Apologies for the late reply, I was out for a few days.

If we make the default value as `false` then we would need to go and explicitly set the value of this prop to `true` in other places where `100%` is required. If we make the default value as `true` then existing implementation works as is and then the place where we want width `auto` we pass the prop `isFullWidth` as `false`. This would make the number of changes very less.

ashit-rath:
Hello @akshat8605, just checking in with you if you need any additional help as it's been a while since our last discussion.

akshat8605:
Hi @ashit-rath, I am on a vacation I will be back by tomorrow and will continue working!

PR #32650, opened 2024-04-12
feat: added updatedAt in actions

### LINKED ISSUE #32553: Make sure action object’s updated_at key has the correct information
We need to return the action updation time for all the actions in the consolidated api response.
This task involves validating and returning this information from the API so that it can be consumed by the client
Ref: https://www.notion.so/Make-sure-action-object-s-updated_at-key-has-the-correct-information-fd574e680ad14654b3ce304e19edbcc2

NilanshBansal:
The value of updatedAt will be consumed by the client to sort the actions based on the recently updated ones in the suggested fetch queries for widgets learnability task.

Fixes #32553 

## Automation

/ok-to-test tags="@tag.ImportExport, @tag.Datasource, @tag.Fork, @tag.Git"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8700455540>
> Commit: 45d7516ccf2b245bf93d66c767306732262ab131
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8700455540&attempt=2" target="_blank">Click here!</a>

## Summary by CodeRabbit

- **New Features**
	- Updated the display of the "Last Updated" timestamp with a specific format and timezone across the application.
- **Bug Fixes**
	- Ensured the "Last Updated" timestamp is accurately updated and displayed when actions are modified.
- **Tests**
	- Added and modified tests to verify the correct handling and display of the "Last Updated" timestamp for actions.


sumitsum (changes requested):
Hi Nilansh, can you please check if it is possible to add any JUnit TC for this ?

rohan-arthur:
test case idea from Sumit: "check that if new action object has updated_at field, it is getting propagated to action dto"

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ConsolidatedAPIServiceImplTest.java
```diff
@@ -565,6 +567,18 @@ public void testPageLoadResponseForEditMode() {
...
 
                     assertNotNull(consolidatedAPIResponseDTO.getPagesWithMigratedDsl());
+                    assertNotNull(consolidatedAPIResponseDTO.getUnpublishedActions());
+                    assertEquals(
+                            1,
+                            consolidatedAPIResponseDTO
+                                    .getUnpublishedActions()
+                                    .getData()
+                                    .size());
+                    assertNotNull(consolidatedAPIResponseDTO
```
sumitsum, line 577: Hi @NilanshBansal APMU, this check is not verifying the correct thing. What needs to be tested is this - is NewAction object has createdAt or updatedAt, is the same present in ActionDTO object as well ? i.e. one way to test it would be to write a test case for the `setTransientFieldsInUnpublishedAction` method .

NilanshBansal: @sumitsum Yes, that makes sense. While I pushed the code, I was also looking for other existing test cases where we could add this check. 
I have added a few tests in the ActionServiceCE_Test.java which will serve the same purpose. 
Can you please re-review now?

sumitsum (approved):

PR #32705, opened 2024-04-16
No domain on nested data classes

sharat87:


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/PEMCertificate.java
```diff
@@ -3,19 +3,13 @@
...
-import lombok.ToString;
 
-@Getter
-@Setter
-@ToString
-@EqualsAndHashCode
+@Data
 @NoArgsConstructor
 @AllArgsConstructor
-public class PEMCertificate implements AppsmithDomain {
```
nidhi-nair, line 18: This is needed to trigger `@Encrypted` processing I think, please check.

sharat87: Thanks for the tip. This sent me digging further, and it turns out we don't use `PEMCertificate` _at all_. So I'm deleting the whole class instead.

That said, can we _please_ rename `AppsmithDomain`? It appears its whole purpose is to mark the classes in which we should look for encrypted fields. The name is very misleading. Would it be okay if I renamed it and opened a PR?

nidhi-nair (approved):

PR #32738, opened 2024-04-17
chore: Move non-permission repository method to generic repo class

abhvsn:
## Description
PR to add separate record class to implement the projection on ApplicationSnapshot repository. 

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8735912421>
> Commit: 5baf4317e0bdb7179fe855c4fe108f9129a31f6b
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8735912421&attempt=1" target="_blank">Click here!</a>

## Summary by CodeRabbit

- **Refactor**
	- Enhanced the application snapshot feature to improve performance and reduce data load by excluding unnecessary data from snapshots.
- **Documentation**
	- Updated import statements and method return types for clarity and consistency.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationControllerCE.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationControllerCE.java
@@ -6,7 +6,6 @@ import com.appsmith.server.applications.base.ApplicationService;
 import com.appsmith.server.constants.FieldName;
 import com.appsmith.server.constants.Url;
 import com.appsmith.server.domains.Application;
-import com.appsmith.server.domains.ApplicationSnapshot;
 import com.appsmith.server.domains.GitAuth;
 import com.appsmith.server.domains.Theme;
 import com.appsmith.server.dtos.ApplicationAccessDTO;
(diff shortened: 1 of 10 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationControllerCE.java
```diff
@@ -232,7 +232,7 @@ public Mono<ResponseDTO<Boolean>> createSnapshot(
 
     @JsonView(Views.Public.class)
     @GetMapping("/snapshot/{id}")
-    public Mono<ResponseDTO<ApplicationSnapshot>> getSnapshotWithoutApplicationJson(
+    public Mono<ResponseDTO<ApplicationSnapshotProjectionWithoutData>> getSnapshotWithoutApplicationJson(
```
sharat87, line 235: Can we please come up with a better name for this please? What happens when we want another projection from the same domain class? This is too generic. 🙂

sharat87 (commented):
I don't have much beyond the name of the `record`, looks fine.

I dug a bit into this API call, since the projection class doesn't seem to be holding any real information at all (`applicationId` is an input already). Turns out the client is only looking at the `updatedAt` field and discarding the rest. I'll confirm this on EE as well and this projection will be even simpler. But that can happen as a separate PR. 🙂

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationControllerCE.java
abhvsn: Hey after your comment I also checked in client side and EE code, looks like only `updatedAt` is getting consumed on client and no usage in EE specific classes. I think instead of coming up with better name this is much easier to just update the record to `updatedAtOnly`, what say?

sharat87 (approved):

Issue #31745, opened 2024-03-13
[Task]: Remove `recentlyUsedWorkspaceIds` and `recentlyUsedAppIds` as no longer used


abhvsn:
### Description
After the homepage revamp project, we have updated the fields to store entities in [hierarchical order](https://github.com/appsmithorg/appsmith/blob/release/app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/ce/RecentlyUsedEntityCE_DTO.java). Because of this change we need to remove  `recentlyUsedWorkspaceIds` and `recentlyUsedAppIds` from the codebase which were used for the 1st version of homepage APIs.

PR #32756, opened 2024-04-18
chore: Deprecate recentlyUsedWorkspaceId and recentlyUsedAppIds from user data

abhvsn:
## Description
PR to remove usage of `recentlyUsedWorkspaceIds` and `recentlyUsedAppIds` from the codebase. We need to clear this debt as after the homepage revamp project, we have updated the fields to store entities in [separate DTO class](https://github.com/appsmithorg/appsmith/blob/release/app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/ce/RecentlyUsedEntityCE_DTO.java) instead of bare list of strings.

Fixes https://github.com/appsmithorg/appsmith/issues/31745

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8752116697>
> Commit: 0be31ca9c9a02030b215f4c5c19f188b2ed481d9
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8752116697&attempt=1" target="_blank">Click here!</a>

## Summary by CodeRabbit

- **New Features**
	- Introduced a new interface for managing user release notes.
	- Added functionality to handle removal of entities from recently used lists more efficiently.

- **Bug Fixes**
	- Corrected API call errors for deprecated methods in application controllers.

- **Refactor**
	- Updated various classes to use new user release notes management system.
	- Renamed methods to better reflect their functionality related to recent workspace and application usage.

- **Tests**
	- Updated unit tests to align with the new changes in application and user data management.


app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/ApplicationControllerCE.java
```diff
@@ -136,9 +136,7 @@ public Mono<ResponseDTO<Application>> delete(@PathVariable String id) {
     @GetMapping("/new")
     public Mono<ResponseDTO<UserHomepageDTO>> getAllApplicationsForHome() {
         log.debug("Going to get all applications grouped by workspace");
-        return applicationFetcher
-                .getAllApplications()
-                .map(applications -> new ResponseDTO<>(HttpStatus.OK.value(), applications, null));
+        return Mono.error(new AppsmithException(AppsmithError.DEPRECATED_API));
```
sharat87, line 139: Can we just remove this endpoint altogather? Then anyone calling it would just get a 404 error, instead of a 500, indicating better that this API is gone, rather than that there's been an error?

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomUserDataRepositoryCEImpl.java
```diff
@@ -18,18 +17,15 @@ public class CustomUserDataRepositoryCEImpl extends BaseAppsmithRepositoryImpl<U
     public Mono<Integer> saveReleaseNotesViewedVersion(String userId, String version) {
         return queryBuilder()
                 .criteria(Bridge.equal(UserData.Fields.userId, userId))
-                .updateFirst(Update.update(UserData.Fields.releaseNotesViewedVersion, version));
+                .updateFirst(new BridgeUpdate().set(UserData.Fields.releaseNotesViewedVersion, version));
```
sharat87, line 20: Let's prefer `Bridge.update()` instead of calling the constructor. Let's reduce the constructor's visibility even, actually.

Why? API consistency with `Bridge.query()`, and having the `Bridge` class be the single API entrypoint for all things we want to do with `Bridge`.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/UserReleaseNotesCEImpl.java
```diff
@@ -0,0 +1,70 @@
...
+                .cache();
+
+        return userMono.flatMap(user -> Mono.zip(
+                        Mono.just(user),
+                        releaseNotesService
+                                .getReleaseNodes()
+                                // In case of an error or empty response from CS Server, continue without this data.
+                                .onErrorResume(error -> Mono.empty())
+                                .defaultIfEmpty(Collections.emptyList()),
+                        userDataMono))
```
sharat87, line 54: This is the same thing as a direct `Mono.zip`, no?

```suggestion
        return Mono.zip(
                        userMono,
                        releaseNotesService
                                .getReleaseNodes()
                                // In case of an error or empty response from CS Server, continue without this data.
                                .onErrorResume(error -> Mono.empty())
                                .defaultIfEmpty(Collections.emptyList()),
                        userDataMono)
```

sharat87 (changes requested):
Minor things, but looks good. Thanks for cleaning up. ❤️

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/UserReleaseNotesCEImpl.java
abhvsn: Yeah, it was plain copy-paste so didn't pay attention here. Updating the code, thanks for catching 👍🏽

sharat87 (approved):

PR #32767, opened 2024-04-18
chore: Introduce IdPoliciesOnly record to handle projections

abhvsn:
## Description
PR to add IdPolicies record to make it easier to translate existing queries using the projections to JPA.

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8736013169>
> Commit: c2a024129eb899057db4dc1598515b90be567076
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8736013169&attempt=1" target="_blank">Click here!</a>

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomDatasourceRepositoryCE.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomDatasourceRepositoryCE.java
@@ -6,14 +6,9 @@ import com.appsmith.server.repositories.AppsmithRepository;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
-import java.util.List;
-import java.util.Set;
-
 public interface CustomDatasourceRepositoryCE extends AppsmithRepository<Datasource> {
 
     Flux<Datasource> findAllByWorkspaceId(String workspaceId, AclPermission permission);
 
     Mono<Datasource> findByNameAndWorkspaceId(String name, String workspaceId, AclPermission aclPermission);
-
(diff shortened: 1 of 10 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomDatasourceRepositoryCE.java
```diff
@@ -6,14 +6,9 @@
...
-import java.util.List;
-import java.util.Set;
-
 public interface CustomDatasourceRepositoryCE extends AppsmithRepository<Datasource> {
 
     Flux<Datasource> findAllByWorkspaceId(String workspaceId, AclPermission permission);
 
     Mono<Datasource> findByNameAndWorkspaceId(String name, String workspaceId, AclPermission aclPermission);
-
-    Flux<Datasource> findAllByIdsWithoutPermission(Set<String> ids, List<String> includeFields);
```
abhvsn, line 18: Removing this as we are moving towards generic query as we are not checking permissions. Also checked this method is not being used in EE.

sharat87 (approved):
This is an awesome refactor. Looks like we've always been passing the same value for the `includedFields` parameter, so a projection makes much more sense. Just checking, Is this true in the EE code as well? Like, we're always passing this same value to this parameter?

abhvsn:
> This is an awesome refactor. Looks like we've always been passing the same value for the `includedFields` parameter, so a projection makes much more sense. Just checking, Is this true in the EE code as well? Like, we're always passing this same value to this parameter?

I have gone through but did not find similar fields getting used on EE, but I'll take this up once again after I'm done with CE repo.

PR #32788, opened 2024-04-18
chore: Page rename API client validation

sharat87:
1. The page names can accept any character except for `/`, backslash and `:`. This is the validation that the server does now. This PR fixes the regex that checks the page name to align with this.
2. When typing a character that's not allowed, we end up with a space at the beginning or end of the page name, that doesn't really make sense, and is confusing. Then we end up with trailing or leading spaces in the page name, making the page name display look misaligned. This PR will make it so that the disallowed characters just can't be typed, and just don't make a difference in the page name.
3. We're also adding server-side validation for the icon slug.

[Slack conversation](https://theappsmith.slack.com/archives/C03RPDB936Z/p1713161688607509).

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8746416301>
> Commit: 21e8d3a0060b9b4721b49149f307354b073332d7
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8746416301&attempt=1" target="_blank">Click here!</a>

## Summary by CodeRabbit

- **New Features**
  - Enhanced page name validation to improve naming consistency across the application.
  - Updated validation for page icons and custom slugs to allow more flexibility and ensure inputs meet new standards.

- **Bug Fixes**
  - Fixed issues in page naming functions to prevent the use of certain special characters and limit the length, enhancing data integrity and user experience.

- **Tests**
  - Added new tests to verify the handling of invalid and empty custom slugs, ensuring robustness in page management functionalities.

### DIFF
--- a/app/client/src/utils/helpers.tsx
+++ b/app/client/src/utils/helpers.tsx
@@ -352,18 +352,9 @@ function getWidgetElementToScroll(
   }
 }
 
-export const resolveAsSpaceChar = (value: string, limit?: number) => {
-  // ensures that all special characters are disallowed
-  // while allowing all utf-8 characters
-  const removeSpecialCharsRegex =
-    /`|\~|\!|\@|\#|\$|\%|\^|\&|\*|\(|\)|\+|\=|\[|\{|\]|\}|\||\\|\'|\<|\,|\.|\>|\?|\/|\""|\;|\:|\s/;
-  const duplicateSpaceRegex = /\s+/;
-  return value
-    .split(removeSpecialCharsRegex)
-    .join(" ")
-    .split(duplicateSpaceRegex)
-    .join(" ")
-    .slice(0, limit || 30);
+export const toValidPageName = (value: string) => {
+  // Ensure that `/`, `\` and `:` are not allowed in page names, aligning with server-side validation.
+  return value.replaceAll(/[/\\:]+/g, "").slice(0, 30);
 };
 
 export const PLATFORM_OS = {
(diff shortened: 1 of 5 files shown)


app/client/src/utils/helpers.tsx
```diff
@@ -352,18 +352,9 @@ function getWidgetElementToScroll(
   }
 }
 
-export const resolveAsSpaceChar = (value: string, limit?: number) => {
-  // ensures that all special characters are disallowed
-  // while allowing all utf-8 characters
-  const removeSpecialCharsRegex =
```
mohanarpit, line 358: @sharat87 Do we also want to handle these special characters in the page name? In the diff we are removing `/ , \ and :` but what about the other special characters?

sharat87: I don't think so. The server never had validation, and this client check is also likely recent. I saw a lot of pages on prod that have `.` in their names. There's even pages named `1. First step`, `2. Second Step`, pages with emoji, etc. The only dangerous characters for us are these, since they are the only ones that can affect a file/URL paths.

Besides, I'd rather not have client validation more strict than server validation. It gives off a false-sense of security.

sharat87: And for space characters, even there I'd say let be. If I want to put in two spaces between words in my page name, why am not allowed? That just seems so arbitrary and unnecessary.

mohanarpit (approved):

app/client/src/utils/helpers.tsx
riodeuno: Shall we `.trim()` this value as well? Or are we okay with prefix and suffix with space characters?

PR #32804, opened 2024-04-19
chore: Add mysql plugin tinyInt1isBit configuration

gzlinzihong:
create table test_tinyint_types (c_tinyint TINYINT, c_tinyInt1isBit TINYINT(1));
> At this time, the query through mysql-plugin will be treated as a Boolean value by default, but its values ​​are not only 0 and 1, but also 2, 3, etc.
> Therefore, when connecting to mysql-plugin, you can support specifying the tinyInt1isBit configuration item to solve this problem.

> When the tinyInt1isBit switch is turned on, the result of the query is a Boolean value 
image
image

> When the tinyInt1isBit switch is turned off, the result of the query is a number.
image
image

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
	- Added a new configuration option in MySQL Plugin to allow conversion of `tinyint1` to `bit`, enhancing data type handling.

- **Tests**
	- Updated tests to cover new scenarios for `tinyint1` data type conversions in MySQL Plugin.

### DIFF
--- a/app/server/appsmith-plugins/mysqlPlugin/src/main/java/com/external/utils/MySqlDatasourceUtils.java
+++ b/app/server/appsmith-plugins/mysqlPlugin/src/main/java/com/external/utils/MySqlDatasourceUtils.java
@@ -251,7 +251,7 @@ public class MySqlDatasourceUtils {
         ob = addSslOptionsToBuilder(datasourceConfiguration, ob);
 
         boolean tinyInt1isBit = true;
-        if (datasourceConfiguration.getProperties().size() > 2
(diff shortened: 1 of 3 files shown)


nidhi-nair:
Thanks for the suggestion here @gzlinzihong, super appreciate how you've already come up with a solution for a problem you faced. 

However, for use cases such as this we recommend that users apply transformations on top of the query responses to modify the data to be represented as required. Please let me know if you need help with setting up such a transformation, but in the mean time I'll be closing this PR.

PR #32818, opened 2024-04-19
chore: Add separate DTO class for projection and move CustomNewActionRepo method to NewActionRepo class

abhvsn:
## Description
PR to move to class based projection for NewActionRepo method instead of relying of list of fields. Also as the permissions were not required converted the existing method to JPA based query. 

## Automation

/ok-to-test tags="@tag.Sanity, @tag.Binding, @tag.Datasource, @tag.GenerateCRUD"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8782210714>
> Commit: 80e40e840989db48954aafed914a801fbba995aa
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8782210714&attempt=1" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
	- Enhanced data fetching capabilities for actions within applications, improving performance and efficiency.

- **Refactor**
	- Optimized data source validation process during application creation to ensure smoother user experiences.


app/server/appsmith-server/src/main/java/com/appsmith/server/projections/IdAndDatasourceIdNewActionView.java
```diff
@@ -0,0 +1,34 @@
+package com.appsmith.server.projections;
```
nidhi-nair, line 1: Do you think it makes sense to organize these projections in the domain packages, seeing as how the nomenclature here is meant to fit only in the context of those domains?

app/server/appsmith-server/src/main/java/com/appsmith/server/projections/IdAndDatasourceIdNewActionView.java
```diff
@@ -0,0 +1,34 @@
...
+ * return the appropriate data from the database.
+ * However, when you try to use a nested projection, Spring Data JPA cannot create a dynamic proxy for the nested
+ * projection interface. This is because the nested projection interface is not a top-level interface. As a result,
+ * Spring Data JPA cannot intercept calls to the getter methods on the nested projection interface and return the
+ * appropriate data from the database.
+ * To work around this limitation, you can define the nested projection as a static inner class of the top-level.
+ * </p>
+ */
+@Getter
+public class IdAndDatasourceIdNewActionView {
```
nidhi-nair, line 20: Is the idea to create a custom type for every combination of projection that we use?

app/server/appsmith-server/src/main/java/com/appsmith/server/projections/IdAndDatasourceIdNewActionView.java
abhvsn: Good idea!

app/server/appsmith-server/src/main/java/com/appsmith/server/projections/IdAndDatasourceIdNewActionView.java
abhvsn: Yeah we need to move to class based approach when it comes to projection going forward as that's the way projections are supported in JPA queries. 
Ref doc: https://docs.spring.io/spring-data/jpa/reference/repositories/projections.html

nidhi-nair (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/projections/IdAndDatasourceIdNewActionView.java
nidhi-nair: Thanks for the reference, helps with context. I've approved the PR so we go ahead with the move, but let's look back at this periodically as you guys cover more queries? I feel like this might end up creating a lot of DTOs to deal with, there should be a better way 🤔

PR #32845, opened 2024-04-22
Configurable init container images

### LINKED ISSUE #32844: [Feature]: Configurable init containers

### Summary

In line with other images in this chart, initContainer images should be configurable. Registry, repository and tag all hould be set separately. An imagePullPolicy would also be useful. 
To ensure backwards compatibility the redis.image and mongodb.image properties should still exist. But in a later major chart version, they should be removed, as these don't really solve anything.

Because right now you can't easily update. When you have a custom registry (with replicated images), you must check the statefulset.yaml file in the helm chart to see what the latest supported version is. To then copy that version and replace docker.io with your own registry. When registry, repository and tag are set in the values.yaml, you can overwrite the registry property and overcome this problem.

kvandenhoute:
## Description
This PR enables configurable init container images. 
In line with other images in this chart, this PR makes sure you can configure registry, repository and tag all separately.
It also adds a imagePullPolicy. 
To ensure backwards compatibility I left the redis.image and mongodb.image properties in there. But imho these should be removed in a future major chart version.

Fixes #32844

## Communication
- [ ] Yes
- [x] No

### DIFF
--- a/deploy/helm/README.md
+++ b/deploy/helm/README.md
@@ -81,6 +81,14 @@ The command uninstalls the release and removes all Kubernetes resources associat
 | `image.repository`	| Appsmith image repository		| `appsmith/appsmith-ce` 	 |
 | `image.tag`					| Appsmith image tag					| `latest` 										      |
 | `image.pullPolicy`	| Appsmith image pull policy	| `IfNotPresent` 							   |
+| `initContainer.redis.registry`		| Redis image registry			| `docker.io` 					  |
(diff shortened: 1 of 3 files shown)


kvandenhoute:
keep open

kvandenhoute:
@sharat87 @pratapaprasanna Possible to have a look at this? Anything you want me to change?

kvandenhoute:
@sharat87 @pratapaprasanna Possible to have a look at this? Anything you want me to change?

kvandenhoute:
bump

PR #32908, opened 2024-04-24
[ops][fix]: hardcode APPSMITH_GIT_ROOT value

pratapaprasanna:
Keeping this open to close exactly after 90 days of merging this [PR](https://github.com/appsmithorg/appsmith/pull/33355)


sharat87 (changes requested):
As discussed on Slack, can we change the solution to this please?

> If there's a custom value configured, we print a warning into the logs. Live with this for ~3months, and then we remove configurability altogather.

So let's just print out a warning message when a custom value for this env variable is defined. Essentially in an `else` block to the existing `if` block.

Something like:
> WARNING: It appears a custom value has been configured for `APPSMITH_GIT_ROOT`. This behaviour is deprecated and will soon be removed.

PR #32933, opened 2024-04-25
fix: collection id not updated in import block API response

AnaghHegde:
Fixes https://github.com/appsmithorg/appsmith/issues/32903

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8892306251>
> Commit: 6b2c1c8e637dc89eebf51d431f5e78f92ae9ac23
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8892306251&attempt=1" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **New Features**
    - Enhanced functionality to set collection IDs during partial imports.
- **Tests**
    - Improved testing for partial imports by adding new assertions to verify action names.


app/server/appsmith-server/src/test/java/com/appsmith/server/solutions/PartialImportServiceTest.java
```diff
@@ -513,6 +513,9 @@ public void testPartialImportWithBuildingBlock_nameClash_success() {
                                     .getLayouts()
                                     .get(0)
                                     .getDsl());
+                    BuildingBlockResponseDTO1.getOnPageLoadActions().forEach(action -> {
+                        assertThat(action.getName()).isIn("updateProductVariant1", "getProducts1", "updateProduct1");
+                    });
```
NilanshBansal, line 518: @AnaghHegde Is this the correct assertion for the above change? 
I believe we are setting the collectionId in the actual fix but here in the test case we are asserting the action name.

app/client/src/ce/api/ApplicationApi.tsx
```diff
@@ -270,7 +269,15 @@ export interface ImportBuildingBlockToApplicationRequest {
...
-  onPageLoadActions: Omit<ActionViewMode, "pageId">[];
+  onPageLoadActions: {
+    id: string;
+    name: string;
+    confirmBeforeExecute: boolean;
+    pluginType: string;
+    collectionId?: string;
+    timeoutInMillisecond: number;
+    jsonPathKeys: string[];
+  }[];
```
rahulbarwal, line 280: There should be an existing interface for this. @jacquesikot can we please double check?

app/client/src/sagas/BuildingBlockAdditionSagas.ts
```diff
@@ -66,24 +67,47 @@ function* runSingleAction(actionId: string) {
   yield take(ReduxActionTypes.RUN_ACTION_SUCCESS);
```
rahulbarwal, line 67: We should also consider run action failure here. @jacquesikot

app/client/src/sagas/BuildingBlockAdditionSagas.ts
```diff
@@ -155,16 +177,23 @@ export function* loadBuildingBlocksIntoApplication(
...
+            .map((action) => ({ id: action.id })),
+        );
+
+        yield runNewlyCreatedJSActions(
+          response.data.onPageLoadActions
+            .filter((action) => action.pluginType === PluginType.JS)
+            .map((action) => ({
+              collectionId: action.collectionId as string, // collectionId exists for JS actions only
+              actionId: action.id,
+            })),
```
rahulbarwal, line 196: We are unnecessarily iterating over the response twice. Just create 2 arrays here and iterate only once.

app/client/src/sagas/BuildingBlockAdditionSagas.ts
```diff
@@ -66,24 +67,47 @@ function* runSingleAction(actionId: string) {
...
+  }[],
+) {
   // Run each action sequentially. We have a max of 2-3 actions per building block.
   // If we run this in parallel, we will have a racing condition when multiple building blocks are drag and dropped quickly.
-  for (const action of newlyAddedActions) {
-    if (action.config.executeOnLoad) {
-      yield runSingleAction(action.config.id);
-    }
+  for (const action of actions) {
+    yield runSingleAction(action.id);
```
rahulbarwal, line 110: When we know that we only expect ids in the destination function, why do we have a middleman function dealing objects that contains just ids. 
Instead just pass ids as array.

rahulbarwal (changes requested):

app/server/appsmith-server/src/test/java/com/appsmith/server/solutions/PartialImportServiceTest.java
AnaghHegde: @NilanshBansal What you said is correct. I will make that change, good to have an extra check. But asserting name is also good enough to verify that the correct actions are sent after refactoring when the name collision happens.

NilanshBansal (approved):
BE changes looks good.

app/client/src/ce/api/ApplicationApi.tsx
jacquesikot: Already checked, no specific interface for this, it combines normal action and JS action

rahulbarwal (approved):

PR #32987, opened 2024-04-26
fix: NOOP spans when context is missing

### LINKED ISSUE #30721: [Bug]:When I git import old version of attached app into release - auto commit does not run, there is no system generated commit either

### Description

For the attached application, when I git import old version into release, the auto commit does not run, and there is no system generated commit for the same. 

### Steps To Reproduce

https://github.com/appsmithorg/appsmith/assets/101863839/1c9938e3-b63a-48e8-b21a-a618fc2d3f0e

Imported application [url](https://release.app.appsmith.com/app/autocommit3/oauth20-65b246611f7b9934024955f8?branch=master)

### Public Sample App

_No response_

### Environment

Release

### Issue video log

_No response_

### Version

Cloud

nidhi-nair:
## Description

When auto-commits are triggered, the execution starts with no MDC context established. The observation helper assumed that this context would always be present and would hence error out during such events. This PR handles scenarios when context is absent by skipping the span entirely. 

If we wish to log auto-commit related spans, we will need to take that up separately by figuring out which trace we'd like to attach the span to.

Fixes #30721 

## Automation

/ok-to-test tags="@tag.Git"

> [!CAUTION]  

## Communication
- [ ] Yes
- [x] No

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ObservationHelperImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ObservationHelperImpl.java
@@ -8,6 +8,7 @@ import io.micrometer.tracing.Tracer;
 import org.slf4j.MDC;
 import org.springframework.stereotype.Component;
 
+import java.util.Map;
 import java.util.Optional;
 
 import static com.appsmith.external.constants.MDCConstants.SPAN_ID;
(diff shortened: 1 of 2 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ObservationHelperImpl.java
```diff
@@ -26,14 +27,19 @@ public ObservationHelperImpl(Optional<Tracer> tracer, CommonConfig commonConfig)
 
     @Override
     public Span createSpan(String name) {
-        TraceContext traceContext = tracer.traceContextBuilder()
-                .traceId(MDC.getCopyOfContextMap().get(TRACE_ID))
-                .spanId(MDC.get(SPAN_ID))
-                .build();
+        Map<String, String> contextMap = MDC.getCopyOfContextMap();
+        if (contextMap != null && contextMap.containsKey(TRACE_ID) && contextMap.containsKey(SPAN_ID)) {
```
sondermanish, line 31: nit pick: let's flip the conditional?

sondermanish (approved):

Issue #21825, opened 2023-03-28
[Task]: Clean widget properties in the DSL


riodeuno:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SubTasks

- Create an architecture for the data flow for the widget properties and configuration.
- Remove properties from the DSL which do not need to be stored on the servers. -- This will reduce the size of the DSLs and make sure that we're differentiating between properties pertinent to the layout and widget functionality, and the widget related configs used in the UI; or only needed during runtime and can be loaded during widget registration
- The properties which we can avoid storing in the DSL are:

```
     name: string;
     iconSVG?: string;
     searchTags?: string[];
     features?: WidgetFeatures;
     isDeprecated?: boolean;
     replacement?: string;
     hideCard?: boolean;
     eagerRender?: boolean;
     needsHeightForContent?: boolean;
```

riodeuno:
cc @rajatagrawal @SatishGandham @sbalaji1192 @ashit-rath

riodeuno:
cc @hetunandu - This may be related to the DSL optimisation.

PR #33013, opened 2024-04-29
feat: On Canvas UI Colors and Widget API to consume these colors

### LINKED ISSUE #33075: [Task]: Widget Dev API to provide on-canvas UI colors
### SubTasks

Different widgets can have different colors in their on canvas ui like widget name component and borders
- [ ] Create an API where the widget developer can specify the colors to use for selected state and other on canvas ui configs.
- [ ] Add color variables for the new widget selection UI.

riodeuno:
## Description
- Create a new set of configurations that can be added to any widget that will configure the following:
   - Selection state colors
   - Focused state colors
   - If the parent widget can be toggled from the widget's widget name component.  
- Create a default configuration to use if widget hasn't specified the colors
- Add CSS color variables for the current selection and focused states
- Add configuration to zone, section and modal widgets. 

Fixes #33075

## Automation

/ok-to-test tags="@tag.Anvil"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8897891359>
> Commit: 85cdd6b0f1fcce6888cfa3e0c10159def7b2565e
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8897891359&attempt=1" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [x] No


riodeuno:
@riodeuno 
Note: Due to the fact that we've updated the main container resizer z-index. Let's confirm if this works in the final PR.

app/client/src/WidgetProvider/factory/index.tsx
```diff
@@ -114,6 +129,7 @@ class WidgetFactory {
       needsHeightForContent: config.needsHeightForContent,
       isSearchWildcard: config.isSearchWildcard,
       needsErrorInfo: !!config.needsErrorInfo,
+      onCanvasUI,
```
marks0351, line 132: I wonder why do we want it to be a part of widget config. since the config is used only by an editor component.

marks0351 (changes requested):

app/client/src/WidgetProvider/factory/index.tsx
```diff
@@ -79,7 +81,20 @@ class WidgetFactory {
   }
 
   private static configureWidget(widget: typeof BaseWidget) {
+    const defaultConfig: WidgetDefaultProps = widget.getDefaults();
     const config = widget.getConfig();
+
+    /**
+     * As this will make all layout system widgets have these properties.
+     * We're going to prioritise #21825.
```
riodeuno, line 89: @marks0351

app/client/src/WidgetProvider/factory/index.tsx
riodeuno: I'm not sure what your concern is. Widgets owning configurations specific to the editor seem like a perfectly fine approach.
If you can share an alternative, it will help me understand your viewpoint.

marks0351: since the scope of using the config is only on the editor component and also for Anvil. The current widget name and how it figures out the parent all this will not be relavants with other layout systems which leads me to belive that we will keep widget name as is in those systems. which is why I think we should just use the `getWidgetHierarchy` to get the hierarchy of a widget and just set the color system of the widget name component.

riodeuno: Yes. WidgetHierarchy is the most straightforward change.
However, there are multiple reasons for using an API approach.
- It may be used by UI modules for module widgets.
- It loosely coupled and keeps the ownership of configurations with the widgets.

The only con that I see is that it will store this configurations in the DSL, but for that we'll be picking up #21825 in the upcoming sprint.

marks0351: approving this. 
however I still think the change in color codes is not based on individual widgets but is clearly based on hierarchy which exclusively only makes sense in Anvil.

marks0351 (approved):

PR #33042, opened 2024-04-29
fix: Empty plugin ids in workspaces are breaking application load

nidhi-nair:
## Description
Some workspaces in seem to exist with plugins associated to them but no id reference. This PR fixes any attempt to read such plugins.

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved plugin mapping by excluding null values, enhancing data accuracy and reliability.


nerbos (approved):
Blind approver expert

nerbos:
If this fix really solves the problem or it's the final solution. Let's please add a test to avoid this regression from happening again!

PR #33081, opened 2024-04-30
fix: block API onPageLoad js actions id mismatch

AnaghHegde:
## Description
The refactored name were not used for the `dslExecutable` objects that were part of the json object. This is json object is used to fetch the onPageLoad actions, the name collision happens only in the database and not in the json object/file

Fixes https://github.com/appsmithorg/appsmith/issues/32903

## Automation

/ok-to-test tags="tag.Templates"

> [!CAUTION]
> 🔴 🔴 🔴 Some tests have failed.
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8905315835>
> Commit: 0dba45e0d26317ca95a3639fb7451c20e591dd05
> Cypress dashboard: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8905315835&attempt=1&selectiontype=test&testsstatus=failed&specsstatus=fail" target="_blank"> Click here!</a>
> The following are new failures, please fix them before merging the PR: <ol>
> <li>cypress/e2e/Regression/ServerSide/GenerateCRUD/MySQL2_Spec.ts </ol>
> To know the list of identified flaky tests - <a href="https://internal.appsmith.com/app/cypress-dashboard/identified-flaky-tests-65890b3c81d7400d08fa9ee3?branch=master" target="_blank">Refer here</a>

## Communication
- [ ] Yes
- [ ] No


app/server/appsmith-server/src/main/java/com/appsmith/server/imports/internal/partial/PartialImportServiceCEImpl.java
```diff
@@ -442,18 +442,22 @@ public Mono<BuildingBlockResponseDTO> importBuildingBlock(BuildingBlockDTO build
                                                             .getRefactoredEntityNameMap()
                                                             .get(dslExecutableDTO.getName())
                                                     != null) {
+                                                String name = dslExecutableDTO
```
nidhi-nair, line 445: @AnaghHegde could you please help me understand what this piece of code is attempting to do?

AnaghHegde: @nidhi-nair This is to get the actionCollection name from the function.

nidhi-nair: Sorry I should have probably phrased that better. We're getting the action collection name yes, but why exactly are we looking for the collection name as an identifier for on page load actions? I'm trying to understand the correlation between grouping these by collections versus treating them as flat actions directly.

AnaghHegde: When the blocks are imported in an existing page with queries widgets etc, there is a chance for name collision. When that happens in the backend flow we refactor the names of action and actionCollection and the DSL as well.
But the onPageLoadActions of the DSL that is yet to be added to the page will have the old name is is to update the correct names in that reference.

PR #33095, opened 2024-05-01
fix: Create module btn UI fix in split screen view

### LINKED ISSUE #33038: [Bug] - Create module button overlapping context menu icon

albinAppsmith:
## Description

Create Module (Beta) button was overlapping on context menu in split screen. This PR addresses this issue.

Fixes #33038

## Automation

/ok-to-test tags="@tag.Sanity, @tag.Workflows"

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No


app/client/src/pages/Editor/QueryEditor/index.tsx
```diff
@@ -89,37 +93,43 @@ function QueryEditor(props: QueryEditorProps) {
...
+      canCreateModuleInstance: isCreatePermitted,
+      canDeleteEntity: isDeletePermitted,
+      entityId: action?.id || "",
+      moduleType: MODULE_TYPE.QUERY,
+    };
+    return (
       <>
         <MoreActionsMenu
           className="t--more-action-menu"
+          convertToModuleProps={convertToModuleProps}
```
ashit-rath, line 107: Instead of passing the props here and then within <MoreActionsMenu> importing the ConvertTModuleInstanceCTA and using there, do you think we should pass the component itself as a prop and name it something like `additionalMenuItem`?


```jsx
// Example name
const convertCTA = isSideBySideEnabled && editorMode === EditorViewMode.SplitScreen 
? <ConvertToModuleInstanceCTA {...convertToModuleProps} />
: null

<MoreActionsMenu
          className="t--more-action-menu"
          convertToModuleProps={convertToModuleProps}
          additionalMenuItem={convertCTA}
          ... // other props
          />

{action?.pluginType !== PluginType.INTERNAL &&
       
            // Need to remove this check once workflow query is supported in module
            convertCTA
          }
```

This way the MoreActionsMenu doesn't need to be aware of a component to import and then apply props to it. What do you think?

ashit-rath: I think I wrote the conditions a bit incorrectly as this was a contrived example to show my thought process

ashit-rath (approved):

PR #33114, opened 2024-05-02
chore: Fix warnings in ConsolidatedAPIServiceImpl

sharat87:
This is an effort to reduce warnings from `javac`, towards adding a linter to the backend.

I recently had to troubleshoot some code in this class and the array based zipping of 10+ Monos was very hard to parse and debug. This PR simplifies that implementation as well as fixes all/most warnings in this class.

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8918809242>
> Commit: f37ef41d41b9925de233f10ca300f89bc3ffe044
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8918809242&attempt=2" target="_blank">Click here!</a>

### DIFF
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/annotations/encryption/EncryptionHandler.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/annotations/encryption/EncryptionHandler.java
@@ -9,6 +9,7 @@ import reactor.util.annotation.NonNull;
 import java.lang.reflect.Field;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
+import java.lang.reflect.WildcardType;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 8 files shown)


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/annotations/encryption/EncryptionHandler.java
```diff
@@ -171,6 +172,10 @@ List<CandidateField> findCandidateFieldsForType(@NonNull Object source) {
                             Type[] typeArguments;
                             ParameterizedType parameterizedType = (ParameterizedType) field.getGenericType();
                             typeArguments = parameterizedType.getActualTypeArguments();
+                            if (typeArguments[1] instanceof WildcardType) {
```
trishaanand, line 175: Is this an intended change? Do we need to add NPE checks before accessing the first element of the array?

sharat87: Yes, application startup fails without this. NPE checks won't be necessary. Since this block is for the `Map` type, it is guaranteed to have two type parameters, one for key type and one for value type.

The code here is looking into the value type of the `Map` field, to see if that could contain any `@Encrypted` fields. But when the `Map` is typed as `Map<?, ?>`, then the type arguments we would get here would be instances of an implementation of `WildcardType`. It doesn't make sense to look inside those to see if it can contain encrypted fields, so we just `return` when we see the wildcard type.

This needed to be introduced here since I'm changing some `Map` fields to `Map<?, ?>` in some entity classes.

trishaanand (approved):

PR #33121, opened 2024-05-02
fix: Filter fields with `@Transient` annotation embedded within jsonb column during DB serialisation

abhvsn:
### Description

As we are using `CustomJsonType` to serialise the jsonb column, it's not respecting the `@Transient` to remove these fields while saving the object to DB. To address this issue we need to implement the custom serialiser which uses reflections to filter fields based on annotation.

### DIFF
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/CustomJsonType.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/CustomJsonType.java
@@ -1,8 +1,11 @@
 package com.appsmith.external.helpers;
 
+import com.appsmith.external.converters.TransientAnnotationSerializer;
+import com.appsmith.external.models.ActionDTO;
 import com.appsmith.external.views.Views;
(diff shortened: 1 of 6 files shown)


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/CustomJsonType.java
```diff
@@ -13,14 +16,20 @@
...
+
     public CustomJsonType() {
         super(makeObjectMapperForDatabaseSerialization());
     }
 
     private static ObjectMapper makeObjectMapperForDatabaseSerialization() {
         final ObjectMapper om = new ObjectMapper();
 
+        SimpleModule module = new SimpleModule();
+        module.addSerializer(ActionDTO.class, serializer);
```
abhvsn, line 29: We can add more classes here as and when required. I have added only ActionDTO to fix `PageServiceTest.clonePage` which was failing because `unpublishedAction.id` was getting saved to DB.

sharat87 (approved):

PR #33123, opened 2024-05-02
fix: datasource documentation link (#30169)

### LINKED ISSUE #30169: [Bug]: Learn more documentation on create datasource takes us to basic application setup documentation page

### Description

We have this banner on create datasource page for whitelisting IP and we have a CTA to learn more, when clicked on learn more it takes me to [this doc page](https://docs.appsmith.com/getting-started/tutorials/start-building#connect-to-a-database), I was expecting this doc page to have info on either datasources or white listing the IP but instead this page has info on basics of application setup. We should fix this and make learn more CTA take us to respective datasource's documentation page: https://theappsmith.slack.com/archives/CGBPVEJ5C/p1704804571691129?thread_ts=1704803786.978579&cid=CGBPVEJ5C

Screenshot 2024-01-09 at 6 02 41 PM

### Steps To Reproduce

1. Create a datasource
2. Observe the banner that appears on top
3. Click on Learn More

### Public Sample App

_No response_

### Environment

Production

### Issue video log

_No response_

### Version

Cloud/ Self-Hosted - 1.8.4

NilanshBansal:
## Description
This PR fixes the link associated with `Learn More` button on the datasource page. Earlier, on clicking Learn More button, there was a hardcoded URL in the code, which redirected the users to documentation homepage. 
With this change, the user will be redirected to the specific plugin page as now the URL configuration is consumed from the database field `plugin.documentationLink`

Fixes #30169   

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8935729769>
> Commit: 03a263decf489c4349cafbd1263cfa60fdf35180
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8935729769&attempt=1" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [ ] No


NilanshBansal:
@sneha122 @rishabhrathod01 the dp will not show these buttons because of [this](https://github.com/appsmithorg/appsmith/blob/4b8e76a2bcccaab49fca43927f0464f55d26379f/app/client/src/pages/Editor/DataSourceEditor/DBForm.tsx#L87) LOC

app/client/src/pages/Editor/DataSourceEditor/DBForm.tsx
```diff
@@ -49,7 +50,11 @@ export const Form = styled.form<{
 
 class DatasourceDBEditor extends JSONtoForm<Props> {
   openDocumentation = () => {
-    openDoc(DocsLink.WHITELIST_IP);
+    store.getState().entities.plugins.list.forEach((plugin: any) => {
+      if (plugin.id === this.props.datasource.pluginId) {
+        openDoc(DocsLink.WHITELIST_IP, plugin?.documentationLink, plugin?.name);
+      }
+    });
```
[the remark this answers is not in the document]
NilanshBansal: This case will never occur because the datasource form will be rendered only when the datasource is present. The learn more button is present in the datasource form

app/client/src/pages/Editor/DataSourceEditor/DBForm.tsx
```diff
@@ -49,7 +50,11 @@ export const Form = styled.form<{
 
 class DatasourceDBEditor extends JSONtoForm<Props> {
   openDocumentation = () => {
-    openDoc(DocsLink.WHITELIST_IP);
+    store.getState().entities.plugins.list.forEach((plugin: any) => {
```
sneha122, line 53: Instead of calling store.getState() here, can we instead add plugins list from mapStateToProps and in this function, we can use it as props.plugins instead? Asking this because all the redux properties needs to be used using mapStateToProps interface which redux provides, this is being followed across frontend code, so here also let's make it consistent. Check this file https://github.com/appsmithorg/appsmith/blob/0dda55780a566f941360be82a085f258e5c8b132/app/client/src/pages/Editor/QueryEditor/Editor.tsx#L294 and https://github.com/appsmithorg/appsmith/blob/0dda55780a566f941360be82a085f258e5c8b132/app/client/src/pages/Editor/QueryEditor/Editor.tsx#L156

CC: @rishabhrathod01

sneha122, line 53: Instead of declaring type as any here, can we please use Plugin interface to define type here? https://github.com/appsmithorg/appsmith/blob/0dda55780a566f941360be82a085f258e5c8b132/app/client/src/api/PluginApi.ts#L22

https://github.com/appsmithorg/appsmith/blob/0dda55780a566f941360be82a085f258e5c8b132/app/client/src/sagas/DatasourcesSagas.ts#L334

app/client/src/pages/Editor/DataSourceEditor/DBForm.tsx
```diff
@@ -49,7 +50,11 @@ export const Form = styled.form<{
...
+      if (plugin.id === this.props.datasource.pluginId) {
```
sneha122, line 54: Instead of `this.props.datasource.pluginId`, can we please use `this.props.datasource?.pluginId`, this way if for whatever reason datasource comes up as undefined, the application will not fail and we will have a protection. For more on ? itself you can check this out https://www.freecodecamp.org/news/how-the-question-mark-works-in-javascript/

app/client/src/pages/Editor/DataSourceEditor/DBForm.tsx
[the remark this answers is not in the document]
sneha122: @NilanshBansal I agree here, it is better to add `.find` method to find the required plugin, rather than `forEach`. With find method as soon as plugin is found, it stops iterating over the array, but with forEach it will continue even though element has been found. Please consider using find method instead. You can modfiy code as follows:

```
const plugin: Plugin = plugins.find(plugin -> plugin.id === this.props.datasource?.pluginId);

If (!!plugin) {
    // openDoc link here
} else {
    // we can just log the error here for better debugging in case of any failure
}
```

sneha122:
@NilanshBansal Do you think we should write any automation test cases for this?

app/client/src/pages/Editor/DataSourceEditor/DBForm.tsx
rishabhrathod01: > Instead of calling store.getState() here, can we instead add plugins list from mapStateToProps and in this function, we can use it as props.plugins instead? 

@sneha122 I don't think we should use mapStateToProps for such cases when we need to read data from state for an instant instead of subscribing it for continuous updates to react upon.  

> Asking this because all the redux properties needs to be used using mapStateToProps interface which redux provides, this is being followed across frontend code, so here also let's make it consistent. 

That is not true, `mapStateToProps` or `useSelector` are meant for the cases when we won't to subscribe to the changes of the state but when we only want to read the data on a particular event using mapStateToProps will be an anti-pattern and will lead to unnecessary re-renders that could be avoided.

rishabhrathod01, line 53:
```suggestion
    const appState: AppState = store.getState();
    appState.entities.plugins.list.forEach((plugin) => {
```

rishabhrathod01: @NilanshBansal we could use AppState here for types, in that case, plugin type will be auto-referred and we won't need to have `(plugin: any)`

NilanshBansal:
> @NilanshBansal Do you think we should write any automation test cases for this?

@sneha122 I am not sure if we should write automation tests for this because, the documentation URL is picked from the database, and the documentation link can change in the future.
For such scenarios, if we change the documentation link from the database, we would also have to update the test cases i.e. code level changes will need to be done for a non-code related change. 
Also, going forward let's say we change the link from `https://docs.appsmith.com` to `https://appsmith-docs.com`, instead of doing any code-level or db changes, we can just add a redirection rule to our DNS settings without need for dev dependency. 
But there are pros and cons to either of the approach.
cc: @sneha122 @rishabhrathod01

rishabhrathod01:
> I am not sure if we should write automation tests for this because, the documentation URL is picked from the database, and the documentation link can change in the future.
For such scenarios, if we change the documentation link from the database, we would also have to update the test cases i.e. code level changes will need to be done for a non-code related change.
Also, going forward let's say we change the link from https://docs.appsmith.com/ to https://appsmith-docs.com/, instead of doing any code-level or db changes, we can just add a redirection rule to our DNS settings without need for dev dependency. But there are pros and cons to either approach. 

IMO we shouldn't test the user education routes in the product. A regression test suite could have these but E2E tests should skip this.

rishabhrathod01 (approved):
LGTM

app/client/src/pages/Editor/DataSourceEditor/DBForm.tsx
sneha122: Resolving this comment after extensive discussion with @rishabhrathod01. Thanks @rishabhrathod01 for clarifying this in detail. As discussed let's add a comment to explain the usage of `store.getState()`.

sneha122 (approved):

app/client/src/pages/Editor/DataSourceEditor/DBForm.tsx
rishabhrathod01: - `mapStateToProps` or `useSelector` are best for cases where we want the react component to re-render and show different UI according to the state change. 
- `store.getState()` are best for cases where we don't want the component to subscribe to change of the state like plugins in above case and instead only read the state on a particular event like onClick on docs link in this case. 

Using `store.getState()` helps us to avoid unnecessary re-renders of the component due to reduce state subscriptions.

PR #33135, opened 2024-05-02
test: Refactor tests ahead of upgrading Spring

sharat87:
Only refactors, only in tests, needed to get these tests to work/pass with the upgraded Spring coming up separately.

### DIFF
--- a/app/server/appsmith-git/src/test/java/com/appsmith/git/helpers/DSLTransformHelperTest.java
+++ b/app/server/appsmith-git/src/test/java/com/appsmith/git/helpers/DSLTransformHelperTest.java
@@ -7,8 +7,6 @@ import org.json.JSONObject;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.extension.ExtendWith;
-import org.springframework.test.context.junit.jupiter.SpringExtension;
 
 import java.io.File;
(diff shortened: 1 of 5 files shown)


nidhi-nair (commented):
Could you please explain what the changes are for exactly, or link to relevant docs?

sharat87:
@nidhi-nair, these tests were either failing to run, or failing to pass with Spring 3.2. I couldn't find anything specifc in release notes or documentation that points to what has changed, but I nevertheless fount them to be good changes to make to our tests anyway.

For example, sim tests with `@ExtendWith(SpringExtension.class)` that didn't need it were okay with Spring 3.0, but they're failing to run with Spring 3.2. I didn't delve into what change in that upgrade caused this, but removing that unused extension gets the test to run.

Another one, for example, in `ActionExecutionSolutionCEImplTest`, we're calling a mocked method _before_ we define the mock behaviour with `Mockito.when`. Again, this is okay with the current version of Mockito, but with what Spring 3.2 brings, we have to define the mock behaviour first. Which, in my opinion, is how it should be anyway.

If it makes sense to add details on the errors in the springup branch that needed these changes, I'll spend some time next week on that. Didn't see the value in that, but happy to do that if you see value.

nidhi-nair:
This is good enough I think, thanks for spending the time to add details! My intent was for us to recognize these rules that we need to be cognizant of moving forward, before the Spring upgrade comes in.

nidhi-nair (approved):

PR #33139, opened 2024-05-02
fix: Remove redundant action collection update call on page load in edit mode

dvj1988:
## Description
If there are any functions in a JS Object that accept an argument, an update action collection for the JavaScript object is triggered when the application first loads. These update calls alter the lastUpdatedAt time of a page merely by opening the application in edit mode, without making any changes.

In getDifferenceInJSCollection, we check if the arguments to the action (function in a JS object) are the same as those in the Redux store. This check is not required as we are already checking the equality of the body and arguments are also part of the body.

This check was always returning false because one of the object has an extra value key
```
x = [{ name: "argument1" }]
y = [{ name: "argument1" , value: undefined }]
```

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8967578757>
> Commit: 5b1cdb75986374be3469b22382c542bf123a8ead
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8967578757&attempt=2" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [x] No


vsvamsi1:
@dvj1988 Can you add some test cases to the case your trying to optimise.

dvj1988:
> @dvj1988 Can you add some test cases to the case your trying to optimise.

This method is already covered in the test cases defined in `app/client/src/utils/JSPaneUtils.test.ts`. The case this PR changes is also covered in the test case `gets updated params value in difference`.

Screenshot 2024-05-06 at 5 21 22 PM

PR #33157, opened 2024-05-03
chore: Disable temporal by default on startup

pratapaprasanna:


deploy/docker/fs/opt/appsmith/templates/docker.env.sh
```diff
@@ -92,5 +92,6 @@ APPSMITH_SUPERVISOR_PASSWORD=$SUPERVISOR_PASSWORD
 APPSMITH_ALLOWED_FRAME_ANCESTORS="'self' *"
 
 APPSMITH_DISABLE_IFRAME_WIDGET_SANDBOX=false
+APPSMITH_TEMPORAL_URL=disabled
```
mohanarpit, line 95: This is not the fix. We should fix the if-condition in the entrypoint.sh [here](https://github.com/appsmithorg/appsmith-ee/blob/release/deploy/docker/fs/opt/appsmith/entrypoint.sh#L480).

PR #33158, opened 2024-05-03
chore: flyway migration setup: create table and unique index migrations

vivonk:
## Description
Setup Flyway migration in appsmith and adding create table and unique index migration for all the entities 

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [x] No


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/ce/V1__createTablesAndRoutines.java
```diff
@@ -0,0 +1,21 @@
...
+ * by new migrations instead of editing this migration related SQL files.
+ */
+public class V1__createTablesAndRoutines extends AppsmithJavaMigration {
+
+    @Override
+    public void migrate(JdbcTemplate jdbcTemplate) throws Exception {
+        String createTablesSQL = loadSQLFileAsString("/sql/createTables.sql");
+        String createFunctionsSQL = loadSQLFileAsString("/sql/createRoutines.sql");
+        jdbcTemplate.execute(createTablesSQL);
+        jdbcTemplate.execute(createFunctionsSQL);
```
abhvsn, line 19: Sorry for the limited knowledge but what's the purpose of the above command?

app/server/appsmith-server/src/main/resources/application.properties
```diff
@@ -14,7 +14,7 @@ spring.main.allow-bean-definition-overriding=true
 spring.data.redis.repositories.enabled=false
 
 #debug=true
-spring.jpa.show-sql=true
+spring.jpa.show-sql=false
```
abhvsn, line 17: Should we keep this enabled as it helps to debug for a few more days, once we are able to interact with UI we can disable this what say?

abhvsn (commented):
Added couple of comments!

@vivonk let's think through on following pointer:
1. How to interlace CE and EE migrations?

app/server/appsmith-server/src/main/resources/application.properties
vivonk: ya works for me, I was using in local dev setup so set this for me personally. Will undo this

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/ce/V1__createTablesAndRoutines.java
vivonk: This will basically execute all the create tables sql

vivonk:
@abhvsn as it happens, all the CE migrations run first and then EE migrations. Similarly, we can name our EE migrations like 
VEE01__name, and CE migrations as V01__name. This was, EE migrations will always run after CE migrations because 01 comes first in order compare to EE01 when sorted. And flyway runs the migration in sorted order of migration version.
cc: @sharat87

abhvsn:
> @abhvsn as it happens, all the CE migrations run first and then EE migrations. Similarly, we can name our EE migrations like VEE01__name, and CE migrations as V01__name. This was, EE migrations will always run after CE migrations because 01 comes first in order compare to EE01 when sorted. And flyway runs the migration in sorted order of migration version. cc: @sharat87

I didn't find if `VEE01`__description is supported in the Flyway naming convention but if it is then I suppose we are sorted here. Also while looking into naming convention found a lot of issues from the users using flyway migration and looks like git hooks can be a good solution here https://flywaydb.org/blog/using-git-hooks-with-flyway.

abhvsn (approved):

PR #33165, opened 2024-05-03
chore: Migrations for adding all the plugins

vivonk:
## Description
Add all the appsmith plugins with latest state in the DB as migration

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [x] No

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/ce/V3__addPlugins.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/ce/V3__addPlugins.java
@@ -1,0 +1,506 @@
+package com.appsmith.server.migrations.ce;
+
+import com.appsmith.external.models.PluginType;
+import com.appsmith.server.domains.Plugin;
+import com.appsmith.server.migrations.AppsmithJavaMigration;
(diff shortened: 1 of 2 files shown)


abhvsn (approved):
LGTM

Tagging @sharat87 as we discussed around having a beans to add this data instead of migration. But I'm now more inclined towards going via migrations approach as this will provide us the standardised way to manage the database changes.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/ce/V3__addPlugins.java
```diff
@@ -65,6 +65,20 @@ private void insertPlugin(Plugin plugin) {
         } catch (DuplicateKeyException e) {
             log.warn("{} plugin already present in database.", plugin.getName());
         }
+        installPluginToAllWorkspaces(plugin);
+    }
+
+    public void installPluginToAllWorkspaces(Plugin plugin) {
```
vivonk, line 71: @abhvsn small change for installing plugins to all the workspaces

Issue #32970, opened 2024-04-25
[Feature]: Support rsa-sha2-256 and rsa-sha2-512


hud-21326:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

Add rsa-sha2-256 or rsa-sha2-512 as options when creating an SSH key in Appsmith for Git integration.

### Why should this be worked on?

Azure DevOps is [currently](https://devblogs.microsoft.com/devops/ssh-rsa-deprecation/) deprecating support for SSH-RSA keys. It also does not support ECDSA keys. Since those are the only two SSH key types that can be generated in Appsmith, this seems to break the integration between DevOps and Appsmith

lbustelo:
+1 We just started experiencing this issue today.

jhu8:
+1 We use Azure as our git provider and started experiencing issues with our git integration today as well.

nidhi-nair:
Sorry that you have had to face such a disruption so abruptly folks! We're looking into this and will keep you updated on progress.

amenasse:
Just wanted to get some clarification on the fix, #33052 seems to address a different problem - atomic commits not being supported by Azure Devops.

When generating keys and connecting does the appsmith git integration use one of the supported SSH signing algorithms mentioned in the initial ticket description ? We are seeing behaviour that could be related to the depreciation schedule mentioned in the article.

nidhi-nair:
Hey @amenasse , thanks for raising your concern. While investigating this problem we discovered that since we introduced atomic pushes, all of Azure operations were affected. The earlier PR was to address that blocker and unblock folks that were only intermittently affected by the brownout phase that Azure's deprecation is in.

We hadn't been able to reproduce and hence successfully test our solution for moving on the SHA2 signing algorithms until very recently, but that's affected our test repositories as well now, so we're on our way to solving the issue. I'll re-open the issue so that the original intent is taken care of as well. Thanks for pitching in. 🙂

PR #33166, opened 2024-05-03
fix: Start using SHA2 instead of SHA1 algorithms for signing during git operations

nidhi-nair:
## Description
Our current Jsch version only has support for SHA-1 which has long since been deprecated. Jsch has been out of support for a while now. The forks from the original repository don't seem to be officially supported by JGit. The JGit project itself has moved on to using Apache Mina as their official solution to sshd connections.

This PR switches our ssh provider to Apache Mina as well, which relies on Java security and hence has quite a lot of flexibility in the JCE providers it supports. We're making use of Sun defaults for RSA, with a little bit of help from Bouncy castle utilities to manage the conversions to and from x.509 patterns for users to copy values and reciprocate on their repositories.

The ECDSA algorithm has been switched to BouncyCastle for the conversion from OpenSSH format to x.509 just so we avoid manual shenanigans on our end to manage the values.

Fixes https://github.com/appsmithorg/appsmith/issues/32970

## Automation

/ok-to-test tags="@tag.Git"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8969063949>
> Commit: e5b9d4d91d261c85f33e4ade3e794674abd14137
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8969063949&attempt=1" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [ ] No


app/server/appsmith-git/src/main/java/com/appsmith/git/helpers/SshTransportConfigCallback.java
```diff
@@ -30,17 +55,72 @@ public SshTransportConfigCallback(String privateKey, String publicKey) {
...
+                    privateKeySpec = new OpenSSHPrivateKeySpec(content);
+                } else {
+                    privateKeySpec = new PKCS8EncodedKeySpec(Base64.getDecoder().decode(privateKey));
+                }
+
+                PrivateKey generatedPrivateKey = keyFactory.generatePrivate(privateKeySpec);
+
+                keyPair = new KeyPair(generatedPublicKey, generatedPrivateKey);
+                return List.of(keyPair);
+            } catch (NoSuchAlgorithmException | InvalidKeySpecException | IOException e) {
```
sondermanish, line 120: Is this error getting consumed somewhere in the backend.
if not then let's chain the exception instead of catching them at once?

sondermanish: this only makes sense if we want to show different errors.

sondermanish (approved):

PR #33197, opened 2024-05-06
chore: Add worklow to check cyclic deps in a PR

dvj1988:
## Description
Add a GitHub workflow to compare the number of cyclic dependencies in a PR to the number in the release branch. If the PR introduces new cyclic dependencies, a comment should be posted in the conversation to alert the developer.

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9228474130>
> Commit: c93f3e5df863ab7e2e287bf093a14a3f5f43e8e8
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9228474130&attempt=1" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [ ] No


.github/workflows/ci-client-cyclic-deps-check.yml
```diff
@@ -0,0 +1,106 @@
...
+
+      # Comment on the PR if cyclic dependencies are found
+      - name: Comment the result on PR
+        if: steps.check_cyclic_deps.outputs.has_more_cyclic_deps == 'true'
+        uses: actions/github-script@v3
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
+            const prNumber = context.payload.pull_request.number;
+            const message = `⚠️ Cyclic Dependency Check:\n\nYou PR has more cyclic depdencies than the release branch.\n\nResolve any new cyclic dependencies your PR has introduced.\n`;
```
riodeuno, line 90: There is a typo here: "Your". Let's use the phrase "This PR has increased the number of cyclic dependencies, when compared with the `release` branch".

Also, Can we link to a notion doc or something that will help the team identify the cyclic dependencies.
For example, run `yarn cyclic-deps` on `release` and run `yarn cyclic-deps` on your branch. Then diff the two files to see the paths that have cyclic dependencies.

app/client/scripts/circular_deps_check.js
```diff
@@ -0,0 +1,11 @@
+const { parseDependencyTree, parseCircular, prettyCircular } = require("dpdm");
+
+const CIRCULAR_DEPS_IN_RELEASE = 2965;
+
+parseDependencyTree("./src", {}).then((tree) => {
+  const circulars = parseCircular(tree);
+  if (circulars.length > CIRCULAR_DEPS_IN_RELEASE) {
+    console.log("More deps than release!");
```
riodeuno, line 8: Since this string is used as a truthy value in CI. Should we make it a simple string like "Failed" or something?

dvj1988: Yeah you are right. Will make this change.

.github/workflows/ci-client-cyclic-deps-check.yml
dvj1988: Got it. Will make this change

riodeuno (dismissed):

riodeuno:
@dvj1988 The cyclic deps has responded to this PR. Are these false positives?

dvj1988:
> @dvj1988 The cyclic deps has responded to this PR. Are these false positives?

@riodeuno Ah, my bad! No its not a false positive. I reduced the number to test the comment. I have reverted it back to the correct threshold in [this commit](https://github.com/appsmithorg/appsmith/pull/33197/commits/158cbbbd9c7d125c62636e155dfd97cd32caff8c)

app/client/scripts/circular_deps_check.js
```diff
@@ -0,0 +1,11 @@
+const { parseDependencyTree, parseCircular, prettyCircular } = require("dpdm");
+
+const CIRCULAR_DEPS_IN_RELEASE = 2965;
```
KelvinOm, line 3: I think having a constant will not be very convenient for us in the future. Especially when we start working on reducing the number of dependencies and we have to edit it manually every time. I think we could checkout to the `release` branch to get the number of dependencies and then compare them with the number of dependencies in the current branch. WDYT? 

@dvj1988 @riodeuno

.github/workflows/ci-client-cyclic-deps-check.yml
```diff
@@ -0,0 +1,106 @@
...
+          key: ${{ github.run_id }}-${{ github.job }}-client
+
+      # Fetch prior run result
+      - name: Get the previous run result
+        id: run_result
+        run: cat ~/run_result 2>/dev/null || echo 'default'
+
+      # In case of prior failure run the job
+      - if: steps.run_result.outputs.run_result != 'success'
+        run: echo "I'm alive!" && exit 0
```
mohanarpit, line 47: Nit: Recommend removing these steps as it's not required in this particular job. We use this for long running jobs (like Cypress test etc) to prevent re-running of the same job after a failure. In this particular case, I don't think this is required.

app/client/scripts/circular_deps_check.js
dvj1988: @KelvinOm Understood. Let me update the PR.

riodeuno (approved):

PR #33216, opened 2024-05-06
fix: Updating invite modal component to fix issue on EE & improving code splitting of a file

ankitakinger:
## Description

Updating invite modal component to fix issue on EE & improving code splitting of a file

Fixes [#3954](https://github.com/appsmithorg/appsmith-ee/issues/3954)

## Automation

/ok-to-test tags="@tag.Workspace, @tag.Git"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/8972376691>
> Commit: cd3d806b5094358894977f28af83c43199ec15d9
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=8972376691&attempt=1" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/client/src/ce/components/gitComponents/GitSettingsCDTab/UnlicensedGitCD.tsx
+++ b/app/client/src/ce/components/gitComponents/GitSettingsCDTab/UnlicensedGitCD.tsx
@@ -9,23 +9,23 @@ import { useAppsmithEnterpriseLink } from "pages/Editor/gitSync/GitSettingsModal
 import React from "react";
 import styled from "styled-components";
 
-const Container = styled.div`
+export const Container = styled.div`
   padding-top: 8px;
   padding-bottom: 16px;
   overflow: auto;
   min-height: calc(360px + 52px);
 `;
 
-const SectionTitle = styled(Text)`
+export const SectionTitle = styled(Text)`
   font-weight: 600;
   margin-bottom: 4px;
 `;
 
-const SectionDesc = styled(Text)`
+export const SectionDesc = styled(Text)`
   margin-bottom: 12px;
 `;
 
-const StyledButton = styled(Button)`
+export const StyledButton = styled(Button)`
   display: inline-block;
 `;
 
--- a/app/client/src/ce/components/gitComponents/GitSettingsCDTab/index.tsx
+++ b/app/client/src/ce/components/gitComponents/GitSettingsCDTab/index.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import UnlicensedGitCD from "./UnlicensedGitCD";
+import UnlicensedGitCD from "@appsmith/components/gitComponents/GitSettingsCDTab/UnlicensedGitCD";
 
 function GitSettingsCDTab() {
   return <UnlicensedGitCD />;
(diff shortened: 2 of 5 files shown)


app/client/src/ce/components/gitComponents/GitSettingsCDTab/UnlicensedGitCD.tsx
```diff
@@ -9,23 +9,23 @@ import { useAppsmithEnterpriseLink } from "pages/Editor/gitSync/GitSettingsModal
 import React from "react";
 import styled from "styled-components";
 
-const Container = styled.div`
+export const Container = styled.div`
```
brayn003, line 12: Any particular reason we are exporting it? Are we importing it somewhere?

app/client/src/ce/components/gitComponents/GitSettingsCDTab/index.tsx
```diff
@@ -1,5 +1,5 @@
 import React from "react";
-import UnlicensedGitCD from "./UnlicensedGitCD";
+import UnlicensedGitCD from "@appsmith/components/gitComponents/GitSettingsCDTab/UnlicensedGitCD";
```
brayn003, line 2: This will import EE + CE code. The idea is to just import CE code

brayn003 (changes requested):

app/client/src/ee/components/gitComponents/GitSettingsCDTab/UnlicensedGitCD.tsx
```diff
@@ -0,0 +1,3 @@
+export * from "ce/components/gitComponents/GitSettingsCDTab/UnlicensedGitCD";
```
brayn003, line 1: I think you are trying to make an EE equivalent for UnlicensedGitCD, however this is a CE only components and file splitting is redundant. That is why it was written the way it was. Do you think we should split a file even when the logic for it will remain the same?

brayn003 (changes requested):

app/client/src/ce/components/gitComponents/GitSettingsCDTab/UnlicensedGitCD.tsx
ankitakinger: Yup importing on respective ee file in EE repo

app/client/src/ce/components/gitComponents/GitSettingsCDTab/index.tsx
ankitakinger: The imports should never be ce code... On both the repos anyway the component is exactly same so this is not a problem. You can refer to other files which we code split in the past to understand this better.

app/client/src/ee/components/gitComponents/GitSettingsCDTab/UnlicensedGitCD.tsx
ankitakinger: I know. This ideally shouldn't have been part of ce folder but should have been directly added in src/components folder. Since that's a bigger change I went ahead doing this as of now, because the entire file path doesn't exist under src/components folder

brayn003 (approved):

PR #33234, opened 2024-05-07
fix:button group spec

jsartisan:
Fixes the button group failing spec


ankitakinger (approved):

app/client/cypress/e2e/Regression/ClientSide/Widgets/Button/ButtonGroup2_spec.ts
```diff
@@ -163,7 +163,7 @@ describe(
       agHelper.ClickButton("Close");
       EditorNavigation.SelectEntityByName("ButtonGroup1", EntityType.Widget);
       agHelper.GetNClick(buttongroupwidgetlocators.buttonSettingInPropPane, 0);
-      propPane.EnterJSContext("onClick", "{{showModal(Modal1.name)}}");
+      propPane.EnterJSContext("onClick", "{{showModal('Modal1')}}");
```
mohanarpit, line 166: @jsartisan Not sure why we are changing this. This was an enhancement done by Aman recently where we can access Modals by using the property `Modal.name` instead of `Modal1`. If this feature isn't working in Cypress test case, it's a regression & we should resolve this problem in the product rather than in the Cypress test.

CC: @AmanAgarwal041

app/client/cypress/e2e/Regression/ClientSide/Widgets/Button/ButtonGroup2_spec.ts
```diff
@@ -23,8 +23,8 @@ describe(
 
     it("1. Add, Delete more buttons to the group and verify", function () {
       agHelper.GetNClick(buttongroupwidgetlocators.newButton);
-      agHelper.TypeText("//*[ @value='Group Button 1']", " New Button");
-      agHelper.AssertContains("Group Button 1 New Button");
+      agHelper.TypeText("//*[ @value='Do Something']", " New Button");
+      agHelper.AssertContains("Do Something New Button");
```
mohanarpit, line 27: @jsartisan Shouldn't this be "Do something"? Across the entire product, we follow sentence casing. Should the default text on the button follow the same conventions?

mohanarpit: Btw, is it possible for us to use a constant variable in the product & Cypress test to avoid such a failure in the future?

jsartisan: The text was suggested by Taras. I'll check with him if we need to change the casing.

jsartisan: > Btw, is it possible for us to use a constant variable in the product & Cypress test to avoid such a failure in the future?

Let me check on this.

jsartisan: Added the label in the messages file and used it everywhere. Thanks for the suggestion.

app/client/cypress/limited-tests.txt
```diff
@@ -1,5 +1,6 @@
 # To run only limited tests - give the spec names in below format:
-cypress/e2e/Regression/ClientSide/Templates/Fork_Template_spec.js
+#cypress/e2e/Regression/ClientSide/Templates/Fork_Template_spec.js
```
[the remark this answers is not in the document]
jsartisan: Done.

ankitakinger (approved):

PR #33293, opened 2024-05-08
chore: Add log when `info.json` is unavailable at specified path

abhvsn:


app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/ProjectProperties.java
```diff
@@ -30,10 +30,12 @@ public ProjectProperties(ObjectMapper objectMapper) {
                 BuildInfo buildInfo = objectMapper.readValue(jsonContent, BuildInfo.class);
                 version = buildInfo.getVersion();
                 commitSha = buildInfo.getCommitSha();
+            } else {
+                log.debug("Unable to find info.json at {}", INFO_JSON_PATH);
             }
         } catch (IOException e) {
             // Ignore the exception and return "UNKNOWN" as the version
-            log.debug("Error reading version from info.json {}", e.getMessage());
+            log.error("Error reading version from info.json {}", e.getMessage(), e);
```
sharat87, line 38: 1. I'd recommend not adding `, e` to this. It'll print the whole stack trace and I don't think that'll give you more information than what `e.getMessage()` would.
2. I'd also recommend removing the `if` check, instead of adding an `else`. If the file isn't there, let it throw an error and we're catching `IOException` anyway. In fact, I'd rather the backend not start at all if `info.json` is missing, but that makes local dev less convenient.

By the way, the `entrypoint.sh` already has a hard-check to fail if the `info.json` file is missing.
https://github.com/appsmithorg/appsmith/blob/f71bdc8b239013ad76a69e0e918a710ffe490342/deploy/docker/fs/opt/appsmith/entrypoint.sh#L462

abhvsn: @sharat87 if we already have the check in place for info.json then do you think there is a scenario where version does not get populated. I'm asking this because in license check call in EE we refer to the same variable https://github.com/appsmithorg/appsmith-ee/blob/5aa31d0e6162c55a13a79072f33cfd4df24cdc61/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/BaseLicenseAPIManagerImpl.java#L29.

abhvsn: > I'd recommend not adding , e to this. It'll print the whole stack trace and I don't think that'll give you more information than what e.getMessage() would.

@sharat87 currently this only prints the info.json path nothing around what failed.

sharat87: Perhaps the license check is happening _before_ the JSON file is read? Is there a possibility of a race condition here? If we can make the dependency explicit, then that question will go away too.

Other than that, I don't see why this file should be missing or invalid.

sharat87 (approved):

Issue #33030, opened 2024-04-29
[Task]: Persist dependency map of a page in backend


dvj1988:
The dependencyMap consumes considerable computation on the frontend. We would like to persist that in the backend for every page in an application. 

For details on the signature and payload of the api calls please refer [this document](https://www.notion.so/appsmith/Solution-design-dependencyMap-0c031675f13140f7a433c731b3e45470?pvs=4#1ee6692b74804422a80cba86cc9ba87f) 

```[tasklist]
### Tasks
- [x] Add dependencyMap in the response of Consolidated view api. (GET /api/v1/consolidated-api/view)
- [x] Add dependencyMap in the response of Consolidated edit api. (GET /api/v1/consolidated-api/edit)
- [x] New API to update the dependency map of a page. (PUT /pages/{pageId}/dependencyMap)
- [x] Add dependency map is the response of get page api (edit mode) (GET /api/v1/pages/{pageId})
- [x] Add dependency map in the response of get page api (view mode) (GET /api/v1/pages/{pageId}/view)
- [x] Publish api changes - Copy the dependency map of a page from unpublished to published.
```

dvj1988:
cc: @rahulbarwal @rajatagrawal

dvj1988:
The updatedAt time looks common for both published and unpublished version of the app. For this feature we would want the updateAt times to be separate for both. 
The updateAt time for the unpublished page should denote the last time the page was updated and the updatedAt time for the published page should denote the last time it was published.

AnaghHegde:
Adding more story points due to last movement review comment which requires change in the flow.

PR #33296, opened 2024-05-08
chore: Add API for storing dependencyMap in page object

AnaghHegde:
## Description
The dependencyMap consumes considerable computation on the frontend. We would like to persist that in the backend for every page in an application.

For details on the signature and payload of the api calls please refer [this document](https://www.notion.so/appsmith/Solution-design-dependencyMap-0c031675f13140f7a433c731b3e45470?pvs=4#1ee6692b74804422a80cba86cc9ba87f)
Slack thread - https://theappsmith.slack.com/archives/C024GUDM0LT/p1714100372372069

Fixes https://github.com/appsmithorg/appsmith/issues/33030

## Automation

/ok-to-test tags="@tag.Git"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9253166755>
> Commit: 03e9f01bec1203feb1278118f2286fc66cc548bd
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9253166755&attempt=1" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [x] No


app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewPageRepositoryCEImpl.java
```diff
@@ -238,4 +254,11 @@ public Flux<NewPage> findAllByApplicationIdsWithoutPermission(
                 .fields(includeFields)
                 .all();
     }
+
+    @Override
+    public Mono<Integer> updateDependencyMap(String pageId, Map<String, List<String>> dependencyMap) {
+        Criteria pageCriteria = where(NewPage.Fields.id).is(pageId);
```
NilanshBansal, line 260: Nit: can we use BridgeUpdate query here. Similar usage code ref below.
https://github.com/appsmithorg/appsmith/blob/3e06f1a5704d1303328751b0093663c14330d8ce/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewActionRepositoryCEImpl.java#L376-L383

Why use Bridge APIs?
https://theappsmith.slack.com/archives/CPQNLFHTN/p1710836011164579

NilanshBansal (changes requested):

NilanshBansal:
@AnaghHegde since you are adding a new field `dependencyMap` in the PageDTO, can you please do a sanity check for the following cases. 
- Is this field exported to the git repository for git-connected apps?
- How will this work for import export? Can consider scenarios - an old JSON imported to the current version, what will be the value of these fields in the DB (I believe this should be null and should be okay but you can still check once).
- On exporting the current version JSON, is the `dependencyMap` field exported?

cc: @rahulbarwal

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CustomNewPageRepositoryCEImpl.java
AnaghHegde: Thanks for flagging this ❤️

NilanshBansal:
> @AnaghHegde since you are adding a new field `dependencyMap` in the PageDTO, can you please do a sanity check for the following cases.
> 
> * Is this field exported to the git repository for git-connected apps?
> * How will this work for import export? Can consider scenarios - an old JSON imported to the current version, what will be the value of these fields in the DB (I believe this should be null and should be okay but you can still check once).
> * On exporting the current version JSON, is the `dependencyMap` field exported?
> 
> cc: @rahulbarwal

@AnaghHegde @rahulbarwal can you please confirm if you have checked this?

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/PageControllerCE.java
```diff
@@ -196,4 +199,15 @@ public Mono<ResponseDTO<ApplicationPagesDTO>> getAllPages(
                 .findApplicationPages(applicationId, pageId, branchName, mode)
                 .map(resources -> new ResponseDTO<>(HttpStatus.OK.value(), resources, null));
     }
+
+    @JsonView(Views.Public.class)
+    @PutMapping("/{defaultPageId}/dependencyMap")
```
mohanarpit, line 204: For my information, any reason we chose to do this as a separate endpoint instead of part of update page layout? Would having a separate endpoint cause a race condition between saving the updated page and saving the new dependency map?

dvj1988: This is a valid concern and we don't have a robust approach at the moment. We are reducing the scope to use the dependency map cache only in the view mode for now. When a user publishes the application, we will trigger the updateDependencyMap API followed by the publish API. The dependency map in the unpublished version will reflect the last published version, and it will be updated with each publish.
We will get back to edit mode dependency maps later when we have a better approach for invalidation of older dependency maps.
@AnaghHegde For the time being we can go ahead with this approach.
cc: @rajatagrawal

AnaghHegde:
@NilanshBansal We are good from git and export point of view.

NilanshBansal (approved):
Considering @mohanarpit's comment [here](https://github.com/appsmithorg/appsmith/pull/33296#discussion_r1606271991) is addressed, this LGTM.

PR #33306, opened 2024-05-09
chore: Prefetch consolidated api using service worker

dvj1988:
## Description
Prefetch the consolidated API using the service worker. This api is prefetched for the view and edit urls.

Fixes [Issue URL](https://github.com/appsmithorg/appsmith/issues/33497)
> [!WARNING]  

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9108829019>
> Commit: 30e2db9961d444b984560bb034ef87b443318213
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9108829019&attempt=2" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [ ] No


app/client/craco.dev.config.js
```diff
@@ -21,4 +21,32 @@ module.exports = merge(common, {
   experiments: {
     cacheUnaffected: true,
   },
+  webpack: {
```
dvj1988, line 24: Revert this. This change is required only for testing

app/client/src/App.css
```diff
@@ -15,6 +15,7 @@
   align-items: center;
   justify-content: center;
   font-size: calc(10px + 2vmin);
+  color: red;
```
dvj1988, line 18: Revert this change

app/client/src/ce/AppRouter.tsx
```diff
@@ -156,6 +156,8 @@ export default function AppRouter() {
     dispatch(initCurrentPage());
   }, []);
 
+  useEffect(() => {}, [safeCrash]);
```
dvj1988, line 159: Revert this change

app/client/src/ce/AppRouter.tsx
```diff
@@ -156,6 +156,8 @@ export default function AppRouter() {
...
+
```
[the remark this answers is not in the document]
dvj1988: This was a dummy commit. I have reverted this change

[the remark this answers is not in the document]
dvj1988: Can you review the PR again and update the summary?

app/client/src/utils/serviceWorkerUtils.js
```diff
@@ -0,0 +1,179 @@
...
+      return this.pendingApiRequests.get(requestKey); // Return the ongoing request
+    }
+
+    // Fetch the request
+    const fetchPromise = fetch(request)
+      .then(async (response) => {
+        // If the response is ok, clone the response
+        if (response.ok) {
+          const clonedResponse = response.clone();
+          // Store in cache with expiration header
```
vsvamsi1, line 164: Why do we need to clone here? I thought the code will look like something like this.
`
    const fetchPromise = await fetch(request)
    if( response.ok){
              const resp = await response.json();

`

app/client/src/utils/serviceWorkerUtils.js
```diff
@@ -0,0 +1,179 @@
...
+
+    // Check for ongoing request
+    const requestKey = `${request.method}:${request.url}`;
+    if (this.pendingApiRequests.has(requestKey)) {
+      return this.pendingApiRequests.get(requestKey); // Return the ongoing request
+    }
+
+    // Fetch the request
+    const fetchPromise = fetch(request)
+      .then(async (response) => {
```
rajatagrawal, line 160: Would it be possible to use async-await here ?

app/client/src/utils/serviceWorkerUtils.js
```diff
@@ -0,0 +1,179 @@
...
+ * Cache strategy for Appsmith API
+ */
+export class AppsmithApiCacheStrategy {
+  /**
+   *
+   * @param {string} cacheName
+   */
+  constructor(cacheName) {
+    // Map to store ongoing API requests
+    this.pendingApiRequests = new Map();
```
rajatagrawal, line 119: Are there advantages of using Map over a simple javascript object ?

app/client/src/utils/serviceWorkerUtils.js
dvj1988: This fn returns the promise and not its result. So the callback had to be defined by chaining.

app/client/src/utils/serviceWorkerUtils.test.js
```diff
@@ -0,0 +1,349 @@
...
+      const result = matchViewerPath(pathName);
+
+      expect(result).toBeTruthy();
+      expect(result.params).toHaveProperty("applicationSlug");
+      expect(result.params).toHaveProperty("pageSlug");
+      expect(result.params).toHaveProperty("pageId", "123");
+    });
+
+    it("should match the custom viewer path", () => {
+      const pathName = "/app/customSlug-custom-456";
```
rajatagrawal, line 325: Is `custom` part of `customSlug` ? Should it be included in result.params ?

app/client/src/utils/serviceWorkerUtils.test.js
```diff
@@ -0,0 +1,349 @@
...
+  describe("matchBuilderPath", () => {
+    it("should match the standard builder path", () => {
+      const pathName = "/app/applicationSlug/pageSlug-123/edit";
+      const options = { end: false };
+      const result = matchBuilderPath(pathName, options);
+
+      expect(result).toBeTruthy();
+      expect(result.params).toHaveProperty("applicationSlug");
+      expect(result.params).toHaveProperty("pageSlug");
+      expect(result.params).toHaveProperty("pageId", "123");
```
rajatagrawal, line 261: Can we have a test case where the pageId is alphanumeric ? It may already by working, but having a test case will just ensure that the code logic will not break for alphanumeric pageIds. The pageIDs would be alphanumeric on prod.

app/client/src/utils/serviceWorkerUtils.test.js
```diff
@@ -0,0 +1,349 @@
...
+
+    it("should not match when no pageId is present", () => {
+      const pathName = "/app/applicationSlug/pageSlug-edit";
+      const options = { end: false };
+      const result = matchBuilderPath(pathName, options);
+
+      expect(result).toBeFalsy();
+    });
+
+    it("should not match when the path is edit widgets", () => {
```
rajatagrawal, line 300: Typo in the test description: `should not` -> `should`

app/client/craco.dev.config.js
rajatagrawal: Does this change need to be reverted ?

app/client/src/serviceWorker.js
```diff
@@ -34,6 +38,31 @@ self.__WB_DISABLE_DEV_DEBUG_LOGS = false;
...
+ * @returns
+ */
+const handleFetchHtml = async (event, request, url) => {
+  const prefetchRequest = getPrefetchRequest(url);
+
+  if (prefetchRequest) {
+    appsmithApiCacheStrategy.handle(prefetchRequest, true);
+  }
+
+  const networkHandler = new NetworkOnly();
```
rajatagrawal, line 60: Is this supposed to be part of if/else above this line ?

app/client/src/utils/serviceWorkerUtils.test.js
```diff
@@ -0,0 +1,349 @@
...
+      const result = getPrefetchRequest(url);
+
+      expect(result).toEqual(expectedResult);
+    });
+  });
+
+  describe("AppsmithApiCacheStrategy", () => {
+    let apiCacheStrategy;
+    let cacheMock;
+
```
vsvamsi1, line 173: Can we have a set of test cases with "AppsmithApiCacheStrategy" integrated in our data layer?

dvj1988: Main bundle (Redux sagas and other modules) do not interact with the service worker code and the integration is through the browser apis. To test the end to end flow we need a browser environment and hence cypress would be the best place for this.

app/client/src/serviceWorker.js
dvj1988: No. This is part of the normal fetch behaviour for all pages and the if block is only meant for edit and view page html requests.

app/client/src/utils/serviceWorkerUtils.js
```diff
@@ -0,0 +1,179 @@
...
+  return decodeURIComponent(params.get(key) || "");
+};
+
+/**
+ * Function to match the path with the builder path
+ * @param {string} pathName
+ * @param {Object} options
+ * @param {boolean} options.end
+ * @returns {Match<object> | boolean}
+ */
```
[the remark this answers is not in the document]
dvj1988: Sure noted.

app/client/src/utils/serviceWorkerUtils.js
dvj1988: There are some [advantages](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#objects_vs._maps). In the current implementation I used it because of the helper methods to read, update and delete. These methods help manipulate the map easier and makes the code more readable.

app/client/src/utils/serviceWorkerUtils.js
dvj1988: > Cloning the response is necessary because request and response streams can only be read once. In order to return the response to the browser and put it in the cache we have to clone it. So the original gets returned to the browser and the clone gets sent to the cache. They are each read once.

This is an excerpt from this [mdn reference](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API/Using_Service_Workers#the_premise_of_service_workers) which explains the reason for cloning.

app/client/craco.dev.config.js
dvj1988: We can leave it as is and enable it for dev mode.

Just FYI there are some warnings that popup in the terminal as result of enabling WorkboxPlugin in dev mode. This is a known issue for webpack watch mode (InjectManifest has been called multiple times, perhaps due to running webpack in --watch mode.)

app/client/src/utils/serviceWorkerUtils.test.js
dvj1988: Resolved

app/client/src/utils/serviceWorkerUtils.test.js
dvj1988: Resolved.

app/client/src/utils/serviceWorkerUtils.js
```diff
@@ -0,0 +1,257 @@
...
+  // If the URL matches the viewer path
+  if (matchViewer && matchViewer.params?.pageId) {
+    const searchParams = getConsolidatedAPISearchParams(matchViewer.params);
+    const requestUrl = `${url.origin}/api/v1/consolidated-api/view?${searchParams}`;
+    const request = new Request(requestUrl, { method: "GET", headers });
+    return request;
+  }
+
+  // Return null if the URL does not match the builder or viewer path
+  return null;
```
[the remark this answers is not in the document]
dvj1988: Resolved.

dvj1988:
Running e2e test cases in [this PR](https://github.com/appsmithorg/appsmith-ee/pull/4175)

rajatagrawal:
@dvj1988 

Can we please add a github issue for this PR before merging ?

rajatagrawal:
@dvj1988

Can we please add a github issue for this PR before merging ?

dvj1988:
> @dvj1988
> 
> Can we please add a github issue for this PR before merging ?

Added an issue to the PR description.

app/client/src/utils/serviceWorkerUtils.js
```diff
@@ -0,0 +1,185 @@
...
+
+  /**
+   * Function to fetch and cache the consolidated API
+   * @param {Request} request
+   * @returns
+   */
+
+  async cacheConsolidatedApi(request) {
+    // Acquire the lock
+    await this.consolidatedApiFetchmutex.acquire();
```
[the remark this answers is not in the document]
rajatagrawal: Should there be a check to ensure that cache is initialized before fetching as well as accessing cache ?

rajatagrawal:
Suggested refactor to comply with single responsibility principle and some suggested naming changes to make code more generic. The code logic is fine and untouched. Just a suggestion on an iteration of code organization. I have suggested the changes in a patch file. The patch can be applied locally to see the suggested changes.

[gitdiffpatch.zip](https://github.com/appsmithorg/appsmith/files/15329125/gitdiffpatch.zip)

app/client/src/utils/serviceWorkerUtils.js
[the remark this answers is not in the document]
rajatagrawal: Another suggested way to overcome this problem could be to add cache initialization in a mutex block

app/client/src/utils/serviceWorkerUtils.js
```diff
@@ -0,0 +1,185 @@
...
+      await this.cache.delete(request);
+    } finally {
+      // Release the lock
+      this.consolidatedApiFetchmutex.release();
+    }
+  }
+
+  async getCachedResponse(request) {
+    // Wait for the lock to be released
+    await this.consolidatedApiFetchmutex.waitForUnlock();
```
rajatagrawal, line 161: Should this be a mutex.acquire instead of waitForUnlock ? Theoretically, before the cache read finishes, another request can be made and the cache be updated, which could lead to corruption of cache data read by this function.

app/client/src/utils/serviceWorkerUtils.js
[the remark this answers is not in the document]
dvj1988: That's a valid point. Will init the cache in the cacheConsolidatedApi fn rather than the constructor.

app/client/src/utils/serviceWorkerUtils.js
dvj1988: The mutex was added to make sure no reads can happen when there is an ongoing write. 
> before the cache read finishes, another request can be made and the cache be updated
Can you give a scenario when this would happen?

dvj1988:
> Suggested refactor to comply with single responsibility principle and some suggested naming changes to make code more generic. The code logic is fine and untouched. Just a suggestion on an iteration of code organization. I have suggested the changes in a patch file. The patch can be applied locally to see the suggested changes.
> 
> [gitdiffpatch.zip](https://github.com/appsmithorg/appsmith/files/15329125/gitdiffpatch.zip)

IMO all changes are nit picks which could go either way. But in any case I can apply the patch.

app/client/src/utils/serviceWorkerUtils.js
rajatagrawal: The read isn't complete but another fetch request comes to service worker

dvj1988: Do you mean a fetch request for a html document?

rajatagrawal: or an ajax call for same api call.

app/client/src/serviceWorker.js
```diff
@@ -52,7 +79,30 @@
...
+
+    // If the response is cached, return the response
+    if (cachedResponse) {
+      return cachedResponse;
+    }
+
+    // If the response is not cached, fetch the response
+    const networkHandler = new NetworkOnly();
+    return networkHandler.handle({ event, request });
+  },
```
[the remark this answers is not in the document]
dvj1988: Error handling on failure is handled in the main bundle. We just need to return the Network handler and the main thread code will take care of gracefully handling this.

app/client/src/serviceWorker.js
```diff
@@ -34,6 +38,29 @@ self.__WB_DISABLE_DEV_DEBUG_LOGS = false;
...
+
+  if (prefetchConsolidatedApiRequest) {
+    consolidatedApiCacheStrategy.cacheConsolidatedApi(
+      prefetchConsolidatedApiRequest,
+    );
+  }
+
+  const networkHandler = new NetworkOnly();
+  return networkHandler.handle({ event, request });
+};
```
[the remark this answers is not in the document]
dvj1988: Resolved

app/client/src/utils/serviceWorkerUtils.js
dvj1988: If it is an HTML fetch, it shouldn't be a problem as the main thread and service worker execution context are reinitialized.

We are defining this as a read-only-once cache to ensure there aren't any stale responses that might accidentally be served to the main thread. If multiple fetches for the consolidated API with the same pageId are initiated, the behavior will be that the first fetch will hit the cache while all others will miss it. In the case of a miss, we will fetch from the server. Locking the read wouldn't help here because we want it to be consumed only once, and all subsequent fetches should go to the network by design.

app/client/craco.dev.config.js
KelvinOm: @dvj1988 Does it affect the build time somehow? Did you check it?

We have some issues with  building time, if it is not necessary, then I would prefer not to add anything here.

KelvinOm (changes requested):

app/client/craco.dev.config.js
dvj1988: I have not seen any issues with the dev server build time or hot reloads. Preferably we want to keep this so that we have the same behaviour of the service worker in the dev and prod modes. 
Earlier we were using service worker to cache the static assets but this PR is prefetching and caching one of the apis. If there are any main bundle changes that would affect the behaviour of the service worker or vice versa we want to start catching them in the dev mode itself.
After the merge I will keep a track of whether this is affecting any devs local setup and can take a measure accordingly.

dvj1988: @KelvinOm Do you have any specific scenario in mind to test?

dvj1988: @KelvinOm By cache do you mean lint cache? 
How do you propose I measure these?

KelvinOm: By cache I mean `babel-loader` cache it stored at `app/client/node_modules/.cache/babel-loader`. It is created when the webpack is running.

> How do you propose I measure these?
I do not know a good way to do this. I'm measuring the time in terminal. We can use `time` like `time yarn start`.  I couldn't run the webpack speed measure plugin because it doesn't work with `craco`.  So we have to measure it manually.

dvj1988: I used the package called `gnomon`. And the time taken looks the same.

Start time without cache was replicated by following the steps
- Checkout the branch
- delete node_modules
- install deps
- comment out the cache settings in craco.dev.config
- start the server

```sh
yarn start | gnomon --type=elapsed-total
```

Tests done in: Apple M1 Pro, RAM 32 GB

|Branch|Start time without cache|Start time with cache |Recompilation Time
|-------------------|----------|----------|----------|
release|190-192s|12-13s|4-6s
chore/prefetch-consolidated-api|190-192s|12-13s|4-6s

Start time without cache - release
Screenshot 2024-05-16 at 4 44 52 PM
Start time with cache - release
Screenshot 2024-05-16 at 4 47 10 PM
Start time without cache - chore/prefetch-consolidated-api
Screenshot 2024-05-16 at 5 39 47 PM
Start time with cache - chore/prefetch-consolidated-api
Screenshot 2024-05-16 at 5 12 39 PM

KelvinOm (approved):

PR #33309, opened 2024-05-09
feat: Moved Tenant Information to Redis for quick access

### LINKED ISSUE #33083: [Task]: Implement Redis caching for tenant configuration
### Description

The tenants/current api is the slowest component of consolidated/view api. Reducing the latency of tenants/current api will reduce the latency of backend in view experience for all users.

Relevant Slack thread : https://theappsmith.slack.com/archives/C024GUDM0LT/p1713933688987279

In the diagram in the slack thread, the blue line just below the magenta line at the top is the contribution of tenants/current to consolidated api. The diagram shows that tenants/current is 90% contributor to consolidated view api.

Upon debugging the cause of latency from tenants/current, it is found that the api makes a DB call to fetch the tenant configuration every time. 

Caching the tenant config ID is already present in the codebase. Extending that approach, we can cache the whole tenant config and update it when a user changes admin settings.

Not making the DB call will reduce the latency of tenants/current api and decrease the latency of consolidated view api as well.

CE PR: https://github.com/appsmithorg/appsmith/pull/33641
Counterpart EE PR: https://github.com/appsmithorg/appsmith-ee/pull/4275

### Subtasks
- Implement caching for fetching tenant configuration
- Invalidate cache if tenant configuration gets updated, update the database and rebuild cache.

NilanshBansal:
## Description

The tenant is fetched multiple times across the appsmith codebase but is rarely updated (from the admin settings). Every time a fetch call to the database is costly both in terms of resources and time taken. 
The consolidated api also makes a call to fetch the tenant and return to the client. To improve the performance of fetching the tenant information, we are moving the tenant information to redis cache for quicker fetch. 
This will improve the performance of the consolidated api and also reduce the time taken by all the different functionalities within the backend codebase which depend on tenant to process further. 

**TL;DR**
Adds tenant information `tenantService.getDefaultTenant()` to redis.

Fixes #33083 

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9079438120>
> Commit: 306e77f1460caf5302e1dd39fc58cce35790a04b
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9079438120&attempt=2" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [x] No


app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CacheableRepositoryHelperCE.java
```diff
@@ -18,4 +19,8 @@ public interface CacheableRepositoryHelperCE {
     Mono<String> getDefaultTenantId();
 
     Mono<String> getInstanceAdminPermissionGroupId();
+
+    Mono<Tenant> fetchCachedTenant(String tenantId);
```
mohanarpit, line 23: Nit: Suggestion renaming this to `fetchTenant`. This is because the function is already in the class `CacheableRepositoryHelper`. Hence, we don't need to add cache to the function name as well.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/TenantRepositoryCE.java
```diff
@@ -5,6 +5,6 @@
 import reactor.core.publisher.Mono;
 
 public interface TenantRepositoryCE extends BaseRepository<Tenant, String>, CustomTenantRepositoryCE {
-
+    // Use tenantService.getDefaultTenant() instead of this method as it is cached to redis.
```
mohanarpit, line 8: Are we deprecating `findBySlug()` function? If yes, I'd recommend either replacing the usage of `findBySlug()` in this PR or adding a `@Deprecated` annotation on this function so that it can be replaced across the board in another PR.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/TenantServiceCEImpl.java
```diff
@@ -39,17 +40,21 @@ public class TenantServiceCEImpl extends BaseService<TenantRepository, Tenant, S
 
     private final FeatureFlagMigrationHelper featureFlagMigrationHelper;
 
+    private CacheableRepositoryHelper cacheableRepositoryHelper;
```
mohanarpit, line 43:
```suggestion
    private final CacheableRepositoryHelper cacheableRepositoryHelper;
```

NilanshBansal:
Test `updateTenantConfiguration_updateStrongPasswordPolicy_success` is failing weirdly with 2 different tenantIds getting populated in redis for a single test case. This is creating inconsistency and leading to failure of this test.

image

rajatagrawal:
@NilanshBansal 

Would it be possible to cache tenant configuration as part of this request https://github.com/appsmithorg/appsmith/blob/217da70d9fd45b8aab3004d68b49d5e8a4087206/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/TenantServiceCEImpl.java#L153 ?



This DB call is made the first time tenant id is needed. It fetches the tenant configuration but only persists id in redis cache. The code can be changed to persist the whole configuration. Thoughts ?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/TenantServiceCEImpl.java
```diff
@@ -151,16 +166,9 @@ public Mono<Tenant> getTenantConfiguration() {
...
-        return repository
-                .findBySlug(FieldName.DEFAULT)
-                .map(tenant -> {
-                    if (tenant.getTenantConfiguration() == null) {
-                        tenant.setTenantConfiguration(new TenantConfiguration());
-                    }
-                    return tenant;
-                })
+        // Fetching Tenant from redis cache
+        return getDefaultTenantId()
```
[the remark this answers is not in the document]
NilanshBansal: @rajatagrawal fetching the defaultTenantId is stored in the instance variable. It is not going to redis or to the database. This is the fastest call that is being made.

NilanshBansal:
> @NilanshBansal
> 
> Would it be possible to cache tenant configuration as part of this [request](https://github.com/appsmithorg/appsmith/blob/217da70d9fd45b8aab3004d68b49d5e8a4087206/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/TenantServiceCEImpl.java#L56)
> 
> This DB call is made the first time tenant id is needed. It fetches the tenant configuration but only persists id in redis cache. The code can be changed to persist the whole configuration. Thoughts ?

@rajatagrawal It is already the fastest call that we are making even faster than redis. Since this is being fetched from the class instance variable and not from any network call. 
Saying that, there are a lot of code level/readability improvements that we can make to this piece of code but purely from a performance standpoint it will not give us any perf boost.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CacheableRepositoryHelperCEImpl.java
```diff
@@ -166,4 +167,31 @@ public Mono<String> getInstanceAdminPermissionGroupId() {
...
+     * If not present in cache, then it fetches the default tenant from the database and adds to redis.
+     * @param tenantId
+     * @return
+     */
+    @Cache(cacheName = "defaultTenant", key = "{#tenantId}")
+    @Override
+    public Mono<Tenant> fetchDefaultTenant(String tenantId) {
+        BridgeQuery<Tenant> defaultTenantCriteria = Bridge.equal(Tenant.Fields.slug, FieldName.DEFAULT);
+        Query query = new Query();
+        query.addCriteria(defaultTenantCriteria);
```
abhvsn, line 182: Let's also include notDeleted criteria to be on safer side.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CacheableRepositoryHelperCEImpl.java
```diff
@@ -166,4 +167,31 @@ public Mono<String> getInstanceAdminPermissionGroupId() {
...
                         inMemoryCacheableRepositoryHelper.setInstanceAdminPermissionGroupId(permissionGroupId));
     }
+
+    /**
+     * Returns the default tenant from the cache if present.
+     * If not present in cache, then it fetches the default tenant from the database and adds to redis.
+     * @param tenantId
+     * @return
+     */
+    @Cache(cacheName = "defaultTenant", key = "{#tenantId}")
```
abhvsn, line 177: Considering there is going to be single defaultTenant does it make sense to remove the key altogether? Or even better have the `cacheName = "tenant" and key = "{#tenantId}"`, as this can be extended when we implement multi-tenancy in future.

abhvsn (changes requested):
@NilanshBansal added couple of minor comments. Also we are updating the tenant in `restartTenant` which is making a call to `BaseService` and not covered as a part of invalidation. I'm little concerned on invalidation part, let's make sure to override the base service and repository methods which are saving and updating the DB object.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CacheableRepositoryHelperCEImpl.java
NilanshBansal: Makes sense. Thanks @abhvsn!

abhvsn (approved):

PR #33332, opened 2024-05-09
chore: WIP Permission group migrations

vivonk:
## Description

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/ce/R__finalMigrations.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/ce/R__finalMigrations.java
@@ -1,0 +1,16 @@
+package com.appsmith.server.migrations.ce;
+
+import com.appsmith.server.migrations.AppsmithJavaMigration;
+import org.springframework.jdbc.core.JdbcTemplate;
+
+public class R__finalMigrations extends AppsmithJavaMigration {
+    private JdbcTemplate jdbcTemplate;
+
+    @Override
+    public void migrate(JdbcTemplate jdbcTemplate) throws Exception {
+        this.jdbcTemplate = jdbcTemplate;
+        // This is a placeholder migration that will be the last migration to run in any CE/EE version.
+        // This migration will be used to add any final migrations that need to be run after all other migrations.
+
+    }
+}
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SeedData.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SeedData.java
@@ -6,17 +6,13 @@ import com.appsmith.server.acl.AclPermission;
 import com.appsmith.server.constants.FieldName;
 import com.appsmith.server.domains.Config;
 import com.appsmith.server.domains.PermissionGroup;
-import com.appsmith.server.domains.PricingPlan;
-import com.appsmith.server.domains.Tenant;
 import com.appsmith.server.domains.Theme;
 import com.appsmith.server.domains.User;
 import com.appsmith.server.dtos.Permission;
 import com.appsmith.server.helpers.InMemoryCacheableRepositoryHelper;
 import com.appsmith.server.repositories.ConfigRepository;
 import com.appsmith.server.repositories.PermissionGroupRepository;
-import com.appsmith.server.repositories.TenantRepository;
 import com.appsmith.server.repositories.ThemeRepository;
-import com.appsmith.server.repositories.UserRepository;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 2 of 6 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/ce/R__finalMigrations.java
```diff
@@ -0,0 +1,16 @@
+package com.appsmith.server.migrations.ce;
+
+import com.appsmith.server.migrations.AppsmithJavaMigration;
+import org.springframework.jdbc.core.JdbcTemplate;
+
+public class R__finalMigrations extends AppsmithJavaMigration {
```
mohanarpit, line 6: @vivonk Out of curiosity, is there a naming convention here that I am not aware of? What does `R__` , `V4__`, etc mean?

vivonk: @mohanarpit The tool that we are using for postgres migration, Flyway, has this naming convention where `V` stands for version and then a text following that for ordering the execution of the migration like V1, V2, V3 etc. and the version can be any text.
Also, R means repeating migrations.
https://www.red-gate.com/blog/database-devops/flyway-naming-patterns-matter

app/server/appsmith-server/src/main/resources/sql/createIndexes.sql
```diff
@@ -1,6 +1,6 @@
 create unique index if not exists u_name on config(name);
 create unique index if not exists u_email on password_reset_token(email);
-create unique index if not exists u_package_name on plugin(package_name);
+-- create unique index if not exists u_package_name on plugin(package_name);
```
abhvsn, line 3: Do we want to remove this unique constraint?

abhvsn (commented):
Minor comments!

app/server/appsmith-server/src/main/resources/sql/createIndexes.sql
vivonk: yes, this needs to be removed. Removing it

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/RepositoryHelperMethods.java
```diff
@@ -14,4 +34,184 @@ public String getDefaultTenantId() {
...
+            tenant.setSlug(rs.getString("slug"));
+
+            // Convert JSON string to Set<Policy>
+            String policiesJson = rs.getString("policies");
+            Set<Policy> policies = mapObject(policiesJson, new TypeReference<Set<Policy>>() {});
+            tenant.setPolicies(policies == null ? new HashSet<>() : policies);
+            return tenant;
+        };
+        try {
+            return jdbcTemplate.queryForObject(sql, rowMapper);
```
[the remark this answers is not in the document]
vivonk: Yes, mapping is not happening. Need to define mappers ourselves. These methods work only for native data type returns.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/ce/V11__cleanPermissionsInPermissionGroups.java
```diff
@@ -0,0 +1,12 @@
+package com.appsmith.server.migrations.ce;
+
+import com.appsmith.server.migrations.AppsmithJavaMigration;
+import org.springframework.jdbc.core.JdbcTemplate;
+
+public class V11__cleanPermissionsInPermissionGroups extends AppsmithJavaMigration {
+    @Override
+    public void migrate(JdbcTemplate jdbcTemplate) throws Exception {
+        String sql = "UPDATE permission_group SET permissions = NULL";
```
sharat87, line 9: This clears the column for the whole table, no? Don't we need this column's data anymore? Can we remove it in today's `release` branch itself then? 🤔

sharat87 (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/ce/V11__cleanPermissionsInPermissionGroups.java
vivonk: We don't reuqire this column anymore, it's deprecated. Removing it from release is a legit solution but that will require efforts and understanding of 1-2 places where this field is used + some initial migrations require this field so removing it is not a easy task. 
This is a tech debt that we are carry forwarding from release to pg.

PR #33355, opened 2024-05-10
[ops]: add warning for custom APPSMITH_GIT_ROOT

pratapaprasanna:


sharat87 (approved):

deploy/docker/fs/opt/appsmith/utils/bin/backup.js
```diff
@@ -134,6 +134,11 @@ async function createGitStorageArchive(destFolder) {
 
   const gitRoot = getGitRoot(process.env.APPSMITH_GIT_ROOT);
 
+  if (gitRoot != "/appsmith-stacks/git-storage/"){
```
sharat87, line 137: This condition will always fail, even if custom git root isn't set. If you see the function `getGitRoot`, we have:

shot-2024-05-27-04-17-05

The default value there doesn't have a trailing `/`. Can we test and verify this once please?

sharat87 (changes requested):

deploy/docker/fs/opt/appsmith/utils/bin/backup.js
pratapaprasanna: Yea got it

sharat87 (approved):

PR #33374, opened 2024-05-13
fix: Changing font causes visual distortion; more noticeable than what used to happen on the base DP

### LINKED ISSUE #33331: [Bug]: Changing font causes visual distortion; more noticeable than what used to happen on the base DP

### Description

changing font causes visual distortion; more noticeable than what used to happen on the base DP

### Steps To Reproduce

1. Change app theme one by one as seen in the video and notice the visual "jiggle" widgets do to adjust spacing 

### Public Sample App

_No response_

### Environment

Deploy Preview

### Severity

Low (Cosmetic UI issues)

### Issue video log

https://www.loom.com/share/fd8ad97824324fe2b74c87a3d58bb526?sid=3539c3ff-46ff-4d26-a5d4-e1e9dbc44541

### Version

cloud

jsartisan:
Fixes #33331

### DIFF
--- a/app/client/packages/design-system/theming/src/hooks/src/useCssTokens.tsx
+++ b/app/client/packages/design-system/theming/src/hooks/src/useCssTokens.tsx
@@ -76,13 +76,11 @@ export function useCssTokens(props: Theme) {
         ${getTypographyCss(typography)}
       `);
     }
-  }, [typography]);
 
-  useEffect(() => {
     setFontFamilyClassName(css`
       ${fontFamilyCss(fontFamily)}
     `);
-  }, [fontFamily]);
+  }, [typography, fontFamily]);
 
   useEffect(() => {
     setProviderClassName(css`
--- a/app/client/src/pages/Editor/IDE/EditorPane/UI/Add.tsx
+++ b/app/client/src/pages/Editor/IDE/EditorPane/UI/Add.tsx
@@ -22,7 +22,7 @@ const AddWidgets = (props: { focusSearchInput?: boolean }) => {
         onCloseClick={closeButtonClickHandler}
         titleMessage={EDITOR_PANE_TEXTS.widgets_create_tab_title}
       />
-      <Flex flexDirection="column" gap="spaces-3" overflowX="scroll">
+      <Flex flexDirection="column" gap="spaces-3" overflowX="auto">
         <UIEntitySidebar focusSearchInput={props.focusSearchInput} isActive />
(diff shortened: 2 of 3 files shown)


app/client/packages/design-system/theming/src/hooks/src/useCssTokens.tsx
```diff
@@ -76,13 +76,11 @@ export function useCssTokens(props: Theme) {
...
       `);
     }
-  }, [typography]);
 
-  useEffect(() => {
     setFontFamilyClassName(css`
       ${fontFamilyCss(fontFamily)}
     `);
-  }, [fontFamily]);
+  }, [typography, fontFamily]);
```
jsartisan, line 83: @KelvinOm Moved the setting font and typography calculation under one useeffect. Hope it won't mess anything else.

KelvinOm: This won't help, since this change won't actually change anything. To be honest, I don't see a way to fix this behavior, because if we change Font Family, browser makes reflow, browser also makes reflow, since we have to apply new font parameters that capsize returned for us. These are two different processes and there is no way to merge them even if we apply all the changes at the same time. So I don't see a way to avoid this. That's how it works. 

We could make it more gentle for the user by hiding applications and showing the loader for a while. I.e. through UI/UX approach. 

cc @jsartisan @ichik @ramsaptami

ichik: Why do we need to apply the changes to font-family immediately instead of waiting for the capsize calculations? Also the visual jiggle seems to be from re-calculations of inner-spacing scale happening in the second pass rather than anything typographic.

KelvinOm: > Why do we need to apply the changes to font-family immediately instead of waiting for the capsize calculations?
It won't change this. We can apply it before, immediately, or after.

Here is an example if we apply font-family with a delay(100 ms).
https://github.com/appsmithorg/appsmith/assets/11555074/8d533e35-ea8f-4bd6-9e59-b9bd283e2b38

> Also the visual jiggle seems to be from re-calculations of inner-spacing scale happening in the second pass rather than anything typographic.
This is in particular due to the fact that we change the font size, line height and compensators (margins).

jsartisan:
Closing this PR as it does not solve the issue.

PR #33429, opened 2024-05-14
test: Enable server tests for the PRs with base PG branch

abhvsn:
## Description
1. PR to enable server tests on PRs with `pg` as the base branch. 
2. Replace the DB URI from `APPSMITH_MONGODB_URI` to `APPSMITH_DB_URL` 

## Automation

/ok-to-test tags="@tag.Sanity, @tag.GenerateCRUD, @tag.Fork"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9174148396>
> Commit: 762b4255f654946a1a47a196df5a1afae5be09f2
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9174148396&attempt=1" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [x] No


app/server/appsmith-server/src/main/resources/application-production.properties
```diff
@@ -1,2 +1,2 @@
 # MongoDB Application Database
-spring.data.mongodb.uri = ${APPSMITH_MONGODB_URI}
\ No newline at end of file
+spring.data.mongodb.uri=${APPSMITH_DB_URI:${APPSMITH_MONGODB_URI}}
```
mohanarpit, line 2: This file is not used. We should make this change to `application.properties` only.

app/server/build.sh
```diff
@@ -34,6 +34,12 @@ if [[ -f .env ]]; then
   source .env
 fi
 
+# Check APPSMITH_DB_URL environment variable for postgres and run the transform script for postgres databases
+if [[ "${APPSMITH_DB_URI}" == *"postgresql://"* ]]; then
+  # If it contains "postgres", run the transform script
+  python3 ./tx/transform.py
+fi
+
```
sharat87, line 42: This is a runtime env variable, and shouldn't influence build. Can we instead change the condition to check if the `transform.py` exists?

```suggestion
if [[ -f tx/transform.py ]]; then
  python3 tx/transform.py
fi

```

app/server/appsmith-server/src/main/resources/application.properties
```diff
@@ -7,6 +7,7 @@ server.max-http-request-header-size=16KB
 spring.lifecycle.timeout-per-shutdown-phase=20s
 
 spring.profiles.active=${ACTIVE_PROFILE:production}
+spring.data.mongodb.uri=${APPSMITH_DB_URI:${APPSMITH_MONGODB_URI}}
```
sharat87, line 10: Let's completely move awa from `APPSMITH_MONGODB_URI`. The backend server should _only_ look at the generic env variable, and the MongoDB specific one at all.

Instead, let's get this fallback functionality in the `entrypoint.sh`. There, we check if the new env variable has a value, otherwise, it's set to the MongoDB env variable's value, and the MongoDB env variable itself, is unset.

This way, we have consistent DB URL fallback implemented for all the services running the container, and not just for the backend.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration056UpdateDbUriEnvVariable.java
sharat87, line 1: Any reason we need this migration, when we have a fallback to read the old env variable anyway?

.github/workflows/server-build.yml
```diff
@@ -159,6 +170,11 @@ jobs:
           APPSMITH_ENVFILE_PATH: /tmp/dummy.env
           APPSMITH_VERBOSE_LOGGING_ENABLED: false
         run: |
+          if [[ "${{ inputs.is-pg-build }}" == "true" ]]; then
+            export APPSMITH_DB_URL="jdbc:postgresql://localhost:5432/postgres"
+          else
+            export APPSMITH_DB_URL="mongodb://localhost:27017/mobtools"
+          fi
           ./build.sh -DskipTests
```
sharat87, line 178: We're building without tests here, right? Is the DB URL needed?

app/server/appsmith-server/src/main/resources/application.properties
sharat87: Additionally, let's have `entrypoint.sh` add the `jdbc:` prefix to the URL if missing.

sharat87 (changes requested):
1. I see both `APPSMITH_DB_URL` and `APPSMITH_DB_URI`. Can we fix this and use `APPSMITH_DB_URL` please?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration056UpdateDbUriEnvVariable.java
abhvsn: @sharat87 added this to avoid the confusion in future, such that after specific version everyone will be using same env for DB URI. Also will be easier to communicate in our docs.

.github/workflows/server-build.yml
abhvsn: Build will fail on pg branch is we don't run the transform script right? That's the hard requirement for now, but I think after reading [this comment](https://github.com/appsmithorg/appsmith/pull/33429/files#r1602687555) we don't have to rely on this at all.

app/server/appsmith-server/src/main/resources/application.properties
abhvsn: @sharat87 updated the entrypoint.sh with the changes requested, can you please take a look. Also, I was thinking if this fallback will help us with local testing though or should we just update the `test.env.example` and ask devs to switch to the updated env manually as we don't run entrypoint for local testing.

sharat87: I'm thinking let's change and announce. Announce to folks that **for local development only**, we need to use the new env variable, and the old one will not work anymore. But the container has backwards compatibility, in that **our users running the container**, don't have to do anything.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration056UpdateDbUriEnvVariable.java
sharat87: I get the problem being solved, but I slightly disagree that it should be solved. The `docker.env` is just one place for env variables. They're also set from ConfigMaps for Kubernetes, in `docker-compose.yml` file, in ECS configuration on AWS console, etc. All we can do here, is update the `docker.env`, which doesn't afford us any comfort that everyone has moved on to the newer env variable.

We'll have to keep the fallback to the old env variable for a _long long_ time to come, anyway. 🙂

app/server/appsmith-server/src/main/resources/application.properties
```diff
@@ -7,6 +7,7 @@ server.max-http-request-header-size=16KB
 spring.lifecycle.timeout-per-shutdown-phase=20s
 
 spring.profiles.active=${ACTIVE_PROFILE:production}
+spring.data.mongodb.uri=${APPSMITH_DB_URL:${APPSMITH_MONGODB_URI}}
```
sharat87, line 10: Let's remove the fallback here, and not have a default value at all. Just like in the old `application-production-properties`.

Rationale at https://github.com/appsmithorg/appsmith/pull/33429#discussion_r1603660557.

deploy/docker/fs/opt/appsmith/entrypoint.sh
```diff
@@ -507,6 +520,7 @@ export APPSMITH_LOG_DIR="${APPSMITH_LOG_DIR:-/appsmith-stacks/logs}"
 mkdir -p "$APPSMITH_LOG_DIR"/{supervisor,backend,cron,editor,rts,mongodb,redis,postgres,appsmithctl}
 
 setup_auto_heal
+configure_database_connection_url
```
sharat87, line 523: Let's move this to above `chece_mongodb_uri`, so that, just in case, the new env variable is available before we do anything with the Database.

sharat87 (changes requested):

app/server/appsmith-server/src/main/resources/application.properties
abhvsn: @sharat87 Does the change in `entrypoint.sh` looks good to you for backward compatibility? If we are aligned I'll remove the migration and rely on `APPSMITH_DB_URL` in `application.properties`.

.github/workflows/server-build.yml
```diff
@@ -133,6 +138,12 @@ jobs:
           distribution: "temurin"
           java-version: "17"
 
+      - name: Conditionally start PostgreSQL
+        run: |
+          if [[ inputs.is-pg-build == 'true' ]]; then
+            docker run -d -p 5432:5432 postgres:alpine
```
sharat87, line 144: Can we verify if this command is working? I tried locally and it doesn't seem to be.

shot-2024-05-16-16-20-26

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration056UpdateDbUriEnvVariable.java
abhvsn: Got it aligned on the approach, I was looking from local dev perspective. I'll continue with the suggestion provided in other thread https://github.com/appsmithorg/appsmith/pull/33429#discussion_r1603660557.

deploy/docker/fs/opt/appsmith/entrypoint.sh
abhvsn: Done, I have made updates in few other places.

.github/workflows/server-build.yml
```diff
@@ -175,6 +175,11 @@ jobs:
           APPSMITH_ENVFILE_PATH: /tmp/dummy.env
           APPSMITH_VERBOSE_LOGGING_ENABLED: false
         run: |
+          if [[ "${{ inputs.is-pg-build }}" == "true" ]]; then
+            export APPSMITH_DB_URL="jdbc:postgresql://localhost:5432/postgres"
```
sharat87, line 179: We'll need to set the username/password here now no? Something like:

```suggestion
            export APPSMITH_DB_URL="jdbc:postgresql://postgres:password@localhost:5432/postgres"
```

app/server/appsmith-server/src/main/resources/application.properties
```diff
@@ -7,6 +7,7 @@ server.max-http-request-header-size=16KB
 spring.lifecycle.timeout-per-shutdown-phase=20s
 
 spring.profiles.active=${ACTIVE_PROFILE:production}
+spring.data.mongodb.uri=${APPSMITH_DB_URL:}
```
sharat87, line 10: Let's please not have a default value here. If this is empty, server startup will fail, so might as well fail earlier here itself. Just like how it was in `application-production.properties`.

```suggestion
spring.data.mongodb.uri=${APPSMITH_DB_URL}
```

deploy/docker/fs/opt/appsmith/entrypoint.sh
```diff
@@ -477,11 +498,19 @@ print_appsmith_info
...
 if [[ -z "${DYNO}" ]]; then
-  # Don't run MongoDB if running in a Heroku dyno.
-  init_mongodb
-  init_replica_set
+  if [[ $isMongoUrl -eq 1 ]]; then
+    # Setup MongoDB and initialize replica set
+    echo "Initializing MongoDB"
+    init_mongodb
+    init_replica_set
+  elif [ $isPostgresUrl -eq 1 ]; then
```
sharat87, line 510:
```suggestion
  elif [[ $isPostgresUrl -eq 1 ]]; then
```

deploy/docker/fs/opt/appsmith/entrypoint.sh
```diff
@@ -168,11 +168,32 @@ unset_unused_variables() {
...
+
+  if [[ "${APPSMITH_DB_URL}" == *"postgresql"* ]]; then
+    isPostgresUrl=1
+    # Check if APPSMITH_DB_URL is a PostgreSQL URL and doesn't start with "jdbc:", prepend "jdbc:"
+    if [[ "${APPSMITH_DB_URL}" != "jdbc:"* ]]; then
+      export APPSMITH_DB_URL="jdbc:${APPSMITH_DB_URL}"
+    fi
+  elif [[ "${APPSMITH_DB_URL}" == *"mongodb"* ]]; then
+    isMongoUrl=1
+  fi
```
sharat87, line 189: Let's make this stricter and define a fallthrough `else` as well. Something like:

```suggestion
  if [[ "${APPSMITH_DB_URL}" == "jdbc:postgresql:"* ]]; then
    isPostgresUrl=1
  elif [[ "${APPSMITH_DB_URL}" == "postgresql:"* ]]; then
    isPostgresUrl=1
    export APPSMITH_DB_URL="jdbc:${APPSMITH_DB_URL}"
  elif [[ "${APPSMITH_DB_URL}" == "mongodb"* ]]; then
    isMongoUrl=1
  fi

```

sharat87 (changes requested):

.github/workflows/server-build.yml
abhvsn: @sharat87 it's failing for the above connection string in my local, but adding a separate username and password works. Let me add 
```
Connection Fail:
spring.datasource.url=jdbc:postgresql://postgres:password@localhost:5432/postgres

Connection successful:
spring.datasource.url=jdbc:postgresql://localhost:5432/postgres
spring.datasource.username=postgres
spring.datasource.password=password
```

From little search I think `jdbc:postgresql://hostname:port/database` is the valid format. I think we need to add more env variable to fix this issue.

abhvsn: Created another PR pointing to PG branch to address this issue https://github.com/appsmithorg/appsmith/pull/33543

abhvsn: @sharat87 extracting username and password worked in my local please check app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/CommonDBConfig.java, we can live with single env APPSMITH_DB_URL for both Mongo and Postgres.

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/CommonDBConfig.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Method to extract Jdbc props from the given DB URL
+     * Expected DB URL: jdbc:postgresql://{username}:{password}@localhost:{port}/{db_name}
+     */
+    public DataSourceProperties extractJdbcProperties(String dbUrl) {
+        DataSourceProperties ds = new DataSourceProperties();
+        try {
+            URI uri = new URI(dbUrl.substring(5)); // remove "jdbc:" prefix
```
sharat87, line 55: Looks like we're stripping the `jdbc:` prefix here, and aren't really using it. Perhaps we don't have to add it to the env variable in the first place? We didn't have it with MongoDB, we don't need it with Postgres, so might as well not add it, right? What say?

sharat87 (changes requested):
Looks good. One more small recommendation and let's get the DP ready. Thanks!

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/CommonDBConfig.java
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Method to extract Jdbc props from the given DB URL
+     * Expected DB URL: postgresql://{username}:{password}@localhost:{port}/{db_name}
+     */
+    public DataSourceProperties extractJdbcProperties(String dbUrl) {
+        DataSourceProperties ds = new DataSourceProperties();
+        try {
+            // remove "jdbc:" prefix if exists
+            URI uri = new URI(dbUrl.replaceFirst("^" + JDBC_PREFIX, ""));
```
sharat87, line 60: Can we completely remove this? Do we expect cases where the `jdbc:` prefix might be there? We can fail in that case saying it's not a valid URL, right?

deploy/docker/fs/opt/appsmith/utils/bin/backup.js
```diff
@@ -207,7 +207,7 @@ function removeSensitiveEnvData(content) {
   // Remove encryption and Mongodb data from docker.env
   const output_lines = []
   content.split(/\r?\n/).forEach(line => {
-    if (!line.startsWith("APPSMITH_ENCRYPTION") && !line.startsWith("APPSMITH_MONGODB")) {
+    if (!line.startsWith("APPSMITH_ENCRYPTION") && !line.startsWith("APPSMITH_MONGODB") && !line.startsWith("APPSMITH_DB")) {
```
sharat87, line 210: For the other two, they're prefixes of _more than one_ env variable. But there's only one env variable that starts with `APPSMITH_DB`, so let's make it specific.

```suggestion
    if (!line.startsWith("APPSMITH_ENCRYPTION") && !line.startsWith("APPSMITH_MONGODB") && !line.startsWith("APPSMITH_DB_URL=")) {
```

sharat87 (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/CommonDBConfig.java
abhvsn: This was added just to support both the url format as I have seen some discussion in couple of threads, so thought of supporting this.

deploy/docker/fs/opt/appsmith/utils/bin/backup.js
abhvsn: I thought in future we may want to move common envs from `APPSMITH_MONGODB*` to more generic `APPSMITH_DB*` hence kept only the prefix. But let's come back to this when required.

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/CommonDBConfig.java
sharat87: I'd say we add support for ignoring that prefix, if its needed in the future. It's _very_ hard to remove something once released, but we can always add support for this later. Since supporting URLs with that prefix isn't adding any value to us today (correct me if I'm wrong here), I'd rather we require URLs to not have `jdbc:` prefix, both for MongoDB and Postgres.

And besides, if we later add support for allowing the `jdbc:` prefix, we do it consistent. We allow it for _both_ Postgres and MongoDB URLs. But that's for later, I'd rather we don't allow it at all today.

Let me know if you disagree. 🙂

abhvsn: Aligned, updated the implementation.

sharat87 (approved):

PR #33433, opened 2024-05-14
chore: Improve Link component

jsartisan:
This PR uses react aria link's component in our Link component

### DIFF
--- a/app/client/packages/design-system/widgets/src/components/Link/src/Link.tsx
+++ b/app/client/packages/design-system/widgets/src/components/Link/src/Link.tsx
@@ -8,7 +8,6 @@ import styles from "./styles.module.css";
 export function Link(props: LinkProps) {
   const {
     children,
-    color = "accent",
     download,
     href,
     ping,
@@ -19,7 +18,7 @@ export function Link(props: LinkProps) {
   } = props;
 
   return (
-    <Text color={color} {...rest}>
+    <Text color="accent" {...rest}>
       <HeadlessLink
         className={styles.link}
         download={download}
--- a/app/client/packages/design-system/widgets/src/components/Link/src/Link.tsx
+++ b/app/client/packages/design-system/widgets/src/components/Link/src/Link.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { Link as AriaLink } from "react-aria-components";
+import { Link as HeadlessLink } from "react-aria-components";
 
 import { Text } from "../../Text";
 import type { LinkProps } from "./types";
@@ -8,6 +8,7 @@ import styles from "./styles.module.css";
 export function Link(props: LinkProps) {
   const {
     children,
+    color = "accent",
     download,
     href,
     ping,
@@ -18,8 +19,8 @@ export function Link(props: LinkProps) {
   } = props;
 
   return (
-    <Text {...rest}>
-      <AriaLink
+    <Text color={color} {...rest}>
+      <HeadlessLink
         className={styles.link}
         download={download}
         href={href}
@@ -29,7 +30,7 @@ export function Link(props: LinkProps) {
         target={target}
       >
         {children}
-      </AriaLink>
+      </HeadlessLink>
     </Text>
   );
 }
--- a/app/client/packages/design-system/widgets/src/components/Link/src/Link.tsx
+++ b/app/client/packages/design-system/widgets/src/components/Link/src/Link.tsx
@@ -10,7 +10,6 @@ export function Link(props: LinkProps) {
     children,
     download,
     href,
-    isDisabled,
     ping,
     referrerPolicy,
     rel,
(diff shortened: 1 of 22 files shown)


app/client/packages/storybook/.storybook/preview-head.html
```diff
@@ -28,7 +28,8 @@
     border: none !important;
   }
 
-  a {
+  /* Note: checking for data-rac attribute to avoid overriding the color of our Link component */
+  a:not([data-rac]) {
     color: var(--color-fg-accent) !important;
```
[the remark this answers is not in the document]
jsartisan: Removing important will make the anchor in props table as blue

CleanShot 2024-05-16 at 11 40 57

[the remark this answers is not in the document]
jsartisan: The` !important` is required here i suppose so that auto-generated styles on the anchor can be overridden.

Instead of `a:not([data-rac]) `, we can use  `a:not(.docs-story a)`. What do you think about this?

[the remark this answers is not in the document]
KelvinOm: You can redefine the styles for Storybrook in the `preview-head.html`. There shouldn't be anything in the component code that works for storybook.

app/client/packages/design-system/widgets/src/components/Link/src/Link.tsx
```diff
@@ -1,17 +1,36 @@
...
 
 import { Text } from "../../Text";
 import type { LinkProps } from "./types";
 import styles from "./styles.module.css";
 
 export function Link(props: LinkProps) {
-  const { children, href, rel, target, ...rest } = props;
+  const {
+    children,
+    color = "accent",
```
KelvinOm, line 11: We discussed this. This prop should be deleted and the link should be accent by default.

KelvinOm (changes requested):

app/client/packages/storybook/.storybook/preview-head.html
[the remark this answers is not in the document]
jsartisan: > You can redefine the styles for Storybrook in the preview-head.html

That's what I did ( if you the file we is preview-head only above ). 🤔 There is no specific code that I wrote in the component to satisfy storybook.

app/client/packages/design-system/widgets/src/components/Link/src/Link.tsx
jsartisan: Removed the prop.

app/client/src/widgets/wds/WDSTableWidget/component/cellComponents/URLCell.tsx
```diff
@@ -16,12 +16,11 @@ export interface NumberCellProps {
 }
 
 function URLCell(props: NumberCellProps & BaseCellComponentProps) {
-  const { allowCellWrapping, cellColor, href, isBold, isItalic, text } = props;
+  const { allowCellWrapping, href, isBold, isItalic, text } = props;
```
jsartisan, line 19: Had to do this because i forked out the branch from the table polish branch.

app/client/packages/storybook/.storybook/preview-head.html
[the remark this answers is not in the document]
KelvinOm: My bad.

app/client/packages/design-system/widgets/src/components/Link/stories/Link.stories.tsx
```diff
@@ -0,0 +1,79 @@
...
+  args: {
+    target: "_blank",
+    href: "https://appsmith.com",
+    lineClamp: 1,
+    children:
+      "Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged.",
+  },
+};
+
+export const Variant: Story = {
```
KelvinOm, line 41: Honestly variant here looks odd. Here for variant we more expect something like a primary|secondary. In fact, we have only 1 variant for link, and this is more about font sizes. Lets rename it to `size`? WDYT?

jsartisan: The link is nothing but Text at the end ( with a href link ).  So the prop names should be similar IMO. In one component ( Text ) we have a `variant` prop, and now in another component ( Link ), we are proposing `size` for the same thing. Maybe we should rename the prop in the text too. Size prop name makes more sense to control the size.

KelvinOm: Yep same thoughts for text component. Let's rename it.

Also in app UI, we use this not as an variant but as a size.
Снимок экрана 2024-05-16 в 15 05 05

jsartisan: Done.

KelvinOm (approved):

PR #33439, opened 2024-05-14
chore: cache templates data

AnaghHegde:
## Description
Cache the templates data to improve the turn around time

Fixes https://github.com/appsmithorg/appsmith/issues/33440

## Automation

/ok-to-test tags="@tag.Templates"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9368135903>
> Commit: 5cbcbd353519ac523a2268cba54815473babead6
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9368135903&attempt=1" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
  - Introduced caching functionality for application templates and data to improve performance.

- **Bug Fixes**
  - Updated error handling for cloud services errors, changing the HTTP status code from 500 to 400.

- **Tests**
  - Added test cases to ensure data retrieval accuracy and validate cache utilization for application templates.


app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CacheableTemplateHelper.java
```diff
@@ -0,0 +1,88 @@
...
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+import java.lang.reflect.Type;
+import java.time.Instant;
+import java.util.List;
+
+@AllArgsConstructor
+public class CacheableTemplateHelper {
+    @Cache(cacheName = "templateMetadata", key = "{#releaseVersion}")
```
mohanarpit, line 29: Since we are using `releaseVersion` as the key, what happens in scenarios where we add a template and the user hasn't updated their Appsmith version? Will they not see the new templates?

AnaghHegde: @mohanarpit Every template that we store in CS has a attribute called min version to identify whether the templates can be rendered in the given Appsmith instance or not. So yes depending the Appsmith instance the newly added template might not be available on some Appsmith instance. 
This was already present and you can see this being passed in the CS API call.

mohanarpit: Disclaimer: I understand this is a draft PR and hence please take this comment as a direction to think in. I think we are talking about different things here. 

I see there is code for caching the templates in the Appsmith server, but there is no code to invalidate this cache. Let's say there is an Appsmith instance v1 deployed. We have 5 templates with min version v1. They will be returned to the Appsmith server & cached there. Now, we create another template with min version v1. So we have 6 templates in total. But since the cache will return the older data, this instance will never see the new template. 

Basically, when will this cache be invalidated?

AnaghHegde: I am aware of the scenario you have mentioned. I have raised a PR to create DP and test the changes. Its not fully done.

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CacheableTemplateHelper.java
```diff
@@ -0,0 +1,111 @@
...
+                    cacheableApplicationJson.setLastUpdated(Instant.now());
+                    return cacheableApplicationJson;
+                })
+                .switchIfEmpty(
+                        Mono.error(new AppsmithException(AppsmithError.NO_RESOURCE_FOUND, "template", templateId)));
+    }
+
+    @CacheEvict(cacheName = "templateMetadata", key = "{#releaseVersion}")
+    public static Mono<Void> clearTemplateMetadataCache(String releaseVersion) {
+        return Mono.empty();
```
NilanshBansal, line 104: Nit: Can you please use `Mono.empty().then()` here to avoid returning a completed Mono.

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CacheableTemplateHelper.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @CacheEvict(cacheName = "templateApplicationData", key = "{#templateId}")
+    public static Mono<Void> clearTemplateApplicationDataCache(String templateId) {
+        return Mono.empty();
```
NilanshBansal, line 109: Same as above. Can you please use `Mono.empty().then()` here to avoid returning a completed Mono.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
```diff
@@ -28,29 +29,28 @@
...
 
-import java.lang.reflect.Type;
 import java.time.Instant;
 import java.util.List;
 import java.util.Map;
 
+import static com.appsmith.server.helpers.CacheableTemplateHelper.clearTemplateApplicationDataCache;
+import static com.appsmith.server.helpers.CacheableTemplateHelper.clearTemplateMetadataCache;
+import static com.appsmith.server.helpers.CacheableTemplateHelper.getApplicationByTemplateId;
+import static com.appsmith.server.helpers.CacheableTemplateHelper.getTemplates;
```
NilanshBansal, line 52: Instead of all these specific imports can you please just import the `CacheableTemplateHelper` and use `cacheableTemplateHelper.funcName()` wherever we are using it.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
```diff
@@ -114,31 +116,17 @@ public Flux<ApplicationTemplate> getSimilarTemplates(String templateId, MultiVal
...
-                    if (clientResponse.statusCode().equals(HttpStatus.OK)) {
-                        return clientResponse.bodyToFlux(ApplicationTemplate.class);
-                    } else if (clientResponse.statusCode().isError()) {
-                        return Flux.error(
-                                new AppsmithException(AppsmithError.CLOUD_SERVICES_ERROR, clientResponse.statusCode()));
+        return getTemplates(releaseNotesService.getRunningVersion(), cloudServicesConfig.getBaseUrl())
+                .flatMap(templates -> {
+                    if (Instant.now().minusSeconds(CACHE_LIFE_TIME_IN_SECONDS).isAfter(templates.getLastUpdated())) {
+                        return clearTemplateMetadataCache(releaseNotesService.getRunningVersion())
+                                .then(Mono.defer(() -> getTemplates(
```
NilanshBansal, line 123: Can you please use `cacheableTemplateHelper.getTemplates()` here.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
```diff
@@ -114,31 +116,17 @@ public Flux<ApplicationTemplate> getSimilarTemplates(String templateId, MultiVal
...
-                .exchangeToFlux(clientResponse -> {
-                    if (clientResponse.statusCode().equals(HttpStatus.OK)) {
-                        return clientResponse.bodyToFlux(ApplicationTemplate.class);
-                    } else if (clientResponse.statusCode().isError()) {
-                        return Flux.error(
-                                new AppsmithException(AppsmithError.CLOUD_SERVICES_ERROR, clientResponse.statusCode()));
+        return getTemplates(releaseNotesService.getRunningVersion(), cloudServicesConfig.getBaseUrl())
+                .flatMap(templates -> {
+                    if (Instant.now().minusSeconds(CACHE_LIFE_TIME_IN_SECONDS).isAfter(templates.getLastUpdated())) {
+                        return clearTemplateMetadataCache(releaseNotesService.getRunningVersion())
```
NilanshBansal, line 122: Can you please use `cacheableTemplateHelper.clearTemplateMetadataCache()` here.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
NilanshBansal, line 123: Can you add a `onError()` block to handle the errors thrown by the `getTemplates()` function. This error catch block could help in handling deserialization errors.
Code Ref: https://github.com/appsmithorg/appsmith/commit/d11d404f3c06109859282601d7bfed4f8feb1922

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
```diff
@@ -162,39 +150,17 @@ public Mono<ApplicationTemplate> getTemplateDetails(String templateId) {
...
-                    return jsonFile;
-                })
-                .switchIfEmpty(
-                        Mono.error(new AppsmithException(AppsmithError.NO_RESOURCE_FOUND, "template", templateId)));
+        return getApplicationByTemplateId(templateId, baseUrl).flatMap(cacheableApplicationJson -> {
+            if (Instant.now()
+                    .minusSeconds(CACHE_LIFE_TIME_IN_SECONDS)
+                    .isAfter(cacheableApplicationJson.getLastUpdated())) {
+                return clearTemplateApplicationDataCache(templateId)
+                        .then(Mono.defer(() -> getApplicationByTemplateId(templateId, baseUrl)))
```
NilanshBansal, line 158: Can you add a `onError()` block to handle the errors thrown by the `getApplicationByTemplateId()` function. You have explicitly thrown an error from the getApplicationByTemplateId() function from [switchIfEmpty()](https://github.com/appsmithorg/appsmith/blob/ad560291b33fb7f4d2bfa9133727452f46da11dc/app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CacheableTemplateHelper.java#L98). 
This error catch block will also help in handling deserialization errors.
Code Ref: https://github.com/appsmithorg/appsmith/commit/d11d404f3c06109859282601d7bfed4f8feb1922

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
```diff
@@ -162,39 +150,17 @@ public Mono<ApplicationTemplate> getTemplateDetails(String templateId) {
...
-                    ApplicationJson jsonFile = gson.fromJson(jsonString, fileType);
-                    return jsonFile;
-                })
-                .switchIfEmpty(
-                        Mono.error(new AppsmithException(AppsmithError.NO_RESOURCE_FOUND, "template", templateId)));
+        return getApplicationByTemplateId(templateId, baseUrl).flatMap(cacheableApplicationJson -> {
+            if (Instant.now()
+                    .minusSeconds(CACHE_LIFE_TIME_IN_SECONDS)
+                    .isAfter(cacheableApplicationJson.getLastUpdated())) {
+                return clearTemplateApplicationDataCache(templateId)
```
NilanshBansal, line 157: Can you please use `cacheableTemplateHelper.clearTemplateApplicationDataCache()` here.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
```diff
@@ -162,39 +150,17 @@ public Mono<ApplicationTemplate> getTemplateDetails(String templateId) {
...
-                            .registerTypeAdapter(Instant.class, new ISOStringToInstantConverter())
-                            .create();
-                    Type fileType = new TypeToken<ApplicationJson>() {}.getType();
-
-                    ApplicationJson jsonFile = gson.fromJson(jsonString, fileType);
-                    return jsonFile;
-                })
-                .switchIfEmpty(
-                        Mono.error(new AppsmithException(AppsmithError.NO_RESOURCE_FOUND, "template", templateId)));
+        return getApplicationByTemplateId(templateId, baseUrl).flatMap(cacheableApplicationJson -> {
```
NilanshBansal, line 153: Can you please use `cacheableTemplateHelper.getApplicationByTemplateId()` here.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
NilanshBansal, line 158: Can you please use `cacheableTemplateHelper.getApplicationByTemplateId()` here.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
```diff
@@ -114,31 +116,17 @@ public Flux<ApplicationTemplate> getSimilarTemplates(String templateId, MultiVal
...
-
-        return WebClientUtils.create(baseUrl + "/api/v1/app-templates?" + uriComponents.getQuery())
-                .get()
-                .exchangeToFlux(clientResponse -> {
-                    if (clientResponse.statusCode().equals(HttpStatus.OK)) {
-                        return clientResponse.bodyToFlux(ApplicationTemplate.class);
-                    } else if (clientResponse.statusCode().isError()) {
-                        return Flux.error(
-                                new AppsmithException(AppsmithError.CLOUD_SERVICES_ERROR, clientResponse.statusCode()));
+        return getTemplates(releaseNotesService.getRunningVersion(), cloudServicesConfig.getBaseUrl())
```
NilanshBansal, line 119: Can you please use `cacheableTemplateHelper.getTemplates()` here.

NilanshBansal (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
AnaghHegde: These are static methods and the standard practice is to use the method name directly by importing each method as static imports.

NilanshBansal: My bad. Thanks for the explanation @AnaghHegde and apologies for the inconvenience.

NilanshBansal:
Adding @abhvsn also as a reviewer to get more eyes to look on this.

app/server/appsmith-server/src/main/java/com/appsmith/server/imports/internal/partial/PartialImportServiceCEImpl.java
```diff
@@ -411,14 +414,18 @@ public Mono<BuildingBlockResponseDTO> importBuildingBlock(BuildingBlockDTO build
         Mono<ApplicationJson> applicationJsonMono =
                 applicationTemplateService.getApplicationJsonFromTemplate(buildingBlockDTO.getTemplateId());
 
+        Stopwatch processStopwatch = new Stopwatch("Download Content from Cloud service");
         return applicationJsonMono.flatMap(applicationJson -> {
+            processStopwatch.stopAndLogTimeInMillis();
+            Stopwatch processStopwatch1 = new Stopwatch("Importing resource in db ");
```
NilanshBansal, line 420: @AnaghHegde why have we used this? Was this only for internal local env monitoring or you want to ship this code to prod to log into mezmo? 
I believe we can remove these extra logs if not required.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ApplicationTemplateServiceUnitTest.java
```diff
@@ -71,6 +72,9 @@ public class ApplicationTemplateServiceUnitTest {
     @MockBean
     private SessionUserService sessionUserService;
 
+    @MockBean
+    private CacheableTemplateHelper cacheableTemplateHelper;
```
NilanshBansal, line 76: @AnaghHegde What is this being used for in the test case?

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ApplicationTemplateServiceUnitTest.java
```diff
@@ -101,7 +105,8 @@ public void initialize() {
                 responseUtils,
                 applicationPermission,
                 objectMapper,
-                sessionUserService);
+                sessionUserService,
+                cacheableTemplateHelper);
     }
 
     private ApplicationTemplate create(String id, String title) {
```
NilanshBansal, line 112: @AnaghHegde Can we add some tests to assert this implementation?
The tests could be asserting the redis key and also the regular application JSON and template validation checks.

NilanshBansal (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CacheableTemplateHelper.java
```diff
@@ -0,0 +1,99 @@
...
+import org.springframework.web.util.UriComponentsBuilder;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+import java.time.Instant;
+
+@AllArgsConstructor
+@Component
+public class CacheableTemplateHelper {
+    @Cache(cacheName = "templateMetadata", key = "{#releaseVersion}")
```
NilanshBansal, line 27: @AnaghHegde can you please add a small comment on what this redis key stores. `templateMetadata` and `templateApplicationData` looks similar to me without diving deep in the code. 
Can you please add the comment from a user perspective what is getting stored here.

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CacheableTemplateHelper.java
```diff
@@ -0,0 +1,99 @@
...
+                .collectList()
+                .map(applicationTemplates -> {
+                    CacheableApplicationTemplate cacheableApplicationTemplate = new CacheableApplicationTemplate();
+                    cacheableApplicationTemplate.setApplicationTemplateList(applicationTemplates);
+                    cacheableApplicationTemplate.setLastUpdated(Instant.now());
+                    return cacheableApplicationTemplate;
+                });
+    }
+
+    @Cache(cacheName = "templateApplicationData", key = "{#templateId}")
```
NilanshBansal, line 56: @AnaghHegde can you please add a small comment on what this redis key stores. `templateMetadata` and `templateApplicationData` looks similar to me without diving deep in the code. 
Can you please add the comment from a user perspective what is getting stored here.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ApplicationTemplateServiceUnitTest.java
AnaghHegde: This was an existing test class. If you look at the code below this, this was a dependency to create an object. Hence added this after this was introduced in `ApplicationTemplateService`

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ApplicationTemplateServiceUnitTest.java
AnaghHegde: I have different opinion here, the JSON validation is already done in multiple places before importing. I don't see the value in doing this. Adding test for redis to verify if the keys are stored in redis or not? I don't think thats required here, we have tests for @Cache annotation already to verify that in the redis cache manager module.

app/server/appsmith-server/src/main/java/com/appsmith/server/imports/internal/partial/PartialImportServiceCEImpl.java
AnaghHegde: @NilanshBansal For the time we need this. It added on purpose.

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CacheableTemplateHelper.java
NilanshBansal: @AnaghHegde can you please respond here, why we have resolved this comment without any info or a new commit adding this?

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CacheableTemplateHelper.java
NilanshBansal: @AnaghHegde can you please respond here, why we have resolved this comment without any info or a new commit adding this?

NilanshBansal:
@AnaghHegde server-unit-tests are failing can you please check?

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CacheableTemplateHelper.java
```diff
@@ -0,0 +1,99 @@
...
+        UriComponents uriComponents = uriComponentsBuilder.build();
+
+        return WebClientUtils.create(baseUrl + "/api/v1/app-templates?" + uriComponents.getQuery())
+                .get()
+                .exchangeToFlux(clientResponse -> {
+                    if (clientResponse.statusCode().equals(HttpStatus.OK)) {
+                        return clientResponse.bodyToFlux(ApplicationTemplate.class);
+                    } else if (clientResponse.statusCode().isError()) {
+                        return Flux.error(
+                                new AppsmithException(AppsmithError.CLOUD_SERVICES_ERROR, clientResponse.statusCode()));
```
abhvsn, line 42: @AnaghHegde Should we log actual error instead of just passing the status code?

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CacheableTemplateHelper.java
```diff
@@ -0,0 +1,99 @@
...
+                    cacheableApplicationJson.setLastUpdated(Instant.now());
+                    return cacheableApplicationJson;
+                })
+                .switchIfEmpty(
+                        Mono.error(new AppsmithException(AppsmithError.NO_RESOURCE_FOUND, "template", templateId)));
+    }
+
+    @CacheEvict(cacheName = "templateMetadata", key = "{#releaseVersion}")
+    public Mono<Void> clearTemplateMetadataCache(String releaseVersion) {
+        return Mono.empty().then();
```
abhvsn, line 92:
```suggestion
        return Mono.empty();
```

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ApplicationTemplateServiceUnitTest.java
abhvsn: @AnaghHegde we are not questioning the caching behaviour but the consumption of it in templates service. For Consider in future if someone adds new nested object which should be part of cached data and forgot to add the Serializable implementation marker which can lead to failure in (de)serialisation. Having the test to assert this functionality can help in catching those issues earlier.

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CacheableTemplateHelper.java
AnaghHegde: I have updated the cacheKeyName, please have a look at the new comment. The name meta is sufficient here.

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CacheableTemplateHelper.java
AnaghHegde: As mentioned above

AnaghHegde:
@abhvsn I have considered the scenario of new attributes being part of the applicationJson object. Hence the data stored in redis is a json `String` and not a java object. If you still think i need to add a test case to assert this behaviour, i would be happy to add a test case.

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CacheableTemplateHelper.java
AnaghHegde: This was the original implementation i had. @NilanshBansal suggested this change, in here - https://github.com/appsmithorg/appsmith/pull/33439#discussion_r1610993634

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CacheableTemplateHelper.java
AnaghHegde: Good catch. I agree with you on this. I will update this error handling.

abhvsn:
> @abhvsn I have considered the scenario of new attributes being part of the applicationJson object. Hence the data stored in redis is a json `String` and not a java object. If you still think i need to add a test case to assert this behaviour, i would be happy to add a test case.

@AnaghHegde I can see we are storing CacheableApplicationTemplate which is a custom object. Also in case of `CacheableApplicationJson` as of today it's true what if tomorrow we introduce some meta object which we want to store alongwith the `applicationJson`, I know this may not be required but a test will give us the confidence that this feature will not break because of any (de)serialization issue.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
```diff
@@ -162,39 +168,37 @@ public Mono<ApplicationTemplate> getTemplateDetails(String templateId) {
...
+                        return cacheableTemplateHelper
+                                .clearTemplateApplicationDataCache(templateId)
+                                .then(Mono.defer(
+                                        () -> cacheableTemplateHelper.getApplicationByTemplateId(templateId, baseUrl)))
+                                .map(CacheableApplicationJson::getApplicationJson);
+                    } else {
+                        return Mono.just(cacheableApplicationJson.getApplicationJson());
+                    }
+                })
+                .onErrorResume(e -> {
```
mohanarpit, line 186: What happens in the scenario that the `templateId` is actually invalid? If there is no template available with that particular templateId, how does CS respond?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
```diff
@@ -162,39 +168,37 @@ public Mono<ApplicationTemplate> getTemplateDetails(String templateId) {
...
-                .get()
-                .retrieve()
-                .bodyToMono(String.class)
-                .map(jsonString -> {
+        return cacheableTemplateHelper
+                .getApplicationByTemplateId(templateId, baseUrl)
+                .flatMap(cacheableApplicationJson -> {
+                    if (Instant.now()
+                            .minusSeconds(CACHE_LIFE_TIME_IN_SECONDS)
+                            .isAfter(cacheableApplicationJson.getLastUpdated())) {
```
mohanarpit, line 176: Thought: If we are storing the template in Redis cache for 24 hours and the Java server is actually pulling the data (without modifying it), is there a need for Redis at all? Can we not store the object as a POJO in memory and simply return that POJO?

AnaghHegde: @mohanarpit That is a valid point. I did think of that and i thought using redis annotation would be easier than this. But i realised about the deserialisation issue and no option to configure time to live after writing the code 😞

mohanarpit: Is there something that prevents us from changing the code right now?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
```diff
@@ -162,39 +168,37 @@ public Mono<ApplicationTemplate> getTemplateDetails(String templateId) {
...
-
-        return webClient
-                .get()
-                .retrieve()
-                .bodyToMono(String.class)
-                .map(jsonString -> {
+        return cacheableTemplateHelper
+                .getApplicationByTemplateId(templateId, baseUrl)
+                .flatMap(cacheableApplicationJson -> {
+                    if (Instant.now()
```
mohanarpit, line 174: Btw, this is not the right way to store time sensitive data in Redis cache. Unfortunately, it's brittle and error prone. 

Redis keys have a property called TTL (Time to live). Redis will automatically delete that key after the specified time period. In order for us to leverage that, we need to enhance the RedisCacheManagerImpl and annotations/Cache.java files. It's a small change that will lead to a much cleaner and longer term code.

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/PageNameIdDTO.java
```diff
@@ -5,11 +5,12 @@
 import lombok.Getter;
 import lombok.Setter;
 
+import java.io.Serializable;
 import java.util.Set;
 
 @Getter
 @Setter
-public class PageNameIdDTO {
+public class PageNameIdDTO implements Serializable {
```
mohanarpit, line 13: Is this change still required? Can we remove any extraneous changes from this PR that is not required?

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CacheableTemplateHelper.java
```diff
@@ -0,0 +1,118 @@
...
+                    CacheableApplicationJson cacheableApplicationJson = new CacheableApplicationJson();
+                    cacheableApplicationJson.setApplicationJson(jsonString);
+                    cacheableApplicationJson.setLastUpdated(Instant.now());
+
+                    // Remove the value from cache if its outdated
+                    if (cacheableApplicationJsonMap.containsKey(templateId)) {
+                        cacheableApplicationJsonMap.replace(templateId, cacheableApplicationJson);
+                    } else {
+                        cacheableApplicationJsonMap.put(templateId, cacheableApplicationJson);
+                    }
```
mohanarpit, line 108: Not sure if this is necessary. We can simply `put` into the hashmap.

AnaghHegde: Yes you are right. Just verified this, put will replace if the value already exists.

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CacheableTemplateHelper.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.Map;
+
+@AllArgsConstructor
+@Slf4j
+public class CacheableTemplateHelper {
+    // Template metadata is used for showing the preview of the template
+
+    public static CacheableApplicationTemplate applicationTemplateList = new CacheableApplicationTemplate();
+
+    public static Map<String, CacheableApplicationJson> cacheableApplicationJsonMap = new HashMap<>();
```
mohanarpit, line 31: Why are these variables public? Do we expect them to be read/updated by external classes?

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ApplicationTemplateServiceUnitTest.java
```diff
@@ -121,12 +84,17 @@ public void getActiveTemplates_WhenRecentlyUsedExists_RecentOnesComesFirst() thr
...
+        cacheableApplicationTemplate.setLastUpdated(Instant.now());
 
-        Mono<List<ApplicationTemplate>> templateListMono = applicationTemplateService.getActiveTemplates(null);
+        Mono<CacheableApplicationTemplate> templateListMono =
+                CacheableTemplateHelper.getTemplates("recently-used", cloudServicesConfig.getBaseUrl());
 
         StepVerifier.create(templateListMono)
-                .assertNext(applicationTemplates -> {
-                    assertThat(applicationTemplates).hasSize(3);
+                .assertNext(cacheableApplicationTemplate1 -> {
```
mohanarpit, line 95: I'm slightly confused about what this test is actually testing. We seem to be setting the templateList in `cacheableApplicationTemplate` and then we are retrieving that from the function and asserting the size is the same. 

Under what conditions will this test fail? Each unit test should actually fail on some condition. Only then can we be certain that the test is truly asserting something.

AnaghHegde: I lack context on why this test class was added. But from reading the code, they are mocking the webclient behaviour to test success and failure. The failure scenario is in the test class `ApplicationTemplateServicePublishTemplateTest`. My hunch is that mocking the webclient did not allow to send different status codes at the same time and hence this was added to a different test class.

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CacheableTemplateHelper.java
AnaghHegde: This was miss from side, fixed the change.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ApplicationTemplateServiceUnitTest.java
mohanarpit: My question wasn't related to the previous version of the test function, but more w.r.t the current changes. If I look at the test function right now, we are basically testing the getter and setter in Java. I don't think this is a useful test. 

Can you please list the test scenarios that you would want to automate & test? Then let's automate it.

AnaghHegde: The only case i can think of is for adding the tests for the the method which calculates the time difference. Apart from this, I can't think of any test cases for the current scenario.

mohanarpit: I can immediately think of a few tests: 

Test #1
1. Assert that `CacheableTemplateHelper` doesn't have the POJO or has an empty POJO.
2. Fetch the templates via the normal flow by mocking CS. 
3. Check if the `CacheableTemplateHelper.getApplicationTemplateList()` is the same as the object returned by the normal flow function. This will ensure that the cache is being set correctly. 

Test #2
1. Set the object in `CacheableTemplateHelper`. This means there is already an object present in cache. 
2. Fetch the templates via the normal flow by mocking CS. 
3. Assert that the result is same as the object set in cache.
4. Assert that the mock should not have been invoked because of cache hit. 

Test #3
1. Set the object in `CacheableTemplateHelper` with an older timestamp. This means there is already an old object present in cache. 
2. Fetch the templates via the normal flow by mocking CS. 
3. Assert that the mock should be invoked because of cache miss. 

Test #4
1. Set the object in `CacheableTemplateHelper` with corrupted object. What happens then? 

Test #5 
1. Set the object in `CacheableTemplateHelper` with null/empty object. What happens then?

Do you think any of these scenarios are relevant? If yes, what could we have done better to ensure you would have come up with these scenarios independently?

AnaghHegde: @mohanarpit Added test cases. Can you check again?

app/server/appsmith-server/src/main/java/com/appsmith/server/imports/internal/partial/PartialImportServiceCEImpl.java
```diff
@@ -414,14 +417,18 @@ public Mono<BuildingBlockResponseDTO> importBuildingBlock(BuildingBlockDTO build
         Mono<ApplicationJson> applicationJsonMono =
                 applicationTemplateService.getApplicationJsonFromTemplate(buildingBlockDTO.getTemplateId());
 
+        Stopwatch processStopwatch = new Stopwatch("Download Content from Cloud service");
```
mohanarpit, line 420: Are these timing changes required? #JustAsking

AnaghHegde: Yes we need this data from prod. I will remove this after 2-3 weeks of collecting this.

app/server/appsmith-server/src/test/java/com/appsmith/server/helpers/CacheableTemplateHelperTemplateJsonDataTest.java
```diff
@@ -0,0 +1,198 @@
...
+    /* Scenarios covered via this test:
+     * 1. CacheableTemplateHelper doesn't have the POJO or has an empty POJO.
+     * 2. Fetch the templates via the normal flow by mocking CS.
+     * 3. Check if the CacheableTemplateHelper.getApplicationTemplateList() is the same as the object returned by the normal flow function. This will ensure that the cache is being set correctly.
+     * 4. From the above steps we now have the cache set.
+     * 5. Fetch the templates again, verify the data is the same as the one fetched in step 2.
+     * 6. Verify the cache is used and not the mock. This is done by asserting the lastUpdated time of the cache.
+     */
+    @Test
+    @Order(1)
```
mohanarpit, line 97: @AnaghHegde Why do we need the `Order` annotation? Can we avoid ordering our tests in a specific order?

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/CacheableApplicationJson.java
```diff
@@ -0,0 +1,13 @@
...
+import lombok.Data;
+
+import java.time.Instant;
+
+@Data
+public class CacheableApplicationJson {
+
+    ApplicationJson applicationJson;
+
+    Instant timeStamp;
```
mohanarpit, line 12: Nit: Minor suggestion to rename the variable to be explicit. Same with the other template POJO.

```suggestion
    Instant cacheExpiryTime;
```

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationTemplateServiceCEImpl.java
mohanarpit: @AnaghHegde Can you please respond here and resolve this comment?

AnaghHegde: @mohanarpit The user will see the 404 error - `"Received error from cloud services`

app/server/appsmith-server/src/test/java/com/appsmith/server/helpers/CacheableTemplateHelperTemplateJsonDataTest.java
AnaghHegde: @mohanarpit Because of the static fields, the values of POJO were not predictable. To prevent this i added order of execution so that we know what is in the POJO when asserting.

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/CacheableTemplateHelper.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+@AllArgsConstructor
+@Slf4j
+@Component
+public class CacheableTemplateHelper {
+    // Template metadata is used for showing the preview of the template
+
+    private static CacheableApplicationTemplate applicationTemplateList = new CacheableApplicationTemplate();
```
mohanarpit, line 37: Can we remove this static variable? `CacheableTemplateHelper` is a Spring bean. This means it's a singleton class that's initialized only once by Spring at boot time. Ideally, we don't want non-final static variables in Spring beans. If you remove the static part, the test cases will also become simpler.

app/server/appsmith-server/src/test/java/com/appsmith/server/helpers/CacheableTemplateHelperTemplateJsonDataTest.java
```diff
@@ -0,0 +1,191 @@
...
+                })
+                .verifyComplete();
+        assertThat(cacheableTemplateHelper.getCacheableApplicationJsonMap().size())
+                .isEqualTo(1);
+    }
+
+    /* Scenarios covered via this test:
+     * 1. Set the cache time to 1 day before the current time.
+     * 2. Fetch the templates again, verify the data is from the mock and not from the cache.
+     */
```
mohanarpit, line 133: @AnaghHegde Is this comment accurate? Same for the other class as well.

app/server/appsmith-server/src/test/java/com/appsmith/server/helpers/CacheableTemplateHelperTemplateMetadataTest.java
```diff
@@ -0,0 +1,163 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+
+@ExtendWith(SpringExtension.class)
+@SpringBootTest
+public class CacheableTemplateHelperTemplateMetadataTest {
+
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+    private static MockWebServer mockCloudServices;
```
mohanarpit, line 37: Here as well. Is there a need for these variables to be static? Can they not be non-static variables?

AnaghHegde: @mohanarpit The MockWebServer needs to be started in the before all the tests and when you use the `BeforeAll` and `@AfterAll` they have to be static. And accessing non static from a static method throws error.

mohanarpit (approved):

mohanarpit:
@AnaghHegde Please resolve the failing tests here. Apart from that, the code LGTM.

mohanarpit (approved):

PR #33441, opened 2024-05-14
feat: Collapsible Bottom View

hetunandu:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/client/src/IDE/Components/BottomView.tsx
+++ b/app/client/src/IDE/Components/BottomView.tsx
@@ -1,0 +1,150 @@
+import React, {
+  type CSSProperties,
+  type RefObject,
+  useEffect,
+  useRef,
+  useState,
+} from "react";
+import styled from "styled-components";
+import Resizer, {
+  ResizerCSS,
+} from "components/editorComponents/Debugger/Resizer";
+import { ActionExecutionResizerHeight } from "pages/Editor/APIEditor/constants";
+import { CodeEditorWithGutterStyles } from "pages/Editor/JSEditor/constants";
+import { ViewHideBehaviour } from "../Interfaces/ViewHideBehaviour";
+import { Button } from "design-system";
+
+const VIEW_MIN_HEIGHT = "38px";
+
+const Container = styled.div`
+  ${ResizerCSS};
+  width: 100%;
+  // Minimum height of bottom tabs as it can be resized
+  min-height: ${VIEW_MIN_HEIGHT};
+  background-color: var(--ads-v2-color-bg);
+  height: ${ActionExecutionResizerHeight}px;
+  border-top: 1px solid var(--ads-v2-color-border);
+`;
+
+const ViewWrapper = styled.div`
+  height: 100%;
+  &&& {
+    ul.ads-v2-tabs__list {
+      margin: 0 ${(props) => props.theme.spaces[11]}px;
+      height: ${VIEW_MIN_HEIGHT};
+    }
+  }
+
+  & {
+    .ads-v2-tabs__list {
+      padding: var(--ads-v2-spaces-1) var(--ads-v2-spaces-7);
+    }
+  }
+
+  & {
+    .ads-v2-tabs__panel {
+      ${CodeEditorWithGutterStyles};
+      overflow-y: auto;
+      height: calc(100% - ${VIEW_MIN_HEIGHT});
+    }
+  }
+`;
+
+const MIN_HEIGHT = {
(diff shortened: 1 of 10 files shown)


app/client/src/IDE/Components/BottomView.tsx
```diff
@@ -0,0 +1,150 @@
...
+import styled from "styled-components";
+import Resizer, {
+  ResizerCSS,
+} from "components/editorComponents/Debugger/Resizer";
+import { ActionExecutionResizerHeight } from "pages/Editor/APIEditor/constants";
+import { CodeEditorWithGutterStyles } from "pages/Editor/JSEditor/constants";
+import { ViewHideBehaviour } from "../Interfaces/ViewHideBehaviour";
+import { Button } from "design-system";
+
+const VIEW_MIN_HEIGHT = "38px";
```
ashit-rath, line 17: Can we just have the number 38? Good practice as  there usually comes a time when you want to reuse this and put to to calculate a value for something else

```
const VIEW_MIN_HEIGHT = 38
```

app/client/src/IDE/Components/BottomView.tsx
```diff
@@ -0,0 +1,150 @@
...
+    right: 0;
+    padding: 9px 11px;
+  }
+`;
+
+const ViewHide = (props: {
+  hideBehaviour: ViewHideBehaviour;
+  isHidden: boolean;
+  onToggle: () => void;
+}) => {
```
ashit-rath, line 82: Can we put the typings in a separate interface than having inline?

app/client/src/IDE/Components/BottomView.tsx
```diff
@@ -0,0 +1,150 @@
...
+        setIcon("arrow-up-s-line");
+      } else {
+        setIcon("arrow-down-s-line");
+      }
+    }
+  }, [props.isHidden]);
+
+  return (
+    <ViewHideButton
+      className="view-hide-button t--view-hide-button"
```
ashit-rath, line 101: Any reason why we aren't using `data-testid` property to add `t--view-hide-button`?

app/client/src/IDE/Components/BottomView.tsx
```diff
@@ -0,0 +1,150 @@
...
+      kind="tertiary"
+      onClick={props.onToggle}
+      size="md"
+      startIcon={icon}
+    />
+  );
+};
+
+const BottomView = (props: Props) => {
+  const panelRef: RefObject<HTMLDivElement> = useRef(null);
```
ashit-rath, line 112: nit: 
```
const panelRef = useRef<HTMLDivElement>(null);
```

app/client/src/IDE/Components/BottomView.tsx
```diff
@@ -0,0 +1,150 @@
...
+  // Handle the height of the tabs when toggling the hidden state
+  useEffect(() => {
+    const panel = panelRef.current;
+    if (!panel) return;
+    if (props.hidden) {
+      panel.style.height = MIN_HEIGHT[props.behaviour];
+    } else {
+      panel.style.height = `${props.height}px`;
+    }
+  }, [props.hidden]);
```
ashit-rath, line 123: `props.behavior` is missing from the dependency list

app/client/src/IDE/Components/BottomView.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  return (
+    <Container
+      className={`select-text ${props.className}`}
+      ref={panelRef}
+      style={props.additionalContainerStyles}
```
ashit-rath, line 129: Rather than accepting inline styles, can we introduce variants / some logical prop that would lead to the desired styling that this component offers?
Example: passing a `bottomSticky: boolean` prop would apply the `position: absolute, bottom: 0` style

ashit-rath:
Few observations:
1. The JS Object bottom tab has some addition height when collapsed, which is not present for the queries section
Screenshot 2024-05-16 at 9 28 37 AM

2. Font weight of the tab's label is not consistent when collapsed and uncollapsed.

app/client/src/IDE/Components/BottomView.tsx
```diff
@@ -0,0 +1,150 @@
+import React, {
```
ashit-rath, line 1: Also can we add some jest test for this component?

ashit-rath (approved):

PR #33477, opened 2024-05-15
chore: Add Swagger UI for server API docs

nidhi-nair:
## Description
Adds Swagger based UI for API docs

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
@@ -196,6 +196,7 @@ public class SecurityConfig {
                                 ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, USER_URL + "/invite/verify"),
                                 ServerWebExchangeMatchers.pathMatchers(HttpMethod.PUT, USER_URL + "/invite/confirm"),
                                 ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, USER_URL + "/me"),
+                                ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, "/v3/**"),
                                 ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, USER_URL + "/features"),
                                 ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, ASSET_URL + "/*"),
                                 ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, ACTION_URL + "/**"),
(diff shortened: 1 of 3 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
```diff
@@ -196,6 +196,7 @@ public SecurityWebFilterChain securityWebFilterChain(ServerHttpSecurity http) {
                                 ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, USER_URL + "/invite/verify"),
                                 ServerWebExchangeMatchers.pathMatchers(HttpMethod.PUT, USER_URL + "/invite/confirm"),
                                 ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, USER_URL + "/me"),
+                                ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, "/v3/**"),
```
nidhi-nair, line 199: This one's so that the redirect to swagger's ui is allowed

app/server/appsmith-server/src/main/resources/application.properties
```diff
@@ -127,3 +127,6 @@ appsmith.newrelic.micrometer.metrics.application.name=${APPSMITH_NEWRELIC_MICROM
 spring.application.name=${OTEL_SERVICE_NAME:appsmith-anonymous}
 appsmith.micrometer.metrics.enabled=${APPSMITH_MICROMETER_METRICS_ENABLED:false}
 appsmith.micrometer.tracing.detail.enabled=${APPSMITH_ENABLE_TRACING_DETAIL:false}
+
+springdoc.api-docs.path=/v3/docs
```
nidhi-nair, line 131: Just reusing the same subpath for ease

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
sharat87: Instead of the wildcard, can we get the exact path configured and use that please? Something like this:

shot-2024-05-15-10-06-16

sharat87 (approved):
Minor thing.

nidhi-nair:
There's actually a bunch of other endpoints swagger needs to load the ui, under the same sub-path is why I chose to just use the parent.

You think I should include all of the following separately? @sharat87 
```
/v3/swagger
/v3/api-docs
/v3/webjars/swagger-ui/index.html
/v3/docs/swagger-config
/v3/docs
```

PR #33480, opened 2024-05-15
CI: Add Repeat for ci-testlimited

yatinappsmith:
Add Repeat for ci-testlimited

### DIFF
--- a/.github/workflows/ci-test-limited.yml
+++ b/.github/workflows/ci-test-limited.yml
@@ -365,7 +365,7 @@ jobs:
           spec: ${{ env.specs_to_run }}
           working-directory: app/client
           env: "NODE_ENV=development"
-          command: npx cypress-repeat run -n ${{ inputs.repeat }} --until-passes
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-test-limited.yml
```diff
@@ -359,6 +365,7 @@ jobs:
           spec: ${{ env.specs_to_run }}
           working-directory: app/client
           env: "NODE_ENV=development"
+          command: npx cypress-repeat run -n ${{ inputs.repeat }} --until-passes --browser ${{ env.BROWSER_PATH }} --config-file cypress_ci_custom.config.ts --spec ${{ env.specs_to_run }} --env "NODE_ENV=development"
```
sharat87, line 368: Looks like a lot of the configuration under `with` is duplicated in CLI arguments here. Should we remove them under `with`, if they are no longer effective, since we're using a custom command?

sharat87 (approved):
Suggestion for a minor cleanup, otherwise looks okay.

PR #33528, opened 2024-05-16
fix: apiKey security issue

### LINKED ISSUE #30009: [Bug]: API Key showing in console log

### Description

I have a datasource that uses an api key for authentication.  When I run a query of that data source the API key is visible in the log.

Internal conversation :- https://theappsmith.slack.com/archives/C040LHZN03V/p1704354949494669 

### Steps To Reproduce

image

1. Create data source 'authenticated api'
2. Set authentication to apikey
3. Give it an apikey
4. Create a query for that data source
5. Run the query
6. Go to your logs
7. Expand the log for the query

Note that if you edit that data source it hides the API key from you.  So if you are hiding it from the user on the edit screen you should not make it visible in the log.

### Public Sample App

_No response_

### Environment

Production

### Issue video log

_No response_

### Version

Self-hosted -  v1.9.44

rishabhrathod01:
## Description
Fixes #30009

### Summary:
This PR addresses the issue of masking sensitive information in query parameters or headers based on the authentication type selected by the user. The changes ensure that sensitive data is properly masked before sending back as response. 

### Changes:
- RequestCaptureFilter.java
Added logic to check the authentication type and mask the appropriate query parameters or headers.

### Testing:
Verified that the masking functionality works as expected for API_KEY authentication types.

## Automation

/test datasource

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9203547387>
> Commit: 9a7fc9cc1942ddc61c54f6ae9451706ad527f49c
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9203547387&attempt=2" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [x] No


NilanshBansal (approved):
Tested the DP, test cases and the code. LGTM

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/helpers/restApiUtils/helpers/RequestCaptureFilter.java
```diff
@@ -50,8 +56,44 @@ public RequestCaptureFilter(ObjectMapper objectMapper) {
         return next.exchange(request);
     }
 
+    private URI createURIWithMaskedQueryParam(URI uriToMask, String queryParamKeyToMask) {
+
+        String query = uriToMask.getQuery();
```
mohanarpit, line 61: @NilanshBansal @rishabhrathod01  We should always avoid doing custom parsing for URL parameters. Please replace this with a single 1 line function `UriComponentsBuilder`. Check this [SO answer](https://stackoverflow.com/a/39348897). 

Custom parsing like this can lead to security flaws, encoding problems, loopholes that we missed, etc. Best to use an established library.

NilanshBansal: Thank You @mohanarpit for highlighting this 🙏 I did raise this in one of my discussions with @rishabhrathod01 on call, but did not press on it much as I didn't foresee the issues that could come up. This is a miss from my end. 
Completely agree with you and the concerns raised by Shri on the failing edge cases.

rishabhrathod01: Thanks @mohanarpit. I was looking for something similar before adding custom logic, I should have done more research on this.

rishabhrathod01: I replaced the custom parsing code with the `UriComponentsBuilder` code now, we don't go into an error state but the output value is still incorrect when the secret value was provided as `value&othervalue=abc`. This is because the value stored in the query param is not encoded which we get using `request.url()`. 

@NilanshBansal I will need your assistance here to resolve this sooner, as the scope has increased a bit here.

PR #33529, opened 2024-05-16
chore: Remove incorrect and unused `update` signature

sharat87:
The `update` method with `String, T, String` signature, is not used in it's form anywhere. This is most likely owing to the implementation being off. The criteria used to find the object to be updated, is not the same as the criteria used to find the updated object to return to the caller.

The query to update is `{value of key field} = id`.

The query to find is `{value of "id" field} = id`.

This discrepency is likely why this function is not used anywhere. The `ApplicationServiceCE` alone has an overridden implementation for this signature, but that's defined with different variable names, and treats the third argument as the Git branch name, not `key`. That function is indeed used, and is unaffected by this change.

All of above is verified on EE as well. No conflicts, no build failure.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserServiceCEImpl.java
@@ -608,7 +608,7 @@ public class UserServiceCEImpl extends BaseService<UserRepository, User, String>
             updates.setName(inputName);
             updatedUserMono = sessionUserService
                     .getCurrentUser()
-                    .flatMap(user -> update(user.getId(), updates)
+                    .flatMap(user -> updateWithoutPermission(user.getId(), updates)
                             .then(
                                     exchange == null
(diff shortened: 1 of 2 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserServiceCEImpl.java
```diff
@@ -608,7 +608,7 @@ public Mono<User> updateCurrentUser(final UserUpdateDTO allUpdates, ServerWebExc
             updates.setName(inputName);
             updatedUserMono = sessionUserService
                     .getCurrentUser()
-                    .flatMap(user -> update(user.getId(), updates, User.Fields.id)
+                    .flatMap(user -> updateWithoutPermission(user.getId(), updates)
```
sharat87, line 611: `UserServiceCE` is the only service that overrides the `update` method, to _do_ a permission check, and has its own `updateWithoutPermission` separately. I intend to fix that inconsistency as well, in a separate PR.

nidhi-nair (approved):

PR #33535, opened 2024-05-17
fix: Show custom widget icon reliably when no widget is found

raza001:
Steps To Reproduce
go to edit mode of the application
In the entity explorer section click on UI tab
search for the widget name which is not present
The custom widget icon will be displayed but seems like broken as in attached screenshot

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  


jsartisan (approved):
LGTM!

just curious @albinAppsmith when did this break? 🤔

albinAppsmith:
@jsartisan I guess this was broken after building blocks.

PR #33538, opened 2024-05-17
feat: Add cleanup stage in transform script

abhvsn:


vivonk (approved):

app/server/tx/transform.py
```diff
@@ -355,10 +357,15 @@ def convert(domain):
     generate_cake_class(domain)
     #use_cake(domain)  # Commenting this out since we want both cake and repo to co-exist now.
 
+def cleanup():
+    print("Cleaning up old cake classes")
+    for filename in glob.iglob('./**/*RepositoryCake.java', recursive=True):
```
sharat87, line 362: I'm guessing the `.` here will be the folder where the script was run from, no? Can we instead use `server_root`? So if the script was run from the repository root, it doesn't end up searching the whole repo, things like `node_modules` also, which won't be useful. So something like:

```suggestion
    for filename in glob.iglob('./**/*RepositoryCake.java', root_dir=server_root, recursive=True):
```

PR #33543, opened 2024-05-17
feat: Make Postgres DB URL configurable via envs

abhvsn:


app/server/envs/test.env.example
```diff
@@ -2,7 +2,9 @@
 
 ACTIVE_PROFILE=test
 
-APPSMITH_MONGODB_URI="mongodb://localhost:27017/appsmith"
```
mohanarpit, line 5: Minor: Do we want to keep the `APPSMITH_MONGODB_URI` in the example env file? This way, it's explicit that this variable still exists for users who may want to run Appsmith with MongoDB.

abhvsn: @mohanarpit Yeah makes sense right now. Once we start to roll out Appsmith with Postgres we can move away from this env to avoid future confusion.

mohanarpit (approved):

abhvsn:
Closing, as changes will be addressed once we merge https://github.com/appsmithorg/appsmith/pull/33429

PR #33567, opened 2024-05-20
fix: Use field visibility and an annotation instrospector to get transient behavior

sharat87:


abhvsn (approved):

abhvsn:
@sharat87 let's create a task to remove `@Transient` from `org.springframework.data.annotation`.

sharat87:
Created https://github.com/appsmithorg/appsmith/issues/33572.

PR #33579, opened 2024-05-20
chore: missing field conflicts

vivonk:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ActionConfiguration.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ActionConfiguration.java
@@ -10,6 +10,8 @@ import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.google.gson.annotations.JsonAdapter;
 import jakarta.persistence.Column;
+import jakarta.persistence.EnumType;
+import jakarta.persistence.Enumerated;
 import jakarta.persistence.Transient;
 import lombok.Getter;
 import lombok.NoArgsConstructor;
@@ -54,6 +56,7 @@ public class ActionConfiguration implements AppsmithDomain, ExecutableConfigurat
     Integer timeoutInMillisecond;
 
     @JsonView({Views.Public.class, FromRequest.class, Git.class})
+    @Enumerated(EnumType.STRING)
     PaginationType paginationType = PaginationType.NONE;
 
     // API fields
@@ -99,6 +102,7 @@ public class ActionConfiguration implements AppsmithDomain, ExecutableConfigurat
     HttpMethod httpMethod;
 
     @JsonView({Views.Public.class, FromRequest.class, Git.class})
+    @Enumerated(EnumType.STRING)
     HttpProtocol httpVersion;
     // Paginated API fields
     @JsonView({Views.Public.class, FromRequest.class, Git.class})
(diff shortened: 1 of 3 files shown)


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ActionConfiguration.java
```diff
@@ -54,6 +56,7 @@ public class ActionConfiguration implements AppsmithDomain, ExecutableConfigurat
     Integer timeoutInMillisecond;
 
     @JsonView({Views.Public.class, FromRequest.class, Git.class})
+    @Enumerated(EnumType.STRING)
```
sharat87, line 59: This _may_ not be needed here. Since `ActionConfiguration` itself is not an `@Entity`, objects of this class will be saved as JSON, as the field of another `@Entity` class. And the JSON serialization is done by Jackson, not Hibernate. Jackson already uses `String` for enum fields.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ActionConfiguration.java
```diff
@@ -99,6 +102,7 @@ public class ActionConfiguration implements AppsmithDomain, ExecutableConfigurat
     HttpMethod httpMethod;
 
     @JsonView({Views.Public.class, FromRequest.class, Git.class})
+    @Enumerated(EnumType.STRING)
```
sharat87, line 105: Same as above. I don't think we need this here.

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Application.java
```diff
@@ -101,9 +101,11 @@ public class Application extends BaseDomain implements Artifact {
     private String slug;
 
     @JsonView({Views.Internal.class, Git.class})
+    @Column(columnDefinition = "jsonb")
```
sharat87, line 104: Unfortunately, we have to add _both_ the annotations togather for it to work. I couldn't figure out a way to have just one annotation do the job of both of these, but it's unlikely we ever want to use just one of these anywhere.

```suggestion
    @Type(CustomJsonType.class)
    @Column(columnDefinition = "jsonb")
```

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Application.java
```diff
@@ -101,9 +101,11 @@ public class Application extends BaseDomain implements Artifact {
...
     AppLayout unpublishedAppLayout;
 
     @JsonView(Views.Internal.class)
+    @Column(columnDefinition = "jsonb")
```
sharat87, line 108: Same as above.

```suggestion
    @Type(CustomJsonType.class)
    @Column(columnDefinition = "jsonb")
```

sharat87 (changes requested):

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ActionConfiguration.java
vivonk: my bad, I was treating this as entity. Have the same in create tables too. Removing it from tables and undo these changes

abhvsn (approved):

Issue #28401, opened 2023-10-26
[Task]: Create Anvil specific Base Widget


marks0351:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### SubTasks

Present BaseWidget is the boilerplate/abstract class on which every widget is built upon.
- Provides static functions for configurations that are defined by each widget 
- Provides Utilities like Updating Widget properties, Updating Meta Properties, Execute actions,etc
- sometimes also has widget specific implementation like Meta widget based utilities only used by List widget

Anvil is the new Editor of Appsmith which will have its own widget suite which mean Anvil widgets are not targeted to work on every layout system(Fixed or Auto Layout).

Having a common BaseWidget to build all widgets is becoming a maintainability issue and also carries the risk of regression in unintended layout systems.
So the proposal is to extract generic utilities out of the present day Base Widget and make sure Anvil has a separate Base Widget that has all the essentials needed but only has Anvil specific implementation.

KelvinOm:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Widget-Design-System-6434e3402be2c1002ea740be/issues/appsmithorg/appsmith/28401?planning-poker) with Zenhub @dhruvikn @jsartisan @KelvinOm

somangshu:
fyi @marks0351

somangshu:
@marks0351 lets prioritise this as a tech debt to be cleaned before moving to R2.

PR #33580, opened 2024-05-20
feat: enable DnD for entire modal widget(including header and footer).

### LINKED ISSUE #33318: [Task]: Trigger Modal layout highlights when moving on the header and footer as well

marks0351:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=DEipBn-)
## Description
In this PR, we are adding an enhancement to DnD of modal widgets.
DnD was implemented only for layouts so in the case of modal, the header and footer are not layouts and DnD is not triggered when a widget is being dragged on top of these components of a modal.

However for the user to be able to efficiently use the DnD of modal, mouse events of the modal are also needed to be processed.

To enable this we have a new hook that wraps all detached widgets in editor mode only, `useAnvilDetachedWidgetsDnD`
It makes sure mouse move events on the widget are dispatched to the dnd listeners via custom event `DETACHED_WIDGET_MOUSE_MOVE_EVENT`

we aslo had to special handle mouse leave and enter utilities to activate a canvas for DnD,
- for all widgets except modals these are handled on the `useAnvilDnDEventCallbacks`
- for modal widgets these events are handled on the `useAnvilDetachedWidgetsDnD`

Fixes #33318
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags="@tag.Anvil"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9171877023>
> Commit: 53925782d15c9998d993f39a0239bee64a7805fc
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9171877023&attempt=1" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/client/src/layoutSystems/anvil/editor/AnvilEditorDetachedWidgetOnion.tsx
+++ b/app/client/src/layoutSystems/anvil/editor/AnvilEditorDetachedWidgetOnion.tsx
@@ -7,6 +7,7 @@ import {
 } from "layoutSystems/anvil/common/hooks/detachedWidgetHooks";
 import { AnvilErrorBoundary } from "../common/widgetComponent/AnvilErrorBoundary";
 import { SKELETON_WIDGET_TYPE } from "constants/WidgetConstants";
+import { useAnvilDetachedWidgetsDnD } from "./hooks/useAnvilDetachedWidgetsDnD";
 
 /**
  * AnvilEditorDetachedWidgetOnion
@@ -23,7 +24,11 @@ export const AnvilEditorDetachedWidgetOnion = (props: BaseWidgetProps) => {
   useObserveDetachedWidget(props.widgetId);
   useHandleDetachedWidgetSelect(props.widgetId);
   useAddBordersToDetachedWidgets(props.widgetId, props.type);
-
+  useAnvilDetachedWidgetsDnD(
+    props.widgetId,
+    props.layout[0].layoutId,
+    !!props.isVisible,
+  );
   return props.type !== SKELETON_WIDGET_TYPE ? (
     <AnvilErrorBoundary {...props}>{props.children}</AnvilErrorBoundary>
   ) : null;
--- a/app/client/src/layoutSystems/anvil/editor/canvasArenas/AnvilModalDropArena.tsx
+++ b/app/client/src/layoutSystems/anvil/editor/canvasArenas/AnvilModalDropArena.tsx
@@ -12,7 +12,6 @@ const StyledModalEditorDropArenaWrapper = styled.div<{ isModalEmpty: boolean }>`
   ${(props) =>
     props.isModalEmpty &&
     `
-    height: 100% !important;
   padding: ${EMPTY_MODAL_PADDING}px;
   `}
 `;
@@ -64,7 +63,10 @@ export const AnvilModalDropArena = ({
   const widget = useSelector(getWidgetByID(modalId));
   const isModalEmpty = widget.children?.length === 0;
   return (
-    <StyledModalEditorDropArenaWrapper isModalEmpty={isModalEmpty}>
+    <StyledModalEditorDropArenaWrapper
+      isModalEmpty={isModalEmpty}
+      style={{ height: "100%" }}
+    >
       <StyledEmptyModalDropArena
         isActive={isCurrentDraggedCanvas}
         isModalEmpty={isModalEmpty}
--- a/app/client/src/layoutSystems/anvil/editor/canvasArenas/AnvilModalDropArena.tsx
+++ b/app/client/src/layoutSystems/anvil/editor/canvasArenas/AnvilModalDropArena.tsx
@@ -9,10 +9,10 @@ export const EMPTY_MODAL_PADDING = 4;
 
 const StyledModalEditorDropArenaWrapper = styled.div<{ isModalEmpty: boolean }>`
   position: relative;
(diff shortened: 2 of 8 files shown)


app/client/src/layoutSystems/anvil/editor/AnvilEditorDetachedWidgetOnion.tsx
```diff
@@ -23,7 +24,11 @@ export const AnvilEditorDetachedWidgetOnion = (props: BaseWidgetProps) => {
   useObserveDetachedWidget(props.widgetId);
   useHandleDetachedWidgetSelect(props.widgetId);
   useAddBordersToDetachedWidgets(props.widgetId, props.type);
-
+  useAnvilDetachedWidgetsDnD(
+    props.widgetId,
+    props.layout[0].layoutId,
+    !!props.isVisible,
```
KelvinOm, line 30: Shouldn't it be just `props.isVisible` without `!!`? Why do we need a double inversion here if it should be the  Boolean type anyway?

```suggestion
    props.isVisible,
```

app/client/src/layoutSystems/anvil/editor/canvasArenas/AnvilModalDropArena.tsx
```diff
@@ -9,10 +9,10 @@ export const EMPTY_MODAL_PADDING = 4;
 
 const StyledModalEditorDropArenaWrapper = styled.div<{ isModalEmpty: boolean }>`
   position: relative;
-  height: 100% !important;
   ${(props) =>
     props.isModalEmpty &&
     `
+    height: 100% !important;
```
KelvinOm, line 15: Why do we need this change? Can we get rid of the `!important` here?

KelvinOm (dismissed):

app/client/src/layoutSystems/anvil/editor/AnvilEditorDetachedWidgetOnion.tsx
marks0351: the type for isVisible is boolean|undefined since its not available for all widgets, so we had to use the `!!`

KelvinOm: Why is it unavailable for some widgets? Can we pass isVisible: false explicitly?

marks0351: The issue here is that we are using the same type definitions for fixed and anvil. so by definition this prop can be undefined, why coz there are canavs widgets in fixed that dont have a isVisible property not sure if there are any more widgets.

as far as setting false explicitly would be a clean up task across the fixed layout paradigm.
instead of doing this, the best thing would be to split withWidgetProps and also have separated type definitions for Anvil. we do have a tech debt task listed for this.

KelvinOm: Understood. There should be a link to BaseWidget task 🙂

marks0351: #28401

app/client/src/layoutSystems/anvil/editor/canvasArenas/AnvilModalDropArena.tsx
KelvinOm: If it's a minor effort, then it's better to get rid of `important`.

app/client/src/layoutSystems/anvil/editor/canvasArenas/hooks/useAnvilDnDEventCallbacks.ts
```diff
@@ -208,10 +211,26 @@ export const useAnvilDnDEventCallbacks = ({
   );
 
   const onMouseOut = useCallback(() => {
-    setDraggingCanvas("");
-  }, [setDraggingCanvas]);
+    if (currentWidgetHierarchy !== widgetHierarchy.WDS_MODAL_WIDGET) {
```
riodeuno, line 214: This seems like an abstraction leak. If the drawer widget or some other widget of the kind needs to be implemented, we'll now have to add another clause for `widgetHierarchy.WDS_DRAWER_WIDGET`. This can get out of hand.

It is fine for now, but I will recommend we create a task and think about how we're going to avoid this abstraction leak. 

The abstraction here is that the system registers widgets while being agnostic of what those widgets are. To prevent abstractions we can extend the widget development API to have widget provide their custom implementations for things like these. Check out `canvasHeightOffset` in the Tabs Widget in the fixed mode. The other ways to avoid such abstraction leaks is to group these widgets into a category, like we've done with `detachFromLayout` property.

riodeuno (approved):

app/client/src/layoutSystems/anvil/editor/canvasArenas/hooks/useAnvilDnDEventCallbacks.ts
marks0351: #33613

KelvinOm (approved):

PR #29151, opened 2023-11-27
fix: reset widget to default  value after setter method 

ohansFavour:
## Description
This PR ensures that widgets are reset to their default value after the `setValue` setter method is used to set its value

#### PR fixes following issue(s)
Fixes #27119 

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- Chore (housekeeping or task changes that don't impact user perception)
- This change requires a documentation update
>
>
>
- [ ] Manual
- [ ] JUnit
- [ ] Jest
- [ ] Cypress
>
>
#### Dev activity
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] PR is being merged under a feature flag

#### QA activity:
- [ ] [Speedbreak features](https://github.com/appsmithorg/TestSmith/wiki/Guidelines-for-test-plans#speedbreakers-) have been covered
- [ ] Test plan covers all impacted features and [areas of interest](https://github.com/appsmithorg/TestSmith/wiki/Guidelines-for-test-plans#areas-of-interest-)
- [ ] Test plan has been peer reviewed by project stakeholders and other QA members
- [ ] Manually tested functionality on DP
- [ ] We had an implementation alignment call with stakeholders post QA Round 2
- [ ] Cypress test cases have been added and approved by SDET/manual QA
- [ ] Added `Test Plan Approved` label after Cypress tests were reviewed
- [ ] Added `Test Plan Approved` label after JUnit tests were reviewed

## Summary by CodeRabbit

- **New Features**
  - Implemented a new reset functionality for widgets, allowing users to revert to default values after changes.
  - Enhanced widget meta updates with a new reset action.

- **Bug Fixes**
  - Added test cases to ensure widget reset functionality works as expected, even after asynchronous operations.

- **Refactor**
  - Refactored evaluation logic to improve handling of widget meta updates and resets.
  - Improved action execution logic for resetting widget properties.

- **Tests**
  - Expanded end-to-end regression tests to cover new reset widget functionality.

- **Documentation**
  - Updated internal documentation to reflect new action types and evaluation processes related to widget meta updates.

PR #33600, opened 2024-05-20
fix: function undefined error after resetWidget is called due to sync…

rishabhrathod01:
## Description

Fixed the error of async function like `Api.run` being `undefined` after running `resetWidget`. 

#### Context:-
we have 2 types of evaluation i.e., sync and async evaluation.

These are dependent on the field we are evaluating, for the data field we use sync evaluation and for the trigger field, we use async evaluation. In Async evaluation, we also define entity functions like `Api.run`.
- In resetWidget code, internally we were switching to sync evaluation and we forgot to add code to switch back to async and due to this code that was after it was throwing an `undefined` error. 

Fixes https://github.com/appsmithorg/appsmith/issues/33601

### Steps to reproduce 

1. Import user’s application ( shared on slack )
2. Go to JSObj_Main JSObject
3. Check the onClickCopySheetTemplate function 
4. Comment the  await Qry_C_CopySheetTemplate.run(); line and run the function.
5. On release it will throw error for Qry_R_SheetTemplate.run(); not being defined but will work fine on DP. 

## Automation

/ok-to-test tags="@tag.JS"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9170093222>
> Commit: f9e8893d12cedbd9d2368cd9609bda7b86816588
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9170093222&attempt=1" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [x] No

### DIFF
--- a/app/client/src/ce/workers/Evaluation/Actions.ts
+++ b/app/client/src/ce/workers/Evaluation/Actions.ts
@@ -98,15 +98,21 @@ export const addDataTreeToContext = (args: {
       EVAL_CONTEXT,
     );
 
-  // if eval is not trigger based i.e., sync eval then we skip adding entity and platform function to evalContext
   if (!isTriggerBased) return;
+  // if eval is not trigger based i.e., sync eval then we skip adding entity function to evalContext
+  addEntityFunctionsToEvalContext(EVAL_CONTEXT, entityFunctionCollection);
+};
 
+export const addEntityFunctionsToEvalContext = (
+  evalContext: EvalContext,
+  entityFunctionCollection: Record<string, Record<string, Function>>,
+) => {
   for (const [entityName, funcObj] of Object.entries(
     entityFunctionCollection,
   )) {
-    EVAL_CONTEXT[entityName] = Object.assign(
+    evalContext[entityName] = Object.assign(
       {},
-      EVAL_CONTEXT[entityName],
+      evalContext[entityName],
       funcObj,
     );
   }
--- a/app/client/src/workers/Evaluation/fns/__tests__/index.test.ts
+++ b/app/client/src/workers/Evaluation/fns/__tests__/index.test.ts
@@ -3,6 +3,7 @@ import { addPlatformFunctionsToEvalContext } from "@appsmith/workers/Evaluation/
 import { setEvalContext } from "workers/Evaluation/evaluate";
 import type { DataTree } from "entities/DataTree/dataTreeTypes";
 import { ENTITY_TYPE } from "entities/DataTree/dataTreeFactory";
+import { RenderModes } from "constants/WidgetConstants";
 
 const dataTree: DataTree = {
   action1: {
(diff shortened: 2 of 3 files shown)


ApekshaBhosale:
@rishabhrathod01 in the issue created you have written "resetWidget function causes the next async method to be undefined". so we need to switch it back to async for next method. why can't we do this before calling next async rather than depending on previous one to be in the right context?

rishabhrathod01:
> in the issue created you have written "resetWidget function causes the next async method to be undefined". so we need to switch it back to async for next method. why can't we do this before calling next async rather than depending on previous one to be in the right context? 

@ApekshaBhosale 

1. As the next async method will be undefined in the eval context, we won't be able to trigger the switch of sync to async evaluation as we won't know when the code got executed. 

For example:- 
```
resetWidget(WidgetName);
Api.run();
```
Here once the resetWidget is executed, we won't be able to switch back to async eval as `Api.run()` itself is not defined to let us know if this was called. Also, we want to default evaluation context to be always async this is because any async operation could be picked from the task queue.

2. By design itself, an async method is not supposed to change the evaluation from async to sync, in the case of `resetWidget` it was a bug that internally it was doing a sync eval but we didn’t switch back.

ApekshaBhosale:
> > in the issue created you have written "resetWidget function causes the next async method to be undefined". so we need to switch it back to async for next method. why can't we do this before calling next async rather than depending on previous one to be in the right context?
> 
> @ApekshaBhosale
> 
> 1. As the next async method will be undefined in the eval context, we won't be able to trigger the switch of sync to async evaluation as we won't know when the code got executed.
> 
> For example:-
> 
> ```
> resetWidget(WidgetName);
> Api.run();
> ```
> 
> Here once the resetWidget is executed, we won't be able to switch back to async eval as `Api.run()` itself is not defined to let us know if this was called. Also, we want to default evaluation context to be always async this is because any async operation could be picked from the task queue.
> 
> 2. By design itself, an async method is not supposed to change the evaluation from async to sync, in the case of `resetWidget` it was a bug that internally it was doing a sync eval but we didn’t switch back.

Understood! then reset Widget should be in async always isn't it? rather than setting it back. AFAIK we were treating all Appsmith platform functions as async and executing them in async context. @rishabhrathod01

rishabhrathod01:
> Understood! then reset Widget should be in async always isn't it? rather than setting it back. AFAIK we were treating all Appsmith platform functions as async and executing them in async context. @rishabhrathod0

@ApekshaBhosale `resetWidget` is an async API as shown in below code screenshot. 
image

We added an sync evaluation inside the resetWidget to fix bug explained in this comment https://github.com/appsmithorg/appsmith/pull/29151/files#r1607899227

ApekshaBhosale (approved):

PR #33603, opened 2024-05-20
ci: Remove inert ADS compliance check

sharat87:
The ADS compliance check relies on the condition `github.pull_request.base.ref == 'true'`, but this variable just doesn't exist. There's no `pull_request` under the `github` context, so this ADS compliance check never worked.

If we want this check, we can add it later as part of the `client-build.compliance.js` script. Here's a PoC for the record, based on the intention I interpreted:

```javascript
function doADSCheck({core, github, context, affectedFiles}) {
  const filesForADSCheck = affectedFiles.filter(f => (f.status === "added" || f.status === "modified") && f.filename.startsWith("app/client/src/"));

  const violatedFiles = new Set();
  for (const f of filesForADSCheck) {
    const content = fs.readFileSync(f.filename, "utf8");
    if (content.match(/(color|Color).*#|border.*#|(color|Color).*"/)) {
      violatedFiles.add(f.filename);
    }
  }

  if (violatedFiles.size === 0) {
    return;
  }

  const body = [
    "🔴 Below files are not compliant with ADS. Please fix and re-trigger ok-to-test",
    ...Array.from(violatedFiles).sort().map(f => "1. " + f),
  ].join("\n");

  github.rest.issues.createComment({
    owner: context.repo.owner,
    repo: context.repo.repo,
    issue_number: prNumber,
    body,
  });
}
```

/test sanity


mohanarpit (approved):

mohanarpit:
@sharat87 I've approved these changes. But wasn't this code working at some point? I'm sure I've seen failure messages in the past when we did violate ADS compliance. Or maybe I dreamt it 😛

sharat87:
@mohanarpit, I'm actually surprised myself cause I also recall seeing failure messages from this. But it looks like the original commit that added this check also had `github.pull_request`. That leads me to wonder if _GitHub_ themselves have changed the API here? 🤔 🤷

PR #33622, opened 2024-05-21
chore: creating super user bean through bean

vivonk:
## Description
This PR replaces repeating updateSuperUser migration that was existing on MongoDB, and instead create bean for the same because of dependency injection used in migration previously.

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/UserConfig.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/UserConfig.java
@@ -1,0 +1,129 @@
+package com.appsmith.server.configurations;
+
+import com.appsmith.server.acl.PolicyGenerator;
+import com.appsmith.server.constants.FieldName;
+import com.appsmith.server.domains.Config;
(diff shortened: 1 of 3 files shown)


abhvsn (commented):
LGTM. 
QQ: Does it make sense to get this change from release branch to decrease the diff and keeping similar way of handling repetitive migrations for both PG and Mongo?

vivonk:
Can we take this up later? But I agree, a better approach is bringing this change from the release branch itself.

abhvsn:
Let's create a ticket to implement this change in release, which will help us to communicate with larger audience to not introduce these repetitive migrations. Approving the change!

abhvsn (approved):

vivonk:
Tracking followup ticket here - https://github.com/appsmithorg/appsmith/issues/33624

PR #33636, opened 2024-05-22
ci: Making minor improvements to the CI workflow files

mohanarpit:
This is trivial housekeeping task to keep the workflow files consistent and remove warnings.

### DIFF
--- a/.github/workflows/ci-test-limited.yml
+++ b/.github/workflows/ci-test-limited.yml
@@ -96,7 +96,7 @@ jobs:
       # In case this is second attempt try restoring status of the prior attempt from cache
       - name: Restore the previous run result
         id: cache-appsmith
-        uses: martijnhols/actions-cache@v3.0.2
+        uses: actions/cache/restore@v4
         with:
           path: |
             ~/run_result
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-test-limited.yml
```diff
@@ -96,7 +96,7 @@ jobs:
       # In case this is second attempt try restoring status of the prior attempt from cache
       - name: Restore the previous run result
         id: cache-appsmith
-        uses: martijnhols/actions-cache@v3.0.2
+        uses: actions/cache/restore@v4
```
sharat87, line 99: There's a lot of other places where we have `actions/cache@v4` for the `run_result` file. I'm guessing we'll fix those in a subsequent PR?

sharat87 (approved):

.github/workflows/ci-test-limited.yml
mohanarpit: Yes. The goal of this & it's [companion PR](https://github.com/appsmithorg/appsmith-ee/pull/4238/) is to remove the warnings that we see in the Github action runs. We'll make `actions/cache` to `actions/cache/restore` at a later time.

PR #33649, opened 2024-05-22
feat: Anvil: Enable parent selection for focused widget

### LINKED ISSUE #33648: [Task]: Enable parent widget selction for a focused widget
### SubTasks

Today, only a selected widget has the possibility of selecting its parent via the widget name component's parent toggle. This was added as there was no way to interact with the widget name component while a widget is focused, the only way to interact was to select the widget.

This has changed in #33646 .
We can now interact with a focused widget's widget name component.

- [ ] Enable the parent widget toggle, for a focused widget, in the widget name component.
- [ ] Disable parent widget toggle if the parent is already selected.

riodeuno:
## Description
- If the parent is already selected, disable parent selection
- Introduce the selector to check if parent is selected.

Fixes #33648 

## Automation

/ok-to-test tags="@tag.Anvil"

> [!CAUTION]  

## Communication
- [ ] Yes
- [x] No

### DIFF
--- a/app/client/src/layoutSystems/anvil/editor/AnvilWidgetName/index.tsx
+++ b/app/client/src/layoutSystems/anvil/editor/AnvilWidgetName/index.tsx
@@ -19,6 +19,7 @@ import type { NameComponentStates } from "./types";
 import { generateDragStateForAnvilLayout } from "layoutSystems/anvil/utils/widgetUtils";
 import { SelectionRequestType } from "sagas/WidgetSelectUtils";
 import { useWidgetSelection } from "utils/hooks/useWidgetSelection";
+import { isWidgetSelected } from "selectors/widgetSelectors";
 
 export function AnvilWidgetName(props: {
   widgetId: string;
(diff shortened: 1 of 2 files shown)


app/client/src/layoutSystems/anvil/editor/AnvilWidgetName/index.tsx
```diff
@@ -42,10 +43,13 @@ export function AnvilWidgetName(props: {
     (state) => getWidgetErrorCount(state, widgetId) > 0,
   );
 
+  const isParentSelected = useSelector(isWidgetSelected(parentId || ""));
```
KelvinOm, line 46: I think it's better to change `isWidgetSelected` function signature and explicitly handle cases where there is no id.

marks0351 (approved):

app/client/src/layoutSystems/anvil/editor/AnvilWidgetName/index.tsx
riodeuno: Yes, I too believe that is the right approach. I'll make the change.

PR #33652, opened 2024-05-22
fix: Remove the strict check for mongo url

abhvsn:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9191016385>
> Commit: c74e98018a6b4bd212e922eb1b3015195a5234a0
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9191016385&attempt=1" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/CommonDBConfig.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/CommonDBConfig.java
@@ -30,7 +30,8 @@ public class CommonDBConfig {
     @Primary
     @Profile("!test")
     public MongoProperties configureMongoDB() {
-        if (!appsmithDbUrl.startsWith("mongodb")) {
+        log.debug("Configuring MongoDB with url: {}", appsmithDbUrl);
+        if (!appsmithDbUrl.contains("mongodb")) {
             return null;
         }
         log.info("Found MongoDB uri configuring now");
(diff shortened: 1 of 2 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/CommonDBConfig.java
```diff
@@ -30,7 +30,8 @@ public class CommonDBConfig {
     @Primary
     @Profile("!test")
     public MongoProperties configureMongoDB() {
-        if (!appsmithDbUrl.startsWith("mongodb")) {
+        log.debug("Configuring MongoDB with url: {}", appsmithDbUrl);
+        if (!appsmithDbUrl.contains("mongodb")) {
```
mohanarpit, line 34: @abhvsn Can you help me understand why this would fix the issue? If the URL doesn't start with `mongodb`, are we creating the default DB URL to contain this string now?

abhvsn: @mohanarpit Nope we are just returning null if the url doesn't contain `mongodb` keyword and server should fail to start. I'm guessing `startsWith` was failing may be because of additional quotes around the DB url string but not sure. To figure out the issue I have added the logs in the entrypoint as well to update the check.

mohanarpit (approved):

PR #33689, opened 2024-05-23
chore: Add new relic to self hosted image

rajatagrawal:
This PR adds new relic apm to self hosted images by default.

In order to enable new relic apm monitoring for a self hosted instance, the user needs to set following environment variables

```
APPSMITH_NEW_RELIC_ENABLED=true // false by default
APPSMITH_NEW_RELIC_APM_LICENSE_KEY="<license key>" // empty by default
APPSMITH_NEW_RELIC_APM_NAME="<name of the apm app>" // empty by default
```


deploy/docker/fs/opt/appsmith/run-java.sh
```diff
@@ -70,6 +70,8 @@ sh /opt/appsmith/run-starting-page-init.sh &
 exec java ${APPSMITH_JAVA_ARGS:-} ${APPSMITH_JAVA_HEAP_ARG:-} \
   --add-opens java.base/java.time=ALL-UNNAMED \
   --add-opens java.base/java.nio=ALL-UNNAMED \
+  -Dnewrelic.config.agent_enabled=${APPSMITH_NEW_RELIC_ENABLED}
+  -javaagent:/opt/newrelic/newrelic.jar
```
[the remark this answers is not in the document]
rajatagrawal: Noted. Thanks. Change made

deploy/docker/base.dockerfile
```diff
@@ -60,6 +60,9 @@ RUN mv /opt/caddy/caddy /opt/caddy/caddy_vanilla
 
 COPY --from=caddybuilder /usr/bin/caddy /opt/caddy/caddy
 
+# Add newrelic related files
+RUN curl -o /opt/newrelic/newrelic.jar https://download.newrelic.com/newrelic/java-agent/newrelic-agent/8.9.1/newrelic-agent-8.9.1.jar
```
sharat87, line 64: Can we ensure the parent folder exists, or are we doing that somewhere already? 🤔

```suggestion
RUN mkdir -p /opt/newrelic \
  curl -o /opt/newrelic/newrelic.jar https://download.newrelic.com/newrelic/java-agent/newrelic-agent/8.9.1/newrelic-agent-8.9.1.jar
```

deploy/docker/fs/opt/appsmith/newrelic-reconfigure.mjs
```diff
@@ -0,0 +1,26 @@
+import * as fs from "fs"
+
+const path = '/opt/newrelic'
+const templatePath = `${path}/newrelic-template.yml`
+const outputPath = `${path}/newrelic.yml`
+
+function create_new_relic_config_yml_file() {
+    if (!(process.env.APPSMITH_NEW_RELIC_ENABLED == true)) {
```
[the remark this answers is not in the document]
sharat87: Can we fix this please?

deploy/docker/fs/opt/appsmith/run-java.sh
```diff
@@ -70,6 +70,8 @@ sh /opt/appsmith/run-starting-page-init.sh &
 exec java ${APPSMITH_JAVA_ARGS:-} ${APPSMITH_JAVA_HEAP_ARG:-} \
   --add-opens java.base/java.time=ALL-UNNAMED \
   --add-opens java.base/java.nio=ALL-UNNAMED \
+  -Dnewrelic.config.agent_enabled=${APPSMITH_NEW_RELIC_ENABLED} \
+  -javaagent:/opt/newrelic/newrelic.jar \
```
sharat87, line 74: Adding this here would mean that these arguments are present even when NewRelic is not configured, right? Can we add something like this just above the `exec` command instead?

```
if [[ $APPSMITH_NEW_RELIC_ENABLED = true ]]; then
  export APPSMITH_JAVA_ARGS+=" -Dnewrelic.config.agent_enabled=true -javaagent:/opt/newrelic/newrelic.jar"
fi
```

Then those arguments will only be added to the `java` command, if NewRelic is enabled.

deploy/docker/fs/opt/appsmith/templates/docker.env.sh
```diff
@@ -93,4 +93,19 @@ APPSMITH_ALLOWED_FRAME_ANCESTORS="'self' *"
...
+APPSMITH_NEW_RELIC_APPLICATION_ID=''
+APPSMITH_NEW_RELIC_BROWSER_AGENT_LICENSE_KEY=''
+APPSMITH_NEW_RELIC_BROWSER_AGENT_ENDPOINT='bam.nr-data.net'
+APPSMITH_NEW_RELIC_OTEL_EXPORTER_OTLP_ENDPOINT='https://otlp.nr-data.net:4317'
+APPSMITH_NEW_RELIC_OTEL_SERVICE_NAME=''
+OTEL_SERVICE_NAME=''
+APPSMITH_NEWRELIC_MICROMETER_SERVICE_NAME=''
+APPSMITH_NEW_RELIC_OTLP_LICENSE_KEY=''
+APPSMITH_MICROMETER_METRICS_ENABLED=false
+
```
sharat87, line 110: Do we need to add these here? Can we not please?

This script is used to generate an initial `docker.env` for new Appsmith installations. Adding new env variables to this file, doesn't mean existing users get them in their `docker.env`.

I'd recommend adding new env variables to this file as much as possible. Instead, if we want these env variables to be set, let's add them to the `run-java.sh`, inside an `if` block that checks if NewRelic is enabled. We only need these env variables for the backend, right? Since that's where the agent is running?

deploy/docker/fs/opt/newrelic/newrelic-template.yml
```diff
@@ -0,0 +1,400 @@
...
+# This section is for settings common to all environments.
+# Do not add anything above this next line.
+common: &default_settings
+
+  # ============================== LICENSE KEY ===============================
+  # You must specify the license key associated with your New Relic
+  # account. For example, if your license key is 12345 use this:
+  # license_key: '12345'
+  # The key binds your agent's data to your account in the New Relic service.
+  license_key: $APPSMITH_NEW_RELIC_APM_LICENSE_KEY
```
sharat87, line 16: Change to literal block so we don't have to add `'` in the script when doing the replacement. Its simpler and safer in most cases.

```suggestion
  license_key: |
    $APPSMITH_NEW_RELIC_APM_LICENSE_KEY
```

deploy/docker/fs/opt/newrelic/newrelic-template.yml
```diff
@@ -0,0 +1,400 @@
...
+  # Set the name of your application as you'd like it show up in New Relic.
+  # If enable_auto_app_naming is false, the agent reports all data to this application.
+  # Otherwise, the agent reports only background tasks (transactions for non-web applications)
+  # to this application. To report data to more than one application
+  # (useful for rollup reporting), separate the application names with ";".
+  # For example, to report data to "My Application" and "My Application 2" use this:
+  # app_name: My Application;My Application 2
+  # This setting is required. Up to 3 different application names can be specified.
+  # The first application name must be unique.
+  app_name: $APPSMITH_NEW_RELIC_APM_NAME
```
sharat87, line 32: Same as above.

```suggestion
  app_name: |
    $APPSMITH_NEW_RELIC_APM_NAME
```

deploy/docker/fs/opt/appsmith/newrelic-reconfigure.mjs
```diff
@@ -0,0 +1,26 @@
...
+    const values = {
+        $APPSMITH_NEW_RELIC_APM_LICENSE_KEY: process.env.APPSMITH_NEW_RELIC_APM_LICENSE_KEY,
+        $APPSMITH_NEW_RELIC_APM_NAME: process.env.APPSMITH_NEW_RELIC_APM_NAME || "Customer Self Hosted Image"
+    }
+
+    let data = fs.readFileSync(templatePath, "utf-8")
+    
+    for (const key in values) {
+        data = data.replace(key, `'${values[key]}'`)
+    }
```
sharat87, line 21: By using literal blocks for these keys in the yaml file, we don't have to add explicit quotes.

```suggestion
    for (const [key, val] of Object.entries(values)) {
        data = data.replace(key, val)
    }
```

sharat87 (changes requested):

deploy/docker/fs/opt/appsmith/templates/docker.env.sh
rajatagrawal: The user will need to override the value of these environment variables. Would that be possible if these variables are added in run-java.sh ?

rajatagrawal (commented):
Reverted to comments

deploy/docker/base.dockerfile
sharat87: Ah I see, yeah that fine then. Thanks.

deploy/docker/fs/opt/appsmith/newrelic-reconfigure.mjs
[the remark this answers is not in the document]
sharat87: @rajatagrawal, environment variables _always_ come in as strings. They're never of any other type. We have to convert them to a type we need, and store them in a separate variable. The `process.env` is _always_ of type `Record<string, undefined | string>`.

deploy/docker/fs/opt/appsmith/run-java.sh
```diff
@@ -64,6 +64,10 @@ while ! curl --fail --silent localhost:"${APPSMITH_RTS_PORT:-8091}"/rts-api/v1/h
 done
 echo 'RTS started.'
 
+if [[ $APPSMITH_NEW_RELIC_ENABLED = true ]]; then
+  export APPSMITH_JAVA_ARGS+="-javaagent:/opt/newrelic/newrelic.jar"
+fi
```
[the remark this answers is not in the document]
sharat87: @rajatagrawal, this as well please.

deploy/docker/fs/opt/appsmith/templates/docker.env.sh
sharat87: Do we expect them to be able to override _all_ of these? That feels like too much flexibility, no?

> Would that be possible if these variables are added in run-java.sh ?

I don't think overriding them in `run-java.sh` makes much sense. 🤔

Other than `OTEL_SERVICE_NAME`, all of them have an `APPSMITH_` prefix, so these are env variables that we own. So, is the backend using them? Isn't it needed by other processes, like Caddy, for example? I see these env variables mentioned in `index.html`, so I'm guessing we need them in Caddy, right?

A more appropriate place would be `entrupoint.sh` itself. Just before we run the script to generate the YAML configuration file, let's set those env variables, that we need to override, and they should be available to all the processes.

Like we won't set `APPSMITH_NEW_RELIC_APM_LICENSE_KEY` at all. Let it come from `docker.env` or some other place, we don't ever set it. We only look for it, and use it if it's set, but we don't ever set it.

Am I making sense?

rajatagrawal: There are three kinds of telemetry that we have.

1. Backend telemetry (APM) - needs `APPSMITH_NEW_RELIC_ENABLED`, `APPSMITH_NEW_RELIC_APM_LICENSE_KEY` and `APPSMITH_NEW_RELIC_APM_NAME`

2. Frontend web vitals : Needs `APPSMITH_NEW_RELIC_ACCOUNT_ID`, `APPSMITH_NEW_RELIC_APPLICATION_ID`, `APPSMITH_NEW_RELIC_BROWSER_AGENT_LICENSE_KEY`

3. Frontend spans and traces: Needs APPSMITH_NEW_RELIC_OTEL_SERVICE_NAME and `APPSMITH_NEW_RELIC_OTLP_LICENSE_KEY`

4. Backend spans and traces: `OTEL_SERVICE_NAME`, `APPSMITH_MICROMETER_METRICS_ENABLED`, `APPSMITH_NEWRELIC_MICROMETER_SERVICE_NAME`

We want to fill these environment variables only if monitoring is required for self hosted instances and their values would be known only then, since I would manually generate apps/license key on newrelic before sharing them with the customer when a customer needs new relic monitoring.

`APPSMITH_NEW_RELIC_BROWSER_AGENT_ENDPOINT` and `APPSMITH_NEW_RELIC_OTEL_EXPORTER_OTLP_ENDPOINT` can be hard coded by us for now since they will always be constant for newrelic.

rajatagrawal: There are four kinds of telemetry that we have.

Backend telemetry (APM) - needs APPSMITH_NEW_RELIC_ENABLED, APPSMITH_NEW_RELIC_APM_LICENSE_KEY and APPSMITH_NEW_RELIC_APM_NAME

Frontend web vitals : Needs APPSMITH_NEW_RELIC_ACCOUNT_ID, APPSMITH_NEW_RELIC_APPLICATION_ID, APPSMITH_NEW_RELIC_BROWSER_AGENT_LICENSE_KEY

Frontend spans and traces: Needs APPSMITH_NEW_RELIC_OTEL_SERVICE_NAME and APPSMITH_NEW_RELIC_OTLP_LICENSE_KEY

Backend spans and traces: OTEL_SERVICE_NAME, APPSMITH_MICROMETER_METRICS_ENABLED, APPSMITH_NEWRELIC_MICROMETER_SERVICE_NAME

We want to fill these environment variables only if monitoring is required for self hosted instances and their values would be known only then, since I would manually generate apps/license key on newrelic before sharing them with the customer when a customer needs new relic monitoring.

APPSMITH_NEW_RELIC_BROWSER_AGENT_ENDPOINT and APPSMITH_NEW_RELIC_OTEL_EXPORTER_OTLP_ENDPOINT can be hard coded by us for now since they will always be constant for newrelic.

deploy/docker/fs/opt/appsmith/run-java.sh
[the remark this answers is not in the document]
rajatagrawal: Done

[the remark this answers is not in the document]
rajatagrawal: @coderabbitai you missed mentioning change true to 'true'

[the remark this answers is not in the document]
rajatagrawal: Thanks @coderabbitai

PR #33734, opened 2024-05-26
chore: additional test cases and method refactor to follow old standards

sondermanish:
## Description
- Added fallback implementation for autocommit eligibility helper to avoid accessing FS for git connected apps when feature flags are switched off
- Added test cases to verify the same
- modified test cases names to follow standards
- refactored method to follow standard

## Automation

/ok-to-test tags="@tag.Git"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9255023455>
> Commit: f378b03595926844a985da73f9bf46176a27d61d
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9255023455&attempt=1" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [ ] No


app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ce/autocommit/AutoCommitEligibilityHelperImpl.java
```diff
@@ -1,57 +1,93 @@
...
+                            branchName,
+                            serverSchemaVersion);
+                    return JsonSchemaVersions.serverVersion > serverSchemaVersion ? TRUE : FALSE;
                 })
+                .defaultIfEmpty(FALSE)
+                .cache();
+
+        return addFileLock(defaultApplicationId)
+                .then(Mono.defer(() -> isServerMigrationRequiredMonoCached))
+                .then(releaseFileLock(defaultApplicationId))
```
nidhi-nair, line 77: Does this need to be deferred ?

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ce/autocommit/AutoCommitEligibilityHelperImpl.java
```diff
@@ -1,57 +1,93 @@
...
 @RequiredArgsConstructor
-public class AutoCommitEligibilityHelperImpl implements AutoCommitEligibiltyHelper {
+public class AutoCommitEligibilityHelperImpl extends AutoCommitEligibilityHelperFallbackImpl
+        implements AutoCommitEligibilityHelper {
 
     private final CommonGitFileUtils commonGitFileUtils;
     private final DSLMigrationUtils dslMigrationUtils;
+    private final RedisUtils redisUtils;
+
+    private Mono<Boolean> addFileLock(String defaultApplicationId) {
```
nidhi-nair, line 40: Why are we having to define this method here?

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ce/autocommit/AutoCommitEligibilityHelperImpl.java
sondermanish: Since I've declared it before hand itself, it's better it's called lazily when required.

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ce/autocommit/AutoCommitEligibilityHelperImpl.java
sondermanish: I've based it as it's been used in `GitServiceCEImpl`,  `GitAutoCommitHandlerCEImpl` we can define it in one common function but outside the scope of this PR

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ce/autocommit/AutoCommitEligibilityHelperImpl.java
nidhi-nair: I meant the ones that are not deferred :)

sondermanish: Why would that be required? add and release only has publisher code then defer would not make the difference, right? Am i missing something?

nidhi-nair: Doesn't matter if this is a publisher or not, since the method will be eagerly evaluated during the assembly phase of parsing through reactive code. We need that artificial kind of delay to get around this indeterminate behaviour.

sondermanish: what I mean is that step 1 would be followed by step 2 and then by step 3. Do you mean to say that step 1 would be executed much earlier than step 2? and then quickly by step 3?
i.e. -> step 1 ................ step2 ... step 3? 
this probably might happen

nidhi-nair (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/git/GitRedisUtils.java
```diff
@@ -26,6 +26,10 @@ public Mono<Boolean> addFileLock(String defaultApplicationId) {
                         }));
     }
 
+    public Mono<Boolean> addFileLockWithoutRetry(String defaultApplicationId) {
```
nidhi-nair, line 29: Approving the PR due to urgency, but please refactor this call to make retry a parameter instead of a separate method.

nidhi-nair (approved):

PR #33735, opened 2024-05-27
chore: Add deviceType and appMode to new relic telemetry data

dvj1988:
## Description
Add deviceType and appMode attributes to new relic spans related to evaluation

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9297145435>
> Commit: 309d8ae5904c55be9e918b7a119cd184b15cce86
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9297145435&attempt=1" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [x] No


app/client/src/sagas/EvaluationsSaga.ts
```diff
@@ -286,6 +286,7 @@ export function* evaluateTreeSaga(
     evalWorker.request,
     EVAL_WORKER_ACTIONS.EVAL_TREE,
     evalTreeRequestData,
+    { appMode },
```
vsvamsi1, line 289: Do we need this argument?

app/client/src/UITelemetry/generateTraces.ts
```diff
@@ -1,24 +1,40 @@
...
 import { SpanKind } from "@opentelemetry/api";
 import { context } from "@opentelemetry/api";
 import { trace } from "@opentelemetry/api";
+import { deviceType } from "react-device-detect";
 
 const GENERATOR_TRACE = "generator-tracer";
+
+const getCommonTelemetryAttributes = () => {
+  return {
+    deviceType,
```
rajatagrawal, line 16: Can we please add appMode also as common telemetry attribute ?

app/client/src/sagas/EvaluationsSaga.ts
dvj1988: Moved the code to add appMode in common attributes

app/client/src/UITelemetry/generateTraces.ts
dvj1988: Resolved.

app/client/src/UITelemetry/generateTraces.ts
```diff
@@ -1,24 +1,53 @@
...
+  Attributes,
+  TimeInput,
+  SpanOptions,
+} from "@opentelemetry/api";
 import { SpanKind } from "@opentelemetry/api";
 import { context } from "@opentelemetry/api";
 import { trace } from "@opentelemetry/api";
+import { matchBuilderPath, matchViewerPath } from "constants/routes";
+import { APP_MODE } from "entities/App";
+import { deviceType } from "react-device-detect";
```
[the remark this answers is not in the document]
dvj1988: Resolved.

rajatagrawal (approved):

dvj1988:
build-deploy-preview skip-tests=true

app/client/src/UITelemetry/generateTraces.ts
dvj1988: @rajatagrawal The test cases were failing because the worker was trying to access the window.location object which is not available in the worker's scope.
This is happening because we are creating the root span inside the request method of the worker service and this makes it impossible to access the common attributes unless they are passed as an additional argument to the request method. I am reverting the code to the previous state for now.

app/client/src/sagas/EvaluationsSaga.ts
dvj1988: Passing an extra argument here to be treated as common telemetry attributes to the request method of GracefulWorkerService. As you mentioned, appMode is part of the worker task request for the EVAL_TREE task but is not present in the request payload for other tasks. This is an optional argument for the request method, and we can pass different attributes for other tasks if needed.

app/client/src/UITelemetry/generateTraces.ts
```diff
@@ -1,24 +1,54 @@
...
+
+import { APP_MODE } from "entities/App";
+import { matchBuilderPath, matchViewerPath } from "constants/routes";
 
 const GENERATOR_TRACE = "generator-tracer";
+
+const getCommonTelemetryAttributes = () => {
+  const pathname = window.location.pathname;
+  const isEditorUrl = matchBuilderPath(pathname);
```
rajatagrawal, line 19: Can we reuse some state variable to determine app mode ? computing a regex for every span may be computation bottleneck especially because this value isn't going to change.

dvj1988: I tried this using the redux store state. But there are some issues while using it in this specific file. I couldn't get around the root cause here. But this is the issue.

While importing the store in the generateTraces file
import store from "store" (similar to `app/client/src/ce/api/ApiUtils.ts`)

I get this warning in the dev server cli logs
```sh
Circular dependency between chunks with runtime (TernWorker, evalWorker, lintWorker, main)
This prevents using hashes of each other and should be avoided.
```
And the browser console throws the error
```
index.js:47 Uncaught Error: Selector creators expect all input-selectors to be functions, instead received the following types: [function, function, undefined]
    at getDependencies (index.js:47:1)
    at index.js:65:1
    at ./src/selectors/onboardingSelectors.tsx (onboardingSelectors.tsx:59:1)
    at options.factory (react refresh:6:1)
    at __webpack_require__ (bootstrap:24:1)
    at fn (hot module replacement:62:1)
    at ./src/selectors/editorSelectors.tsx (editorContextSelectors.ts:145:1)
    at options.factory (react refresh:6:1)
    at __webpack_require__ (bootstrap:24:1)
    at fn (hot module replacement:62:1)
    ````

rajatagrawal: Can we try this approach https://github.com/appsmithorg/appsmith/blob/fb0bbdfa5cf4360cb291f93ac1d08a1fc1cd95ee/app/client/src/widgets/IframeWidget/component/index.tsx#L137

It may work.

dvj1988: useSelector is a react hook and can be use only in react components.

rajatagrawal (approved):

PR #33777, opened 2024-05-28
chore: Modified logging context to have only requestId and userEmail

nidhi-nair:
## Description
We're modifying the response headers to not hold any info on context, and the logging pattern to only display `requestId`, `userEmail`, `traceId` and `spanId`.

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/filters/MDCFilter.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/filters/MDCFilter.java
@@ -20,7 +20,6 @@ import reactor.util.context.Context;
 import java.util.Map;
 import java.util.UUID;
 
-import static com.appsmith.external.constants.MDCConstants.SESSION_ID_LOG;
 import static com.appsmith.external.constants.MDCConstants.THREAD;
 import static com.appsmith.external.constants.MDCConstants.USER_EMAIL;
 import static java.util.stream.Collectors.toMap;
@@ -44,7 +43,6 @@ public class MDCFilter implements WebFilter {
 
     public static final String REQUEST_ID_HEADER = "X-Request-Id";
 
-    private static final String REQUEST_ID_LOG = "requestId";
     private static final String SESSION = "SESSION";
 
     @Override
@@ -52,7 +50,7 @@ public class MDCFilter implements WebFilter {
         try {
             // Using beforeCommit here ensures that the function `addContextToHttpResponse` isn't run immediately
             // It is only run when the response object is being created
-            exchange.getResponse().beforeCommit(() -> addContextToHttpResponse(exchange.getResponse()));
+            // exchange.getResponse().beforeCommit(() -> addContextToHttpResponse(exchange.getResponse()));
             return ReactiveSecurityContextHolder.getContext()
                     .map(ctx -> ctx.getAuthentication().getPrincipal())
                     .flatMap(principal -> {
@@ -75,22 +73,13 @@ public class MDCFilter implements WebFilter {
             contextMap.put(USER_EMAIL, user.getEmail());
         }
 
-        final String internalRequestId = request.getHeaders().getFirst(INTERNAL_REQUEST_ID_HEADER);
-        if (!StringUtils.isEmpty(internalRequestId)) {
-            contextMap.put(INTERNAL_REQUEST_ID_HEADER, internalRequestId);
-            contextMap.put(REQUEST_ID_LOG, internalRequestId);
-        }
+        final String internalRequestId = getOrCreateInternalRequestId(request);
+        contextMap.put(INTERNAL_REQUEST_ID_HEADER, internalRequestId);
 
         final String requestId = request.getHeaders().getFirst(REQUEST_ID_HEADER);
         if (!StringUtils.isEmpty(requestId)) {
             contextMap.put(REQUEST_ID_HEADER, requestId);
         }
-
-        contextMap.put(SESSION_ID_LOG, getSessionId(request));
-
-        // This is for the initial thread that started the request,
-        // any reactive forking will generate a new thread
-        contextMap.put(THREAD, Thread.currentThread().getName());
 
         // Set the MDC context here for regular non-reactive logs
         MDC.setContextMap(contextMap);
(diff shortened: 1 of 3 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/filters/MDCFilter.java
```diff
@@ -44,15 +43,14 @@ public class MDCFilter implements WebFilter {
...
 
-    private static final String REQUEST_ID_LOG = "requestId";
     private static final String SESSION = "SESSION";
 
     @Override
     public Mono<Void> filter(ServerWebExchange exchange, WebFilterChain chain) {
         try {
             // Using beforeCommit here ensures that the function `addContextToHttpResponse` isn't run immediately
             // It is only run when the response object is being created
-            exchange.getResponse().beforeCommit(() -> addContextToHttpResponse(exchange.getResponse()));
```
nidhi-nair, line 55: We're removing this addition since the only field we were adding to response was the last used thread which is irrelevant information.

app/server/appsmith-server/src/main/java/com/appsmith/server/filters/MDCFilter.java
```diff
@@ -75,23 +73,14 @@ private Context addRequestHeadersToContext(
             contextMap.put(USER_EMAIL, user.getEmail());
         }
 
-        final String internalRequestId = request.getHeaders().getFirst(INTERNAL_REQUEST_ID_HEADER);
-        if (!StringUtils.isEmpty(internalRequestId)) {
-            contextMap.put(INTERNAL_REQUEST_ID_HEADER, internalRequestId);
-            contextMap.put(REQUEST_ID_LOG, internalRequestId);
-        }
+        final String internalRequestId = getOrCreateInternalRequestId(request);
```
nidhi-nair, line 76: This method was unused, but from our discussion it seemed like the intention was to always have this fallback.

app/server/appsmith-server/src/main/resources/application.properties
```diff
@@ -31,7 +31,7 @@ appsmith.codec.max-in-memory-size=${APPSMITH_CODEC_SIZE:150}
 logging.level.root=info
 logging.level.com.appsmith=debug
 logging.level.com.external.plugins=debug
-logging.pattern.console=[%d{ISO8601, UTC}] %X - %m%n
+logging.pattern.console=[%d{ISO8601, UTC}] requestId=%X{X-Appsmith-Request-Id} userEmail=%X{userEmail} traceId=%X{traceId} spanId=%X{spanId} - %m%n
```
nidhi-nair, line 34: This shows up as shown below in the absence of observability.
Screenshot 2024-05-28 at 9 11 22 AM

app/server/appsmith-server/src/main/java/com/appsmith/server/filters/MDCFilter.java
sharat87: This was a miss from me. I intended to delete that function, but missed.

The internal request ID comes in from Caddy, and _only_ from Caddy. Any reason we want to setup a fallback here? If it's only for local testing, when server is run without Caddy, do we need this there?

If we really want backend to _also_ generate internal request IDs, I'd like a way to distinguish the two. Like with a `-backend` suffix etc. Since we're not including it in the response headers, it should be okay.

app/server/appsmith-server/src/main/java/com/appsmith/server/filters/MDCFilter.java
sharat87: Can we remove this code then (instead of commenting it out)?

app/server/appsmith-server/src/main/java/com/appsmith/server/filters/MDCFilter.java
sharat87, line 1: (Writing here since GitHub wouldn't let me post on that line)

Looks like we're still adding the thread in response headers, although we're not adding it to the context map. Not needed I think, no?

sharat87 (commented):
Curious, I see we're removing the thread from the context map, so should we cleanup some tests in `MDCConfigTest`, as stale now? 🤔

app/server/appsmith-server/src/main/java/com/appsmith/server/filters/MDCFilter.java
nidhi-nair: Yep just wanted to confirm if this is okay.

app/server/appsmith-server/src/main/java/com/appsmith/server/filters/MDCFilter.java
nidhi-nair: I see, the caddy bit I missed, yeah in that case I'll delete the method entirely.

app/server/appsmith-server/src/main/java/com/appsmith/server/filters/MDCFilter.java
nidhi-nair: Hey just double checking on this, you've mentioned that adding the X-appsmith-request-id value to response was an expectation [here](https://theappsmith.slack.com/archives/CPQNLFHTN/p1716531188129379?thread_ts=1716530153.424049&cid=CPQNLFHTN). It's not happening right now, but can you confirm we don't need it?

app/server/appsmith-server/src/main/java/com/appsmith/server/filters/MDCFilter.java
nidhi-nair: Nothing is getting added to response headers anymore because of the commented line

sharat87: Yeah but that code that was using it was still there at the time. Look like we're now deleting that also, so this is resolved. Thanks!

app/server/appsmith-server/src/main/java/com/appsmith/server/filters/MDCFilter.java
sharat87: I think there's a miss here.

Caddy should be responsible for that. Caddy should add it to the request, before the request hits backend, and also add it to the response from the backend, before it goes out. Caddy should be doing both.

But it looks like it may not. I'll investigate this and solve it separately. Either way, I'd rather keep the ownership of this header's value in one place, unless we have a strong case for it to be generated in the backend as well.

sharat87 (approved):

PR #33797, opened 2024-05-28
chore: fix table with and without connected data have different heights

### LINKED ISSUE #33758: [Bug]: Table with and without connected data have different heights

### Description

Table with and without connected data have different heights

### Steps To Reproduce

image

### Public Sample App

_No response_

### Environment

Deploy Preview

### Severity

### Issue video log

_No response_

### Version

cloud

jsartisan:
Fixes #33758

/ok-to-test tags="@tag.Anvil"
> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9267730971>
> Commit: 5305b21086cc8965be0f9159d11ebeadf2503878
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9267730971&attempt=1" target="_blank">Click here!</a>

### DIFF
--- a/app/client/src/widgets/wds/WDSTableWidget/component/cellComponents/EmptyCell.tsx
+++ b/app/client/src/widgets/wds/WDSTableWidget/component/cellComponents/EmptyCell.tsx
@@ -6,6 +6,7 @@ import type { ReactTableColumnProps } from "../Constants";
 import { MULTISELECT_CHECKBOX_WIDTH, StickyType } from "../Constants";
 import { EmptyCell, EmptyRow } from "../TableStyledWrappers";
 import { renderBodyCheckBoxCell } from "./SelectionCheckboxCell";
+import { Text } from "@design-system/widgets";
 
 const addStickyModifierClass = (
   columns: ReactTableColumnProps[],
@@ -29,6 +30,7 @@ export const renderEmptyRows = (
   borderRadius: string,
   style?: CSSProperties,
   prepareRow?: (row: Row<Record<string, unknown>>) => void,
+  isHeaderRow: boolean = false,
 ) => {
   const rows: string[] = new Array(rowCount).fill("");
 
@@ -45,7 +47,7 @@ export const renderEmptyRows = (
         },
       };
       return (
-        <div {...rowProps} className="tr" key={index}>
+        <tr {...rowProps} className="tr" key={index}>
           {multiRowSelection && renderBodyCheckBoxCell(false)}
           {row.cells.map(
             (cell: Cell<Record<string, unknown>>, cellIndex: number) => {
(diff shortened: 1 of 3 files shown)


app/client/src/widgets/wds/WDSTableWidget/component/cellComponents/EmptyCell.tsx
```diff
@@ -153,13 +158,16 @@ export const renderEmptyRows = (
                     ? "td hidden-cell"
                     : `td${addStickyModifierClass(columns, colIndex)}`
                 }
-                role="cell"
+                role={isHeaderRow ? "columnheader" : "cell"}
```
jsartisan, line 161: Need to add `role="columnheader"` in case of header cell so that our CSS properly selects thead's cells.

app/client/src/widgets/wds/WDSTableWidget/component/cellComponents/EmptyCell.tsx
```diff
@@ -153,13 +158,16 @@ export const renderEmptyRows = (
...
                 {...stickyAttributes}
+                data-column-type="text"
                 key={colIndex}
                 sticky={column?.sticky ?? StickyType.NONE}
                 style={{ ...distanceFromEdge }}
                 width={column.width}
-              />
+              >
+                <Text data-empty-text="">&#8203;</Text>
```
jsartisan, line 169: Need to Text component so that height of the empty cell is same as the filled one.

app/client/src/widgets/wds/WDSTableWidget/component/cellComponents/EmptyCell.tsx
```diff
@@ -74,13 +76,16 @@ export const renderEmptyRows = (
                       ? "td hidden-cell"
                       : `td${addStickyModifierClass(columns, cellIndex)}`
                   }
+                  data-column-type="text"
```
jsartisan, line 79: need to add this attribute so that our css picks this element to style.

KelvinOm (commented):
@jsartisan Why do we need all these changes if we plan to add the default data into the table? All this code for an empty table should go away, right?

jsartisan:
> @jsartisan Why do we need all these changes if we plan to add the default data into the table? All this code for an empty table should go away, right?

@KelvinOm Right but this changes also makes sense for deploy mode so that empty table is of same height. It's not just for edit mode.

KelvinOm:
> > @jsartisan Why do we need all these changes if we plan to add the default data into the table? All this code for an empty table should go away, right?
> 
> @KelvinOm Right but this changes also makes sense for deploy mode so that empty table is of same height. It's not just for edit mode.

So if we have default data, they will also displayed in deploy mode, we should not have a case with an empty table. Or am I missing something?

jsartisan:
> > > @jsartisan Why do we need all these changes if we plan to add the default data into the table? All this code for an empty table should go away, right?
> > 
> > @KelvinOm Right but this changes also makes sense for deploy mode so that empty table is of same height. It's not just for edit mode.
> 
> So if we have default data, they will also displayed in deploy mode, we should not have a case with an empty table. Or am I missing something? 

Consider a table that is binded with a data source and and data is empty from it.

KelvinOm (approved):

Issue #33766, opened 2024-05-27
[Bug]: Forking to same workspace or different workspace brings up error on Airgapped image


btsgh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Forking to same workspace or different workspace brings up error on Air-gapped image

### Steps To Reproduce

1. Pull air-gapped nightly.
2. Create a new app with one datasource.
3. Fork the app into the same or a new workspace.
4. Note that this errors out with a message - we could not connect to our servers - check network connection. 
5. On refresh though, we see that the app is created in the workspace
6. Reproducible with simple json files also.

Attaching json:
[newapp.json](https://github.com/appsmithorg/appsmith/files/15456424/newapp.json)

### Environment

Production

### Severity

High (Blocker to building or releasing)

### Version

Airgap

btsgh:
https://github.com/appsmithorg/appsmith/assets/101863839/bab12c9e-0162-4e28-b462-b600844c2ab9

btsgh:
Slack [thread](https://theappsmith.slack.com/archives/CTHN8GX5Y/p1716966252788189?thread_ts=1716818085.111689&cid=CTHN8GX5Y) with more discussion around this.

PR #33831, opened 2024-05-29
fix: ignore DSL parsing errors in the forking flow

AnaghHegde:
## Description
The expectation for any server side flow today that is using `updateLayout` exclusively is to ignore DSL parsing or similar kind of errors and proceed assuming that the next client side interaction will fix the problem. The actual DSL issue is a problem coming up from incorrectly written DSL migrations which for these apps won't make a difference even if we fix them retroactively. 

Fixes https://github.com/appsmithorg/appsmith/issues/33766

## Automation

/ok-to-test tags="@tag.Fork"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9300457429>
> Commit: 93cfeab4ff0a9822ace1c1890910f69e3a2ae66a
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9300457429&attempt=2" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [ ] No


app/server/appsmith-server/src/main/java/com/appsmith/server/fork/internal/ApplicationForkingServiceCEImpl.java
```diff
@@ -348,6 +348,10 @@ public Mono<List<String>> forkApplications(
                 })
                 .flatMapIterable(tuple2 -> clonedPages)
                 .flatMap(clonedPage -> updateLayoutService.updatePageLayoutsByPageId(clonedPage.getId()))
+                .onErrorResume(throwable -> {
+                    log.error("Error while parsing DSL {} ", throwable.getMessage());
```
nidhi-nair, line 352: Let's catch a more specific error, other kinds of failures shouldn't be ignored

AnaghHegde: Done

app/server/appsmith-server/src/main/java/com/appsmith/server/fork/internal/ApplicationForkingServiceCEImpl.java
```diff
@@ -349,8 +349,11 @@ public Mono<List<String>> forkApplications(
                 .flatMapIterable(tuple2 -> clonedPages)
                 .flatMap(clonedPage -> updateLayoutService.updatePageLayoutsByPageId(clonedPage.getId()))
                 .onErrorResume(throwable -> {
```
nidhi-nair, line 351: Can we use a comparison by type instead of a string compare here please?

AnaghHegde: I thought of that but the type of exception will be same right? For example - if the DSL is corrupted or has a missing element still the parsing will fail. Only the message will be different. Hence i am doing comparison by string and not type.

nidhi-nair: I think it's safe to assume that the only kind of failures we expect to see here are incorrect bindingPaths. If there were other kinda of failures, they would surface in normal updateLayout calls as well.

AnaghHegde: If i have to filter by exception type it is going to be `AppsmithException` and even if this flow fails for any other reason also the exception type will still be same. Hence i am filtering via the error message.

nidhi-nair: I'm sorry, looks like we're misunderstanding what a type check means. An error type check in Appsmith is always meant to check for `AppsmithError` types, since all of our caught exceptions would be `AppsmithException` type. By making a string check here, we're making the code brittle because tomorrow when the error message copy is altered for any reason, this flow will stop catching that error. Is this making sense?

Let's also add a test case to avoid this breaking in the future.

app/server/appsmith-server/src/test/java/com/appsmith/server/solutions/ForkingServiceTest.java
```diff
@@ -0,0 +1,97 @@
+package com.appsmith.server.solutions;
```
nidhi-nair, line 1: Nit, this test belongs in the fork package since we have moved to structuring classes by features and domains

app/server/appsmith-server/src/main/java/com/appsmith/server/fork/internal/ApplicationForkingServiceCEImpl.java
```diff
@@ -348,6 +348,16 @@ public Mono<List<String>> forkApplications(
                 })
                 .flatMapIterable(tuple2 -> clonedPages)
                 .flatMap(clonedPage -> updateLayoutService.updatePageLayoutsByPageId(clonedPage.getId()))
+                .onErrorResume(throwable -> {
+                    if (throwable instanceof AppsmithException e) {
```
nidhi-nair, line 352: Nit, this can be a single check instead of two successive checks.

nidhi-nair (changes requested):
LGTM, needs a few nitpick changes.

app/server/appsmith-server/src/main/java/com/appsmith/server/fork/internal/ApplicationForkingServiceCEImpl.java
AnaghHegde: @nidhi-nair Done. Can you check the PR again?

nidhi-nair (approved):

PR #33901, opened 2024-06-01
test: fix flaky iconButton2 spec

NandanAnantharamu:
RCA: exeternal website used in this case "[Yahoo.com](http://yahoo.com/)"

Solution:

so replacing third party url should fix our problem.
Will use intercept/TED link to solve this as we did for our flaky tests earlier

### DIFF
--- a/app/client/cypress/e2e/Regression/ClientSide/Widgets/Others/IconButton_2_spec.ts
+++ b/app/client/cypress/e2e/Regression/ClientSide/Widgets/Others/IconButton_2_spec.ts
@@ -270,13 +270,22 @@ describe(
       propPane.ToggleJSMode("onClick", true);
       propPane.UpdatePropertyFieldValue(
         "onClick",
-        `{{navigateTo('www.yahoo.com', {}, 'SAME_WINDOW');}}`,
+        `{{navigateTo('http://host.docker:4200/', {}, 'NEW_WINDOW');}}`,
       );
       deployMode.DeployApp(
         locators._widgetInDeployed(draggableWidgets.ICONBUTTON),
       );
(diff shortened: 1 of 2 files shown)


app/client/cypress/e2e/Regression/ClientSide/Widgets/Others/IconButton_2_spec.ts
```diff
@@ -270,13 +270,22 @@ describe(
       propPane.ToggleJSMode("onClick", true);
       propPane.UpdatePropertyFieldValue(
         "onClick",
-        `{{navigateTo('www.yahoo.com', {}, 'SAME_WINDOW');}}`,
+        `{{navigateTo('http://host.docker:4200/', {}, 'NEW_WINDOW');}}`,
```
ApekshaBhosale, line 273: should we be hardcoding this URL? can we not take it from config file or something? @NandanAnantharamu

ApekshaBhosale (changes requested):

app/client/cypress/e2e/Regression/ClientSide/Widgets/Others/IconButton_2_spec.ts
NandanAnantharamu: the link doesnt matter what we enter as the call is intercepted it wont matter what we enter here. let me know if you still want me to change @ApekshaBhosale

ApekshaBhosale: got it

ApekshaBhosale (approved):

PR #33921, opened 2024-06-03
chore: added unique index for slug field in tenant (#33377)

### LINKED ISSUE #33377: [Task]: Add mongoDB unique index for slug field in Tenant collection
### SubTasks

The `Tenant` domain `slug` field has an `@Unique` [annotation](https://github.com/appsmithorg/appsmith/blob/d1c9bde439138f0d6d4b414ff001f0fc07c93027/app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Tenant.java#L21) but there is no unique index for slug field in the mongo database. We should add an index in the database to maintain consistentency. 

image

NilanshBansal:
Fixes #33377 

## Automation

/ok-to-test tags="@tag.All"

> [!CAUTION]  

## Communication
- [ ] Yes
- [x] No

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration056AddUniqueIndexForSlugInTenant.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration056AddUniqueIndexForSlugInTenant.java
@@ -22,6 +22,6 @@ public class Migration056AddUniqueIndexForSlugInTenant {
 
     @Execution
     public void executeMigration() {
-        ensureIndexes(mongoTemplate, Tenant.class, makeIndex(Tenant.Fields.slug));
+        ensureIndexes(mongoTemplate, Tenant.class, makeIndex(Tenant.Fields.slug).unique());
     }
 }
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration056AddUniqueIndexForSlugInTenant.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration056AddUniqueIndexForSlugInTenant.java
@@ -1,0 +1,27 @@
+package com.appsmith.server.migrations.db.ce;
+
+import com.appsmith.server.domains.Tenant;
+import io.mongock.api.annotations.ChangeUnit;
+import io.mongock.api.annotations.Execution;
+import io.mongock.api.annotations.RollbackExecution;
+import org.springframework.data.mongodb.core.MongoTemplate;
+
+import static com.appsmith.server.migrations.DatabaseChangelog1.ensureIndexes;
+import static com.appsmith.server.migrations.DatabaseChangelog1.makeIndex;
+
+@ChangeUnit(order = "056", id = "add-unique-index-for-slug-in-tenant")
+public class Migration056AddUniqueIndexForSlugInTenant {
+    private final MongoTemplate mongoTemplate;
+
+    public Migration056AddUniqueIndexForSlugInTenant(MongoTemplate mongoTemplate) {
+        this.mongoTemplate = mongoTemplate;
+    }
+
+    @RollbackExecution
+    public void executionRollback() {}
+
+    @Execution
+    public void executeMigration() {
+        ensureIndexes(mongoTemplate, Tenant.class, makeIndex(Tenant.Fields.slug));
+    }
+}


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration056AddUniqueIndexForSlugInTenant.java
```diff
@@ -0,0 +1,27 @@
...
+    public Migration056AddUniqueIndexForSlugInTenant(MongoTemplate mongoTemplate) {
+        this.mongoTemplate = mongoTemplate;
+    }
+
+    @RollbackExecution
+    public void executionRollback() {}
+
+    @Execution
+    public void executeMigration() {
+        ensureIndexes(mongoTemplate, Tenant.class, makeIndex(Tenant.Fields.slug));
```
subrata71: Is the intention to create only index or unique index? The action doesn't match the name of this migration.

subrata71: Don't you need to add any other fields e.g. `deleted`, `deletedAt`, `instanceId` etc?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration056AddUniqueIndexForSlugInTenant.java
```diff
@@ -0,0 +1,27 @@
...
+import io.mongock.api.annotations.ChangeUnit;
+import io.mongock.api.annotations.Execution;
+import io.mongock.api.annotations.RollbackExecution;
+import org.springframework.data.mongodb.core.MongoTemplate;
+
+import static com.appsmith.server.migrations.DatabaseChangelog1.ensureIndexes;
+import static com.appsmith.server.migrations.DatabaseChangelog1.makeIndex;
+
+@ChangeUnit(order = "056", id = "add-unique-index-for-slug-in-tenant")
+public class Migration056AddUniqueIndexForSlugInTenant {
```
subrata71, line 13: What if there are already records in the `tenant` collection which violate the uniqueness in the slug field? Do you think writing another migration would be required for the sanity of the data?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration056AddUniqueIndexForSlugInTenant.java
```diff
@@ -0,0 +1,27 @@
...
+        this.mongoTemplate = mongoTemplate;
+    }
+
+    @RollbackExecution
+    public void executionRollback() {}
+
+    @Execution
+    public void executeMigration() {
+        ensureIndexes(mongoTemplate, Tenant.class, makeIndex(Tenant.Fields.slug));
+    }
```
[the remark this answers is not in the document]
NilanshBansal: Done

subrata71 (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration056AddUniqueIndexForSlugInTenant.java
NilanshBansal: Thanks for the comment @subrata71! Done that now.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration056AddUniqueIndexForSlugInTenant.java
NilanshBansal: There are no records in the Tenant collection which violates this because the field was defined as @Unique always in the Domain.
https://github.com/appsmithorg/appsmith/blob/4b50a7f0ec0c9fd89bd093e87b2b5a052f189bac/app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Tenant.java#L23

subrata71: Apologies if I misunderstood it. But using the `@Unique` annotation on the POJO level doesn't ensure uniqueness in the DB, right? It only ensures uniqueness at the application level. Please correct me if I am wrong here. @NilanshBansal

NilanshBansal:
Closing this PR.
Ref: https://github.com/appsmithorg/appsmith/issues/33377#issuecomment-2144763283

PR #33975, opened 2024-06-04
chore: fix export service

vivonk:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/BearerTokenAuth.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/BearerTokenAuth.java
@@ -16,6 +16,6 @@ import lombok.ToString;
 @NoArgsConstructor
 @AllArgsConstructor
 public class BearerTokenAuth extends AuthenticationDTO {
-    @Encrypted @JsonView({Views.Public.class, FromRequest.class})
+    @Encrypted @JsonView({Views.Internal.class, FromRequest.class})
     String bearerToken;
 }
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/BearerTokenAuth.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/BearerTokenAuth.java
@@ -2,6 +2,7 @@ package com.appsmith.external.models;
 
 import com.appsmith.external.annotations.encryption.Encrypted;
 import com.appsmith.external.views.FromRequest;
+import com.appsmith.external.views.Views;
 import com.fasterxml.jackson.annotation.JsonView;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
@@ -15,6 +16,6 @@ import lombok.ToString;
 @NoArgsConstructor
 @AllArgsConstructor
 public class BearerTokenAuth extends AuthenticationDTO {
-    @Encrypted @JsonView(FromRequest.class)
+    @Encrypted @JsonView({Views.Public.class, FromRequest.class})
     String bearerToken;
 }
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/domains/GitArtifactMetadata.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/domains/GitArtifactMetadata.java
@@ -71,7 +71,7 @@ public class GitArtifactMetadata implements AppsmithDomain {
     @JsonView(Views.Public.class)
     String docUrl;
 
-    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd'T'HH:mm:ssX", timezone = "UTC")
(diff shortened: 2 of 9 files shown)


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/BearerTokenAuth.java
```diff
@@ -15,6 +16,6 @@
 @NoArgsConstructor
 @AllArgsConstructor
 public class BearerTokenAuth extends AuthenticationDTO {
-    @Encrypted @JsonView(FromRequest.class)
+    @Encrypted @JsonView({Views.Public.class, FromRequest.class})
```
abhvsn, line 19: Is this expected on client side, if not can we change the view to internal in that case?

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/GitArtifactMetadata.java
```diff
@@ -71,7 +71,7 @@ public class GitArtifactMetadata implements AppsmithDomain {
     @JsonView(Views.Public.class)
     String docUrl;
 
-    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd'T'HH:mm:ssX", timezone = "UTC")
+    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd'T'HH:mm:ss.SSSX", timezone = "UTC")
```
abhvsn, line 74: Any reason we are changing the granularity here?

app/server/appsmith-server/src/test/java/com/appsmith/server/exports/internal/ExportServiceTests.java
```diff
@@ -1816,6 +1815,9 @@ public void exportApplication_WithPageIcon_ValidPageIcon() {
                 .assertNext(applicationJson -> {
                     List<NewPage> pages = applicationJson.getPageList();
                     assertThat(pages).hasSize(2);
+                    // sort based on page name so that first page is the default created page
+                    pages.sort(Comparator.comparing(
+                            page -> page.getUnpublishedPage().getName()));
```
abhvsn, line 1820: Are we expecting some sorting in exported file, else this test would have been too flaky. I would rather want this to fix in release by filtering the page with name first and then asserting for the icon. With sorting we will again hit with flakiness in future.

abhvsn (commented):
Minor changes requested.

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/GitArtifactMetadata.java
vivonk: What was happening that test cases that are somewhere comparing the `lastUpdatedAt`, have precision of milliseconds and since this date was just at seconds, it was causing false failures. So I don't think this precious is a good solution for us in long term.

app/server/appsmith-server/src/test/java/com/appsmith/server/exports/internal/ExportServiceTests.java
vivonk: I think this is caused because of MongoDB data retrieval vs Postgres data retrieval. In both databases, the order of records retrieved is different so we want to be on the same page at least on postgres.
But agree, long-term, get this fix from release branch itself. 
I have noted it and will be following up on it soon after server tests.

vivonk: Resolve with the following code and soon will be bringing this change from release
```

                    NewPage page = pages.stream()
                            .filter(page1 ->
                                    page1.getUnpublishedPage().getName().equals("page_" + randomId))
                            .findFirst()
                            .orElse(null);
                    assertThat(page).isNotNull();
                    assertThat(page.getUnpublishedPage().getName()).isEqualTo("page_" + randomId);
                    assertThat(page.getUnpublishedPage().getIcon()).isEqualTo("flight");
```

app/server/appsmith-server/src/test/java/com/appsmith/server/imports/internal/ImportServiceTests.java
```diff
@@ -4597,6 +4597,7 @@ public void exportApplication_WithPageIcon_ValidPageIcon() {
         StepVerifier.create(resultMono)
                 .assertNext(applicationJson -> {
                     List<NewPage> pages = applicationJson.getPageList();
+                    pages.sort(Comparator.comparing(p -> p.getUnpublishedPage().getName()));
```
abhvsn, line 4600: @vivonk let's update the assertion here as well. Also share the ticket when we want to make the same change in release.

vivonk: Thanks for pointing this out

abhvsn (approved):

PR #33987, opened 2024-06-05
chore: Update postgres version references

abhvsn:
## Description
PR to update the version references for postgres. This will help us in keeping the diff minimum between release and pg branch as we will be upgrading the postgres version to 15 on pg branch.

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **Chores**
  - Updated Docker scripts to use a generic placeholder for PostgreSQL version, enhancing flexibility and future-proofing the setup.

### DIFF
--- a/deploy/docker/fs/opt/appsmith/entrypoint.sh
+++ b/deploy/docker/fs/opt/appsmith/entrypoint.sh
@@ -424,17 +424,18 @@ init_postgres() {
       # Postgres does not allow it's server to be run with super user access, we use user postgres and the file system owner also needs to be the same user postgres
       chown postgres:postgres "$POSTGRES_DB_PATH"
 
+      # Please refer to base.Dockerfile for the installation of postgres
       # Initialize the postgres db file system
-      su postgres -c "/usr/lib/postgresql/13/bin/initdb -D $POSTGRES_DB_PATH"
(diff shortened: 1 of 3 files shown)


deploy/docker/fs/opt/appsmith/entrypoint.sh
```diff
@@ -424,17 +424,18 @@ init_postgres() {
       # Postgres does not allow it's server to be run with super user access, we use user postgres and the file system owner also needs to be the same user postgres
       chown postgres:postgres "$POSTGRES_DB_PATH"
 
+      # Please refer to base.Dockerfile for the installation of postgres
```
sharat87, line 427:
```suggestion
      # Please refer to base.dockerfile for the installation of Postgres
```

sharat87 (approved):
Looks good. Minor, can we use `current` instead of `current_version`, may be? Unless we expect a conflict on that name, shorter and equally informative I think.

So, since this involves a change in `base.dockerfile`, we'll have to do it with two phases. I'll go ahead and "merge" the changes for `base.dockerfile`, then we'll let the base image with symlink build, and _then_ we can let the rest of the changes in this PR go in, so we'll merge this PR. As is, if we merge the whole thing in one go, all DP builds using the current base image, will fail to find the symlink. 🙂

PR #33995, opened 2024-06-05
fix: fix application fork tests pg

vivonk:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/UploadedFile.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/UploadedFile.java
@@ -1,9 +1,10 @@
 package com.appsmith.external.models;
 
 import com.appsmith.external.annotations.encryption.Encrypted;
+import com.appsmith.external.views.FromRequest;
 import com.appsmith.external.views.Views;
-import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonView;
+import jakarta.persistence.Transient;
 import lombok.AllArgsConstructor;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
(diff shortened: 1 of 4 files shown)


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/UploadedFile.java
```diff
@@ -26,11 +27,11 @@ public class UploadedFile implements AppsmithDomain {
     @JsonView(Views.Public.class)
     String name;
 
-    @Encrypted @JsonProperty(access = JsonProperty.Access.WRITE_ONLY)
-    @JsonView(Views.Public.class)
+    @Encrypted @JsonView({Views.Internal.class, FromRequest.class})
```
abhvsn, line 30: @vivonk let's get this change from release branch. I'm suggesting this because when we change views we are changing the way client-server talk to each other. This can only be catch via cypress tests today and unfortunately we are not there yet on pg branch. Moreover debugging cyress tests are harder compared to server tests in my experience may be because I haven't spent enough time there.

From prima-facie I can see we have added `FromRequest` only for base64Content but not on other field i.e. `name`, are we not expecting this to be updated by client? Also can you help me in which controller we are updating these fields and if `FromRequest` view is applied on the DTO object that is being de-serialised.

abhvsn (changes requested):
Changes requested via release branch.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/UploadedFile.java
```diff
@@ -26,11 +27,11 @@ public class UploadedFile implements AppsmithDomain {
...
     String base64Content;
 
     @JsonView(Views.Internal.class)
+    @Transient
```
sharat87, line 34: Can you check again once, this shouldn't be necessary now.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/UploadedFile.java
vivonk: UploadedFile is part of datasource configuration deep down and datasource related controllers manipulate datasourceConfiguration. That makes, datasource controller is your PoC here.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/UploadedFile.java
vivonk: this is still needed else two tests are failing with JsonObject cannot be converted into byte array

abhvsn (approved):

PR #34013, opened 2024-06-06
chore: adding baseline data

vivonk:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No


app/server/envs/dev.env.example
```diff
@@ -28,3 +28,4 @@ APPSMITH_CLOUD_SERVICES_BASE_URL="https://release-cs.appsmith.com"
 
 #APPSMITH_RECAPTCHA_SITE_KEY=""
 #APPSMITH_RECAPTCHA_SECRET_KEY=""
+APPSMITH_BASE_DATA_PATH="deploy/docker/fs/opt/appsmith"
```
sharat87, line 31: Any reason we want to make this configurable? Can we not hard-code this in the migration and let it be? 🤔

deploy/docker/fs/opt/appsmith/baseline-ce-data/actionCollection.jsonl
sharat87, line 1: Can we remove the empty files in this folder please?

sharat87 (changes requested):

app/server/envs/dev.env.example
vivonk: @sharat87 Hard coding is not an easy job. Reason is: working directory can be variable. 
If I am running tests in IntellIIJ, working directory is different, it's module directory.
If I am running server as service in IntellIJ, working directory is root folder.
Because of the execution reason, we want this path as configuration.

Another way around is moving these jsonl to resources because file size is so small that we don't have any additional size impact.
cc: @abhvsn

deploy/docker/fs/opt/appsmith/baseline-ce-data/actionCollection.jsonl
vivonk: yes, was thinking the same

app/server/envs/dev.env.example
sharat87: Thanks for sharing the reasoning. In that case I'd still urge to make this smart and not configurable.

Idea is that what can be computed and figured out smartly, we'll have the computers do the work for us. Otherwise, we'll be expecting every developer to manually understand and fiddle with this configuration. Extra manual effort for something the program should've been able to deduce by itself.

What say? (I'm also changing `baseline-ce-data` to just `baseline-ce`, shorter and equally informative I think, but feel free to disagree 🙂)

shot-2024-06-06-08-33-36

```java
    private static Path findEffectiveDataPath() {
        // MongoDB export placed in Stacks volume, in Docker container.
        final Path mongoDataInStacks = Path.of("appsmith-stacks", MONGO_DATA_NAME);
        if (Files.exists(mongoDataInStacks)) {
            return mongoDataInStacks;
        }

        // Find where `opt/appsmith` folder is located.
        // In Docker container, we'll find it at `/opt/appsmith`.
        // In dev systems, we look for the `deploy` folder, navigating upwards.
        Path optAppsmith = Path.of("opt", "appsmith");
        if (!Files.exists(optAppsmith)) {
            Path current = Path.of(".").toAbsolutePath();
            while (!Files.exists(current.resolve("deploy"))) {
                current = current.getParent();
                if (current == null) {
                    throw new RuntimeException("No baseline folder found");
                }
            }
            optAppsmith = current.resolve(Path.of("deploy", "docker", "opt", "appsmith"));
        }

        // Is there MongoDB export in optAppsmith?
        if (Files.exists(optAppsmith.resolve(MONGO_DATA_NAME))) {
            return optAppsmith.resolve(MONGO_DATA_NAME);
        }

        // Otherwise, just return baseline data path.
        return optAppsmith.resolve("baseline-" + ProjectProperties.EDITION.toLowerCase());
    }
```

sharat87 (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/ce/V002__loadMongoData.java
```diff
@@ -34,21 +34,45 @@ public class V002__loadMongoData extends AppsmithJavaMigration {
...
+        return optAppsmith.resolve("baseline-" + ProjectProperties.EDITION.toLowerCase());
+    }
 
     @Override
     public void migrate(JdbcTemplate jdbcTemplate) throws Exception {
-        final boolean isOperatingOnBaselineData = !MONGO_DATA_ROOT.toFile().exists();
-
-        final Path effectiveDataRoot = isOperatingOnBaselineData ? BASELINE_DATA_ROOT : MONGO_DATA_ROOT;
+        final Path effectiveDataRoot = findEffectiveDataPath();
+        final boolean isCustomerExistingDataPresent = effectiveDataRoot.endsWith(MONGO_DATA_NAME);
```
sharat87, line 75: The boolean value is being inverted here, so it looks like we're doing ID replacements in MongoDB exported data, and not in the baseline data. Can you verify the logic on this once please?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/ce/V002__loadMongoData.java
```diff
@@ -34,21 +34,45 @@ public class V002__loadMongoData extends AppsmithJavaMigration {
...
+    private static final String MONGO_DATA_NAME = "mongo-data";
 
     final Map<String, String> idMap = new HashMap<>();
 
-    final Path BASELINE_DATA_ROOT =
-            Path.of("/Users/shri/work/appsmith-ee-pg/deploy/docker/fs/opt/appsmith/utils/bin/baseline-"
-                    + ProjectProperties.EDITION.toLowerCase() + "-data");
+    private static Path findEffectiveDataPath() {
+        // MongoDB export placed in Stacks volume, in Docker container.
+        final Path mongoDataInStacks = Path.of("appsmith-stacks", MONGO_DATA_NAME);
```
sharat87, line 43: Looks like this is building relative paths. 🤦

```suggestion
        final Path mongoDataInStacks = Path.of("/", "appsmith-stacks", MONGO_DATA_NAME);
```

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/ce/V002__loadMongoData.java
```diff
@@ -34,21 +34,45 @@ public class V002__loadMongoData extends AppsmithJavaMigration {
...
+        if (Files.exists(mongoDataInStacks)) {
+            return mongoDataInStacks;
+        }
+
+        // Find where `opt/appsmith` folder is located.
+        // In Docker container, we'll find it at `/opt/appsmith`.
+        // In dev systems, we look for the `deploy` folder, navigating upwards.
+        Path optAppsmith = Path.of("opt", "appsmith");
```
sharat87, line 51:
```suggestion
        Path optAppsmith = Path.of("/", "opt", "appsmith");
```

app/server/envs/dev.env.example
sharat87, line 1: Can we have this file not show up in the diff please?

app/server/envs/test.env.example
sharat87, line 1: Can we have this file not show up in the diff please? It doesn't look like there's any significant changes. Thanks!

deploy/docker/fs/opt/appsmith/baseline-ce/newAction.jsonl
sharat87, line 1: Empty file_

sharat87 (changes requested):

sharat87 (approved):

PR #34089, opened 2024-06-07
fix: display label instead of value in select column in table widget #26188

SaiCharanChetpelly31:
Fixed issue [26188](https://github.com/appsmithorg/appsmith/issues/26188)

This pull request adds a new feature called "Display as" for the select column type in the table widget.
image

- Whenever a column is set to select, by default we show labels and not values.
- image
- For existing tables, since we have created this problem we ensure we don't break their apps. We will give an option called "Display As" as suggested in the original issue and for new select columns, it will be Label and for existing columns, it will be Value.
- image
- We make it clear that the computed value here refers to the default value/display value. Displaying this as the helper text below the computed value property only for select column type.
- Added a helper text below the computed value field that indicates that the computed value here is a default value/display value
- image
- Validation:When computed value is not present in the select option then we throw an error in select option. This is irrespective of displayAs property.
- image
-For new row select options, there is no validation of computed value. Since are adding new data into the table and it has to adhere to the select options that we provide.
image

- For displaying value in cell,
       When display as is set to label, we show empty values for missing labels in the widget
       image
       When display as is set to value, we show the computed value as it is even if labels are missing
       image

- When computed value is not present in the select options, we still make the select options available. We do this because its not the select options that has issue but its computed value.
- 
image

## Summary by CodeRabbit

- **New Features**
  - Enhanced `SelectCell` component to allow customization of display behavior, showing either the label or value of selected options.

- **Bug Fixes**
  - Fixed issues related to the display properties of select columns in the `TableWidgetV2` component.

- **Tests**
  - Added tests for the display properties of select columns in the `TableWidgetV2` component.

- **Refactor**
  - Updated property configurations and utility functions to improve handling of select column display properties.


app/client/src/constants/PropertyControlConstants.tsx
```diff
@@ -114,8 +116,6 @@ export interface PropertyPaneControlConfig {
    */
   controlConfig?: Record<string, unknown>;
   defaultValue?: unknown;
-  /** used to mark a property as reusable so that it can be reused in next dropping widget */
-  isReusable?: boolean;
```
jsartisan, line 118: Any reason for removing this?

app/client/src/widgets/TableWidgetV2/widget/propertyConfig/PanelConfig/Data.ts
```diff
@@ -395,24 +409,18 @@ export default {
       enableSearch: true,
       dropdownHeight: "156px",
       controlType: "DROP_DOWN",
-      customJSControl: "TABLE_COMPUTE_VALUE",
```
jsartisan, line 398: This is very risky.

jsartisan:
Hey sorry but I am closing the PR as TableWidget is a very complex widget and requires a lot of context to fix bugs in it.
It would be great if you could pick issues marked with a "Good first" label.

app/client/src/widgets/TableWidgetV2/widget/propertyConfig/PanelConfig/Data.ts
SaiCharanChetpelly31: @jsartisan I will try to fix the review comments. can you please reopen the PR?

PR #34090, opened 2024-06-07
fix: fixing null check failure on pg branch

vivonk:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the stability of the layout retrieval process by adding null checks for both `publishedPage` and `unpublishedPage` and their respective `layouts`. This prevents potential errors when these elements are missing.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/domains/NewPage.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/domains/NewPage.java
@@ -50,6 +50,9 @@ public class NewPage extends BranchAwareDomain implements Context {
     @JsonView(Views.Internal.class)
     @Override
     public Layout getLayout() {
+        if (this.getPublishedPage() == null || this.getPublishedPage().getLayouts() == null) {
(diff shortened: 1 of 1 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/domains/NewPage.java
```diff
@@ -50,6 +50,9 @@ public String getArtifactId() {
     @JsonView(Views.Internal.class)
     @Override
     public Layout getLayout() {
+        if (this.getPublishedPage() == null || this.getPublishedPage().getLayouts() == null) {
```
abhvsn, line 53: We should be checking for unpublished page.

abhvsn (changes requested):

abhvsn (approved):

PR #34099, opened 2024-06-07
chore: exclude button from tab order

KelvinOm:
## Description
Removed the option to select the tab button in edit mode. 

This is the first PR. There will be separate PRs for each widget. 

## Automation

/ok-to-test tags="@tag.Anvil"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9445961677>
> Commit: ccba28bd2b17a7afdc27fb59371f7e72bd370169
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9445961677&attempt=1" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **New Features**
  - Introduced `isEditMode` getter in widgets to enhance edit mode detection.
  
- **Improvements**
  - Updated button component properties for better consistency and usability.
  - Enhanced widget interaction logic for more accurate behavior in different modes.

- **Removed**
  - Eliminated unnecessary `minWidth`, `maxWidth`, and `minHeight` properties from button components.

### DIFF
--- a/app/client/src/widgets/BaseWidget.tsx
+++ b/app/client/src/widgets/BaseWidget.tsx
@@ -368,6 +368,12 @@ abstract class BaseWidget<
     return this.props.layoutSystemType === LayoutSystemTypes.AUTO;
   }
 
+  get isEditMode() {
+    return (
+      this.props.renderMode === RenderModes.CANVAS && !this.props.isPreviewMode
+    );
+  }
+
   updateOneClickBindingOptionsVisibility(visibility: boolean) {
     const { updateOneClickBindingOptionsVisibility } = this.context;
     if (visibility) {
@@ -549,6 +555,7 @@ export interface WidgetProps
     DataTreeEvaluationProps {
   key?: string;
   isDefaultClickDisabled?: boolean;
+  disableWidgetInteraction?: boolean;
 
   [key: string]: any;
 }
--- a/app/client/src/widgets/BaseWidget.tsx
+++ b/app/client/src/widgets/BaseWidget.tsx
@@ -368,12 +368,6 @@ abstract class BaseWidget<
     return this.props.layoutSystemType === LayoutSystemTypes.AUTO;
   }
 
-  get isEditMode() {
-    return (
-      this.props.renderMode === RenderModes.CANVAS && !this.props.isPreviewMode
-    );
-  }
-
   updateOneClickBindingOptionsVisibility(visibility: boolean) {
     const { updateOneClickBindingOptionsVisibility } = this.context;
     if (visibility) {
@@ -555,7 +549,6 @@ export interface WidgetProps
     DataTreeEvaluationProps {
   key?: string;
   isDefaultClickDisabled?: boolean;
-  disableWidgetInteraction?: boolean;
 
   [key: string]: any;
 }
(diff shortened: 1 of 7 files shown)


app/client/src/widgets/BaseWidget.tsx
```diff
@@ -368,6 +368,12 @@ abstract class BaseWidget<
     return this.props.layoutSystemType === LayoutSystemTypes.AUTO;
   }
 
+  get isEditMode() {
+    return (
+      this.props.renderMode === RenderModes.CANVAS && !this.props.isPreviewMode
```
KelvinOm, line 373: I'm not sure these checks are enough. We have `isEditOnlyModeSelector` selector, it has a rather complicated logic for checking the same thing. I'm not sure that we need all this. 

cc @riodeuno @marks0351

riodeuno: @KelvinOm @marks0351 Where is going to be used?
In the first version of Appsmith back in 2020, we had something called `getCanvasView` for this. But, I don't see which widget in WDS will need this.

KelvinOm: @riodeuno In edit mode, we don't allow to set focus on a widget using tabulation.

[See this prop ](https://github.com/appsmithorg/appsmith/pull/34099/files#diff-17b513ce1c79052950d3ba72981d75c19ac965e04afa06d3121a80f89c1976b0R156)

`excludeFromTabOrder={this.isEditMode}`

marks0351: @KelvinOm can't we make use of [`allowWidgetInteraction`](https://github.com/appsmithorg/appsmith/blob/release/app/client/src/layoutSystems/anvil/index.ts#L18)  be enough for this usecase?

marks0351: we have added the `isEditOnlyModeSelector`, to deal with some usecases in the layout components to avoid prop drilling. for widgets all info is already available and present on the widget props it selves we append them in [withWidgetProps](https://github.com/appsmithorg/appsmith/blob/release/app/client/src/widgets/withWidgetProps.tsx#L246)

app/client/src/widgets/BaseWidget.tsx
```diff
@@ -549,6 +555,7 @@ export interface WidgetProps
     DataTreeEvaluationProps {
   key?: string;
   isDefaultClickDisabled?: boolean;
+  disableWidgetInteraction?: boolean;
```
KelvinOm, line 558: Added it explicitly to the interface. The line below allows us to add anything. 😔

marks0351: `disableWidgetInteraction` is an optional argument however it will only be available for anvil widgets, so I would suggest to add an interface for now to coallate all anvil based props.
Ideally once we have separated Basewidget, all this would be easier to implement.

KelvinOm: What other props are specific to Anvil? Due to the fact that `[key: string]: any;` allows us to send anything, it is difficult to find all these things. However, this should be a separate task.

app/client/src/widgets/BaseWidget.tsx
marks0351: @KelvinOm do we still need this change?

marks0351 (commented):
have added one last [comment](https://github.com/appsmithorg/appsmith/pull/34099#discussion_r1634372931).

app/client/src/widgets/BaseWidget.tsx
KelvinOm: @marks0351 Oddd, I remember deleting it. You are right, I don't need that anymore. Thanks for pointing it out. I have fixed it.

marks0351 (approved):

PR #34143, opened 2024-06-10
fix: pass correct widgetId for reflow of building blocks

### LINKED ISSUE #34139: [Bug]: building blocks do not get dropped on the canvas when there is reflow

### Description

When i drag a building block into the canvas with existing widgets, and the building block causes any other widget on the canvas to move, the building block drop errors out and does not work.

### Steps To Reproduce

1. Go to a new workspace and create a new app
2. Drag and drop a text widget or any other widget unto the canvas
3. Drag a building block unto the canvas, using the block to move the already dropped widget
4. Watch that the block fails to get dropped and shows an error

### Public Sample App

_No response_

### Environment

Production

### Severity

### Issue video log

_No response_

### Version

Cloud - 1.28.0

jacquesikot:
## Description
**Problem**
When dragging a dropping a building block on the canvas with reflow (i.e when other widgets are moved) the process fails. The building block does not show, the skeleton loader does not go away and an error message is shown.

**Why**
The flow for adding building blocks with reflow does not contain a widgetId in the newWidget object, instead, the payload carries a canvasId parameter.

**Solution**
I have added a check to add the correct value for the widgetId in the flow for dropping building blocks on canvas with reflow.

Fixes #34139 

## Automation

/ok-to-test tags="@tag.Templates"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9448541112>
> Commit: fc8cfe04c784888514f119b0b51b67ea60014910
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9448541112&attempt=1" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the stability of the widget loading process by ensuring the correct widget ID is assigned when adding and moving building blocks to the canvas.

### DIFF
--- a/app/client/src/sagas/BuildingBlockSagas/BuildingBlockAdditionSagas.ts
+++ b/app/client/src/sagas/BuildingBlockSagas/BuildingBlockAdditionSagas.ts
@@ -381,7 +381,10 @@ export function* addAndMoveBuildingBlockToCanvasSaga(
   );
   yield call(
     loadBuildingBlocksIntoApplication,
-    actionPayload.payload.newWidget,
+    {
+      ...actionPayload.payload.newWidget,
+      widgetId: actionPayload.payload.canvasId,
+    },
     skeletonWidget.widgetId,
   );
 }


rahulbarwal (changes requested):
Can we add a cypress test for both scenarios of dropping so that this does not regress:
1. Without reflow(I think we already have it)
2. With reflow

rahulbarwal (approved):

PR #34176, opened 2024-06-11
fix: tag copied cut widgets with layout system type

### LINKED ISSUE #34159: [Task]: Tag copied widgets with layout system type

### Summary

When pasting widgets into a page, we need to be careful about which layout system the originally copied widgets belong.
For this, we will now being tagging the copied widgets with the layout system type, and confirm if the page in which the widgets are being pasted, belong to the same layout system.

We have Anvil and Fixed layout systems today. To make sure that these two layout systems can co-exist, we need certain safeguards with respect to their interoperability.
By making sure that we’re not moving widgets to and from different layout systems, we’ll be maintaining the integrity of the system and prevent the system from breaking due to incompatible widgets.

### LINKED ISSUE #34160: [Task]: Tag exported partial apps with the layout system type
### SubTasks

- [ ] When importing a partial app into an incompatible layout system, we risk breaking systems. So, if a widget is added to an export, the layout system should also be tagged along with the exported json.
- [ ] When importing a partial app, we need to make sure that the layout systems of the target app is compatible with the one being imported.
- [ ] This check is not necessary if no widgets are being exported.

riodeuno:
## Description
- Adds a tag to copied and exported (partial export) widget data, that contains the layout system type of the source page.
- When pasting or importing the above tag is used to verify if the layout system type of the current page and source page matches.
- If it doesn't match, a message is shown in a toast notification and the paste/import skips adding widgets.
- In the case of partial import, everything else is imported except for widgets, if the layout systems donot match.

Note: Since partial import/export use the paste saga to add widgets to the page, the following have been handled:
- The message shown to the user is generic and doesn't specify paste or import.
- Both fixes have been added in the same PR.
  
Fixes #34159 
Fixes #34160

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [x] No


riodeuno:
Sounds like a neat idea. I'll do this when looking into separating Anvil copy/cut/paste from Fixed mode. Paste is already separated. I'll deal with the copy and cut flows as well.

marks0351 (approved):

marks0351 (approved):

PR #34185, opened 2024-06-11
feat: side-by-side edit mode hover analytics

### LINKED ISSUE #33159: [Task]: Analytics for side by side canvas
### SubTasks

Must haves:
- [ ] Trigger an event whenever a user hovers on the Canvas when in side-by-side mode
- [ ] This event should have a property which mentions if the user navigated or not
- [ ] This event should have a property which mentions if a user hovers on a widget or not
Nice to have:
- [ ] A property which lists the types of widgets the user hovered on.

Examples:
If a user hovers on the canvas, but on no widgets: Event triggers 
Event fired CANVAS_HOVER {
  navigated: false, 
  widgetHover: false,
  widgetTypes: []
}

If a user navigates using cmd+click on a text widget:
Event fired CANVAS_HOVER {
  navigated: true, 
  widgetHover: true,
  widgetTypes: [Text]
}

If a user doesn't navigate but only hovers on 3 widgets: table1, table2, and select1
Event fired CANVAS_HOVER {
navigated: false, 
widgetHover: true,
widgetTypes: [Table, Select]
}

alex-golovanov:
## Description
The purpose of this PR is to collect data about hovered canvas and widgets and send it to analytics platform.

Fixes #33159

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9584384829>
> Commit: 9cec247f7f1e82e0cc23fa5aa5499008bdc58964
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9584384829&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: ``

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **New Features**
  - Introduced analytics tracking for side-by-side hover actions in the IDE canvas.
  - Added a new `AnalyticsWrapper` component for handling analytics events within layout systems.
  - Implemented a custom hook `useIsInSideBySideEditor` to check for side-by-side editor mode.
  - Added new CSS styles for layout systems.

- **Bug Fixes**
  - Improved handling of widget hover events and analytics tracking in side-by-side editor mode.

- **Refactor**
  - Refactored logic for determining the current entity info and segment state in the IDE.
  - Updated import paths and reordered imports for better organization.


app/client/src/utils/hooks/useIsInSideBySideEditor.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Hook to check if current component is in side-by-side editor mode.
+ */
+const useIsInSideBySideEditor = () => {
+  const { pathname } = useLocation();
+  const ideViewMode = useSelector(getIDEViewMode);
+
+  const isInSideBySideEditor =
+    ideViewMode === EditorViewMode.SplitScreen &&
+    pathname.includes(JS_COLLECTION_EDITOR_PATH);
```
hetunandu, line 17: The correct definition here would be
( isSideBySideEnabled && appState === EditorState.Editor && editorMode === EditorViewMode.SplitScreen && segment !== EditorEntityTab.UI && )

use hooks: useCurrentAppState and useCurrentEditorState


this hook should be placed in file `pages/Editor/IDE/hooks.ts`

app/client/src/reducers/uiReducers/analyticsReducer.ts
```diff
@@ -1,19 +1,59 @@
+import type { ReduxAction } from "@appsmith/constants/ReduxActionConstants";
 import { ReduxActionTypes } from "@appsmith/constants/ReduxActionConstants";
 import { createReducer } from "utils/ReducerUtils";
 
 export type SegmentState = "INIT_SUCCESS" | "INIT_UNCERTAIN";
 
 export const initialState: AnalyticsReduxState = {
   telemetry: {},
+  ideCanvasSideBySideHover: {
```
hetunandu, line 9: This state does not belong here. Can we move it to `uiReducers/ideReducer.ts`?

app/client/src/actions/analyticsActions.ts
```diff
@@ -7,3 +7,22 @@ export const segmentInitSuccess = () => ({
...
+  type: ReduxActionTypes.SEND_ANALYTICS_FOR_SIDE_BY_SIDE_HOVER,
+});
+
+export const recordAnalyticsForSideBySideNavigation = () => ({
+  type: ReduxActionTypes.RECORD_ANALYTICS_FOR_SIDE_BY_SIDE_NAVIGATION,
+});
+
+export const resetAnalyticsForSideBySideHover = () => ({
+  type: ReduxActionTypes.RESET_ANALYTICS_FOR_SIDE_BY_SIDE_HOVER,
+});
```
hetunandu, line 28: We should probably avoid doing this via redux actions and sagas. Incase it is found that is the best way to do it, such actions need a new file that is specific for analytics tracking related actions and not this current file which has actions to setup and manage the analytics lib

hetunandu (changes requested):

app/client/src/layoutSystems/CanvasFactory.tsx
```diff
@@ -27,7 +36,27 @@ const LayoutSystemBasedCanvas = memo((props: WidgetProps) => {
     ],
   );
   const { Canvas, propertyEnhancer } = canvasSystem;
-  return <Canvas {...propertyEnhancer(props)} />;
+
+  const isInSideBySideEditor = useIsInSideBySideEditor();
```
marks0351, line 40: For Canvas, I would suggest to make changes on the editor layer of each layout system. 
we have three layout systems, Fixed, Auto Layout and Anvil.
Fixed is the legacy layout system by default, Auto Layout has been deprecated and Anvil is being actively developed.

So Ideally these changes would go on the FixedLayoutEditorCanvas and AnvilEditorCanvas. 

Why not in this layer?
because this canvas factory is common for all layout systems and also for all modes(edit/preview/deploy). since our requirement is only to collect analytics in edit mode, adding these enhancements only to the editor would be better. 
WDYT?

app/client/src/layoutSystems/common/widgetName/index.tsx
```diff
@@ -181,6 +192,13 @@ export function WidgetNameComponent(props: WidgetNameComponentProps) {
     showAsSelected,
   ]);
 
+  const isInSideBySideEditor = useIsInSideBySideEditor();
+  useEffect(() => {
```
marks0351, line 196: I would suggest to add a separate middleware saga that takes ReduxActionTypes.FOCUS_WIDGET which is dispatched whenever a widget is hovered.

marks0351 (changes requested):

app/client/src/reducers/uiReducers/analyticsReducer.ts
alex-golovanov: It seems like ideReducer is not a good place for this, as we're collecting analytics, which do not have direct relation to IDE from code composition perspective (even though we're collecting data on IDE usage).

app/client/src/actions/analyticsActions.ts
alex-golovanov: Could you elaborate on this?  How should we approach building and accessing common state for unrelated components?

app/client/src/layoutSystems/CanvasFactory.tsx
alex-golovanov: Makes sense, will move it directly to editors.

app/client/src/layoutSystems/common/widgetName/WidgetNameLayer.tsx
```diff
@@ -1,4 +1,4 @@
-import WidgetNameComponent from "layoutSystems/common/widgetName";
+import WidgetNameComponent from "./";
```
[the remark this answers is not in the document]
alex-golovanov: It was intentional, as I think that module-like/alias imports should be used only for external entities, but I will return the old import, since it's a common approach on the project.

app/client/src/layoutSystems/common/widgetName/index.tsx
alex-golovanov: That is a great suggestion.  I've initially looked at ReduxActionTypes.FOCUS_WIDGET, but discarded it as I thought that it was not hover but focus, it appears to be otherwise and perfectly suits this task (seems like not the best name for this action type).

marks0351: U r right. when this action was first introducted it indeed worked like focus, however recent implementations have changed how it functions.

app/client/src/layoutSystems/CanvasFactory.tsx
alex-golovanov: Done.

app/client/src/layoutSystems/common/widgetName/index.tsx
alex-golovanov: Refactored, please have a look.

marks0351 (approved):

hetunandu (dismissed):

hetunandu (approved):

PR #34197, opened 2024-06-12
fix: set eval version for workflows artefact 

### LINKED ISSUE #34171: [Bug]: adding `"\n"` to the workflow code causes the evaluation to fail

### Description

When I add `\n` to my workflow code, the evaluation fails.
Side effect: When the PUT call to update the actions in JS object happens, it removes the function which has errors in it.

### Steps To Reproduce

Create a workflow and use `\n` in it.

### Public Sample App

_No response_

### Environment

Production

### Severity

Critical (Broken Production apps)

### Issue video log

_No response_

### Version

1.28_SNAPSHOT

ayushpahwa:
## Description
Eval version 2 is used to properly handle the escape characters like `\n` etc in the eval flow. If not set, it is defaulting to behaviour for eval version 1. This PR introduces the type fixes to allow for eval version to be set via the API. The main change is in the EE PR and this PR is to update the type. 

Fixes #34171 

## Automation

/test sanity

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9478023950>
> Commit: c5edfb33f93802d2a025a064ddaa4593ea7932a4
> Cypress dashboard url: <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9478023950&attempt=1" target="_blank">Click here!</a>

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **Refactor**
  - Consolidated `EvaluationVersion` type imports to a single source for improved maintainability.
  - Promoted `DEFAULT_EVALUATION_VERSION` to be exportable for broader use across the application.

### DIFF
--- a/app/client/src/ce/constants/WorkflowConstants.ts
+++ b/app/client/src/ce/constants/WorkflowConstants.ts
@@ -18,7 +18,7 @@ export interface Workflow {
   tokenGenerated: boolean;
   // Evaluation version of the application. Used to ensure the escape characters are properly evaluated.
   // PR for reference: https://github.com/appsmithorg/appsmith/pull/8796
-  evaluationVersion?: EvaluationVersion;
+  evaluationVersion: EvaluationVersion;
   token?: string;
 }
 
--- a/app/client/src/ce/constants/WorkflowConstants.ts
+++ b/app/client/src/ce/constants/WorkflowConstants.ts
@@ -1,4 +1,4 @@
-import type { EvaluationVersion } from "@appsmith/api/ApplicationApi";
+import type { EvaluationVersion } from "constants/EvalConstants";
 
 type ID = string;
 
--- a/app/client/src/ce/constants/WorkflowConstants.ts
+++ b/app/client/src/ce/constants/WorkflowConstants.ts
@@ -16,6 +16,8 @@ export interface Workflow {
   slug: string; // Slug of the workflow (Not in use currently).
   mainJsObjectId: string; // ID of the main JS object.
   tokenGenerated: boolean;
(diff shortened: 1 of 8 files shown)


rajatagrawal (approved):

app/client/src/ce/constants/WorkflowConstants.ts
```diff
@@ -14,6 +16,9 @@ export interface Workflow {
   slug: string; // Slug of the workflow (Not in use currently).
   mainJsObjectId: string; // ID of the main JS object.
   tokenGenerated: boolean;
+  // Evaluation version of the application. Used to ensure the escape characters are properly evaluated.
+  // PR for reference: https://github.com/appsmithorg/appsmith/pull/8796
+  evaluationVersion?: EvaluationVersion;
```
rajatagrawal, line 21: Is it required to have evaluationVersion as optional ? Can it be non optional ?

ayushpahwa: It is going to be optional till server starts sending it in. In EE PR, I am setting a default value. Once server starts sending, it will be enforced.

rajatagrawal: If there is a default value provided, then it can be assumed that evaluation version would always be present. Would that be the case ? From client's perspective, evaluation version is always present, either server value or default value.

ayushpahwa: Updated

rajatagrawal (approved):

PR #34209, opened 2024-06-12
fix: visually tag anvil application card

### LINKED ISSUE #34162: [Task]: Tag an Anvil application card in the applications page
### SubTasks

- [ ] To easily differentiate between a Fixed layout system app and an Anvil app, the apps listed in the applications page need to be visually differentiated. We will be adding a tag for now to make this distinction clear to users.

riodeuno:
## Description
Visually tag Anvil apps to differentiate them from non-anvil apps.
Screenshot from 2024-06-12 16-43-58

Fixes #34162 

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [x] No

### DIFF
--- a/app/client/src/components/common/Card.tsx
+++ b/app/client/src/components/common/Card.tsx
@@ -28,6 +28,7 @@ type CardProps = PropsWithChildren<{
   titleTestId: string;
   isSelected?: boolean;
   hasEditPermission?: boolean;
+  showAnvilTag: boolean;
 }>;
 
 interface NameWrapperProps {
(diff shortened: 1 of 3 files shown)


app/client/src/components/common/Card.tsx
```diff
@@ -28,6 +28,7 @@ type CardProps = PropsWithChildren<{
   titleTestId: string;
   isSelected?: boolean;
   hasEditPermission?: boolean;
+  showAnvilTag: boolean;
```
ashit-rath, line 31: Can we make this prop optional and have the default value as `false`? This prop is not relevant for `Card` usage in  PackageCard and WorkflowCard in EE and would definitely break TS checks when this gets merged into EE

riodeuno: Sounds good.
I'll make this change.

PR #34220, opened 2024-06-13
chore: Remove `deleted` boolean field

sharat87:
Remove the `deleted` field since

1. it'll be removed from `release` branch soon.
2. we aren't creating the corresponding `deleted` column in database tables anyway.


sharat87:
Merging since the tests are passing locally.

PR #34224, opened 2024-06-13
fix: display label instead of value in select column of table widget

SaiCharanChetpelly31:
Fixed issue [26188](https://github.com/appsmithorg/appsmith/issues/26188)

This pull request adds a new feature called "Display as" for the select column type in the table widget.
image

- Whenever a column is set to select, by default we show labels and not values.
- image
- For existing tables, since we have created this problem we ensure we don't break their apps. We will give an option called "Display As" as suggested in the original issue and for new select columns, it will be Label and for existing columns, it will be Value.
- image
- We make it clear that the computed value here refers to the default value/display value. Displaying this as the helper text below the computed value property only for select column type.
- Added a helper text below the computed value field that indicates that the computed value here is a default value/display value
- image
- Validation:When computed value is not present in the select option then we throw an error in select option. This is irrespective of displayAs property.
- image
-For new row select options, there is no validation of computed value. Since are adding new data into the table and it has to adhere to the select options that we provide.
image

- For displaying value in cell,
       When display as is set to label, we show empty values for missing labels in the widget
       image
       When display as is set to value, we show the computed value as it is even if labels are missing
       image

- When computed value is not present in the select options, we still make the select options available. We do this because its not the select options that has issue but its computed value.
- 
image

## Summary by CodeRabbit

- **New Features**
  - Enhanced table widget with new `selectDisplayAs` property to display options as label or value.
  - Updated `PropertyControl` component to support enhanced `helperText` functionality.

- **Improvements**
  - Added helper text `COMPUTED_VALUE_SELECT_HELPER_TEXT` for clearer display of computed values in the table widget.

- **Bug Fixes**
  - Fixed dropdown selection functionality in the `AggregateHelper` class.

- **Tests**
  - Introduced comprehensive tests for new `selectDisplayAs` property and related functionalities in the table widget.

### DIFF
--- a/app/client/cypress/e2e/Regression/ClientSide/Widgets/TableV2/columnTypes/Select3_spec.ts
+++ b/app/client/cypress/e2e/Regression/ClientSide/Widgets/TableV2/columnTypes/Select3_spec.ts
@@ -1,0 +1,228 @@
+import {
+  entityExplorer,
+  propPane,
+  agHelper,
+  draggableWidgets,
+  table,
+  locators,
+} from "../../../../../../support/Objects/ObjectsCore";
+import {
+  getWidgetSelector,
+  WIDGET,
+} from "../../../../../../locators/WidgetLocators";
+
+const tableData = `[
+    {
+      "id": 2,
+      "name": "saicharan",
+      "gender": "M"
+    },
+    {
+      "id": 3,
+      "name": "Raju",
+      "gender": "M"
+    },
+    {
+      "id": 4,
+      "name": "Rajesh",
+      "gender": "M"
+    },
+    {
+      "id": 5,
+      "name": "Lahari",
+      "gender": "F"
+    },
+    {
+      "id": 6,
+      "name": "Sneha",
+      "gender": "F"
+    },
+    {
+      "id": 7,
+      "name": "Varshini",
+      "gender": "F"
+    },
+    {
+      "id": 8,
+      "name": "Vidmahi",
+      "gender": "F"
+    },
+    {
+      "id": 9,
+      "name": "nikhil",
+      "gender": "M"
+    }
+  ]`;
+const updatedTableData = `[
+    {
+      "id": 2,
+      "name": "saicharan",
+      "gender": "27"
+    },
+    {
+      "id": 3,
+      "name": "Raju",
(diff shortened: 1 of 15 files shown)


SaiCharanChetpelly31:
HI @ApekshaBhosale @marks0351 can you please review this PR?

jsartisan:
Let me check this today.

jsartisan:
@SaiCharanChetpelly31 I tried the PR on my local and couldn't get it working. For some reason, the select options always shows failed to validate for me
CleanShot 2024-06-20 at 12 46 10

SaiCharanChetpelly31:
@jsartisan can you try giving options in this way?
image

jsartisan:
Hey @SaiCharanChetpelly31 I am closing the PR as it does not work as intended. Also please check our [contribution guidelines. ](https://github.com/appsmithorg/appsmith/issues/34308)

SaiCharanChetpelly31:
@jsartisan can you please try to give array of objects?
image
image

jsartisan:
I tried the same. Passed an array of objects. It still does not work. 😕
CleanShot 2024-06-20 at 12 57 40

Anyway, I would suggest you to pick something from the good first issue. By the way, I am really impressed that you understood the code of the table widget without any help. Great job on it but really sorry, we won't be accepting table widget contribution PR for now. It really consumes a lot of time for us and we are tight on bandwidth right now 🥲 Really sorry for this.

SaiCharanChetpelly31:
> I tried the same. Passed an array of objects. It still does not work. 😕 CleanShot 2024-06-20 at 12 57 40
> 
> Anyway, I would suggest you to pick something from the good first issue. By the way, I am really impressed that you understood the code of the table widget without any help. Great job on it but really sorry, we won't be accepting table widget contribution PR for now. It really consumes a lot of time for us and we are tight on bandwidth right now 🥲 Really sorry for this.

@jsartisan Sure i will pick good first issues.
It is giving error for you because **it can accept only string or number for value property. But you gave boolean.**
If it is working as intended, can you please add a comment for the same?

Thanks in advance.

jsartisan:
Tried that too
CleanShot 2024-06-20 at 13 19 05

PR #34229, opened 2024-06-13
chore: Use `JsonForDatabase` in `BaseAppsmithRepositoryCEImpl`

sharat87:


sharat87:
Merging as tests pass locally.

Issue #34064, opened 2024-06-07
[Task]: Use standard naming convention strategy for index names in postgres


vivonk:
### SubTasks

- Go through existing indexes and add `NULLS NOT DISTINCT`
- Renaming indexes as per this convention - https://stackoverflow.com/questions/4107915/postgresql-default-constraint-names/4108266#4108266

vivonk:
This is completed, closing this

Issue #34042, opened 2024-06-06
[Task]: Take care of pointers for postgres migrations - Debt


vivonk:
### SubTasks

- Convert all the repeated migrations.
- index Constraints.
- When importing baseline data, let `created_at` take current time and `updated_at` be `null`.
-  (EE only) The `TenantConfiguration.pluginToken` should be encrypted. How do we do this in the `loadMongoData` migration? This isn't getting encrypted in `release` branch today, a bug. That's why we found it with our regex.
If we fix that bug, and we need this encrypted, we'll need to solve for this as well.

vivonk:
Regarding 
>  (EE only) The TenantConfiguration.pluginToken should be encrypted. How do we do this in the loadMongoData migration? This isn't getting encrypted in release branch today, a bug. That's why we found it with our regex.
If we fix that bug, and we need this encrypted, we'll need to solve for this as well.

Since this field itself will be removed going forward and there is no use of this field, we can live with non-encrypted values as well.

vivonk:
Taken care, closing them

PR #34235, opened 2024-06-13
chore: improvements in codebase - index, migrations for postgres

vivonk:
Fixes https://github.com/appsmithorg/appsmith/issues/34042
Fixes https://github.com/appsmithorg/appsmith/issues/34064
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/ce/V003__createIndexes.sql
```diff
@@ -1,14 +1,10 @@
...
-create unique index if not exists u_workspace_app_deleted_git_application_metadata on application(workspace_id, name, deleted_at, (git_application_metadata ->> 'remoteUrl'), (git_application_metadata ->>'branchName'));
-create unique index if not exists u_workspace_app_git_application_metadata on application(workspace_id, name, (git_application_metadata ->> 'remoteUrl'), (git_application_metadata ->>'branchName')) WHERE deleted_at is NULL;
-create unique index if not exists u_dsConfigStructure_dsId_envId on datasource_storage_structure(datasource_id, environment_id);
-create unique index if not exists u_applicationId_chunkOrder on application_snapshot(application_id, chunk_order);
+create unique index if not exists config_name_key on config(name) ;
+create unique index if not exists prt_email_key on password_reset_token(email);
+create unique index if not exists user_email_key on "user"(email);
+create unique index if not exists sequence_name_key on sequence(name);
+create unique index if not exists plugin_name_package_name_version_idx on plugin(plugin_name, package_name, version) NULLS NOT DISTINCT;
```
sharat87, line 5: Any reason we're using `_idx` here? It's a unique index like every other unique index, right? 🤔

vivonk: Actually I have a thought here. Since calling all the single field indexes with `_key` suffix make sense, at the same time, multiple field indexes as `_key` sound a bit difference so I chose to stick with `_idx` incase of multi-field constraint. 
if you think `_key` is better for all, I will rename them.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/ce/V002__loadMongoData.java
```diff
@@ -143,6 +145,9 @@ private void moveForTable(Path jsonlPath, JdbcTemplate jdbcTemplate, boolean isC
                         data.put(snakeKey, value);
                     }
                 }
+                if (shouldUpdateTimestamps(isCustomerExistingDataPresent, tableName)) {
+                    updateTimestamps(data);
+                }
```
sharat87, line 150: Can we instead check if `columnTypes` map has `created_at` key, and if yes, set it to current time in `data`, and same for `updated_at`? This way, we don't have to maintain the `NON_BASE_DOMAIN_TABLES`. What say?

Something like:

```java
if (columnTypes.containsKey("created_at")) {
    data.put("created_at", Instant.now().toString());
}
if (columnTypes.containsKey("updated_at")) {
    data.put("updated_at", null);
}
```

sharat87 (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/ce/V002__loadMongoData.java
vivonk: Ah, great, agree with you there

sharat87 (approved):

Issue #34041, opened 2024-06-06
[Task]: Add new assert for password encryption in DatasourceServiceTest 


vivonk:
### SubTasks

- Update test
- checkEncryptionOfAuthenticationDTOTest
- checkEncryptionOfAuthenticationDTOAfterUpdate

vivonk:
Fixed in the above PR, closing this

PR #34253, opened 2024-06-14
chore: password encrypted value check test datasource

vivonk:
Fixes https://github.com/appsmithorg/appsmith/issues/34041
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/server/appsmith-server/src/test/java/com/appsmith/server/services/DatasourceServiceTest.java
+++ b/app/server/appsmith-server/src/test/java/com/appsmith/server/services/DatasourceServiceTest.java
@@ -1,5 +1,6 @@
 package com.appsmith.server.services;
 
+import com.appsmith.external.helpers.EncryptionHelper;
 import com.appsmith.external.models.ActionConfiguration;
 import com.appsmith.external.models.ActionDTO;
 import com.appsmith.external.models.Connection;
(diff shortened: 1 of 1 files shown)


app/server/appsmith-server/src/test/java/com/appsmith/server/services/DatasourceServiceTest.java
```diff
@@ -1086,17 +1098,41 @@ public void checkEncryptionOfAuthenticationDTOTest() {
...
                     assertThat(authentication.getUsername()).isEqualTo(username);
-                    // TODO :: Fetch record from DB and check if password is encrypted
-                    assertThat(authentication.getPassword()).isEqualTo(password);
+                    assertThat(EncryptionHelper.decrypt(authentication.getPassword()))
+                            .isEqualTo(password);
                 })
                 .verifyComplete();
     }
 
+    private DatasourceStorage getDatasourceStorage(String datasourceId, String envId) {
```
sharat87, line 1115: Instead of changing so much in the test, can we make a minimal and focused change please? Ideally, we don't want to touch the tests as much as possible between `release` and `pg` so we can be sure we're asserting the same way as much as possible.

So in this case, the problem is only the "password" assertion part of the whole test. So let's just get the password from the database, and assert just that. Something like this can get the password you need:

shot-2024-06-15-06-54-50

Then, only the password assertion needs to be changed to this:

```java
assertThat(EncryptionHelper.decrypt(getDatasourcePassword(savedDatasource.getId(), defaultEnvironmentId)))
                            .isEqualTo(password);
```

This way, we only fetch the very minimum piece of information we want from the database.

What do you think?

sharat87 (changes requested):
Reporting a suggestion, let me know if you disagree. 🙂

app/server/appsmith-server/src/test/java/com/appsmith/server/services/DatasourceServiceTest.java
vivonk: Agree, making these changes

app/server/appsmith-server/src/test/java/com/appsmith/server/services/DatasourceServiceTest.java
```diff
@@ -84,6 +87,8 @@ public class DatasourceServiceTest {
     @Autowired
     DatasourceService datasourceService;
 
+    ObjectMapper objectMapper = new ObjectMapper();
```
sharat87, line 90: Looks like we don't need to add this.

sharat87 (changes requested):
Looks good. One minor item and we should merge.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/DatasourceServiceTest.java
vivonk: Agree

sharat87 (commented):
❤️

sharat87 (approved):

PR #34338, opened 2024-06-19
perf: add react-hooks/exhaustive-deps rule

### LINKED ISSUE #34337: Add `react-hooks/exhaustive-deps` rule with warn level
**Problem**
Right now there are no automated checks for hook dependencies. This leads to incorrect cache invalidation and, subsequently, can cause (is causing?) unpredictable UI behaviour.

**Solution**
Add `react-hooks/exhaustive-deps`.  At this point in time we need to run it in `warn` level, until critical mass is reached, at which point we can switch it to `error`.

Is it worth using in warning mode? Absolutely. Developers will get immediate feedback on hook dependencies and will be able to compose their code correctly.

alex-golovanov:
## Description
Adds ESLint rule `react-hooks/exhaustive-deps` with warn level. Context in the task below.

Fixes #34337 

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9584250724>
> Commit: daa705074504f398b80b1175440d00011b5d865d
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9584250724&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: ``

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **Chores**
  - Added ESLint rule `react-hooks/exhaustive-deps` with a severity level of "warn" to enforce exhaustive dependencies in React hooks.

### DIFF
--- a/app/client/.eslintrc.base.json
+++ b/app/client/.eslintrc.base.json
@@ -13,7 +13,6 @@
     "react",
     "@typescript-eslint",
     "prettier",
-    "react-hooks",
     "sort-destructure-keys",
     "cypress",
     "testing-library",
@@ -26,7 +25,8 @@
     "plugin:testing-library/react",
     // Note: Please keep this as the last config to make sure that this (and by extension our .prettierrc file) overrides all configuration above it
     // https://www.npmjs.com/package/eslint-plugin-prettier#recommended-configuration
-    "plugin:prettier/recommended"
+    "plugin:prettier/recommended",
+    "plugin:react-hooks/recommended"
   ],
   "parserOptions": {
     "ecmaVersion": 2020, // Allows for the parsing of modern ECMAScript features
@@ -54,8 +54,6 @@
     "jest/no-disabled-tests": "error",
     // enforce `import type` for all type-only imports so the bundler knows to erase them
     "@typescript-eslint/consistent-type-imports": "error",
-    "react-hooks/rules-of-hooks": "error",
-    "react-hooks/exhaustive-deps": "warn",
     "react/jsx-boolean-value": "error",
     "react/self-closing-comp": "error",
     "react/jsx-sort-props": "error",
--- a/app/client/.eslintrc.base.json
+++ b/app/client/.eslintrc.base.json
@@ -23,10 +23,10 @@
     "plugin:@typescript-eslint/recommended",
     "plugin:cypress/recommended",
     "plugin:testing-library/react",
+    "plugin:react-hooks/recommended",
     // Note: Please keep this as the last config to make sure that this (and by extension our .prettierrc file) overrides all configuration above it
     // https://www.npmjs.com/package/eslint-plugin-prettier#recommended-configuration
-    "plugin:prettier/recommended",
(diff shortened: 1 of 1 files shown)


app/client/.eslintrc.base.json
```diff
@@ -55,6 +55,7 @@
     // enforce `import type` for all type-only imports so the bundler knows to erase them
     "@typescript-eslint/consistent-type-imports": "error",
     "react-hooks/rules-of-hooks": "error",
+    "react-hooks/exhaustive-deps": "warn",
```
KelvinOm, line 58: @alex-golovanov We are using the react-hook plugin, and [this rule has already been configured](https://github.com/facebook/react/blob/main/packages/eslint-plugin-react-hooks/src/index.js#L18) as you requested. Is there a problem with it, or something else? Why have these changes been made?

KelvinOm (changes requested):

app/client/.eslintrc.base.json
alex-golovanov: > @alex-golovanov We are using the react-hook plugin, and [this rule has already been configured](https://github.com/facebook/react/blob/main/packages/eslint-plugin-react-hooks/src/index.js#L18) as you requested. Is there a problem with it, or something else? Why have these changes been made?

Yes, the `react-hooks` plugin is, in fact, being used, however this particular rule is not turned on.

alex-golovanov: @KelvinOm just to clarify, the rule that you are pointing at is in `recommended` preset, which is not being used in our config.

KelvinOm: So let's add `"plugin:react-hooks/recommended"` to extends?

alex-golovanov: > So let's add `"plugin:react-hooks/recommended"` to extends?

Sure, using preset may be better for future proofing, though at this point it's highly unlikely to change. Still, I'll oblige.

app/client/.eslintrc.base.json
```diff
@@ -26,7 +25,8 @@
     "plugin:testing-library/react",
     // Note: Please keep this as the last config to make sure that this (and by extension our .prettierrc file) overrides all configuration above it
     // https://www.npmjs.com/package/eslint-plugin-prettier#recommended-configuration
-    "plugin:prettier/recommended"
+    "plugin:prettier/recommended",
+    "plugin:react-hooks/recommended"
```
KelvinOm, line 29: `"plugin:prettier/recommended"` should always be the last line, check out the comment above.

KelvinOm (changes requested):

app/client/.eslintrc.base.json
alex-golovanov: Tbh, I don't think it should be there at all, but that's a topic for another discussion :)

KelvinOm (approved):

PR #34354, opened 2024-06-20
feat: Add user context to permissions for repository methods

abhvsn:
## Description
We are using non-reactive driver (JDBC) for Postgres which makes the DB calls synchronous in nature. This is forcing us to add blocking calls to fetch the permission groups for the user. As a part of this call we are fetching the current user from the ReactiveSecurityContext, but with blocking call this results in null value. To fix this we are going ahead with adding the user object in the AclPermission in the reactive chain itself to avoid the dependency of ReactiveSecurityContext in the synchronous call.

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/BaseAppsmithRepositoryCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/BaseAppsmithRepositoryCEImpl.java
@@ -207,21 +207,47 @@ public abstract class BaseAppsmithRepositoryCEImpl<T extends BaseDomain> impleme
 
     protected Set<String> getCurrentUserPermissionGroupsIfRequired(
             AclPermission permission, boolean includeAnonymousUserPermissions) {
-        return permission == null ? Set.of() : getCurrentUserPermissionGroups(includeAnonymousUserPermissions);
+        return permission == null ? Set.of() : getPermissionGroupsForUser(includeAnonymousUserPermissions, permission.getUser());
     }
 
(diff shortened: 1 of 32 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/BaseAppsmithRepositoryCEImpl.java
```diff
@@ -207,21 +207,47 @@ protected Set<String> getCurrentUserPermissionGroupsIfRequired(AclPermission per
...
-                .map(ctx -> ctx.getAuthentication().getPrincipal())
-                .map(principal -> includeAnonymousUserPermissions
-                        ? getAllPermissionGroupsForUser((User) principal)
-                        : getStrictPermissionGroupsForUser((User) principal))
-                .block();
-        return permissionGroups == null ? Collections.emptySet() : permissionGroups;
+    protected Set<String> getPermissionGroupsForUser(boolean includeAnonymousUserPermissions, User user) {
+        if (user == null) {
+            return Mono.just(Collections.emptySet());
+        }
```
sharat87, line 220: Doesn't this produce a compile error?

sharat87 (changes requested):

abhvsn:
Closing this as we moved to adding the context via QueryAllParams instead of AclPermissions and the [change](https://github.com/appsmithorg/appsmith/pull/33693) is already merged.

Issue #33572, opened 2024-05-20
[Task]: Remove uses of the incorrect `@Transient` annotation


sharat87:
### SubTasks

- [ ] Move all `org.springframework.data.annotation.Transient` uses to `jakarta.persistence.Transient`.
- [ ] Uncomment the check for this in `CustomJsonType` class.

PR #34404, opened 2024-06-23
feat: Updated the @Transient from the jakarta dependency.

Naveen-Goud:
**Description**

This is the PR for the [TASK](#33572 )

**Changes in PR:**
- Added the Jakarta dependency in pom.xml of appsmith-server and appsmith-interfaces.
- Replaced the @Trasient from the jakarta by removing from springframework.

Difference between **org.springframework.data.annotation.Transient** and **jakarta.persistence.Transient**
@org.springframework.data.annotation.Transient specifically states to the spring framework that the Object Mapper you are using should not include this value when converting from Java Object to JSON.

Also note if you attempted to persist the above entity into the database, it would still save it to the database because @org.springframework.data.annotation.Transient only applys to Object mapping frameworks not JPA.

**fixes**- #33572 

Please review this PR- @sharat87 

## Summary by CodeRabbit

- **Chores**
  - Updated dependencies to use `jakarta.persistence-api` version `3.2.0`.
  - Refactored annotations from `org.springframework.data.annotation.Transient` to `jakarta.persistence.Transient` for multiple models.

### DIFF
--- a/app/server/appsmith-interfaces/pom.xml
+++ b/app/server/appsmith-interfaces/pom.xml
@@ -250,7 +250,12 @@
             <groupId>io.micrometer</groupId>
             <artifactId>micrometer-tracing</artifactId>
         </dependency>
-
+        <!-- https://mvnrepository.com/artifact/jakarta.persistence/jakarta.persistence-api -->
+        <dependency>
+            <groupId>jakarta.persistence</groupId>
+            <artifactId>jakarta.persistence-api</artifactId>
+            <version>3.2.0</version>
+        </dependency>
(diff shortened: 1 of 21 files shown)


sharat87:
Hey @Naveen-Goud, thank you for your work and for opening a PR. However, that task is not intended for the `release` branch. It is for a separate project, that's still in its own PR, and hasn't been merged to `release`. We don't want this change on the `release` branch as it is today.

On that note, I'll be closing this PR. Please pick tasks that have the "Good first issue" label on them to avoid this in the future. Thank you!

PR #34409, opened 2024-06-24
Improved signup and login pages performance

### LINKED ISSUE #15744: [Task]: Make signup and login pages faster

### SubTasks

We load the main bundle for login and signup pages.

1. The bundle is huge, and we don't need all that code for this page
2. It takes the browser about 1-3 seconds to process and understand the script.

Fixing these will help us shave about 2-3 seconds from the load time.
While doing these, ensure that our split code is preloaded so that they have it ready when they finally login.

AnnaHariprasad5123:
Fixes #15744

**What's in this PR?**

- Added code splitting and fallback mechanism for auth pages.
- Integrated Webpack plugins to improve auth pages load speed.
- Added preconnect links for external resources.

**Screenshots :** 

**Earlier :** 
image

**Now :** 
image

Thank you 

## Summary by CodeRabbit

- **New Features**
  - Improved performance by converting user authentication components to lazy loading.

- **Optimizations**
  - Enhanced CSS and JavaScript optimizations through new webpack plugins.
  - Increased memory allocation for build scripts to improve build stability and performance.
  - Added preconnect links for faster resource fetching.

- **Dependencies**
  - Updated and added several dependencies for better optimization and performance.

### DIFF
--- a/app/client/craco.common.config.js
+++ b/app/client/craco.common.config.js
@@ -2,6 +2,11 @@ const CracoAlias = require("craco-alias");
 const CracoBabelLoader = require("craco-babel-loader");
 const path = require("path");
 const webpack = require("webpack");
+const CrittersPlugin = require("critters-webpack-plugin");
+const CssMinimizerPlugin = require("css-minimizer-webpack-plugin");
+const TerserPlugin = require("terser-webpack-plugin");
+const PurgeCSSPlugin = require("purgecss-webpack-plugin").PurgeCSSPlugin;
+const glob = require("glob");
 
 module.exports = {
   devServer: {
@@ -73,6 +78,23 @@ module.exports = {
         ],
       },
       optimization: {
+        minimize: true,
+        minimizer: [
+          new CssMinimizerPlugin(),
+          new TerserPlugin({
+            terserOptions: {
+              compress: {
+                unused: true,
+                dead_code: true,
+                warnings: false,
+              },
+              output: {
+                comments: false,
+              },
+            },
+            extractComments: false,
+          }),
+        ],
         splitChunks: {
           cacheGroups: {
             icons: {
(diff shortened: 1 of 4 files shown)


app/client/craco.common.config.js
```diff
@@ -73,6 +78,23 @@ module.exports = {
         ],
       },
       optimization: {
+        minimize: true,
```
KelvinOm, line 81: As far as I know, CRACO config already has appropriate settings. Do we really need to redefine them? Also, for dev mode we do not need all this optimizations. So we need to add a check for it.

cc @alex-golovanov

KelvinOm (changes requested):

app/client/craco.common.config.js
alex-golovanov: Yes, indeed, craco has this optimisation. Moreover, we only want to enable this in production mode (currently it is set up this way).

I should mention that currently there is an active effort to move from craco to webpack, so these settings will be transparent in the future.

app/client/package.json
```diff
@@ -109,13 +108,14 @@
...
     "craco-babel-loader": "^1.0.4",
+    "critters-webpack-plugin": "^3.0.2",
     "cssnano": "^6.0.1",
     "cypress-log-to-output": "^1.1.2",
     "d3-geo": "^3.1.0",
     "dayjs": "^1.10.6",
     "deep-diff": "^1.0.2",
-    "design-system": "npm:@appsmithorg/design-system@2.1.41",
-    "design-system-old": "npm:@appsmithorg/design-system-old@1.1.17",
+    "design-system": "npm:@appsmithorg/design-system@2.1.39",
```
alex-golovanov, line 117: These downgrades are probably unintentional, doesn't look that we should downgrade these packages.

app/client/public/index.html
```diff
@@ -7,6 +7,8 @@
       content="width=device-width, initial-scale=1, shrink-to-fit=no"
     />
     <title>Appsmith</title>
+    <link rel="preconnect" href="https://www.google.com" />
+    <link rel="preconnect" href="https://assets.appsmith.com" />
```
alex-golovanov, line 11: I'm not familiar with this URL (`assets.appsmith.com`), what is it for? It appears that it may change depending on current location/environment of the FE app.

AnnaHariprasad5123: In Lighthouse metrics, LCP suggested these changes to reduce the load time. So I defined them. After that, I didn't get them in LCP.

PR #34548, opened 2024-06-27
feat: snowflake key auth server

sneha122:
## Description
This PR adds new authentication method for snowflake: Use of public and private keys to authorise a snowflake database. 

Snowflake provides a couple of authentication mechanisms in order to authenticate the DB. Currently appsmith only provides a way to authorise using username and password. This PR adds support for private key authentication, where by user can set a public key on their DB and they can use corresponding private key in appsmith to authorise the datasource.

In snowflake DB form, we have added a new field for authentication type which has two options: Basic and Key pair. Basic will allows us to authenticate using username and password. Key pair will provide us a way to upload the private key along with input field for adding passphrase (In case of encrypted private key).

More info: https://docs.snowflake.com/en/user-guide/key-pair-auth

https://github.com/appsmithorg/appsmith/assets/30018882/99774925-12a3-4cc0-af0a-614c3574cdc3

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9743830603>
> Commit: cb64fffc9e181558525e020a4597b616eeacea7a
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9743830603&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Datasource`

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
  - Enhanced Snowflake integration with improved connection creation and authentication handling (key pair and basic auth types).

- **Bug Fixes**
  - Fixed issues related to handling authentication objects within datasource configuration to ensure more reliable connections.

- **Documentation**
  - Added new error messages for passphrase requirements and incorrect passphrase or private key for encrypted private keys.

- **Tests**
  - Updated tests to verify the correct setting of authentication types in the Snowflake plugin.


app/server/appsmith-plugins/snowflakePlugin/src/main/java/com/external/plugins/exceptions/SnowflakeErrorMessages.java
```diff
@@ -36,4 +36,8 @@ public class SnowflakeErrorMessages extends BasePluginErrorMessages {
     public static final String DS_MISSING_PASSWORD_ERROR_MSG = "Missing password for authentication.";
 
     public static final String DS_MISSING_PRIVATE_KEY_ERROR_MSG = "Missing private key for authentication.";
+
+    public static final String PASSPHRASE_IS_REQUIRED_FOR_ENCRYPTED_PRIVATE_KEY =
```
NilanshBansal, line 40: @sneha122 Nit: Can we please follow the same nomenclature as the other error messages above.
Maybe `DS_MISSING_PASSPHRASE_ENCRYPTED_PRIVATE_KEY` and `DS_INCORRECT_PRIVATE_KEY_OR_PASSPHRASE`

app/server/appsmith-plugins/snowflakePlugin/src/main/java/com/external/plugins/SnowflakePlugin.java
```diff
@@ -491,5 +459,108 @@ public Mono<DatasourceStructure> getStructure(
                     })
                     .subscribeOn(scheduler);
         }
+
+        private Mono<HikariConfig> getHikariConfig(
+                DatasourceConfiguration datasourceConfiguration, Properties properties) {
+            HikariConfig commonConfig = getCommonHikariConfig(properties);
+            Mono<HikariConfig> configMono = Mono.just(commonConfig);
```
NilanshBansal, line 466: @sneha122 any particular use case to convert commonConfig to Mono?

app/server/appsmith-plugins/snowflakePlugin/src/main/java/com/external/plugins/SnowflakePlugin.java
```diff
@@ -491,5 +459,108 @@ public Mono<DatasourceStructure> getStructure(
...
+
+            String authenticationType =
+                    datasourceConfiguration.getAuthentication().getAuthenticationType();
+            if (authenticationType != null) {
+                switch (authenticationType) {
+                    case BASIC:
+                        configMono = configMono.flatMap(config -> getBasicAuthConfig(config, datasourceConfiguration));
```
NilanshBansal, line 473: @sneha122 Here, we are again doing a flatMap on the converted Mono to get the config value again. Can we just not keep the commonConfig as it is and pass it to getBasicAuthConfig() function?

app/server/appsmith-plugins/snowflakePlugin/src/main/java/com/external/plugins/SnowflakePlugin.java
```diff
@@ -159,58 +157,31 @@ public Mono<ActionExecutionResult> execute(
...
-                        // Set up the connection URL
-                        StringBuilder urlBuilder = new StringBuilder(
-                                "jdbc:snowflake://" + datasourceConfiguration.getUrl() + ".snowflakecomputing.com?");
-                        config.setJdbcUrl(urlBuilder.toString());
+            return getHikariConfig(datasourceConfiguration, properties)
+                    .flatMap(config -> Mono.fromCallable(() -> {
+                                // Set up the connection URL
+                                StringBuilder urlBuilder = new StringBuilder("jdbc:snowflake://"
+                                        + datasourceConfiguration.getUrl() + ".snowflakecomputing.com?");
+                                config.setJdbcUrl(urlBuilder.toString());
```
NilanshBansal, line 165: @sneha122 Nit: L163-165 are redundant as it is already being set from the getKeyPairAuthConfig(). 
Ref: https://github.com/appsmithorg/appsmith/blob/42675823c4b1a863cfcc494fe562a58a8bc4fa4e/app/server/appsmith-plugins/snowflakePlugin/src/main/java/com/external/plugins/SnowflakePlugin.java#L493-L494

For Basic Auth also, this can be set in the function getBasicAuthConfig() or maybe removed from the getKeyPairAuthConfig() function as well to remove the duplication.

NilanshBansal (changes requested):
Also, can we please add some tests for the following cases:
- when passphrase is wrong for encrypted private key. 
- when private key or passphrase is wrong
- positive successful test case with all the fields and no errors.

app/server/appsmith-plugins/snowflakePlugin/src/main/java/com/external/plugins/SnowflakePlugin.java
sneha122: No reason actually, let me revert this back, thanks for pointing out

app/server/appsmith-plugins/snowflakePlugin/src/main/java/com/external/plugins/SnowflakePlugin.java
sneha122: Yes I will do that

app/server/appsmith-plugins/snowflakePlugin/src/main/java/com/external/plugins/SnowflakePlugin.java
sneha122: @NilanshBansal So in basic auth, we need to set this url on config object where as in key pair auth, we need to set jdbc url on the datasource object. Let me refactor this code a bit to make it more clear.

sneha122:
> Also, can we please add some tests for the following cases:
> 
> * when passphrase is wrong for encrypted private key.
> * when private key or passphrase is wrong
> * positive successful test case with all the fields and no errors.

@NilanshBansal Will take this up separately as it may not be as straightforward as it looks.

app/server/appsmith-plugins/snowflakePlugin/src/main/java/com/external/plugins/SnowflakePlugin.java
```diff
@@ -37,8 +37,7 @@
 import java.sql.Connection;
 import java.util.*;
 
-import static com.appsmith.external.constants.Authentication.BASIC;
-import static com.appsmith.external.constants.Authentication.SNOWFLAKE_KEY_PAIR_AUTH;
+import static com.appsmith.external.constants.Authentication.*;
```
NilanshBansal, line 40: @sneha122 Nit: Can we please add specific imports instead of this wildcard import. The Authentication.java has a lot of static constants which are getting imported here and we are not utilising them.

app/server/appsmith-plugins/snowflakePlugin/src/main/java/com/external/plugins/SnowflakePlugin.java
```diff
@@ -491,5 +459,112 @@ public Mono<DatasourceStructure> getStructure(
...
+        private Mono<HikariConfig> getHikariConfig(
+                DatasourceConfiguration datasourceConfiguration, Properties properties) {
+            HikariConfig commonConfig = getCommonHikariConfig(properties);
+            Mono<HikariConfig> configMono = Mono.empty();
+
+            String authenticationType =
+                    datasourceConfiguration.getAuthentication().getAuthenticationType();
+            if (authenticationType != null) {
+                switch (authenticationType) {
+                    case DB_AUTH:
```
NilanshBansal, line 472: Is the authentication type DB_AUTH or BASIC, can you please confirm once.

NilanshBansal (commented):
Minor qs, rest lgtm

app/server/appsmith-plugins/snowflakePlugin/src/main/java/com/external/plugins/SnowflakePlugin.java
sneha122: @NilanshBansal Checked this it is getting mapped to dbauth, hence updated the code as well

app/server/appsmith-plugins/snowflakePlugin/src/main/java/com/external/plugins/SnowflakePlugin.java
sneha122: Yes will update this

NilanshBansal (approved):
Changes LGTM. Also tested on DP

PR #34566, opened 2024-06-27
chore: Change `policies` set to `policyMap`

sharat87:
Migrate `policies` from being a `Set<Policy>` to a `policyMap`, that's a `Map<String, Policy>`. The key defines the permission whereas the value is the actual policy object. This makes it easier to use this data in code, to query as a MongoDB nested document, and as a Postgres `jsonb` column. Win-win!

The migration runs for all collections in parallel, with following stats taken on 1-Jul-2024.

1. Release data, only non-deleted documents, all collections, parallel, 2m 42s.
2. Prod replica data, only non-deleted documents, all collections, parallel, (15m).
3. Release data, only non-deleted documents, all collections, serial, ~9m.

This PR is only part of the solution. We have backwards compatible `getPolicies()` and `setPolicies()` methods so the diff isn't too big, but a follow-up PR will remove those two, and migration all usages to use the map instead.

This PR only (almost) strictly includes changes that were necessary for the build and tests to pass.

**/test all**

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/10301601562>
> Commit: d58ac59ea9bffb170f979176cdfc1ff378b440ef
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=10301601562&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Thu, 08 Aug 2024 12:46:36 UTC

## Summary by CodeRabbit

- **New Features**
  - Introduced migration logic to convert `policies` to `policyMap`, enhancing data structure for policy management.

- **Documentation**
  - Updated summaries and documentation to reflect changes in policy handling.

- **Refactor**
  - Replaced direct field references with constant values for policy fields in migration files.
  - Adjusted method signatures to utilize the new `policyMap` structure.
  - Improved policy management logic for better immutability and safety.

- **Tests**
  - Added new test cases for policy management functionality.
  - Updated existing test assertions to ensure proper policy handling with the new data structure.


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/ce/ActionCE_DTO.java
```diff
@@ -169,6 +171,27 @@ public class ActionCE_DTO implements Identifiable, Executable {
...
+
+    @Deprecated(forRemoval = true, since = "Use policyMap instead")
+    public void setPolicies(Set<Policy> policies) {
+        if (policies == null) {
+            policyMap = null;
+        } else {
+            policyMap.clear();
+            for (Policy policy : policies) {
+                policyMap.put(policy.getPermission(), policy);
+            }
```
nsarupr, line 191: Possible NPE.
Can you please use the similar code used above? Or make this an empty map when a null value is being set.

app/server/appsmith-server/src/main/java/com/appsmith/server/dtos/PageDTO.java
```diff
@@ -81,6 +82,27 @@ public class PageDTO {
...
+    public void setPolicies(Set<Policy> policies) {
+        if (policies == null) {
+            policyMap = null;
+        } else {
+            policyMap.clear();
+            for (Policy policy : policies) {
+                policyMap.put(policy.getPermission(), policy);
+            }
+        }
+    }
```
nsarupr, line 104: Similar comment as above.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
```diff
@@ -243,8 +243,12 @@ public void addInstanceConfigurationPlaceHolder(MongoTemplate mongoTemplate) {
                 Set.of(new Permission(savedInstanceConfig.getId(), MANAGE_INSTANCE_CONFIGURATION)));
 
         Query adminUserQuery = new Query();
-        adminUserQuery.addCriteria(
-                where(BaseDomain.Fields.policies).elemMatch(where("permission").is(MANAGE_INSTANCE_ENV.getValue())));
+        // Oring for backward compatibility where policies are stored in a set instead of map
```
nsarupr, line 246: nit: 
```suggestion
        // OR-ing for backward compatibility where policies are stored in a set instead of map
```

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
```diff
@@ -222,7 +222,7 @@ public void addAnonymousUser(MongoTemplate mongoTemplate) {
         }
     }
 
-    @ChangeSet(order = "029", id = "add-instance-config-object", author = "")
+    @ChangeSet(order = "029", id = "add-instance-config-object", author = "", runAlways = true)
```
nsarupr, line 225: Are there any side-effects to running this migration every time?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
```diff
@@ -243,8 +243,12 @@ public void addInstanceConfigurationPlaceHolder(MongoTemplate mongoTemplate) {
                 Set.of(new Permission(savedInstanceConfig.getId(), MANAGE_INSTANCE_CONFIGURATION)));
 
         Query adminUserQuery = new Query();
-        adminUserQuery.addCriteria(
-                where(BaseDomain.Fields.policies).elemMatch(where("permission").is(MANAGE_INSTANCE_ENV.getValue())));
+        // Oring for backward compatibility where policies are stored in a set instead of map
+        Criteria policyCriteria =
+                where("policies").elemMatch(where("permission").is(MANAGE_INSTANCE_ENV.getValue()));
```
nsarupr, line 248: Let's define a constant value for this as this is being used in multiple places.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration028TagUserManagementRolesWithoutDefaultDomainTypeAndId.java
```diff
@@ -40,7 +40,7 @@ public void tagUserManagementRolesWithoutDefaultDomainTypeAndId() {
                 .is(RESET_PASSWORD_USERS.getValue())
                 .andOperator(notDeleted());
         Query queryExistingUsersWithResetPasswordPolicy = new Query(resetPasswordPolicyExistsAndNotDeleted);
-        queryExistingUsersWithResetPasswordPolicy.fields().include(User.Fields.policies);
+        queryExistingUsersWithResetPasswordPolicy.fields().include("policies");
```
nsarupr, line 43: Let's use constant.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration057PolicySetToPolicyMap.java
nsarupr, line 1: Let's add comments to make this file readable for future visits.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/BaseAppsmithRepositoryCEImpl.java
```diff
@@ -105,11 +105,8 @@ public static Criteria userAcl(Set<String> permissionGroups, AclPermission permi
             return null;
         }
         // Check if the permission is being provided by any of the permission groups
-        return Criteria.where(BaseDomain.Fields.policies)
-                .elemMatch(Criteria.where("permissionGroups")
-                        .in(permissionGroups)
-                        .and("permission")
-                        .is(permission.getValue()));
+        return Criteria.where(BaseDomain.Fields.policyMap + "." + permission.getValue() + ".permissionGroups")
```
nsarupr, line 108: There is a method called `dotted(String... args)`, which can help you get the string which you need here.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/PolicySolutionCEImpl.java
```diff
@@ -68,8 +68,10 @@ public <T extends BaseDomain> T addPoliciesToExistingObject(Map<String, Policy>
             policyMap1.put(entry.getKey(), policy);
         }
 
+        final Set<Policy> policies = new HashSet<>(obj.getPolicies());
```
nsarupr, line 71: Since the methods `getPolicies()` and `setPolicies(Set<Policy> policies)` are already marked for deprecation, should we not use them over here?
Or if we want to update this method at a later stage, lets add test cases for now, so that it is easier when we want to update this code here.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/NewPageServiceTest.java
```diff
@@ -291,9 +291,8 @@ public void publishPages_WhenUserDoesNotHavePermissionOnPages_NotPublished() {
                             .get();
 
                     firstPage
-                            .getPolicies()
-                            .forEach(policy -> policy.setPermission(
-                                    pagePermission.getReadPermission().getValue()));
+                            .getPolicyMap()
+                            .remove(pagePermission.getEditPermission().getValue());
```
nsarupr, line 295: Let's not make changes to the test case here. 
Can we instead still use the ReadPermission?

nsarupr (changes requested):
Code change requests added

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/BaseDomain.java
```diff
@@ -98,6 +100,34 @@ public boolean isDeleted() {
...
+    public Set<Policy> getPolicies() {
+        return policyMap == null ? null : Set.copyOf(policyMap.values());
+    }
+
+    // TODO Abhijeet: Remove this method once we have migrated all the usages of policies to policyMap
+    @JsonView({Views.Internal.class})
+    @Deprecated(forRemoval = true, since = "Use policyMap instead")
+    public void setPolicies(Set<Policy> policies) {
+        if (policies == null) {
+            policyMap = null;
```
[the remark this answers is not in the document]
abhvsn: This null check is for the `policies` the else null check is happening on `policyMap` where we want to set the value. We are keeping the null value as is just to replicate the behaviour what we have today.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
abhvsn: We are expecting the policies field will be removed from the code-base do we still want to add the constant, for the field which is removed from the codebase?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration028TagUserManagementRolesWithoutDefaultDomainTypeAndId.java
abhvsn: Same as above.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/DatabaseChangelog2.java
abhvsn: I'm going with creating an interface with deprecated field constants, so that future ourselves will think twice before making any change.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration028TagUserManagementRolesWithoutDefaultDomainTypeAndId.java
```diff
@@ -40,7 +41,7 @@ public void tagUserManagementRolesWithoutDefaultDomainTypeAndId() {
                 .is(RESET_PASSWORD_USERS.getValue())
                 .andOperator(notDeleted());
         Query queryExistingUsersWithResetPasswordPolicy = new Query(resetPasswordPolicyExistsAndNotDeleted);
-        queryExistingUsersWithResetPasswordPolicy.fields().include(User.Fields.policies);
+        queryExistingUsersWithResetPasswordPolicy.fields().include(POLICIES);
```
sharat87, line 44: We're calling `.getPolicies` on the objects from this query a little further down. That means if the document in DB has `policyMap` instead of `policies`, it won't get the data it should, right? Should we also `.include` the `policyMap` field in this query here?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration029PopulateDefaultDomainIdInUserManagementRoles.java
```diff
@@ -45,7 +46,7 @@ public void populateDefaultDomainIdInUserManagementRoles() {
                 .is(RESET_PASSWORD_USERS.getValue())
                 .andOperator(notDeleted());
         Query queryExistingUsersWithResetPasswordPolicy = new Query(resetPasswordPolicyExistsAndNotDeleted);
-        queryExistingUsersWithResetPasswordPolicy.fields().include(User.Fields.policies);
+        queryExistingUsersWithResetPasswordPolicy.fields().include(POLICIES);
```
sharat87, line 49: Same as above. Should we `.include` the `policyMap` field as well?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration030TagUsersWithNoUserManagementRoles.java
```diff
@@ -48,10 +50,8 @@ public void tagUsersWithNoUserManagementRoles() {
 
         Criteria criteriaUsersWithNoUserManagementRoles =
                 Criteria.where(BaseDomain.Fields.id).nin(userIdsWithUserManagementRoles);
-        Criteria criteriaUsersPoliciesExists =
-                Criteria.where(BaseDomain.Fields.policies).exists(true);
-        Criteria criteriaUsersPoliciesNotEmpty =
-                Criteria.where(BaseDomain.Fields.policies).not().size(0);
+        Criteria criteriaUsersPoliciesExists = Criteria.where(POLICIES).exists(true);
+        Criteria criteriaUsersPoliciesNotEmpty = Criteria.where(POLICIES).not().size(0);
```
sharat87, line 54: Just checking, don't we need the `policyMap` criteria here?

sharat87 (commented):
Just checking if we've reviewed the migrations that touch `User` objects, and if they need any changes.

(I'm not allowed to approve this PR since I'm the author).

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration028TagUserManagementRolesWithoutDefaultDomainTypeAndId.java
abhvsn: @sharat87 yeah sounds good! 
Another query considering the migration for moving from `policies` to `policyMap` will still be pending because of the ordering, we will end up in a state with null policies as getter will look for the policyMap which is still not present yet.

sharat87: Not sure I understand. Are we talking about the query in the migration... or are you asking a question? 🤔 😬

abhvsn: I was asking the question, what happens when the migration for policies to policyMap is still pending, but we are upgrading from the version where both `Migration028TagUserManagementRolesWithoutDefaultDomainTypeAndId.java`	 and `Migration057PolicySetToPolicyMap.java` are pending.

sharat87: Oh yeah, excellent question. That means we can't use the `User` class at all here, right? And operate on `Document` or `JSONNode` objects directly? 😞

abhvsn: @sharat87 this means wherever we are using getPolicies in migrations needs to be replaced with `JSONNode`, will this make the migrations brittle. I think this is the reason we never used to remove the field completely instead just deprecate it.

sharat87: But does keeping the field and deprecating it solve it? If it does, I'm happy to go that route. Let's try it out thin.

abhvsn: @sharat87 This will mean following updates:
1. Instead of removing policies add deprecation warning
2. Don't introduce backward compatibility methods getPolicies and setPolicies
3. Replace `getPolicies -> getPolicyMap` and `setPolicies -> setPolicyMap` except in migrations, which is what we planned to do once this PR is merged

I know this will make the PR diffs larger but I'm still inclined to this than removing the field completely. Thoughts!

sharat87: Oh we can still have the fallbacks, no? Like, in the `getPolicies()`, we check if `this.policies` is `null`. If it is, then we look into `policyMap` as we are right now. In `setPolicies()`, we _only_ update `this.policyMap` and explicitly set `this.policies` to `null`. Nothing much else would have to change, no?

abhvsn: Sounds good 👍

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration057PolicySetToPolicyMap.java
```diff
@@ -0,0 +1,121 @@
...
+
+    private static final Set<String> EE_COLLECTION_NAMES = Set.of(
+            // Empty here
+            );
+
+    @RollbackExecution
+    public void rollbackExecution() {}
+
+    @Execution
+    public void execute() {
```
nsarupr, line 69: In case of a failure, the migration isn't idempotent.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration057PolicySetToPolicyMap.java
```diff
@@ -0,0 +1,121 @@
...
+                        }
+                    } else {
+                        log.error(errorPrefix, error);
+                    }
+                    return Mono.error(error);
+                })
+                .block();
+    }
+
+    private static Mono<Void> executeForCollection(ReactiveMongoTemplate mongoTemplate, String collectionName) {
```
nsarupr, line 92: Can you please add comments to the migration to make this more explicable?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration057PolicySetToPolicyMap.java
```diff
@@ -0,0 +1,121 @@
...
+        final ArrayOperators.ArrayToObject operator =
+                ArrayOperators.ArrayToObject.arrayValueOfToObject(VariableOperators.Map.itemsOf(POLICIES)
+                        .as("this")
+                        .andApply(agg -> new Document("k", "$$this.permission").append("v", "$$this")));
+
+        final Mono<UpdateResult> convertToMap = mongoTemplate.updateMulti(
+                // Migrate the policies field to the policyMap field only if the policies field exists and is not empty
+                new Query(where(dotted(POLICIES, "0"))
+                        .exists(true)
+                        .and(BaseDomain.Fields.deletedAt)
```
nsarupr, line 104: Use the isDeleted() criteria method.
Also, can you create a criteria and query separately to make the code more readable.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration057PolicySetToPolicyMap.java
```diff
@@ -0,0 +1,121 @@
...
+                .block();
+    }
+
+    private static Mono<Void> executeForCollection(ReactiveMongoTemplate mongoTemplate, String collectionName) {
+        log.info("Migrating policies to policyMap in {}", collectionName);
+
+        final ArrayOperators.ArrayToObject operator =
+                ArrayOperators.ArrayToObject.arrayValueOfToObject(VariableOperators.Map.itemsOf(POLICIES)
+                        .as("this")
+                        .andApply(agg -> new Document("k", "$$this.permission").append("v", "$$this")));
```
nsarupr, line 98: Explanation needed.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/BaseAppsmithRepositoryCEImpl.java
```diff
@@ -105,11 +106,8 @@ public static Criteria userAcl(Set<String> permissionGroups, AclPermission permi
             return null;
         }
         // Check if the permission is being provided by any of the permission groups
-        return Criteria.where(BaseDomain.Fields.policies)
-                .elemMatch(Criteria.where("permissionGroups")
-                        .in(permissionGroups)
-                        .and("permission")
-                        .is(permission.getValue()));
+        return Criteria.where(dotted(BaseDomain.Fields.policyMap, permission.getValue(), "permissionGroups"))
```
nsarupr, line 109: Let's use the constant values and not the Field values.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/PolicySolutionCEImpl.java
```diff
@@ -59,17 +59,21 @@ public <T extends BaseDomain> T addPoliciesToExistingObject(Map<String, Policy>
         // TODO: Investigate a solution without using deep-copy.
         // TODO: Do we need to return the domain object?
```
nsarupr, line 60: Can you also take a look at the TODOs?

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/PolicySolutionCEImpl.java
```diff
@@ -59,17 +59,21 @@ public <T extends BaseDomain> T addPoliciesToExistingObject(Map<String, Policy>
...
         final Map<String, Policy> policyMap1 = new HashMap<>();
-        for (Map.Entry<String, Policy> entry : policyMap.entrySet()) {
-            Policy entryValue = entry.getValue();
-            Policy policy = Policy.builder()
-                    .permission(entryValue.getPermission())
-                    .permissionGroups(new HashSet<>(entryValue.getPermissionGroups()))
-                    .build();
-            policyMap1.put(entry.getKey(), policy);
+        if (policyMap != null) {
```
nsarupr, line 62: Why add a null check? Are we at any point passing a null value to the `addPoliciesToExistingObject` method? If it is only being used in the test case, we can go ahead and mark the policyMap parameter non-nullable.

app/server/appsmith-server/src/test/java/com/appsmith/server/services/NewPageServiceTest.java
```diff
@@ -287,10 +287,10 @@ public void publishPages_WhenUserDoesNotHavePermissionOnPages_NotPublished() {
...
                             .get();
 
-                    firstPage
-                            .getPolicies()
-                            .forEach(policy -> policy.setPermission(
-                                    pagePermission.getReadPermission().getValue()));
+                    firstPage.setPolicies(firstPage.getPolicies().stream()
+                            .peek(policy -> policy.setPermission(
+                                    pagePermission.getReadPermission().getValue()))
+                            .collect(Collectors.toUnmodifiableSet()));
```
nsarupr, line 293: Can we split it into creating policies and setting them, to increase code readability

nsarupr (changes requested):
Added code changes and code comment requests.

app/server/appsmith-server/src/main/java/com/appsmith/server/constants/DeprecatedFieldName.java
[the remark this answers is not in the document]
abhvsn: So that we don't have to add public static for each constant.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration025RemoveUnassignPermissionFromUnnecessaryRoles.java
```diff
@@ -40,7 +41,7 @@ public void executeMigration() {
 
         Query queryInterestingPermissionGroups = new Query(workspaceDeveloperAndAppViewerRolesCriteria);
         queryInterestingPermissionGroups.fields().include("id");
-        queryInterestingPermissionGroups.fields().include("policies");
+        queryInterestingPermissionGroups.fields().include(POLICIES, PermissionGroup.Fields.policyMap);
```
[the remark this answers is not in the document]
abhvsn: I'm still split on this because in the current PR it was pain to go in each class and look where all we are using the policies as we deliberatly updating the previous migrations. I know this is a anti-pattern and will take your suggestion but still I thought it was a good way to find out all the dependencies across the codebase.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration057PolicySetToPolicyMap.java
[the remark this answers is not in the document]
abhvsn: Thanks for reminding this 🙏🏽. When we wrote that migration focus was on optimising the speed and the PR is also evolved since then, makes sense to split now.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration057PolicySetToPolicyMap.java
abhvsn: When are we going to stop relying on deprecated field 😅. Also in the base domain we use deletedAt to refer deleted documents.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/PolicySolutionCEImpl.java
abhvsn: Oh this was present earlier nothing new added here.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration057PolicySetToPolicyMap.java
abhvsn: It is right? We are updating the objects if: 
- policies exists and is not empty
- not archived
- policyMap does not exists or is empty 

Once the object is updated we expect to have a policyMap object referenced from policies.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/PolicySolutionCEImpl.java
abhvsn: Yeah, it was added as part of tests that we added in this PR. I'm OK with adding a `@NotNull` as we are not expecting the null value.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration025RemoveUnassignPermissionFromUnnecessaryRoles.java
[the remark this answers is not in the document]
abhvsn: Also extending this thought we should rely on constants at migration class not even for other classes like FieldName. Reason being if tomorrow we want to remove the field completely from the code means we need to remove from the constants class as well, where we will end up in updating the migration again.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration053AddApplicationDeletePagePermissionToApplications.java
```diff
@@ -36,7 +35,7 @@ public class Migration053AddApplicationDeletePagePermissionToApplications {
     private static final String APPLICATION_DELETE_PAGES_VALUE = AclPermission.APPLICATION_DELETE_PAGES.getValue();
     private static final String DELETE_APPLICATIONS_VALUE = AclPermission.DELETE_APPLICATIONS.getValue();
 
-    private static final String POLICIES = BaseDomain.Fields.policies;
+    private static final String POLICIES = "policies";
```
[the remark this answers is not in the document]
abhvsn: Keeping this as is, because this is the ideal way to add the constants in migration classes.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration057PolicySetToPolicyMap.java
abhvsn: We are getting following exception after adding the notDeleted() criteria:
```
Caused by: org.springframework.data.mongodb.InvalidMongoDbApiUsageException: Due to limitations of the com.mongodb.BasicDocument, you can't add a second 'null' criteria; Query already contains '{ "$or" : [{ "policyMap" : { "$exists" : false}}, { "policyMap" : {}}]}'

```

Looks like we have to resort to explicit check for `deletedAt: null`.

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/BaseAppsmithRepositoryCEImpl.java
abhvsn: Any reason for this request? If we are not using the Fields constants in the codebase then why did we introduced it in first place? Also just to be on same page we are talking about `policyMap` field.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration025RemoveUnassignPermissionFromUnnecessaryRoles.java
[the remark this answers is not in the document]
nsarupr: Do you think a set of fieldNames at the migration level help, rathen than adding them to classes individually?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration031CreateUserManagementRolesForUsersTaggedIn030.java
```diff
@@ -50,7 +51,7 @@ public void createUserManagementRolesForUsersTaggedInMigration030() {
 
         Query queryUsersTaggedInMigration030 = new Query(criteriaUsersTaggedInMigration030);
         queryUsersTaggedInMigration030.fields().include(User.Fields.id);
-        queryUsersTaggedInMigration030.fields().include(User.Fields.policies);
+        queryUsersTaggedInMigration030.fields().include(POLICIES, "policyMap");
```
nsarupr, line 54: As mentioned in the [comment](https://github.com/appsmithorg/appsmith/pull/34566#discussion_r1690818846) above, should we use the constants at migration level, so that we do not miss on them.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration047AddMissingFieldsInDefaultAppsmithAiDatasource.java
```diff
@@ -123,7 +124,7 @@ public void addMissingFieldsInDefaultAppsmithAiDatasource() {
      */
     private Pair<Boolean, Set<String>> getAllApplicationIdsOfDatasourceAndCheckIfAnyAppPublic(String datasourceId) {
         Query newActionsQuery = new Query().addCriteria(newActionCriteria(datasourceId));
-        newActionsQuery.fields().include(FieldName.ID, FieldName.APPLICATION_ID, NewAction.Fields.policies);
+        newActionsQuery.fields().include(FieldName.ID, FieldName.APPLICATION_ID, POLICIES, "policyMap");
```
nsarupr, line 127: Missed on using the constant value here as well.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration057PolicySetToPolicyMap.java
```diff
@@ -0,0 +1,120 @@
...
+
+        // Migrate the policies field to the policyMap field only if the policies field exists and is not empty
+        // Also policyMap is used as idempotency key to avoid updating same docs in case of failures
+        Criteria validPolicyCriteria = where(dotted(POLICIES, "0")).exists(true);
+        Criteria policyMapNotExists = new Criteria()
+                .orOperator(where(POLICY_MAP).exists(false), where(POLICY_MAP).is(new HashMap<>()));
+        Criteria notDeletedCriteria = where("deletedAt").isNull();
```
nsarupr, line 102: Let us create another method in Migration helper so that it can be reused again.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration057PolicySetToPolicyMap.java
nsarupr: This looks good to me.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration057PolicySetToPolicyMap.java
nsarupr: Does this mean that we will need to update the notDeleted() method, because it will end up giving this error at other places as well?

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/BaseAppsmithRepositoryCEImpl.java
nsarupr: I was talking about the `permissionGroup` field

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/PolicySolutionCEImpl.java
nsarupr: Adding the annotation would make more sense, because otherwise it looks like we are adding an additional test case just to check for the `Null` values which is also just being used inside a test case.

AnaghHegde (approved):

PR #34577, opened 2024-06-28
test: updated js methods to refer ts locators for Git Tests

NandanAnantharamu:
RCA:
Locator was updated in the Ts helper and there was Js helper which was not updated

Solution:
Replaced js locator with ts locator

### DIFF
--- a/app/client/cypress/locators/gitSyncLocators.js
+++ b/app/client/cypress/locators/gitSyncLocators.js
@@ -20,7 +20,7 @@ export default {
   bottomBarCommitButton: ".t--bottom-bar-commit",
   bottomBarMergeButton: ".t--bottom-bar-merge",
   bottomBarPullButton: ".t--bottom-bar-pull",
-  mergeBranchDropdownDestination: ".t--merge-branch-dropdown-destination",
+  mergeBranchDropdownDestination: "[data-testid=t--merge-branch-dropdown-destination]",
   mergeCTA: "[data-testid=t--git-merge-button]",
   loaderQuickGitAction: ".t--loader-quick-git-action",
   connetStatusbar: ".t--connect-statusbar",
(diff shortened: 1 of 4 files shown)


app/client/cypress/locators/gitSyncLocators.js
```diff
@@ -20,7 +20,7 @@ export default {
   bottomBarCommitButton: ".t--bottom-bar-commit",
   bottomBarMergeButton: ".t--bottom-bar-merge",
   bottomBarPullButton: ".t--bottom-bar-pull",
-  mergeBranchDropdownDestination: ".t--merge-branch-dropdown-destination",
+  mergeBranchDropdownDestination: "[data-testid=t--merge-branch-dropdown-destination]",
```
mohanarpit, line 23: Why are we updating the value of the locator here instead of using the variable from the file `GitSync.ts`? Why have the same value in 2 variables in 2 files? This is tech debt that we should address & remove.

NandanAnantharamu: Along with locator we have plan to remove the method as part of js to ts migration, so we will take this up during that activity. This was a quick fix to unblock promotion. Let me know your thoughts @mohanarpit

mohanarpit: Remove which method? When will the JS -> TS migration happen? Also, why isn't the quick fix to simply change the import in the JS test file? Is that a big change to do?

NandanAnantharamu: cy.merge() to be removed. We are planning this migration post flaky fix task. This import thing i can check and see if there are no major changes will go ahead

NandanAnantharamu: done @mohanarpit but the failing test is not related to this locator

nidhi-nair (approved):

mohanarpit (approved):

PR #34588, opened 2024-06-28
chore: Fix datasource creation throwing 400

sharat87:
The recent changes to deny any requests that have unrecognized fields in the body, is making the datasource creation API fail. This is because we are sending extra fields that shouldn't be sent.

[Slack conversation](https://theappsmith.slack.com/archives/CPQNLFHTN/p1719577339838649?thread_ts=1719466294.012589&cid=CPQNLFHTN).

**/test sanity datasource**

> [!WARNING]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9714799545>
> Commit: 85495fbcf6ec641162c7c30ea8f758dcaee380e3
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9714799545&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Sanity, @tag.Datasource`
> It seems like **no tests ran** 😔. We are not able to recognize it, please check <a href="https://github.com/appsmithorg/appsmith/actions/runs/9714799545" target="_blank">workflow here</a>.

## Summary by CodeRabbit

- **New Features**
  - Enhanced data security by adding a method to sanitize and retain only necessary authentication information for various authentication types.

### DIFF
--- a/app/client/src/api/DatasourcesApi.ts
+++ b/app/client/src/api/DatasourcesApi.ts
@@ -76,10 +76,10 @@ class DatasourcesApi extends API {
     }
 
     const clean: any = {
-      authenticationType: authentication.authenticationType,
+      authenticationType: authentication.authenticationType ?? "dbAuth",
     };
 
-    switch (authentication.authenticationType) {
+    switch (clean.authenticationType) {
       case "dbAuth":
         clean.authType = authentication.authType;
         clean.username = authentication.username;
--- a/app/client/src/api/DatasourcesApi.ts
+++ b/app/client/src/api/DatasourcesApi.ts
@@ -50,7 +50,7 @@ class DatasourcesApi extends API {
             datasourceConfiguration: {
               ...storage.datasourceConfiguration,
               isValid: undefined,
-              authentication: this.cleanAuthenticationObject(
+              authentication: DatasourcesApi.cleanAuthenticationObject(
                 storage.datasourceConfiguration.authentication,
               ),
               connection: storage.datasourceConfiguration.connection && {
--- a/app/client/src/api/DatasourcesApi.ts
+++ b/app/client/src/api/DatasourcesApi.ts
@@ -50,6 +50,9 @@ class DatasourcesApi extends API {
             datasourceConfiguration: {
               ...storage.datasourceConfiguration,
               isValid: undefined,
+              authentication: this.cleanupAuthenticationObject(
+                storage.datasourceConfiguration.authentication,
+              ),
               connection: storage.datasourceConfiguration.connection && {
                 ...storage.datasourceConfiguration.connection,
                 ssl: {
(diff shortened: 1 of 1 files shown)


app/client/src/api/DatasourcesApi.ts
```diff
@@ -50,6 +50,9 @@ class DatasourcesApi extends API {
             datasourceConfiguration: {
               ...storage.datasourceConfiguration,
               isValid: undefined,
+              authentication: this.cleanAuthenticationObject(
```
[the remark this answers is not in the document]
sharat87: Done.

app/client/src/api/DatasourcesApi.ts
```diff
@@ -50,6 +50,9 @@ class DatasourcesApi extends API {
             datasourceConfiguration: {
               ...storage.datasourceConfiguration,
               isValid: undefined,
+              authentication: DatasourcesApi.cleanAuthenticationObject(
```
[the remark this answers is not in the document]
sharat87: Invalid.

app/client/src/api/DatasourcesApi.ts
```diff
@@ -67,6 +70,65 @@ class DatasourcesApi extends API {
     return API.post(DatasourcesApi.url, datasourceConfig);
   }
 
+  static cleanAuthenticationObject(authentication: any): any {
```
mohanarpit, line 73: Is the usage of `any` avoidable here? Ideally would like to have as much type safety at compile time as possible.

sharat87: I tried looking into it to setup a type, but, in my opinion, the type definitions to represent this just aren't there today. It's one of the reasons it was so hard to identify what fields are being sent to the server, because more often than not, the type of the request body wasn't an accurate representation of the runtime object fields.

I've raised it in the Slack group I just added you to as well. I'm considering tackling this in small pieces after I'm done with API hardening on the server side.

But for now, I don't have better options here, without blowing this PR up in size that is. Is that okay?

app/client/src/api/DatasourcesApi.ts
```diff
@@ -67,6 +70,65 @@ class DatasourcesApi extends API {
...
+        clean.bearerToken = authentication.bearerToken;
+        break;
+      case "snowflakeKeyPairAuth":
+        clean.username = authentication.username;
+        clean.privateKey = authentication.privateKey;
+        clean.passphrase = authentication.passphrase;
+    }
+
+    return clean;
+  }
```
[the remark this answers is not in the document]
sharat87: I'm not very certain this "improves" code readability, and especially maintainability (refactoring becomes impossible), so I'd rather not do this.

app/client/src/api/DatasourcesApi.ts
mohanarpit: Hmm ... OK. I'm approving this for now. In the future, I think we can potentially copy over the object structures from the server to construct the types where they don't exist on the client.

mohanarpit (approved):

app/client/src/api/DatasourcesApi.ts
sharat87: Oh indeed. I've been thinking about generating an `Api.ts` with methods the same names as backend controller classes, full with type definitions for query params, request body, and response body. Almost like gRPC (I think). We'll get there. Right after I'm done hardening the server a bit more, I'll start biting off at the client. 🙂

PR #34623, opened 2024-07-02
chore: Add `in` clause support for enums datatype for bridge apis

abhvsn:
## Description
PR to improve  enum filtering to ensure better compatibility with Hibernate queries even when using `@Enumerated(String.class)`.

EE corresponding PR: https://github.com/appsmithorg/appsmith-ee/pull/4576

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved enum filtering to ensure better compatibility with Hibernate queries even when using `@Enumerated(String.class)`.


app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ce/bridge/BridgeQuery.java
```diff
@@ -77,6 +77,12 @@ public BridgeQuery<T> in(@NonNull String key, @NonNull Collection<String> value)
         return this;
     }
 
+    // Filtering for enums does not work with hibernate even if the field is annotated with @Enumerated(String.class)
+    public BridgeQuery<T> in(@NonNull String key, @NonNull List<Enum<?>> value) {
```
sharat87, line 81: Can we make this a `Collection` instead of `List`, so it's similar to the other `in` method signatures, and, more importantly, I can pass `Set` objects as well?

sharat87 (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ce/bridge/BridgeQuery.java
abhvsn: @sharat87 I tried that but I'm getting following error `'in(String, Collection<String>)' clashes with 'in(String, Collection<Enum<?>>)'; both methods have same erasure` with the Collection as an arg.
Screenshot 2024-07-02 at 3 52 27 PM

sharat87: Ohh I see. Got it. Good point. Then I'd recommend either changing that existing method's signature from `Collection<String>` to `Collection<Object>` instead. But if that feels to open up too much, my alternative suggestion there would be to use separate method names. Like `.in` and `.enumIn` or something like that.

But, can be taken up separately. Approving now to unblock.

sharat87 (dismissed):

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ce/bridge/BridgeQuery.java
abhvsn: Yeah agreed with separate names added a commit.

sharat87 (approved):

PR #34628, opened 2024-07-02
fix: added backwards compatibility for older snowflake datasources

### LINKED ISSUE #34627: Add backwards compatibility to avoid data migration
https://www.notion.so/Add-backwards-compatibility-to-avoid-data-migration-3f117de33c5c4a29985b54a09f7f1f4e

sneha122:
## Description
This PR adds backwards compatibility for older snowflake datasources so that we wont need migration. 

With new updates on snowflake plugin for key pair authentication, we have introduced a new field in authentication object called authenticationType, this field is responsible for telling us whether it's a basic authentication or key pair authentication. For older datasources, this field wont be there, so in order to ensure that those datasource continue to work smoothly, we have added a fallback mechanism, where if any datasource does not have authentication field, it will be considered as basic auth and we will set its properties accordingly. This PR adds that support

### Steps to test:
1. Create a snowflake datasource along with queries on app.appsmith.com
2. Attach these queries to table widget so they run on page load
3. Export this app and get json
4. Import this json on the DP of this PR, the queries should work

Fixes #34627
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9759249212>
> Commit: f68972710918b450d989c1d28a9286a397fc08bd
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9759249212&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Datasource`

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced backward compatibility for Snowflake plugins by handling cases where `authenticationType` is not present in older datasources.

- **New Features**
  - Improved datasource configuration by cleaning the authentication object in the `DatasourcesApi`.

- **Refactor**
  - Simplified authentication type handling by removing unused `USERNAME_PASSWORD` and `KEY_PAIR` enums in `DBAuth`.

### DIFF
--- a/app/server/appsmith-plugins/snowflakePlugin/src/main/java/com/external/plugins/SnowflakePlugin.java
+++ b/app/server/appsmith-plugins/snowflakePlugin/src/main/java/com/external/plugins/SnowflakePlugin.java
@@ -204,13 +204,16 @@ public class SnowflakePlugin extends BasePlugin {
                 DatasourceConfiguration datasourceConfiguration, Properties properties) {
             // Only for username password auth, we need to set these properties, for others
             // like key-pair auth, authentication specific properties need to be set on config itself
+            // check for authenticationType == null is added to provide backwards compatibility
             AuthenticationDTO authentication = datasourceConfiguration.getAuthentication();
-            switch (authentication.getAuthenticationType()) {
-                case DB_AUTH:
-                    DBAuth dbAuth = (DBAuth) authentication;
-                    properties.setProperty("user", dbAuth.getUsername());
-                    properties.setProperty("password", dbAuth.getPassword());
-                    break;
+            String authenticationType =
+                    datasourceConfiguration.getAuthentication().getAuthenticationType();
+            Boolean isDBAuth =
+                    (authenticationType != null && DB_AUTH.equals(authenticationType)) || authenticationType == null;
+            if (isDBAuth) {
+                DBAuth dbAuth = (DBAuth) authentication;
+                properties.setProperty("user", dbAuth.getUsername());
+                properties.setProperty("password", dbAuth.getPassword());
             }
             properties.setProperty(
                     "warehouse",
@@ -477,6 +480,10 @@ public class SnowflakePlugin extends BasePlugin {
                     default:
                         break;
                 }
+            } else {
+                // Else cases added for older datasources, they would not have authenticationType in their config object
+                // As this is newly introduced
+                configMono = getBasicAuthConfig(commonConfig, datasourceConfiguration);
             }
             return configMono;
         }
--- a/app/server/appsmith-plugins/snowflakePlugin/src/main/java/com/external/plugins/SnowflakePlugin.java
+++ b/app/server/appsmith-plugins/snowflakePlugin/src/main/java/com/external/plugins/SnowflakePlugin.java
@@ -204,13 +204,9 @@ public class SnowflakePlugin extends BasePlugin {
                 DatasourceConfiguration datasourceConfiguration, Properties properties) {
             // Only for username password auth, we need to set these properties, for others
             // like key-pair auth, authentication specific properties need to be set on config itself
-            // check for authenticationType == null is added to provide backwards compatibility
             AuthenticationDTO authentication = datasourceConfiguration.getAuthentication();
-            String authenticationType =
-                    datasourceConfiguration.getAuthentication().getAuthenticationType();
-            Boolean isDBAuth =
-                    (authenticationType != null && DB_AUTH.equals(authenticationType)) || authenticationType == null;
-            if (isDBAuth) {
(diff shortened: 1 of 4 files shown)


app/server/appsmith-plugins/snowflakePlugin/src/main/java/com/external/plugins/SnowflakePlugin.java
```diff
@@ -204,13 +204,16 @@ public Properties addAuthParamsToConnectionConfig(
...
-            switch (authentication.getAuthenticationType()) {
-                case DB_AUTH:
-                    DBAuth dbAuth = (DBAuth) authentication;
-                    properties.setProperty("user", dbAuth.getUsername());
-                    properties.setProperty("password", dbAuth.getPassword());
-                    break;
+            String authenticationType =
+                    datasourceConfiguration.getAuthentication().getAuthenticationType();
+            Boolean isDBAuth =
+                    (authenticationType != null && DB_AUTH.equals(authenticationType)) || authenticationType == null;
```
NilanshBansal, line 212: @sneha122 can we simplify this condition by just
`authenticationType == null || DB_AUTH.equals(authenticationType)`

NilanshBansal, line 212: @sneha122 Can we also set the datasourceConfiguration.authentication.authenticationType as  "dbAuth" here, so that when the datasource gets updated the client sends this value to be set in the field. 
Slowly this would lead to all our database documents to be updated with the right value.

app/server/appsmith-plugins/snowflakePlugin/src/main/java/com/external/plugins/SnowflakePlugin.java
```diff
@@ -477,6 +480,10 @@ private Mono<HikariConfig> getHikariConfig(
                     default:
                         break;
                 }
+            } else {
+                // Else cases added for older datasources, they would not have authenticationType in their config object
+                // As this is newly introduced
+                configMono = getBasicAuthConfig(commonConfig, datasourceConfiguration);
```
NilanshBansal, line 486: @sneha122 is there a way we can omit this else condition? Can we add the datasourceConfiguration.authentication.authenticationType as "dbAuth" as we encounter null value there, so that we don't need to add specific conditions everywhere in the code. 

You can check by creating a getter function for `getAuthenticationType()` in the domain object itself which will avoid all these checks anywhere this value is getting accessed.

app/server/appsmith-plugins/snowflakePlugin/src/main/java/com/external/plugins/SnowflakePlugin.java
sneha122: Will DB_AUTH.equals(authenticationType) fail if authenticationType is null?

sneha122: I dont think we should do this, this is getting executed in the context of execute call i.e. when query is executed, we shouldn't implicitly update the database here (This action should be done only by user explicitly saving the datasource)

app/server/appsmith-plugins/snowflakePlugin/src/main/java/com/external/plugins/SnowflakePlugin.java
sneha122: I dont think we should do this for reasons mentioned above, other than that, authenticationType is a generic property that exists on AuthenticationDTO, If we move this logic in the object itself, it would be applicable for all datasources that use AuthenticationDTO. I believe logic that is specific to snowflake should remain in its own module as opposed to moving it to generic place.

app/server/appsmith-plugins/snowflakePlugin/src/main/java/com/external/plugins/SnowflakePlugin.java
NilanshBansal: No, because the OR condition second part will only be evaluated if the first part is FALSE.
Also, the condition will eitherways not fail if we do a string comparison with null as it will just return False

NilanshBansal: @sneha122 I meant that update the value in the DTO object itself, and not database. 
Just sending this value to the client in the response and I believe the client updates the react store with the api response and whenever there is an update api call triggered, client will send this as payload to the server.

NilanshBansal (approved):

PR #34641, opened 2024-07-02
chore: Remove `updateAndReturn` and start to rely on `updateById`

abhvsn:
## Description
PR to remove `updateAndReturn` and instead start relying on `updateById` repository method. Both of these method is doing the similar op, hence the cleanup. Also on pg branch `updateAndReturn` needs to be fixed if we want to keep supporting both the methods hence the cleanup approach is taken.  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **Refactor**
  - Replaced `updateAndReturn` method with `updateById` for updating tenant configuration.

- **Bug Fixes**
  - Improved the way tenant configurations are updated to enhance reliability and maintainability.

### DIFF
--- a/app/server/appsmith-server/src/test/java/com/appsmith/server/services/ce/TenantServiceCETest.java
+++ b/app/server/appsmith-server/src/test/java/com/appsmith/server/services/ce/TenantServiceCETest.java
@@ -9,7 +9,6 @@ import com.appsmith.server.domains.TenantConfiguration;
 import com.appsmith.server.exceptions.AppsmithException;
 import com.appsmith.server.helpers.FeatureFlagMigrationHelper;
 import com.appsmith.server.helpers.UserUtils;
-import com.appsmith.server.helpers.ce.bridge.Bridge;
 import com.appsmith.server.repositories.CacheableRepositoryHelper;
 import com.appsmith.server.repositories.TenantRepository;
 import com.appsmith.server.repositories.UserRepository;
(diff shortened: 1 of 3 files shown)


app/server/appsmith-server/src/test/java/com/appsmith/server/services/ce/TenantServiceCETest.java
```diff
@@ -88,9 +87,9 @@ public void setup() throws IOException {
         assert tenant != null;
         originalTenantConfiguration = tenant.getTenantConfiguration();
 
-        tenantRepository
-                .updateAndReturn(tenant.getId(), Bridge.update().set(Tenant.Fields.tenantConfiguration, null), null)
-                .block();
+        Tenant update = new Tenant();
+        update.setTenantConfiguration(null);
+        tenantRepository.updateById(tenant.getId(), update, null).block();
```
sharat87, line 92: Do you think we should prefer `Bridge.update()` where we can, instead of using a resource object? Using the resource object to do updates has always been flaky (default field values in the entity class also get updated if we're not careful). But we didn't have good tools with MongoDB to avoid that. Now, I think we should slowly deviate away from it. What say?

sharat87 (changes requested):

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ce/TenantServiceCETest.java
abhvsn: @sharat87 do you mean introducing another method something like `updateById(String id, BridgeUpdate update, AclPermission permission)`?

abhvsn: Also is it okay to let this pass and take that up as a separate task which will bring down the usage of updating via resource object significantly?

sharat87 (approved):

abhvsn:
Closing this as we fixed the issue on pg branch. We will take up separate refactor for reducing the dependency on updateById method which updates the DB object based on the resource instead of UpdateObject.

PR #34649, opened 2024-07-02
test: small change in common method

NandanAnantharamu:
/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9764002269>
> Commit: 9d629c9dd28abe15aadd859dc68c079462539bc8
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9764002269&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of waiting for elements to appear on the Home Page by adding a conditional check, reducing potential errors related to undefined elements.

### DIFF
--- a/app/client/cypress/support/Pages/HomePage.ts
+++ b/app/client/cypress/support/Pages/HomePage.ts
@@ -522,7 +522,7 @@ export class HomePage {
       "not.include",
       "edit",
     );
-    this.agHelper.WaitUntilEleAppear(element ?? this.locator._backToEditor);
+    if (element) this.agHelper.WaitUntilEleAppear(this.locator._backToEditor);
     this.agHelper.AssertElementExist(this.deployHelper._deployPageWidgets);
     this.agHelper.AssertElementVisibility(this.deployHelper._deployPageWidgets);
     this.agHelper.AssertElementVisibility(this.deployHelper._appViewPageName);


app/client/cypress/support/Pages/HomePage.ts
```diff
@@ -522,7 +522,7 @@ export class HomePage {
       "not.include",
       "edit",
     );
-    this.agHelper.WaitUntilEleAppear(element ?? this.locator._backToEditor);
+    if (element) this.agHelper.WaitUntilEleAppear(this.locator._backToEditor);
```
ApekshaBhosale, line 525: @NandanAnantharamu is the condition right? if element is present then wait for till elem appear?

ApekshaBhosale (changes requested):

app/client/cypress/support/Pages/HomePage.ts
NandanAnantharamu: @ApekshaBhosale earlier condition was working even without element, added this to make sure the page is loaded completely

ApekshaBhosale: but isn't condition should be if element is not present then wait for it to appear? what is the point of wait untill appear if element is already there @NandanAnantharamu

NandanAnantharamu: ok let me update with AssertElement visibility and then click on backToEditor

NandanAnantharamu: @ApekshaBhosale so keeping the changes as is as discussed as its a necessity for one of the tests we have

ApekshaBhosale (approved):

PR #34652, opened 2024-07-03
chore: Remove unused `new` from JSON exports/responses

sharat87:
This field isn't used on client and so shouldn't be sent across.

EE PR at https://github.com/appsmithorg/appsmith-ee/pull/4466. All pass.

**/test all**

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9853291119>
> Commit: 22f738390a9e442c3e5d92f9007763d8c1e767b4
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9853291119&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Tue, 09 Jul 2024 09:02:56 UTC

## Summary by CodeRabbit

- **Bug Fixes**
  - Resolved JSON formatting issues in several export fixture files.
  - Removed unnecessary field values from action configurations to ensure cleaner and more efficient data handling.

- **Refactor**
  - Updated the `isNew` method's annotation in `BaseDomain.java` to improve JSON serialization behavior.


app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/BaseDomain.java
```diff
@@ -78,7 +79,7 @@ public abstract class BaseDomain implements Persistable<String>, AppsmithDomain,
     protected Set<Policy> policies = new HashSet<>();
 
     @Override
-    @JsonView(Views.Public.class)
+    @JsonIgnore
     public boolean isNew() {
         return this.getId() == null;
     }
```
abhvsn, line 85: If this is not being used anywhere in the code does it make sense to remove this piece of code completely.

abhvsn (commented):
Minor comment!

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/models/BaseDomain.java
sharat87: Oh it's being used in the server code, quite a bit. It's also important in `pg`, where we use this in our overridden `.save` method in `BaseCake`. Just that the client doesn't need it, so we should send it across. Server very much uses and needs it.

abhvsn (approved):

PR #34708, opened 2024-07-04
chore: Compute default value for jsaction params

ashit-rath:
## Description
This PR adds the values to jsArguments. The logic for this is
- If the value is string then it is kept as is
- For non-strings they are wrapped with `{{ }}` do maintain the data type integrity when evaluated.

This property is currently not used anywhere in the platform and this is intended to be used by js modules to identify the default values of parameters and provide support to alter then in a UI in the app.

This PR also splits `workers/Evaluation/getJSActionForEvalContext.ts` to override in the EE for modules

PR for https://github.com/appsmithorg/appsmith-ee/pull/4612

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9919551354>
> Commit: c6ab372477fb3fd2f1ce171729af4fa64ac2a487
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9919551354&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Sat, 13 Jul 2024 12:16:08 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
  - Added support for additional node types (`RestElement`, `ObjectPattern`, `ArrayPattern`) in our AST processing.
  - Introduced `addPropertiesToJSObjectCode` function to enhance JavaScript object property management.

- **Updates**
  - Enhanced `myFun2` function with new parameters and default values to improve flexibility and usage.
  - Improved `parseJSObject` function with additional parameters for better functionality.

- **Tests**
  - Added a new test suite for `addPropertiesToJSObjectCode` function to ensure robust property management in JavaScript objects.


app/client/packages/ast/src/index.ts
```diff
@@ -530,14 +532,22 @@ export const getFunctionalParamsFromNode = (
...
-          // });
+          const defaultValueInString = code.slice(
+            paramNode.right.start,
+            paramNode.right.end,
+          );
+          const defaultValue =
+            paramNode.right.type === "Literal" &&
+            typeof paramNode.right.value === "string"
+              ? paramNode.right.value
+              : `{{${defaultValueInString}}}`;
```
rishabhrathod01, line 546: Could you elaborate on how adding `{{ }}` helps to maintain datatype integrity as mentioned in the description?

app/client/packages/ast/src/jsObject/index.ts
```diff
@@ -122,9 +122,7 @@ export const parseJSObject = (code: string) => {
     };
 
     if (isPropertyAFunctionNode(node.value)) {
-      // if in future we need default values of each param, we could implement that in getFunctionalParamsFromNode
-      // currently we don't consume it anywhere hence avoiding to calculate that.
-      const params = getFunctionalParamsFromNode(node.value);
+      const params = getFunctionalParamsFromNode(node.value, true, code);
```
rishabhrathod01, line 125: Do we need the defaultValue that will be configured here? I would like to understand how it will be used.

app/client/packages/ast/src/index.ts
ashit-rath: When this is consumed/evaluated, having the `{{ }}` makes sure it is computed in the right data since everything is stored as string.
For example if there is a function `sum(a = 10, b = 20)`
The values would be stored as `{{10}}` and `{{20}}`. So when this is evaluated; the result is always `10` and `20`.
But if the bindings are removed then it is always treated as string and the result after eval is always `"10"` and `"20"`.

app/client/packages/ast/src/jsObject/index.ts
ashit-rath: The reason why it is added here is to avoid double AST parsing. The JS module needs to know the arguments and default values; if it is not calculated here then it has to be re-parsed in the saga where the arguments and values are stored.. This would lead to double parsing and since we already have the logic here and the default value usage is not present anywhere; it seemed a better idea from perf POV to not redo the same the in a different place as the logic would be exactly same.

ashit-rath: Check the EE PR for more details of usage of this.

rishabhrathod01: @ashit-rath won't this add a small performance degrade for parsing of JSObjects?

rishabhrathod01 (dismissed):
LGTM

rishabhrathod01 (approved):

PR #34736, opened 2024-07-04
ci: Support for limited tests run from PR description

sharat87:
Specify the specs we want to run, in a plain code fence, where the first line has to exactly be `/test`. The remaining lines have to be one spec file per line.

Like this:

```
/test
cypress/e2e/Regression/ClientSide/Fork/ForkApplication_spec.ts
cypress/e2e/Regression/ClientSide/JSLibrary/Library_spec.ts
```

1. It's not possible to run with tags and specs and the same time. This will need a slightly larger effort. But we'll get there.
2. The status message shows spec files like they are space-separated, and isn't very easy to read. Don't take it too seriously yet, we'll improve.
3. Globs _should_ also work fine. The `spec` content, _almost_ directly lands into the `--spec` argument of Cypress. See docs at https://docs.cypress.io/guides/guides/command-line#cypress-run-spec-lt-spec-gt.
4. Multiple runners is disabled for this, have to solve it separately. The `ci-test-limited.yml` also doesn't do multiple runners either.

> [!WARNING]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9803353429>
> Commit: e99ae387c9a239ae51155af45977c285474cfbaf
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9803353429&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: 
> Spec: cypress/e2e/Regression/ClientSide/Fork/ForkApplication_spec.ts
> cypress/e2e/Regression/ClientSide/JSLibrary/Library_spec.ts
> 
> It seems like **no tests ran** 😔. We are not able to recognize it, please check <a href="https://github.com/appsmithorg/appsmith/actions/runs/9803353429" target="_blank">workflow here</a>.
> <hr>Fri, 05 Jul 2024 05:20:23 UTC

## Summary by CodeRabbit

- **Chores**
  - Enhanced GitHub Actions workflows to improve test specification handling and reporting.
  - Refactored tag parsing logic in scripts for more accurate and informative outputs.

### DIFF
--- a/.github/workflows/pr-automation.yml
+++ b/.github/workflows/pr-automation.yml
@@ -61,6 +61,7 @@ jobs:
             Commit: ${{ github.event.pull_request.head.sha }}
             Workflow: `${{ github.workflow }}`
             Tags: `${{ steps.parseTags.outputs.tags }}`
+            Spec: `${{ steps.parseTags.outputs.spec }}`
         with:
           script: |
             require("write-cypress-status.js")({core, context, github}, "important", process.env.BODY)
--- a/.github/workflows/pr-automation.yml
+++ b/.github/workflows/pr-automation.yml
@@ -19,6 +19,7 @@ jobs:
         shell: bash
     outputs:
       tags: ${{ steps.parseTags.outputs.tags }}
+      spec: ${{ steps.parseTags.outputs.spec }}
       matrix: ${{ steps.checkAll.outputs.matrix }}
     steps:
 
@@ -72,5 +73,6 @@ jobs:
     secrets: inherit
     with:
       tags: ${{ needs.parse-tags.outputs.tags}}
+      spec: ${{ needs.parse-tags.outputs.spec}}
       matrix: ${{ needs.parse-tags.outputs.matrix}}
       is-pg-build: ${{ github.event.pull_request.base.ref == 'pg' }}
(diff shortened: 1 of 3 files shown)


riodeuno:
@sharat87 I am not sure if I got the message in the PR description correctly. The question is -- we're getting the message that "no tests ran", is that expected because we're only running individual specs right now?

I don't see a problem if this is actually the case, as only power users would use the spec list way of testing and would very likely know that the "no tests ran" message is a consequence of not running tags.

ApekshaBhosale:
@sharat87 one small thing, can we use less machines based on number of specs to run? we have reduced to only one machine for `ci-test-limited`. Before it used to be three and majorly only one was required, wasting 2 resources.

sharat87:
Thanks for checking this out folks! ❤️ 

@riodeuno, I should've mentioned this, please ignore the Cypress report part of the PR description. It's not behaving correctly because there's changes in `pr-automation.yml` that have to be in `release` branch before the rest of the functionality in this PR can take any affect. I've verified that part before opening this up for review. I can force-merge the changes in just the `pr-automation.yml` file, and then test this PR here. Then the Cypress report should be more accurate.

@ApekshaBhosale, yes, indeed. This is how `ci-test-limited` is also operating today. It has no `matrix` in it, and it only runs on one runner, irrespective of how many spec files we're running. So I kept it to that in this PR. I want to add support for multiple runners for this, in a future PR. But since this aligns with what we have today, I preferred to keep the PR focused just to that.

Let me know what you guys think! 🙏

yatinappsmith (approved):

PR #34770, opened 2024-07-08
ci: Clean up disk usage before saving container to disk

sharat87:
Saving the Docker image to disk is failing on EE `pg` because of lack of disk space. So we clean up all the unused Docker images just before doing this.

## Summary by CodeRabbit

- **Chores**
  - Updated workflow to remove a specific Docker image before saving a new one, improving Docker image management.

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -113,6 +113,10 @@ jobs:
       # Saving the docker image to tar file
       - name: Save Docker image to tar file
         run: |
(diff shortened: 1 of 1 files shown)


abhvsn (approved):
@sharat87 is this happening because we started to bundle both Mongo and PG server builds in the docker image or something else is a culprit? Wanted to get the understanding why now 🙂

sharat87:
@abhvsn, that's right. The image is now _too_ big for this. 😞

PR #34788, opened 2024-07-08
chore: Don't duplicate the URL parsing Regexes

sharat87:
This code duplication meant that we missed updating the regexes to the ones that support UUIDs, and so the caching functionality broke.

This PR removes the duplication and imports the regex constants from a single place.

/test sanity

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9850040595>
> Commit: d8261a1b5ed55c7258260695fea2f2ca32dc0484
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9850040595&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Sanity`
> Spec:
> <hr>Tue, 09 Jul 2024 03:00:41 UTC

## Summary by CodeRabbit

- **Refactor**
  - Improved maintainability by importing path constants from a centralized location.
  - Replaced hardcoded `pageId` and `applicationId` with variables for better code clarity and flexibility.

### DIFF
--- a/app/client/src/ce/utils/serviceWorkerUtils.ts
+++ b/app/client/src/ce/utils/serviceWorkerUtils.ts
@@ -2,13 +2,14 @@ import { Mutex } from "async-mutex";
 import { APP_MODE } from "entities/App";
 import type { Match, TokensToRegexpOptions } from "path-to-regexp";
 import { match } from "path-to-regexp";
-
-const BUILDER_PATH = `/app/:applicationSlug/:pageSlug(.*\-):pageId/edit`;
-const BUILDER_CUSTOM_PATH = `/app/:customSlug(.*\-):pageId/edit`;
-const VIEWER_PATH = `/app/:applicationSlug/:pageSlug(.*\-):pageId`;
-const VIEWER_CUSTOM_PATH = `/app/:customSlug(.*\-):pageId`;
-const BUILDER_PATH_DEPRECATED = `/applications/:applicationId/pages/:pageId/edit`;
-const VIEWER_PATH_DEPRECATED = `/applications/:applicationId/pages/:pageId`;
+import {
+  BUILDER_PATH,
+  BUILDER_CUSTOM_PATH,
+  VIEWER_PATH,
+  VIEWER_CUSTOM_PATH,
+  BUILDER_PATH_DEPRECATED,
+  VIEWER_PATH_DEPRECATED,
+} from "@appsmith/constants/routes/appRoutes";
 
 interface TMatchResult {
   pageId?: string;
--- a/app/client/src/ce/utils/serviceWorkerUtils.test.ts
+++ b/app/client/src/ce/utils/serviceWorkerUtils.test.ts
@@ -18,10 +18,13 @@ import { Request as NFRequest, Response as NFResponse } from "node-fetch";
   delete: jest.fn(),
 };
 
+const applicationId = "b0123456789abcdef0000000";
+const pageId = "a0123456789abcdef0000000";
+
 describe("serviceWorkerUtils", () => {
   describe("matchBuilderPath", () => {
     it("should match the standard builder path", () => {
-      const pathName = "/app/applicationSlug/pageSlug-123/edit";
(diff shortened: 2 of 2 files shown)


dvj1988:
@sharat87 Can you also create a EE test PR also for these change? Last time when I tried importing the regexes from appRoutes there were some issues in the build. I would like to do a manual test before we merge.

app/client/src/ce/utils/serviceWorkerUtils.ts
```diff
@@ -2,13 +2,14 @@ import { Mutex } from "async-mutex";
 import { APP_MODE } from "entities/App";
```
riodeuno, line 2: @dvj1988 @sharat87 Is it a good idea to tie the service worker utilities to our main app's code? I'm wondering if it effects the chucks or the build process? Basically, servicer worker registration and code in the wild seems to be separated from the main codebase.

sharat87:
Thanks @dvj1988 for the heads-up. Indeed there was a failure, and I fixed it in this EE PR: https://github.com/appsmithorg/appsmith-ee/pull/4565.

app/client/src/ce/utils/serviceWorkerUtils.ts
sharat87: @riodeuno, I suspected that too. I think it's quite likely as well. But from the code, it looks like we're mostly importing constant strings, both into application code and service worker code, so I think this is okay. I'd rather constants are imported from a single source for both. It'll mean they'll effectively be downloaded twice, once for application and once for service worker, but I think this is okay. Especially since we _need_ the constant value in service worker code, so might as well import from a common source.

I don't know enough to be sure about the above though, so I'd invite more opinions on this. 🙏

riodeuno: Agreed.
We can run `yarn analyze` and see what it looks like.
Looking at it without analyzing it --
The only thing that comes to mind is that the serviceworker js file is created at build time, once generated there shouldn't be any ramifications on the chunks for our codebase. Based on this, this change is fine.

It could be that service worker utils are chunked along with our main app codebase. We're talking about ~250 LoC. The trade-off here is code structure that allows for better maintainability and fewer side effects vs saving the minified version of ~250 lines of unminified code from being downloaded in the chunk. I'm happy to trade in favour of better code structure.

@dvj1988 I too am looking for your opinion on this.

riodeuno: Actually, scratch that. I'm fairly certain that serviceworkerutils would not get into our main codebase chunks. Approving this PR.

riodeuno (approved):

app/client/src/ce/utils/serviceWorkerUtils.ts
dvj1988: I think the constants that are imported are duplicated in both the service worker bundle as well as the main bundle. There is a minor duplication of code in the bundle but as you guys have mentioned it better to have a single source of truth.

This is from https://ce-34788.dp.appsmith.com/pageService.js
Screenshot 2024-07-09 at 2 14 32 PM

This is from https://ce-34788.dp.appsmith.com/static/js/main.46d3445c.js
Screenshot 2024-07-09 at 2 15 36 PM

I believe the reason why we have duplication is because, the two bundles are generated as 2 different entry points in weback and the chunks generated in one entry point is not reused in another.

Also if you check the service worker we can see chunk details of the main bundle being injected into the service worker. So that means the service worker is bundled after the main bundle chunks are created.
Screenshot 2024-07-09 at 2 20 09 PM

dvj1988:
> Thanks @dvj1988 for the heads-up. Indeed there was a failure, and I fixed it in this EE PR: [appsmithorg/appsmith-ee#4565](https://github.com/appsmithorg/appsmith-ee/pull/4565).

Give me time to manually test both ce and ee DPs today.

dvj1988 (approved):
LGTM.

PR #34855, opened 2024-07-10
fix: modal rename update references

### LINKED ISSUE #34823: [Bug]: Changing Modal name does not change its references even after the Modal1.name code refactor

### Description

Hi Team,
I can see the recent changes to move away from using Modal name as string, and instead using an attribute of the Modal widget, eg Modal1.name
But i still face issues after renaming the Modal name. All the widgets or code that is using Modal1.name is not changing to the name Modal widget name, something like Test_Modal.name

Screenshot 2024-07-09 at 4 29 45 PM

### Steps To Reproduce

1. Drag and Drop a button
2. Configure its onClick action to showModal and create new Modal
3. Change modal's name from Modal1 to some other name like "Test_Modal"
4. Now try to close the Modal using the Close button or icon.
5. An error will be shown saying could not find Modal1

On inspecting either of the close button's onClick event, you will see that the reference is still to closeModal(Modal1.name)
and not closeModal(Test_Modal.name)

With the refactor changes done to use Modal.name instead of string, the references should have been changed all across. 
But that does not seem to be happening

### Public Sample App

_No response_

### Environment

Production

### Severity

Low (Cosmetic UI issues)

### Issue video log

_No response_

### Version

Cloud

AmanAgarwal041:
## Description
Whenever the modal widget was dropped, it had cross icon and close button to close the current modal which had reference to the current modal. On changing the modal name the references to both the button was not getting changed. This was happening only when before making changes to both the buttons, the modal name was changed. 
This PR fixes the references for the both the buttons and verifies the flow through cypress test.

Fixes #34823 
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags="@tag.Widget"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/9909276940>
> Commit: d24b817d4b69570dedc455553382a6b09ec14b0f
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=9909276940&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Widget`
> Spec:
> <hr>Fri, 12 Jul 2024 15:13:57 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
  - Added test scenarios to verify that renaming a modal updates all references inside its widgets.

- **Bug Fixes**
  - Enhanced the `ModalWidget` component to properly manage `onClick` events for `iconChild` and `cancelBtnChild` configurations.

- **Tests**
  - Implemented tests to ensure the correct behavior of modal widget properties and interactions during rename events.


app/client/cypress/e2e/Regression/ClientSide/BugTests/Bug_34823_ModalRename_Spec.ts
ApekshaBhosale, line 1: let's please change the spec name and not put it in BugTests if that's possible. spec doesn't give any idea what case it's solving at one glance. Going forward we don't want to have this at all @AmanAgarwal041

ApekshaBhosale (changes requested):

app/client/cypress/e2e/Regression/ClientSide/BugTests/Bug_34823_ModalRename_Spec.ts
rajatagrawal: @coderabbitai Can you make a suggestion to change the name of the spec file that makes it more informative ?

app/client/src/widgets/ModalWidget/widget/index.tsx
```diff
@@ -206,6 +211,11 @@ export class ModalWidget extends BaseWidget<ModalWidgetProps, WidgetState> {
                             propertyName: "onClick",
                             propertyValue: `{{closeModal(${parent.widgetName}.name);}}`,
                           },
+                          {
```
rajatagrawal, line 214: Can we please write a unit test instead of a cypress test that asserts the response of getDefaults function and asserts two cases:

1. Asserts that getDefaults() function returns values equal to the current values.
2. Asserts that getDefaults() function now includes the new json properties added to blueprint operation section ?

cc : @ApekshaBhosale

sagar-qa007 (approved):

AmanAgarwal041:
> Can we please have a unit test instead of a cypress test ? This PR is adding to an existing functionality to our widgets which can be sufficiently tested with a unit test.
> 
> The cypress test is also testing drag and drop, renaming widget , submit and other user flows that can be skipped.

@rajatagrawal I my opinion this should be an integration test as the default properties are same as before, we just have added another key to `dynamicTriggerPathList`. This is a flow that needs to be test if the references of the modal are updated on the components inside modal widget on renaming or not. 
Can you elaborate more about how we can do it via jest test case ?

rajatagrawal (approved):

Issue #34890, opened 2024-07-12
[Task]: Race condition in setup worker and first eval action


ashit-rath:
When a package editor is loaded direct from the url and not from the workspace home page; the first eval doesn't always start due to a race condition.

Race condition:
In order for the eval to start there are 2 pre-requisites
1. Worker setup
2. Dispatch of first eval action

The above 2 needs to happen in sequence because if the first eval action is dispatched and the worker is not ready then the evaluation would not work.
The dispatching of first eval action is always triggered when all the entities of a page/package/workflow are fetched from the server ensuring data completeness before triggering first data tree setup in eval.

But due to some reason the worker setup takes a very long time when any editor is accessed directly from the url and it can go upwards of 4000ms. Due to this delay in some cases package editor entities are fetching completes before the worker setup and fires the first eval action which is missed by post worker setup code flow.

In the screenshot `fetchAllModuleEntityCompletion` indicates the completion of fetching of entities and dispatching of the first eval action.

<img width="2000" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNjdVQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--832e044604e7f9e0477f8e594788e68138b9cf9a/Screenshot%202024-07-12%20at%2011.32.12%E2%80%AFAM.png" alt="Screenshot 2024-07-12 at 11.32.12 AM.png" />

PR #34891, opened 2024-07-12
chore: Buffer first eval action to fix race condition in packages

ashit-rath:
## Description
Check the issue description for the actual problem.

Solution:
To mitigate the race condition; instead of using `take` post setup worker an `actionChannel` is used to create a queue to detect any first eval action dispatched while the code path completes the worker setup. This queue is then picked by and the action is processed and the `actionChannel` is closed.

The baseline theory of the solution is to create a queue of any first eval action dispatch throughout the lifecycle of the setup process but act only when the setup is complete

Fixes https://github.com/appsmithorg/appsmith/issues/34890

## Automation

/ok-to-test tags="@tag.All"

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/client/src/sagas/EvaluationsSaga.ts
+++ b/app/client/src/sagas/EvaluationsSaga.ts
@@ -646,6 +646,7 @@ function* evalAndLintingHandler(
 }
 
 function* evaluationChangeListenerSaga(): any {
+  const firstEvalActionChannel = yield actionChannel(FIRST_EVAL_REDUX_ACTIONS);
   // Explicitly shutdown old worker if present
   yield all([call(evalWorker.shutdown), call(lintWorker.shutdown)]);
   const [evalWorkerListenerChannel] = yield all([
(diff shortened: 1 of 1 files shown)


app/client/src/sagas/EvaluationsSaga.ts
```diff
@@ -669,8 +670,9 @@ function* evaluationChangeListenerSaga(): any {
   yield spawn(handleEvalWorkerRequestSaga, evalWorkerListenerChannel);
 
   const initAction: EvaluationReduxAction<unknown> = yield take(
-    FIRST_EVAL_REDUX_ACTIONS,
+    firstEvalActionChannel,
   );
+  firstEvalActionChannel.close();
```
dvj1988, line 675: @ashit-rath Curious to know, what happens if we don't close the actionChannel here?

ashit-rath: AFAIK It may lead to memory leak as it would still be subscribed to all the action.

dvj1988 (approved):

PR #34992, opened 2024-07-17
ci: Add a pre-push hook to check for only ce changes

rajatagrawal:
## Description
This PR checks for file changes when pushing in CE repo and passes if the changes are present only in CE folder. To override the check, developers can use --no-verify check when pushing the changes. This PR is an extension of https://github.com/appsmithorg/appsmith/pull/32791

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **Chores**
  - Implemented a pre-push Git hook to check for specific file changes and provide feedback before pushing to a remote repository.

### DIFF
--- a/app/client/.husky/pre-push
+++ b/app/client/.husky/pre-push
@@ -1,0 +1,95 @@
+#!/bin/bash
+
+echo "Pre push hook called"
+
+# An example hook script to verify what is about to be pushed.  Called by "git
+# push" after it has checked the remote status, but before anything has been
+# pushed.  If this script exits with a non-zero status nothing will be pushed.
+#
+# This hook is called with the following parameters:
+#
+# $1 -- Name of the remote to which the push is being done
+# $2 -- URL to which the push is being done
+#
+# If pushing without using a named remote those arguments will be equal.
+#
+# Information about the commits which are being pushed is supplied as lines to
+# the standard input in the form:
+#
+#   <local ref> <local oid> <remote ref> <remote oid>
+
+appsmith_ee_url="appsmith-ee.git"
+appsmith_ce_url="appsmithorg/appsmith.git"
+
+# Define the null SHA. 
+null_sha="0000000000000000000000000000000000000000"
+
+# Function to get list of files between two commits
+do_commits_contain_ee_files() {
+    # Store the commit hashes
+    from_commit=$1
+    to_commit=$2
+    string_to_match="app/client/src/ee"
+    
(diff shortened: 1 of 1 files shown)


app/client/.husky/pre-push
```diff
@@ -0,0 +1,95 @@
...
+    
+    # Get the list of files between the two commits
+    files=$(git diff --name-only $from_commit $to_commit)
+
+    # Iterate over each file
+    for file in $files; do
+        # Check if the file path contains the string
+        if [[ "$file" == *"$string_to_match"* ]]; then
+            echo "File '$file' matches the string '$string_to_match'"
+            return 0
```
KelvinOm, line 57: As I understand, we need to check all files but here we just check only one file and than exit with success. Is this the expected behavior or am I reading the code incorrectly?

app/client/.husky/pre-push
```diff
@@ -0,0 +1,95 @@
...
+        fi
+    done
+    return 1
```
KelvinOm, line 60: Let's add a message about what went wrong and how to fix it.

app/client/.husky/pre-push
rajatagrawal: I am looping over all affected files in line 53

app/client/.husky/pre-push
rajatagrawal: The message is added on line 89. Does it look sufficient ? I can elaborate it if needed.

app/client/.husky/pre-push
KelvinOm: I got it. Let's say we take the first file then checked it then we get into this if and exit. Don't we need to check the rest of the files?

app/client/.husky/pre-push
KelvinOm: Got it. That's enough.

app/client/.husky/pre-push
rajatagrawal: No. I think if there is even one file that is EE folder that would be a sufficient condition for the hook to fail I think. The hook can let the user know of other errors when try to push again after fixing the first error.

KelvinOm: Got it. I was confused by the fact that if we return 0 here, then the output to the script itself will be 1.

KelvinOm (approved):

PR #35087, opened 2024-07-22
feat: set app max width

### LINKED ISSUE #34993: feat: set application max width

znamenskii-ilia:
## Description
Add a new application setting which would allow users to set max width of their applications.

Fixes #34993

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/10159296604>
> Commit: d2a98dcaecda2903dd8233f1f53ea119721095fa
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=10159296604&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Tue, 30 Jul 2024 09:50:48 UTC

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **New Features**
	- Introduced a new constant for application maximum width settings, allowing for `Unlimited`, `Large`, and `Medium` options.
	- Added an `AppMaxWidthSelect` component for users to select maximum width configurations from a dropdown menu.
	- Enhanced `Canvas` and `MainContainerWrapper` components to accept and utilize maximum width settings for improved layout adaptability.
	- Implemented a new selector to retrieve the maximum width setting for use in various components.
	- Enhanced the theme property pane to allow users to customize the application layout through maximum width selection.
	- Updated `CanvasFactory` to utilize the application theme settings for a more responsive layout.

- **Bug Fixes**
	- Improved initialization logic to prevent null reference issues related to theme settings.

- **Tests**
	- Enhanced test cases to verify inclusion of maximum width settings in exported applications and during application imports.

### DIFF
--- a/app/client/packages/design-system/theming/src/token/src/types.ts
+++ b/app/client/packages/design-system/theming/src/token/src/types.ts
@@ -140,3 +140,11 @@ export type Typography = {
 };
 
 export type IconStyle = "outlined" | "filled";
+
+export const APP_MAX_WIDTH = {
+  Unlimited: "UNLIMITED",
+  Large: "LARGE",
+  Medium: "MEDIUM",
+} as const;
+
+export type APP_MAX_WIDTH = (typeof APP_MAX_WIDTH)[keyof typeof APP_MAX_WIDTH];
--- a/app/client/src/layoutSystems/anvil/viewer/canvas/AnvilViewerCanvas.tsx
+++ b/app/client/src/layoutSystems/anvil/viewer/canvas/AnvilViewerCanvas.tsx
@@ -4,6 +4,32 @@ import type { BaseWidgetProps } from "widgets/BaseWidgetHOC/withBaseWidgetHOC";
 import { getAnvilCanvasId } from "./utils";
 import { LayoutProvider } from "layoutSystems/anvil/layoutComponents/LayoutProvider";
 import { AnvilDetachedWidgets } from "./AnvilDetachedWidgets";
+import styled from "styled-components";
+import { APP_MAX_WIDTH } from "@design-system/theming";
+
+const appMaxWidthToCSSValue = (maxWidth: APP_MAX_WIDTH): string => {
+  switch (maxWidth) {
+    case APP_MAX_WIDTH.Unlimited:
+      return "auto";
+    case APP_MAX_WIDTH.Large:
+      return "1080px";
+    case APP_MAX_WIDTH.Medium:
+      return "800px";
+    default: {
+      const exhaustiveCheck: never = maxWidth;
+      throw new Error(`Unhandled maxWidth: ${exhaustiveCheck}`);
+    }
+  }
+};
+
+const RootStyled = styled.div<{
+  maxWidth?: APP_MAX_WIDTH;
+}>`
+  max-width: ${({ maxWidth }) =>
+    maxWidth ? `${appMaxWidthToCSSValue(maxWidth)}` : "auto"};
+  margin: ${({ maxWidth }) => (maxWidth ? "0 auto" : "inherit")};
+`;
+
 export const AnvilViewerCanvas = React.forwardRef(
   (props: BaseWidgetProps, ref: React.ForwardedRef<HTMLDivElement>) => {
     const className: string = useMemo(
@@ -14,14 +40,15 @@ export const AnvilViewerCanvas = React.forwardRef(
     return (
       <>
         <AnvilDetachedWidgets />
-        <div
+        <RootStyled
           className={className}
           id={getAnvilCanvasId(props.widgetId)}
+          maxWidth={props.maxWidth}
           ref={ref}
           tabIndex={0} //adding for accessibility in test cases.
         >
           <LayoutProvider {...props} />
-        </div>
+        </RootStyled>
       </>
     );
   },
--- a/app/client/src/pages/Editor/Canvas.tsx
+++ b/app/client/src/pages/Editor/Canvas.tsx
@@ -42,10 +42,10 @@ const appMaxWidthToCSSValue = (maxWidth: APP_MAX_WIDTH): string => {
       return "1080px";
     case APP_MAX_WIDTH.Medium:
       return "800px";
-    default:
+    default: {
       const exhaustiveCheck: never = maxWidth;
-
       throw new Error(`Unhandled maxWidth: ${exhaustiveCheck}`);
+    }
   }
 };
 
--- a/app/client/src/pages/Editor/Canvas.tsx
+++ b/app/client/src/pages/Editor/Canvas.tsx
@@ -9,6 +9,7 @@ import { combinedPreviewModeSelector } from "selectors/editorSelectors";
 import { getSelectedAppTheme } from "selectors/appThemingSelectors";
 import { getViewportClassName } from "layoutSystems/autolayout/utils/AutoLayoutUtils";
 import {
+  APP_MAX_WIDTH,
   ThemeProvider as WDSThemeProvider,
   useTheme,
 } from "@design-system/theming";
@@ -24,6 +25,7 @@ import { focusWidget } from "actions/widgetActions";
 interface CanvasProps {
   widgetsStructure: CanvasWidgetStructure;
   canvasWidth: number;
+  maxWidth?: APP_MAX_WIDTH;
   enableMainCanvasResizer?: boolean;
 }
 
@@ -32,18 +34,37 @@ const StyledWDSThemeProvider = styled(WDSThemeProvider)`
   display: flex;
 `;
 
+const appMaxWidthToCSSValue = (maxWidth: APP_MAX_WIDTH): string => {
+  switch (maxWidth) {
+    case APP_MAX_WIDTH.Unlimited:
+      return "auto";
+    case APP_MAX_WIDTH.Large:
+      return "1080px";
+    case APP_MAX_WIDTH.Medium:
+      return "800px";
+    default:
+      const exhaustiveCheck: never = maxWidth;
+
+      throw new Error(`Unhandled maxWidth: ${exhaustiveCheck}`);
+  }
+};
+
 const Wrapper = styled.section<{
   background: string;
   width: number;
+  maxWidth?: APP_MAX_WIDTH;
   $enableMainCanvasResizer: boolean;
 }>`
   flex: 1;
   background: ${({ background }) => background};
   width: ${({ $enableMainCanvasResizer, width }) =>
     $enableMainCanvasResizer ? `100%` : `${width}px`};
+  max-width: ${({ maxWidth }) =>
+    maxWidth ? `${appMaxWidthToCSSValue(maxWidth)}` : "auto"};
+  margin: ${({ maxWidth }) => (maxWidth ? "0 auto" : "inherit")};
 `;
 const Canvas = (props: CanvasProps) => {
-  const { canvasWidth } = props;
+  const { canvasWidth, maxWidth } = props;
   const isPreviewMode = useSelector(combinedPreviewModeSelector);
   const isAppSettingsPaneWithNavigationTabOpen = useSelector(
     getIsAppSettingsPaneWithNavigationTabOpen,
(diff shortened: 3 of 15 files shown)


app/client/src/pages/Editor/WDSThemePropertyPane/components/AppMaxWidthSelect.tsx
```diff
@@ -0,0 +1,46 @@
+import { APP_MAX_WIDTH } from "@design-system/theming";
+import { Option, Select } from "design-system";
+import React from "react";
+
+interface AppMaxWidthSelectProps {
+  value: APP_MAX_WIDTH;
+  onSelect: (value: string) => void;
+}
+
+const resolveOptionLabelText = (option: APP_MAX_WIDTH) => {
```
alex-golovanov, line 10: Since they map one to one, it seems that a config object would be a more concise way to do the same and you wouldn't have to deal with unresolved `default`.

```ts
const optionLabels = {
  [APP_MAX_WIDTH.Unlimited]: "Unlimited",
  [APP_MAX_WIDTH. Large]: "Large",
  [APP_MAX_WIDTH.Medium]: "Medium",
} as const;

const optionLabel = optionLabels[APP_MAX_WIDTH.Unlimited];
```

znamenskii-ilia: How would you guarantee that there is a translation for each enum item?

KelvinOm: I'd like to add my two cents. It is better not to use enums at all. As for `as const`, it is also better not to abuse them, since this inflates final bundle. The only use case that makes sense for `as const` is when you need to iterate over the values and somehow use them. For everything else, it is enough to use literal union types. It is convenient and safe. 

Related discussions: 
https://theappsmith.slack.com/archives/CGBPVEJ5C/p1686750585115319
https://theappsmith.slack.com/archives/CPG2ZTXEY/p1678081588554009
https://www.notion.so/appsmith/Appsmith-front-end-code-style-guide-86db29586bac4ba3b2e044c315cae892?pvs=4#99a5fe0226954e128b58c6ad013a995e

alex-golovanov: > Could you share your ideas on wow would we guarantee that there is a translation for each enum item?

Sure, something like this will suffice:

```ts
enum APP_MAX_WIDTH  {
  Unlimited = "UNLIMITED",
  Large = "LARGE",
  Medium = "MEDIUM",
};

const optionLabels: Record<APP_MAX_WIDTH, string> = {
  [APP_MAX_WIDTH.Unlimited]: "Unlimited",
  [APP_MAX_WIDTH.Large]: "Large",
  [APP_MAX_WIDTH.Medium]: "Medium",
};

const optionLabel = optionLabels[APP_MAX_WIDTH.Unlimited];
```

alex-golovanov: > I'd like to add my two cents. It is better not to use enums at all. As for `as const`, it is also better not to abuse them, since this inflates final bundle. The only use case that makes sense for `as const` is when you need to iterate over the values and somehow use them. For everything else, it is enough to use literal union types. It is convenient and safe.
> 
> Related discussions: https://theappsmith.slack.com/archives/CGBPVEJ5C/p1686750585115319 https://theappsmith.slack.com/archives/CPG2ZTXEY/p1678081588554009 https://www.notion.so/appsmith/Appsmith-front-end-code-style-guide-86db29586bac4ba3b2e044c315cae892?pvs=4#99a5fe0226954e128b58c6ad013a995e

In this particular case you do need an config for reference, whether it's an object or an enum. I don't buy the hate towards enums, there are, literally, no objective arguments against **string** enums, while they are superior to objects since they offer both value and type. Not to mention you can easily derive a union type from an emum, check for inclusion etc, while unions are much simpler in nature and are not intended for anything that requires iteration.

My five cents - don't discriminate types. Use what is most appropriate for the situation.

znamenskii-ilia: > As for as const, it is also better not to abuse them, since this inflates final bundle.

Could you clarify what you mean? AFAIK TS just strips `as const` and that is it. 

> The only use case that makes sense for as const is when you need to iterate over the values and somehow use them.

That is my case. I need to render Select option for each value.

app/client/src/pages/Editor/Canvas.tsx
```diff
@@ -32,18 +34,37 @@ const StyledWDSThemeProvider = styled(WDSThemeProvider)`
   display: flex;
 `;
 
+const appMaxWidthToCSSValue = (maxWidth: APP_MAX_WIDTH): string => {
```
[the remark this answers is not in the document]
znamenskii-ilia: Thanks for sharing your opinion about ideal place for this function, however I personally find this place right.

app/client/packages/design-system/theming/src/token/src/types.ts
```diff
@@ -140,3 +140,11 @@ export type Typography = {
...
 
 export type IconStyle = "outlined" | "filled";
+
+export const APP_MAX_WIDTH = {
+  Unlimited: "UNLIMITED",
+  Large: "LARGE",
+  Medium: "MEDIUM",
+} as const;
+
+export type APP_MAX_WIDTH = (typeof APP_MAX_WIDTH)[keyof typeof APP_MAX_WIDTH];
```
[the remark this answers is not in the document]
znamenskii-ilia: > Using enum seems like a more straight forward solution, is there a benefit to writing it this way, through object and creating a separate union type? 

Yes, less generated code. Even though it is not a big problem imo. The other reason is personal preference, which is fine if this is not documented in the style guide.

> One thing is for sure though, constants should not be in types.ts.

That is not constants that is enum.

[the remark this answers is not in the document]
alex-golovanov: > That is not constants that is enum.

It is an object and most certainly not an [enum](https://www.typescriptlang.org/docs/handbook/enums.html), hence it does not belong in types.  Adding `as const` makes it readonly, similar result could be achieved using `Object.freeze`.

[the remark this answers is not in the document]
znamenskii-ilia: We mean different thing by enum. As I understood when you say enum you mean TS enum (link to which you shared). I, by enum, any [enumiration](https://en.wikipedia.org/wiki/Enumerated_type#:~:text=In%20computer%20programming%2C%20an%20enumerated,or%20enumerators%20of%20the%20type.).

This is valid JS enum for me:
```
const statuses = {
  active: 1,
  deleted: 2,
};
```
I'd prefer to have them somewhere their own dedicated file (`editor/domain/enum.ts`), but we don't have one yet.

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Application.java
```diff
@@ -479,6 +479,9 @@ public static class ThemeSetting {
         @JsonView({Views.Public.class, Git.class})
         IconStyle iconStyle;
 
+        @JsonView({Views.Public.class, Git.class})
+        AppMaxWidth appMaxWidth = AppMaxWidth.LARGE;
```
znamenskii-ilia, line 483: Hi @nidhi-nair, I've added `AppMaxWidth` prop to AppThemeSettings. The requirement is that it has `LARGE` value by default. That is, this value is set when a user create an app. However I noticed that until a user changes some AppThemeSettings, they are not set to frontend.

Here is Response of newly created app
Screenshot 2024-07-23 at 20 16 18

Here is Response after I updated some AppThemeSettings
Screenshot 2024-07-23 at 20 16 43

I was wondering why we don't send AppThemeSettings to frontend?

riodeuno: @znamenskii-ilia I believe, when creating the application, these settings [need to be sent along with the create application payload](https://github.com/appsmithorg/appsmith/blob/84cd0de775b376c31d3fa2e41c769c8ebc58dcab/app/client/src/ce/sagas/ApplicationSagas.tsx#L512). So, these changes won't be reflected until one of the following happens:
1. This has been set when creating the application (via the create application payload)
2. This has been updated by the user (via the update application payload)

znamenskii-ilia: Tried to fix it [here](https://github.com/appsmithorg/appsmith/pull/35087/files#diff-7d1cb305ad7ef7996a9d6a1c6968199007ac6746b6d996c37756992db0f04eceR31), but it doesn't work

app/client/packages/design-system/theming/src/token/src/types.ts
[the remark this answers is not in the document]
alex-golovanov: Please forgive me, but It like calling apples oranges. We're in TS domain here and calling objects enums because they _can_ have enumerable properties is a bit of a stretch.

app/server/appsmith-server/src/main/java/com/appsmith/server/domains/Application.java
abhvsn: @znamenskii-ilia Currently this is not being sent in the response because it's not getting saved with the create call. To fix this we need changes in both client and server. Current sample request payload:
```
{
    "workspaceId": "66a0d7d10751785f3ea70f88",
    "name": "Untitled application 2",
    "color": "#ECECEC",
    "icon": "team",
    "positioningType": "FIXED",
    "showNavbar": null
}
```

If we want this to be available for client while creating the application you can add the respective field in `ApplicationCreationDTO` class. Something like:
```
ApplicationCreationDTO(
        @NotBlank @Size(max = 99) String workspaceId,
        @NotBlank @Size(max = 99) String name,
        @IconName String icon,
        @Pattern(regexp = "#[A-F0-9]{6}") String color,
        Application.AppPositioning positioningType,
        Application.ThemeSetting.AppMaxWidth appMaxWidth,		// New field introduced 
        Boolean showNavbar)
```   
Also update the `toApplication()` method to set the `themeSetting` field in the application:
```
final Application.ThemeSetting themeSetting = new Application.ThemeSetting();
        themeSetting.setAppMaxWidth(appMaxWidth);
        applicationDetail.setThemeSetting(themeSetting);
```

znamenskii-ilia: @abhvsn  I wonder why we need to send default values from frontend and can't just add them on backend side?

Have you change to check the fix I made?

app/client/packages/design-system/theming/src/token/src/types.ts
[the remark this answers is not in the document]
somangshu: I agree that these are personal observations and we should set such standards outside this PR and as a group. 

Please comments on the main thread that Pawan raised to reinitiate this conversation with the wider audience.

KelvinOm (dismissed):

app/client/src/layoutSystems/anvil/viewer/canvas/AnvilViewerCanvas.tsx
```diff
@@ -4,6 +4,32 @@ import type { BaseWidgetProps } from "widgets/BaseWidgetHOC/withBaseWidgetHOC";
...
+  switch (maxWidth) {
+    case APP_MAX_WIDTH.Unlimited:
+      return "auto";
+    case APP_MAX_WIDTH.Large:
+      return "1080px";
+    case APP_MAX_WIDTH.Medium:
+      return "800px";
+    default: {
+      const exhaustiveCheck: never = maxWidth;
+      throw new Error(`Unhandled maxWidth: ${exhaustiveCheck}`);
```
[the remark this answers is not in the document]
znamenskii-ilia: This will break exhaustive checking. It works only because of `never` type

KelvinOm (approved):

Issue #34996, opened 2024-07-17
[Bug]: appsmith backup fails while taking backup


pratapaprasanna:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

When we try to take a backup of appsmith we see that it fails. 

However, when we try 
```
export APPSMITH_DB_URL=mongodb://appsmith:YQrNSIqBadOHG@localhost:27017/appsmith

root@ee240a85f275:/opt/appsmith# appsmithctl backup
.
.
Finished taking an encrypted backup archive at /appsmith-stacks/data/backup/appsmith-backup-2024-07-17T11-38-20.213Z.tar.gz.enc
Starting the cleanup task after taking a backup.
Cleanup task completed.
```

the value into APPSMITH_DB_URL as mentioned [here](https://github.com/appsmithorg/appsmith/blob/a46c7311790c0f8a68bc2ea28ada67333e9d8a88/deploy/docker/fs/opt/appsmith/utils/bin/mongo_shell_utils.js#L9) is not computed correctly 

### Steps To Reproduce

login to a container using 
```
sudo docker exec -it -u root ee240a85f275 bash
```
and do 

```
root@ee240a85f275:/opt/appsmith# appsmithctl backup
Available free space at /appsmith-stacks
4349689856


Exporting database
2024-07-17T11:41:40.201+0000	error parsing command line options: error parsing uri: scheme must be "mongodb" or "mongodb+srv"
2024-07-17T11:41:40.201+0000	try 'mongodump --help' for more information
Starting the cleanup task after taking a backup.
Cleanup task completed.
```

### Environment

Production

### Severity

High (Blocker to building or releasing)

### Version

selfhosted

alfonsoasenjo:
the issue starts on 1.29, previous versions working fine

PR #35162, opened 2024-07-24
fix: failing backup and restore appsmithctl

pratapaprasanna:
Fixes: #34996

### DIFF
--- a/deploy/docker/fs/opt/appsmith/utils/bin/utils.js
+++ b/deploy/docker/fs/opt/appsmith/utils/bin/utils.js
@@ -32,21 +32,22 @@ function start(apps) {
   console.log("Started " + appsStr);
 }
 
-
 function getDburl() {
+  let dbUrl = '';
   try {
     let env_array = fs.readFileSync(Constants.ENV_PATH, 'utf8').toString().split("\n");
-    for(i in env_array) {
-        if (env_array[i].startsWith("APPSMITH_MONGODB_URI") || env_array[i].startsWith("APPSMITH_DB_URL")) {
-          var dbUrl = env_array[i].toString().split("=")[1];
-        }
+    for (let i in env_array) {
+      if (env_array[i].startsWith("APPSMITH_MONGODB_URI") || env_array[i].startsWith("APPSMITH_DB_URL")) {
+        dbUrl = env_array[i].toString().split("=")[1];
+        break; // Break early when the desired line is found
+      }
     }
   } catch (err) {
-    console.error(err);
+    console.error("Error reading the environment file:", err);
   }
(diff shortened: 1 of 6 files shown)


deploy/docker/fs/opt/appsmith/utils/bin/utils.js
```diff
@@ -31,6 +32,26 @@ function start(apps) {
...
+        break; // Break early when the desired line is found
+      }
+    }
+  } catch (err) {
+    console.error("Error reading the environment file:", err);
+  }
+  let dbEnvUrl = process.env.APPSMITH_DB_URL || process.env.APPSMITH_MONGO_DB_URI;
+  if (dbEnvUrl && dbEnvUrl !== "undefined") {
+    dbUrl = dbEnvUrl;
+  }
```
vsvamsi1, line 51: Utility functions should generally be pure and env variables should not be a part of it. If you were to drive the utility function the env values as a parameter is okay, however i think you can move this function into a separate file and not be a part of the utility file.

deploy/docker/fs/opt/appsmith/utils/bin/utils.js
```diff
@@ -31,6 +32,26 @@ function start(apps) {
   console.log("Started " + appsStr);
 }
 
+function getDburl() {
+  let dbUrl = '';
```
vsvamsi1, line 36: I have a nitpick comment. 
It appears to me that dbUrl variable job is to be just eventually returned, i think we can remove this and directly the return the result.

vsvamsi1: An example 

```
function getDburl() {

  try {
    const env_array = fs.readFileSync(Constants.ENV_PATH, 'utf8').toString().split("\n");
    for (let i in env_array) {
      if (env_array[i].startsWith("APPSMITH_MONGODB_URI") || env_array[i].startsWith("APPSMITH_DB_URL")) {
       // return early when the desired line is found
       return  env_array[i].toString().split("=")[1];
        
      }
    }
  } catch (err) {
    console.error("Error reading the environment file:", err);
  }
}

  let dbEnvUrl = process.env.APPSMITH_DB_URL || process.env.APPSMITH_MONGO_DB_URI;
  if (dbEnvUrl && dbEnvUrl !== "undefined") {
   return dbEnvUrl;
  }

 return "";
}
```

deploy/docker/fs/opt/appsmith/utils/bin/utils.js
pratapaprasanna: Merging this for the time being because of the urgency. I will address this in the next release
thanks @vsvamsi1

PR #35323, opened 2024-07-31
fix: collapse params & query params in action selector if value is not changed

### LINKED ISSUE #32872: Collapse the params text area field in the action selector as it is an optional field.
The params section has been a source of confusion as seen in user tests, repeatedly. For an optional field and one that is not at all used -- it occupies prominent space in the action selector.  We could address this by hiding this section by default when there are no params configured. (This should be in-line with the behaviour of a JS function which does not have any params configured)

https://pasteboard.co/SzHNaENv54iU.png

AmanAgarwal041:
## Description
Problem statement:
The params section has been a source of confusion as seen in user tests, repeatedly. For an optional field and one that is not at all used -- it occupies prominent space in the action selector. We could address this by hiding this section by default when there are no params configured. (This should be in-line with the behaviour of a JS function which does not have any params configured)

Solution:
This PR solves this problem by collapsing the params section whenever we `select a query` to execute using action selector or whenever we are using `navigateTo` we collapse query params as well, this way we are not confusing new users by providing lots of options in the beginning and users can eventually discover this feature as when they require it.

Fixes #32872 
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags="@tag.JS, @tag.Datasource, @tag.Binding, @tag.Widget"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/10210743155>
> Commit: ad5b310ef29d4e86de826cdb441a7040e11f72f8
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=10210743155&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.JS, @tag.Datasource, @tag.Binding, @tag.Widget`
> Spec:
> <hr>Fri, 02 Aug 2024 07:26:10 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
	- Enhanced the Field component to support new field types and improved data presentation.
	- Introduced a collapsible UI for the TextView component, enhancing interactivity and user experience.

- **Improvements**
	- Added the ability to track value changes via a new `isValueChanged` property in the TextViewProps interface.
	- Updated element selectors for greater flexibility and robustness in UI interactions.

- **Bug Fixes**
	- Standardized behavior for detecting value changes across different field types, ensuring consistency in user interactions.
	- Improved test coverage for the TextView component to ensure proper handling of dynamic input conditions.

### DIFF
--- a/app/client/src/components/editorComponents/ActionCreator/Field/index.tsx
+++ b/app/client/src/components/editorComponents/ActionCreator/Field/index.tsx
@@ -161,8 +161,8 @@ export function Field(props: FieldProps) {
       });
       break;
     case FieldType.PARAMS_FIELD:
-      viewElement = (view as (props: SelectorViewProps) => JSX.Element)({
-        options: options as TreeDropdownOption[],
+      viewElement = (view as (props: TextViewProps) => JSX.Element)({
+        exampleText: exampleText,
         label: label,
         get: (value: string) => getterFunction(value, props.field.position),
         set: (value: string | DropdownOption) => {
@@ -174,7 +174,9 @@ export function Field(props: FieldProps) {
           props.onValueChange(finalValueToSet, false);
         },
         value: value,
-        defaultText: defaultText,
+        isValueChanged: (value: string) => {
+          return value !== getterFunction("");
+        },
       });
       break;
     case FieldType.KEY_VALUE_FIELD:
@@ -190,12 +192,29 @@ export function Field(props: FieldProps) {
         defaultText: defaultText,
       });
       break;
+    case FieldType.QUERY_PARAMS_FIELD:
+      viewElement = (view as (props: TextViewProps) => JSX.Element)({
+        label: label,
+        toolTip: toolTip,
+        exampleText: exampleText,
+        get: getterFunction,
+        set: (value: string | DropdownOption, isUpdatedViaKeyboard = false) => {
+          const finalValueToSet = fieldConfig.setter(value, props.value);
+          props.onValueChange(finalValueToSet, isUpdatedViaKeyboard, true);
+        },
+        value: value,
+        additionalAutoComplete: props.additionalAutoComplete,
+        dataTreePath: props.dataTreePath,
+        isValueChanged: (value: string) => {
+          return value !== getterFunction("");
+        },
+      });
+      break;
     case FieldType.ALERT_TEXT_FIELD:
     case FieldType.URL_FIELD:
     case FieldType.KEY_TEXT_FIELD_STORE_VALUE:
     case FieldType.KEY_TEXT_FIELD_REMOVE_VALUE:
     case FieldType.VALUE_TEXT_FIELD:
-    case FieldType.QUERY_PARAMS_FIELD:
     case FieldType.DOWNLOAD_DATA_FIELD:
     case FieldType.DOWNLOAD_FILE_NAME_FIELD:
     case FieldType.COPY_TEXT_FIELD:
--- a/app/client/src/components/editorComponents/ActionCreator/Field/index.tsx
+++ b/app/client/src/components/editorComponents/ActionCreator/Field/index.tsx
@@ -206,6 +206,8 @@ export function Field(props: FieldProps) {
         additionalAutoComplete: props.additionalAutoComplete,
         dataTreePath: props.dataTreePath,
         isValueChanged: (value: string) => {
+          // This function checks whether the param value is changed from the default value.
+          // getterFunction("") -> passing empty string will return the default value
           return value !== getterFunction("");
         },
       });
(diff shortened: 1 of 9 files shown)


sagar-qa007:
@AmanAgarwal041 can we also add unit test here if possible?

app/client/src/components/editorComponents/ActionCreator/Field/index.tsx
```diff
@@ -190,12 +192,29 @@ export function Field(props: FieldProps) {
...
+        get: getterFunction,
+        set: (value: string | DropdownOption, isUpdatedViaKeyboard = false) => {
+          const finalValueToSet = fieldConfig.setter(value, props.value);
+          props.onValueChange(finalValueToSet, isUpdatedViaKeyboard, true);
+        },
+        value: value,
+        additionalAutoComplete: props.additionalAutoComplete,
+        dataTreePath: props.dataTreePath,
+        isValueChanged: (value: string) => {
+          return value !== getterFunction("");
```
sneha122, line 209: @AmanAgarwal041 What exactly are we doing here in isValueChanged? Can we add appropriate comments of this new property? It's unclear at this point what getterFunction is doing and why we are passing empty value to it here

sneha122 (changes requested):
@AmanAgarwal041 Let's add jest test cases for this

sneha122:
@rishabhrathod01 I have reviewed the PR, still adding you as a reviewer for this PR as you have more context around the action selectors and params.

app/client/src/components/editorComponents/ActionCreator/Field/index.tsx
```diff
@@ -190,12 +192,29 @@ export function Field(props: FieldProps) {
         defaultText: defaultText,
       });
       break;
+    case FieldType.QUERY_PARAMS_FIELD:
```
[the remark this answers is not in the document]
AmanAgarwal041: It was already there. It is not created newly. I just had to create `isValueChanged` function for `QUERY_PARAMS_FIELD` specifically that is why I added a specific case. @sneha122

app/client/src/components/editorComponents/ActionCreator/Field/index.tsx
AmanAgarwal041: `isValueChanged` as the name suggests checks whether the value is changed from the default value. But surely add the comment for getterfunction. @sneha122

AmanAgarwal041:
> @AmanAgarwal041 Let's add jest test cases for this

@sneha122 The estimated for this issue is wrong considering we want to add a jest case for the same. As the implementation that was already suggested did not work properly. Do you think that jest test case is important right now ?

sneha122:
> > @AmanAgarwal041 Let's add jest test cases for this
> 
> @sneha122 The estimated for this issue is wrong considering we want to add a jest case for the same. As the implementation that was already suggested did not work properly. Do you think that jest test case is important right now ?

@AmanAgarwal041 We can update the estimate for this and add appropriate comment for the same, but we should definitely add jest case for this.

app/client/src/components/editorComponents/ActionCreator/Field/index.tsx
```diff
@@ -190,12 +192,29 @@ export function Field(props: FieldProps) {
...
+        exampleText: exampleText,
+        get: getterFunction,
+        set: (value: string | DropdownOption, isUpdatedViaKeyboard = false) => {
+          const finalValueToSet = fieldConfig.setter(value, props.value);
+          props.onValueChange(finalValueToSet, isUpdatedViaKeyboard, true);
+        },
+        value: value,
+        additionalAutoComplete: props.additionalAutoComplete,
+        dataTreePath: props.dataTreePath,
+        isValueChanged: (value: string) => {
```
rishabhrathod01, line 208: - Do we need the `isValueChange` property? I believe the component already has the value and here we just need to check if the value is not empty. Do you think, this could be done even without the isValueChange method?

rishabhrathod01: @AmanAgarwal041 can you help me understand with an example what is default value and why we need to track the change of default value? 

- What I understand is that the component will receive the params binding value and in those cases, we want the default state of the collapsible input to be open else the default state will be closed.

AmanAgarwal041: @rishabhrathod01 We want the accordion to be expanded only when the user has changed the input text. That is why this function works to track whether the value has been changed from default value or not.

rishabhrathod01 (approved):
LGTM

sneha122 (approved):

PR #35468, opened 2024-08-07
chore: Postgres transaction POC

abhvsn:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!WARNING]
> Invalid tags. Please use `/ok-to-test tags="@tag.All"` or `/test all` in the PR body to run all tests.
> [Tags documentation](https://www.notion.so/appsmith/7c0fc64d4efb4afebf53348cd6252918).
> [List of valid tags](https://github.com/appsmithorg/appsmith/blob/release/app/client/cypress/tags.js).
> <hr>Mon, 12 Aug 2024 14:55:44 UTC

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
@@ -11,6 +11,7 @@ import com.appsmith.server.domains.NewAction;
 import com.appsmith.server.domains.NewPage;
 import com.appsmith.server.domains.Theme;
 import com.appsmith.server.dtos.ActionCollectionDTO;
+import com.appsmith.server.helpers.ce.bridge.BridgeUpdate;
 import com.appsmith.server.repositories.ActionCollectionRepository;
 import com.appsmith.server.repositories.ApplicationRepository;
 import com.appsmith.server.repositories.AppsmithRepository;
@@ -95,6 +96,16 @@ public class TransactionAspect {
                     // Check if it's a write operation
                     boolean isWriteOp = isWriteOp((MethodSignature) joinPoint.getSignature());
 
+                    Object[] args = getArgs(joinPoint.getArgs());
+                    boolean isNewWriteOp = isWriteOp(args, (MethodSignature) joinPoint.getSignature());
+
+                    // TODO - remove this once the values are consistent with the new method to check write operation
+                    if (isNewWriteOp != isWriteOp) {
+                        log.error(
+                                "Mismatch in write operation detection. isNewWriteOp: {}, isWriteOp: {}",
+                                isNewWriteOp,
+                                isWriteOp);
+                    }
                     // If the operation is read, operation
                     // 1. Get the object from DB
                     // 2. Check if the object is already present in the context:
@@ -114,17 +125,13 @@ public class TransactionAspect {
                         //      - If end up in switchIfEmpty means no object is present in the DB and should mark this
                         //         as a new object and store the object in DB
                         //      - If object is present in the DB, then store the initial state in the context
-                        // 3. Return the object
-
-                        BaseDomain domain = (BaseDomain) joinPoint.getArgs()[0];
-                        if (isByIdMethod((MethodSignature) joinPoint.getSignature())) {
-                            BaseDomain domainObject = domain;
-                            String updateById = (String) joinPoint.getArgs()[1];
-                            domainObject.setId(updateById);
-                            addEntityToContextMap(transactionContext, domainObject);
+                        if ((args[2] != null && isUpdateOp(args[2]))
+                                || (isArchiveOp((MethodSignature) joinPoint.getSignature()))) {
+                            addEntityToMapUpdateAndArchiveOp(transactionContext, (String) args[1]);
                             return ((Mono<?>) joinPoint.proceed(joinPoint.getArgs()));
                         }
 
+                        BaseDomain domain = (BaseDomain) args[0];
                         if (domain.getId() == null) {
                             domain.setId(generateId());
                         }
@@ -145,26 +152,26 @@ public class TransactionAspect {
                 try {
                     if (!context.isEmpty() && context.hasKey("transactionContext")) {
                         Map<String, DBOps> transactionContext = context.get("transactionContext");
-                        boolean isWriteOp = isWriteOp((MethodSignature) joinPoint.getSignature());
+                        Object[] args = getArgs(joinPoint.getArgs());
+                        boolean isWriteOp = isWriteOp(args, (MethodSignature) joinPoint.getSignature());
 
                         Flux flux = (Flux<?>) joinPoint.proceed(joinPoint.getArgs());
 
                         if (!isWriteOp) {
                             return flux.map(obj -> addNonWriteEntityToContextMap(transactionContext, obj));
                         } else {
-                            BaseDomain domain = (BaseDomain) joinPoint.getArgs()[0];
-                            if (isByIdMethod((MethodSignature) joinPoint.getSignature())) {
-                                BaseDomain domainObject = domain;
-                                String updateById = (String) joinPoint.getArgs()[1];
-                                domainObject.setId(updateById);
-                                addEntityToContextMap(transactionContext, domainObject);
+                            if ((args[2] != null && isUpdateOp(args[2]))
+                                    || (isArchiveOp((MethodSignature) joinPoint.getSignature()))) {
+                                addEntityToMapUpdateAndArchiveOp(transactionContext, (String) args[1]);
                                 return flux;
                             }
 
+                            BaseDomain domain = (BaseDomain) args[0];
                             if (domain.getId() == null) {
                                 domain.setId(generateId());
                             }
                             addEntityToContextMap(transactionContext, domain);
+
                             return flux;
                         }
                     }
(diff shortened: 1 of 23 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,274 @@
...
+
+    private final ApplicationRepository applicationRepository;
+
+    private final NewPageRepository newPageRepository;
+
+    private final NewActionRepository newActionRepository;
+
+    private final ActionCollectionRepository actionCollectionRepository;
+
+    private Map<Class<?>, AppsmithRepository<?>> repoByEntityClass;
```
abhvsn, line 64: Let's create a separate class something like RepositoryFactory to keep the changes within this class limited to aspect only.

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,274 @@
...
+    public Object handleTransaction(ProceedingJoinPoint joinPoint) throws Throwable {
+
+        Class<?> returnType =
+                ((MethodSignature) joinPoint.getSignature()).getMethod().getReturnType();
+
+        if (Mono.class.isAssignableFrom(returnType)) {
+
+            return Mono.deferContextual(context -> {
+                try {
+                    // transaction context is not maintained
```
abhvsn, line 90:
```suggestion
                    // Check if the transaction context is available, if not execute the DB call and return
```

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,274 @@
...
+                    if (context.isEmpty() || !context.hasKey("transactionContext")) {
+                        return (Mono<?>) joinPoint.proceed(joinPoint.getArgs());
+                    }
+                    Map<String, DBOps> transactionContext = context.get("transactionContext");
```
abhvsn, line 94: Let's pull out this constant string at the class level and refer the same in other flows.

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,274 @@
...
+                    // Check if it's a write operation
+                    boolean isWriteOp = isWriteOp((MethodSignature) joinPoint.getSignature());
+
+                    // If the operation is read, operation
+                    // 1. Get the object from DB
+                    // 2. Check if the object is already present in the context:
+                    //      - If not, store the object in the context and return
+                    //      - If yes, then return the object as is. We want to just maintain the initial state of the
+                    // object as we are concerned with the objects initial state before the transaction started
```
abhvsn, line 103: Let's also add the comment why only initial state is required here.

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,274 @@
...
+                    if (!isWriteOp) {
+                        return ((Mono<?>) joinPoint.proceed(joinPoint.getArgs()))
+                                .map(obj -> addNonWriteEntityToContextMap(transactionContext, obj));
```
abhvsn, line 106:
```suggestion
                                .map(obj -> updateContextMapWithReadOperation(transactionContext, obj));
```

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,274 @@
...
+                        // 2. Check if the object is already present in the context
+                        //    a. If yes, execute the DB call
+                        //    b. If not get the initial state of the object from DB using findById method on the
+                        // repository class
+                        //      - If end up in switchIfEmpty means no object is present in the DB and should mark this
+                        //         as a new object and store the object in DB
+                        //      - If object is present in the DB, then store the initial state in the context
+                        // 3. Return the object
+
+                        BaseDomain domain = (BaseDomain) joinPoint.getArgs()[0];
```
abhvsn, line 119: We should never end up in a state where args are not present but should we add the null check just in case?

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,274 @@
...
+                        if (isByIdMethod((MethodSignature) joinPoint.getSignature())) {
+                            BaseDomain domainObject = domain;
+                            String updateById = (String) joinPoint.getArgs()[1];
```
abhvsn, line 122: We may not have control over here where are we defining the id param in method definition. Can we come up with better approach like going through params and then finding out the arg at that location can be 1 approach. Lets also update the name of the variable to `id`.

abhvsn: Also will the `id` be present at index 1 or 0?

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
abhvsn: Let's update the implementation here to make sure the arg is instance of BaseDomain before blindly assigning it.

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,274 @@
...
+
+                            if (domain.getId() == null) {
+                                domain.setId(generateId());
+                            }
+                            addEntityToContextMap(transactionContext, domain);
+                            return flux;
+                        }
+                    }
+
+                    return (Flux<?>) joinPoint.proceed(joinPoint.getArgs());
```
abhvsn, line 172: Similar comments as mentioned in the Mono block.

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,274 @@
...
+
+                } catch (Throwable e) {
+                    log.error(
+                            "Error occurred while adding the user context to the permission object when invoking function {}",
```
abhvsn, line 176: Let's change the error, to more relevant one.

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,274 @@
...
+                            joinPoint.getSignature().getName(),
+                            e);
+                    return Flux.error(e);
+                }
+            });
+        }
+        return joinPoint.proceed(joinPoint.getArgs());
+    }
+
+    private String generateId() {
```
abhvsn, line 186: Let's make the method in BaseCake as public static and refer that here to follow `DRY` principal.

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,274 @@
...
+        return UUID.randomUUID().toString();
+    }
+
+    private Object addNonWriteEntityToContextMap(Map<String, DBOps> transactionContext, Object obj) {
+        if (obj instanceof BaseDomain) {
+            DBOps dbOps = new DBOps();
+            dbOps.setEntity(obj);
+            dbOps.setModified(true);
```
abhvsn, line 194: Actually this is read operation so we should not set the modified field, right?

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,274 @@
...
+        return methodName.contains("save")
+                || methodName.contains("update")
+                || methodName.contains("delete")
+                || methodName.contains("insert")
+                || methodName.contains("archive");
+    }
+
+    private boolean isByIdMethod(MethodSignature signature) {
+        String methodName = signature.getMethod().getName();
+        return methodName.contains("updateById") || methodName.contains("archiveById");
```
abhvsn, line 239: Same as above let's improve this instead of relying on method name.

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,274 @@
...
+        return obj;
+    }
+
+    private boolean isWriteOp(MethodSignature signature) {
+        String methodName = signature.getMethod().getName();
+        return methodName.contains("save")
+                || methodName.contains("update")
+                || methodName.contains("delete")
+                || methodName.contains("insert")
+                || methodName.contains("archive");
```
abhvsn, line 234: Isn't there a better way to judge this based on args instead of relying on method name? Basically we can look for all the args if that contains object which is instance of BaseDomain or BridgeUpdate etc and mark these accordingly.

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,274 @@
...
+                transactionContext.put(id, dbOps);
+            }
+            return obj;
+        }
+        return obj;
+    }
+
+    private Object addEntityToContextMap(Map<String, DBOps> transactionContext, Object obj) {
+        AppsmithRepository<?> repository = getDomainClassFromObject(obj);
+        if (repository == null) {
```
abhvsn, line 206: We will end up in this scenario if are not able to figure out which entity we are updating.

abhvsn (commented):
@AnaghHegde it feels like we are targeting only the methods which are part of import flow. Can we please make this generic especially the wrote operations to scale well across other flows as well?

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
AnaghHegde: Yes that is correct

PR #35476, opened 2024-08-07
chore: Ability to run PR automations every time PR updates

hetunandu:
## Description

By changing the workflow trigger and the way to check for correct labels, we are trying to allow the PR integrations tests to run every time the PR is updated. That means, if you push changes, and the PR already has ok-to-test label, it will rerun the tests based on new changes

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/10296910575>
> Commit: c59583ba28c8cdff35a12422b8c73838c6efdc85
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=10296910575&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Sanity`
> Spec:
> <hr>Thu, 08 Aug 2024 06:32:48 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
	- Introduced a new job, `checkTestLabel`, to verify the presence of the "ok-to-test" label on pull requests.
	- Added a new job, `mark-stale`, to mark the Cypress status as stale when the label is absent.
- **Improvements**
	- Updated the `parse-tags` job to run only after the `checkTestLabel` job, enhancing workflow control.
	- Modified the conditions for executing the `parse-tags` job based on the output of the `checkTestLabel` job.
- **Style**
	- Minor formatting adjustments for job dependencies in the workflow.

### DIFF
--- a/.github/workflows/pr-automation.yml
+++ b/.github/workflows/pr-automation.yml
@@ -35,7 +35,7 @@ jobs:
             }
 
   mark-stale:
-    needs: [ check-label ]
+    needs: [ checkTestLabel ]
     if: needs.checkTestLabel.outputs.ok_to_test == 'true'
     runs-on: ubuntu-latest
     steps:
@@ -53,7 +53,7 @@ jobs:
             }
 
   parse-tags:
-    needs: [ check-label ]
+    needs: [ checkTestLabel ]
     if: needs.checkTestLabel.outputs.ok_to_test == 'true'
     runs-on: ubuntu-latest
     permissions:
--- a/.github/workflows/pr-automation.yml
+++ b/.github/workflows/pr-automation.yml
@@ -13,8 +13,6 @@ jobs:
     outputs:
       ok_to_test: ${{ steps.checkLabel.outputs.ok_to_test }}
     steps:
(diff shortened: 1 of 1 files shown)


nidhi-nair:
Isn't this wasteful of CI resources @hetunandu ? We want to encourage developers to commit to remote as often as possible, even with broken builds. Why are we opting to occupy our runners unnecessarily? Perhaps we can think of a CLI option to mock the label removal and addition instead?

hetunandu:
@nidhi-nair this PR would only reduce the steps to test in my opinion. It solves these problems

- It is still dependant on the dev input. Only when the label exists, it would trigger the cypress runs
- It will automatically trigger the tests on new pushes. This is something that is anyways done manually by removing and adding labels. 
- Correctly marks the PR to be not ready to merge by marking the Cypress test results stale.

I am not sure how a CLI process to remove and add back the label would be any better?

KelvinOm (approved):

PR #35614, opened 2024-08-12
chore: Add tenant cache eviction for the existing policies to policyMap migration

abhvsn:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration059PolicySetToPolicyMap.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration059PolicySetToPolicyMap.java
@@ -1,5 +1,7 @@
 package com.appsmith.server.migrations.db.ce;
 
+import com.appsmith.server.domains.Tenant;
+import com.appsmith.server.repositories.CacheableRepositoryHelper;
 import com.mongodb.client.result.UpdateResult;
 import io.mongock.api.annotations.ChangeUnit;
 import io.mongock.api.annotations.Execution;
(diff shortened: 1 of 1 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration059PolicySetToPolicyMap.java
```diff
@@ -80,6 +85,13 @@ public void execute() {
                     return Mono.error(error);
                 })
                 .block();
+
+        // Evict the default tenant from the cache to ensure that the updated tenant object is fetched from the database
+        Query tenantQuery = new Query();
+        tenantQuery.addCriteria(where(Tenant.Fields.slug).is(DEFAULT));
+        Tenant defaultTenant = mongoTemplate.findOne(tenantQuery, Tenant.class).block();
```
AnaghHegde, line 92: Why is this a block call?

AnaghHegde (commented):
Minor comment rest look good to me

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration059PolicySetToPolicyMap.java
abhvsn: @AnaghHegde here the MongoTemplate that is injected is a reactive one, in other cases we inject the non-reactive template. The blocking is required just to make sure the eviction happens after the migration is completed. This can be done using .then but thought it won't make any difference as the earlier call is also blocking in nature.

AnaghHegde (approved):

PR #35645, opened 2024-08-13
feat: added properties for executeAction mixpanel event (#35291)

### LINKED ISSUE #35291: [Task]: Update the event ExecuteAction to include new properties about bindings
### SubTasks

- [ ] Include the `request` parameter for all execute action events (currently it isn't firing for anonymous users)
- [ ] Add a new property which shows all the parameters. This should be an array for instance:  {{Table1.pageSize}}, {{Table1.pageOffset}},{{Select1.selectedOptionValue}}
- [ ] Have a new property which shows the raw query without the substitutions. For instance: `SELECT * FROM TABLE WHERE country =  {{Select1.selectedOptionValue}}`, along with request, which currently returns: `SELECT * FROM TABLE WHERE country = "China"`

### Why?
If we look at the queries users are writing, it can be a great indicator of what progress they are making in their app. For instance, if they are using the pageSize and pageOffset property in a select query, they are doing server side pagination. This will allow us to measure:
(a) How many users are progressing through their app building journey
(b) Are we noticing patterns of what order users are building out this functionality
(c) What stage do users get to before they get stuck, or lost

All these tracking can be added to the same `Execute Action` event already present in the backend.

NilanshBansal:
## Description
This PR adds new properties to the `Execute Action` mixpanel event so that we can measure the following:
- How many users are progressing through their app building journey
- Are we noticing patterns of what order users are building out this functionality
- What stage do users get to before they get stuck, or lost

New Properties added are: 
- ACTION_EXECUTION_REQUEST_PARAMS_VALUE_MAP
- ACTION_EXECUTION_INVERT_PARAMETER_MAP
- ACTION_EXECUTION_RAW_QUERY

Fixes #35291 

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/10372829069>
> Commit: e8d8a02d38c9306eebef201a84a3570db9cd572d
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=10372829069&attempt=2" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Sanity`
> Spec:
> <hr>Tue, 13 Aug 2024 16:11:44 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
	- Introduced new constants for action execution parameters, enhancing clarity and maintainability.
	- Added the capability to log raw action configurations in analytics events for improved tracking of action executions.

- **Improvements**
	- Enhanced analytics data handling by conditionally managing properties based on the hosting environment, ensuring better data privacy for self-hosted applications.
	- Expanded functionality of action execution by integrating new configuration options, improving overall performance.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCEImpl.java
@@ -208,9 +208,6 @@ public class AnalyticsServiceCEImpl implements AnalyticsServiceCE {
             final String hashedUserId = hash(userId);
             // Remove params map and request key properties, if it's self-hosted as it contains user's evaluated params
             analyticsProperties.remove("request");
-            if (analyticsProperties.containsKey(FieldName.ACTION_EXECUTION_REQUEST_PARAMS_VALUE_MAP)) {
-                analyticsProperties.remove(FieldName.ACTION_EXECUTION_REQUEST_PARAMS_VALUE_MAP);
-            }
             for (final Map.Entry<String, Object> entry : analyticsProperties.entrySet()) {
                 if (userId.equals(entry.getValue())) {
                     analyticsProperties.put(entry.getKey(), hashedUserId);
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCEImpl.java
@@ -206,7 +206,7 @@ public class AnalyticsServiceCEImpl implements AnalyticsServiceCE {
         // Hash usernames at all places for self-hosted instance
         if (shouldHashUserId(event, userId, hashUserId, commonConfig.isCloudHosting())) {
             final String hashedUserId = hash(userId);
-            // Remove params map and request key properties, if it's self-hosted as it contains user's evaluated params
+            // Remove request key, if it's self-hosted as it contains user's evaluated params
             analyticsProperties.remove("request");
             for (final Map.Entry<String, Object> entry : analyticsProperties.entrySet()) {
                 if (userId.equals(entry.getValue())) {
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCEImpl.java
@@ -207,6 +207,10 @@ public class AnalyticsServiceCEImpl implements AnalyticsServiceCE {
         if (shouldHashUserId(event, userId, hashUserId, commonConfig.isCloudHosting())) {
             final String hashedUserId = hash(userId);
             analyticsProperties.remove("request");
+            // Remove params map key property if it's self-hosted
+            if (analyticsProperties.containsKey(FieldName.ACTION_EXECUTION_REQUEST_PARAMS_VALUE_MAP)) {
+                analyticsProperties.remove(FieldName.ACTION_EXECUTION_REQUEST_PARAMS_VALUE_MAP);
+            }
             for (final Map.Entry<String, Object> entry : analyticsProperties.entrySet()) {
                 if (userId.equals(entry.getValue())) {
                     analyticsProperties.put(entry.getKey(), hashedUserId);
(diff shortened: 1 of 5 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCEImpl.java
```diff
@@ -207,6 +207,10 @@ public Mono<Void> sendEvent(String event, String userId, Map<String, ?> properti
         if (shouldHashUserId(event, userId, hashUserId, commonConfig.isCloudHosting())) {
             final String hashedUserId = hash(userId);
             analyticsProperties.remove("request");
+            // Remove params map key property if it's self-hosted
```
sneha122, line 210: @NilanshBansal Why are we removing this for self hosted? Can we please let the comment also reflect why we are doing this?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCEImpl.java
```diff
@@ -207,6 +207,10 @@ public Mono<Void> sendEvent(String event, String userId, Map<String, ?> properti
...
+            if (analyticsProperties.containsKey(FieldName.ACTION_EXECUTION_REQUEST_PARAMS_VALUE_MAP)) {
```
sneha122, line 211: @NilanshBansal sendEvent is generic event, can we please remove this property in action execution specific functions?

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImpl.java
```diff
@@ -789,6 +789,7 @@ protected Mono<ActionExecutionResult> getActionExecutionResult(
                     final DatasourceStorage datasourceStorage = tuple.getT2();
                     final PluginExecutor pluginExecutor = tuple.getT3();
                     final Plugin plugin = tuple.getT4();
+                    final String rawQuery = actionDTO.getActionConfiguration().getBody();
```
sneha122, line 792: For UQI queries like gsheet mongo, the body may be empty and the actual query data may be getting saved to bodyForm, in that case we don't want that query data in the event?

NilanshBansal: @sneha122 Thanks for highlighting this! I am checking

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImpl.java
```diff
@@ -807,6 +821,7 @@ protected Mono<ActionExecutionResult> getActionExecutionResult(
                                             executeActionDTO, actionDTO, datasourceStorage, plugin, pluginExecutor)
                                     .timeout(Duration.ofMillis(timeoutDuration)));
 
+                    ActionConfiguration finalRawActionConfiguration = rawActionConfiguration;
```
sneha122, line 824: Why have we created this finalRawActionConfiguration? Why can't we simply pass rawActionConfiguration? Also why is the deep copying necessary? As far as I can see in the code, we are only passing this data in sendEvent function, we are not manipulating it inside

NilanshBansal: @sneha122 we have to add `finalRawActionConfiguration` as the variable to be used in lambda expression should be final or effectively final. 
SCR-20240813-sav

We have deep copied the object as the actionDTO was getting sent to `[verifyDatasourceAndMakeRequest](https://github.com/appsmithorg/appsmith/blob/496cecf7c6bba83b9f5dcadc07759998c0b09100/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImpl.java#L806)()` and is then sent as parameter to other nested functions leading to the shallow copy getting updated with the executed query and not the raw binded query.

sneha122: Got it, makes sense! Thank you

sneha122 (dismissed):

Issue #35366, opened 2024-08-02
[Task]: Transaction Management for ImportExport


AnaghHegde:
### SubTasks

- All the sub flows for entities
- Delete flow for pages `deleteUnpublishedPageByBranchAndDefaultPageId`
- Git service:
    - `updateProtectedBranches`

PR #35707, opened 2024-08-15
chore: Transaction support for PG via Spring AOP

AnaghHegde:
## Description
We are utilising spring AOP Around advice to handle the Transaction in pg. You can read more about the Spring AOP here - https://docs.spring.io/spring-framework/docs/4.3.15.RELEASE/spring-framework-reference/html/aop.html

We are using the `Around` advice for our use case. It has the opportunity to do work both before and after the method executes, and to determine when, how, and even if, the method actually gets to execute at all. Using this property we are intercepting all the calls done to the Cake/Repo class in the ImportExport service. We also use the reactive thread context to avoid processing every single call made to the Cake class. With this we can capture the state of the entities that are to be modified/updated/created before the actual call is executed and store this data for future processing when something goes wrong in the Transaction flow. This is done via `TransactionAspect` class. 

One has to worry about the state of the db only when something fails inside the Transaction flow. So when something fails , the above captured state is used to restore the previous state of the entities. This is done via the `TransactionHandler` class.

Around advice is declared using the `@Around` annotation. The first parameter of the advice method must be of type `ProceedingJoinPoint`. Within the body of the advice, calling `proceed()` on the `ProceedingJoinPoint` causes the underlying method to execute. The proceed method may also be called passing in an Object[] - the values in the array will be used as the arguments to the method execution when it proceeds.

A sample example on how the method interception rules are used with `@Around` annotation. 
`execution(* com.xyz.service.AccountService.*(..))` this means the execution of any method defined by the AccountService interface will be intercepted.

The current implementation of `@Cache` annotation also uses the Spring AOP Around advice. 

Fixes https://github.com/appsmithorg/appsmith/issues/35366

## Automation

/ok-to-test tags="@tag.Git, @tag.ImportExport, @tag.Templates"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/10458709051>
> Commit: 3fd921d87a2ee124b0125202c32abd840259d0f7
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=10458709051&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Git, @tag.ImportExport, @tag.Templates`
> Spec:
> <hr>Mon, 19 Aug 2024 18:42:11 UTC

## Communication
- [ ] Yes
- [ ] No


app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/TransactionHandler.java
```diff
@@ -0,0 +1,226 @@
...
+                                archiveAction(action.getId());
+                            } else {
+                                saveActionToDb(action);
+                            }
+                        }
+                    });
+                    return true;
+                })
+                .then();
+    }
```
abhvsn, line 225: Looks like a lot of code is repeated here, can we replicate what we have done for `getById` above?

abhvsn (commented):
Few more comments.

@AnaghHegde if it helps we can also note the initial time before the flow is triggers say t1:
1. If the object's `createdAt` >= t1 this we can be marked as created. 
2. If `createdAt < t1 and updatedAt > t1` these can be marked as updated but the only caveate here will be to keep the copy of the object before the update operation is executed.

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,274 @@
...
+            dbOps.setModified(true);
+        } else {
+            dbOps.setEntity(obj);
+            dbOps.setNew(true);
+        }
+        transactionContext.put(getObjectId(dbOps), dbOps);
+    }
+
+    private boolean isWriteOp(Object[] args, MethodSignature signature) {
+        // Special case like findCustomJsLib accepting the BaseDomain object as parameter instead of the UUID,
```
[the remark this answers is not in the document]
AnaghHegde: Yeah even i was thinking about changing the method name from release. Since this is going to take sometime, i will remove this change once the refactor is done on the release.

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,274 @@
...
+        return joinPoint.proceed(joinPoint.getArgs());
+    }
+
+    private Object[] getArgs(Object[] args) {
+        // To store the baseDomain and the id of the object and BridgeUpdate
+        // when the BaseDomain is not present, in the case of updateById methods
+        Object[] newArgs = new Object[3];
+        for (Object arg : args) {
+            if (arg instanceof BaseDomain domain) {
+                newArgs[0] = domain;
```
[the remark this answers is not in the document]
AnaghHegde: My understanding is that Updates are BridgeUpdate and not BaseDomain.

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/RepositoryFactoryImpl.java
```diff
@@ -0,0 +1,40 @@
+package com.appsmith.server.helpers;
```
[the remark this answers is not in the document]
AnaghHegde: No, i think this is a utility/helper function.

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,274 @@
...
+        return newArgs;
+    }
+
+    private String generateId() {
+        return UUID.randomUUID().toString();
+    }
+
+    private boolean isUUIDString(String id) {
+        try {
+            UUID.fromString(id);
```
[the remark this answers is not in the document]
AnaghHegde: I guess it will fail but i can confirm only after testing on the adaptable image.

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,314 @@
...
+            return false;
+        }
+    }
+
+    // To Support the mongo _ids as well
+    private boolean isObjectIdString(String id) {
+        return OBJECTID_PATTERN.matcher(id).matches();
+    }
+
+    private void addEntityToMapUpdateAndArchiveOp(Map<String, DBOps> transactionContext, EntityData entityData) {
```
abhvsn, line 188: Correct me but this method is only updating the `entityData` can we please move this to `getArgs()` instead so that only 1 method will be owner of updating this object?

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,314 @@
...
+            if (!transactionContext.containsKey(id)) {
+                transactionContext.put(id, dbOps);
+            }
+            return obj;
+        }
+        return obj;
+    }
+
+    private void addEntityToContextMap(Map<String, DBOps> transactionContext, Object obj, EntityData entityData) {
+        AppsmithRepository<?> repository = getDomainClassFromObject(obj);
```
abhvsn, line 213: This won't work in case of update via BridgeUpdate object. As per current implementation we are just adding the id in the `BaseDomain` so it will end up in null value always. As suggested above can we rely on `getArgs` to fill up all these details in entityObject.

abhvsn: Also to get the class you can rely on target, something like below:
```
        Object target = joinPoint.getTarget();
        Class<?> targetClass = target.getClass();

        if (BaseCake.class.isAssignableFrom(targetClass)) {
            Type genericSuperclass = targetClass.getGenericSuperclass();
            if (genericSuperclass instanceof ParameterizedType) {
                ParameterizedType parameterizedType = (ParameterizedType) genericSuperclass;
                Type[] actualTypeArguments = parameterizedType.getActualTypeArguments();
                if (actualTypeArguments.length > 0) {
                    Class<?> entityClass = (Class<?>) actualTypeArguments[0];
                }
            }
        }
```

Here entityClass will be something like Tenant.class etc.
Screenshot 2024-08-16 at 7 15 28 PM

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,314 @@
...
+                        // 1. Extract the id of the object
+                        // 2. Check if the object is already present in the context
+                        //    a. If yes, execute the DB call
+                        //    b. If not get the initial state of the object from DB using findById method on the
+                        // repository class
+                        //      - If end up in switchIfEmpty means no object is present in the DB and should mark this
+                        //         as a new object and store the object in DB
+                        //      - If object is present in the DB, then store the initial state in the context
+                        if (isArchiveOp((MethodSignature) joinPoint.getSignature())) {
+                            addEntityToMapUpdateAndArchiveOp(transactionContext, entityData);
```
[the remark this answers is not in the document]
AnaghHegde: After the below comment this is not required anymore.

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,291 @@
...
+                            joinPoint.getSignature().getName(),
+                            e);
+                    return Flux.error(e);
+                }
+            });
+        }
+        return joinPoint.proceed(joinPoint.getArgs());
+    }
+
+    private EntityData getArgs(Object[] args) {
```
abhvsn, line 138:
```suggestion
    private EntityData getEntityData(Object[] args) {
```

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,291 @@
...
+
+        if (entityData.getBaseDomain() != null) {
+            // When the update method is called with id not present in baseDomain
+            if (entityData.getBaseDomain().getId() == null && entityData.getId() != null) {
+                entityData.getBaseDomain().setId(entityData.getId());
+            } else if (entityData.getBaseDomain().getId() == null && entityData.getId() == null) {
+                // When the object is new and not present in the DB, we need to generate the id
+                entityData.getBaseDomain().setId(generateId());
+                entityData.setId(entityData.getBaseDomain().getId());
+            }
```
abhvsn, line 161: What about `baseDomain.getId() != null && entityData.getId() == null`?

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,291 @@
...
+        } else if (entityData.getUpdate() != null) {
+            // When the updateById is called with the BridgeUpdate object
+            entityData.setBaseDomain(new BaseDomain() {});
+            entityData.getBaseDomain().setId(entityData.getId());
+        } else if (entityData.getId() != null) {
+            // When the id is passed as a parameter to the method for archiveById or deleteById
+            entityData.setBaseDomain(new BaseDomain() {});
+            entityData.getBaseDomain().setId(entityData.getId());
+        }
```
abhvsn, line 170: Considering the statements in the else-if block identical can we merge this? Also make sure to log errors when the we are not able to extract the id in any situation.

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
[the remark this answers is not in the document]
abhvsn: @AnaghHegde You can test this with MongoDB dump data from your local.

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,291 @@
...
+            if (entityData.getBaseDomain().getId() == null && entityData.getId() != null) {
+                entityData.getBaseDomain().setId(entityData.getId());
+            } else if (entityData.getBaseDomain().getId() == null && entityData.getId() == null) {
+                // When the object is new and not present in the DB, we need to generate the id
+                entityData.getBaseDomain().setId(generateId());
+                entityData.setId(entityData.getBaseDomain().getId());
+            }
+        } else if (entityData.getUpdate() != null) {
+            // When the updateById is called with the BridgeUpdate object
+            entityData.setBaseDomain(new BaseDomain() {});
```
abhvsn, line 164: Instead of creating a new object of the abstract domain please refer [this comment](https://github.com/appsmithorg/appsmith/pull/35707/files#r1719868103) to find out the entity class and create a new object of that class.

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
```diff
@@ -0,0 +1,323 @@
...
+        }
+        transactionContext.put(getObjectId(dbOps), dbOps);
+    }
+
+    private boolean isWriteOp(EntityData entityData, MethodSignature signature) {
+        // Special case like findCustomJsLib accepting the BaseDomain object as parameter instead of the UUID,
+        // hence the need to check for method name as well
+        if (entityData.getBaseDomain() != null
+                && isSaveOrCreateOp(entityData.getBaseDomain())
+                && !signature.getMethod().getName().contains(FIND)) {
```
abhvsn, line 240: @AnaghHegde I'm guessing as we are not creating the entities in `getEntityData` we may end up marking read queries like `getById` with write operation. This should not be the issue today as we generally end up using find keyword but this looks flaky. Please take a look into this, basically start with read op first like if the BridgeUpdate or BaseDomain is not found may be treated as read operation.

app/server/appsmith-server/src/test/java/com/appsmith/server/transactions/ImportApplicationTransactionServiceTest.java
```diff
@@ -188,12 +182,14 @@ public void importApplication_transactionExceptionDuringListActionSave_omitTrans
...
         // Check  if expected exception is thrown
         StepVerifier.create(resultMono)
-                .expectErrorMatches(error -> error instanceof AppsmithException
-                        && error.getMessage()
-                                .equals(AppsmithError.GENERIC_JSON_IMPORT_ERROR.getMessage(
-                                        createdWorkspace.getId(), "")))
+                .expectErrorMatches(error ->
+                        error instanceof AppsmithException && error.getMessage().equals(errorMessage))
                 .verify();
     }
```
abhvsn, line 194: Can we test this functionality by adding more tests around the update operation.

abhvsn (commented):
Can we please add tests to check if the transaction is working with update, insert(covered) and archive action?

app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/TransactionAspect.java
AnaghHegde: @abhvsn You cannot treat it as a read when the BridgeUpdate or BaseDomain is not present. It can be archive as well. The scenario you are mentioning is an edge case and needs to be handled by using the method name as a signature.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/TransactionHandlerCEImpl.java
```diff
@@ -0,0 +1,238 @@
...
+        actionCollectionRepository.archiveById(actionCollectionId);
+    }
+
+    private void saveActionToDb(NewAction action) {
+        newActionRepository.save(action);
+    }
+
+    private void archiveAction(String actionId) {
+        newActionRepository.archiveById(actionId);
+    }
```
abhvsn, line 117: Repetitive code and can be improved with something like what we have done for `getById`

abhvsn (approved):

PR #35717, opened 2024-08-16
perf: JSONForm widget optimise calling of clear errors

ashit-rath:
## Description
During on page load for a jsonForm widget, the `clearErrors` method was getting called even if there wasn't any error to clear. This led to significant number of this function execution for a large scale application

For more info - https://theappsmith.slack.com/archives/C024GUDM0LT/p1723206363693759

Fixes https://github.com/appsmithorg/appsmith/issues/35739

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/10440802784>
> Commit: 8f449b75c15bd57a65ca75c5b7cffc7417f645c5
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=10440802784&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.JSONForm`
> Spec:
> <hr>Sun, 18 Aug 2024 13:21:29 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
	- Improved error handling in form validation to ensure errors are cleared only when necessary.
  
- **Bug Fixes**
	- Enhanced logic for retrieving current error states to provide a more robust user experience when filling out forms.

- **Tests**
	- Expanded test coverage for form validation logic, including various scenarios to ensure correct behavior under different conditions.

### DIFF
--- a/app/client/src/widgets/JSONFormWidget/fields/useRegisterFieldValidity.ts
+++ b/app/client/src/widgets/JSONFormWidget/fields/useRegisterFieldValidity.ts
@@ -51,16 +51,6 @@ function useRegisterFieldValidity({
         Sentry.captureException(e);
       }
     }, 0);
-
-    setMetaInternalFieldState((prevState) => {
-      const metaInternalFieldState = klona(prevState.metaInternalFieldState);
-      set(metaInternalFieldState, `${fieldName}.isValid`, isValid);
-
-      return {
-        ...prevState,
-        metaInternalFieldState,
-      };
-    });
   }, [
     isValid,
     fieldName,
@@ -70,6 +60,18 @@ function useRegisterFieldValidity({
     clearErrors,
     setError,
   ]);
+
+  useEffect(() => {
+    setMetaInternalFieldState((prevState) => {
+      const metaInternalFieldState = klona(prevState.metaInternalFieldState);
+      set(metaInternalFieldState, `${fieldName}.isValid`, isValid);
+
+      return {
+        ...prevState,
+        metaInternalFieldState,
+      };
+    });
+  }, [fieldName, isValid, setMetaInternalFieldState]);
 }
 
 export default useRegisterFieldValidity;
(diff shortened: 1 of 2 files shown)


app/client/src/widgets/JSONFormWidget/fields/useRegisterFieldValidity.ts
```diff
@@ -58,7 +61,15 @@ function useRegisterFieldValidity({
...
-  }, [isValid, fieldName, fieldType]);
+  }, [
+    isValid,
+    fieldName,
+    fieldType,
+    setMetaInternalFieldState,
+    error,
+    clearErrors,
+    setError,
+  ]);
```
vsvamsi1, line 72: Are all the additional deps required here? because fulfilling all the deps used in this useEffect we are functionally altering the behaviour of what the useEffect was before. In addition we may cause a performance regression at this line 
`const metaInternalFieldState = klona(prevState.metaInternalFieldState)`
as now the useEffect triggered more often.

ashit-rath: Yes they are required based on the new linting rules we have and also because of best practices for hooks dependencies. 
This will not cause perf regressions since the additional dependencies other than the `error` are memoized functions for react-hooks-form and the the `error` object is also memoised by react-hooks-form (verified this). 

One thing we can do is split the efffct intro 2 effects since the logic and intent of the `setMetaInternalFieldState` in not dependent on the setError and clearError.

vsvamsi1: Oh thats great that the deps are memoised. As you have mentioned, It will be great if we can split this useEffect, `setMetaInternalFieldState` will be more resilient to any performance regressions in the case where someone introduces changes to cause one of the deps to change references.

vsvamsi1 (approved):

PR #35727, opened 2024-08-16
[Draft]: Add TreeStructure entity

znamenskii-ilia:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!WARNING]
> Tests have not run on the HEAD d6f7ebee4e48142e69405ff9f981a039ed1d4296 yet
> <hr>Mon, 19 Aug 2024 15:12:00 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
	- Introduced new tree manipulation capabilities with functions for adding, removing, finding, and updating nodes.
	- Added a structured approach for managing page elements, including support for dynamic properties.
	- Enhanced the type system for page elements to ensure hierarchical relationships and integrity.

- **Documentation**
	- Added definitions for new interfaces and types to improve clarity and usability of tree and page structures.

- **Bug Fixes**
	- Improved type safety and validation mechanisms for tree and page element operations.

### DIFF
--- a/app/client/src/modules/editor/common/libs/tree/Tree.ts
+++ b/app/client/src/modules/editor/common/libs/tree/Tree.ts
@@ -1,11 +1,12 @@
-import type { TreeNode } from "./types";
+import { isParentTreeNode } from "./Node";
+import type { ParentTreeNode, TreeNode } from "./types";
 
 /**
  * @param targetNode The node to add a child to
  * @param node The child node to add
  * @returns The new node with the added child
  */
-const appendChildNode = <TNode extends TreeNode>(
+export const appendChildNode = <TNode extends ParentTreeNode>(
   targetNode: TNode,
   node: TreeNode,
 ): TNode => ({
@@ -21,18 +22,18 @@ const appendChildNode = <TNode extends TreeNode>(
  * @param node The child node to add
  * @returns The new tree with the added node
  */
-export const addNode = <TNode extends TreeNode>(
+export const addNode = <TNode extends ParentTreeNode>(
   rootNode: TNode,
   targetNodeId: string,
   node: TreeNode,
 ): TNode => {
   if (rootNode.id === targetNodeId) return appendChildNode(rootNode, node);
 
-  if (rootNode.isLeaf) return rootNode;
+  const newChildren = rootNode.children.map((child) => {
+    if (!isParentTreeNode(child)) return child;
 
-  const newChildren = rootNode.children.map((child) =>
-    addNode(child, targetNodeId, node),
-  );
+    return addNode(child, targetNodeId, node);
+  });
 
   return { ...rootNode, children: newChildren };
 };
@@ -45,13 +46,19 @@ export const addNode = <TNode extends TreeNode>(
  * @returns The new tree with the removed node
  */
 export const removeNode = (
-  rootNode: TreeNode,
+  rootNode: ParentTreeNode,
   nodeId: string,
-): TreeNode | null => {
+): ParentTreeNode | null => {
   if (rootNode.id === nodeId) return null;
 
   const newChildren = rootNode.children
-    .map((child) => removeNode(child, nodeId))
+    .map((child) => {
+      if (child.id === nodeId) return null;
+
+      if (!isParentTreeNode(child)) return child;
+
+      return removeNode(child, nodeId);
+    })
     .filter((child) => child !== null);
 
   return { ...rootNode, children: newChildren };
@@ -66,14 +73,16 @@ export const removeNode = (
  * @returns The found node or null
  */
 export const findNode = (
-  rootNode: TreeNode,
+  rootNode: ParentTreeNode,
   nodeId: string,
 ): TreeNode | null => {
   if (rootNode.id === nodeId) return rootNode;
 
-  if (rootNode.isLeaf) return null;
-
   for (const child of rootNode.children) {
+    if (child.id === nodeId) return child;
+
+    if (!isParentTreeNode(child)) continue;
+
     const node = findNode(child, nodeId);
 
     if (node) return node;
@@ -91,7 +100,7 @@ export const findNode = (
  * @returns The new tree with the moved node
  */
 export const moveNode = (
-  rootNode: TreeNode,
+  rootNode: ParentTreeNode,
   nodeId: string,
   targetNodeId: string,
 ) => {
@@ -105,7 +114,11 @@ export const moveNode = (
 
   // To move node it first needs to be copied to the new parent
   // and then removed from the old parent
-  return removeNode(addNode(rootNode, targetNodeId, nodeToMove), nodeId);
+  const updatedRootNode = removeNode(rootNode, nodeId);
+
+  if (!updatedRootNode) return rootNode;
+
+  return addNode(updatedRootNode, targetNodeId, nodeToMove);
 };
 
 /**
@@ -117,17 +130,21 @@ export const moveNode = (
  * @returns The new tree with the updated node
  */
 export const updateNode = (
-  rootNode: TreeNode,
+  rootNode: ParentTreeNode,
   nodeId: string,
   data: Partial<TreeNode>,
-) => {
+): ParentTreeNode => {
   const nodeToUpdate = findNode(rootNode, nodeId);
 
   if (!nodeToUpdate) return rootNode;
 
   const updatedNode: TreeNode = { ...nodeToUpdate, ...data };
 
-  if (updatedNode.parentId === null) return updatedNode;
+  if (updatedNode.parentId === null) return rootNode;
 
-  return moveNode(rootNode, nodeId, updatedNode.parentId);
+  const updatedRootNode = removeNode(rootNode, nodeId);
+
+  if (!updatedRootNode) return rootNode;
+
+  return addNode(updatedRootNode, updatedNode.parentId, updatedNode);
 };
--- a/app/client/src/modules/editor/common/libs/tree/Tree.ts
+++ b/app/client/src/modules/editor/common/libs/tree/Tree.ts
@@ -1,0 +1,133 @@
+import type { TreeNode } from "./types";
+
+/**
+ * @param targetNode The node to add a child to
+ * @param node The child node to add
+ * @returns The new node with the added child
+ */
+const appendChildNode = <TNode extends TreeNode>(
+  targetNode: TNode,
+  node: TreeNode,
+): TNode => ({
+  ...targetNode,
+  children: [...targetNode.children, node],
+});
+
+/**
+ * Adds node to the given tree
+ *
+ * @param rootNode The root node of the tree
+ * @param targetNodeId The id of the node to add a child to
+ * @param node The child node to add
+ * @returns The new tree with the added node
+ */
+export const addNode = <TNode extends TreeNode>(
+  rootNode: TNode,
+  targetNodeId: string,
+  node: TreeNode,
+): TNode => {
+  if (rootNode.id === targetNodeId) return appendChildNode(rootNode, node);
+
+  if (rootNode.isLeaf) return rootNode;
+
+  const newChildren = rootNode.children.map((child) =>
+    addNode(child, targetNodeId, node),
+  );
+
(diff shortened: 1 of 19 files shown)


app/client/src/modules/editor/domain/entities/pageStructure/PageElement.ts
```diff
@@ -0,0 +1,119 @@
...
+  | SectionPageElement
+  | ZonePageElement
+  | ButtonPageElement;
+
+interface PageElementTypeMap {
+  CANVAS_WIDGET: CanvasPageElement;
+  SECTION_WIDGET: SectionPageElement;
+  ZONE_WIDGET: ZonePageElement;
+  WDS_BUTTON_WIDGET: ButtonPageElement;
+}
```
znamenskii-ilia, line 91: I believe this can be generated from PageElement. Will check later

app/client/src/modules/editor/domain/entities/pageStructure/PageElement.ts
```diff
@@ -0,0 +1,119 @@
...
+export interface PageElementWithDynamic {
+  dynamicBindingPathList: DynamicPath[];
+  dynamicTriggerPathList: DynamicPath[];
+  dynamicPropertyPathList: DynamicPath[];
+}
+
+export interface BasePageElement extends TreeNode<PageElementType> {
+  name: string;
+  version: number;
+}
```
znamenskii-ilia, line 30: I kept as similar to what we have now, that is flat. Later I'd like to store all Node data into separate field

app/client/src/modules/editor/domain/entities/pageStructure/PageStructure.ts
```diff
@@ -0,0 +1,50 @@
...
+} from "./PageElement";
+import { addNode, findNode } from "../../../common/libs/tree/Tree";
+
+export type PageStructure = CanvasPageElement;
+
+const allowedChildrenMap: { [key in PageElementType]?: PageElementType[] } = {
+  CANVAS_WIDGET: ["SECTION_WIDGET"],
+  SECTION_WIDGET: ["ZONE_WIDGET"],
+  ZONE_WIDGET: ["WDS_BUTTON_WIDGET"],
+};
```
znamenskii-ilia, line 14: I'm not sure if we have strict rules about elements nesting. So it is configurable now. Later we can define more strict rules and control them by types as well.

app/client/src/modules/editor/common/libs/tree/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from "./Node";
+export * from "./Tree";
+export * from "./types";
```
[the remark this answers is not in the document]
riodeuno: We can extract this at any point. Eventually this should be a package, so it is your decision on when to do this.

app/client/src/modules/editor/common/libs/tree/types.ts
```diff
@@ -0,0 +1,21 @@
...
+  children: TreeNode[];
+}
+
+export interface ParentTreeNode<TType extends string = string>
+  extends TreeNode<TType> {
+  isLeaf: false;
+  children: TreeNode[];
+}
+
+export interface LeafTreeNode<TType extends string = string>
```
znamenskii-ilia, line 15: Not sure about this name. Leaf means just the last node in a branch. For us it is a node that can't have nested nodes. I will prob rename it to TerminalNode

app/client/src/modules/editor/domain/entities/pageStructure/PageStructure.ts
riodeuno: These were designed at a stage when we weren't sure how the layout mechanics would be. So, we designed something generic to avoid multiple rewrites. This is an artefact of that decision. Initially these were supposed to be defined in the layout types. I believe the devs who worked on this, made it more specific. We will need to revisit how layouts work today, we may not need to keep it generic if it helps performance and maintainability.

app/client/src/modules/editor/domain/entities/pageStructure/PageElement.ts
```diff
@@ -0,0 +1,119 @@
...
+): element is PageElementTypeMap[TType] => element.type === type;
+
+export const isEmptyParentElement = isEmptyTreeNode;
+
+// TYPE GUARDS
+export const isParentPageElement = (
+  element: PageElement,
+): element is PageElement & ParentPageElement => isParentTreeNode(element);
+
+export const isDynamicEnabled = (element: PageElementWithDynamic): boolean =>
```
riodeuno, line 118: I believe we can rename it to `hasDynamicProperties`. Dynamic properties are properties that are configured using JS code. The paths are set in node configurations as a caching mechanism. These properties can always be regenerated by inspecting the properties of the node and their values. We may or may not need these in nodes depending on how we architecture the dependecypath generation for evaluations.

app/client/src/modules/editor/domain/entities/pageStructure/PageElement.ts
```diff
@@ -0,0 +1,119 @@
...
+import type { Colors } from "../../enums/Colors";
+import type { DynamicPath } from "./DynamicPath";
+
+export type PageElementType =
+  | "CANVAS_WIDGET"
+  | "SECTION_WIDGET"
+  | "ZONE_WIDGET"
+  | "WDS_BUTTON_WIDGET";
+
+export interface PageElementWithDynamic {
```
riodeuno, line 21: How about `PageElementWithDynamicProperties` or `PageElementWithDynamicConfigurations`?

app/client/src/modules/editor/domain/entities/pageStructure/PageStructure.ts
```diff
@@ -0,0 +1,50 @@
...
+): boolean => {
+  return !allowedChildrenMap[rootPageElement.type]?.includes(pageElement.type);
+};
+
+export const addPageElement = <TElement extends PageElement>(
+  rootPageElement: TElement,
+  targetPageElementId: string,
+  pageElement: PageElement,
+): TElement => {
+  const targetPageElement = findPageElement(
```
riodeuno, line 34: Do both `findPageElement` and `addNode` recursively traverse the tree? If so, I was wondering if we can optimise this. At this point, I'd like to investigate how many callees this function would have to know if the optimisation is required. Alternatively, if the optimisation is cheap, we can just do it and leave it at that. WDYT?

app/client/src/modules/editor/domain/entities/pageStructure/PageElement.ts
```diff
@@ -0,0 +1,119 @@
...
+  | ButtonPageElement;
+
+interface PageElementTypeMap {
+  CANVAS_WIDGET: CanvasPageElement;
+  SECTION_WIDGET: SectionPageElement;
+  ZONE_WIDGET: ZonePageElement;
+  WDS_BUTTON_WIDGET: ButtonPageElement;
+}
+
+export const generatePageElementId = (prefix = ""): string =>
```
riodeuno, line 93: If there is no prefix provided, would this generate `_<randomuuid>`? Do you think we can remove the `_` if there is no prefix?

app/client/src/modules/editor/domain/entities/pageStructure/PageElement.ts
znamenskii-ilia: Renamed

app/client/src/modules/editor/domain/entities/pageStructure/PageElement.ts
znamenskii-ilia: Renamed to `PageElementWithDynamicProperties`

app/client/src/modules/editor/domain/entities/pageStructure/PageElement.ts
znamenskii-ilia: Updated

app/client/src/modules/editor/domain/entities/pageStructure/PageStructure.ts
znamenskii-ilia: Speaking about optimization. I intentionally used nested Tree as a default structure because it matches what we store on BE side. I wanted to avoid premature optimization and situation which we have how when we turn tree into a flat object, then again into a tree, and again into a flat object (I saw it some in code). When we measure and see perf issues we will be able to easily replace Tree with FlatTree, without touching clients of PageStructure.

app/client/src/modules/editor/common/libs/tree/types.ts
znamenskii-ilia: What I wanted to say is that the term `Leaf` might confuse developers as well. Image you have a Node with no children (It is called EmptyNode in my code, which is not a standard name), but which can have it. It is a Leaf node, because it doesn't have children atm. So `leafness` for me is about state. I'd like to have name for Node that cannot/allowed to have children at all. I don't know "official" name for that, because trees are supposed to grow with no limits. But for our purposes we may introduce such term. We can use in Tree or only in PageStructure. I'd use term `Terminal` for that, but ok with any other.
This may seem insignificant, but I expect that we will extend the use of these terms to other parts of the application (including the UI, for example Button is a TerminalPageComponent, since can't have children).

riodeuno: > trees are supposed to grow with no limits
Yes, in general and in Fixed mode, this can happen. However, as I mentioned this to @nidhi-nair in the slack message. The max depth of the tree in Anvil will be 4. There are no plans of there being a scenario where it could grow more in the near future. The only scenario where it could grow is the HoC (data view widget, etc) which hasn't been discussed at length yet.

>This may seem insignificant, but I expect that we will extend the use of these terms to other parts of the application 
Sounds good to me.

KelvinOm: > because trees are supposed to grow with no limits

Even taking into account the fact that at the moment we can have max depth only 4, this can change at any time. To be honest, I don't quite understand the value in labeling a node without children with specific flag. Why is that needed? I think developers are able to solve such things themselves when writing a code.

KelvinOm (commented):
To be honest, it is quite difficult for me to review this concept without it's practical application. Overall, everything looks good, but I'd like to see a practical application of this code.

app/client/src/modules/editor/common/libs/tree/Tree.ts
```diff
@@ -0,0 +1,133 @@
...
+ * @param targetNode The node to add a child to
+ * @param node The child node to add
+ * @returns The new node with the added child
+ */
+const appendChildNode = <TNode extends TreeNode>(
+  targetNode: TNode,
+  node: TreeNode,
+): TNode => ({
+  ...targetNode,
+  children: [...targetNode.children, node],
```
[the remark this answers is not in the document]
znamenskii-ilia: You are right, it should add to the end, as it does now.. Your code adds it to the beginning of the array. Or I'm missing something?

app/client/src/modules/editor/common/libs/tree/types.ts
znamenskii-ilia: You mean `isLeaf` field? It is quite standard one, but I also was thinking about removing it. Leafness of a node can be derived from presence of `children` field. Basically I see two options to mark a node is leaf
option 1
```
export interface TreeNode<TType extends string = string> {
  // ...
  children: TreeNode[];
}

export interface ParentTreeNode<TType extends string = string>
  extends TreeNode<TType> {
  isLeaf: false;
}

export interface LeafTreeNode<TType extends string = string>
  extends TreeNode<TType> {
  isLeaf: true;
}
```

option 2
```
export interface TreeNode<TType extends string = string> {
  // ...
}

export interface ParentTreeNode<TType extends string = string>
  extends TreeNode<TType> {
  children: TreeNode[];
}

export interface LeafTreeNode<TType extends string = string>
  extends TreeNode<TType> {
  children: never;
}
```

I'm going to keep them both until the meeting on Wen where we are going to discuss PageStructure with BE and git team.

PR #35786, opened 2024-08-20
chore: Add cleanup and update stage for tenant at the server restart

abhvsn:
## Description
PR to add the cleanup and update default tenant at the server restart. This was required because whenever we update tenant via migrations the result gets reverted because the cached tenant is not getting updated as we generally end up updating only the MongoDB state and not the cached entry. 

/test Sanity
 
> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/10471754877>
> Commit: e288cfe1084bdbf2104a656bf60eb9c24708fe77
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=10471754877&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Sanity`
> Spec:
> <hr>Tue, 20 Aug 2024 13:01:42 UTC

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **New Features**
	- Implemented a mechanism to refresh tenant policies and manage cache cleanup during server restarts, ensuring accurate policy enforcement for multi-tenant applications.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/TenantConfig.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/TenantConfig.java
@@ -2,6 +2,7 @@ package com.appsmith.server.configurations;
 
 import com.appsmith.server.domains.Tenant;
 import com.appsmith.server.helpers.CollectionUtils;
+import com.appsmith.server.repositories.CacheableRepositoryHelper;
 import com.appsmith.server.repositories.TenantRepository;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
@@ -18,6 +19,7 @@ import static com.appsmith.external.models.BaseDomain.policySetToMap;
 public class TenantConfig implements ApplicationListener<ApplicationStartedEvent> {
 
     private final TenantRepository tenantRepository;
+    private final CacheableRepositoryHelper cachableRepositoryHelper;
 
     // Method to cleanup the cache and update the default tenant policies if the policyMap is empty. This will make sure
     // cache will be updated if we update the tenant via startup DB migrations.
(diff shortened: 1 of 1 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/TenantConfig.java
```diff
@@ -0,0 +1,37 @@
...
+import lombok.RequiredArgsConstructor;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import reactor.core.publisher.Mono;
+
+import static com.appsmith.external.models.BaseDomain.policySetToMap;
+
+@Configuration
+@RequiredArgsConstructor
+public class TenantConfig {
```
nidhi-nair, line 15: Nit: Can we make this a listener on the Spring context refresh event so that the execution period for this eviction is predictable?

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/TenantConfig.java
```diff
@@ -0,0 +1,37 @@
...
+
+    private final TenantService tenantService;
+    private final CacheableRepositoryHelper cacheableRepositoryHelper;
+
+    // Bean to cleanup the cache and update the default tenant policies on every server restart. This will make sure
+    // cache will be updated if we update the tenant via migrations.
+    @Bean
+    public void cleanupAndUpdateRefreshDefaultTenantPolicies() {
```
nidhi-nair, line 23: Does this mean that every time any pod restarts it will reset the cache? Do you see any downside to this with the current stability issue?

nidhi-nair: If this is by type, shall we consider making this a conditional eviction instead?

abhvsn: We are cleaning up the cache so yeah definitely for the first call we will hit the DB, but we will have correct data. Ideally we want to hook this with the DB migrations which updates the default tenant. Any suggestions from your side?

nidhi-nair (approved):

nidhi-nair (approved):

AnaghHegde (approved):

PR #35845, opened 2024-08-22
chore: Input morphing

jsartisan:


ichik (changes requested):
1. Wrong thumnails and icons (correct one should be available in the Figma file that we extract icons from)
2. Sorting in the `Inputs` section. Regular `Input` should be first, rest can be alphabetized for now.
3. As far as I can see morphing the input changes neither the name nor the icon in the explorer. Those should also update (unless builder set some custom name manually).
4. Input type can be moved out of data section to be its own thing.

sagar-qa007:
@jsartisan Please add unit test case if possible.

znamenskii-ilia (approved):

ichik (approved):
Please add a follow-up to handle widget names better when morphing.

jsartisan:
Created the ticket - https://github.com/appsmithorg/appsmith/issues/35894

PR #35862, opened 2024-08-23
Fix/select label key function

### LINKED ISSUE #35269: [Bug]: Label key functions not working as expected

### Description

This worked on a version prior to 1.30, but the bug is present on 1.30 and 1.33.

When using a function to get a dynamic label key or value key, it is not interpreted as a key but instead as the label itself:
image

### Steps To Reproduce

1. Create a select widget
2. Create a JS object called "test"
3. Create an array in the JS object "select: [{label: 'first label', key: 'first key'}]",
4. Bind the select widget's source data to {{test.select}}
5. Set the label key to JS
6. Enter the word "label" as the entry in label key
7. Note the widget now says "first label" in the dropdown
image

8. Enter a function that should return the string "label" as the entry in label key
9. Note that the widget now says "label" in the dropdown
image

For some reason now returning a string 'label' is not being interpreted the same way as just writing 'label' in the box.

The reason I have this function there is because Appsmith gives me says there are errors in the log when the query that populates my source data hasn't been run yet.  I can just ignore the errors, I just added this length check so I could cut down on the number of displayed errors and hopefully only see the ones that matter.

However, if I actually did need to have a dynamic label or value for whatever reason, this would be a blocker.

It would also be nice if it just didn't give label key and value key errors if the source data length is 0 / empty.

### Public Sample App

_No response_

### Environment

Production

### Severity

Medium (Frustrating UX)

### Issue video log

_No response_

### Version

Self-Hosted 1.33

jacquesikot:
## Description
**Problem**
When the Select widget's labelKey or valueKey properties are switched to JS mode, passing a function that returns a string as the key causes the dropdown options to break. Instead of correctly mapping the labels and values, the options are incorrectly returned as an array of the key itself.

**Root Cause**
The issue lies in the `getOptions` function in derived.js. The logic that computes the select options based on the labelKey and valueKey was not handling cases where these properties are functions that return arrays. Instead of mapping through the source data and returning the corresponding values for each key, the function was mistakenly returning the entire array.

**Solution**
The `getOptions` function in derived.js has been updated to correctly handle cases where labelKey or valueKey are arrays. The function now maps through the sourceData and computes the correct labels and values based on the provided properties, ensuring that the dropdown options are displayed as expected.

Fixes #35269

## Automation

/ok-to-test tags="@tag.Sanity, @tag.Select, @tag.Widget, @tag.Binding"

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/client/cypress/e2e/Regression/ClientSide/Widgets/Select/Select2_Spec.ts
+++ b/app/client/cypress/e2e/Regression/ClientSide/Widgets/Select/Select2_Spec.ts
@@ -305,5 +305,26 @@ describe(
         1,
       );
     });
+
+    it("10. Validate using function inside label key works correctly", () => {
+      deployMode.NavigateBacktoEditor();
+      EditorNavigation.SelectEntityByName("Select1", EntityType.Widget);
+      propPane.ToggleJSMode("Label key", true);
+      propPane.UpdatePropertyFieldValue("Label key", "{{(() => 'name')()}}");
+      agHelper.SelectDropDown("Blue");
+      agHelper.ReadSelectedDropDownValue().then(($selectedValue) => {
+        expect($selectedValue).to.eq("Blue");
+      });
+    });
+
+    it("11. Validate using function inside value key works correctly", () => {
+      EditorNavigation.SelectEntityByName("Select1", EntityType.Widget);
+      propPane.ToggleJSMode("Value key", true);
+      propPane.UpdatePropertyFieldValue("Value key", "{{(() => 'code')()}}");
+      agHelper.SelectDropDown("Blue");
+      agHelper.ReadSelectedDropDownValue().then(($selectedValue) => {
+        expect($selectedValue).to.eq("Blue");
+      });
+    });
   },
 );
--- a/app/client/src/widgets/SelectWidget/widget/derived.js
+++ b/app/client/src/widgets/SelectWidget/widget/derived.js
@@ -8,13 +8,13 @@ export default {
     if (typeof props.optionLabel === "string") {
       labels = sourceData.map((d) => d[props.optionLabel]);
     } else if (_.isArray(props.optionLabel)) {
-      labels = props.optionLabel;
+      labels = sourceData.map((d, i) => d[props.optionLabel[i]]);
     }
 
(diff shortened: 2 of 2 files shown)


app/client/cypress/e2e/Regression/ClientSide/Widgets/Select/Select2_Spec.ts
```diff
@@ -305,5 +305,26 @@ describe(
...
+      EditorNavigation.SelectEntityByName("Select1", EntityType.Widget);
+      propPane.ToggleJSMode("Label key", true);
+      propPane.UpdatePropertyFieldValue("Label key", "{{(() => 'name')()}}");
+      agHelper.SelectDropDown("Blue");
+      agHelper.ReadSelectedDropDownValue().then(($selectedValue) => {
+        expect($selectedValue).to.eq("Blue");
+      });
+    });
+
+    it("11. Validate using function inside value key works correctly", () => {
```
sagar-qa007, line 320: @jacquesikot can you please elaborate/define "works correctly" in title? Same for above test case.

app/client/cypress/e2e/Regression/ClientSide/Widgets/Select/Select2_Spec.ts
```diff
@@ -305,5 +305,26 @@ describe(
         1,
       );
     });
+
+    it("10. Validate using function inside label key works correctly", () => {
```
sagar-qa007, line 309: Just want to confirm that, do we need cypress test case here or unit test will work here?

sagar-qa007 (changes requested):

app/client/cypress/e2e/Regression/ClientSide/Widgets/Select/Select2_Spec.ts
jacquesikot: @sagar-qa007 We need to check the behaviour after changing the property pane. I think we need a cypress test for that.

rahulbarwal (changes requested):
Lets combine the test cases 10 & 11 and add a test for the specific problem that the user has shared.

rahulbarwal (approved):

PR #35918, opened 2024-08-28
chore: adding telemetry for klona

vsvamsi1:
## Description
Klona operations are generally expensive, for that reason we will be tracking the performance of all klonas in the main thread.

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/10608431935>
> Commit: cab2716f74f3747d3df64342f4d7f4972358a3e6
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=10608431935&attempt=2" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Thu, 29 Aug 2024 06:26:43 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
	- Introduced telemetry tracking for cloning operations across various components, enhancing performance monitoring and debugging capabilities.

- **Bug Fixes**
	- Improved observability of cloning operations, aiding in identifying potential performance bottlenecks.


app/client/src/widgets/JSONFormWidget/fields/ArrayField.tsx
```diff
@@ -229,10 +250,17 @@ function ArrayField({
...
+      const newValues = startAndEndSpanForFn(
+        "klona",
+        {
+          codeSegment: "ArrayField.remove.newValues",
+        },
+        () =>
+          klona(
+            // TODO: Fix this the next time the file is edited
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            values.filter((_val: any, index: number) => index !== removedIndex),
```
znamenskii-ilia, line 262: May I ask you to fix `TODO`?

app/client/src/utils/helpers.tsx
```diff
@@ -828,7 +829,13 @@ export const mergeWidgetConfig = (target: any, source: any) => {
   // TODO: Fix this the next time the file is edited
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   const sectionMap: Record<string, any> = {};
```
znamenskii-ilia, line 831: May I ask you to fix TODO?

znamenskii-ilia:
Many files with `// TODO: Fix this the next time the file is edited` were changed. May I ask you to fix at least some of them?

dvj1988 (approved):

znamenskii-ilia (changes requested):
Pls check my comment about Todos

app/client/src/utils/helpers.tsx
```diff
@@ -818,6 +823,32 @@ export function isValidColor(color: string) {
   return color?.includes("url") || validateColor(color) || isEmptyOrNill(color);
 }
 
+function klonaWithTelemtryWrapper<T>(
```
rajatagrawal, line 826: Typo in the word `telemetry`

rajatagrawal: Would it make sense to add an attribute to the span which differentiates between various kinds of klonas ? like we may want to compare performance of `klonaFull` vs `klonaRegular` ?

rajatagrawal (changes requested):
Left feedback

app/client/src/utils/helpers.tsx
dvj1988: > Would it make sense to add an attribute to the span which differentiates between various kinds of klonas ? like we may want to compare performance of klonaFull vs klonaRegular ?
I don't think that would be necessary here since we already know where each klona function is used from the span attribute. Also, comparisons among klona functions would only make sense if we could measure the latency using the same object.

rajatagrawal (approved):

vsvamsi1:
> Many files with `// TODO: Fix this the next time the file is edited` were changed. May I ask you to fix at least some of them?
Sure I can take this up in a separate PR.

PR #35940, opened 2024-08-28
Test/fix checkbox grouptests

NandanAnantharamu:
RCA:
checkbox should be checked when clicked


NandanAnantharamu:
closing as its redundant

PR #35965, opened 2024-08-29
chore: cache dependencyMap computation on client

dvj1988:
## Description
This PR includes the changes to enable browser side caching of dependency map in published mode.

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/11338708443>
> Commit: af545a2309e259bd2b44533b6572a2cfbb53b51d
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=11338708443&attempt=3" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Tue, 15 Oct 2024 07:42:33 UTC

## Communication
- [ ] Yes
- [x] No

## Release Notes

- **New Features**
	- Introduced asynchronous profiling for web worker operations, enhancing telemetry capabilities.
	- Updated handling of the `EVAL_TREE` action to support asynchronous processing, improving performance.
	- Added a caching mechanism for computation results, optimizing data retrieval and management.

- **Bug Fixes**
	- Improved accuracy of `applicationId` extraction for fetching published page resources.

- **Tests**
	- Enhanced test cases for `DataTreeEvaluator` and caching mechanisms to reflect realistic application states.

- **Chores**
	- Updated various components to support new asynchronous logic and caching properties.


app/client/src/workers/common/DependencyMap/index.ts
```diff
@@ -45,27 +52,60 @@ export function createDependencyMap(
...
+        dataTreeEvalRef.errors.push(...errors);
+
+        if (errors.length) {
+          shouldCache = false;
+        }
+      }
+    });
+
+    DependencyMapUtils.makeParentsDependOnChildren(dependencyMap);
+
```
vsvamsi1, line 97: Can this aspect be part of the cached result, since this is also an expensive computation.

app/client/src/workers/common/DependencyMap/index.ts
```diff
@@ -45,27 +52,60 @@ export function createDependencyMap(
...
+    });
 
-    for (const path of Object.keys(entityDependencies)) {
-      const pathDependencies = entityDependencies[path];
-      const { errors, references } = extractInfoFromBindings(
-        pathDependencies,
+  if (dependencyMapCache) {
+    Object.entries(dependencyMapCache).forEach(([path, references]) => {
+      dependencyMap.addDependency(path, references);
+    });
```
vsvamsi1, line 68: For a large graph this is also an expensive computation, can the cache return the graph directly. So the cache would consist  of dependencyMap with the following properties
`

  constructor() {
    this.#nodes = new Map();
    this.#dependencies = new Map();
    this.#invalidDependencies = new Map();
    this.#dependenciesInverse = new Map();
    this.#invalidDependenciesInverse = new Map();
  }
`

app/client/src/workers/common/AppComputationCache/index.ts
```diff
@@ -0,0 +1,294 @@
...
+
+    if (!shouldCache) {
+      return null;
+    }
+
+    const cacheKey = this.generateCacheKey([appId, pageId, appMode, cacheName]);
+
+    try {
+      const cached = await this.store.getItem<ICachedData<T>>(cacheKey);
+      if (isNull(cached)) {
```
vsvamsi1, line 163: I assumed setting and retrieving items from local storage are synchronous operations.

app/client/src/workers/common/AppComputationCache/index.ts
```diff
@@ -0,0 +1,294 @@
...
+    const shouldCache = this.isComputationCached({
+      cacheName,
+      appMode,
+    });
+
+    if (!shouldCache) {
+      return null;
+    }
+
+    const cacheKey = this.generateCacheKey([appId, pageId, appMode, cacheName]);
```
vsvamsi1, line 159: Would it better if the cacheKey to have the timestamp as part of the cacheKey, because right now if there is cache miss we are reading the entire stale value below and there is a parsing cost associated to reading from localStorage or indexedDb. If its was part of the cacheKey we would directly get a null during a cache miss and infer that the cache is stale with the added benefit of not parsing the stale result.

vsvamsi1: Also if the app or page gets deleted, can we have code to delete the respective entries from the browser local storage. This is to prevent a memory leak in the browser storage.

app/client/src/workers/common/AppComputationCache/index.ts
dvj1988: We are using the package `localforage` to implement the cache. The package selects the appropriate driver (IndexedDB, WebSQL, or LocalStorage) based on browser compatibility. All the APIs of localforage are promise/callback based, and since promises are cleaner and easier to maintain, I used them.

app/client/src/workers/common/DependencyMap/index.ts
dvj1988: `addDependency` takes around 15ms and `addNodes` around 73ms for Xolair. These are are the two methods we continue to use even if dependencyMap is cached.

@vsvamsi1 Instead of adding all I would suggest let's instrument these separately and take a call based on the p95 latencies.

app/client/src/workers/common/AppComputationCache/index.ts
dvj1988: > Would it better if the cacheKey would have the timestamp as part of the cacheKey, because right now if there is cache miss we are reading the entire stale value below and there is a parsing cost associated to reading from localStorage or indexedDb. If its was part of the cacheKey we would directly get a null during a cache miss and infer that the cache is stale with the added benefit of not parsing the stale result. So the performance of cache miss and hit would be fast in this manner.

Implementing it the way you have mentioned requires an extra cleanup fn to remove invalid caches. I have added the same and made sure the clean up fn is debounced so that it doesn't interfere with the eval cycles in most cases.

dvj1988: > Also if the app or page gets deleted, can we have code to delete the respective entries from the browser local storage. This is to prevent a memory leak in the browser storage.

The app viewer might not be the user responsible for deleting a page/app. Therefore a better way to remove invalid or unused cached values would be to implement a LRU cache. I have added the necessary metadata to implement this in a later version.

app/client/src/workers/common/AppComputationCache/index.ts
```diff
@@ -0,0 +1,306 @@
...
+    cacheName,
+    cacheProps,
+  }: {
+    cacheProps: ICacheProps;
+    cacheName: EComputationCacheName;
+  }) {
+    const { appId, appMode, instanceId, pageId, timestamp, workspaceId } =
+      cacheProps;
+
+    const timeStampEpoch = new Date(timestamp).getTime();
```
vsvamsi1, line 175: When sending the cache props can we directly send it as an epoch, to simply the format.

dvj1988: I would like to understand the reasoning behind this. We need to do the transformation either sending it to the worker thread or in the cache computation util.

app/client/src/workers/common/AppComputationCache/index.ts
```diff
@@ -0,0 +1,306 @@
...
+        const keyParts = key.split(AppComputationCache.CACHE_KEY_DELIMITER);
+        const cacheKeyTimestamp = parseInt(keyParts[5], 10);
+
+        return (
+          keyParts[0] === cacheProps.instanceId &&
+          keyParts[1] === cacheProps.workspaceId &&
+          keyParts[2] === cacheProps.appId &&
+          keyParts[3] === cacheProps.pageId &&
+          keyParts[4] === cacheProps.appMode &&
+          cacheKeyTimestamp !== new Date(cacheProps.timestamp).getTime()
```
vsvamsi1, line 279: Can you elaborate what entries are trying to delete.

app/client/src/workers/common/AppComputationCache/index.ts
```diff
@@ -0,0 +1,306 @@
...
+      return null;
+    }
+
+    const cacheKey = this.generateCacheKey({
+      cacheProps,
+      cacheName,
+    });
+
+    try {
+      const cached = await this.store.getItem<ICachedData<T>>(cacheKey);
```
vsvamsi1, line 141: Can we also track the performance of this function.

vsvamsi1:
Minor NIT but the PR looks good.

app/client/src/workers/common/AppComputationCache/index.ts
vsvamsi1: I feel its better before because serialisation of date object may add some latency(very minor). Epoch is a primitive and it can be transmitted faster.

vsvamsi1 (approved):
Minor nit picks but the PR looks good.

app/client/src/workers/common/AppComputationCache/index.ts
dvj1988: The cache key includes the `lastDeployedAt` timestamp, so the dependency map of a newly deployed app will be added as a new entry in the storage. Therefore, we need to delete all previous entries that don't match the current `lastDeployedAt`  timestamp.

app/client/src/workers/common/AppComputationCache/index.ts
dvj1988: `timestamp` is an alias to the `lastDeployedAt` value of an application. Its in the ISO string format (not a Date object), which is also a (string) primitive.  So I don't think it would add any latency here.

dvj1988:
Still on-going

rishabhrathod01:
@dvj1988 please let me know once the PR is ready for review.

dvj1988:
> @dvj1988 please let me know once the PR is ready for review. 

Changes are complete and its ready for review.

rajatagrawal (approved):
LGTM

rishabhrathod01 (approved):
LGTM

PR #35966, opened 2024-08-29
chore: Commit secrets

abhvsn:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/server/secrets.py
+++ b/app/server/secrets.py
@@ -1,0 +1,1 @@
+adobe_client_secret='ASDfwadANRHLasdwerfNXPbbtV5fmseX/GJC1P/'


app/server/secrets.py
```diff
@@ -0,0 +1 @@
+adobe_client_secret='ASDfwadANRHLasdwerfNXPbbtV5fmseX/GJC1P/'
```
sagar-qa007, line 1: @abhvsn Do we think secret should be pass like this? Can we please do this we GITHUB SECRETS ?

sagar-qa007 (changes requested):

app/server/secrets.py
abhvsn: @sagar-qa007 this was the test if GH can catch these dummy secrets and create an alert for us. I'm still exploring how does the secret scanning works in GH, and experimenting along the lines.  

Ref doc: https://docs.github.com/en/code-security/secret-scanning/introduction/supported-secret-scanning-patterns#supported-secrets

PR #36064, opened 2024-09-03
test: updated test with intercept

NandanAnantharamu:
RCA:
Widget was not loading as a result test is flaky

Solution:
Adding an intercept to make sure the widget is loaded

EE PR: https://github.com/appsmithorg/appsmith-ee/pull/5020

> [!WARNING]
> Tests have not run on the HEAD 125780e27b5b8d4898fc7d48e8f9f5a8d92d0138 yet
> <hr>Wed, 04 Sep 2024 07:15:27 UTC

### DIFF
--- a/app/client/cypress/e2e/Regression/ClientSide/Widgets/Custom/CustomWidgetEditorPropertyPane_spec.ts
+++ b/app/client/cypress/e2e/Regression/ClientSide/Widgets/Custom/CustomWidgetEditorPropertyPane_spec.ts
@@ -1,9 +1,15 @@
 import {
   agHelper,
+  assertHelper,
   deployMode,
   entityExplorer,
+  locators,
   propPane,
 } from "../../../../../support/Objects/ObjectsCore";
+import {
+  PageLeftPane,
+  PagePaneSegment,
+} from "../../../../../support/Pages/EditorNavigation";
 
 describe(
   "Custom widget Tests",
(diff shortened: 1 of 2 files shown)


sagar-qa007 (approved):

app/client/cypress/e2e/Regression/ClientSide/Widgets/Custom/CustomWidgetEditorPropertyPane_spec.ts
```diff
@@ -26,6 +32,9 @@ describe(
     };
 
     it("shoud check that default model changes are converyed to custom component", () => {
+      agHelper.AssertElementExist(locators._widgetInDeployed("customwidget"));
+      cy.intercept("https://api.segment.io/v1/b").as("widgetLoad");
```
mohanarpit, line 36: @NandanAnantharamu @ApekshaBhosale @sagar-qa007 
1. Why would this network call impact the loading of a widget on the canvas?
2. Why aren't we intercepting Segment API calls globally and only doing it for this specific test?

NandanAnantharamu: @mohanarpit , actually we were seeing flakiness while the custom widget was loading intermittently hence we added this intercept, and regarding point 2 i agree if this approach works will add it globally.

mohanarpit: I don't think this is the correct fix. Segment API calls are already redirected to TED on localhost. Refer [this line](https://github.com/appsmithorg/appsmith/blob/release/.github/workflows/ci-test-custom-script.yml#L192) in our CI. Hence it should ideally not be causing issues in widget rendering. 

Is this a potential product bug that we've uncovered?

NandanAnantharamu: True .. i am running the test multiple times in CI to figure out the solution as its flaky. As of now no credible evidence to say its a product bug

Issue #36102, opened 2024-09-03
[Task]: Optimise fetching application to improve FPL


subrata71:
### SubTasks

https://www.notion.so/appsmith/Optimizing-fetching-application-as-part-of-the-pages-API-aabc24b41160425d87a3604c80fd6777

PR #36118, opened 2024-09-04
chore: introduce caching and projection to optimise FPL

subrata71:
## Description
This PR aims to enhance the performance of the pages span in consolidated-API view mode. It does so by brining the following changes

- **Add Projection:** Implement projections to enhance performance for `appsmith.consolidated-api.view.pages.getpage`.
- **Implement Caching:** Introduce caching to eliminate the `getpage` query, thereby optimizing `appsmith.consolidated-api.view.application_id` and related spans.
    
    We will implement caching to store `defaultApplicationId` information in memory. This cache will reduce the need for the time-consuming `getpage` query. On a cache miss, the system will first fetch the page and then the branched application. If the information is present in the cache, it will directly retrieve the branched application, leveraging the cached data for improved efficiency.

With these changes, the pages API is showing improvements. A few screenshots from local environment:

Screenshot 2024-09-04 at 6 57 01 PM
Screenshot 2024-09-04 at 6 57 46 PM
Screenshot 2024-09-04 at 7 03 17 PM
Screenshot 2024-09-04 at 7 12 47 PM

Fixes https://github.com/appsmithorg/appsmith/issues/36102

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/10778438378>
> Commit: 25e71b594261a318ef6204603a225c3136892a89
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=10778438378&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Sanity`
> Spec:
> <hr>Mon, 09 Sep 2024 18:03:52 UTC

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **New Features**
	- Enhanced page retrieval capabilities with additional parameters for improved data handling.
	- Introduced a caching mechanism for optimized retrieval of application IDs in view mode.
	- Added support for bulk cache eviction through the `@CacheEvict` annotation.

- **Bug Fixes**
	- Adjusted method signatures in tests to align with updated service methods, ensuring proper functionality.

- **Documentation**
	- Updated test cases to reflect changes in method parameters for clarity and accuracy.

- **Chores**
	- Refactored method calls across various services and tests to incorporate new parameters and improve overall performance.


app/server/appsmith-server/src/main/java/com/appsmith/server/newpages/base/NewPageServiceCE.java
```diff
@@ -71,7 +69,8 @@ Mono<PageDTO> findByNameAndApplicationIdAndViewMode(
 
     Mono<String> getNameByPageId(String pageId, boolean isPublishedName);
 
-    Mono<NewPage> findByBranchNameAndBasePageId(String branchName, String defaultPageId, AclPermission permission);
+    Mono<NewPage> findByBranchNameAndBasePageId(
```
sneha122, line 72: @subrata71 I am assumming basePageId and defaultPageId are the same thing right? Confirming because method name has basePageId in it and in parameters we are naming this variable as defaultPageId

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ConsolidatedAPIServiceCEImpl.java
```diff
@@ -197,16 +213,39 @@ public Mono<ConsolidatedAPIResponseDTO> getConsolidatedInfoForPageLoad(
...
+
+        // Check if the application ID was found in the cache.
         if (isBlank(baseApplicationId)) {
+            // Cache miss detected. We need to fetch the application ID and update the cache.
+
+            // Retrieve the application ID by querying the new page service using branch name, base page ID, and
+            // application mode.
             branchedApplicationMonoCached = newPageService
                     .findByBranchNameAndBasePageIdAndApplicationMode(branchName, basePageId, mode)
-                    .map(NewPage::getApplicationId)
```
sneha122, line 203: @subrata71 I am using this same method in my PR to fetch the page and instead of getting applicationId I am using it to get pageId as I need that in order to get actions for the current page, can we please make it generic so that we can use this method to fetch either applicationId field or pageId field?

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ConsolidatedAPIServiceImplTest.java
```diff
@@ -406,16 +406,16 @@ public void testPageLoadResponseForEditMode() {
...
         Application mockApplication = new Application();
-        mockApplication.setId("mockApplicationId");
+        mockApplication.setId("mockApplicationId123");
         doReturn(Mono.just(mockApplication))
                 .when(spyApplicationService)
                 .findByBranchedApplicationIdAndApplicationMode(anyString(), any());
 
         NewPage mockNewPage = new NewPage();
-        mockNewPage.setApplicationId("mockApplicationId");
+        mockNewPage.setApplicationId("mockApplicationId123");
```
sneha122, line 415: Curious, why are we changing this id from mockApplicationId to mockApplicationId123?

app/server/appsmith-server/src/main/java/com/appsmith/server/newpages/base/NewPageServiceCE.java
subrata71: Yes, 'basePageId' and 'defaultPageId' refer to the same thing.

app/server/appsmith-server/src/main/java/com/appsmith/server/newpages/base/NewPageServiceCEImpl.java
```diff
@@ -511,19 +506,21 @@ public Mono<String> getNameByPageId(String pageId, boolean isPublishedName) {
...
     @Override
-    public Mono<NewPage> findByBranchNameAndBasePageId(String branchName, String basePageId, AclPermission permission) {
+    public Mono<NewPage> findByBranchNameAndBasePageId(
+            String branchName, String basePageId, AclPermission permission, List<String> projectedFieldNames) {
 
         if (!StringUtils.hasText(basePageId)) {
             return Mono.error(new AppsmithException(AppsmithError.INVALID_PARAMETER, FieldName.PAGE_ID));
         } else if (!StringUtils.hasText(branchName)) {
-            return this.findById(basePageId, permission)
+            return repository
```
NilanshBansal, line 515: @subrata71 Will this change affect some newrelic logging? As I can see, that in the service layer findById method we have the `FETCH_PAGE_FROM_DB` trace getting logged.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ConsolidatedAPIServiceCEImpl.java
```diff
@@ -82,6 +84,20 @@ public class ConsolidatedAPIServiceCEImpl implements ConsolidatedAPIServiceCE {
...
+     *
+     * <p>This cache uses an LRU (Least Recently Used) eviction policy and is configured to
+     * hold a maximum of 330,000 entries. Each entry consists of a key-value pair where both
+     * the key and the value are UUIDs (Version 4). Since each UUID is 16 bytes, a single
+     * key-value pair consumes 32 bytes of memory.</p>
+     *
+     * <p>The total capacity of the cache is therefore approximately 10 MB (330,000 entries
+     * * 32 bytes per entry = 10,560,000 bytes, which is roughly 10.08 MB).</p>
+     */
+    private final Cache<String, String> viewModeDefaultPageIdToDefaultAppIdLRUCache =
```
NilanshBansal, line 98: @subrata71 since this is a static cache, this will not be persisted across multiple pods right? Should we instead use an in-memory cache?

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ConsolidatedAPIServiceImplTest.java
subrata71: @sneha122 
I think I should revert this change. These test cases were failing locally and I changed these values and it started working. Let me verify and revert

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ConsolidatedAPIServiceCEImpl.java
subrata71: @NilanshBansal 
We intentionally used this in-memory cache due to its limited scope, which is restricted to view mode. Even if there’s a cache miss, the path taken is optimized compared to before. On a cache hit, it follows a more efficient path, and for subsequent requests, it continues to use this faster route. A cache miss only occurs once per pod and isn’t a major concern in my opinion.

In case of cache-miss, cache is updated [here](https://github.com/appsmithorg/appsmith/pull/36118/files#diff-e90ea02fa44fc7063128c97a99a0fb540688bb8ccd84b82f7a62a7b8b5ad275bR234-R240)

NilanshBansal: @subrata71 I think this can have an impact on the data served to app viewers. Consider the case where after deploying the app, the app url created for app viewers contain that same pageId and later on that page was deleted and app was redeployed.
In that scenario, the old app link will still continue to work because our cache would have the page entry mapped to applicationId which should not be the case.
Also, the experience here would be different for each app viewer as the request can reach different pods and each pod will have its own static cache which means some people could view the page and some people could see Page Not Found error.
Further discussions on [this](https://theappsmith.slack.com/archives/C024GUDM0LT/p1725517232433509?thread_ts=1725513570.678919&cid=C024GUDM0LT) thread.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ConsolidatedAPIServiceCEImpl.java
subrata71: I have added projection in this query and removed the map operator. This way it already becomes generic. As you have already merged your PR I will adjust it if required. @sneha122

app/server/appsmith-server/src/test/java/com/appsmith/server/services/ConsolidatedAPIServiceImplTest.java
subrata71: Reverted

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ConsolidatedAPIServiceCEImpl.java
subrata71: Hey @NilanshBansal 
I have replaced the in-memory cache implementation by Redis as we discussed and invalided the cache on deployment of application. Can you please check now?

app/server/appsmith-server/src/main/java/com/appsmith/server/newpages/base/NewPageServiceCEImpl.java
subrata71: This has the logging considering branch. We previously decided to avoid adding metrics to the repository layers, as they can be tracked separately through DB call instrumentation. However, it's instrumented in the service layer, so everything is covered.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/constants/spans/ce/ApplicationSpanCE.java
```diff
@@ -1,7 +1,10 @@
 package com.appsmith.external.constants.spans.ce;
 
+import static com.appsmith.external.constants.spans.ConsolidatedApiSpanNames.APPLICATION_ID_SPAN;
 import static com.appsmith.external.constants.spans.ConsolidatedApiSpanNames.CONSOLIDATED_API_PREFIX;
 
 public class ApplicationSpanCE {
     public static final String APPLICATION_FETCH_FROM_DB = CONSOLIDATED_API_PREFIX + "app_db";
+    public static final String APPLICATION_ID_FETCH_REDIS_SPAN = APPLICATION_ID_SPAN + ".r_redis";
+    public static final String APPLICATION_ID_UPDATE_REDIS_SPAN = APPLICATION_ID_SPAN + ".u_redis";
```
sneha122, line 9: @subrata71 instead of r_redis and u_redis, can we write it completely like read_redis and update_redis, this way reading the perf metrics data would be simpler.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationPageServiceCEImpl.java
```diff
@@ -1084,6 +1087,12 @@ protected Mono<Tuple2<Mono<Application>, ApplicationPublishingMetaDTO>> publishA
 
                     Mono<Boolean> archivePageMono;
 
+                    Mono<Boolean> evictDeletedDefaultPageIdsMono = Flux.fromIterable(publishedPageIds)
+                            .flatMap(toBeDeletedPageId -> cacheManager.evict(
```
NilanshBansal, line 1091: @subrata71 it is recommended to use @cache and @cacheEvict decorators already created. Can we please use these instead of the cacheManager directly. It helps in having consistent implementation across.

app/server/appsmith-interfaces/src/main/java/com/appsmith/external/constants/spans/ce/ApplicationSpanCE.java
subrata71: This got truncated in the traces breakdown. Anyway, I will update it accordingly

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationPageServiceCEImpl.java
subrata71: I can't actually use them in my use case as caching and evicting are done after doing some process in the data, right?

app/server/appsmith-server/src/main/java/com/appsmith/server/newpages/base/NewPageServiceCEImpl.java
```diff
@@ -511,19 +506,21 @@ public Mono<String> getNameByPageId(String pageId, boolean isPublishedName) {
     }
 
     @Override
-    public Mono<NewPage> findByBranchNameAndBasePageId(String branchName, String basePageId, AclPermission permission) {
+    public Mono<NewPage> findByBranchNameAndBasePageId(
```
sneha122, line 509: @subrata71 Curious, we have updated this method to accept projectionFields param to only send the required data, but in consolidatedAPIService, we are using `findByBranchNameAndBasePageIdAndApplicationMode` method instead, how will the added projections help for consolidatedAPI?

subrata71: Inside this method it passes the projected names. Can you please check the implementation of findByBranchNameAndBasePageIdAndApplicationMode?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ConsolidatedAPIServiceCEImpl.java
```diff
@@ -69,6 +70,8 @@
 import static com.appsmith.external.constants.spans.ConsolidatedApiSpanNames.THEMES_SPAN;
 import static com.appsmith.external.constants.spans.ConsolidatedApiSpanNames.USER_PROFILE_SPAN;
 import static com.appsmith.external.constants.spans.ConsolidatedApiSpanNames.WORKSPACE_SPAN;
+import static com.appsmith.external.constants.spans.ce.ApplicationSpanCE.APPLICATION_ID_FETCH_REDIS_SPAN;
+import static com.appsmith.external.constants.spans.ce.ApplicationSpanCE.APPLICATION_ID_UPDATE_REDIS_SPAN;
```
NilanshBansal, line 74: Can we please import these from EE file.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationPageServiceCEImpl.java
```diff
@@ -90,6 +91,7 @@
 import static com.appsmith.server.constants.CommonConstants.EVALUATION_VERSION;
 import static com.appsmith.server.helpers.ObservationUtils.getQualifiedSpanName;
 import static com.appsmith.server.helpers.ce.PolicyUtil.policyMapToSet;
+import static com.appsmith.server.services.ce.ConsolidatedAPIServiceCEImpl.VIEW_MODE_DEFAULT_PAGE_ID_TO_APP_ID_CACHE_NAME;
```
NilanshBansal, line 94: Can we please import this from EE file. 
Also, do you think we should move this to the helpers file? Usually we do not import constants defined in Impl files.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationPageServiceCEImpl.java
NilanshBansal: @subrata71 we can still use the annotatated methods, we can define our required methods in CacheableRepositoryHelper and invoke the respective method as we have invoked the cacheManager.evict() here.
Since, it is getting called through Flux.fromIterable() invoking that method should not be a concern. 
Adding @abhvsn also to check this, as he also has good context on why we prefer using Annotated methods

NilanshBansal:
Adding @abhvsn also for a quick review

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ConsolidatedAPIServiceCEImpl.java
subrata71: Can you please share the reasoning behind it? @NilanshBansal

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ApplicationPageServiceCEImpl.java
abhvsn: Agreed with Nilansh here, we can separate out the caching part in other class so as to avoid responsibility clutter in service classes. Also if we want to implement the bulk evict we can improve the `@CacheEvict` to accept the list of keys may be if it's not currently supported.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ConsolidatedAPIServiceCEImpl.java
NilanshBansal: The reasoning behind this is to apply the semantic standard. 
We don't mention imports from ce, acc to the architecture that we have created, the EE classes extend the CE ones. 

From a nomenclature perspective also, importing from ApplicationSpan is much better than to import from ApplicationSpanCE and this can be checked throughout the codebase. 

While theoretically the current code is also correct, I am highlighting more about the nomenclature part

app/server/appsmith-server/src/main/java/com/appsmith/server/newpages/base/NewPageServiceCEImpl.java
subrata71: https://theappsmith.slack.com/archives/C024GUDM0LT/p1725623463468619?thread_ts=1725599127.309689&cid=C024GUDM0LT

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CacheableRepositoryHelperCEImpl.java
```diff
@@ -199,4 +202,16 @@ public Mono<Tenant> fetchDefaultTenant(String tenantId) {
     public Mono<Void> evictCachedTenant(String tenantId) {
         return Mono.empty().then();
     }
+
+    @Cache(cacheName = CACHE_DEFAULT_PAGE_ID_TO_DEFAULT_APPLICATION_ID, key = "{#basePageId}")
+    @Override
+    public Mono<String> fetchBaseApplicationId(String basePageId, String baseApplicationId) {
```
NilanshBansal, line 208: @subrata71 AFAIK, this method is defined to fetch the applicationId for a pageId right? How do we already have the applicationId in the parameter?

app/server/appsmith-server/src/main/java/com/appsmith/server/repositories/ce/CacheableRepositoryHelperCEImpl.java
```diff
@@ -199,4 +202,16 @@ public Mono<Tenant> fetchDefaultTenant(String tenantId) {
...
+        return !StringUtils.hasText(baseApplicationId) ? Mono.empty() : Mono.just(baseApplicationId);
```
NilanshBansal, line 209: @subrata71 should there not be a logic to fetch the applicationId from the database based on the pageId, if it is not present in the cache?

subrata71: The implementation is handled within the service layer of the consolidated API. This approach is somewhat complex because the goal is to cache only the applicationId rather than the entire application document. I aimed to avoid transferring part of the business logic to the cacheable repository layer. If you’d like more details, I’d be happy to discuss further. @NilanshBansal

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/ConsolidatedAPIServiceCEImpl.java
```diff
@@ -198,28 +202,60 @@ public Mono<ConsolidatedAPIResponseDTO> getConsolidatedInfoForPageLoad(
 
         /* Fetch default application id if not provided */
         Mono<Application> branchedApplicationMonoCached;
+        Mono<String> baseApplicationIdMono = Mono.just("NA");
```
NilanshBansal, line 205: @subrata71 will using an empty string be better than "NA" for applicationId?

subrata71: Updated the logic with empty string

NilanshBansal (approved):

PR #36187, opened 2024-09-09
chore: Plugin Action Editor Context

### LINKED ISSUE #36152: [Task]: Create Plugin Action context
We will create a new react context called Plugin Action. This will be used to pass down the Plugin Action and its other details that would be needed to render the Plugin Action Editor component. This will be a common component between Apis and Queries. This context can be used inside all IDEs

hetunandu:
## Description

- Introduce the PluginActionEditor module structure
- Add basic handling and states in the PluginActionContext
- Update AppIDE to use the new Editor when the feature flag is active. This will later be updated and the component will be used from the route level itself

Fixes #36152

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/10805408539>
> Commit: 1b8259b32c0df05137b01e25d7ff942b936e8110
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=10805408539&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Datasource`
> Spec:
> <hr>Wed, 11 Sep 2024 05:53:02 UTC

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **New Features**
	- Introduced a `PluginActionEditor` component for managing plugin actions.
	- Added `PluginActionContext` for improved state management of plugin actions.
	- Implemented feature flag checks to conditionally render the redesigned action editor interface in both the `ApiEditorWrapper` and `QueryEditor` components.

- **Documentation**
	- Added centralized export files for easier access to new components and context providers.

- **Bug Fixes**
	- Enhanced error handling for missing plugin settings and configurations in the `PluginActionEditor`.


app/client/src/PluginActionEditor/PluginActionContext.tsx
```diff
@@ -0,0 +1,62 @@
...
+interface PluginActionContext {
+  action: Action;
+  editorConfig: unknown[];
+  settingsConfig: unknown[];
+  plugin: Plugin;
+  datasource?: Datasource;
+}
+
+// No need to export this context to use it. Use the hook defined below instead
+const PluginActionContext = createContext<PluginActionContext | null>(null);
```
albinAppsmith, line 20: Interface and const have same name. I suggest keeping it different.

app/client/src/PluginActionEditor/PluginActionEditor.tsx
```diff
@@ -0,0 +1,72 @@
...
+      </CenteredWrapper>
+    );
+  }
+
+  if (!action || !plugin) {
+    // Handle not found
+    return <EntityNotFoundPane />;
+  }
+  if (!settingsConfig || !editorConfig) {
+    throw Error("Plugin config for action not found");
```
albinAppsmith, line 56: Should we show the same not found component here? Or a better component to handle this error.

app/client/src/PluginActionEditor/PluginActionEditor.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { getIsEditorInitialized } from "selectors/editorSelectors";
+import Spinner from "components/editorComponents/Spinner";
+import CenteredWrapper from "components/designSystems/appsmith/CenteredWrapper";
+
+interface ChildrenProps {
+  children: React.ReactNode[];
+}
+
+const PluginActionEditor = (props: ChildrenProps) => {
+  const { pathname } = useLocation();
```
ashit-rath, line 24: With this refactor, should we rely on the path to identify the action? Can we pass the action as a prop to the `PluginActionEditor`? 
Reason for asking this is the path has to conform to a particular standard for this editor to work (path should have :entityId or :queryId). If we make this agnostic then this editor can be rendered anywhere and has no dependency on the url

app/client/src/PluginActionEditor/PluginActionEditor.tsx
```diff
@@ -0,0 +1,72 @@
...
+    throw Error("Plugin config for action not found");
+  }
+
+  return (
+    <PluginActionContextProvider
+      action={action}
+      datasource={datasource}
+      editorConfig={editorConfig}
+      plugin={plugin}
+      settingsConfig={settingsConfig}
```
ashit-rath, line 65: If you check `ModuleQueryEditor`, the `settingsConfig` is passed as a prop to the editor.

app/client/src/PluginActionEditor/PluginActionEditor.tsx
ayushpahwa: Another suggestion: AFAIK each action now has a context object to define which parent it belongs to. Maybe we can use that instead of the URL. (I don't know if this functionality exists for packages however, need to confirm with someone from backend team)

app/client/src/PluginActionEditor/PluginActionEditor.tsx
ayushpahwa: +1

ayushpahwa (commented):
How are we planning to add the customisations for toolbars (like create module option, removing edit options etc)

app/client/src/PluginActionEditor/PluginActionEditor.tsx
hetunandu: We will remove that

hetunandu:
> How are we planning to add the customisations for toolbars (like create module option, removing edit options etc)

@ayushpahwa we are planning to add that business logic closer to the actual functionality. So there will be a new component created called `AppPluginActionToolbar` that would wrap the `PluginActionToolbar` and add the changes there

app/client/src/PluginActionEditor/PluginActionEditor.tsx
ashit-rath: What I meant was the Module editors show only specific settings for each pluginType hence they are passed from top bottom to decouple and avoid if-else inside the editors. if the plan is to remove that then what's the plan to keep the editor config agnostic?
Check `useModuleFallbackSettingsForm.ts` for more info.

app/client/src/PluginActionEditor/PluginActionEditor.tsx
hetunandu: I will make this change in the next PR to update the PluginActionEditor

app/client/src/PluginActionEditor/PluginActionEditor.tsx
hetunandu: Got it. Yup I have this considered. Will make this close to the PluginActionToolbar where this will be used

ashit-rath (approved):

albinAppsmith (approved):

PR #36203, opened 2024-09-09
chore: Migration for missing datasource configuration on default rest datasources for git connected app.

sondermanish:
## Description

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags="@tag.Git, @tag.ImportExport"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/10789950948>
> Commit: 69729ce5e51ee6127f292a1dce468c492db9e413
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=10789950948&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Git, @tag.ImportExport`
> Spec:
> <hr>Tue, 10 Sep 2024 10:07:42 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
	- Enhanced application JSON migration process to support additional contextual parameters, improving accuracy and relevance.
	- Introduced a new helper class to facilitate datasource configuration during migrations.

- **Bug Fixes**
	- Improved handling of incompatible schemas during migration, ensuring robust error management.

- **Tests**
	- Adjusted test cases to accommodate changes in method signatures for migration processes, ensuring continued functionality.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/JsonSchemaMigration.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/JsonSchemaMigration.java
@@ -27,18 +27,6 @@ public class JsonSchemaMigration {
                 && (applicationJson.getServerSchemaVersion() <= jsonSchemaVersions.getServerVersion());
     }
 
-    /**
-     * This is a temporary check which is being placed for the compatibility of server versions in scenarios
-     * where user is moving a json from an instance which has
-     * release_autocommit_feature_enabled true to an instance which has the flag as false. In that case the server
-     * version number of json would be 8 and in new instance it would be not compatible.
-     * @param applicationJson
-     * @return
-     */
-    private boolean isAutocommitVersionBump(ApplicationJson applicationJson) {
-        return jsonSchemaVersions.getServerVersion() == 7 && applicationJson.getServerSchemaVersion() == 8;
-    }
-
     private void setSchemaVersions(ApplicationJson applicationJson) {
         applicationJson.setServerSchemaVersion(getCorrectSchemaVersion(applicationJson.getServerSchemaVersion()));
         applicationJson.setClientSchemaVersion(getCorrectSchemaVersion(applicationJson.getClientSchemaVersion()));
@@ -114,7 +102,7 @@ public class JsonSchemaMigration {
      * @param artifactExchangeJson : the json to be imported
      * @return transformed artifact exchange json
      */
-    @Deprecated
+    @Deprecated(forRemoval = true, since = "Use migrateArtifactJsonToLatestSchema")
     public ArtifactExchangeJson migrateArtifactToLatestSchema(ArtifactExchangeJson artifactExchangeJson) {
 
         if (!ArtifactType.APPLICATION.equals(artifactExchangeJson.getArtifactJsonType())) {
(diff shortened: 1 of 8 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/JsonSchemaMigration.java
```diff
@@ -53,24 +57,53 @@ public Mono<? extends ArtifactExchangeJson> migrateArtifactExchangeJsonToLatestS
...
+                    if (!isCompatible(appJson)) {
+                        return Mono.empty();
                     }
 
-                    return null;
+                    // Taking a tech debt over here for import of file application.
+                    // All migration above version 9 is reactive
+                    // TODO: make import flow migration reactive
+                    return Mono.just(migrateServerSchema(appJson))
+                            .flatMap(migratedApplicationJson -> {
```
nidhi-nair, line 81: Can we make this an additional systematic path similar to the json file migrations? This is a legitimate case of a new data point that needs to be added to the repository, and, although is expected fewer times than file migrations, should be accounted for as well.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/JsonSchemaMigration.java
```diff
@@ -145,11 +180,36 @@ private ApplicationJson migrateServerSchema(ApplicationJson applicationJson) {
...
+
+        return applicationJson;
+    }
+
+    /**
+     * This is non reactive way of adding migrations in the
+     * @param applicationJson
+     * @return
+     */
+    private ApplicationJson nonReactiveServerMigrationForImport(ApplicationJson applicationJson) {
```
nidhi-nair, line 202: Why is this a separate method?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/MigrationHelperMethods.java
```diff
@@ -1231,4 +1234,93 @@ public static void setThemeSettings(Application.ThemeSetting themeSetting) {
...
+        }
+
+        for (NewAction action : actionList) {
+            if (action.getUnpublishedAction() == null
+                    || action.getUnpublishedAction().getDatasource() == null) {
+                continue;
+            }
+
+            Datasource actionDatasource = action.getUnpublishedAction().getDatasource();
+            if (actionDatasource.getDatasourceConfiguration() == null
```
nidhi-nair, line 1286: Would we need this to run for ds config with empty url as well? For apps like Akhil's that got affected by the previous version?

sondermanish: we can run for that as well, but do we have any unintended consequences?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/JsonSchemaMigration.java
sondermanish: So what you mean is that let's add a reactive migration set separately itself?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/MigrationHelperMethods.java
nidhi-nair: Triggering it again for something that already exists on git is going to only get the correct value from DB. The extent of unintended consequence would be in case the user was supposed to truly see that as an uncommitted change but we went ahead and forcefully committed it. We can live with that, considering that at the current moment the data is already lost for the user, and there is no "right version" on the repo, except in their history.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/JsonSchemaMigration.java
nidhi-nair: Yes please

nidhi-nair (approved):

Issue #31108, opened 2024-02-14
[Task]: Sort workspaces on homepage in fixed alphabetical order


pranayagarwal96:
### SubTasks

Currently, workspaces on homepage are sorted in order of last edited. However, basis feedback received from users, they would rather prefer to have the workspaces in fixed order, so that they don't have to keep searching in the list every time (as the order keeps dynamically changing basis usage)

abhvsn:
Need to explicitly maintain the context when switching back to homepage, if we go ahead with this implementation https://theappsmith.slack.com/archives/C02K2MZERSL/p1709111336331609

abhvsn:
Moving this to product backlog as we don't have the bandwidth to address the issue described above around maintaining the context, and this is not going to be a backend only code change.

bookcaseballpen:
I think this sorting feature is critial.
our company ops currenlty also using appsmith admin tools. they are not tech savy. and they keep complainning the change of app locations.

bookcaseballpen:
@pranayagarwal96 
Hi, 
I am looking for a  fix for both. 
but the applicaiton order inside a workspace is more critical than than the workspace order to our team.

pranayagarwal96:
Got it - unfortunately, this is not in our near-term pipeline but we do intend to provide different sorting options to user for both apps and workspaces over time

holylander:
I'd suggest to just have them sorted alphabetically ( very easy to implement in JS), plus also adding a small simple filter box, that search by workspace name

daaliachhak17:
https://zemoso.atlassian.net/browse/TNOSB-431- ticket created in Jira

MajaharZemoso:
@pranayagarwal96 
I am picking up this issue and i will be focusing on sorting workspace in alphabetical order.

MajaharZemoso:
@Nikhil-Nandagopal @abhvsn @pranayagarwal96 
I wanted to clarify our approach: Are we planning to sort the applications also, or are we focusing solely on sorting by workspace?

holylander:
my two cents is that having an fixed sorting helps UX. Sorting by default name is a good example, this way, on the next visit, user knows where the workspaces are etc.
On the other hand, a search box by name would be also very convenient

Nikhil-Nandagopal:
@MajaharZemoso we can do it for apps next, let's start with a PR for workspaces

MarkoSulamagi:
+1 

Looking forward to the solution.

[...]

PR #36226, opened 2024-09-10
fix: added the method to sort the workspace after fetching them.

Naveen-Goud:
**Description:**
Currently, workspaces on sidebar and apps are sorted in order of last edited. We should sort them in alphabetical order

fixes: [#31108](https://github.com/appsmithorg/appsmith/issues/31108)

**updates in Pr:**
1.added a new service which return workspaces of the user in alphabetical order along with its interface
2.added this service call in controller.
3.added the test case for this changes.

**snapshot:**
image

## Summary by CodeRabbit

- **New Features**
  - Workspaces are now displayed in alphabetical order instead of by recently used, enhancing organization and accessibility for users.

- **Bug Fixes**
  - Improved retrieval method for user workspaces to ensure accurate alphabetical sorting.

- **Tests**
  - Added tests to validate the new alphabetical sorting functionality for user workspaces.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserWorkspaceServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserWorkspaceServiceCEImpl.java
@@ -32,6 +32,7 @@ import reactor.util.function.Tuple2;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


Naveen-Goud:
Hii @riteshkew @pranayagarwal96 , could you assign reviewer to this PR.

thank you.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserWorkspaceServiceCEImpl.java
```diff
@@ -414,4 +415,31 @@ public Mono<List<Workspace>> getUserWorkspacesByRecentlyUsedOrder() {
...
     }
+
+    /*
+     * Returns a list of workspaces for the current user, sorted in alphabetical order.
+     *
+     * @return Mono containing the list of workspaces
+     */
+    @Override
+    public Mono<List<Workspace>> getUserWorkspaceInAlphabeticalOrder() {
+        Mono<List<String>> workspaceIdsMono = userDataService
```
abhvsn, line 426: @Naveen-Goud considering we are now sorting entities based on the alphabetical order we can omit the step to fetch `workspaceIdsMono` which populates the list of workspaceIds from the recently used list.

abhvsn (changes requested):
@Naveen-Goud can we also do the same for fetching applications on homepage? You will find more details about this in `WorkspaceControllerCE.workspacesForHome`.

One more suggestion, as we are not confident about the user perception for this change we can utilise the feature flagging to selectively target the users. 
```
Mono<Boolean> isAphabeticalOrderingEnabled = featureFlagService.check(FeatureFlagEnum.release_alphabetical_ordering_enabled);
                    return isAphabeticalOrderingEnabled.flatMap(isEnabled -> {
                        if (isEnabled) {
                            // If alphabetical ordering is enabled, then we need to sort the pages in alphabetical order    
                        }
                        // If alphabetical ordering is disabled, then we need to sort the pages in recently used order
                    }
```

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserWorkspaceServiceCEImpl.java
Naveen-Goud: Hii @abhvsn ,I have debugged the code and observed that , Its considering all the workspaces not only the recent used  , below is the screen short, before sorting the workspaces.

image

so this method sorts all the workspaces not only the recently used.

thank you

abhvsn: @Naveen-Goud I meant we should not need `workspaceIdsMono` at all because we will not be referring to recentlyUsedEntities which are stored in the userData object.

abhvsn:
@Naveen-Goud please let us know if you are still working on this ticket, asking because we haven't seen any activity on this PR recently.

Naveen-Goud:
Hii @abhvsn , I have updated the method . could you review the changes.

below is the snap shot after the changes.

image


thank you.

PR #36312, opened 2024-09-13
feat:sort apps in a workspace in alphabetical order

MajaharZemoso:
[Issue](https://github.com/appsmithorg/appsmith/issues/31108)

## Describtion

- added a new service which return applications of the user in alphabetical order
- added this service call in controller.
- added the test case for this changes.

### Screenshots
image

## Summary by CodeRabbit

- **New Features**
	- Introduced a new method to retrieve applications within a workspace, sorted alphabetically by name.
	- Enhanced application retrieval options to improve organization and accessibility for users.

- **Bug Fixes**
	- Updated application retrieval logic to ensure applications are displayed in alphabetical order instead of by recent usage.

- **Tests**
	- Added tests to verify that applications are returned in alphabetical order within a workspace.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/applications/base/ApplicationServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/applications/base/ApplicationServiceCEImpl.java
@@ -57,6 +57,7 @@ import reactor.core.publisher.Mono;
 
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


MajaharZemoso:
@appsmithorg/query-js-pod , Please add reviewer for this pr if you have enough bandwidth.
Thanks in advance.

app/server/appsmith-server/src/main/java/com/appsmith/server/applications/base/ApplicationServiceCEImpl.java
```diff
@@ -211,6 +212,56 @@ public Flux<Application> findByWorkspaceIdAndBaseApplicationsInRecentlyUsedOrder
...
+                        .filter(application -> {
+                            /*
+                             * Filter applications based on the following criteria:
+                             * - Applications that are not connected to Git.
+                             * - Applications that, when connected, revert with default branch only.
+                             */
+                            return !GitUtils.isApplicationConnectedToGit(application)
+                                    || GitUtils.isDefaultBranchedApplication(application);
+                        })
+                        .sort(Comparator.comparing(
```
mohanarpit, line 260: Instead of sorting in application memory, we should perform this operation during the database query. In this scenario, we should query MongoDB to get the applications for the workspace in a sorted manner. This is because it's much faster to perform this operation in a database rather than in application memory. Also, moving such ops to the DB ensures long-term stability of this function.

MajaharZemoso: Thank you for the suggestion! I've updated the code to perform sorting at the MongoDB query level. Please let me know if further adjustments are needed.

abhvsn:
@MajaharZemoso thanks for your interest but we have received another contribution for the same issue, and decided to go ahead with the other PR as the other one was raised earlier. 
You can look for other `good first issues`. Sorry again we should have assigned this ticket to single owner before asking for the contribution to avoid this situation.

Issue #35794, opened 2024-08-20
Date picker shouldn't appear on top of the date field


carinanfonseca:
Steps:
1. Make sure you have an editable date-type column
2. Click on the edit icon 
3. The date picker appears on top of the field hiding the field, which is not a good experience. Expected: It should appear on the right.

[Private Zenhub Video](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMDV3Qmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4943418b584a84a1d00755907315f4f8802951e4/Screen%20Recording%202024-08-19%20at%2011.46.36.mov)

deepthin3:
Hi - We are not able to access the video - It is giving the error "Permission denied". Can you please help with a loom video

venkatalakshmimaddu:
@carinanfonseca I am picking up this issue

venkatalakshmimaddu:
@carinanfonseca I have rasied the pr for this issue 
[pr link](https://github.com/appsmithorg/appsmith/pull/36359)
It seems no reviewer has been assigned yet. Could you please assign reviewer for this pr

ALOK9442:
hey @carinanfonseca @Nikhil-Nandagopal , if no one is working on this issue, can you assign it to me?

PR #36359, opened 2024-09-17
fix: Date picker shouldn't appear on top of the date field

venkatalakshmimaddu:
Fixes: #35794 

Hi @carinanfonseca 

**What's in the pr :** 
- Initially, the date picker was appearing on top of the date field.
- Updated the date picker component by adding props to adjust the popover position to the right.
- Now, the date picker popover consistently appears on the right side of the date field.

**Before :** 

image

**After :**

image

## Summary by CodeRabbit

- **New Features**
	- Enhanced the Date Picker component with improved popover positioning and behavior.
	- The popover now stays within viewport boundaries and does not flip when overflowing.

### DIFF
--- a/app/client/src/widgets/DatePickerWidget2/component/index.tsx
+++ b/app/client/src/widgets/DatePickerWidget2/component/index.tsx
@@ -2,7 +2,7 @@ import React from "react";
 import styled from "styled-components";
 import { IntentColors } from "constants/DefaultTheme";
 import type { IRef, Alignment } from "@blueprintjs/core";
-import { ControlGroup, Classes } from "@blueprintjs/core";
+import { ControlGroup, Classes, Position } from "@blueprintjs/core";
 import type { ComponentProps } from "widgets/BaseComponent";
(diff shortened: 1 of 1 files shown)


app/client/src/widgets/DatePickerWidget2/component/index.tsx
```diff
@@ -390,6 +390,11 @@ class DatePickerComponent extends React.Component<
                 usePortal: !this.props.withoutPortal,
                 canEscapeKeyClose: true,
                 portalClassName: `${DATEPICKER_POPUP_CLASSNAME}-${this.props.widgetId}`,
+                position: Position.RIGHT,
```
sagar-qa007, line 393: Any test needs to add or updated here?

sagar-qa007 (changes requested):

app/client/src/widgets/DatePickerWidget2/component/index.tsx
venkatalakshmimaddu: No need of test cases

PR #36412, opened 2024-09-19
chore: Refactor API

### LINKED ISSUE #36481: [Task]: Refactor APIs for better types
### Problem 1

Consider the following API code ( PagesAPI for e.g ):

```js
class PageApi extends Api {
  static async fetchPage(
    pageRequest: FetchPageRequest,
  ): Promise<AxiosPromise<FetchPageResponse>> {
    ...
  }

  static savePage(
    request: SavePageRequest,
  ): AxiosPromise<SavePageResponse> | undefined {
    ...
  }
  ```
  
 For every function, we define the return type explicitly and repeat the `Promise<AxiosPromise<...>>` part . This is not actually required. Axios as an library is properly typed. It supports generics to tell the types of the response.
We just need to pass the type through generics:

```js
function fetchPage(pageRequest: FetchPageRequest) {
 return api.get<FetchPageResponse>(url);
}
```

### Problem 2

Another problem with the types is that what we are defining explicitly for API functions is wrong. In the interceptors, we manipulate axios response to return `.data` part so that we can just do `response.data` instead of `response.data.data`. But this manipulation is not reflected in types which makes all the developers define types explicitly everywhere.

### Problem 3

There is another problem with the code, there is no need to define class when all the functions are just static. We can simply just export functions instead of adding static members to the class. 

So the API refactor has following tasks:
 - Refactor the code around APIs so that we don't use explicit return types instead we used axios generics to tell the type of the resposne.
 - Refactor the API to only use normal functions instead of static functions of a class.
 - Fixes around response to match what we do in the interceptor.

jsartisan:
Fixes #36481 

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/11028542060>
> Commit: ed537d3958a3eba4502cbc32daf60c4cd814002d
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=11028542060&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Wed, 25 Sep 2024 08:56:31 UTC

## Summary by CodeRabbit

- **New Features**
	- Enhanced API interaction with new default configurations for requests.
	- Improved error handling with a centralized interceptor for managing various API response errors.
	- Introduced access control for specific API endpoints through blocked and enabled route management.
	- Streamlined environment-specific configurations for better maintainability.
	- Added functionalities for managing application themes, including fetching, updating, and deleting themes.
	- Introduced new API functions for retrieving consolidated page load data for viewing and editing.
	- Centralized access point for API services related to theming and consolidated page load data.
	- New modular structure for API request and response interceptors to improve organization and maintainability.

- **Tests**
	- Added unit tests for both API response and request interceptors to ensure correct functionality and error handling.


ayushpahwa (commented):
Hey, I couldn't find the context for the change to give review. Will it be possible to add some info+related issue number for better understanding. In case, the PR is still WIP would suggest moving to draft to avoid notifications.

app/client/src/api/Api.ts
```diff
@@ -21,16 +19,7 @@ export const apiRequestConfig = {
 
 const axiosInstance: AxiosInstance = axios.create();
 
-const requestInterceptors = [
```
jsartisan, line 24: reusing the refactored interceptors.

jsartisan:
@ayushpahwa Hey, here is the context - https://github.com/appsmithorg/appsmith/issues/36481

app/client/src/ce/api/ApiUtils.ts
```diff
@@ -1,289 +1,7 @@
...
-    return Promise.resolve({ data: null, status: 200 });
-  }
-
-  return config;
-};
-
-// Request interceptor will add a timer property to the request.
-// this will be used to calculate the time taken for an action
-// execution request
-export const apiRequestInterceptor = (config: AxiosRequestConfig) => {
```
jsartisan, line 97: Refactored all these to tiny interceptors for better readability.

app/client/src/sagas/AppThemingSaga.tsx
```diff
@@ -75,11 +80,10 @@ export function* fetchAppThemes(action: ReduxAction<FetchAppThemesAction>) {
   try {
     const { applicationId, themes } = action.payload;
 
-    const response: ApiResponse<AppTheme> = yield call(
-      getFromServerWhenNoPrefetchedResult,
-      themes,
-      async () => ThemingApi.fetchThemes(applicationId),
-    );
+    const response: SagaReturnType<typeof AppThemingApi.fetchThemes> =
```
jsartisan, line 83: This is important, Instead of defining the type explicitly, we can cheat and grab the type from the functions with `SagaReturnType` and passing the function in it.

app/client/src/api/core/api.ts
```diff
@@ -0,0 +1,28 @@
+import type { AxiosResponseData } from "api/types";
+
+import { apiFactory } from "./factory";
+
+const apiInstance = apiFactory();
+
+export async function get<T>(...args: Parameters<typeof apiInstance.get>) {
+  return apiInstance.get<T, AxiosResponseData<T>>(...args);
```
jsartisan, line 8: Note that we pass `AxiosResponseData` as the second parameter to the generic. The reason is we intercept the axios response and return `data` of it ( so that we just need to do `.data` instead `.data.data`. So we need to pass the response type to match what we do in interceptor.

hetunandu (dismissed):
LGTM

app/client/src/api/core/api.ts
ayushpahwa: Would it make sense to add this as comment in the function?

ayushpahwa (approved):

app/client/src/api/core/api.ts
jsartisan: yea. it makes sense. Let me add it.

ayushpahwa (approved):

PR #36453, opened 2024-09-20
chore: change the jsonl dump path to appsmith stacks for pg

AnaghHegde:
## Description
The jsonl files are dumped to the appsmith-stack directory. This to avoid manually copying the dump files to appsmith-stacks since the /opt/appsmith is not a persistent storage. This will cause issues in cluster or HA set up.


abhvsn (changes requested):
@AnaghHegde we will have to update the path which we look for when MongoLoadData migration runs. For all the existing instances we will have to cleanup the specific entry otherwise we may end up with checksum mismatch.

AnaghHegde:
@abhvsn This condition already exists in the loadMongoData migration class.

abhvsn:
> @abhvsn This condition already exists in the loadMongoData migration class.

Rechecked an yes you are right, this is already handled. Another issue, we are using the same script to generate the baseline data which is generated from local and then copy pasted to `/opt/appsmith` as the appsmith-stacks directory is not present. Can we handle this scenario to copy the data based on where the script is invoked?

PR #36455, opened 2024-09-20
fix: Invalid Date Display in Table Widget's Date Column When Using Unix Timestamp (ms)

### LINKED ISSUE #35631: [Bug]: incorrect Unix Timestamp value when updating a Date type column

### Description

When I update a Date type column in a table with data extracted from a database, the Unix timestamp value returned is incorrect.

E.g., when setting 2024-08-13 in the datepicker, the updated value internally is `8iAug 13, 2024i02con20`.

This makes impossible the update of the date for any table row.

### Steps To Reproduce

1. Create a table widget.
2. Extract data from a DB with any date column.
3. For that date column, set type as Date, date format as Unix Timestamp (ms) and display format as YYYY-MM-DD
4. Make the column editable.
5. Change the value of the column using the datepicker.
6. Now, if you check the computed value, it is incorrect.

imagen

### Public Sample App

_No response_

### Environment

Production

### Severity

### Issue video log

_No response_

### Version

Appsmith v1.31.1

jacquesikot:
## Description

**Problem**
When populating a table widget with data that includes dates in Unix timestamp (milliseconds) format, setting the column type to "Date," the input format to "Unix timestamp (ms)," and selecting a display format leads to an issue during inline editing. While the date picker behaves correctly, selecting a new date results in the table cell showing an "Invalid Date" error.

**Root Cause**
The platform currently uses DateInputFormat.MILLISECONDS for Unix timestamp (ms) formatting. However, this value is not a valid option for the moment.format() function, which expects the input format to be 'x' for Unix timestamps in milliseconds. This mismatch leads to the "Invalid Date" error.

**Solution**
Modify the logic to map DateInputFormat.MILLISECONDS to the correct moment format string 'x'.
Adjust the table's transformDataPureFn to correctly process and display dates in Unix timestamp (ms) format after inline edits, ensuring the moment library can handle the input properly.

**Outcome**
This fix ensures that when a user selects a date via the date picker in inline editing mode, the selected date is displayed correctly in the table without any errors.

Fixes #35631

## Automation

/ok-to-test tags="@tag.Sanity, @tag.Binding, @tag.Table, @tag.Datepicker"

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No


app/client/cypress/e2e/Regression/ClientSide/Widgets/TableV2/Date_column_editing_2_spec.ts
```diff
@@ -92,5 +92,114 @@ describe(
           expect($textData).to.include("YYYY-MM-DDTHH:mm:ss.SSSZ"),
         );
     });
+
+    it("4. should allow input format of Unix Timestamp(ms) and not throw Invalid Value error when inline editing", () => {
+      // Get string format for tomorrow date - Sat Sep 21 2024 format
+      const tomorrow = new Date();
+      tomorrow.setDate(tomorrow.getDate() + 1);
+      const formattedTomorrowDateVerbose = tomorrow
```
sagar-qa007, line 100: Can we please add this in helper function with better parameters to make a good usage of it?

app/client/cypress/e2e/Regression/ClientSide/Widgets/TableV2/Date_column_editing_2_spec.ts
```diff
@@ -92,5 +92,114 @@ describe(
...
+
+      EditorNavigation.SelectEntityByName("Table1", EntityType.Widget);
+
+      propPane.NavigateBackToPropertyPane();
+
+      // Update table data
+      propPane.UpdatePropertyFieldValue(
+        "Table data",
+        `
+        {{
```
sagar-qa007, line 122: Can we please existing fixture here? or atleast add this in fixture folder and update in test.

app/client/src/widgets/TableWidgetV2/component/cellComponents/DateCell.tsx
```diff
@@ -235,8 +240,11 @@ export const DateCell = (props: DateComponentProps) => {
     setShowRequiredError(false);
     setHasFocus(false);
 
-    const formattedDate = date ? moment(date).format(inputFormat) : "";
-
+    if (inputFormat === DateInputFormat.MILLISECONDS)
+      momentAdjustedInputFormat = MomentDateInputFormat.MILLISECONDS;
+    const formattedDate = date
+      ? moment(date).format(momentAdjustedInputFormat)
+      : "";
```
sagar-qa007, line 247: Query : Blank value return means error will come or what will be the output?

sagar-qa007 (changes requested):

app/client/src/widgets/TableWidgetV2/widget/reactTableUtils/transformDataPureFn.tsx
```diff
@@ -38,14 +38,16 @@ export const transformDataPureFn = (
...
                   : column.metaProperties.inputFormat;
-
                 if (
                   type !== DateInputFormat.EPOCH &&
                   type !== DateInputFormat.MILLISECONDS
                 ) {
                   inputFormat = type;
                   moment(value as MomentInput, inputFormat);
-                } else if (!isNumber(value)) {
+                } else if (
```
rahulbarwal, line 47: This condition and the condition above is confusing, can we simplify this logic?
Also, the top of this function there is a comment stating the need to unit test this, while we are at it, lets take care of that as well.
One suggestion, right now the variable `type` is of `any`, which is not ideal, either declare an interface for it, or explain in comments what all values can it assume.

app/client/src/widgets/TableWidgetV2/component/cellComponents/DateCell.tsx
rahulbarwal: empty value will come. essentially, the date itself is not provided, it is a falsey value(null/undefined/""/0)

app/client/src/widgets/TableWidgetV2/component/cellComponents/DateCell.tsx
```diff
@@ -235,8 +240,11 @@ export const DateCell = (props: DateComponentProps) => {
     setShowRequiredError(false);
     setHasFocus(false);
 
-    const formattedDate = date ? moment(date).format(inputFormat) : "";
-
+    if (inputFormat === DateInputFormat.MILLISECONDS)
+      momentAdjustedInputFormat = MomentDateInputFormat.MILLISECONDS;
```
rahulbarwal, line 244: What is rationale behind this condition? Its unclear; add comment.
also, we have an if statement, lets put it inside block scope `{}`.

rahulbarwal (changes requested):

app/client/cypress/e2e/Regression/ClientSide/Widgets/TableV2/Date_column_editing_2_spec.ts
jacquesikot: This is done, and implemented inside helper.ts in the TableV2 cypress tests folder.

app/client/cypress/e2e/Regression/ClientSide/Widgets/TableV2/Date_column_editing_2_spec.ts
jacquesikot: moved to a new file fixtures.ts for TableV2 cypress tests

app/client/src/widgets/TableWidgetV2/component/cellComponents/DateCell.tsx
jacquesikot: Comments are added and block scope is added

app/client/src/widgets/TableWidgetV2/widget/reactTableUtils/transformDataPureFn.tsx
jacquesikot: Unit test has also been added

app/client/src/widgets/TableWidgetV2/widget/reactTableUtils/fixtures.ts
```diff
@@ -0,0 +1,145 @@
...
+    lll: "2024-09-25", // LLL format to date
+  },
+  {
+    epoch: "2024-09-22", // Converted from epoch to date
+    milliseconds: "2024-09-22", // Converted from milliseconds to date
+    iso_8601: "2024-09-23", // ISO 8601 to date
+    yyyy_mm_dd: "2024-09-23", // No transformation needed
+    lll: "2024-09-23", // LLL format to date
+  },
+];
```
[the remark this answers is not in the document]
rahulbarwal: @jacquesikot see if this makes sense.

app/client/src/widgets/TableWidgetV2/component/cellComponents/DateCell.tsx
```diff
@@ -235,7 +241,16 @@ export const DateCell = (props: DateComponentProps) => {
     setShowRequiredError(false);
     setHasFocus(false);
 
-    const formattedDate = date ? moment(date).format(inputFormat) : "";
+    // If the input format is milliseconds or epoch, convert the date to the required format
+    if (inputFormat === DateInputFormat.MILLISECONDS) {
+      momentAdjustedInputFormat = MomentDateInputFormat.MILLISECONDS;
+    } else if (inputFormat === DateInputFormat.EPOCH) {
+      momentAdjustedInputFormat = MomentDateInputFormat.SECONDS;
+    }
```
rahulbarwal, line 249: Nit: this can be refactored as a separate function.

app/client/cypress/e2e/Regression/ClientSide/Widgets/TableV2/Date_column_editing_2_spec.ts
```diff
@@ -92,5 +94,46 @@ describe(
           expect($textData).to.include("YYYY-MM-DDTHH:mm:ss.SSSZ"),
         );
     });
+
+    it("4. should allow input format of Unix Timestamp(ms) and not throw Invalid Value error when inline editing", () => {
```
rahulbarwal, line 98: Q: what other types of values are supported? If there are more than 2 lets test all of them one by one.

rahulbarwal (commented):
Minor suggestions.

rahulbarwal (commented):
Minor suggestions.

app/client/src/widgets/TableWidgetV2/component/cellComponents/DateCell.tsx
rahulbarwal: then finally the assignment will look like:
```js
function refactoredMethod (inputFormat: TYPE) {
    // If the input format is milliseconds or epoch, convert the date to the required format
    if (inputFormat === DateInputFormat.MILLISECONDS) {
      return MomentDateInputFormat.MILLISECONDS;
    } else if (inputFormat === DateInputFormat.EPOCH) {
      return MomentDateInputFormat.SECONDS;
    }
    return inputFormat;
}

const momentAdjustedInputFormat = refactoredMethod(inputFormat);
```

jacquesikot: This change has been implemented

app/client/src/widgets/TableWidgetV2/widget/reactTableUtils/transformDataPureFn.tsx
```diff
@@ -45,8 +44,6 @@ export const transformDataPureFn = (
                 ) {
                   inputFormat = type;
                   moment(value as MomentInput, inputFormat);
-                } else if (!isNumber(value)) {
-                  isValidDate = false;
```
rahulbarwal, line 49: If we remove this, then essentially there is no use of `isValidDate`, right? 

The only part where we make it false, is in the `catch` block. there are less chances of it entering that. 

We can either get rid of variable or check its meaning and identify if it should be false.

app/client/cypress/e2e/Regression/ClientSide/Widgets/TableV2/Date_column_editing_2_spec.ts
jacquesikot: All values have been added to a new test file - Date_column_types_validation_spec

app/client/src/widgets/TableWidgetV2/widget/reactTableUtils/transformDataPureFn.tsx
jacquesikot: The logic is expecting EPOCH and MILLISECONDS to return as a number, which it doesn't, it only returns as a string, making the logic there redundant.

app/client/src/widgets/TableWidgetV2/widget/reactTableUtils/fixtures.ts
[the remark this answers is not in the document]
jacquesikot: @rahulbarwal This is not correct from coderabbitai. The epoch values evaluate to 2022

rahulbarwal (approved):
LGTM. 
@sagar-qa007 would appreciate your review of cypress changes.

app/client/cypress/e2e/Regression/ClientSide/Widgets/TableV2/Date_column_types_validation_spec.ts
```diff
@@ -0,0 +1,167 @@
...
+    before(() => {
+      entityExplorer.DragNDropWidget("tablewidgetv2", 350, 500);
+      EditorNavigation.SelectEntityByName("Table1", EntityType.Widget);
+      propPane.ToggleJSMode("Table data", true);
+      propPane.UpdatePropertyFieldValue("Table data", tableV2DateTestData);
+    });
+
+    afterEach(() => {
+      propPane.NavigateBackToPropertyPane(false);
+    });
```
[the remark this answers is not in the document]
sagar-qa007: @jacquesikot can you please check why we need afterEach here?

sagar-qa007 (changes requested):

app/client/cypress/e2e/Regression/ClientSide/Widgets/TableV2/fixtures.ts
```diff
@@ -0,0 +1,69 @@
+export const tableV2DateTestData = `
```
sagar-qa007, line 1: Please move this fixture in fixture folder.

app/client/cypress/e2e/Regression/ClientSide/Widgets/TableV2/helpers.ts
```diff
@@ -0,0 +1,33 @@
+/**
```
sagar-qa007, line 1: Please move this helper file in helper folder.

sagar-qa007 (changes requested):

app/client/cypress/e2e/Regression/ClientSide/Widgets/TableV2/fixtures.ts
jacquesikot: Done

app/client/cypress/e2e/Regression/ClientSide/Widgets/TableV2/helpers.ts
jacquesikot: Cannot find any dedicated helper folder - except the helper folder for table widget

rahulbarwal (approved):

rahulbarwal (dismissed):

app/client/cypress/e2e/Regression/ClientSide/Widgets/TableV2/helpers.ts
sagar-qa007: You can use this : app/client/cypress/support/Pages/Table.ts

jacquesikot: This is done

sagar-qa007 (dismissed):

sagar-qa007 (approved):

PR #36480, opened 2024-09-23
feat: added LRU cache for mockDB connections

### LINKED ISSUE #36474: [Task]: Implement LRU caching to cache and evict mock mongoDB connections
### SubTasks

Reason: We have recently seen a surge in the number of connections to the mongo movies mockDB leading to the max limit reached.
This directly impacted the cypress tests and the usage of the mockDB for the release, prod and self hosted instances.
 
On debugging we found the mockDB connections were not getting closed, hence to resolve this we need to implement the LRU caching for the mockDB connections and close the least recently used connections.

Ref: https://theappsmith.slack.com/archives/C02MUD8DNUR/p1726855216593569?thread_ts=1725892395.532789&cid=C02MUD8DNUR

NilanshBansal:
## Description
This PR implements an LRU (Least Recently Used) caching mechanism for the Mongo mockDB connections which get auto-cleaned up every 2 hours based on their access time. 
This is done to stop the overpopulation of the open dangling connections to the mockDB resulting in the max connection limit.
The Caching Implementation used is [Google Guave Cache](https://github.com/google/guava/wiki/CachesExplained). 
Also refer - [Working of Guava cache ](https://medium.com/@alxkm/introduction-to-caching-with-google-guava-a-simple-and-flexible-solution-2c721427e72e)

image

Fixes #36474

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/11066462645>
> Commit: 38fcf572b32f1b5d7544828ccf00f2b6fbaa180e
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=11066462645&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Sanity`
> Spec:
> <hr>Fri, 27 Sep 2024 08:01:28 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
	- Introduced a new constant for the MongoDB plugin to enhance plugin identification.
	- Added a `DatasourcePluginContext` class to encapsulate datasource plugin context, including connection details and creation time.
	- Implemented a caching mechanism for datasource contexts to optimize connection management and reduce excessive database connections.
	- Added functionality to identify mock MongoDB datasources.

- **Bug Fixes**
	- Enhanced cleanup processes for stale connections in the caching system.


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
```diff
@@ -158,9 +212,15 @@ public Mono<DatasourceContext<Object>> getCachedDatasourceContextMono(
...
                                         datasourceContext)
                                 .cache(); /* Cache the value so that further evaluations don't result in new connections */
+                        DatasourcePluginContext<Object> datasourcePluginContext = new DatasourcePluginContext<>();
+                        datasourcePluginContext.setConnection(datasourceContext.getConnection());
+                        datasourcePluginContext.setPluginId(plugin.getId());
 
                         if (datasourceContextIdentifier.isKeyValid() && shouldCacheContextForThisPlugin(plugin)) {
+                            datasourceContextMap.put(datasourceContextIdentifier, datasourceContext);
                             datasourceContextMonoMap.put(datasourceContextIdentifier, datasourceContextMonoCache);
+                            datasourcePluginContextMapLRUCache.put(
```
NilanshBansal, line 222: WIP: To check for the mock mongo datasource connections and add only them to the `datasourcePluginContextMapLRUCache`

NilanshBansal: Done

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
```diff
@@ -67,6 +89,50 @@ public DatasourceContextServiceCEImpl(
...
+        log.debug(
+                "Removing Datasource Context from cache and closing the open connection for DatasourceId: {} and environmentId: {}",
+                datasourceContextIdentifier.getDatasourceId(),
+                datasourceContextIdentifier.getEnvironmentId());
+
+        // Close connection and remove entry from both cache maps
+        final Object connection = datasourceContextMap.containsKey(datasourceContextIdentifier)
+                        && datasourceContextMap.get(datasourceContextIdentifier) != null
+                ? datasourceContextMap.get(datasourceContextIdentifier).getConnection()
+                : null;
```
subrata71, line 112: Nit: Can you please improve the readability by moving this logic to a method?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
```diff
@@ -67,6 +89,50 @@ public DatasourceContextServiceCEImpl(
...
+                pluginExecutorHelper
+                        .getPluginExecutor(pluginMono)
+                        .flatMap(
+                                pluginExecutor -> Mono.fromRunnable(() -> pluginExecutor.datasourceDestroy(connection)))
+                        .onErrorResume(e -> {
+                            log.error("Error destroying stale datasource connection", e);
+                            return Mono.empty();
+                        })
+                        .subscribe(); // Trigger the execution
+            } catch (Exception e) {
```
subrata71, line 127: QQ: Isn't the try-catch block redundant here since errors that occur during the reactive pipeline will be handled by the onErrorResume operator.?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
```diff
@@ -29,15 +34,32 @@
...
+     * This cache is used to store the datasource context for a limited time and then destroy the least recently used
+     * connection.The cleanup process is performed after every 2 hours.
+     * The purpose of this is to prevent the large number of open dangling connections to the movies mockDB.
+     * The removalListener method is called when the connection is removed from the cache.
+     */
+    protected final Cache<DatasourceContextIdentifier, DatasourcePluginContext> datasourcePluginContextMapLRUCache =
+            CacheBuilder.newBuilder()
+                    .removalListener(createRemovalListener())
+                    .expireAfterAccess(2, TimeUnit.HOURS)
+                    .build();
```
subrata71, line 61: This is an LRU cache with unbounded capacity. What happens if the maximum number of connections is reached before the 2-hour expiration? Will the system handle situations where the cache grows indefinitely, potentially leading to resource exhaustion? Do you think that it's worth considering implementing a maximum size limit for the cache to prevent this issue.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
```diff
@@ -308,6 +391,8 @@ public Mono<DatasourceContext<?>> getDatasourceContext(DatasourceStorage datasou
         } else {
             if (isValidDatasourceContextAvailable(datasourceStorage, datasourceContextIdentifier)) {
                 log.debug("Resource context exists. Returning the same.");
+                log.debug("Accessing the LRU cache to update the last accessed time");
```
subrata71, line 394: Nit: Is this log really necessary?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
subrata71: Can we write a test to verify this change?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
NilanshBansal: Yes, the try-catch is redundant here @subrata71. Updating this 👍

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
NilanshBansal: Not necessary, I am updating this to a code comment.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
NilanshBansal: > This is an LRU cache with unbounded capacity. What happens if the maximum number of connections is reached before the 2-hour expiration? Will the system handle situations where the cache grows indefinitely, potentially leading to resource exhaustion? Do you think that it's worth considering implementing a maximum size limit for the cache to prevent this issue.

Thanks for highlighting this @subrata71, very valid point but I have purposely not set the size of this cache. The reasoning behind it is to use a much shorter `expireAfterAccess` time i.e. 2 hours. As of today, our current connection limit gets hit in about 72 hours of non-stop pods running. With the current implementation this expiryTime is much much shorter. We are not seeing such increase in usage for mockDB to cap it in a size limit. For the same reason, I have not set the explicitly set the cache size. 
Also, if in future we make changes to the `DatasourcePluginContext` to add more data to the cache, the size limit may affect the ability to cache. 
But saying that, it is just a configuration, if required we can update it anytime. What are your thoughts on this @subrata71?

subrata71 (commented):
I overlooked this point during my review. This is an in-memory cache, not shared across different pods. Doesn’t that mean the cache will only be effective for subsequent requests to the same pod? Each pod will have its own cache, which could lead to excessive caching and undermine the original goal.

NilanshBansal:
> I overlooked this point during my review. This is an in-memory cache, not shared across different pods. Doesn’t that mean the cache will only be effective for subsequent requests to the same pod? Each pod will have its own cache, which could lead to excessive caching and undermine the original goal.

@subrata71 yes the cache is not shared across the different pods and as per the current and projected traffic on mockDB connections we are reaching the limit of 3k connections in 72 hours combined on all pods.
 The eviction time added in the above configuration is 2 hours per pod which can only lead to excessive caching only when there are 3k connections in less than 2 hours combined on all pods which is not the case that we are projecting. 
 Saying that, do you still find the need to add a cache size limit @subrata71 ?

subrata71:
@NilanshBansal 
Could you please add more logging to track when the cache expires? Including information like the cache size would help with monitoring across pods and allow for better adjustments to the caching implementation.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
NilanshBansal: As per discussion with @subrata71, asserting this in a test will be very challenging and connections being used from the other maps is already being asserted. Hence, we are proceeding with the following: 
1. Adding size limit to the cache.
2. Adding logs to know the cache size after addition and eviction.

NilanshBansal:
As per discussion with @subrata71, I have added the maximum size limit of the cache per pod and also added the logs around the cache size after addition and eviction. 
@subrata71 can you please re-review 🙏

NilanshBansal:
Tag.All has already passed on this PR hence updated with Tag.Sanity as only log level changes are done post it.
SCR-20240927-i0c

subrata71 (approved):
LGTM

PR #36482, opened 2024-09-23
fix: batched evaluate action params to a chunk size

### LINKED ISSUE #36383: [Bug]: The stack trace limit gets exceeded  if the data send to the execute call is large

### Description

When we have an action with large number of data to be passed to the execute action call, it loops to a function `updateBlobDataFromUrls` which performs some functionality and then stores that value in an array. If the length of the array to be looped is quite large then the stack trace limit gets exceeded which results in no execute call getting fired or anything further.

### Steps To Reproduce

1. Create an api action which takes an object and one of the keys should accept array of strings i.e. `{ "sids": ["hello", "hello2"] }`
2. Run the query with array items more than 800 strings
3. Observe the execute calls never goes in and the loader icon gets stuck for a particular timeout

### Public Sample App

_No response_

### Environment

Production

### Severity

Low (Cosmetic UI issues)

### Issue video log

_No response_

### Version

Both

AmanAgarwal041:
## Description

In this PR, we fix the payload evaluation issue by adding a batching mechanism. 

Root cause the issue here was that using yield call in loop tends to call stack overflow which is a known open issue in the redux-saga library. The popular workaround to this is to add `yield delay(0` after n number of batch. 

Fixes #36383

## Automation

/test all

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/11134117696>
> Commit: a3a583f880cb9d057a2f4668f1ccb0fc10714321
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=11134117696&attempt=2" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Wed, 02 Oct 2024 01:49:36 UTC

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **Improvements**
	- Enhanced processing efficiency by introducing a batching mechanism, allowing smoother handling of large data sets without blocking the event loop.


ayushpahwa (approved):

rishabhrathod01:
Merging this as the planned unit test couldn't help in asserting the call stack issue.

PR #36605, opened 2024-09-30
chore: Datepicker Widget

jsartisan:
## Summary by CodeRabbit

- **New Features**
  - Introduced a new `Calendar` component for enhanced date selection.
  - Added a `DatePicker` component with customizable options and validation.
  - Implemented a `FieldCalendarPopover` for displaying calendars in a dialog format.
  - New icons and thumbnails for the DatePicker have been added.

- **Bug Fixes**
  - Improved validation logic for date selection in the DatePicker.

- **Documentation**
  - Added Storybook stories for the DatePicker component to demonstrate various use cases.

- **Style**
  - Introduced new styles for the calendar and date picker components for better UI presentation.

- **Chores**
  - Updated configuration files for the new DatePicker widget to enhance functionality and accessibility.

> [!WARNING]
> Tests have not run on the HEAD 4daa071d4c5f37059d592786ade0270badbffe05 yet
> <hr>Mon, 30 Sep 2024 12:26:29 UTC


app/client/packages/storybook/.storybook/preview-head.html
```diff
@@ -11,16 +11,6 @@
     font-family: var(--font-family) !important;
   }
 
-  th, td,
```
jsartisan, line 14: Removed this as it was interfering with calender cells styles.

app/client/src/widgets/WidgetUtils.ts
```diff
@@ -989,3 +989,45 @@ export const checkForOnClick = (e: React.MouseEvent<HTMLElement>) => {
...
+ *  isValidDate: "{{ this.value === 1 }}"
+ * }
+ * ```
+ *
+ * Main rule to remember is don't use deconstruct the props like `const { value } = props;` in the derived property function. Directly access props like `props.value`
+ * (Reason is we replace props.value with this.value in the derived property function)
+ * @param propertyFns
+ * @returns
+ */
+export function parseDerivedProperties(propertyFns: Record<string, unknown>) {
```
jsartisan, line 1017: @riodeuno Need your review on this. 
So, in all widgets, we used to use `parseDerivedProperties` and copy it to almost every widget (  A sign that it must be abstracted to a common function. ). I tried to abstract that logic to a separate function that can be used. Instead of loading the `derived.js` file with raw loader, we can just simply load it and parse it. What do you think of this approach?

app/client/src/modules/ui-builder/ui/wds/WDSInputWidget/widget/index.tsx
```diff
@@ -52,8 +53,10 @@ class WDSInputWidget extends WDSBaseInputWidget<InputWidgetProps, WidgetState> {
   }
 
   static getDerivedPropertiesMap(): DerivedPropertiesMap {
+    const parsedDerivedProperties = parseDerivedProperties(derivedPropertyFns);
```
jsartisan, line 56: @riodeuno This is how we can use it.

app/client/packages/design-system/widgets/src/components/Calender/src/CalendarHeading.tsx
jsartisan, line 1: Need this component because calender consumes HeadingContext to display the months in popover header. so had to create this so that I could wrap the text in our Text component.

app/client/packages/design-system/widgets/src/components/DatePicker/src/DatePicker.tsx
```diff
@@ -0,0 +1,70 @@
...
+import React from "react";
+import {
+  DateInput,
+  DateSegment,
+  Group,
+  DatePicker as HeadlessDatePicker,
+} from "react-aria-components";
+
+import type { DatePickerProps } from "./types";
+import styles from "../../ComboBox/src/styles.module.css";
```
jsartisan, line 19: Just reusing the styels from combobox.

app/client/packages/design-system/widgets/src/components/FieldCalenderPopover/src/FieldCalenderPopover.tsx
jsartisan, line 1: IMO, this component is just useless. I created it because we created FieldListPopover  I get it we did so that we can reuse the same component in select as well right? We should have just created the List and add our styles and use it in Select and Comobox. 
So instead of FieldListPopver, the the combobox, we just do:

```jsx
```

app/client/src/widgets/WidgetUtils.ts
```diff
@@ -989,3 +989,45 @@ export const checkForOnClick = (e: React.MouseEvent<HTMLElement>) => {
...
+ * ```
+ *
+ * It will return
+ * ```js
+ * {
+ *  isValidDate: "{{ this.value === 1 }}"
+ * }
+ * ```
+ *
+ * Main rule to remember is don't use deconstruct the props like `const { value } = props;` in the derived property function. Directly access props like `props.value`
```
riodeuno, line 1012: Can we write a test for this? Or at a lint rule?

app/client/src/widgets/WidgetUtils.ts
riodeuno: @jsartisan I really appreciate this change. 
Let's make sure that this is used across all widgets in a separate PR and cleanup our code. 

This function may find itself in another file in the future, but for now this looks fine to me.

app/client/src/widgets/WidgetUtils.ts
jsartisan: I don't think we can add a guard with a test. The lint rule would be better. I'll pick that up as a separate task.

jsartisan:
@ichik Can you check if everything is ok visually here?

ichik:
@jsartisan 
Suggestion (design-wise): 
1. Switch the inner button to `ghost` perhaps even with `neutral` color instead of `accent`
2. Date selection should show up on focus by default (optionally turning that off with widget settings)
3. Dates in previous/next months should have `-subtle` color
4. The widget should be put into `Time` category in the widget insert panel

Bugs and other observations:
1. There's no way to select the time in the dropdown, only keyboard entry
2. Date format seems to have no effect on the widget until you start interacting with precision prop (precision prop settings take priority over format and may even contradict it)
3. Date range selection: will we do a separate widget or add it to this one as a prop?
4. Related to the previous functionality: would be good to have access to a prop that controls how many months are shown in the dropdown

PR #36704, opened 2024-10-04
Task: added heartbeat for monitoring

ApekshaBhosale:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/client/src/Heartbeat.tsx
+++ b/app/client/src/Heartbeat.tsx
@@ -1,0 +1,34 @@
+import { useEffect } from "react";
+import log from "loglevel";
+const Heartbeat = () => {
+  const heartbeatUrl =
+    "https://uptime.betterstack.com/api/v1/heartbeat/71SZmsvAVW73QvEPQzLvxwEL";
+  const interval = 30 * 1000; // 30 secs in milliseconds
+
+  const sendHeartbeat = async () => {
+    try {
+      const response = await fetch(heartbeatUrl);
+
+      if (response.ok) {
+        log.info("Heartbeat sent successfully");
+      } else {
+        log.error("Failed to send heartbeat", response.status);
+      }
+    } catch (error) {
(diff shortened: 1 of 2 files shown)


app/client/src/Heartbeat.tsx
```diff
@@ -0,0 +1,34 @@
+import { useEffect } from "react";
+import log from "loglevel";
+const Heartbeat = () => {
+  const heartbeatUrl =
```
ashit-rath, line 4: Do you think we should keep this as env variable? It would be easier to swap out or even have a condition that if this url is missing then don't ping which may help use to pause this service instantly if something goes wrong?

Same for the interval

ApekshaBhosale: why would it be missing? we don't want to monitor any other instance. can you give some example where adding this on env variable would be useful? @ashit-rath

ashit-rath: > why would it be missing?

What I meant was if we do not set the env then it's absence can be treated as not to trigger this service. 

> can you give some example where adding this on env variable would be useful?

Let's say we want to swap out the service provider or the url of the service changes or we want to reduce/increase the timeout or we want to check the heartbeat for a different instance using a different account/workspace/however then choose to differentiate instance like sentry does. Usually external services are passed down as an env like sentry / algolia / smartlook etc.

ApekshaBhosale: this is part of status.appsmith.com. i don't think we will change the provider and other instances we won't be monitoring this way. we have something called monitoring agent coming up and we will be using that

ashit-rath (approved):

PR #35124, opened 2024-07-23
feat: display label in table select cell

jacquesikot:
## Description
**Issue**
In the table widget, the select column currently displays the value from the dropdown when a user makes a selection, rather than the label. This behaviour is inconsistent with the standalone select widget, which correctly renders the label upon selection.

**Fix**
This PR addresses the inconsistency by modifying the table widget's select column to display the label of the selected item, while maintaining the table cell value and ensuring alignment with the behaviour of the standalone select widget.

**Tested Cases**
Manual Test Cases for Table Select Widget Improvement

1. Default Value Display
**Objective:** Ensure that a new table displays the correct default label key from the selectOptions in the cell.
**Steps:**
- Drop a new table widget.
- Add sample data to the table.
- Set a column (e.g., gender) to select type.
- Verify that the table cell displays the value from the label key of the `selectOptions` by default.

2. Binding Check for Selected Row
**Objective:** Ensure that the table binding for the selected row reflects the correct value key from the selectOptions.
**Steps:**
- Drop a new table widget.
- Add sample data to the table.
- Set a column (e.g., gender) to select type.
- Select a row and verify that `Table1.selectedRow.gender` matches the value key from the `selectOptions`.

3. Updating Table Cell Content
**Objective:** Verify that updating the table cell content via the dropdown updates the cell with the correct label key property.
**Steps:**
- Drop a new table widget.
- Add sample data to the table.
- Set a column (e.g., gender) to select type.
- Change the content of the table cell using the dropdown.
- Confirm that the cell content updates accordingly with the value key property.

4. Add New Row Functionality
**Objective:** Ensure that adding a new row works as expected with the enhanced functionality.
**Steps:**
- Drop a new table widget.
- Add sample data to the table.
- Set a column (e.g., gender) to select type.
- Add a new row to the table.
- Verify that the new row uses the label key property from the `selectOptions` in the table display.

5. Deployed Mode Verification
**Objective:** Verify that the display functionality works correctly in deployed mode.
**Steps:**
- Drop a new table widget.
- Add sample data to the table.
- Set a column (e.g., gender) to select type.
- Change the displayAs property to both value and label.
- Deploy the table.
- Verify that the table displays correctly in both value and label modes in the deployed environment.

Fixes #26188

## Automation

/ok-to-test tags="@tag.Table, @tag.Binding, @tag.Sanity"

### :mag: Cypress test results

> [!TIP]
> 🟢 🟢 🟢 All cypress tests have passed! 🎉 🎉 🎉
[...]

PR #36755, opened 2024-10-08
feat: Allow filtering of table select column label

### LINKED ISSUE #36635: [Bug]: Table with select column search and filter over the value, not over the label

### Description

When searching or filtering over a table with a select column type, those filters are applied using the `value` value of the select, but they should use the `label` value, i.e., what the user sees in the table.
imagen

### Steps To Reproduce

1. Create a table widget.
2. Assign data, e.g., `{{ [{ "column": "foo", }] }}`.
3. Set column data type to `select`.
4. Set options for this table, e.g., `{{ { "label":"bar", "value":"foo" } }}`.
5. When searching or filtering `bar`, nothing appears. When searching or filtering `foo`, the row with the value `bar` appears.

### Public Sample App

_No response_

### Environment

Production

### Severity

Medium (Frustrating UX)

### Issue video log

_No response_

### Version

Appsmith v1.38

jacquesikot:
## Description

**Problem**
When filtering a table with a select column type, users expect to filter by the visible label values shown in each cell. Currently, however, filtering is applied to the underlying option values rather than the displayed labels, leading to unexpected filter results for end-users.

**Root Cause**
In a previous update ([PR #35124](https://github.com/appsmithorg/appsmith/pull/35124)), the table cell display for select columns was changed to show labels instead of values. However, the filtering logic was not updated accordingly, so the table still filtered on the original option values, creating a mismatch between displayed and filtered content.

**Solution**
This PR modifies the displayedRow property within the table widget to use the label property instead of the value key when filtering or searching select column data. This ensures that table filtering and searching now align with the visible label values in the select columns, providing a more intuitive user experience.

Fixes #36635 

## Automation

/ok-to-test tags="@tag.Sanity, @tag.Table, @tag.Select, @tag.Binding"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/11234735437>
> Commit: fd6c179ffb2c61d23cb98fd749c8df49cebcfcdd
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=11234735437&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Sanity, @tag.Table, @tag.Select, @tag.Binding`
> Spec:
> <hr>Tue, 08 Oct 2024 12:48:01 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
	- Introduced a new test case to verify filtering functionality for the "role" column in the Table Widget.
	- Enhanced filtering mechanism to support multiple label values for select columns.

- **Bug Fixes**
	- Removed outdated search functionality to streamline user experience.

- **Refactor**
	- Restructured existing test cases for improved clarity and flow.

### DIFF
--- a/app/client/src/widgets/TableWidgetV2/widget/derived.js
+++ b/app/client/src/widgets/TableWidgetV2/widget/derived.js
@@ -684,7 +684,7 @@ export default {
        * For select columns with label and values, we need to include the label value
        * in the search
        */
-      let labelValueForSelectCell = "";
+      let labelValuesForSelectCell = {};
       /*
        * Initialize an array to store keys for columns that have the 'select' column type
        * and contain selectOptions.
@@ -716,7 +716,7 @@ export default {
             primaryColumns[key].selectOptions,
           );
 
-          let selectOptions;
+          let selectOptions = {};
 
           /*
            * If selectOptions is an array, check if it contains nested arrays.
@@ -732,7 +732,8 @@ export default {
               });
 
               if (option) {
-                labelValueForSelectCell = option.label;
+                labelValuesForSelectCell[primaryColumns[key].alias] =
+                  option.label;
               }
             } else {
               /* Handle the case where selectOptions is a flat array - selectOptions is plain JSON */
@@ -742,7 +743,8 @@ export default {
               );
 
               if (option) {
-                labelValueForSelectCell = option.label;
+                labelValuesForSelectCell[primaryColumns[key].alias] =
+                  option.label;
               }
             }
           } else {
@@ -753,7 +755,8 @@ export default {
             );
 
             if (option) {
-              labelValueForSelectCell = option.label;
+              labelValuesForSelectCell[primaryColumns[key].alias] =
+                option.label;
             }
           }
         });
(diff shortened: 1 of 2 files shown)


app/client/src/widgets/TableWidgetV2/widget/derived.js
```diff
@@ -684,7 +684,7 @@ export default {
        * For select columns with label and values, we need to include the label value
        * in the search
        */
-      let labelValueForSelectCell = "";
+      let labelValuesForSelectCell = {};
```
rahulbarwal, line 687: Please update the comment on top of this line. And check if there are any other comments that can be updated.

rahulbarwal (dismissed):
Lets update all the comments. Otherwise good to go.

app/client/src/widgets/TableWidgetV2/widget/derived.js
jacquesikot: Thanks, its done

rahulbarwal (approved):

sagar-qa007 (approved):

PR #36801, opened 2024-10-10
chore: Updating some states related to action execution

ankitakinger:
## Description

Updating some states related to action execution under modularised flow: isRunning, isRunDisabled, onDebugClick, actionResponse.

Fixes [#36802](https://github.com/appsmithorg/appsmith/issues/36802)

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/11275569168>
> Commit: 2b23612de534793e5852fbafd8e54821e09a290f
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=11275569168&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Sanity`
> Spec:
> <hr>Thu, 10 Oct 2024 14:38:26 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
	- Enhanced context management in the Plugin Action Editor to include action response data.
	- Introduced a new click handler for the "Run" button in the Plugin Action Toolbar with improved state management.
	- Improved logic for action execution blocking based on running state and user permissions.
	- Added a custom hook to determine whether plugin action execution should be blocked.

- **Bug Fixes**
	- Ensured the context value updates correctly when action response changes.

- **Documentation**
	- Updated context type declarations to reflect new properties.


app/client/src/ce/PluginActionEditor/components/PluginActionResponse/hooks/usePluginActionResponseTabs.tsx
```diff
@@ -42,8 +49,36 @@ function usePluginActionResponseTabs() {
...
+  const actionDatasourceName = action?.datasource.name || "";
+
+  // if the url is empty and the action's datasource name is the default datasource name (this means the api does not have a datasource attached)
+  // or the user does not have permission,
+  // we block action execution.
+  const blockExecution =
+    (!actionDatasourceUrl &&
+      !actionDatasourceUrlPath &&
+      actionDatasourceName === DEFAULT_DATASOURCE_NAME) ||
+    !isExecutePermitted;
```
hetunandu, line 74: This is only related to API actions and should be extracted out into its own hook. Also should this condition not be applied at the toolbar run level as well?

hetunandu (changes requested):

app/client/src/ce/PluginActionEditor/components/PluginActionResponse/hooks/usePluginActionResponseTabs.tsx
ankitakinger: Yes you're correct. It will be needed there too. Although, I just checked, both Query and Api uses different conditions for blocking execution. Should we return the value from the hook based on action type?

hetunandu: If they rely on completely different logic to decide this, it should be extracted out into separate utils.

ankitakinger: I've kept it in the same hook because `PluginActionToolbar` is common between Api and Query and so we will end up having Action type checked there if we have 2 separate hooks.

hetunandu (approved):

PR #36899, opened 2024-10-16
chore: add soft token + add subtle button variant

jsartisan:
/ok-to-test tags="@tag.Anvil"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/11362114942>
> Commit: e1fc66eb329aa58c08ae6147cc6a5bb1d8107a17
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=11362114942&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Anvil`
> Spec:
> <hr>Wed, 16 Oct 2024 09:07:10 UTC

### DIFF
--- a/app/client/packages/design-system/widgets/src/components/Text/src/types.ts
+++ b/app/client/packages/design-system/widgets/src/components/Text/src/types.ts
@@ -5,7 +5,7 @@ import type {
 import type { ReactNode } from "react";
 import type { COLORS } from "../../../shared";
 
-export interface TextProps extends React.HTMLAttributes<HTMLElement> {
+export interface TextProps {
   /** size variant of the text
    * @default body
    */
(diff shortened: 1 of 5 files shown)


jsartisan:
Note: This PR adds a new subtle variant that we will use for suggested prompt buttons in AIChat component.
In the next PR in ee, I'll replace the `AssistantSuggestionButton` with our button with subtle variant.

app/client/packages/design-system/widgets/src/components/Text/src/types.ts
```diff
@@ -5,7 +5,7 @@ import type {
 import type { ReactNode } from "react";
 import type { COLORS } from "../../../shared";
 
-export interface TextProps extends React.HTMLAttributes<HTMLElement> {
```
znamenskii-ilia, line 8: I wonder why this was removed? This change will not allow to pass standard html attributes to Text

jsartisan: So what was happening was initially when creating markdown components for paragraphs, I was using `...props` and passing it to our `Text` which resulted in type conflicts as props of paragraph were `HTMLParagraphElement`  and our `Text`  couldn't accept those props. So I added the above code to fix it. But later I removed `...props` part. So not required anymore.

Having the above causes issues in storybook. It shows all props of HTMLelement in the docs. 🫠

CleanShot 2024-10-16 at 14 11 00

KelvinOm (approved):

PR #36915, opened 2024-10-16
chore: refactor query duplication flow

ayushpahwa:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/client/src/PluginActionEditor/hooks/index.ts
+++ b/app/client/src/PluginActionEditor/hooks/index.ts
@@ -1,5 +1,6 @@
 export { useActionSettingsConfig } from "ee/PluginActionEditor/hooks/useActionSettingsConfig";
 export { useHandleDeleteClick } from "ee/PluginActionEditor/hooks/useHandleDeleteClick";
+export { useHandleDuplicateClick } from "ee/PluginActionEditor/hooks/useHandleDuplicateClick";
 export { useHandleRunClick } from "ee/PluginActionEditor/hooks/useHandleRunClick";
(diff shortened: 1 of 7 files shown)


app/client/src/PluginActionEditor/hooks/useHandleDuplicateClick.ts
```diff
@@ -0,0 +1,26 @@
+import { copyActionRequest } from "actions/pluginActionActions";
+import { usePluginActionContext } from "PluginActionEditor/PluginActionContext";
+import { useCallback } from "react";
+import { useDispatch } from "react-redux";
+
+function useHandleDuplicateClick() {
```
sagar-qa007, line 6: @ayushpahwa any unit test case coverage needed here?

sagar-qa007 (changes requested):

hetunandu (approved):

app/client/src/PluginActionEditor/hooks/useHandleDuplicateClick.ts
ayushpahwa: This is just a hook which is passing on the action as per the input. I have added test for the rendering of the action menu which was the actual change and also covers the functionality of this hook. This test can be enhanced to introduce coverage of other options also but since they are already covered in cypress, I have just left a TODO comment for now for the same.

hetunandu (approved):

sagar-qa007 (approved):

PR #36937, opened 2024-10-17
chore: extract useCurrentAppState hook to reduce cyclic dependency

hetunandu:
## Description

Extracts an IDE hook `useCurrentAppState` as it was used in multiple places. This usage was causing cyclic dependencies in various places as it was included in a file which was imported as well multiple times. This one extraction removes more than 100 cycle dependencies.

## Automation

/ok-to-test tags="@tag.All"

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No


ankitakinger (approved):
LGTM

Issue #36879, opened 2024-10-15
[Bug]: Duplicate JS Object action creation and unable to delete the duplicate action


rishabhrathod01:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

When a user creates a duplicate action in JSObject, the duplicate action is retained in the DB and cannot be deleted. After this, the updates are not sent to the server for body changes of JSObject action. 

### Steps To Reproduce

https://github.com/user-attachments/assets/33e3633d-522c-4ebb-a690-e34edde3af4e

### Environment

Production

### Severity

High (Blocker to building or releasing)

### Version

1.42

rishabhrathod01:
Adding `.distinct(actionDTO -> actionCollectionDTO.getName() + "." + actionDTO.getName())` after `Flux.fromIterable(actionCollectionDTO.getActions())` in `updateUnpublishedActionCollection` method inside `LayoutCollectionServiceCEImpl.java` resolves this issue. We could think about the response code and any warning error we would like to send back.

PR #36958, opened 2024-10-17
fix: resolve redundant JSObject action saving

rishabhrathod01:
## Description

In this change, we avoid storing duplicate actions or variables in the `parsedBody` after AST parsing. This prevents duplicate actions from being sent to the server on save. 

This change also removes `parsedFunction` from the action object of parsedBody as it is not being used.

Fixes https://github.com/appsmithorg/appsmith/issues/36879

## Test Cases

- [x] After parsing JS Object make sure parsedBody doesn't have duplicate actions or variables

## Automation

/ok-to-test tags="@tag.JS"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/11485245933>
> Commit: c12ee7007d8f860aa3bbf2cc1889bd4b762785a6
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=11485245933&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.JS`
> Spec:
> <hr>Wed, 23 Oct 2024 18:20:11 UTC

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **New Features**
	- Introduced enhanced methods for creating, updating, and moving action collections, improving validation and consistency.
	- Added a new interface for better handling of JavaScript actions and variables, enhancing data processing efficiency.

- **Bug Fixes**
	- Improved error reporting from pages to collections, ensuring accurate tracking of action errors.
	- Added tests to ensure duplicate actions are correctly handled during updates.

- **Documentation**
	- Updated comments and documentation for clarity on new functionalities and changes.

### DIFF
--- a/app/client/src/workers/Evaluation/JSObject/index.ts
+++ b/app/client/src/workers/Evaluation/JSObject/index.ts
@@ -21,6 +21,9 @@ import JSObjectCollection from "./Collection";
 import ExecutionMetaData from "../fns/utils/ExecutionMetaData";
 import { jsPropertiesState } from "./jsPropertiesState";
 import { getFixedTimeDifference } from "workers/common/DataTreeEvaluator/utils";
+interface ParseJSAction extends ParsedJSSubAction {
+  parsedFunction: unknown;
+}
 
 /**
  * Here we update our unEvalTree according to the change in JSObject's body
@@ -114,18 +117,17 @@ export function saveResolvedFunctionsAndJSUpdates(
         JSObjectName: entityName,
         JSObjectASTParseTime,
       });
-      // TODO: Fix this the next time the file is edited
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      const actions: any = [];
-      // TODO: Fix this the next time the file is edited
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      const variables: any = [];
+
+      const actionsMap: Record<string, ParseJSAction> = {};
+      const variablesMap: Record<string, { name: string; value: unknown }> = {};
 
       if (success) {
         if (!!parsedObject) {
           jsPropertiesState.update(entityName, parsedObject);
           parsedObject.forEach((parsedElement) => {
             if (isJSFunctionProperty(parsedElement)) {
+              if (actionsMap[parsedElement.key]) return;
+
               try {
                 ExecutionMetaData.setExecutionMetaData({
                   enableJSVarUpdateTracking: false,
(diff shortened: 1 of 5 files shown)


rishabhrathod01:
image

Executing both functions always executes the later part, according to this we show an error for the 2nd function. We should avoid overriding the first function definition.

rishabhrathod01 (commented):
> image
> 
> Executing both functions always executes the 2nd function body part, according to this we show an error for the 2nd function. We should avoid overriding the first function definition.

Issues still retains fix this before merging.

app/client/src/workers/Evaluation/JSObject/index.ts
```diff
@@ -21,6 +21,9 @@ import JSObjectCollection from "./Collection";
 import ExecutionMetaData from "../fns/utils/ExecutionMetaData";
 import { jsPropertiesState } from "./jsPropertiesState";
 import { getFixedTimeDifference } from "workers/common/DataTreeEvaluator/utils";
+interface ParseJSAction extends ParsedJSSubAction {
+  parsedFunction: unknown;
```
sneha122, line 25: @rishabhrathod01 Setting this unknown here, will it cause any compile time or runtime issues when the parsedBody is used in subsequent computations? Will I need to check any test cases here specifically?

sneha122:
@rishabhrathod01 Should we add unit test cases for the changes?

app/client/src/workers/Evaluation/JSObject/index.ts
rishabhrathod01: > Will it cause any compile time or runtime issues when the parsedBody is used in subsequent computations? 

No, the `parsedFunction` is not consumed in subsequent computation.

> Will I need to check any test cases here specifically?

No

app/client/src/workers/Evaluation/JSObject/index.ts
```diff
@@ -312,8 +310,6 @@ export function parseJSActions(
     parsedBody.actions = parsedBody.actions.map((action) => {
       return {
         ...action,
-        // parsedFunction - used only to determine if function is async
-        parsedFunction: undefined,
```
sneha122, line 316: @rishabhrathod01 Since we have removed this, do you think there will be any issues in determining if function is async or not as per the comment above this code?

sneha122: Read the PR description, since it's not being used, resolving this comment.

sneha122 (approved):

app/client/src/workers/Evaluation/JSObject/index.ts
rishabhrathod01: @sneha122 we have also removed the sync-async determination from our codebase. So this won't be required anymore.

PR #36980, opened 2024-10-19
chore: fix blueprint operation inconsistency for zone

jsartisan:
In anvil, when widgets are dropped on canvas, we create a zone which has `visual separation` marked as true by default. 
Now, There is a usecase for the chat widget in which we want to modify the zone`s `Visual separation` to false on creation of zone. 

The code for bleuprint operation for chat widget would like this:

```js
blueprint: {
    operations: [
      {
        type: BlueprintOperationTypes.MODIFY_PROPS,
        fn: (
          widget: FlattenedWidgetProps,
          widgets: CanvasWidgetsReduxState,
          parent: FlattenedWidgetProps,
          layoutSystemType: LayoutSystemTypes,
        ) => {
          if (layoutSystemType !== LayoutSystemTypes.ANVIL) return [];

          const updates: UpdatePropertyArgs[] = [];
          const parentId = widget.parentId;

          if (!parentId) return updates;

          const parentWidget = widgets[parentId];

          // we want to proceed only if the parent is a zone widget and has no children
          if (
            parentWidget.children?.length === 0 &&
            parentWidget.type === "ZONE_WIDGET"
          ) {
            updates.push({
              widgetId: parentId,
              propertyName: "elevatedBackground",
              propertyValue: false,
            });
          }

          return updates;
        },
      },
    ],
  },
  ```

This should work fine, but in the code where we create zone and attaching widgets to it, after running the blueprint operations, we were not using the correct updated zone that returned from blueprint operations of the child widgets. This PR uses the correct zone.

Also there is a case where blueprint operation is not able to update the existing widgets because all properties of existing widgets were readonly. So cloned the widgets from redux before passing to widget addition saga functions.

/ok-to-test tags="@tag.All"

## Summary by CodeRabbit

- **New Features**
	- Improved clarity in widget handling by renaming parameters related to dragged widgets.
	- Streamlined the process of adding widgets to zones by simplifying parameter structures.

- **Bug Fixes**
	- Enhanced immutability in widget property updates within the state management process.

- **Style**
	- Updated CSS styles for the `.markdown` class, removing unnecessary pseudo-elements for improved formatting.

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/11474751370>
> Commit: 8a385eec313142b40f848eed20310d3774c0705c
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=11474751370&attempt=3" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Wed, 23 Oct 2024 09:47:24 UTC

## Summary by CodeRabbit

- **New Features**
	- Improved clarity in widget handling by renaming parameters related to dragged widgets.
	- Enhanced functionality for adding widgets to zones by simplifying the data structure used.
	- Implemented safer state manipulation for widget addition using a deep copy approach.

- **Bug Fixes**
	- Addressed potential issues with direct state mutation during widget addition.

- **Style**
	- Updated CSS styles for the Markdown component by removing unnecessary pseudo-elements.

### DIFF
--- a/app/client/packages/design-system/widgets/src/components/Markdown/src/styles.module.css
+++ b/app/client/packages/design-system/widgets/src/components/Markdown/src/styles.module.css
@@ -1,12 +1,6 @@
 .markdown {
   color: var(--color-fg);
 
-  &::after,
-  &::before {
-    /* This is required to remove the compensators of capsizing that comes up due to use of `wds-body-text` class */
-    content: none !important;
-  }
-
   table {
     border: var(--border-width-1) solid var(--color-bd);
     border-collapse: separate;
--- a/app/client/src/layoutSystems/anvil/integrations/sagas/anvilWidgetAdditionSagas/index.ts
+++ b/app/client/src/layoutSystems/anvil/integrations/sagas/anvilWidgetAdditionSagas/index.ts
@@ -27,6 +27,7 @@ import log from "loglevel";
 import { generateDefaultLayoutPreset } from "layoutSystems/anvil/layoutComponents/presets/DefaultLayoutPreset";
 import { addWidgetsToPreset } from "layoutSystems/anvil/utils/layouts/update/additionUtils";
 import { addNewAnvilWidgetToDSL } from "./helpers";
+import { klona } from "klona";
 
 // The suggested widget functionality allows users to bind data from the Query pane
 // to a new or existing widget on the Canvas.
@@ -109,7 +110,8 @@ export function* getUpdatedListOfWidgetsAfterAddingNewWidget(
   isSection: boolean, // Indicates if the drop zone is a section
 ) {
   const { alignment, canvasId } = highlight;
-  const allWidgets: CanvasWidgetsReduxState = yield select(getWidgets);
+  const allWidgetsFromRedux: CanvasWidgetsReduxState = yield select(getWidgets);
+  const allWidgets = klona(allWidgetsFromRedux) as CanvasWidgetsReduxState;
 
   const parentWidgetWithLayout = allWidgets[canvasId];
 
--- a/app/client/src/layoutSystems/anvil/utils/layouts/update/sectionUtils.ts
+++ b/app/client/src/layoutSystems/anvil/utils/layouts/update/sectionUtils.ts
@@ -19,7 +19,7 @@ import { addNewAnvilWidgetToDSL } from "layoutSystems/anvil/integrations/sagas/a
 export function* createSectionAndAddWidget(
   allWidgets: CanvasWidgetsReduxState,
   highlight: AnvilHighlightInfo,
-  widgets: WidgetLayoutProps[],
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

File: app/client/packages/design-system/widgets/src/components/Markdown/src/styles.module.css [hunk 1]
```diff
@@ -1,12 +1,6 @@
 .markdown {
   color: var(--color-fg);
 
-  &::after,
```
jsartisan, line 4: This is not required. Not related to the current PR but added  in this PR just to avoid another PR.

app/client/src/layoutSystems/anvil/integrations/sagas/anvilWidgetAdditionSagas/index.ts
```diff
@@ -109,7 +110,8 @@ export function* getUpdatedListOfWidgetsAfterAddingNewWidget(
   isSection: boolean, // Indicates if the drop zone is a section
 ) {
   const { alignment, canvasId } = highlight;
-  const allWidgets: CanvasWidgetsReduxState = yield select(getWidgets);
+  const allWidgetsFromRedux: CanvasWidgetsReduxState = yield select(getWidgets);
+  const allWidgets = klona(allWidgetsFromRedux) as CanvasWidgetsReduxState;
```
jsartisan, line 114: @riodeuno Is this safe? The use case is aichatwidget wants to modify the parent zone's elevatedBackground to false. ( that code is written in aichatbot blueprint operation ).
So the problem I was facing is if there is an existing zone that is empty and I drop aichatwidget into it, I was not able to modify the zone elevated background, the redux state was coming from redux and it was using immer, due to which the property was readonly.

riodeuno: @jsartisan This is a problem because we will quickly lose track of where the original values were updated. Ideally, we should let the functions return what needs to be updated, then in this function we should use something like `produce` from `immer` to update the appropriate paths.

I'm not a 100% sure of what is happening in this PR, so I can't provide an alternative. Could you elaborate in the description on "how" we're going about changing the zone's `elevatedBackground` property?

riodeuno:
@jsartisan @KelvinOm 

Note:
The approach for `BlueprintOperations` needs a review. The way it integrates with the system is not clean and has resulted in some hacks like the following:
- Sagas are directly called from Blueprint, causing it to be coupled with Datasource system.
- Immutable data from widgets is cloned to be updated by Blueprint, resulting in mutation functions. (Not very functional and testable)

I'd suggest we unblock AI Chat with the changes in this PR, and revisit the system to understand how to modularise this particular service.

KelvinOm (approved):

PR #37018, opened 2024-10-23
chore: Create query selector

jsartisan:
/ok-to-test tags="@tag.All"

This PR updates the WidgetQueryGeneratorForm to only allow queries so that it looks like a query selector. This is needed for chat widget where the widget can only bind Appsmith AI queries.

CleanShot 2024-10-23 at 12 30 32

PS: The above image is an example where TableWidget had allowedDatasourcestypes as "Mysql" ( just for testing )

## Summary by CodeRabbit

- **New Features**
	- Introduced a new dropdown variant "Connect to query" for improved datasource selection.
	- Enhanced handling of dynamic query options and datasource filtering based on context.

- **Improvements**
	- Updated the Widget Query Generator Form to support new properties for better data management.
	- Improved control flow in dropdown options to ensure appropriate options are displayed based on the current context.
	- Enhanced the One Click Binding Control to utilize new properties for better interaction with the Widget Query Generator Form.

- **Bug Fixes**
	- Refined logic for displaying datasource options, preventing unnecessary options from appearing.

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/11493055126>
> Commit: d6e322c649ef07c5fa58cd16fa0a5d97e55f7129
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=11493055126&attempt=2" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Thu, 24 Oct 2024 06:53:37 UTC


app/client/src/actions/datasourceActions.ts
```diff
@@ -32,9 +32,10 @@ export const createDatasourceFromForm = (
   };
 };
 
-export const createTempDatasourceFromForm = (
```
jsartisan, line 35: Types are a little weird in data source code. The code uses just plugin and type in the function, yet we pass` Datasource | CreateDatasourceConfig` here. I think this is done just to satisfy the typescript.

jsartisan:
> What happens if the Appsmith AI Datasource doesn't exist?
You will see an empty dropdown in that case. That state is properly handled in UI as of now. Btw, as soon as you drop the chat widget, we create a new Appsmith AI query ( @KelvinOm do we also create datasource if it does not exists? )

> if the user wants to use their own datasource?
This case won't happen. the AIChat widget explicitly require AppsmithAI datasoruce cause that datasources is handling lot of stuff under the hood. So, it can't be binded with other datasources like MySQL, postgres.

The way the widget tells that it only needs AppsmithAI is by specifying in the propertyPaneConfig like this:

```js
{
	propertyName: "chatQuery",
	helpText: "The query to be used for the chat",
	label: "Chat Query",
	controlType: "ONE_CLICK_BINDING_CONTROL",
	controlConfig: {
	  showEditFieldsModal: false, 
	  datasourceDropdownVariant: DROPDOWN_VARIANT.CONNECT_TO_QUERY
	  excludePrimaryColumnFromQueryGeneration: true,
	  isConnectableToWidget: false,
	  allowedDatasourceTypes: [PluginName.APPSMITH_AI], // <-------------------
	  getQueryBindingValue: (query: ActionData | ModuleInstanceData) => {
	    return query.config.name;
	  },
	},
}
 ```

KelvinOm:
> do we also create datasource if it does not exists?

@jsartisan  Yes, if datasource is not exists, then we create it, if there is one, we add query to it.

KelvinOm (approved):

PR #37062, opened 2024-10-24
fix: avoiding unnecessary update page layouts in js action update

### LINKED ISSUE #37046: [Task]: Avoid update page layout from getting called multiple times
### SubTasks

While evaluating the metrics for JS object update call, we came across a strange observation, when updating JsObject which has 10 existing actions, update page layout was happening with each action update and it was also happening at the end when all actions and action collection is updated.

This task is to remove those unnecessary page layout updates with action update as final page layout update will happen after all updates are done.

Screenshot 2024-10-22 at 6 47 46 PM

With this we hope to improve the JSobject update API performance by upto 50%.

sneha122:
## Description

When JS object is updated we have a JS object update API call going in (api/v1/collections/actions) This API does mainly does following things:
1. Updates all actions (Js functions) of the JS object
2. Updates action collection (Js Object)
3. Updates the page layout and on load actions

The issue was with 1st part where, every time we would update the action (JS function), we were also calling update page layout (3rd part), which means if we are updating 20 actions, update page layout was getting called 20 times for each action update. This is in addition to page layout update that happens after action collection is updated. 

This PR fixes that issue and removes those redundant calls for updating page layout with each action update.

Fixes #37046 
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags="@tag.JS, @tag.Sanity"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/11513970737>
> Commit: 95cc7e7e2f4fecc399437d0b8ce64f3fa880e3d9
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=11513970737&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.JS, @tag.Sanity`
> Spec:
> <hr>Fri, 25 Oct 2024 08:27:48 UTC

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **New Features**
	- Enhanced action processing by introducing conditional layout updates based on action type, specifically for JavaScript actions.

- **Bug Fixes**
	- Improved efficiency by preventing unnecessary layout updates when actions are of type JavaScript.

- **Tests**
	- Added new test coverage for layout update functionality, ensuring correct invocation of layout services during action collection updates.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/LayoutActionServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/LayoutActionServiceCEImpl.java
@@ -248,9 +248,10 @@ public class LayoutActionServiceCEImpl implements LayoutActionServiceCE {
                 .name(UPDATE_SINGLE_ACTION)
                 .tap(Micrometer.observation(observationRegistry))
                 .flatMap(updatedAction -> {
-                    // JS actions are skipped because when JSobject is updated, we first update all actions and action
-                    // collection
-                    // and then
+                    // Update page layout is skipped for JS actions here because when JSobject is updated, we first
+                    // update all actions, action
+                    // collection and then we update the page layout, hence updating page layout with each action update
+                    // is not required here
                     if (action.getPluginType() != PluginType.JS) {
                         updateLayoutService
                                 .updatePageLayoutsByPageId(pageId)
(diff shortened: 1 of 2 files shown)


nidhi-nair (changes requested):
@sneha122 this change needs to be accompanied by test cases.

sneha122:
> @sneha122 this change needs to be accompanied by test cases.

@nidhi-nair Do you have any suggestions for the test cases? What assertions can be made? I was also thinking of test cases but thought this is simply removing the redundant calls

NilanshBansal (dismissed):
LGTM under the assumption that existing tests does cover this change. 

@sneha122 please check above comment from Nidhi and move forward only post resolution of that.

Issue #37043, opened 2024-10-23
[Bug]: Migration of the workflows to postgres does not migrate schedules


abhvsn:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

During the release-ee migration we observed the fields which are created after the loadMongoData migration is executed e.g. `workflow.schedules` which was added as part of `V006_02__addSchedulesColumnInWorkflow.sql`, does not get carried forward to destination instance. This is because at the time of populating the data from jsonl files to postgres table, it does not contain the required field. 

So going forward till the `pg` branch is not merged to `release` we will have to update the `V001__createTables.sql`, `V001_01__adaptCETables.sql` and `V001_02__createTables.sql` migrations as necessary and also add alter table statements to import the data correctly in the destination instance. 

### Steps To Reproduce

1. Create a workflow schedule on MongoDB
2. Run the migration script
3. Check the migrated data on postgres, for which schedules field will be missing

### Environment

Production

### Severity

Medium (Frustrating UX)

### Version

v1.47

PR #37103, opened 2024-10-25
chore: Add flyway repair for the migration

AnaghHegde:
## Description
The reason for doing this is that we want to introduce the spotless on the SQL files. Post this, the existing migrations checksum will change and the validation fails because of this. Hence introducing the `flyway.repair()`
We have considered the other alternatives like `ignoreMigration` (not available in OSS), change baseline version(not working out ex ecpected), make flyway to ignore the validation on few files by setting checksum as `NULL`

In this setup, calling migrate() explicitly after repair() ensures that all pending migrations are applied immediately after any necessary repairs to the schema history. This approach is beneficial because:

Order of Execution: Spring Boot’s default Flyway configuration may automatically call migrate() at application startup if spring.flyway.enabled=true, but this will not account for any issues that repair() might have resolved.
Ensures Continuity: Calling repair() fixes schema history inconsistencies like checksum mismatches, and calling migrate() right afterward ensures that migrations proceed smoothly with any errors addressed first. This can prevent unexpected issues in production environments.
Creating the Flyway Instance: In the configuration, creating the Flyway instance allows you to customize the setup before Spring Boot runs any migrations. Specifically:

Custom Configuration: By creating the Flyway instance yourself, you can set properties directly on it (like baselineVersion, locations, dataSource, etc.) based on what’s defined in application.properties. This setup also gives flexibility to add conditional logic, if needed, for different environments.
Control over Repair and Migration: By defining your own Flyway bean, you gain explicit control over when repair() and migrate() are called and in what order, which can be important if you need to fix schema inconsistencies before applying any migrations.

## Automation

/ok-to-test tags=""

> [!WARNING]
> Tests have not run on the HEAD c23d8ef59fcf832aa37e5b9d12b76ba39281c5c4 yet
> <hr>Mon, 04 Nov 2024 09:41:52 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
	- Introduced a new configuration for database migration management with `FlywayRepairConfig`.
	- Enhanced application readiness handling with new event-driven approaches in `UserConfig`, `ThemesConfig`, and `SeedMongoData`.
	- Added a test configuration for Flyway migrations to ensure a clean database state during tests.

- **Bug Fixes**
	- Improved clarity in configuration properties comments.

- **Chores**
	- Removed outdated Flyway configuration properties to streamline database migration processes.
	- Updated SQL debugging configuration to utilize environment variables.
	- Added a new property to enable Flyway migrations during application tests.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/FlywayRepairConfig.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/FlywayRepairConfig.java
@@ -2,6 +2,7 @@ package com.appsmith.server.configurations;
 
 import org.flywaydb.core.Flyway;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.ApplicationRunner;
 import org.springframework.boot.autoconfigure.jdbc.DataSourceProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
@@ -18,17 +19,24 @@ public class FlywayRepairConfig {
 
     @Bean
     public Flyway flyway() {
-        Flyway flyway = Flyway.configure()
+        return Flyway.configure()
                 .dataSource(
                         dataSourceProperties.getUrl(),
                         dataSourceProperties.getUsername(),
                         dataSourceProperties.getPassword())
-                .locations("classpath:com/appsmith/server/migrations")
+                .locations("classpath:db/migration") // Adjust path if needed
                 .baselineOnMigrate(true)
                 .baselineVersion("0")
                 .load();
-        flyway.repair(); // Optionally repair the schema history if needed
-        flyway.migrate();
-        return flyway;
+    }
+
+    // ApplicationRunner bean to run migrations at startup
+    // TODO - Remove this once the pg and release branch is merged
+    @Bean
+    public ApplicationRunner flywayMigrationRunner(Flyway flyway) {
+        return args -> {
+            flyway.repair(); // Fixes any inconsistencies in the schema history
+            flyway.migrate();
+        };
     }
 }
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/FlywayRepairConfig.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/FlywayRepairConfig.java
@@ -1,0 +1,34 @@
+package com.appsmith.server.configurations;
+
+import org.flywaydb.core.Flyway;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.jdbc.DataSourceProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class FlywayRepairConfig {
+
+    private final DataSourceProperties dataSourceProperties;
+
+    @Autowired
+    public FlywayRepairConfig(DataSourceProperties dataSourceProperties) {
+        this.dataSourceProperties = dataSourceProperties;
+    }
+
+    @Bean
+    public Flyway flyway() {
+        Flyway flyway = Flyway.configure()
+                .dataSource(
+                        dataSourceProperties.getUrl(),
+                        dataSourceProperties.getUsername(),
+                        dataSourceProperties.getPassword())
+                .locations("classpath:com/appsmith/server/migrations")
+                .baselineOnMigrate(true)
+                .baselineVersion("0")
+                .load();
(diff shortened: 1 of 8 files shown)


abhvsn (commented):
@AnaghHegde now this is a bean how do we guarantee that flyway migration runs before the other beans are created?

AnaghHegde:
@abhvsn Thanks for highlighting this. This is fixed by using the ApplicationRunner.

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/TenantConfig.java
```diff
@@ -17,7 +17,7 @@
 @Configuration
 @RequiredArgsConstructor
 @Slf4j
-public class TenantConfig implements ApplicationListener<ApplicationStartedEvent> {
+public class TenantConfig implements ApplicationListener<ApplicationReadyEvent> {
```
abhvsn, line 20: I think this was a deliberate choice to avoid any conflicts to fix the tenant object before the application ready event. Any reasone we are changing this?

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/ThemesConfig.java
```diff
@@ -31,20 +33,32 @@
 @Configuration
 @Slf4j
 @RequiredArgsConstructor
-public class ThemesConfig {
+public class ThemesConfig implements ApplicationListener<ApplicationReadyEvent> {
```
abhvsn, line 36: Let's keep all the repetitive migration on ApplicationStartedEvent so that we will be sure on before the InstanceConfig all the migrations are completed.

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/FlywayRepairConfig.java
```diff
@@ -0,0 +1,42 @@
...
+                .locations("classpath:com/appsmith/server/migrations")
+                .baselineOnMigrate(true)
+                .baselineVersion("0")
+                .load();
+    }
+
+    // ApplicationRunner bean to run migrations at startup
+    // TODO - Remove this once the pg and release branch is merged
+    @Bean
+    public ApplicationRunner flywayMigrationRunner(Flyway flyway) {
```
abhvsn, line 36: Can you provide some insights when the flyway bean will be instantiated on the spring lifecycle events and at what point we are expecting the flyway will execute the migrations?

abhvsn (commented):
Left few comments for clarity.

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/TenantConfig.java
AnaghHegde: This is due to the migration issue. This would otherwise throw error that table does not exists.

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/ThemesConfig.java
AnaghHegde: This is due to the migration issue. This would otherwise throw error that table does not exists.

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/FlywayRepairConfig.java
AnaghHegde: @abhvsn This is done during ApplicationStartedEvent and flyway executes the migration and then the repetable migrations are executed.

abhvsn: @AnaghHegde I don't think this will work as expected when we add a new field to the entity class as the structure of the table will be affected and repository beans will start complaining. Can we make sure the flyway migration is executed even before the beans are executed just like what it's present today?

AnaghHegde: @abhvsn How does this cause the ` repository beans will start complaining`

abhvsn: @AnaghHegde IMO this will be the case when we update the entity object which getting stored in the DB.

abhvsn (changes requested):
@AnaghHegde let's speed up this, migration script is at risk because of this change.

abhvsn:
Blocking: https://github.com/appsmithorg/appsmith/issues/37043

PR #37108, opened 2024-10-28
fix: resolve empty table dropdown issue with dynamic select options in add new row functionality

### LINKED ISSUE #23470: [Bug]: New Row Select Options Are Not Supporting Dynamic Data If TableData is Empty

### Description

When the table for which a Select column is populated dynamically contains no data, and a user tries to add a new row, the select options are not displayed. However, this works fine if the Table already contains at least one row.

[Front thread](https://app.frontapp.com/open/msg_1sj5ykwc?key=FRaV8YIIHi4a0ocls0zJ7Zrd5lihh9Rr)

### Steps To Reproduce

1. Add a Table widget to the canvas and bind it to a query.
2. Allow adding a new row, make the columns editable, and set one of the columns to be of type select.
3. Bind the options of the select column to another query.
4. Ensure that the query that populates the table does not return any data and try to add a new row. Note that no select options are shown for the column.

### Public Sample App

_No response_

### Environment

Production

### Issue video log

https://www.loom.com/share/f4b09c520cfb4d15b362775382121e06

### Version

Appsmith v1.9.19-SNAPSHOT

Front logo [Front conversations](https://app.frontapp.com/open/top_avn4c)

jacquesikot:
## Description
**Problem**
When using a Table widget's Select column type with dynamic options, the computed value binding fails to handle empty table states correctly. If the table has no data (`processedTableData` is empty), the dynamic options evaluation still attempts to map over the non-existent table data, resulting in an empty array instead of the expected options.

**Root Cause**
The issue stems from the `getComputedValue` function always using the table mapping binding prefix:
```typescript
{{${tableName}.processedTableData.map((currentRow, currentIndex) => (
  // dynamic options here
))}}
```
This creates an unnecessary dependency on table data even when the dynamic options don't reference `currentRow` or `currentIndex`, causing evaluation to fail when the table is empty.

### Problematic Evaluation
When the table is empty, expressions like this in table widget computed properties:
```typescript
{{[
  { label: "Released", value: "Released" },
  { label: "Not Released", value: "Not Released" }
]}}
```
Would evaluate to an empty array `[]` because it's wrapped in a `.map()` over empty table data.

**Solution**
Updated the binding logic to account for scenarios where table does not have data and return the evaluated string directly in an IIFE

1. Updated the binding prefix and suffix
```typescript
static getBindingPrefix = (tableName: string) => {
  return `{{(() => { const tableData = ${tableName}.processedTableData || []; return tableData.length > 0 ? tableData.map((currentRow, currentIndex) => (`;
};

static getBindingSuffix = (stringToEvaluate: string) => {
  return `)) : ${stringToEvaluate} })()}}`;
};
```

2. Refactored `getComputedValue` and `getInputComputedValue` to implement the new bindings
3. Created a migration and migration test for the DSL change

This ensures that:
- Dynamic options not dependent on table context evaluate correctly even with empty tables
- The component maintains consistent behaviour across all table states

The solution prevents unnecessary table data dependencies while preserving the ability to use table-specific values when required.

Fixes #23470  

## Automation

/ok-to-test tags="@tag.Table, @tag.Binding, @tag.Select, @tag.Sanity, @tag.Widget"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/13514895959>
> Commit: 0d2e78a0a7be63d4f70fc3499829621bd1761b3d
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=13514895959&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Table, @tag.Binding, @tag.Select, @tag.Sanity, @tag.Widget`
> Spec:
> <hr>Tue, 25 Feb 2025 07:52:52 UTC

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **New Features**
	- Enhanced test coverage for adding new rows in the `TableV2` widget, ensuring proper UI behavior when no data exists.
- **Bug Fixes**
	- Improved validation of UI elements based on the "Allow adding a row" property.
- **Refactor**
	- Streamlined logic for handling computed values in the `ComputeTablePropertyControlV2`, improving readability and functionality.


app/client/src/components/propertyControls/TableComputeValue.tsx
```diff
@@ -175,24 +175,51 @@ class ComputeTablePropertyControlV2 extends BaseControl<ComputeTablePropertyCont
...
+    return this.wrapInDoubleCurlyBraces(stringToEvaluate);
+  };
+
+  shouldReturnValueDirectly = (value: string) => {
+    return (
+      !isDynamicValue(value) && !this.props.additionalControlData?.isArrayValue
+    );
+  };
+
+  containsTableSpecificBindings = (stringToEvaluate: string) => {
```
jsartisan, line 206: The functions that are extracted are just too shallow. I would do such abstractions only if they are used at atleast 3 places or doing something complex. But in this case, it is very straight forward, so extracting thesr won't help much, insteads, makes the code a little complex.

app/client/src/components/propertyControls/TableComputeValue.tsx
```diff
@@ -175,24 +175,51 @@ class ComputeTablePropertyControlV2 extends BaseControl<ComputeTablePropertyCont
...
 
+    // Convert string to JavaScript expression
     const stringToEvaluate = stringToJS(value);
 
+    // Return empty if the evaluated string is empty
     if (stringToEvaluate === "") {
       return stringToEvaluate;
     }
 
+    // Handle specific cases with "currentIndex" or "currentRow"
```
jsartisan, line 191: The comment is vey vague. Can you add a little more context why we need to handle this specific case?

app/client/src/components/propertyControls/TableComputeValue.tsx
rahulbarwal: @jsartisan I want to understand this better. Can you tell in detail, or share a link that explains this concept?

jsartisan: The main idea is that modules should be deep. Methods like wrapInDoubleCurlyBraces and shouldReturnValueDirectly are too shallow - they don't hide enough complexity. These shallow methods actually increase cognitive load rather than reducing it.

I definitely recommend reading "Philosphy of software design".

jsartisan (approved):

app/client/src/components/propertyControls/TableComputeValue.tsx
```diff
@@ -157,21 +155,26 @@ class ComputeTablePropertyControlV2 extends BaseControl<ComputeTablePropertyCont
...
+
+    // Find the first closing parenthesis after the map
+    const valueEnd = propertyValue.indexOf("))", valueStart);
+
+    if (valueEnd === -1) return propertyValue;
+
+    // Extract the value between the map parentheses
+    const evaluateString = propertyValue.substring(valueStart, valueEnd);
+
+    return JSToString(evaluateString);
```
rahulbarwal, line 177: This has become complex from its previous implementation.
Few pointers:
1. Lot of strings, which ideally should be constants.
2. Recommendation: use string interpolation more(as it was done earlier)
3. Add better comments.

app/client/src/components/propertyControls/TableComputeValue.tsx
```diff
@@ -182,15 +185,23 @@ class ComputeTablePropertyControlV2 extends BaseControl<ComputeTablePropertyCont
       return value;
     }
 
+    // Convert string to JavaScript expression
```
rahulbarwal, line 188: This comment does not tell me anything that code does not.

app/client/src/components/propertyControls/TableComputeValue.tsx
```diff
@@ -182,15 +185,23 @@ class ComputeTablePropertyControlV2 extends BaseControl<ComputeTablePropertyCont
...
     const stringToEvaluate = stringToJS(value);
 
+    // Return empty if the evaluated string is empty
```
rahulbarwal, line 191: same as above.

app/client/src/components/propertyControls/TableComputeValue.tsx
```diff
@@ -99,11 +99,13 @@ function InputText(props: InputTextProp) {
...
-  static getBindingPrefix(tableName: string) {
-    return `{{${tableName}.processedTableData.map((currentRow, currentIndex) => ( `;
-  }
+  static getBindingPrefix = (tableName: string) => {
+    return `{{(() => { const tableData = ${tableName}.processedTableData || []; return tableData.length > 0 ? tableData.map((currentRow, currentIndex) => (`;
+  };
 
-  static bindingSuffix = `))}}`;
+  static getBindingSuffix = (stringToEvaluate: string) => {
+    return `)) : ${stringToEvaluate} })()}}`;
```
rahulbarwal, line 107: Shouldn't `stringToEvaluate` be present in both conditions of ternary operator?

rahulbarwal (changes requested):

app/client/src/components/propertyControls/TableComputeValue.tsx
jacquesikot: It is, it was added after the getBindingPrefix in the buildTableSpecificBinding below. I have moved it here for easier understanding

app/client/src/components/propertyControls/TableComputeValue.tsx
jacquesikot: Got it

app/client/src/components/propertyControls/TableComputeValue.test.tsx
```diff
@@ -0,0 +1,172 @@
...
+	]
+    `;
+    const computedValue = ComputeTablePropertyControlV2.getTableComputeBinding(
+      tableName,
+      input,
+    );
+
+    expect(
+      ComputeTablePropertyControlV2.getInputComputedValue(computedValue),
+    ).toBe(`{{${input}}}`);
```
rahulbarwal, line 38: Input initialization and output verification repitition is unnecessary. 
What I mean is you can have something like:
```
const checks = ["currentRow.price",  `
    [
		{
			"value": "male",
			"label": "male"
		},
		{
			"value": "female",
			"label": "female"
		}
	]
    `, ["123", "-456", "0.123", "-0.456"], 
.... and so on
]

for(const input  of checks) {
	const computedValue = ComputeTablePropertyControlV2.getTableComputeBinding(
      tableName,
      input,
    );

expect(
      ComputeTablePropertyControlV2.getInputComputedValue(computedValue),
    ).toBe(`{{${input}}}`);
}
```

this way you will have only one unit test with a lot of inputs.

rahulbarwal (changes requested):

rahulbarwal (approved):

PR #37121, opened 2024-10-29
chore: Removed manipulation for custom traceparent header

nidhi-nair:
## Description
All Appsmith systems should always rely on a single traceparent header to propagate traces, removing incorrect manipulation to bring back consistency.

## Automation

/ok-to-test tags=""

> [!WARNING]
> Tests have not run on the HEAD 3aeba4a6182aa740dbcf56ec6ef10e9542a4e16e yet
> <hr>Wed, 30 Oct 2024 03:57:35 UTC

## Communication
- [ ] Yes
- [ ] No

## Release Notes

- **New Features**
  - Introduced new observability properties for enhanced metrics tracking, including support for New Relic.
  - Added a new constant for trace management to improve logging capabilities.

- **Bug Fixes**
  - Removed outdated web filter to streamline HTTP request tracing.

- **Configuration Changes**
  - Updated service names in observability configurations for clarity.
  - Simplified tracing configuration to include all spans without filtering.

These updates enhance the application's observability, allowing for better tracking and monitoring of performance metrics.


app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/TracingConfig.java
```diff
@@ -42,19 +36,4 @@ ObservationPredicate noActuatorServerObservations() {
             }
         };
     }
-
-    @Bean
-    SpanExportingPredicate onlyAppsmithSpans() {
```
ApekshaBhosale, line 47: @nidhi-nair now this has been removed. we will allow all spans? won't there be too many and cause clutter?

ApekshaBhosale (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/TracingConfig.java
nidhi-nair: I'd like to see if this is still true post spring upgrade since the DP is not showing as many spans anymore. We can revisit this if we see a spike perhaps? This allows us to fill the gaps in the trace lineage.

ApekshaBhosale: yes makes sense

ApekshaBhosale (approved):

PR #37128, opened 2024-10-30
fix: JSON form validation trigger on child component update

### LINKED ISSUE #28018: [Bug]: 'Required' Field Validation in JSON Form Not Working Correctly If Fields Are Hidden

### Description

If we have 2 fields and want to alternatively make them required and visible, based on the selection from a dropdown, the Required validation does not work as expected and the Submit button stays disabled.
It seems that once a field was marked as Required and it gets programmatically hidden, the Required mark does not get removed programmatically as it should (it ignores the JS condition).
See the screen recording and sample app for more details.
[Front thread](https://app.frontapp.com/open/cnv_153r29e4?key=lREPusUPGTClzGmI2uGzY9hO7igwZZUK)

### Steps To Reproduce

1. Use a **JSON Form** having one select "Type" (options Red and Blue) and two **Input widgets** labeled "Red" and "Blue".
2. If the value of "Type" is "Red", then the "Red" field should be visible and required; also the "Blue" field should be non-required and hidden.
3. If the value of "Type" is "Blue", then the "Blue" field should be visible and required; also the "Red" field should be non-required and hidden.
4. For the Red Field in the required js put `{{formData.Type == "Red"}}` and visibility js put `{{formData.Type == "Red"}}`
5. For the Blue Field in the required js put `{{formData.Type == "Blue"}}` and visibility js put `{{formData.Type == "Blue"}}`
6. Now if the Type is selected as "Red", only the Red field should be visible and required, but this is not happening if we switch between red and blue in the type, it also considers both fields required validation.

### Public Sample App

https://app.appsmith.com/app/jsonform-validation-required/page1-65280f6880b32971dff37e3c

### Environment

Production

### Issue video log

https://www.loom.com/share/50c7b4b6562544dfa598216aa82094fc

### Version

Cloud v1.9.40

rahulbarwal:
## Description
<ins>Problem</ins>

Form validation was not triggered when the child component was updated, resulting in inconsistent data consistency.

<ins>Root cause</ins>

The `Form` component in `app/client/src/widgets/JSONFormWidget/component/Form.tsx` did not include the `trigger` function from the `methods` object, preventing form validation from being triggered on child component updates.

<ins>Solution</ins>

This PR adds the `trigger` function from the `methods` object to the `Form` component, ensuring form validation is triggered correctly when the child component is updated. 

* Adds unit tests for `Form` component as well

Fixes #28018
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/11697880527>
> Commit: 1c38b05c04d9ed2c83d09a90d5fec720e30d4daf
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=11697880527&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.JSONForm`
> Spec:
> <hr>Wed, 06 Nov 2024 06:06:41 UTC

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **New Features**
	- Enhanced form validation lifecycle management with the introduction of the `useUnmountFieldValidation` hook for better handling of field validation upon unmounting.
	- Improved testability of the form component through the inclusion of `data-testid` attributes for the submit and reset buttons.

- **Bug Fixes**
	- Resolved edge cases in form validation, ensuring components function correctly with changing props and handle empty schemas gracefully.

- **Tests**
	- Introduced a comprehensive suite of unit tests for the `Form` component, covering various scenarios including validation and visibility management.
	- Added tests for the new `useUnmountFieldValidation` hook to ensure correct validation behavior during unmounting.


app/client/src/widgets/JSONFormWidget/component/Form.tsx
```diff
@@ -273,6 +273,12 @@ function Form<TValues = any>(
     onFormValidityUpdate(!isFormInValid);
   }, [onFormValidityUpdate, isFormInValid]);
 
+  useEffect(() => {
+    // This effect is to handle an edge-case:
+    // https://github.com/appsmithorg/appsmith/issues/28018
+    trigger();
```
ashit-rath, line 279: I didn't quite understood the solution here, why are we re-validating the whole form when children changes? Isn't it a problem of when reset button is pressed the validation persists?

jacquesikot: @ashit-rath I don't think so, the problem is the validation not getting triggered when children changes. When reset button is pressed, then the validation change is triggered.

app/client/src/widgets/JSONFormWidget/component/Form.tsx
```diff
@@ -273,6 +273,12 @@ function Form<TValues = any>(
...
+  }, [children, trigger]);
+
```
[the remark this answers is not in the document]
jacquesikot: @rahulbarwal I agree with this. Can we find some state that changes when formData changes, and then add that as a dependency in the useEffect, instead of this hasVisibilityChanged solution above?

jacquesikot (dismissed):
Can we find some state that changes when formData changes, and then add that as a dependency in the useEffect

app/client/src/widgets/JSONFormWidget/component/Form.tsx
rahulbarwal: @ashit-rath 
Let me take you with a dry run of whats happening in the original issue #28018
1. User select red and get one input which bound to red. It is empty hence, errors has `{red: required:true}`.
2. user adds value, errors are cleared. and user can submit.
3. Now they click reset, the input is cleared and errorState has this: `{red: required:true}`
4. Now user switches to blue, and gets input bound to blue, which adds `{blue: required:true}`
5. Over all errorState looks liks: `{red: required:true, blue: required:true}`
6. Now even if you type in blue input, the red error never goes away. 

I am open to ideas if this call to `trigger` can be optimized. Looking forward to your ideas.

ashit-rath: Thanks @rahulbarwal and @jacquesikot for explaining this in detail, I understood the problem. I believe this is a reset problem; if you look at the `RESET_OPTIONS` which gets passed to `onReset` has `keepErrors: true`. Have we checked if this was set to `false` for the case when button is clicked; solves the problem?

rahulbarwal: Thanks @ashit-rath for this input. 
`keepErrors: false` did not work, and I think I know why: 
In step `3` above, when user resets the form, the red input is visible and Here, whether we keep or discard errors does not matter. 
IT is on step `4` that multiple errors enter the formState(where we are not doing the reset).

rahulbarwal: Thanks @ashit-rath, for this feedback, I've incorporated these changes. please review again.

app/client/src/widgets/JSONFormWidget/component/Form.tsx
```diff
@@ -273,6 +274,60 @@ function Form<TValues = any>(
     onFormValidityUpdate(!isFormInValid);
   }, [onFormValidityUpdate, isFormInValid]);
 
+  useEffect(() => {
+    if (!schema || Object.keys(schema).length === 0) {
+      return;
+    }
+
+    const computeFieldVisibilityMap = (
```
ashit-rath, line 282: Why aren't we doing it at the field level? Check `useUpdateInternalMetaState`/`useRegisterFieldValidity`.

rahulbarwal: Done.

ashit-rath (approved):

PR #37377, opened 2024-11-13
chore: Remove `/restart` API

sharat87:
This PR removes the `/restart` API. Currently, the client decides whether a restart is needed after saving some settings. Depending on that decision, a call is made to `/restart` and the server then restarts. This PR flips this by having the server take the decision, and merely informing the client of the decision.

This way, we don't need a `/restart` API. Fewer things there are, fewer things we have to protect.

## Automation

/test sanity settings

> [!WARNING]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/11819840608>
> Commit: 4195fdf1d1dbd234bdcfde462b1036e4fd8fee25
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=11819840608&attempt=2" target="_blank">Cypress dashboard</a>.
> Tags: @tag.Sanity, @tag.Settings
> Spec: 
> It seems like **no tests ran** 😔. We are not able to recognize it, please check <a href="https://github.com/appsmithorg/appsmith/actions/runs/11819840608" target="_blank">workflow here</a>.
> <hr>Thu, 14 Nov 2024 08:57:56 UTC

## Communication
- [ ] Yes
- [x] No

### DIFF
--- a/app/client/cypress/e2e/Regression/ClientSide/AdminSettings/Admin_settings_spec.js
+++ b/app/client/cypress/e2e/Regression/ClientSide/AdminSettings/Admin_settings_spec.js
@@ -181,9 +181,6 @@ describe("Admin settings page", { tags: ["@tag.Settings"] }, function () {
       fromAddress = uuid;
       cy.get(adminsSettings.fromAddress).clear().type(`${uuid}@appsmith.com`);
     });
-    cy.intercept("POST", "/api/v1/admin/restart", {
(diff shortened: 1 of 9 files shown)


sharat87:
Super user signup fails because it sets the `APPSMITH_ADMIN_EMAILS` env variable, which is triggering a restart. It shouldn't do a restart in that case.

Issue #37151, opened 2024-10-31
[Task] Analyse the JSObject creation flow metrics


rishabhrathod01:
This task analyses the logs when:

- JS object is created and we copy paste another huge JS object into this one, typically when we paste a jsobject which creates 35-40 new actions, the update API times out.

sneha122:
The log analysis is done by following the traces mentioned here: https://onenr.io/0VwgZL0K0QJ

Considering JS object which is creating 40 new actions: https://onenr.io/0oQDqnmGNwy
Each action is created with createAction function call. From the traces above we can see that if create action takes ~17s of time, 1st component takes around ~14s, 2nd component takes around 2s and last one takes around few ms.
<img width="783" alt="Screenshot 2024-11-11 at 8 04 03 PM" src="https://github.com/user-attachments/assets/c807a25d-79b7-4108-8252-254f952eeddd">

After analysing the logs, we can see that there are three components that are causing delays:
1. When we check if the name is allowed or not -> This component takes 80-90% of the total create action time, if we optimise this function we can take care of most of the timeout issues
3. Fetching the page data from DB -> This is the second time consuming part, it takes around 5-7% of the create action time
4. Saving the actions in DB -> This takes around 3-5% of the create action time (Only in exceptional cases it takes more)

We should definitely prioritise and fix 1st component and subsequently rest of the components can be taken up.

CC: @rishabhrathod01

sneha122:
Closing this issue and will create separate tasks for each optimisation proposed here

PR #37391, opened 2024-11-14
chore: perf optimisation for js action creation phase 1

### LINKED ISSUE #37365: [Task]: Optimise isNameAllowed method in JS object action creation
### SubTasks

The log analysis is done by following the traces mentioned here: https://onenr.io/0VwgZL0K0QJ

Considering JS object which is creating 40 new actions: https://onenr.io/0oQDqnmGNwy
Each action is created with createAction function call. From the traces above we can see that if create action takes ~17s of time, 1st component takes around ~14s, 2nd component takes around 2s and last one takes around few ms.
Screenshot 2024-11-11 at 8 04 03 PM

This task is to optimise following method -> check if the name is allowed or not -> This component takes 80-90% of the total create action time, if we optimise this function we can take care of most of the timeout issues

sneha122:
## Description

[During analysis of action creation flow metrics](https://github.com/appsmithorg/appsmith/issues/37151#issuecomment-2468354426), we observed that RefactoringService.isNameAllowed is taking 80-90% of the total JS object action time. This PR optimises this part in a way that for any jsobject action, instead of fetching all actions from DB and comparing it 
to see if current action name is allowed, we simply do that check in memory where for current action collection, if any action names are being duplicated, we throw the error.

We could make this change easily because recently we merged a [PR](https://github.com/appsmithorg/appsmith/pull/36958) which removes the actions with duplicate name from client payload whenever Js object update API is called, with this change, we can guarantee that for any JS object update call, all actions inside it will always have unique names. This PR makes the similar check on backend where if any action has duplicate name within collection, we throw an error and don't store that action in the DB.

We may need to consider following test case in both before and after implementation of this approach. This can be covered during PR testing:
What happens if the client sends multiple requests to add a new function in an existing collection. That is, as a result of the debounce logic, if the server receives 2 consecutive requests with a populated collection but without actionId associated to either request.

Relevant thread: https://theappsmith.slack.com/archives/C040LHZN03V/p1731571364933089

Fixes #37365 
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags="@tag.JS"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/11911295324>
> Commit: d5c75edd301e75b2432b642f366bc80c6fea6b89
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=11911295324&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.JS`
> Spec:
> <hr>Tue, 19 Nov 2024 11:14:16 UTC

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **New Features**
	- Enhanced validation to prevent the creation of actions with duplicate names in action collections.
	- Simplified handling of JavaScript actions, allowing them to bypass certain validation checks.

- **Bug Fixes**
	- Improved error handling during action updates and collection modifications to ensure better logging and management of failures.

- **Tests**
	- Added tests to verify that duplicate action names trigger appropriate error messages, enhancing the robustness of the action collection feature.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/LayoutCollectionServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/LayoutCollectionServiceCEImpl.java
@@ -317,7 +317,7 @@ public class LayoutCollectionServiceCEImpl implements LayoutCollectionServiceCE 
 
         // If duplicate action name exists, throw an error
         final Map<String, Long> actionNameCountMap = actionCollectionDTO.getActions().stream()
-                .collect(Collectors.groupingBy(ActionDTO::getName, Collectors.counting()));
+                .collect(Collectors.groupingBy(ActionDTO::getValidName, Collectors.counting()));
         List<String> duplicateNames = actionNameCountMap.entrySet().stream()
                 .filter(entry -> entry.getValue() > 1)
                 .map(Map.Entry::getKey)
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/LayoutCollectionServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/LayoutCollectionServiceCEImpl.java
@@ -315,18 +315,9 @@ public class LayoutCollectionServiceCEImpl implements LayoutCollectionServiceCE 
         final Set<String> baseActionIds = new HashSet<>();
         baseActionIds.addAll(validBaseActionIds);
 
-        // If duplicate action name exists, throw an error
-        final Map<String, Long> actionNameCountMap = actionCollectionDTO.getActions().stream()
-                .collect(Collectors.groupingBy(ActionDTO::getValidName, Collectors.counting()));
-        List<String> duplicateNames = actionNameCountMap.entrySet().stream()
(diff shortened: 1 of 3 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/LayoutCollectionServiceCEImpl.java
```diff
@@ -335,11 +339,23 @@ public Mono<ActionCollectionDTO> updateUnpublishedActionCollection(
                                 actionDTO.setPluginType(actionCollectionDTO.getPluginType());
                                 actionDTO.setPluginId(actionCollectionDTO.getPluginId());
                                 actionDTO.setBranchName(branchedActionCollection.getBranchName());
+
+                                boolean isDuplicateActionName = existingActions.stream()
```
nidhi-nair, line 343: Why are we checking only against existing names and not against all incoming actions?

sneha122: makes sense, ill check for all

NilanshBansal (approved):
LGTM

PR #37666, opened 2024-11-25
fix: Restore doesn't work when backup file is renamed

sharat87:
When backups are renamed from their original generated name, restore operation can't process that backup anymore. This PR should fix that.

## Automation

/test sanity

> [!CAUTION]  

## Communication
- [ ] Yes
- [x] No


sharat87:
@coderabbitai review

mohanarpit (approved):

PR #37747, opened 2024-11-26
fix: Resolve list widget Error

### LINKED ISSUE #37683: [Bug]: List widget redirects to error page

### Description

The list widget redirects to an error page due to auto-height updates.

### Steps To Reproduce

- Add a List Widget and a Button Widget.
- Configure the Button Widget to set the visibility of the List Widget to true when clicked.
- Configure the List Widget so that clicking on any of its items sets its visibility to false.
- In preview mode:
  - Click on any item in the List Widget to hide it except the last item.
  - Click the Button Widget to display the List Widget again.
  - Observe the error that occurs, as shown in the attached video.

### Public Sample App

_No response_

### Environment

Production

### Severity

### Issue video log

https://www.loom.com/share/ebf896dc35f04ac187350c2d47be178a?sid=77601f0a-76ec-4ebe-8493-f13f39d1f8ee

### Version

Cloud

saharAdem:
## Description
This PR resolves an issue with the List widget caused by infinite auto height update loops. The child widgets of the List widget already have the isMetaWidget flag, which prevents auto height updates. However, the List widget itself did not have this flag, leading to the issue. This fix ensures that the isMetaWidget flag is also applied to the List widget, preventing the infinite update loop.

Fixes #37683

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **Tests**
	- Added a new Cypress test specification for the List widget's auto-height functionality
	- Verifies widget behavior during interactions and visibility changes

- **Bug Fixes**
	- Improved logic for handling auto-height updates in List widgets
	- Prevented potential infinite loop scenarios during widget interactions

### DIFF
--- a/app/client/cypress/e2e/Regression/ClientSide/BugTests/ListWidget_Bug37683_spec.ts
+++ b/app/client/cypress/e2e/Regression/ClientSide/BugTests/ListWidget_Bug37683_spec.ts
@@ -1,0 +1,42 @@
+import {
+    agHelper,
+    locators,
+    propPane,
+  } from "../../../../support/Objects/ObjectsCore";
+ 
+  const widgetSelector = (name: string) => `[data-widgetname-cy="${name}"]`;
+  const containerWidgetSelector = `[type="CONTAINER_WIDGET"]`;
+ 
+  describe("Bug 37683: List widget auto-height update infinite loop error", () => {
+    it("The list widget height should not trigger an auto-height update", () => {
+      cy.dragAndDropToCanvas("listwidgetv2", {
+        x: 300,
+        y: 600,
+      });
+      propPane.EnterJSContext(
+        "onitemclick",
+        `{{List1.setVisibility(false)}}`,
+        true,
+        false,
+      );
+      cy.dragAndDropToCanvas("buttonwidget", {
+        x: 600,
+        y: 600,
+      });
+      propPane.EnterJSContext(
+        "onClick",
+        `{{List1.setVisibility(true)}}`,
+        true,
+        false,
+      );
+      agHelper.GetNClick(locators._enterPreviewMode);
+      cy.get(`${widgetSelector("List1")} ${containerWidgetSelector}`)
+        .eq(1)
+        .click({ force: true });
+      cy.wait(4000);
+      agHelper.GetNClick(`${widgetSelector("Button1")}`);
+      agHelper.AssertElementVisibility(locators._widgetInDeployed("list1"));
+      agHelper.AssertElementVisibility(locators._widgetInDeployed("button1"));
+      agHelper.GetNClick(locators._exitPreviewMode);
+    });
+  });
--- a/app/client/src/widgets/ListWidgetV2/widget/defaultProps.ts
+++ b/app/client/src/widgets/ListWidgetV2/widget/defaultProps.ts
@@ -56,6 +56,7 @@ export default {
   minWidth: FILL_WIDGET_MIN_WIDTH,
   responsiveBehavior: ResponsiveBehavior.Fill,
   flexVerticalAlignment: FlexVerticalAlignment.Top,
+  isMetaWidget: true,
   dynamicBindingPathList: [
     {
       key: "currentItemsView",
--- a/app/client/src/widgets/ListWidgetV2/widget/defaultProps.ts
+++ b/app/client/src/widgets/ListWidgetV2/widget/defaultProps.ts
@@ -56,7 +56,6 @@ export default {
   minWidth: FILL_WIDGET_MIN_WIDTH,
   responsiveBehavior: ResponsiveBehavior.Fill,
   flexVerticalAlignment: FlexVerticalAlignment.Top,
-  isMetaWidget: true,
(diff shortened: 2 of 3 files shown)


saharAdem:
@albinAppsmith review

rahulbarwal (commented):
Thanks @saharAdem.
While we are reviewing this for any side effects. Can you please add some automation tests to ensure that we don't miss this again?

saharAdem:
Hi @rahulbarwal,
I added tests. Could you please check the PR?

jacquesikot:
@saharAdem I think the solution is incorrect here, we shouldn't add `isMetaWidget` to the default props as it would lead to isMetaWidget: true for every dropped list widget and this will have some implications.
I was checking out the issue and it seems if I have 2 buttons outside the list, use one to close and the other to open; it does not crashes so something else goes wrong when the list item is clicked and List1.setVisibility(false) is triggered.

jacquesikot (changes requested):
I think the solution is incorrect here, we shouldn't add isMetaWidget to the default props as that would lead to isMetaWidget: true for every dropped list widget and this will have some implications

saharAdem:
I believe the issue is caused by an infinite loop of auto-height updates. I will investigate further. However, this has occurred multiple times when using the list widget.

saharAdem:
> @saharAdem I think the solution is incorrect here, we shouldn't add `isMetaWidget` to the default props as it would lead to isMetaWidget: true for every dropped list widget and this will have some implications. I was checking out the issue and it seems if I have 2 buttons outside the list, use one to close and the other to open; it does not crashes so something else goes wrong when the list item is clicked and List1.setVisibility(false) is triggered.

@jacquesikot Thanks for your explanation. The issue occurs due to auto-height updates in the list widget, which causes an infinite loop. 

Any list widget should be prevented from auto height updates. The inner widgets of the list widget are already excluded from auto-height updates using the `isMetaWidget` property that added in the [MetaWidgetGenerator.ts](https://github.com/appsmithorg/appsmith/blob/release/app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.ts) file 
. as outlined [here](https://github.com/appsmithorg/appsmith/blob/dfd7fdee97fc2afc3c08a6924f232c02a91f6874/app/client/src/widgets/withWidgetProps.tsx#L295-L310) 


```
         if (shouldCollapseWidgetInViewOrPreviewMode) {
        // This flag (isMetaWidget) is used to prevent the Auto height saga from updating
        // the List widget Child Widgets. Auto height is disabled in the List widget and
        // this flag serves as a way to avoid any unintended changes to the child widget's height.
        if (
          widgetProps.bottomRow !== widgetProps.topRow &&
          !widgetProps.isMetaWidget
        ) {
          dispatch({
            type: ReduxActionTypes.UPDATE_WIDGET_AUTO_HEIGHT,
            payload: {
              widgetId: props.widgetId,
              height: 0,
            },
          });
        }
```

Since the existing code and logic in the [MetaWidgetGenerator.ts](https://github.com/appsmithorg/appsmith/blob/release/app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.ts) are based on applying `isMetaWidget` to the list widget's child widgets, and to avoid unintended consequences, auto-height updates for the list widget itself should also be disabled. I have added some failing use cases to the issue #37683  which you can review there.

torcoste:
> @jacquesikot 
> it seems if I have 2 buttons outside the list, use one to close and the other to open; it does not crashes so something else goes wrong when the list item is clicked and List1.setVisibility(false) is triggered.

I have a hypothesis: when List1 visibility is changed to false inside the list item click handler, List1 is demounted from the page before some side effects of list item click handler are executed. Thus, when List1 is mounted on the page again, a crash occurs.

Factors that indicate this hypothesis is true: 
- crash does not occur in developer mode, where when visability is changed, the widget is not demounted, but only becomes semi-transparent.
- If before mounting List1 back on the page we reset its state using resetWidget, there is no crash.
```js
resetWidget("List1").then(() => List1.setVisibility(true))
```

I attach an appsmith application that demonstrates this workaround
[ListWidgetIssue37747-workarround.json](https://github.com/user-attachments/files/18489388/ListWidgetIssue37747-workarround.json)

I hope this helps solve the problem sooner 🙏

saharAdem:
Hey @jacquesikot,  any updates on this?

jacquesikot:
Thanks @saharAdem @torcoste I am investigating this further.

jacquesikot:
Thanks @saharAdem and @torcoste for your inputs, I have raised a PR to fix this issue. I have added a detailed description of my final findings and solution. https://github.com/appsmithorg/appsmith/pull/39262

adrianwysocki1993:
[heart]         Adrian Wysocki reacted to your message:
________________________________

adrianwysocki1993:
[celebrate]     Adrian Wysocki reacted to your message:
________________________________

jacquesikot:
@saharAdem Please try this [dp](https://ce-39262.dp.appsmith.com/) too, let me know if it works for you, thanks

PR #37776, opened 2024-11-27
chore: draft GS changes

sneha122:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/FileListMethod.java
+++ b/app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/FileListMethod.java
@@ -40,7 +40,7 @@ public class FileListMethod implements ExecutionMethod, TriggerMethod {
     public WebClient.RequestHeadersSpec<?> getExecutionClient(WebClient webClient, MethodConfig methodConfig) {
         UriComponentsBuilder uriBuilder = getBaseUriBuilder(
                 this.BASE_DRIVE_API_URL,
-                "?includeItemsFromAllDrives=true&supportsAllDrives=true&orderBy=name&q=mimeType%3D'application%2Fvnd.google-apps.spreadsheet'%20and%20trashed%3Dfalse",
+                "?includeItemsFromAllDrives=true&supportsAllDrives=true&corpora=allDrives&orderBy=name&q=mimeType%3D'application%2Fvnd.google-apps.spreadsheet'%20and%20trashed%3Dfalse",
                 true);
 
         return webClient
--- a/app/server/appsmith-plugins/googleSheetsPlugin/src/main/resources/form.json
+++ b/app/server/appsmith-plugins/googleSheetsPlugin/src/main/resources/form.json
@@ -38,6 +38,18 @@
             {
               "label": "Read / Write / Delete | Selected google sheets",
               "value": "https://www.googleapis.com/auth/drive.file"
+            },
+            {
+              "label": "Read / Write / Delete | All google sheets",
+              "value": "https://www.googleapis.com/auth/spreadsheets,https://www.googleapis.com/auth/drive"
+            },
+            {
+              "label": "Read / Write | All google sheets",
+              "value": "https://www.googleapis.com/auth/spreadsheets,https://www.googleapis.com/auth/drive.readonly"
+            },
+            {
+              "label": "Read | All google sheets",
+              "value": "https://www.googleapis.com/auth/spreadsheets.readonly,https://www.googleapis.com/auth/drive.readonly"
             }
           ],
           "initialValue": "https://www.googleapis.com/auth/drive.file",


app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/FileListMethod.java
```diff
@@ -37,10 +39,16 @@ public boolean validateExecutionMethodRequest(MethodConfig methodConfig) {
     }
 
     @Override
-    public WebClient.RequestHeadersSpec<?> getExecutionClient(WebClient webClient, MethodConfig methodConfig) {
+    public WebClient.RequestHeadersSpec<?> getExecutionClientWithFlags(
+            WebClient webClient, MethodConfig methodConfig, Map<String, Boolean> featureFlagMap) {
+        // Flags are needed here for google sheets integration to support shared drive behind a flag
+        // Once thoroughly tested, this flag can be removed
+        Boolean isSharedDriveSupportEnabled = featureFlagMap.getOrDefault(
+                FeatureFlagEnum.release_google_sheets_shared_drive_support_enabled.toString(), Boolean.FALSE);
```
abhvsn, line 47: Any reason we are not passing the Map<FeatureFlagEnum, Boolean>? That way we don't have to use `toString()`.

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/TriggerMethod.java
```diff
@@ -21,7 +22,17 @@ public interface TriggerMethod {
     /**
      * Returns with the specification required to hit that particular trigger request
      */
-    WebClient.RequestHeadersSpec<?> getTriggerClient(WebClient webClient, MethodConfig methodConfig);
+    default WebClient.RequestHeadersSpec<?> getTriggerClient(WebClient webClient, MethodConfig methodConfig) {
```
abhvsn, line 25: If this is internal to GSheet plugin do we need this method anymore? Can we just rely on `getTriggerClientWithFlags`?

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/plugins/GoogleSheetsPlugin.java
```diff
@@ -81,6 +81,23 @@ public Mono<ActionExecutionResult> executeParameterized(
...
                 DatasourceConfiguration datasourceConfiguration,
                 ActionConfiguration actionConfiguration) {
+            return executeParameterizedWithFlags(
+                    connection, executeActionDTO, datasourceConfiguration, actionConfiguration, null);
+        }
+
+        @Override
+        public Mono<TriggerResultDTO> trigger(
+                Void connection, DatasourceConfiguration datasourceConfiguration, TriggerRequestDTO request) {
+            return triggerWithFlags(connection, datasourceConfiguration, request, null);
```
abhvsn, line 91: Should we throw unsupported method exception here considering we will always be using the method with feature flag param? Are there any flows where we are still using this method?

app/server/appsmith-server/src/main/java/com/appsmith/server/plugins/solutions/PluginTriggerSolutionCEImpl.java
```diff
@@ -74,6 +78,11 @@ public Mono<TriggerResultDTO> trigger(
         Mono<PluginExecutor> pluginExecutorMono =
                 pluginMono.flatMap(plugin -> pluginExecutorHelper.getPluginExecutor(Mono.just(plugin)));
 
+        // Flags are needed here for google sheets integration to support shared drive behind a flag
+        // Once thoroughly tested, this flag can be removed
```
abhvsn, line 82: Let's add the `// TODO message...` to quickly find these tech-debts in future.

abhvsn (commented):
Is it possible to add the tests specifically for the feature flag scenario?

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/FileListMethod.java
sneha122: CachedFeatures has following structure where features are stored as Map<String, Boolean>

```
@Data
@NoArgsConstructor
@AllArgsConstructor
public class CachedFeatures implements Serializable {
    Map<String, Boolean> features;
    Instant refreshedAt;
}
```

Hence I did toString, maybe I can change this to .name() on enum instead of toString() as done in this code:
`featureMap -> featureMap.getOrDefault(featureEnum.name(), Boolean.FALSE)` will update this.

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/TriggerMethod.java
sneha122: This is being overridden by FileInfoMethod, GetStructureMethod and RowsGetMethod,hence kept it as is. Only for FileListMethod we need it with flags hence created separate method with flags and kept the getTriggerClient as is. Hope that makes sense

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/plugins/GoogleSheetsPlugin.java
sneha122: This is being used in DatasourceTriggerSolution test cases and I feel instead of saying unsupported method, we can fall back to the default not supported version of drive, hence added this method here, let me know if that sounds good

app/server/appsmith-server/src/main/java/com/appsmith/server/plugins/solutions/PluginTriggerSolutionCEImpl.java
sneha122: Will do that

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/config/FileListMethod.java
sneha122: Done, can you please check

sneha122:
> Is it possible to add the tests specifically for the feature flag scenario?

Can I add test cases for this in a separate PR? As this is a release blocker I would like to get this merged ASAP, WDYT?

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/plugins/GoogleSheetsPlugin.java
abhvsn: We should generally rely on fail early, which provides you the correct code path where the issue is occurring. I was trying to understand if we are in any situation going to hit this method in GSheets plugin, ideally if not we may have to trace manually to get to the correct trace instead of relying on the stack-trace we get by default from throwing exception.

abhvsn (approved):

app/server/appsmith-plugins/googleSheetsPlugin/src/main/java/com/external/plugins/GoogleSheetsPlugin.java
sneha122: Makes sense, this is not explicitly being called for gsheetplugin, it's only in the test cases and to handle test cases I have made this fallback to default

PR #37858, opened 2024-11-29
ci: Only allow appsmith.com emails to signup on DPs

sharat87:
If we need someone without such an email to signup, let someone _with_ such an email signup as superuser first, then they can add more admin users in Admin Settings.

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **New Features**
	- Enhanced deployment configuration for the Appsmith application, allowing signup only from specified domains (`appsmith.com`).
  
- **Chores**
	- Maintained existing AWS and Kubernetes configurations, ensuring seamless deployment operations.


mohanarpit:
@sharat87 What's the rationale behind this change?

mohanarpit:
@sharat87 Can you help me understand the rationale behind this change?

sharat87:
Hey, @mohanarpit, sorry, seems I've missed this twice! The rationale is that DPs are an internal detail, and internal environments. With the POV of secure by default, and considering the expectation for such internal DPs is for the use of internal engineers, I'm setting this up. In fact, I wanted to setup email validation also to be enabled by default, but that felt like too much just right now.

What say?

PR #38141, opened 2024-12-13
chore: fix select widget bug

jsartisan:


app/client/packages/design-system/widgets/src/components/ComboBox/src/styles.module.css
```diff
@@ -0,0 +1,11 @@
+div.comboboxInputGroup {
```
jsartisan, line 1: All these styles are needed so that the popover takes full width.

app/client/packages/design-system/widgets/src/components/ComboBox/src/ComboBox.tsx
```diff
@@ -51,7 +54,22 @@ export const ComboBox = (props: ComboBoxProps) => {
         size={size}
       />
       <FieldError>{errorMessage}</FieldError>
-      <Popover UNSTABLE_portalContainer={root}>
+      <Popover
+        UNSTABLE_portalContainer={root}
+        className={styles.comboboxPopover}
+        containerPadding={0}
+        crossOffset={
```
jsartisan, line 61: This is done so that combobox popover is properly positioned

app/client/packages/design-system/widgets/src/components/ComboBox/src/styles.module.css
```diff
@@ -0,0 +1,3 @@
+.comboboxPopover {
+  width: calc(var(--trigger-width) + var(--inner-spacing-2) * 2);
```
jsartisan, line 2: Since we change how the component is styled, these styles are need so that popover takes full width.

app/client/packages/design-system/widgets/src/components/Input/src/styles.module.css
```diff
@@ -194,31 +164,17 @@
...
+.inputGroup:has(> .input[data-size="large"]) {
   block-size: calc(
     var(--body-line-height) + var(--body-margin-start) + var(--body-margin-end)
   );
   padding-block: var(--inner-spacing-3);
   padding-inline: var(--inner-spacing-3);
 }
-
-/**
- * ----------------------------------------------------------------------------
```
jsartisan, line 213: moved these styles to select folder.

znamenskii-ilia (approved):

app/client/packages/design-system/widgets/src/components/ComboBox/src/ComboBox.tsx
KelvinOm: I believe we we can replace calculations for `crossOffset` and `offset` just by adding CSS. Like this `margin-inline-start: calc(-1* var(--inner-spacing-2));`. I think this way the code will be somewhat cleaner.

app/client/packages/design-system/widgets/src/components/Input/src/styles.module.css
```diff
@@ -57,59 +56,27 @@
...
  */
-.inputGroup[data-hovered]
-  .input:not(
-    :is(
-        [data-focused],
-        [data-readonly],
-        [data-disabled],
-        [data-focus-within],
-        :has(~ input[data-disabled="true"])
+.inputGroup[data-hovered]:has(
```
KelvinOm, line 71: Why is this change necessary?

app/client/packages/design-system/widgets/src/components/Menu/src/Menu.tsx
```diff
@@ -18,8 +19,11 @@ export const Menu = (props: MenuProps) => {
   return (
     <MenuNestingContext.Provider value={nestingLevel + 1}>
       {/* Only the parent Popover should be placed in the root. Placing child popoves in root would cause the menu to function incorrectly */}
-      <Popover UNSTABLE_portalContainer={isRootMenu ? root : undefined}>
-        <HeadlessMenu className={listStyles.listBox} {...props}>
+      <Popover
+        UNSTABLE_portalContainer={isRootMenu ? root : undefined}
+        maxHeight={maxHeight}
```
KelvinOm, line 24: Why is this implemented only for `menu`? Don't we need this for other lists? It may also be worth considering a default constant for this. What do you think?

app/client/packages/design-system/widgets/src/components/Input/src/styles.module.css
jsartisan: We now style the container based on the state of the input. Previously we were styling the input only. It's the same change we discussed on the call.

app/client/packages/design-system/widgets/src/components/ComboBox/src/ComboBox.tsx
jsartisan: Updated.

jsartisan:
@KelvinOm added on useRootContainer hook as well and used it in select, menu, combobox.

app/client/packages/design-system/widgets/src/components/Menu/src/Menu.tsx
jsartisan: Added POPOVER_LIST_BOX_MAX_HEIGHT and used it in menu, select and combobox.

ichik (approved):
Needs a follow-up to address the fact that Windows doesn't provide flag emojis.

KelvinOm (approved):

PR #38345, opened 2024-12-24
feat: custom eslint rule infra

ayushpahwa:
## Description
This PR introduces the custom ESLint rule to [handle floating promises](https://typescript-eslint.io/rules/no-floating-promises/) in our JS Objects (using async function calls without await inside async functions). Along with this, some refactors were needed

1) Function to dynamically enable/disable the rules based on context ( using editor type as the context for now, can be updated as per requirements).
2) Generate a list of async functions from the global data input for evaluation.
3) Package added for estree types.

Note: This custom rule is only used in our EE codebase for now. We can enable it in CE in case the requirement arises.

Fixes https://github.com/appsmithorg/appsmith/issues/37255

## Automation

/test sanity js

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/12558146123>
> Commit: 36b66109c7ef70dc1eb97b445661ac69f881775a
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=12558146123&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Sanity, @tag.JS`
> Spec:
> <hr>Tue, 31 Dec 2024 10:54:23 UTC

## Communication
- [ ] Yes
- [x] No

## Release Notes

- **New Features**
	- Added a custom ESLint rule to enforce proper Promise handling
	- Enhanced linting capabilities with more granular error detection

- **Improvements**
	- Updated linting utility functions to support more flexible configuration
	- Introduced more detailed context-aware linting options

- **Development**
	- Added TypeScript type definitions for ESTree
	- Expanded test coverage for linting utilities

- **Chores**
	- Refactored linting-related utility functions and constants


app/client/src/plugins/Linting/customRules/no-floating-promises.ts
```diff
@@ -0,0 +1,138 @@
+import type { Rule } from "eslint";
+import type * as ESTree from "estree";
+
+export const noFloatingPromisesLintRule: Rule.RuleModule = {
```
KelvinOm, line 4: We have a [package for custom rules](https://github.com/appsmithorg/appsmith/tree/release/app/client/packages/eslint-plugin). However, I'm not entirely sure that these rules are necessary at all. Why can't we just use [this plugin](https://github.com/eslint-community/eslint-plugin-promise)?

ayushpahwa: Hey @KelvinOm thanks for pointing out the eslint plugin, I'll try them out. As for the package for custom rules, these are for the developers building the TS code of appsmith. The eslint rule that I added is for the appsmith users who write JS code inside the JSObjects of apps, packages and workflows. If we mix them, there might be confusion related to how the rules are being applied. However we do need to communicate this properly, would updates to readme help here?

ayushpahwa: I tried to see if the plugins can work. The issue seems to be in the way we lint our code in appsmith. Each function is linted one at a time with limited scope of the rest of the world. Also most of the async functions like Query.run etc are not part of the defined codebase that gets linted, it needs to be passed down as global context. Hence these out of the box plugins don't fit in this particular case.

KelvinOm: Thanks! LGTM!

KelvinOm (approved):

PR #38443, opened 2025-01-01
chore: flexible return for test payload extraction func

### LINKED ISSUE #37742: [Task] Change default payload of test run

ayushpahwa:
## Description
The `getTestPayloadFromCollectionData` extracts the `testPayload` from passed `collectionData` and returns an empty string if no information is present. This PR updates the function definition to allow different default value based on the function call.

Fixes #37742 

## Automation

/test sanity

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/12572189037>
> Commit: 26ea6e2b17c9a31e836ced4e3787a0723db63680
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=12572189037&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Sanity`
> Spec:
> <hr>Wed, 01 Jan 2025 16:34:44 UTC

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced function handling for undefined collection data
  - Improved key iteration in action execution utilities

- **Documentation**
  - Added JSDoc comment to clarify function behavior

### DIFF
--- a/app/client/src/ce/utils/actionExecutionUtils.ts
+++ b/app/client/src/ce/utils/actionExecutionUtils.ts
@@ -100,8 +100,9 @@ export function isBrowserExecutionAllowed(..._args: any[]) {
 
 /**
  * Function to extract the test payload from the collection data
- * @param collectionData from the js Object
- * @param defaultValue to be returned if no information is found
+ * @param [collectionData] from the js Object
+ * @param [defaultValue=""] to be returned if no information is found,
+ * returns an empty string by default
  * @returns stored value from the collectionData
  * */
 export const getTestPayloadFromCollectionData = (
--- a/app/client/src/ce/utils/actionExecutionUtils.ts
+++ b/app/client/src/ce/utils/actionExecutionUtils.ts
@@ -9,6 +9,7 @@ import { getCurrentEnvironmentDetails } from "ee/selectors/environmentSelectors"
 import type { Plugin } from "api/PluginApi";
 import { get, isNil } from "lodash";
 import type { JSCollectionData } from "ee/reducers/entityReducers/jsActionsReducer";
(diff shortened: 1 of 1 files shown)


app/client/src/ce/utils/actionExecutionUtils.ts
```diff
@@ -98,17 +99,24 @@ export function isBrowserExecutionAllowed(..._args: any[]) {
...
+/**
+ * Function to extract the test payload from the collection data
+ * @param [collectionData] from the js Object
+ * @param [defaultValue=""] to be returned if no information is found,
+ * returns an empty string by default
+ * @returns stored value from the collectionData
+ * */
 export const getTestPayloadFromCollectionData = (
   collectionData: JSCollectionData | undefined,
+  defaultValue = "",
```
ashit-rath, line 111: I didn't quite understand the reasoning behind adding this. Basically we are passing a default value to a function which it will return when the collectionData is not there or something else is missing. Can't this function return `null` and the caller of this function can just do `getTestPayloadFromCollectionData() || defaultValue`?

ayushpahwa: The main use is in the EE side of this change. Empty string is treated as a [falsy value](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) and for my use case I want the default value to be used only if the item is undefined. While I could do that check during consumption of this functions, but since the check is already happening here I preferred to update the function definition.

ashit-rath (approved):

PR #38572, opened 2025-01-10
fix: JSObject diff error when actions are missing

ashit-rath:
## Description
This PR fixes 2 issues
<em>Issue 1<em> - No function available in the dropdown of a valid JS Module in JS Module editor
Due to some reason the actions property in the JSObject is missing instead of being `[]`, due to this  when the check is done for getting the diff `const preExisted = jsAction.actions.find((js) => js.name === action.name);` the `.find` is on an `undefined` value. The actual reason of the `actions` to be missing is yet to be discovered but this is added to add guard against such failures.

<em>Issue 2<em> - When a function is removed from a JSModule, a silent error is thrown by the splice
`TypeError: Cannot delete property '0' of [object Array]`
The reason for this is that the jsObject are directly being mutated for an object where the `writable` property is `false`.
Upon verifying the code block removed from line 145-155 has no use since it just updates the jsActions object which neither is used by subsequent code of the function nor the caller of this function.
Also splicing directly on redux selector data mutates the redux store directly which is a bad practice.
So technically it's a piece of code that is unused, thus removing it fixes the problem

Fixes https://github.com/appsmithorg/appsmith/issues/38683

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/12802464049>
> Commit: 316f0aecfe19b1d1db80baf21fa16312ae5f883f
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=12802464049&attempt=2" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Thu, 16 Jan 2025 11:54:27 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling in JS collection updates by ensuring proper array initialization
	- Enhanced action comparison logic to handle undefined action arrays more gracefully

- **Refactor**
	- Simplified action comparison and deletion process in JS collection utilities


vsvamsi1 (approved):

app/client/src/utils/JSPaneUtils.tsx
```diff
@@ -142,17 +144,6 @@ export const getDifferenceInJSCollection = (
     }
   }
 
-  if (toBearchivedActions.length > 0) {
-    for (let i = 0; i < toBearchivedActions.length; i++) {
-      const action = toBearchivedActions[i];
-      const deleteArchived = jsAction.actions.findIndex((js) => {
```
ApekshaBhosale, line 148: don't remove completely but here this is needed jsAction.actions.splice(deleteArchived, 1); @ashit-rath

ApekshaBhosale (changes requested):

app/client/src/utils/JSPaneUtils.tsx
ApekshaBhosale: Understood! no problem with this.

PR #38582, opened 2025-01-10
fix: Separate Plugin Entity from Action

hetunandu:
## Description

Entities/Actions had type definitions of Plugins as well. There were also some Plugin definitions in the PluginApi File. 
I have separated them out into a single place. This reduces a number of cyclical dependencies

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/12772696355>
> Commit: 8b261ed2d5eaf373e70d6592f949547d2adfb982
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=12772696355&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Datasource`
> Spec:
> <hr>Tue, 14 Jan 2025 17:58:36 UTC

## Communication
- [ ] Yes
- [ ] No


app/client/src/IDE/Components/Sidebar/Sidebar.tsx
```diff
@@ -19,15 +18,15 @@ const Container = styled(Flex)`
 // the button should be selected and only handle calling the onClick
 export interface IDESidebarButton
   extends Omit<SidebarButtonProps, "onClick" | "selected"> {
-  state: EditorState;
+  state: string;
```
dvj1988, line 21: @hetunandu Can you explain this change? Looks like we are replacing an enum type with string.

hetunandu: This enum was causing a bit of a cycle. It was also restricting this component to only work with this enum hence making it tightly coupled. I am removing the restriction yet keeping the functionality the same. Later on, we will move this component into ADS to ensure it is separated from business logic

hetunandu: Removed from this PR

dvj1988 (approved):
LGTM 🚀

PR #38794, opened 2025-01-22
fix: when value of array is empty, no need for data type recalculation

### LINKED ISSUE #37246: [Bug]: Lose binding to a script or api with no errors

### Description

I have a dropdown that I bound to a script, when I set them up I see a message that it will load on page start up, but then after a few uses I see this message and then it stops working and I see only the green red blue options: 

image

image

image

It's very annoying and it makes developing on appsmith very difficult. This is not the first time this happens too and I had to try so many things to bypass this problem last time, it shouldn't be this difficult and bindings shouldn't spontaneously be lost. 

Please help fix this bug. 

About the version of appsmith, where can I see what version is running? 

### Steps To Reproduce

add a script binding to a drop down to bring it data
see a message it will load on page startup
use it several times
see a message it will not load on page start up anymore and default data is shown in the dropdown from now on

### Public Sample App

_No response_

### Environment

Production

### Severity

Low (Cosmetic UI issues)

### Issue video log

_No response_

### Version

Don't know.

rahulbarwal:
## Description
<ins>Problem</ins>
When we have a `select/multi-select` field type and the source data gives empty array to the `JSONFormWidget`, the widget tries to gauge the sub type for value inside the array, and since it is empty it got `undefined`

This led to re-evaluation of property config for the field. 

<ins>Solution</ins>
- Added a check to prevent unnecessary recalculation of sub data types when arrays are empty in `checkIfArrayAndSubDataTypeChanged`.

- Changed parameter type from `any` to `unknown` in `dataTypeFor` and `subDataTypeFor` functions for improved type safety.

- Added and refactored unit tests.

This refactor enhances type safety and optimizes performance in the schema parsing logic.

Fixes #37246  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/12905055896>
> Commit: cad7015881c16a9af273b84dcf44cc33e32fb7d9
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=12905055896&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.JSONForm`
> Spec:
> <hr>Wed, 22 Jan 2025 09:50:23 UTC

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **Tests**
	- Enhanced test suite for determining field types.
	- Added comprehensive test cases covering primitive values, email formats, date formats, array types, object types, and edge cases.

- **Refactor**
	- Updated type annotations in schema parser functions to improve type safety.
	- Changed parameter types from `any` to `unknown`.
	- Added clarifying comment for handling empty arrays in type checking.

### DIFF
--- a/app/client/src/widgets/JSONFormWidget/schemaParser.ts
+++ b/app/client/src/widgets/JSONFormWidget/schemaParser.ts
@@ -183,9 +183,7 @@ export const getSourceDataPathFromSchemaItemPath = (
   return sourceDataPath;
 };
 
-// TODO: Fix this the next time the file is edited
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export const dataTypeFor = (value: any) => {
+export const dataTypeFor = (value: unknown) => {
   const typeOfValue = typeof value;
 
   if (Array.isArray(value)) return DataType.ARRAY;
@@ -195,13 +193,11 @@ export const dataTypeFor = (value: any) => {
   return typeOfValue as DataType;
 };
 
-// TODO: Fix this the next time the file is edited
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export const subDataTypeFor = (value: any) => {
+export const subDataTypeFor = (value: unknown) => {
   const dataType = dataTypeFor(value);
 
   if (dataType === DataType.ARRAY) {
-    return dataTypeFor(value[0]);
+    return dataTypeFor((value as unknown[])[0]);
   }
 
   return undefined;
@@ -351,6 +347,13 @@ export const checkIfArrayAndSubDataTypeChanged = (
 
   const currSubDataType = subDataTypeFor(currentData);
   const prevSubDataType = subDataTypeFor(prevData);
+
+  /**
+   * If the array is empty, then we don't need to check for sub data type
+   * as it would be an empty array, which will always be `undefined`.
+   * which leads to unnecessary re calculation of a type(https://github.com/appsmithorg/appsmith/issues/37246)
+   */
+  if (currentData.length === 0 || prevData.length === 0) return false;
 
   return currSubDataType !== prevSubDataType;
 };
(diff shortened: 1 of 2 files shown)


jacquesikot (approved):

app/client/src/widgets/JSONFormWidget/schemaParser.ts
```diff
@@ -352,6 +348,13 @@ export const checkIfArrayAndSubDataTypeChanged = (
   const currSubDataType = subDataTypeFor(currentData);
   const prevSubDataType = subDataTypeFor(prevData);
 
+  /**
+   * If the array is empty, then we don't need to check for sub data type
+   * as it would be an empty array, which will always be `undefined`.
+   * which leads to unnecessary re calculation of a type(https://github.com/appsmithorg/appsmith/issues/37246)
+   */
+  if (currentData.length === 0 || prevData.length === 0) return false;
```
ashit-rath, line 356: Do you think we should move this condition above the subDataTypeFor and exit early?

rahulbarwal: @ashit-rath In the scenario when it does not work, the source data for JSON from looks like this:
Screenshot 2025-01-23 at 10 27 24 AM

* Here if I select row at index 0, then the value it receives in this line is []. 
* From the previous evaluation check JSONForm knows that value type is `Multiselect` and subtype is `string`
* For the new row, value is `Multiselect` but subtype is `undefined`
* This needs to revaluation of the type for the JSON form(which subsequently leads to resetting of data in property pane)

What we want to prevent is the check diff between prev and current when one of them turns out to be empty. I  don't think that making `subDataTypeFor` exit early will help us in ☝️ .

Issue #37648, opened 2024-11-22
Fix focus retention of tabs while switching between IDEs


ankitakinger:
When you switch between IDEs, the old IDE loses the focus retention of tabs. For example, you're on app editor and on page 2, with 6 queries/apis open (tabs). Now, you switch to home screen and open a package editor, and then back to the app editor, the opened page as well as its open tabs are no longer retained. 

This also happens when you simply move from editor to homepage and back to editor. Another scenario this happens is on page refresh. And yet another one, if you try to connect and authorize a Google spreadsheet as a datasource, once you are back on the app editor after authorisation is complete, the opened tabs are no longer retained (open bug [#31590](https://github.com/appsmithorg/appsmith/issues/31590))

This needs to be fixed by maybe using local storage or you can explore other solutions.

Also, handle the state that maintains which tabs are open so that when user switches between say data and editor tabs in the sidebar, the open entities should remain open. The focus retention across IDEs will be handled in another task.

hetunandu:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/IDE-Pod-60507ad1d4b98d00150a2858/issues/appsmithorg/appsmith/37648?planning-poker) with Zenhub @alex-golovanov @ankitakinger

PR #38798, opened 2025-01-22
fix: App card routing

hetunandu:
## Description

fixes an issue where App card click causing the whole app to remount and init all redux state

The performance of opening an app in edit has improved drastically post this change

https://github.com/user-attachments/assets/e6e40044-b9c3-4fda-8454-53cba95c4a71

Reverts code back to https://github.com/appsmithorg/appsmith/pull/15168
Changed with https://github.com/appsmithorg/appsmith/pull/27223/files#diff-d346c04dfdbc5fc695a1c5e7308754cd2c676a6ea31e1783b28792d7c3895dd3R461

Precursor for #37648

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/13003965904>
> Commit: fdb8faf000454d1c1e0fbf637981d95f22d32734
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=13003965904&attempt=3" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Tue, 28 Jan 2025 06:43:19 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **Refactor**
	- Updated event handling for application launch and edit functions.
	- Simplified prop access within the application card component.
	- Enhanced application selection and editing functionality in the HomePage class.
	- Modified navigation mechanism in the AppsmithLink component.

- **Bug Fixes**
	- Improved event handling to prevent default actions and support opening links in new tabs.
	- Refined visibility checks for UI components after application creation and editing.

- **New Features**
	- Added a method to select an application for editing in the HomePage class.
	- Introduced a validation step to check the visibility of the application name after launching.


app/client/src/pages/Applications/ApplicationCard.tsx
```diff
@@ -477,26 +477,38 @@ export function ApplicationCard(props: ApplicationCardProps) {
...
+  const launchMobileApp = useCallback(
+    (e) => {
+      e.preventDefault();
+      e.stopPropagation();
+      setURLParams();
+      history.push(viewModeURL);
+      dispatch(getCurrentUser());
+    },
+    [dispatch, props.application.defaultBasePageId, setURLParams, viewerParams],
+  );
```
ankitakinger, line 511: @hetunandu this is removing the functionality to open app in new tab which works on release.

ankitakinger (approved):

ankitakinger (dismissed):

app/client/src/pages/Applications/ApplicationCard.tsx
vsvamsi1: Hi @hetunandu can you point me out the action which is causing the remount. You are pushing the a new route so the remount is inevitable?

dvj1988:
@hetunandu Is it correct to assume that the entire page was re-rendering because we were passing the href to the "Edit" and "Launch" CTAs on the Application card?
If we are using React Router to navigate to the application's edit and view modes, do we still need to pass the href to these CTAs? From the code, it appears that we are passing the href but then preventing the default behavior, which would otherwise navigate to the new URL.

hetunandu:
@dvj1988 is correct here @vsvamsi1 

href causes the React Router / Browser to load the App.tsx to mount again instead just the sub route. 

I am preventing default unless the user attempts to open the App in a new tab / window cause that anyways needs a full App mount.

there still is a href passed which makes it a link. I think this helps in browser understanding it is a link but if needed we can remove that as well.

ankitakinger (approved):

PR #38891, opened 2025-01-29
chore: Migrate Tenant to Organization

trishaanand:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No


app/server/appsmith-server/src/main/java/com/appsmith/server/acl/AclPermission.java
```diff
@@ -121,7 +121,7 @@ public enum AclPermission {
     READ_PERMISSION_GROUPS("read:permissionGroups", PermissionGroup.class),
 
     // Manage tenant permissions
-    MANAGE_TENANT("manage:tenants", Tenant.class),
+    MANAGE_TENANT("manage:tenants", Organization.class),
```
abhvsn, line 124: @trishaanand curious are we going to create another PR to move the existing permission nomenclature with the corresponding migration changes required?

trishaanand: Oh it would be part of this PR itself. Its just a WIP.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration065_CopyTenantIdToOrganizationId.java
```diff
@@ -0,0 +1,242 @@
...
+                new Query(where("tenantId").exists(true).and("organizationId").exists(false));
+
+        // Create an update pipeline that copies the value
+        List<Document> pipeline = Arrays.asList(new Document("$set", new Document("organizationId", "$tenantId")));
+
+        try {
+            long updatedCount = mongoTemplate
+                    .getCollection(mongoTemplate.getCollectionName(domainClass))
+                    .updateMany(query.getQueryObject(), pipeline)
+                    .getModifiedCount();
```
abhvsn, line 164: Do you think it will be a good idea to batch the updates, to reduce memory usage?

trishaanand: @abhvsn : From what I understand, updateMany underlying implementation uses batching (mongo db level) and is the most efficient way to update documents. Do you have suggestions for improvement?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration065_CopyTenantIdToOrganizationId.java
```diff
@@ -0,0 +1,242 @@
...
+                            JsonNode jsonNode = objectMapper.readTree(jsonStr);
+                            ObjectNode objectNode = (ObjectNode) jsonNode;
+
+                            if (objectNode.has("tenantId") && !objectNode.has("organizationId")) {
+                                // Copy tenantId value to organizationId
+                                objectNode.put(
+                                        "organizationId",
+                                        objectNode.get("tenantId").asText());
+                                // Remove tenantId
+                                objectNode.remove("tenantId");
```
abhvsn, line 218: Will there be any impact if older pods remove the cache because of the de-serialisation exception while migration is going on?

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration065_CopyTenantIdToOrganizationId.java
abhvsn: I think that's `bulkWrite` not `updateMany`, can you share if there is any reference. Otherwise for batched process I was thinking breaking the query in 2 DB calls:
1. Find (batching will be done for this)
2. updateMany 

This will increase the time considerable as we are hitting the DB couple of time only if the collection is bigger, which is what we are trying to optimise the memory footprint here.

abhvsn (approved):
Added minor comment for migrations, otherwise LGTM.

trishaanand:
Cypress tests started failing just now. They are passing on EE. Unrelated failures.

PR #38940, opened 2025-01-31
fix: connection pool config added for dynamoDB to avoid stale connections

sneha122:
## Description
This PR configures connection time to live property for DynamoDB plugin. This is configured to be 1 day which means after 1 day, active/idle connections will be closed. This is done to ensure that we don't face issues if the connection becomes stale.  

Fixes #[`Issue Number`](https://github.com/appsmithorg/appsmith-ee/issues/6030)  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags="@tag.All"

> [!CAUTION]  

## Communication
- [ ] Yes
- [x] No

### DIFF
--- a/app/server/appsmith-plugins/dynamoPlugin/src/main/java/com/external/plugins/DynamoPlugin.java
+++ b/app/server/appsmith-plugins/dynamoPlugin/src/main/java/com/external/plugins/DynamoPlugin.java
@@ -29,6 +29,7 @@ import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
 import software.amazon.awssdk.core.SdkBytes;
 import software.amazon.awssdk.core.SdkField;
 import software.amazon.awssdk.core.SdkPojo;
+import software.amazon.awssdk.http.apache.ApacheHttpClient;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
 import software.amazon.awssdk.services.dynamodb.DynamoDbClientBuilder;
@@ -42,6 +43,7 @@ import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.lang.reflect.WildcardType;
 import java.net.URI;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -70,6 +72,7 @@ public class DynamoPlugin extends BasePlugin {
     private static final String DYNAMO_TYPE_BINARY_SET_LABEL = "BS";
     private static final String DYNAMO_TYPE_MAP_LABEL = "M";
     private static final String DYNAMO_TYPE_LIST_LABEL = "L";
+    private static final Duration CONNECTION_TIME_TO_LIVE = Duration.ofDays(1);
 
     public DynamoPlugin(PluginWrapper wrapper) {
         super(wrapper);
@@ -285,7 +288,15 @@ public class DynamoPlugin extends BasePlugin {
             log.debug(Thread.currentThread().getName() + ": datasourceCreate() called for Dynamo plugin.");
             return Mono.fromCallable(() -> {
                         log.debug(Thread.currentThread().getName() + ": creating dynamodbclient from DynamoDB plugin.");
-                        final DynamoDbClientBuilder builder = DynamoDbClient.builder();
+
+                        // Configuring connection time to live as 1 day so that we don't face issues with stale
+                        // connections
+                        // in the connection pool.
+                        // 1 Day is added randomly as a time which is not too less or too high
+                        final DynamoDbClientBuilder builder = DynamoDbClient.builder()
+                                .httpClient(ApacheHttpClient.builder()
+                                        .connectionTimeToLive(CONNECTION_TIME_TO_LIVE)
+                                        .build());
 
                         if (!CollectionUtils.isEmpty(datasourceConfiguration.getEndpoints())) {
                             final Endpoint endpoint =
--- a/app/server/appsmith-plugins/dynamoPlugin/pom.xml
+++ b/app/server/appsmith-plugins/dynamoPlugin/pom.xml
@@ -32,6 +32,23 @@
                 </exclusion>
             </exclusions>
         </dependency>
+
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>apache-client</artifactId>
+            <version>2.15.3</version>
+            <scope>compile</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>slf4j-api</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>com.fasterxml.jackson.core</groupId>
+                    <artifactId>jackson-databind</artifactId>
+                </exclusion>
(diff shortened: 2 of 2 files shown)


app/server/appsmith-plugins/dynamoPlugin/src/main/java/com/external/plugins/DynamoPlugin.java
```diff
@@ -285,7 +288,15 @@ public Mono<DynamoDbClient> datasourceCreate(DatasourceConfiguration datasourceC
...
                         log.debug(Thread.currentThread().getName() + ": creating dynamodbclient from DynamoDB plugin.");
-                        final DynamoDbClientBuilder builder = DynamoDbClient.builder();
+
+                        // Configuring connection time to live as 1 day so that we don't face issues with stale
+                        // connections
+                        // in the connection pool.
+                        // 1 Day is added randomly as a time which is not too less or too high
+                        final DynamoDbClientBuilder builder = DynamoDbClient.builder()
+                                .httpClient(ApacheHttpClient.builder()
+                                        .connectionTimeToLive(CONNECTION_TIME_TO_LIVE)
```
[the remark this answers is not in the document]
sneha122: @NilanshBansal It won't have any impact, it will be similar to how connections in connection pool works, this connection will have a max life time of 1 day and post that it will be closed if unused and new one will be created. This is done to avoid stale connections. We can create a DP and test it out there

app/server/appsmith-plugins/dynamoPlugin/src/main/java/com/external/plugins/DynamoPlugin.java
```diff
@@ -280,12 +284,71 @@ public Mono<ActionExecutionResult> execute(
...
+                    .subscribeOn(scheduler);
+        }
+
         @Override
         public Mono<DynamoDbClient> datasourceCreate(DatasourceConfiguration datasourceConfiguration) {
             log.debug(Thread.currentThread().getName() + ": datasourceCreate() called for Dynamo plugin.");
             return Mono.fromCallable(() -> {
                         log.debug(Thread.currentThread().getName() + ": creating dynamodbclient from DynamoDB plugin.");
-                        final DynamoDbClientBuilder builder = DynamoDbClient.builder();
+
```
NilanshBansal, line 343: @sneha122 do we need any code changes in this method as the one with the featureflagmap is already overridden now?

sneha122: No we don't need any changes here, let me remove that

app/server/appsmith-plugins/dynamoPlugin/src/main/java/com/external/plugins/DynamoPlugin.java
```diff
@@ -280,12 +284,71 @@ public Mono<ActionExecutionResult> execute(
...
+                         * {@link https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/http/apache/ApacheHttpClient.Builder.html#connectionTimeToLive(java.time.Duration)}
+                         */
+                        Boolean isDynamoConnectionToLiveEnabled = featureFlagMap.getOrDefault(
+                                FeatureFlagEnum.release_dynamodb_connection_time_to_live_enabled.name(), Boolean.FALSE);
+                        DynamoDbClientBuilder builder;
+                        if (isDynamoConnectionToLiveEnabled) {
+                            log.debug("DynamoDB client builder created with custom connection time to live");
+                            builder = DynamoDbClient.builder()
+                                    .httpClient(ApacheHttpClient.builder()
+                                            .connectionTimeToLive(CONNECTION_TIME_TO_LIVE)
```
NilanshBansal, line 308: @sneha122 have we already validated if this TTL property works with our datasource context and is actually closing connections? Also, are the connections getting rebuild when the old connection is closed from DynamoDB? 
If not, can we create a DP and ask the user to test it out first on the DP if it resolves the issue?

app/server/appsmith-plugins/dynamoPlugin/src/main/java/com/external/plugins/DynamoPlugin.java
```diff
@@ -283,6 +284,57 @@ public Mono<ActionExecutionResult> execute(
...
+                                    DynamoErrorMessages.MISSING_REGION_ERROR_MSG);
+                        }
+
+                        builder.region(Region.of(authentication.getDatabaseName()));
+
+                        builder.credentialsProvider(StaticCredentialsProvider.create(AwsBasicCredentials.create(
+                                authentication.getUsername(), authentication.getPassword())));
+
+                        return builder.build();
+                    })
```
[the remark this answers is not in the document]
sneha122: datasourceCreate with just one param will never be called as we have already overridden method with two params, do we still need to move this functionality in that case?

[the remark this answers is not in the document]
NilanshBansal: The code is getting duplicated, so I would recommend to just keep it within one single function. For now, we might not be using the second function, but we don't know when there is a need to update the common code which will make it difficult to update the code at both places.

app/server/appsmith-plugins/dynamoPlugin/src/main/java/com/external/plugins/DynamoPlugin.java
sneha122: What I have validated so far is I had created one app with multiple page running queries and deployed it, queries work fine on day 1 and day 2 as well, this means that even after connection expiration the queries work fine.

Do you have any other ideas on how to validate this? 

Secondly we can ask the user to test out the issue on DP but this issue occurs intermittently and that begs the question of how long we wanna keep this DP open. What Id suggest here is that since we are adding this behind flag anyways, we can ask the user to create the same scenario on our cloud version, that way we won't have to maintain the DP, WDYT?

app/server/appsmith-plugins/dynamoPlugin/src/main/java/com/external/plugins/DynamoPlugin.java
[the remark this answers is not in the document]
sneha122: Makes sense, let me do that

app/server/appsmith-plugins/dynamoPlugin/src/main/java/com/external/plugins/DynamoPlugin.java
NilanshBansal: @sneha122 In case we move this code to prod, the user needs to upgrade their instance or are they already using appsmith cloud? If they need to update their instance to a newer version (which is again some work to be done by them), I would recommend getting this tested on a DP first from them. 
If they are using appsmith cloud, we should be good to go ahead.

NilanshBansal: Right, but why do we want to merge this PR if we have not validated the fix with the user? What additional advantage are we getting from taking this to production? 
You can still take a call on what you feel is correct based on the current situation 🙏

NilanshBansal (approved):
Approved as this is only a temporary change and feature flag will be removed in the next 2 sprints as per discussion with @sneha122

PR #39207, opened 2025-02-12
feat: Introduce infinite scroll feature behind feature flag. 

### LINKED ISSUE #39079: Feature flag to control visibility of this feature(Infinite scrolling).

rahulbarwal:
## Description

- Adding `infiniteScrollEnabled` prop to TableWidgetProps
- Added feature flag: `release_tablev2_infinitescroll_enabled` to control this. 

Fixes #39079 
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags="@tag.Table, @tag.Sanity"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/13305863325>
> Commit: 182c03d37644cc0b7f5563ed5c9e59c4c06667b6
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=13305863325&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Table, @tag.Sanity`
> Spec:
> <hr>Thu, 13 Feb 2025 11:35:05 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
	- Introduced an option to enable infinite scrolling in table views. When activated, tables automatically adjust scrolling behavior and pagination controls, enhancing data browsing.
- **Tests**
	- Updated test cases to verify the new infinite scrolling functionality, ensuring it is disabled by default.


app/client/src/widgets/TableWidgetV2/component/Table.tsx
```diff
@@ -487,7 +489,7 @@ export function Table(props: TableProps) {
                 subPage={subPage}
                 tableSizes={tableSizes}
                 totalColumnsWidth={totalColumnsWidth}
-                useVirtual={shouldUseVirtual}
```
AmanAgarwal041, line 490: Why did we remove this and changed to `false` always ?

app/client/src/widgets/TableWidgetV2/component/StaticTable.tsx
```diff
@@ -81,6 +81,7 @@ const StaticTable = (props: StaticTableProps, ref: React.Ref<SimpleBar>) => {
         getTableBodyProps={props.getTableBodyProps}
         height={props.height}
         isAddRowInProgress={props.isAddRowInProgress}
+        isInfiniteScrollEnabled={false}
```
AmanAgarwal041, line 84: As we have added the feature flag, shouldn't we make use of that here instead of making it `false` always ?

rahulbarwal: Static table is never going to use infinite scroll. So basically the idea behind infinite scroll is that it will always use virtual table. 

Why do we have to pass this when static table is not using this variable? So this is just baggage that we have from earlier structuring of code. I am planning to handle this in my next PR where I will try to do a refactor if possible.

app/client/src/widgets/TableWidgetV2/component/Table.tsx
rahulbarwal: Yeah, this one as well you have to scroll a little above to see why when exactly is this component rendered and you will notice that this should use virtual variable is being used as a ternary operator. 

So basically if this is false, this component will never come into picture. 
https://github.com/appsmithorg/appsmith/pull/39207/files/037e61374e5ee5c1f9ce565a830dc82eedd88d19#diff-08edd82840d934f409a516b170d42e1cb44db77af820eb4c58d0cd6816ab6cb8L458

rahulbarwal: I agree with both of your comments here. It's just that right now both the static table and the virtual table are using the same table body underneath and there is no differentiation of props being passed. This is just something that needs to be cleaned up with a proper refactor which seemed out of scope of this PR.

app/client/src/widgets/TableWidgetV2/component/StaticTable.tsx
AmanAgarwal041: If this prop is a new prop then we can mark this as the optional prop which would be undefined by default which will solve the purpose instead of passing false to this prop explicitly.

rahulbarwal: This is what I am forced to do right now but I don't like it to be honest. Ideally I should not have to write any piece of code that I don't need. 
If instead of hard coding it to false, if I give it a variable that means that the JS compiler has to evaluate that value before passing it down here. Which in this case might be a moot point but still in principle I should not have to write what I don't need. 

What I am proposing is eliminating this unnecessary code in a refactor PR.

rahulbarwal: We met over a call and agreed to make this an optional argument and deal with this whole component structure in a separate PR.

AmanAgarwal041 (approved):

PR #39301, opened 2025-02-16
fix: sourcemaps

KelvinOm:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No


app/client/config/webpack.config.js
```diff
@@ -157,7 +154,9 @@ module.exports = function (webpackEnv) {
                   ],
                 ],
           },
-          sourceMap: isEnvProduction ? true : isEnvDevelopment,
+          sourceMap:
+            process.env.REACT_APP_ENVIRONMENT === "PRODUCTION" ||
```
alex-golovanov, line 158: Please, extract these to a `const`.

Are there more environments than dev and production?

app/client/config/webpack.config.js
```diff
@@ -157,7 +154,9 @@ module.exports = function (webpackEnv) {
                   ],
                 ],
           },
-          sourceMap: isEnvProduction ? true : isEnvDevelopment,
+          sourceMap:
```
alex-golovanov, line 157: Is this really needed, I thought it should automatically resolve through `devtool`.

app/client/config/webpack.config.js
```diff
@@ -188,7 +191,10 @@ module.exports = function (webpackEnv) {
     mode: isEnvProduction ? "production" : "development",
     // Stop compilation early in production
     bail: isEnvProduction,
-    devtool: devtool,
+    devtool:
```
alex-golovanov, line 194: Can't we just define `devtool` above and use it here? These long conditions are super hard to read.

`const devtool = isEnvProduction ? "source-map" : "cheap-module-source-map";`

app/client/config/webpack.config.js
KelvinOm: We have a strange story here. We have `webpackEnv` variables that define the build mode. But before eject we also used `REACT_APP_ENVIRONMENT` to add or override some things. In this PR I have added `REACT_APP_ENVIRONMENT` to make it works as before. Ideally, we need to rethink approach and add a specific variable that will be used only for sourcemaps. To do this, we also need to change the github workflow files, but the purpose of this PR is just to make the sourcemaps work.

app/client/config/webpack.config.js
KelvinOm: Nope. We don't always need sourcemaps. For example, when creating a DP, this condition should be false in order for the build to be created faster.

app/client/config/webpack.config.js
KelvinOm: I'm not sure. I've left everything as it is for now.

alex-golovanov (approved):

PR #39397, opened 2025-02-21
chore: add rollup config for wds

jsartisan:
With this PR, we want to publish WDS as a npm package. To do that, we are now adding a rollup setup to compile and build the package. 

One thing to note:
Rollup needed `7.24.0` version of browserlist and we had to update browserlist version in resolutions in root package.json. Now since we updated browserlist, certain code that was written with old format ( code was coming from blueprint's hotkeys component ) started failing and we had to refactor. It included the code around hotkeys component. It was deprecatdd by blueprint and we were still using. We have now refactored  those part of code to use the `useHotKeys` hook.

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/13538917574>
> Commit: 9b0b51791e4a95574c9729245ba09994ab371b71
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=13538917574&attempt=2" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Wed, 26 Feb 2025 09:12:45 UTC


app/client/packages/design-system/widgets/package.json
```diff
@@ -34,12 +39,31 @@
     "usehooks-ts": "*"
   },
   "devDependencies": {
+    "@babel/core": "^7.23.9",
```
KelvinOm, line 42: We have a conflict with package versions. The build (`yarn start`) does not work with these versions. We have to use the same package versions that are currently used in the app.

Снимок экрана 2025-02-24 в 11 19 23

app/client/package.json
```diff
@@ -423,4 +423,4 @@
     "@types/react": "^17.0.2",
     "postcss": "8.4.31"
   }
-}
+}
```
KelvinOm, line 426: After installing the dependencies, the GIT shows 3 modified files, could you check and fix it?

Снимок экрана 2025-02-24 в 11 21 23

app/client/packages/design-system/widgets/package.json
```diff
@@ -4,12 +4,17 @@
   "main": "src/index.ts",
   "author": "Valera Melnikov <valera@appsmith.com>, Pawan Kumar <pawan@appsmith.com>",
   "license": "MIT",
+  "type": "module",
```
KelvinOm, line 7: We need to check if this affects the current work of the package in the main app.

KelvinOm (changes requested):

app/client/packages/design-system/widgets/package.json
jsartisan: It affected the jest.config.js. I have changed to use es module format instead of cjs.

KelvinOm: So is it not for build or publishing? Can you help me to understand what the issue is? Can you make a screenshot?

app/client/src/components/editorComponents/GlobalSearch/GlobalSearchHotKeys.tsx
```diff
@@ -1,5 +1,5 @@
-import React from "react";
-import { Hotkey, Hotkeys, HotkeysTarget } from "@blueprintjs/core";
+import React, { useMemo } from "react";
+import { useHotkeys, type HotkeyConfig } from "@blueprintjs/core";
```
KelvinOm, line 2: Why we need all this changes?

app/client/packages/design-system/widgets/package.json
jsartisan: We added this "type:module", because the script "rollup.config.js" is using import/export syntax. 

We can avoid adding "type:module", if we use syntax like `require("rollup:)` there in rollup.config.js.

Since we added thee "type:module", the jest.config.js also needs to be written in esm format.

KelvinOm (approved):

PR #39489, opened 2025-02-28
chore: beta tags enabled for certain integrations

### LINKED ISSUE #39486: [Task]: Implement beta tags for the newly launched integrations
### SubTasks

- [x] Add beta tags for the released integrations
- [x] If the integration is not released for a user, then keep showing premium or soon tags as per the earlier implementation

Figma design: https://www.figma.com/design/kbU9xPv44neCfv1FFo9Ndu/User-Activation?node-id=6695-26424&t=FyZRh5SvhOFH1zlL-4

AmanAgarwal041:
## Description
- This PR appends a beta tag for the premium integrations if they are rolled out to the users. This also contains the functionality of disabling beta tags from any of the rolled integrations.
- This PR also solves a bug of duplicate premium integrations (released and upcoming with different letter casing).

Fixes #39486 
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/13638370133>
> Commit: 145758f4eb05b1dc39eb01dcf108c45419a4f058
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=13638370133&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Datasource`
> Spec:
> <hr>Mon, 03 Mar 2025 19:50:43 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
  - Introduced a visual "Beta" label to indicate beta status on certain integrations.
  - Enhanced the interface by conditionally displaying the new Beta tag alongside existing status indicators.
  - Defined a set of notable beta integrations to be marked with the Beta tag.
  - Applied updated styling for the Beta tag to ensure a consistent, clear display across the interface.
  - Added functionality to check if plugins are in a beta state.


app/client/src/pages/Editor/IntegrationEditor/PremiumDatasources/Constants.ts
```diff
@@ -41,3 +41,5 @@ export const PREMIUM_INTEGRATION_CONTACT_FORM =
 
 export const SCHEDULE_CALL_URL =
   "https://calendly.com/carina-neves-fonseca/appsmith";
+
+export const BETA_PLUGINS = ["Zendesk", "Salesforce", "Slack"];
```
sneha122, line 45: @AmanAgarwal041 This will be needed for JIRA as well, can we make this dynamic similar to premium integrations? Otherwise everytime a plugin is added or removed, we would have to make changes to code

AmanAgarwal041: @sneha122 How are we gonna handle that when do we want to remove beta tags from some datasource and keep for others ? Which is want happens here through this variable, either this or we would have to implement it from backend.

AmanAgarwal041: @sneha122 Since we have plugins returned from the server which we show by looping through it and show beta tags if those integrations are enabled for it. Hence, filtering it dynamically through a function would be an overhead. What we can do is have a function which returns whether an integration is enabled for beta tag or not. Does it make sense ?

AmanAgarwal041: @sneha122 I think if we don't want to create any subsequent prs then we would have to implement it using backend, that way when we enable any integrations, we would mark the integration as beta. But in that case, it would be difficult for us to remove beta tag for a specific integration as it will involve another migration to do that. 
Anyhow we would have to address one or the other thing using another pr or script.

sneha122: Got it thanks @AmanAgarwal041 With current approach, we are maintaining list of released integrations on frontend , currently which is empty and so beta tag will be added to all external saas integrations but as and when we decide to release certain integrations, we will have to add those in this list and then raise a Pr

sneha122 (approved):

PR #39517, opened 2025-03-03
fix: map widget marker onClick duplication

### LINKED ISSUE #39514: [Bug]: Map widget is having performance issues with onMarkerClick event

### Description

I am facing performance issues when using a Map widget with configured onMarkerClick.
So I configured the onMarkerClick to showAlert the title of the marker. I was able to notice that whenever I clicked the marker, two alerts were being thrown every time.

Also on using this functionality for a while, the page freezes and I get a popup stating that my App has hanged and I either need to wait or kill the Application.

Facing this issue in both Edit and View mode of the application

https://theappsmith.slack.com/archives/C02RCR2PURZ/p1740717067284009

### Steps To Reproduce

1. Drag and Drop a Map widget
2. Set some default markers in its setting
3. Configure onMarkerClick event to showAlert with the Marker's title
4. Now click on any Marker. You will see two alerts being thrown everytime
5. Keep clicking different markers, and after sometime the screen freezes and App starts to hang

### Public Sample App

_No response_

### Environment

Production

### Severity

Medium (Frustrating UX)

### Issue video log

_No response_

### Version

Cloud - 1.62

jacquesikot:
## Description
When clicking on a marker in the Map Widget, the onMarkerClick event was being triggered twice for a single click, causing duplicate actions and potentially unexpected behavior in applications.

```js
// In MapWidget/widget/index.tsx
onMarkerClick = (lat?: number, long?: number, title?: string) => {
  this.disableDrag(true);
  const selectedMarker = {
    lat: lat,
    long: long,
    title: title,
  };
  console.log("🚀 ~ MapWidget ~ title:", title); // This was logging twice

  this.props.updateWidgetMetaProperty("selectedMarker", selectedMarker, {
    triggerPropertyName: "onMarkerClick",
    dynamicString: this.props.onMarkerClick,
    event: {
      type: EventType.ON_MARKER_CLICK,
    },
  });
};
```

## Root Cause
This happens because:
1. The Marker component was adding duplicate event listeners to the Google Maps marker.
2. Two separate useEffect hooks were both adding "click" event listeners:
   - One during marker initialisation
   - Another in a separate useEffect that tracks the onClick prop

```js
// First listener in initialization useEffect
googleMapMarker.addListener("click", () => {
  if (onClick) onClick();
});

// Second listener in a separate useEffect
marker.addListener("click", () => {
  if (onClick) onClick();
});
```
3. When a marker was clicked, both event listeners fired, causing the onClick callback to execute twice.
4. This resulted in onMarkerClick being called twice for a single user interaction.

## Solution
Modify the `Marker.tsx` component to avoid adding duplicate event listeners:
1. Remove the click event listener from the initialisation useEffect
2. Add proper cleanup for event listeners using clearListeners and removeListener
3. Store the listener reference to properly remove it in the cleanup function
4. Apply the same pattern to the dragend event listener

```js
// track on onclick - with proper cleanup
useEffect(() => {
  if (!marker) return;

  // Clear existing click listeners before adding a new one
  google.maps.event.clearListeners(marker, "click");
  
  // Add the new click listener
  const clickListener = marker.addListener("click", () => {
    if (onClick) onClick();
  });

  // Return cleanup function
  return () => {
    google.maps.event.removeListener(clickListener);
  };
}, [marker, onClick]);
```

## Why this works
1. Properly cleans up existing listeners before adding new ones
2. Ensures only one click handler is active at any time
3. Prevents event handler duplication during component re-renders

## Testing
Added a Cypress test that verifies the marker click event triggers exactly once per click:
1. Creates a map widget with a marker
2. Uses a JS Object to track click count
3. Verifies the click count increments by exactly 1 for each click
4. Ensures no duplicate events are triggered

Fixes #39514 

## Automation

/ok-to-test tags="@tag.Widget, @tag.Sanity, @tag.Maps, @tag.Binding"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/13673511245>
> Commit: 3c025ebd9dd85a3c409bc1582e13e324d8dbb2cd
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=13673511245&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Widget, @tag.Sanity, @tag.Maps, @tag.Binding`
> Spec:
> <hr>Wed, 05 Mar 2025 11:49:24 UTC

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced the behavior of map markers so they respond more reliably to user interactions like clicking and dragging, addressing issues that could affect overall stability.

- **Tests**
  - Added comprehensive tests to ensure marker functionality remains consistent and dependable during use.


app/client/src/widgets/MapWidget/component/Marker.tsx
```diff
@@ -79,19 +75,32 @@ const Marker: React.FC<MarkerProps> = (options) => {
...
+    google.maps.event.clearListeners(marker, "dragend");
+    const dragEndListener = marker.addListener(
+      "dragend",
+      (e: google.maps.MapMouseEvent) => {
+        if (onDragEnd) onDragEnd(e);
+      },
+    );
+
+    return () => {
+      google.maps.event.removeListener(dragEndListener);
```
rahulbarwal, line 101: In this effect we are adding drag and listener to a marker but when the component is dismounted we are removing the listener from the maps object. Is this correct? Shouldn't we be keeping track of all the markers and all the listeners and clearing those listeners as soon as the marker itself gets removed?

app/client/src/widgets/MapWidget/component/Marker.tsx
```diff
@@ -79,19 +75,32 @@ const Marker: React.FC<MarkerProps> = (options) => {
   useEffect(() => {
     if (!marker) return;
 
-    marker.addListener("click", () => {
+    google.maps.event.clearListeners(marker, "click");
+    const clickListener = marker.addListener("click", () => {
       if (onClick) onClick();
     });
+
```
rahulbarwal, line 82: Similar question as above.

rahulbarwal (commented):
It seems we are adding listeners to markers and removing them from Google Maps. Just validate the correctness of it.

rahulbarwal (approved):

app/client/src/widgets/MapWidget/component/Marker.tsx
jacquesikot: Not necessarily. The approach ensures that only a single listener for each event is mounted in a single time, and all listeners are named and cleaned up appropriately.

PR #39757, opened 2025-03-17
chore: remove analytics execution from the critical path

vsvamsi1:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No


app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
```diff
@@ -557,17 +557,20 @@ public Mono<DatasourceContext<?>> getDatasourceContext(DatasourceStorage datasou
...
-            executePluginDTO.setPluginVersion(plugin.getVersion());
-            executePluginDTO.setDatasource(new RemoteDatasourceDTO(
-                    datasourceStorage.getDatasourceId(), datasourceStorage.getDatasourceConfiguration()));
-            datasourceContext.setConnection(executePluginDTO);
-
-            return datasourceContext;
-        });
+        return configService
+                .getInstanceId()
+                .subscribeOn(Schedulers.boundedElastic())
```
ApekshaBhosale, line 562: @vsvamsi1 why did you introduced new thread here?

ApekshaBhosale (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
vsvamsi1: Because this is a `configService.getInstanceId()` a db call, Schedulers.boundedElastic() threads are designed blocking/longer execution tasks.

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImpl.java
```diff
@@ -1097,16 +1109,16 @@ private Mono<ActionExecutionRequest> sendExecuteAnalyticsEvent(
             request.setProperties(stringProperties);
         }
 
-        return Mono.justOrEmpty(actionDTO.getApplicationId())
+        Mono<Application> applicationMono = Mono.justOrEmpty(actionDTO.getApplicationId())
                 .flatMap(applicationService::findById)
```
ApekshaBhosale, line 1113: @vsvamsi1 this is also a db call but we use the same thread so better take @NilanshBansal opinion here

ApekshaBhosale (changes requested):

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImpl.java
NilanshBansal: @ApekshaBhosale the changes look good as applicationMono is only getting used once so we need not to cache it.
cc: @vsvamsi1

NilanshBansal (approved):
LGTM

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImpl.java
NilanshBansal: @ApekshaBhosale logically yes, we can add thread scheduling here also but since we are already using ReactiveCrudRepository, it automatically handles the thread scheduling and it is non blocking. 
The call goes to `repository.findById(id)` which by default handles scheduling

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/DatasourceContextServiceCEImpl.java
NilanshBansal: @vsvamsi1 we can avoid subscribing this on a scheduler as the ReactiveCrudRepository is already non blocking and handling the database calls
cc: @ApekshaBhosale

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImpl.java
vsvamsi1: @NilanshBansal  I have a follow up question, all the monos chained to first db call will they be still operating on the same thread(bounded elastic one)?

NilanshBansal (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImpl.java
NilanshBansal: @vsvamsi1 no. The other monos will be running on the reactive threads. Only the db calls with the reactivecrudrepository automatically gets scheduled

vsvamsi1: Okay makes sense thank you

ApekshaBhosale (approved):

PR #39803, opened 2025-03-19
fix: fix axis label configuration for ECharts with additional properties

### LINKED ISSUE #38974: [Bug]: Y labels for chart are truncated

### Description

When I create a chart with values in certain ranges, the labels of the Y axis are truncated or completely replaced with an ellipse. Here are a few examples:

Image

Image

Image

Weirdly this doesn't affect very large labels, but relatively small ones.

### Steps To Reproduce

- Create a chart component
- Replace the chart data with 
```
[
  {
    "x": "Product1",
    "y": 4
  },
  {
    "x": "Product2",
    "y": 8
  },
  {
    "x": "Product3",
    "y": 12
  }
]
```
The chart labels are truncated.

### Public Sample App

_No response_

### Environment

Production

### Severity

Medium (Frustrating UX)

### Issue video log

_No response_

### Version

Cloud v1.60.0-Snapshot

ALOK9442:
Fixes #38974   
Fixes the issue where Y-axis labels were getting truncated, as mentioned in the issue.

Also resolves the problem of X-axis labels not being truncated when displaying large values. The fix ensures that truncation occurs in two cases:
When the orientation is set to "auto," truncation happens for relatively larger values.
For other orientations, truncation follows the same width constraints as before. 
## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
  - Enhanced X-axis label display for charts, offering improved control over label spacing, width, and overlap when using automatic orientation.
  - Adjusted Y-axis label rendering by removing explicit truncation, resulting in a modified display behavior when labels exceed allocated space.

### DIFF
--- a/app/client/src/widgets/ChartWidget/component/LayoutBuilders/EChartsYAxisLayoutBuilder.ts
+++ b/app/client/src/widgets/ChartWidget/component/LayoutBuilders/EChartsYAxisLayoutBuilder.ts
@@ -45,7 +45,6 @@ export class EChartsYAxisLayoutBuilder {
       nameGap: this.nameGap,
       axisLabel: {
         width: this.labelsWidth,
-        overflow: "truncate",
       },
     };
   };
--- a/app/client/src/widgets/ChartWidget/component/LayoutBuilders/EChartsXAxisLayoutBuilder.ts
+++ b/app/client/src/widgets/ChartWidget/component/LayoutBuilders/EChartsXAxisLayoutBuilder.ts
@@ -32,7 +32,12 @@ export class EChartsXAxisLayoutBuilder {
 
   axisLabelConfig = (allocatedXAxisHeight: number) => {
     if (this.props.labelOrientation == LabelOrientation.AUTO) {
-      return {};
+      return {
+        width: allocatedXAxisHeight,
+        overflow: "truncate",
+        hideOverlap: true,
+        interval: 0,
+      };
     } else {
       return {
         width:
@@ -40,6 +45,8 @@ export class EChartsXAxisLayoutBuilder {
           this.defaultHeightForXAxisName -
           this.gapBetweenLabelAndName,
         overflow: "truncate",
+        hideOverlap: true,
+        interval: 0,
       };
     }
   };


ALOK9442:
Hey @rahulbarwal I have raised the pr, can you please check it. 
Thanks

app/client/src/widgets/ChartWidget/component/LayoutBuilders/EChartsYAxisLayoutBuilder.ts
```diff
@@ -45,7 +45,6 @@ export class EChartsYAxisLayoutBuilder {
       nameGap: this.nameGap,
       axisLabel: {
         width: this.labelsWidth,
-        overflow: "truncate",
```
rajatagrawal, line 48: @ALOK9442 

We should debug why the label width is incorrect. Calculating the correct width for labels and the chart area should fix the bug. Removing truncate will make other use cases fail.

Same approach for x label truncation issue too.

rajatagrawal: Once the fix is done, I would like to request you to test the client unit test cases pass as well since there are a lot of unit test cases for the chart widget which check various scenarios which should pass after the changes. Thank you!

ALOK9442: @rajatagrawal 
thanks for pointing out the issue and for x axis we are having that issue in auto orientation only so we need to add the truncate properties in it and else I am debugging what could be the problem as you are correct about not removing truncate as it will affect the other edge cases

ALOK9442:
Hey @rajatagrawal, since this is assigned to you, I wanted to share a few observations:

1. The width calculation currently considers commas in numbers. For example, in a large number like 120,000,000, the measureText function includes the commas while measuring the width, which may lead to inaccuracies.

2. The font being used appears as 14px undefined. While the font size is correctly set to 14px, the font family is not being applied properly. This happens because it is assigned as fontFamily: "fontfamily", which causes measureText to fall back to the default 10px sans-serif, leading to incorrect width calculations.

although I am still working on it

ALOK9442:
hey @rajatagrawal , I think using a formatter would be a better approach as the width is dynamic and we are calculating it on the basis of text width and that's why even if we truncate on the basis  of this dynamic width we will still face inconsistency for bigger numbers, let me know what do you think about this aproach

PR #40079, opened 2025-04-04
chore: Adding handling for empty environment variable for SSO configuration

trishaanand:
## Description
Enable form login spec is failing after disconnecting github which is leading to the server not starting. This fixes by eliminating any client iwth null or empty value for client id

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced third-party login reliability by automatically applying default values for incomplete OAuth2 settings, ensuring a smoother and more robust authentication experience.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/oauth2/OAuth2PropertiesCustomizer.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/oauth2/OAuth2PropertiesCustomizer.java
@@ -1,0 +1,141 @@
+package com.appsmith.server.authentication.oauth2;
+
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.config.BeanFactoryPostProcessor;
+import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.context.EnvironmentAware;
+import org.springframework.core.Ordered;
+import org.springframework.core.annotation.Order;
+import org.springframework.core.env.ConfigurableEnvironment;
+import org.springframework.core.env.EnumerablePropertySource;
+import org.springframework.core.env.Environment;
+import org.springframework.core.env.MapPropertySource;
+import org.springframework.core.env.MutablePropertySources;
+import org.springframework.core.env.PropertySource;
+import org.springframework.stereotype.Component;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Customizes OAuth2 client properties to handle empty client IDs.
+ * This component runs early in the Spring initialization process and replaces
+ * empty client ID values with "missing_value_sentinel" to trigger default values.
+ *
+ * Handles any OAuth provider registered using the standard Spring Security OAuth2 properties.
+ */
+@Component
+@Order(Ordered.HIGHEST_PRECEDENCE)
+@Slf4j
+public class OAuth2PropertiesCustomizer implements BeanFactoryPostProcessor, EnvironmentAware, ApplicationContextAware {
+
+    private static final String MISSING_VALUE_SENTINEL = "missing_value_sentinel";
+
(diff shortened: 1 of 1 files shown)


trishaanand:
/ci-test-limit-count run_count=1 update_snapshot=true specs_to_run=cypress/e2e/Regression/ClientSide/FormLogin/EnableFormLogin_spec.js

app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/oauth2/OAuth2PropertiesCustomizer.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    @Override
+    public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory) throws BeansException {
+        log.info("OAuth2PropertiesCustomizer: Processing OAuth2 properties early in application startup");
+
+        if (!(environment instanceof ConfigurableEnvironment)) {
+            log.warn("Environment is not a ConfigurableEnvironment, skipping OAuth2 property customization");
+            return;
+        }
```
nsarupr, line 67: Can this lead to any errors, if the properties are not set?

app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/oauth2/OAuth2PropertiesCustomizer.java
```diff
@@ -0,0 +1,141 @@
...
+                    if (name.startsWith(OAUTH_CLIENT_ID_PREFIX) && name.endsWith(OAUTH_CLIENT_ID_SUFFIX)) {
+                        properties.add(name);
+                    }
+                }
+            }
+        }
+
+        log.debug("Found OAuth client ID properties: {}", properties);
+        return properties;
+    }
```
nsarupr, line 124: @this method is just finding all the different clients for oauth, right? Like google, oidc, saml etc? If yes, can we rename the method accordingly. Or add examples of what might be returned.

app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/oauth2/OAuth2PropertiesCustomizer.java
```diff
@@ -0,0 +1,141 @@
...
+    private String extractProviderName(String propertyName) {
+        Matcher matcher = CLIENT_ID_PATTERN.matcher(propertyName);
+        if (matcher.matches() && matcher.groupCount() >= 1) {
+            return matcher.group(1);
+        }
+
+        // Fallback if regex fails
+        String withoutPrefix = propertyName.substring(OAUTH_CLIENT_ID_PREFIX.length());
+        return withoutPrefix.substring(0, withoutPrefix.indexOf('.'));
+    }
```
nsarupr, line 140: This method will only receive the inputs of type `spring.security.oauth2.client.registration.github.client-id` so do we still need to handle the fallbacks?

app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/oauth2/OAuth2PropertiesCustomizer.java
trishaanand: If this breaks, it would break during spring upgrade. And we should be able to catch it with SSO cypress tests.

app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/oauth2/OAuth2PropertiesCustomizer.java
trishaanand: No. We use spring configuration for registering any oauth clients. So till spring changes this property, we are sorted. If spring changes its property, then we would also need to update our application.properties.

nsarupr (approved):
Approving for unblocking the release.

app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/oauth2/OAuth2PropertiesCustomizer.java
trishaanand: Yes. It would return all of our application.properties known configs.

PR #40102, opened 2025-04-05
feat: Optimize Infinite Scroll Implementation in Table Widget

rahulbarwal:
## Description
This PR improves the infinite scroll implementation in the Table Widget by optimizing the scroll handling logic and removing unnecessary UI elements. The changes focus on making the infinite scroll more efficient and removing redundant code.

### 1. Enhanced Infinite Scroll Hook
- Replaced `useInfiniteScroll` with a more efficient `useEnhancedInfiniteScroll` hook
- Optimized ref usage by combining `isProcessingRef` and `lastLoadTriggerRef` into a single `loadingStateRef`
- Improved scroll position tracking logic to prevent unnecessary re-renders
- Added better handling of loading states and scroll triggers

### 2. BaseVirtualList Component Updates
- Removed the `LoadMoreButton` component and its related logic
- Added new props `onScroll` and `isItemLoaded` for better scroll handling
- Simplified the row rendering logic by removing conditional rendering for load more button
- Removed `LOAD_MORE_BUTTON_ROW` constant as it's no longer needed
- Added scroll event handling with debouncing to prevent excessive updates

### 3. InfiniteScrollBody Component
- Integrated the new `useEnhancedInfiniteScroll` hook
- Added proper handling of loading states and end of data conditions

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags="@tag.Table, @tag.Datasource, @tag.Sanity"

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No


app/client/src/widgets/TableWidgetV2/component/VirtualTable/InifiniteScrollBody/useInfiniteScroll.tsx
```diff
@@ -1,23 +1,113 @@
...
+   * We implement debouncing to avoid triggering unnecessary load more events, incorporating an additional timeout of 100 milliseconds to further prevent this.
+   * There is also a ref that indicates whether a load more request has just been triggered, serving as a safety net to prevent multiple simultaneous requests.
+   */
+  const debouncedLoadMore = useCallback(
+    debounce(() => {
+      if (!isLoading && !endOfData && !haveWeJustTriggeredLoadMoreRef.current) {
+        haveWeJustTriggeredLoadMoreRef.current = true;
+        loadMore();
+        setTimeout(() => {
+          haveWeJustTriggeredLoadMoreRef.current = false;
```
AmanAgarwal041, line 41: Suppose the loadMore request takes time, more than 100ms then will the debouncedLoadMore function not get triggered more than once ?

AmanAgarwal041: Should we have a callback or something implementation to make sure the loadMore ref is set `false` only when the loadmore finished the `haveWeJustTriggeredLoadMoreRef` becomes false ?

rahulbarwal: "Load more" is simply another term for "next page click." With the next page click, you can perform multiple operations. 
I believe we don't have that level of control to understand when it ends. 

A delay of 100 milliseconds seems like a reasonable starting point, but I'm open to alternatives. If you have a better approach, please suggest it; otherwise, I consider this a sufficient starting point.

PR #40301, opened 2025-04-17
chore: add dsl version validation for app computation cache

dvj1988:
## Description
This PR adds dsl version to the app computation cache. If there is a mismatch in dsl version the cache is updated with the new value.

Change also includes error handling and reporting for cases when there are exceptions while fetching the cache.

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/14533292018>
> Commit: fadb0c528b85b846799db2cbbd9da4d01834627f
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=14533292018&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Fri, 18 Apr 2025 11:07:11 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
  - Added explicit DSL version tracking to application state and evaluation context for improved cache validation.

- **Bug Fixes**
  - Improved error handling for cache operations, ensuring errors are logged and recorded without disrupting core functionality.
  - Enhanced cache validation to prevent usage of invalid or mismatched cache entries.

- **Tests**
  - Expanded and updated test coverage for cache validity, error scenarios, and DSL version handling.

- **Refactor**
  - Strengthened cache property validation and error propagation for more robust caching behavior.


app/client/src/ce/selectors/entitiesSelector.ts
```diff
@@ -1689,8 +1689,8 @@ export const getQuerySegmentItems = createSelector(
           : datasourceIdToNameMap[action.config.datasource.id] ?? "AI Queries";
       } else {
         group =
-          datasourceIdToNameMap[action.config.datasource?.id] ??
-          action.config.datasource?.name;
+          action.config.datasource?.name ??
+          datasourceIdToNameMap[action.config.datasource?.id];
```
dvj1988, line 1693:
```suggestion
          datasourceIdToNameMap[action.config.datasource?.id] ??
          action.config.datasource?.name;
```

app/client/src/workers/common/DependencyMap/index.ts
```diff
@@ -59,13 +63,23 @@ export async function createDependencyMap(
...
+  let dependencyMapCache: Record<string, string[]> | null = null;
+
+  try {
+    dependencyMapCache = await appComputationCache.getCachedComputationResult<
       Record<string, string[]>
     >({
       cacheProps,
       cacheName: EComputationCacheName.DEPENDENCY_MAP,
     });
+  } catch (error) {
```
vsvamsi1, line 75: So the changes over here is if we are unable to retrieve, we would get a null in the dependencyMapCache and subseqeunetly reconstruct the tree. Is my understanding correct @Diljit.

app/client/src/workers/common/AppComputationCache/index.ts
```diff
@@ -230,10 +238,8 @@ export class AppComputationCache {
 
       return computationResult;
     } catch (error) {
-      loglevel.error("Error getting cache result:", error);
-      const fallbackResult = await computeFn();
-
-      return fallbackResult;
+      loglevel.error(error);
```
vsvamsi1, line 241: If it can't retrieve it it can recompute the result like what the earlier code is doing.

app/client/src/workers/common/AppComputationCache/AppComputationCache.test.ts
```diff
@@ -240,13 +271,50 @@ describe("AppComputationCache", () => {
       expect(keysMock).toHaveBeenCalledTimes(1);
     });
 
+    test("should call deleteInvalidCacheEntries on dsl version mismatch after 10 seconds", async () => {
```
vsvamsi1, line 274: Is 10 seconds an arbitary number where the main thread becomes idle to delete the cache?

app/client/src/workers/common/AppComputationCache/index.ts
dvj1988: I have moved the fallback logic to where it is consumed. We need to throw the error here so that we can catch it where it is used. This would help us then send the error to the main catch block in eval tree and then eventually send it back to the main thread.

app/client/src/workers/common/DependencyMap/index.ts
dvj1988: If the cache is invalid or not present the appcomputation cache will send back null. The catch block is when we have an error thrown by the app computation cache.

app/client/src/workers/common/AppComputationCache/AppComputationCache.test.ts
dvj1988: yes it is arbitrary. The idea is to make it async.

I think this logic is incorrect now. There needs to be a change here. Will update it accordingly after testing this flow on ee.

vsvamsi1 (approved):

Issue #39831, opened 2025-03-20
[Task]: Add Run behavior in all Action Settings config


hetunandu:
### SubTasks

Update the configs in all the action settings json (Queries and JS) to have `runBehavior` select. And remove `executeOnLoad` setting.

PR #40339, opened 2025-04-22
feat: Replacing `executeOnLoad` with `runBehavior` in all action config settings

ankitakinger:
## Description

Replacing `executeOnLoad` with `runBehavior` in all action config settings

Figma: https://www.figma.com/design/qFXimqMSH45aviBOdTfDXu/Reactive-Queries-JS?node-id=61-16773&t=rv594VlMSm75Bvcb-0

Fixes [#39831](https://github.com/appsmithorg/appsmith/issues/39831)

## Automation

/ok-to-test tags="@tag.All"

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/client/src/components/formControls/DropDownControl.tsx
+++ b/app/client/src/components/formControls/DropDownControl.tsx
@@ -20,7 +20,7 @@ import {
 } from "workers/Evaluation/formEval";
 import type { Action } from "entities/Action";
 import type { SelectOptionProps } from "@appsmith/ads";
-import { Icon, Option, OptGroup, Select } from "@appsmith/ads";
+import { Icon, Option, OptGroup, Select, Flex, Text } from "@appsmith/ads";
 import { getFormConfigConditionalOutput } from "selectors/formSelectors";
 import { fetchFormDynamicValNextPage } from "actions/evaluationActions";
 import { objectKeys } from "@appsmith/utils";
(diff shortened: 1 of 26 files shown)


app/client/src/components/formControls/DropDownControl.tsx
```diff
@@ -495,7 +508,24 @@ function renderDropdown(
...
+      key={option.value}
+      label={option.label}
+      value={option.value}
+    >
+      <Flex flexDirection="column">
+        <Flex>
+          {option.icon && <Icon color={option.color} name={option.icon} />}
+          <OptionLabel>{option.label}</OptionLabel>
+        </Flex>
+        <OptionSubText>{option.subText}</OptionSubText>
```
AmanAgarwal041, line 525: Can we move this condition inside <Option children only instead of repeating `<Option` as any changes in the Option component , would be a repetitive one ? @ankitakinger

ankitakinger: Yes you're correct. The reason I did it this way (temporarily) is because I'm not completely sure where all this will impact and if any UI issues get introduced due to my change. So for safety did it this way as `subText` is right now being used only in our usecase.

AmanAgarwal041: Ohh! got it

AmanAgarwal041 (approved):

PR #40393, opened 2025-04-25
chore: Update type for run behavior for some instances

ankitakinger:
## Description

Update type for run behavior for some instances

Fixes #

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/client/src/actions/pluginActionActions.ts
+++ b/app/client/src/actions/pluginActionActions.ts
@@ -20,7 +20,7 @@ import type { EventLocation } from "ee/utils/analyticsUtilTypes";
 import type { GenerateDestinationIdInfoReturnType } from "ee/sagas/helpers";
 import type { Span } from "instrumentation/types";
 import type { EvaluationReduxAction } from "./EvaluationReduxActionTypes";
-import type { ActionRunBehaviour } from "PluginActionEditor/constants/PluginActionConstants";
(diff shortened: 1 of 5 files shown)


app/client/src/actions/pluginActionActions.ts
```diff
@@ -20,7 +20,7 @@ import type { EventLocation } from "ee/utils/analyticsUtilTypes";
 import type { GenerateDestinationIdInfoReturnType } from "ee/sagas/helpers";
 import type { Span } from "instrumentation/types";
 import type { EvaluationReduxAction } from "./EvaluationReduxActionTypes";
-import type { ActionRunBehaviour } from "PluginActionEditor/constants/PluginActionConstants";
+import type { ActionRunBehaviourType } from "PluginActionEditor/constants/PluginActionConstants";
```
AmanAgarwal041, line 23: Can we move the type inside types directory, as it will help us reduce the cyclic dependency in the EE repo ?

ankitakinger: Actually the old type for `executeOnLoad` was in this file and not declaring it there is giving some weird error on frontend about initialization. Hence I added this type there itself. 

Will check it out in depth in a separate PR.

AmanAgarwal041 (approved):

Issue #38787, opened 2025-01-21
[Bug]: Appsmith-EE helm chart refers to non-existing non-root user and group


ssichynskyi:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

In Appsmith-EE values.yaml there are 2 blocks instructing the user on how to run under non-root:
```yaml
podSecurityContext: {}
  # For non-root user
  #  fsGroup: 1001
  #  sysctls:
  #  - name: net.ipv4.ip_unprivileged_port_start
  #    value: "80"
securityContext: {}
  # For non-root user
  # runAsNonRoot: true
  # runAsUser: 1000
```
When I uncomment them and use like this:
```yaml
    # ....
    helm:
      releaseName: appsmith-ee
      values: |-
        # ....
        podSecurityContext:
          fsGroup: 1001
          sysctls:
          - name: net.ipv4.ip_unprivileged_port_start
            value: "80"
        securityContext:
          runAsNonRoot: true
          runAsUser: 1000
```
However has nor /etc/passwd on appsmith container is not aware about user with ID 1000.
When SSH in appsmith, the prompt is as follows:
```
groups: cannot find name for group ID 1001
I have no name!@appsmith-ee-0:/opt/appsmith$ ...
```

### Steps To Reproduce

1. Deploy the Appsmith-EE on k8s using the most recent helm chart (3.6.0) using abovementioned configs
2. SSH into the appsmith pod

### Environment

Production

### Severity

Medium (Frustrating UX)

### Version

Helm chart 3.6.0.

ssichynskyi:
I want to add a bit here. When running under this non-root user, it is not able to create a working dir for alloy. And therefore, the `appsmith-ee` is not in the running state.
Log looks like this:

```
appsmith 2025-03-11 08:37:44,671 INFO success: alloy entered RUNNING state, process has stayed up for > than 0 seconds (startsecs)
appsmith alloy stderr | Error: failed to create the remotecfg service: mkdir data-alloy: permission denied
appsmith 2025-03-11 08:37:44,674 INFO exited: alloy (exit status 1; not expected)
```
This error log disappears as soon as appsmith pod is switched under the root user. This is serious security issue, please increase the priority @Nikhil-Nandagopal

wyattwalter:
[...]

PR #40642, opened 2025-05-12
fix: allow running as non-root

wyattwalter:
## Description

Allows the Appsmith container to run as a non-root user, specified at runtime through either docker-compose or Kubernetes pod security context. I didn't specify the user in the `Dockerfile` because environments like OpenShift choose a user at runtime, so it can't be known at build time.

This needs to be followed by an update to docs and changes in the Helm chart to finish it off, but that has a separate release cycle and this needs to go ahead of that.

Ideally we would run as non-root by default, but since there's data persisted on the filesystem automatically transitioning the default is impossible without a lot of pain. This moves us in that direction and enables it in the future if we go down that path.

Required to fix https://github.com/appsmithorg/appsmith/issues/38787

## Automation

/ok-to-test tags=""

> [!WARNING]
> Tests have not run on the HEAD 50ba745c5bb7709c60ce5194437f921f1a95c980 yet
> <hr>Thu, 15 May 2025 15:56:31 UTC

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **New Features**
  - Added support for user identity emulation when running as a non-root user, improving compatibility in certain deployment environments.
  - Prevented embedded database initialization when running as a non-root user to ensure proper operation.

- **Chores**
  - Installed additional system packages to the base image for enhanced functionality.
  - Optimized image size by consolidating and improving cleanup steps during the build process.
  - Updated base image and refined installation commands for improved build consistency.

### DIFF
--- a/deploy/docker/base.dockerfile
+++ b/deploy/docker/base.dockerfile
@@ -40,14 +40,6 @@ RUN set -o xtrace \
     /var/lib/apt/lists/* \
     /tmp/*
 
-# libnss_wrapper.so is written to an architecture-specific directory, so we symlink to it in a common location to make it easier to activate
-ENV NSS_WRAPPER_SYMLINK=/usr/local/lib/libnss_wrapper.so
-RUN NSS_WRAPPER_LIB=$(find /usr/lib -name libnss_wrapper.so -type f 2>/dev/null | head -n1) && \
-    ln -sf "$NSS_WRAPPER_LIB" $NSS_WRAPPER_SYMLINK
-# these env vars need to be set for NSS Wrapper to work but don't matter until LD_PRELOAD is set which is optionally done at runtime
-ENV NSS_WRAPPER_PASSWD="${TMP}/passwd"
-ENV NSS_WRAPPER_GROUP="${TMP}/group"
-
 ENV PATH="/usr/lib/postgresql/14/bin:${PATH}"
 
 # Install Java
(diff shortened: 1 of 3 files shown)


deploy/docker/base.dockerfile
```diff
@@ -21,14 +21,32 @@ RUN set -o xtrace \
     supervisor curl nfs-common gnupg \
     gettext \
     ca-certificates \
+    libnss-wrapper \
+    git \
```
wyattwalter, line 25: Moving git here from the Dockerfile at the root. Once this is in the master branch we can delete the install there.

deploy/docker/base.dockerfile
```diff
@@ -21,14 +21,32 @@ RUN set -o xtrace \
     supervisor curl nfs-common gnupg \
     gettext \
     ca-certificates \
+    libnss-wrapper \
```
wyattwalter, line 24: This package is new.

deploy/docker/base.dockerfile
```diff
@@ -65,23 +76,20 @@ RUN set -o xtrace \
   && mkdir -p /opt/caddy \
   && version="$(curl --write-out '%{redirect_url}' 'https://github.com/caddyserver/caddy/releases/latest' | sed 's,.*/v,,')" \
   && curl --location "https://github.com/caddyserver/caddy/releases/download/v$version/caddy_${version}_linux_$(uname -m | sed 's/x86_64/amd64/; s/aarch64/arm64/').tar.gz" \
-  | tar -xz -C /opt/caddy
-
-RUN mv /opt/caddy/caddy /opt/caddy/caddy_vanilla
+    | tar -xz -C /opt/caddy && \
+  mv /opt/caddy/caddy /opt/caddy/caddy_vanilla
```
wyattwalter, line 80: This removed an extra layer, removing ~50MB from the image.

Issue #36858, opened 2024-10-14
[Bug]: Automatically set focus when adding a new row to a table widget


samiphoenix84:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

When adding a new row to a table, it would be useful if the focus was automatically set to the first editable field of the new row. This will avoid the need for the user to manually click on the input, providing a more intuitive user interface.

### Steps To Reproduce

Create a table widget.
Connect to a datasource
Allow adding new rows
Click 'Add new Row'

### Environment

Production

### Severity

Low (Cosmetic UI issues)

### Version

Cloud

PrasadMadine:
Hello @Nikhil-Nandagopal , I'm interested in picking up this issue.

rahulbarwal:
@PrasadMadine sure, go ahead

PrasadMadine:
Hi @rahulbarwal , Since the InlineCellEditor of PlainTextCell is re-rendering for every editable column of new added row the auto focus is being catched at the last cell rather than first cell.

![image](https://github.com/user-attachments/assets/994d668f-000f-482b-8b8e-65bf76897f78)

rahulbarwal:
That would not be correct experience. 
Instead of autofocus, why not force focus into the very first column we know to be editable. 
Note: this should happen only when user clicks on `Add new row`, after that user is free to put focus anywhere.

SamC1832js:
HI @rahulbarwal , can I work on this issue?

rahulbarwal:
@SamC1832js thanks for your interest in contributing! Please feel free to go ahead.

SamC1832js:
@rahulbarwal  Hi repo maintainer, can you help me out?

I ran into an issue during the local environment setup. On `dev.appsmith.com`, I couldn't connect to any data source. Here's a checklist of what I've done so far:

### **Checklist**
#### **Client**
- **WSL2**: Yes
- **Node.js**: 20.11.1 (Yes)
- **Docker Desktop**: Yes
- **WSL Distro**: Yes
- **Enable integration with the WSL Distro in Docker Desktop**: Yes
- **Install Remote-WSL extension in VSCode**: Yes

#### **Within WSL**
- **Repo cloned**: Yes
- **mkcert**: Yes
- **envsubst**: Yes
- **Yarn installed as a global npm package**: Yes
- **Local HTTPS certificates**: Yes
- **Backend server using Docker image**: Yes
- **Build & run the code using Yarn**: Yes

[...]

PR #40643, opened 2025-05-12
Focus first editable cell tableV2 widget : #36858

DejoeNithin:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/client/src/widgets/TableWidgetV2/component/TableContext.tsx
+++ b/app/client/src/widgets/TableWidgetV2/component/TableContext.tsx
@@ -45,6 +45,7 @@ export interface TableContextState
     width: number;
   };
   tableSizes: TableSizes;
+  focusFirstEditableCellInNewRow: () => void;
 }
 
 export const TableContext = createContext<TableContextState | undefined>(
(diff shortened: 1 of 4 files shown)


app/client/src/widgets/TableWidgetV2/component/TableContext.tsx
```diff
@@ -64,9 +65,30 @@ export const TableProvider = ({ children, ...state }: TableProviderProps) => {
...
+      if (!table) return;
+
+      const firstEditableInput = table.querySelector(
+        ".tableWrap .new-row .t--inlined-cell-editor input:not([disabled])",
+      );
+
+      if (firstEditableInput) (firstEditableInput as HTMLElement).focus();
+    }, 0);
+  }, [state.widgetId]);
+
```
rahulbarwal, line 83: Direct DOM manipulation is an anti-pattern in React and should be avoided whenever possible. 

Can you explore a solution using refs instead? 
I understand it can be complex, given that the first editable row can be at any position in the column list. 

However, I believe this is the best approach, as it allows us to implement the functionality in a React-compliant manner without relying on `setTimeouts` or creating brittle dependencies on selectors of the widget.

rahulbarwal (changes requested):
I greatly appreciate your time and effort in attempting to solve this problem, as well as your initiative in writing tests without needing reminders. However, I believe there may be a better approach available, and we should strive to find it.

app/client/src/widgets/TableWidgetV2/component/TableContext.tsx
AmanAgarwal041: Additionally, please try to implement the solution inside the `handleAddNewRow` function itself , to provide the integrated implementation as, passing it to the context will include one more property to be handled when the single use case for this function is to focus the element only on adding new row. 
For ref : https://github.com/appsmithorg/appsmith/blob/release/app/client/src/widgets/TableWidgetV2/widget/index.tsx#L2948-L2966 is the function where the focus implementation should be done. You can create a new function for focusing inside this class just call this function here instead of passing it down to the context.
What do you think ?

PR #40702, opened 2025-05-20
Fix consolidated api unauthorized check

dvj1988:
## Summary
- return 401 Unauthorized when `/consolidated-api/view` is called with a `defaultPageId` but without authentication

## Testing
- `mvn test` *(fails: mvn not found)*

> [!WARNING]
> Tests have not run on the HEAD 923e8b4be7698af8e48f26e8061ab89e722b6bf3 yet
> <hr>Tue, 20 May 2025 07:14:48 UTC

## Summary by CodeRabbit

- **New Features**
  - Added authentication check to the view mode endpoint. Unauthenticated users now receive a 401 Unauthorized response when accessing certain data.
- **Bug Fixes**
  - Improved security by ensuring sensitive data is only accessible to authenticated users.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ConsolidatedAPIController.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ConsolidatedAPIController.java
@@ -8,6 +8,7 @@ import com.appsmith.server.domains.ApplicationMode;
 import com.appsmith.server.dtos.ConsolidatedAPIResponseDTO;
 import com.appsmith.server.dtos.ResponseDTO;
 import com.appsmith.server.services.ConsolidatedAPIService;
+import com.appsmith.server.services.SessionUserService;
 import com.fasterxml.jackson.annotation.JsonView;
 import io.micrometer.observation.ObservationRegistry;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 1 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ConsolidatedAPIController.java
```diff
@@ -100,40 +105,69 @@ public Mono<ResponseEntity<ResponseDTO<ConsolidatedAPIResponseDTO>>> getAllDataF
...
-                    }
-
-                    ResponseDTO<ConsolidatedAPIResponseDTO> responseDTO =
-                            new ResponseDTO<>(HttpStatus.OK, consolidatedAPIResponseDTO);
-
-                    return new ResponseEntity<>(responseDTO, headers, HttpStatus.OK);
-                })
+        Mono<Boolean> isAuthenticatedMono = sessionUserService
+                .getCurrentUser()
+                .map(user -> true)
```
trishaanand, line 110: This is incorrect.You would always be returned a user (either the logged in user, or a user object whose email is anonymousUser

dvj1988: In that case can we send a 401 for anonymous users? We should able to distinguish an anonymous user from a logged in user right?

PR #40793, opened 2025-05-29
fix: allow running as non-root (#40642)

pratapaprasanna:
## Description

Allows the Appsmith container to run as a non-root user, specified at runtime through either docker-compose or Kubernetes pod security context. I didn't specify the user in the `Dockerfile` because environments like OpenShift choose a user at runtime, so it can't be known at build time.

This needs to be followed by an update to docs and changes in the Helm chart to finish it off, but that has a separate release cycle and this needs to go ahead of that.

Ideally we would run as non-root by default, but since there's data persisted on the filesystem automatically transitioning the default is impossible without a lot of pain. This moves us in that direction and enables it in the future if we go down that path.

Required to fix https://github.com/appsmithorg/appsmith/issues/38787

## Automation

/ok-to-test tags=""

> [!WARNING]
> Tests have not run on the HEAD
50ba745c5bb7709c60ce5194437f921f1a95c980 yet
> <hr>Thu, 15 May 2025 15:56:31 UTC

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

- **New Features**
- Added support for user identity emulation when running as a non-root user, improving compatibility in certain deployment environments.
- Prevented embedded database initialization when running as a non-root user to ensure proper operation.

- **Chores**
- Installed additional system packages to the base image for enhanced functionality.
- Optimized image size by consolidating and improving cleanup steps during the build process.
- Updated base image and refined installation commands for improved build consistency.

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/deploy/docker/base.dockerfile
+++ b/deploy/docker/base.dockerfile
@@ -21,6 +21,8 @@ RUN set -o xtrace \
     supervisor curl nfs-common gnupg \
     gettext \
     ca-certificates \
+    libnss-wrapper \
+    git \
   # Install MongoDB v6, Redis, PostgreSQL v14
   && curl -fsSL https://www.mongodb.org/static/pgp/server-6.0.asc | gpg --dearmor -o /usr/share/keyrings/mongodb-server-6.0.gpg \
   && echo "deb [ arch=amd64,arm64 signed-by=/usr/share/keyrings/mongodb-server-6.0.gpg ] https://repo.mongodb.org/apt/ubuntu jammy/mongodb-org/6.0 multiverse" | tee /etc/apt/sources.list.d/mongodb-org-6.0.list \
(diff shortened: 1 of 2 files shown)


moinmoin:
Regarding Openshift, this **update seems like a hack**. Openshift allows to "pin" the UID of an namespace.
https://docs.redhat.com/en/documentation/openshift_container_platform/4.16/html/authentication_and_authorization/managing-pod-security-policies
see -> "nonroot"

After adapting the Appsmith Dockerfile for a long time already in order to make a non-root build, checking for components NOT running as root would be much better, than hacking around it. I remember that caddy used port 80 as default. This would be an easier fix than adding libnss to the mix.

PR #40821, opened 2025-06-02
feat: handled error condition from cs

### LINKED ISSUE #40820: [Task]: Replace error thrown when no plugins found from CS in cronjob to log message
### SubTasks

Replace error thrown when no plugins found from CS in cronjob to log message
- Due to higher frequency of the occurrence of the above scenario we are getting lot of error messages in our logs, these should be actually log messages as this is the expected behaviour.

NilanshBansal:
## Description
This PR improves error handling when fetching plugins from CS. 
- Distinguish between legitimate "no plugins found" cases (not an error) and actual cloud service errors
- Return Mono.empty() with log info when no plugins are found (normal case)
- Propagate errors with Mono.error() when cloud services returns actual errors
- Preserve proper error details in exception messages

Fixes #40820 
Corresponding EE PR: https://github.com/appsmithorg/appsmith-ee/pull/7696

## Automation

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/15393377154>
> Commit: 8a2c61f27a7069d2f4babed0ccc47ac8e43459e7
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=15393377154&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Sanity`
> Spec:
> <hr>Mon, 02 Jun 2025 14:01:39 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling when fetching plugins from cloud services. Users will now receive clearer error notifications if there is an issue, rather than silently seeing an empty plugin list.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ce/PluginScheduledTaskUtilsCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ce/PluginScheduledTaskUtilsCEImpl.java
@@ -47,15 +47,20 @@ public class PluginScheduledTaskUtilsCEImpl implements PluginScheduledTaskUtilsC
                         clientResponse.bodyToMono(new ParameterizedTypeReference<ResponseDTO<List<Plugin>>>() {}))
                 .map(listResponseDTO -> {
                     if (listResponseDTO.getData() == null) {
-                        log.error(
-                                "Error fetching plugins from cloud-services. Error: {}",
-                                listResponseDTO.getErrorDisplay());
+                        String errorMessage = listResponseDTO.getErrorDisplay();
+                        log.error("Error fetching plugins from cloud-services. Error: {}", errorMessage);
+
+                        // If there's an actual error message, propagate it as an error
+                        if (errorMessage != null && !errorMessage.isEmpty()) {
+                            throw new RuntimeException("Cloud Services error: " + errorMessage);
+                        }
+
+                        // Otherwise, return empty list (no plugins found)
                         return Collections.<Plugin>emptyList();
                     }
                     return listResponseDTO.getData();
                 })
                 .map(plugins -> {
-
                     // Parse plugins into map for easier manipulation
                     return plugins.stream()
                             .collect(Collectors.toMap(
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ce/PluginScheduledTaskUtilsCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ce/PluginScheduledTaskUtilsCEImpl.java
@@ -48,8 +48,6 @@ public class PluginScheduledTaskUtilsCEImpl implements PluginScheduledTaskUtilsC
                 .map(listResponseDTO -> {
                     if (listResponseDTO.getData() == null) {
                         String errorMessage = listResponseDTO.getErrorDisplay();
-                        log.error("Error fetching plugins from cloud-services. Error: {}", errorMessage);
-
                         // If there's an actual error message, propagate it as an error
                         if (errorMessage != null && !errorMessage.isEmpty()) {
                             throw new RuntimeException("Cloud Services error: " + errorMessage);


app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ce/PluginScheduledTaskUtilsCEImpl.java
```diff
@@ -47,15 +47,20 @@ protected Mono<Map<PluginScheduledTaskCEImpl.PluginIdentifier, Plugin>> fetchPlu
                         clientResponse.bodyToMono(new ParameterizedTypeReference<ResponseDTO<List<Plugin>>>() {}))
                 .map(listResponseDTO -> {
                     if (listResponseDTO.getData() == null) {
-                        log.error(
-                                "Error fetching plugins from cloud-services. Error: {}",
-                                listResponseDTO.getErrorDisplay());
+                        String errorMessage = listResponseDTO.getErrorDisplay();
+                        log.error("Error fetching plugins from cloud-services. Error: {}", errorMessage);
```
trishaanand: This is not required. The error would be logged (if present) by the run time exception that you are throwing in line 55

NilanshBansal: @trishaanand the log statement at line 51 also logs in case of empty plugin list from the CS. There are two conditions here: 
1. If there is no new update from the CS plugins, it returns empty list
2.  If there is an error from the CS API
The line 51 logs in both cases. For the first case, it is a single log statement without the log trace.

trishaanand: Since scenario 1 is still in play at this point, do you think we should be printing an error here? Also, can you confirm that log.error is NOT printing the stack trace at line 51?

NilanshBansal: @trishaanand just checked, we were duplicating the error log. Have removed it now. Please ignore my previous comment.

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/ce/PluginScheduledTaskUtilsCEImpl.java
```diff
@@ -47,15 +47,18 @@ protected Mono<Map<PluginScheduledTaskCEImpl.PluginIdentifier, Plugin>> fetchPlu
                         clientResponse.bodyToMono(new ParameterizedTypeReference<ResponseDTO<List<Plugin>>>() {}))
                 .map(listResponseDTO -> {
                     if (listResponseDTO.getData() == null) {
-                        log.error(
-                                "Error fetching plugins from cloud-services. Error: {}",
-                                listResponseDTO.getErrorDisplay());
+                        String errorMessage = listResponseDTO.getErrorDisplay();
+                        // If there's an actual error message, propagate it as an error
+                        if (errorMessage != null && !errorMessage.isEmpty()) {
+                            throw new RuntimeException("Cloud Services error: " + errorMessage);
```
rahulbarwal, line 53: Are we certain that this condition only occurs when cloud services generate an error and not under any other circumstances? 

This seems like a generic approach to determine whether an issue is related to a cloud service error. We need to ensure that the error message will only be populated in the case of a cloud service error. If this is not the case, we should consider a more precise solution.

NilanshBansal: Yes @rahulbarwal as per the cloud services response format, we always send errorMessage in case of any errors. Hence this condition works fine as we are only interacting with cs here.

Issue #40813, opened 2025-05-30
[Bug]: UI shows table being updated twice after Query execution


sneha122:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

After execution is complete and the Table widgets are updated, notice in the video that the Query1 still shows to be executing in the UI while in reality its done executing.

https://www.notion.so/appsmith/203fe271b0e280bd8852f2321a6e829a?v=203fe271b0e280ed8015000cb9056bfb&p=203fe271b0e280109c15fddf8a9e7236&pm=s

### Steps To Reproduce

After execution is complete and the Table widgets are updated, notice in the video that the Query1 still shows to be executing in the UI while in reality its done executing.

### Environment

Deploy Preview

### Severity

High (Blocker to building or releasing)

### Version

NA

ankitakinger:
Re-opening as the previous change had to be reverted:
https://github.com/appsmithorg/appsmith/pull/40922

ankitakinger:
Re-opening as the previous change had to be reverted:
#40922

ankitakinger:
Fix added in reactivity-fix PR: https://github.com/appsmithorg/appsmith-ee/pull/7440

PR #40239, opened 2025-04-14
fix: code editor changes

vsvamsi1:
## Description
We've encountered a bug in the REST API plugin where rapidly changing the URL input causes the save status to get stuck in the loading state. This happens because the evaluation is debounced, and by the time it's ready to run, the inputs may have changed in a way that prevents the evaluation from being triggered. However, we still initiate a saga that tracks the terminal state and controls the loading status. Since the evaluation never actually occurs, the terminal state is never reached, causing the loading status to remain stuck.

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  
> _If no issue exists, please create an issue first, and check with the maintainers if the issue is valid._

## Automation

/ok-to-test tags="@tag.All"

### :mag: Cypress test results

> [!CAUTION]
> 🔴 🔴 🔴 Some tests have failed.
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/15130914842>
> Commit: d77321feaaaec58ae0d377d595174fcaa8630ae3
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=15130914842&attempt=1&selectiontype=test&testsstatus=failed&specsstatus=fail" target="_blank">Cypress dashboard</a>.
> Tags: @tag.All
> Spec: 
> The following are new failures, please fix them before merging the PR: <ol>
> <li>cypress/e2e/Regression/Apps/PromisesApp_spec.js
> <li>cypress/e2e/Regression/ClientSide/ActionExecution/ActionSelector_JsToNonJSMode_4_spec.ts
> <li>cypress/e2e/Regression/ClientSide/ActionExecution/FrameworkFunctions_CopyToClipboardFunctions_spec.ts
> <li>cypress/e2e/Regression/ClientSide/ActionExecution/FrameworkFunctions_DownloadFunctions_spec.ts
> <li>cypress/e2e/Regression/ClientSide/ActionExecution/FrameworkFunctions_LocalStoreValueFunctions_spec.ts
> <li>cypress/e2e/Regression/ClientSide/ActionExecution/FrameworkFunctions_ResetWidgetFunctions_spec.ts
> <li>cypress/e2e/Regression/ClientSide/ActionExecution/StoreValue_spec.ts
> <li>cypress/e2e/Regression/ClientSide/Autocomplete/Autocomplete_setters_spec.ts
[...]

PR #40849, opened 2025-06-03
chore: Adding debounce to `onValueChange` for input widgets

ankitakinger:
## Description

Adding debounce to `onValueChange` for input widgets to fix multiple Execute API calls happening in reactive queries flow.

Fixes [#40813](https://github.com/appsmithorg/appsmith/issues/40813)

## Automation

/ok-to-test tags="@tag.All"

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/client/src/widgets/BaseInputWidget/component/index.tsx
+++ b/app/client/src/widgets/BaseInputWidget/component/index.tsx
@@ -9,7 +9,7 @@ import {
   ControlGroup,
   Tag,
 } from "@blueprintjs/core";
-import _, { isNil } from "lodash";
+import _, { debounce, isNil } from "lodash";
 
 import type { ComponentProps } from "widgets/BaseComponent";
 import { Colors } from "constants/Colors";
@@ -430,7 +430,7 @@ class BaseInputComponent extends React.Component<
 > {
   constructor(props: BaseInputComponentProps) {
     super(props);
-    this.state = { showPassword: false };
+    this.state = { showPassword: false, inputValue: this.props.value || "" };
   }
 
   componentDidMount() {
@@ -450,6 +450,8 @@ class BaseInputComponent extends React.Component<
         );
       }
     }
+
+    this.setState({ inputValue: this.props.value || "" });
   }
 
   componentWillUnmount() {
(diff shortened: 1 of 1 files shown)


app/client/cypress/e2e/Regression/ClientSide/Widgets/Input/Inputv2_spec.js
```diff
@@ -385,7 +385,9 @@ describe(
       cy.get(widgetInput).clear();
       cy.wait(300);
       // Input text and hit enter key
-      cy.get(widgetInput).type("test{enter}");
+      cy.get(widgetInput).type("test{enter}", {
+        delay: 300,
```
sneha122: @ankitakinger Can we add a comment here for why delay of 300ms is added here? We can also take it out in constant file similar to what was done in this [PR](https://github.com/appsmithorg/appsmith/pull/40239), so that if in future, the value 300 changes in the code, it will be updated for cypress as well

app/client/cypress/e2e/Regression/ClientSide/Widgets/ListV2/Listv2_BasicClientSideData_spec.js
```diff
@@ -76,7 +76,7 @@ describe(
       cy.get(".t--draggable-inputwidgetv2").each(($inputWidget, index) => {
         cy.wrap($inputWidget)
           .find("input")
-          .type(index + 1);
+          .type(index + 1, { delay: 300 });
```
sneha122: Also as done in this [PR](https://github.com/appsmithorg/appsmith/pull/40239), can we make sure that delay is a bit more than 300? This is to avoid flakiness because if in race condition situation the input change hasn't happened but cypress is done waiting for 300 ms then the test would become flaky, WDYT?

app/client/cypress/support/Pages/AggregateHelper.ts
```diff
@@ -973,7 +976,10 @@ export class AggregateHelper {
     force = false,
   ) {
     this.ClearTextField(selector, force, index);
-    return this.TypeText(selector, totype, index);
+    return this.TypeText(selector, totype, {
+      index,
+      delay: 200,
```
sneha122: Is there any reason why this is 200 where as our debounce is 300?

app/client/cypress/e2e/Regression/ClientSide/Widgets/ListV2/Listv2_BasicClientSideData_spec.js
ankitakinger: I think as of now lets keep it 300 itself... If we add more then it might cause more failures. If we see flakiness later, we can take an action accordingly that time. As I've ran all tests multiple times now, and those don't seem to be flaky.

app/client/cypress/support/Pages/AggregateHelper.ts
ankitakinger: Because here it was working out with 200. But I'll change it to keep it consistent.

app/client/cypress/e2e/Regression/ClientSide/Widgets/Input/Inputv2_spec.js
ankitakinger: Done

sneha122 (approved):
LGTM

PR #40878, opened 2025-06-05
fix: Use non blocking SCAN command instead of blocking keys for redis…

trishaanand:
… pattern match

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/test all

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No


trishaanand:
> Oh wow I thought keys() is a flux and this change should not have any effect but after going through the implementation realised that it's actually equivalent to:
> 
> ```
> Mono<List<>>.flatMap(list -> Flux.fromIterable(list))
> ```
> 
> I don't know why people blocks the reactive chain and load all the items in memory again converting this to Flux 🤔.

@abhvsn : This is not blocking on the server but rather KEYS command is blocking on redis whereas SCAN is a non blocking asynchronous op.

Issue #31925, opened 2024-03-19
[Bug]: `currentItem is not defined` in Empty List Widget


amelia-c0n:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

When a List widget is empty, the error message `currentItem is not defined` gets displayed even though the container and widgets are hidden. This is confusing to users and leads to a poor customer experience.

![image](https://github.com/appsmithorg/appsmith/assets/101268423/c46c4afc-4830-4e95-90dc-221ae71e250a)

### Steps To Reproduce

1. Add a **List widget** to the canvas.
2. Ensure that the query/API that populates it returns no data or hardcode the **Items** using `[]`.

### Environment

Production

### Severity

Medium (Frustrating UX)

### Version

v1.17

sonali-humanedge:
Hey @Nikhil-Nandagopal @somangshu @pranavkanade @sribalajig any progress on this issue?

szalapski:
I would add that even making the expression more robust still false-alarms the linter. For example, `{{currentItem?.id ?? ""}}` still gives the same linter error **even though the expression can never be null or undefined.**

So I am hitting this too. It creates noise in the linter, which will lead to ignoring the linter altogether, which will lead to bad apps, which will lead to losing money.  Please fix!

Also, please tell me of any workaround you can think of, such as further improving the expression.

PR #40886, opened 2025-06-06
feat: Implement empty meta widget generation for handling empty data scenarios

rahulbarwal:
## Description
<ins>Problem</ins>

Errors appeared in the console when the List widget in edit mode received an empty array as data.

<ins>Root cause</ins>

In edit mode, the List widget uses template widgets to create meta widgets. When the list becomes empty, meta widgets are not generated, but template widgets with `currentItem` bindings remain. Since `currentItem` requires meta widgets for evaluation, this results in errors.

<ins>Solution</ins>

This PR handles the generation of empty meta widgets when the List widget receives empty data, ensuring `currentItem` bindings don’t break. It introduces a `generateEmptyMetaWidgets` method in `MetaWidgetGenerator`, enhances the List widget to track and respond to empty data states, and resets the meta widget cache when transitioning between empty and non-empty states.

Fixes #`Issue Number`  
_or_  
Fixes https://github.com/appsmithorg/appsmith/issues/31925
> [!WARNING]  

## Automation

/ok-to-test tags="@tag.List"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/15756809644>
> Commit: 58818e09dd83faaa92ee0dad53aab93dcf5f55a5
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=15756809644&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.List`
> Spec:
> <hr>Thu, 19 Jun 2025 12:11:58 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of empty list data on the first page, ensuring widgets display correctly when no data is present.
  - Enhanced consistency in detecting and rendering empty lists across the List Widget.

- **New Features**
  - Added a utility function to accurately identify fully empty lists, enabling more predictable widget behavior.
  - Introduced an option to handle empty list cases by injecting placeholder data for consistent widget generation.

- **Tests**
  - Added a test to verify the display of "No data to display" message when the list is empty and confirm no errors occur.

### DIFF
--- a/app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.test.ts
+++ b/app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.test.ts
@@ -167,6 +167,7 @@ const init = ({
   const options = klona({
     ...DEFAULT_OPTIONS,
     ...optionsProps,
+    isEmptyListWidgetCase: false,
   });
   const cache = passedCache || new Cache();
 
@@ -850,6 +851,7 @@ describe("#generate", () => {
         levelData,
         pageSize: 3,
         widgetName: "List6",
+        isEmptyListWidgetCase: false,
       })
       .generate();
 
@@ -897,6 +899,7 @@ describe("#generate", () => {
         levelData,
         pageSize: 3,
         widgetName: "List6",
+        isEmptyListWidgetCase: false,
       })
       .generate();
 
@@ -957,6 +960,7 @@ describe("#generate", () => {
         levelData,
         pageSize: 3,
         widgetName: listWidgetName,
+        isEmptyListWidgetCase: false,
       })
       .generate();
 


app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.ts
```diff
@@ -469,6 +469,120 @@ class MetaWidgetGenerator {
...
+      ...(metaWidget.dynamicBindingPathList || []),
+      { key: "currentItem" },
+    ];
+  };
+
+  /**
+   * Process dynamic paths for empty data state
+   * This is similar to addDynamicPathsProperties but handles empty data scenario
+   */
+  private addDynamicPathsPropertiesForEmptyData = (
```
ashit-rath: What is the difference between this and the existing addDynamicPathProperties?
Why are currentView and levelPaths not part of this?

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -602,6 +605,25 @@ class ListWidget extends BaseWidget<
   };
 
   generateMetaWidgets = () => {
+    const { listData } = this.props;
+    const hasEmptyData = !listData || listData.length === 0;
+
+    // Check if data is empty and generate empty meta widgets to prevent currentItem binding errors
+    if (hasEmptyData) {
+      this.previouslyHadEmptyData = true;
+      this.generateEmptyMetaWidgets();
```
ashit-rath: Weird idea: what if instead of having a specific function for generating empty meta widgets; when data is empty or undefined we simply pass `[{}]` as data to the list widget generator. We just need the first row of widgets which it can produce. Not 100% sure if it would work or not but the idea is to play with the data property and make the existing logic work rather than handling it as a special case.

rahulbarwal: Understood. That makes sense. I will try this approach. I don't have an answer to your previous question at the moment, as I may have overlooked it while working on this solution with AI assistance. I will address this, along with your last question. Now that we are aligned on pursuing this solution, I will explore ways to make it foolproof.

app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.ts
```diff
@@ -1106,14 +1106,32 @@ class MetaWidgetGenerator {
...
+    } else {
+      const shouldAddDataCacheToBinding = this.shouldAddDataCacheToBinding(
+        metaWidgetId,
+        key,
+      );
+
+      dataBinding = shouldAddDataCacheToBinding
+        ? `{{${JSON.stringify(this.cachedKeyDataMap[key])}}}`
+        : `{{${this.widgetName}.listData[${metaWidgetName}.currentIndex]}}`;
+    }
```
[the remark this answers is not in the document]
rahulbarwal: When we receive an empty object as the first item in the data array, it should be treated as empty data specifically for the list project. In this PR, I am adding this empty object to ensure it is recognized as empty when passed through, which serves as a case for generating empty meta widgets. There is no scenario in which a user can have a data list containing a single item that is an empty object.

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -559,6 +563,10 @@ class ListWidget extends BaseWidget<
     } = this.props;
     const pageSize = this.pageSize;
 
+    if (isListFullyEmpty(listData, pageNo)) {
+      listData.push({});
```
ashit-rath: We shouldn't mutate the props

app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.ts
```diff
@@ -1106,14 +1106,32 @@ class MetaWidgetGenerator {
...
+      this.pageNo === 1;
 
-    const dataBinding = shouldAddDataCacheToBinding
-      ? `{{${JSON.stringify(this.cachedKeyDataMap[key])}}}`
-      : `{{${this.widgetName}.listData[${metaWidgetName}.currentIndex]}}`;
+    let dataBinding: string;
+
+    if (isEmptyDataCase) {
+      // For empty data case, set currentItem to empty object
+      dataBinding = "{{{}}}";
```
ashit-rath, line 1124: I didn't quite understand why we want to do this? In worst case the the `listData` would be `[]` so the result of the binding in line 1131 after eval would be `undefined` and this is causing error in the first row meta widgets?

app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.ts
```diff
@@ -1106,14 +1106,32 @@ class MetaWidgetGenerator {
...
-      metaWidgetId,
-      key,
-    );
+    /**
+     * Check if we're in the empty data case (when original listData was empty, we add [{}])
+     * The page condition is included because there may be instances when a user uses navigation controls and lands on a specific page number,
+     * such as page 3, which returns an empty response.
+     * In this case, the list data will be empty, and we want to avoid generating empty meta widgets for that scenario.
+     */
+    const isEmptyDataCase =
```
ashit-rath: Do you think instead of detecting the empty case like this we just directly send it from the Listwidget as a prop to MetaWidgetGenerator like `isListDataEmpty`?

app/client/src/widgets/ListWidgetV2/widget/index.tsx
rahulbarwal: Agreed, made changes to mutate just the data for metawidget generator.

app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.ts
rahulbarwal: When the list data is empty, we have sufficient guardrails to prevent reaching this execution point. The list data must contain elements for the system to access and add properties of the current item.

app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.ts
rahulbarwal: I made a similar change. Even with a property indicating whether the list data is empty, I still need to pass a single empty object within an array as the list data prop. Overall, I don't think it matters whether we send this as an extra variable or calculate it on the spot.

ashit-rath: What I was trying to point out was passing a prop and using makes it more readable than checking empty case like this but I'll let you take that call

app/client/src/widgets/ListWidgetV2/widget/index.tsx
```diff
@@ -558,6 +562,11 @@ class ListWidget extends BaseWidget<
       serverSidePagination = false,
     } = this.props;
     const pageSize = this.pageSize;
+    const data = this.props.listData;
+
+    if (data && isListFullyEmpty(data, pageNo)) {
+      data.push({});
```
ashit-rath: @rahulbarwal Doesn't this still mutate the `listData` property?

app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.ts
rahulbarwal: Implemented relevant changes

app/client/src/widgets/ListWidgetV2/widget/index.tsx
rahulbarwal: Yes, resolved it using ternary operator.

app/client/src/widgets/ListWidgetV2/MetaWidgetGenerator.ts
rahulbarwal: Without the if-else check, the binding on line 1131 is not an empty array; it is bound to the current list widget, as shown in the screenshot. It appears that this binding is shared with the actual widget. When the eval runs, it likely causes a crash because the list.data property is unavailable. The current item variable has been removed. Any insights you could provide on this matter would be appreciated.
Screenshot 2025-06-19 at 4 49 26 PM

ashit-rath (dismissed):

PR #40905, opened 2025-06-10
fix: improve null safety in action execution analytics data collection

### LINKED ISSUE #40904: [Task]: Resolve NPEs in action execution log for grafana sendExecuteAnalyticsEvent()
### SubTasks

We observed some NPEs for action execution in grafana logs. 
https://appsmithmonitoringcloud.grafana.net/goto/gvZfE5YHR?orgId=1

Ref: https://theappsmith.slack.com/archives/CPQNLFHTN/p1749459309955479

NilanshBansal:
## Description
Fix NPE in ActionExecutionSolutionCEImpl by replacing `Map.of()` with explicit HashMap entries. This change improves error resilience by adding null checks with default values for all map entries and makes the code more maintainable by separating each property assignment into individual operations.

Fixes #40904 

## Automation

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImpl.java
@@ -1127,27 +1127,21 @@ public class ActionExecutionSolutionCEImpl implements ActionExecutionSolutionCE 
                             ? ApplicationMode.PUBLISHED.toString()
                             : ApplicationMode.EDIT.toString();
 
-                    final Map<String, Object> data = new HashMap<>(Map.of(
-                            "username",
-                            user.getUsername(),
-                            "type",
-                            pluginType,
-                            "pluginName",
-                            plugin.getName(),
-                            "name",
-                            actionDTO.getName(),
-                            "datasource",
-                            Map.of("name", datasourceStorage.getName()),
-                            "workspaceId",
-                            application.getWorkspaceId(),
-                            "appId",
-                            actionDTO.getApplicationId(),
-                            FieldName.APP_MODE,
-                            appMode,
-                            "appName",
-                            application.getName(),
-                            "isExampleApp",
-                            application.isAppIsExample()));
+                    final Map<String, Object> data = new HashMap<>();
+                    data.put("username", ObjectUtils.defaultIfNull(user.getUsername(), "anonymous"));
+                    data.put("type", ObjectUtils.defaultIfNull(pluginType, "UNKNOWN_PLUGIN_TYPE"));
+                    data.put("pluginName", ObjectUtils.defaultIfNull(plugin.getName(), "unknown"));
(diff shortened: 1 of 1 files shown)


abhvsn (commented):
@NilanshBansal though `Map.of()` doesn't allow the null key and values,`HashMap` does. Also in the analyticsService we are converting the null values to empty string to avoid the NPE of segment library [here](https://github.com/appsmithorg/appsmith-ee/blob/a90f26ef2c54f0852ad9b3b30b12f34241503aa6/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCEImpl.java#L225).  Do you think instead of adding the fallback here for each field we should add this selectively only for the required fields where the fallback is not an empty string? Currently the code looks un-necessarily boated.

NilanshBansal:
> @NilanshBansal though `Map.of()` doesn't allow the null key and values,`HashMap` does. Also in the analyticsService we are converting the null values to empty string to avoid the NPE of segment library [here](https://github.com/appsmithorg/appsmith-ee/blob/a90f26ef2c54f0852ad9b3b30b12f34241503aa6/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/AnalyticsServiceCEImpl.java#L225). Do you think instead of adding the fallback here for each field we should add this selectively only for the required fields where the fallback is not an empty string? Currently the code looks un-necessarily boated.

Yes @abhvsn that makes sense! Let me remove the fallback values just make the changes for map.of() to hashmap

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImpl.java
```diff
@@ -1160,37 +1154,28 @@ private Mono<ActionExecutionRequest> sendExecuteAnalyticsEvent(
...
-                            timeElapsed,
-                            "actionCreated",
-                            DateUtils.ISO_FORMATTER.format(actionDTO.getCreatedAt()),
-                            "actionId",
-                            ObjectUtils.defaultIfNull(actionDTO.getId(), "")));
-                    data.putAll(Map.of(
+                            ObjectUtils.defaultIfNull(actionExecutionResult.getIsExecutionSuccess(), false));
+                    data.put("statusCode", ObjectUtils.defaultIfNull(actionExecutionResult.getStatusCode(), ""));
+                    data.put("timeElapsed", timeElapsed);
+                    data.put("actionCreated", DateUtils.ISO_FORMATTER.format(actionDTO.getCreatedAt()));
```
abhvsn, line 1163: @NilanshBansal I'm guessing actionDTO will never be null otherwise we will again end up in NPE.

abhvsn (approved):

app/server/appsmith-server/src/main/java/com/appsmith/server/solutions/ce/ActionExecutionSolutionCEImpl.java
NilanshBansal: Yes that's correct @abhvsn! `actionDTO` will never be null.

PR #41031, opened 2025-06-24
chore: cache theme value properties, since it is a frequent property

vsvamsi1:
## Description
Added caching of theme property value since it is a frequent expression, it constitutes 20% of all binding expressions for a large customer app. Expecting a 400ms reduction in LCP for a large customer app.

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/15880337835>
> Commit: 11fab20fe285aa1b3b59c164179902628d35d97d
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=15880337835&attempt=2" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Wed, 25 Jun 2025 17:45:37 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- **New Features**
  - Improved performance when evaluating theme-related properties by introducing caching for repeated values.

- **Chores**
  - Added a utility to identify specific theme-related unevaluated values.

### DIFF
--- a/app/client/src/ce/entities/DataTree/utils.ts
+++ b/app/client/src/ce/entities/DataTree/utils.ts
@@ -84,3 +84,14 @@ export function isWidgetActionOrJsObject(
 ): entity is ActionEntity | WidgetEntity | JSActionEntity {
   return isWidget(entity) || isAction(entity) || isJSAction(entity);
 }
+
+const THEME_PROPERTIES = new Set([
+  "{{appsmith.theme.fontFamily.appFont}}",
+  "{{appsmith.theme.boxShadow.appBoxShadow}}",
+  "{{appsmith.theme.colors.primaryColor}}",
+  "{{appsmith.theme.borderRadius.appBorderRadius}}",
+]);
+
+export function isThemeUnevaluatedValue(value: string): boolean {
+  return THEME_PROPERTIES.has(value);
+}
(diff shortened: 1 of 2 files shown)


riodeuno:
@vsvamsi1 Coderabbit has a review comment.

Have we confirmed the performance gains in this?
Does this change anything in the product behaviour w.r.t to theme changes? It seems that the cache, is an in memory cache and only lasts until the page is refreshed. Can you think of an edge scenario where the theme needs to update, but it doesn't? When we change the theme __ in any way __ do the changes propagate?

vsvamsi1:
HI @riodeuno 
This optimisation focuses on the first evaluation, where all expressions within the unevaluated tree are processed. I’ve observed that five specific theme properties are consistently present across most widgets. In large applications like Xolair, there are approximately 700 such expressions during the first evaluation.

An in-memory cache is sufficient for this scenario. It remains simple, covers both the first evaluation and subsequent evaluation cycles when the theme is modified. Since the cache is used within a single evaluation cycle, I believe we don't need a more complex solution, such as maintaining a persistent cache across multiple cycles with action tracking for invalidation. Introducing that level of complexity doesn't seem necessary for this use case.

Issue #41125, opened 2025-07-22
[Bug]: Reactive dependency misuse error toast is displayed when user generate CRUD page using Sample mongo movies datasource


RakshaKShetty:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

Reactive dependency misuse error toast is displayed when user generate CRUD page using Sample mongo movies datasource
<img width="1440" height="900" alt="Image" src="https://github.com/user-attachments/assets/e49ad959-7cb7-451e-b656-e9098eade244" />

### Steps To Reproduce

1. go to app.appsmith.com
2. Create a new workspace and then a new app
3. Go to datasource page scroll to bottom 
4. We see sample data source 
5. Select movies from Mongo 
6. Now generate a CRUD page
7. The error toast is displayed as in attached screenshot

### Environment

Production

### Severity

Medium (Frustrating UX)

### Version

Cloud

Issue #41113, opened 2025-07-15
[Bug]: Unexpected error and cyclic dependency crops up in the following template app.


btsgh:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Description

On navigating to the Customers/Orders/Products/Returns page of the attached app json, we see a cyclic dependency as well as an unexpected error 

### Steps To Reproduce

Import the attached json
Navigate to any of these pages Customers/Orders/Products/Returns
Note the cyclic dependency / unexpected error popping up

[23AprReg.json](https://github.com/user-attachments/files/21226361/23AprReg.json)

https://github.com/user-attachments/assets/9a395748-6372-42fa-8ce7-10815af4ec54

### Environment

Production

### Severity

High (Blocker to building or releasing)

### Version

Cloud

PR #41144, opened 2025-07-29
fix: Updating logic for reactive actions to fix cyclic dependency issue with App templates and Generate page flow of a DB

ankitakinger:
## Description

Updating logic for reactive actions to fix cyclic dependency issue with App templates and Generate page flow of a DB. Currently, both flows were leading to cyclic dependency errors which shouldn't show up.
App template used - Order Fulfilment Tracker
DB used - Mock DB Movies

Fixes [#41125](https://github.com/appsmithorg/appsmith/issues/41125) [41113](https://github.com/appsmithorg/appsmith/issues/41113)

EE PR for tests: https://github.com/appsmithorg/appsmith-ee/pull/8029

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/16590882275>
> Commit: 907935727e70fac4dc1a8efe70cd8a3cd5da262b
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=16590882275&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Tue, 29 Jul 2025 09:46:35 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

* **Refactor**
  * Improved dependency detection logic for actions and JS actions, refining how data paths are identified and handled.
  * Unified data path detection by using a shared utility function across the application.
  * Enhanced filtering of entities during dependency calculations for greater accuracy.

* **Bug Fixes**
  * Corrected detection of reactive dependency misuse to reduce false positives for certain entity types.

### DIFF
--- a/app/client/src/ce/workers/common/DependencyMap/utils/getEntityDependenciesByType.ts
+++ b/app/client/src/ce/workers/common/DependencyMap/utils/getEntityDependenciesByType.ts
@@ -132,16 +132,6 @@ export function getJSDependencies(
     dependencies = { ...dependencies, [fullPropertyPath]: newDeps };
   }
 
-  for (const funcName of jsActionConfig.actionNames) {
-    const func = jsEntity[funcName];
-
-    if (func) {
-      dependencies[`${jsObjectName}.${funcName}.data`] = [
-        `${jsObjectName}.${funcName}`,
-      ];
-    }
-  }
-
   return dependencies;
 }
 export function getActionDependencies(
--- a/app/client/src/entities/DependencyMap/DependencyMapUtils.ts
+++ b/app/client/src/entities/DependencyMap/DependencyMapUtils.ts
@@ -5,6 +5,7 @@ import {
   getEntityNameAndPropertyPath,
   IMMEDIATE_PARENT_REGEX,
   isActionConfig,
+  isDataPath,
   isJSActionConfig,
 } from "ee/workers/Evaluation/evaluationUtils";
 import type { ConfigTree } from "entities/DataTree/dataTreeTypes";
@@ -167,10 +168,6 @@ export class DependencyMapUtils {
     return false;
   }
 
-  static isDataPath(path: string) {
-    return path.endsWith(".data");
-  }
-
   static detectReactiveDependencyMisuse(
     dependencyMap: DependencyMap,
     configTree: ConfigTree,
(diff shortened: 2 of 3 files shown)


app/client/src/ce/workers/common/DependencyMap/utils/getEntityDependenciesByType.ts
```diff
@@ -132,16 +132,6 @@ export function getJSDependencies(
     dependencies = { ...dependencies, [fullPropertyPath]: newDeps };
   }
 
-  for (const funcName of jsActionConfig.actionNames) {
-    const func = jsEntity[funcName];
-
```
ApekshaBhosale, line 137: why is this removed? @ankitakinger

app/client/src/entities/DependencyMap/DependencyMapUtils.ts
```diff
@@ -213,27 +211,34 @@ export class DependencyMapUtils {
...
-            new Error(
-              `Reactive dependency misuse: '${node}' depends on both trigger path '${runPath}' and data path '${dataPath}' from the same entity. This can cause unexpected reactivity.`,
-            ),
-            { node, triggerPath: runPath, dataPath },
-          );
+        const { entityName: depName } = getEntityNameAndPropertyPath(dep);
+        const entity = configTree[depName];
+
+        // to show cyclic dependency errors only for Action calls and not JSObject.body or JSObject
+        if (entity && entity.ENTITY_TYPE === "ACTION") {
```
ApekshaBhosale: use enum type. don't directly compare

app/client/src/entities/DependencyMap/DependencyMapUtils.ts
```diff
@@ -256,7 +261,15 @@ export class DependencyMapUtils {
...
       if (!entityConfig) return;
 
-      if (!isActionConfig(entityConfig) && !isJSActionConfig(entityConfig)) {
+      if (isWidget(entityConfig)) {
+        return;
+      }
+
+      // to not calculate transitive dependencies for JSObject.body and JSObject
+      if (
+        isJSActionConfig(entityConfig) &&
```
ApekshaBhosale, line 270: why this specific condition required?

app/client/src/ce/workers/common/DependencyMap/utils/getEntityDependenciesByType.ts
ankitakinger: Because this is what was causing error in the template. Without this, if I add `this.myFun1.data` in the `myFunc1` definition, it still throws a cyclic dependency error as expected.

ApekshaBhosale: this is common code which is used multiple places. all conditions use cases are fine?

app/client/src/entities/DependencyMap/DependencyMapUtils.ts
ankitakinger: Because `JSObject` as entity and `JSObject.body` as entity were also being calculated for transitive dependencies and we don't need those, we only need the dependencies for functions. This was leading to errors even when say `myFun1` is using `Query1.data` and `myFun2` is using `Query1.run`, which is incorrect as its not cyclic.

app/client/src/ce/workers/common/DependencyMap/utils/getEntityDependenciesByType.ts
ankitakinger: Yes all sample apps that we had created for testing, all are working as expected without this. And all existing unit and integration tests are also passing, as this was added by us in the reactive actions PR itself.

app/client/src/entities/DependencyMap/DependencyMapUtils.ts
```diff
@@ -215,7 +215,7 @@ export class DependencyMapUtils {
         const entity = configTree[depName];
 
         // to show cyclic dependency errors only for Action calls and not JSObject.body or JSObject
-        if (entity && entity.ENTITY_TYPE === "ACTION") {
+        if (entity && entity.ENTITY_TYPE) {
```
ApekshaBhosale, line 218: why are not comparing now? @ankitakinger you removed conditions itself

ankitakinger: Its not required. I tested again. JSObject.body doesn't have an ENTITY_TYPE for which it was failing, so just ENTITY_TYPE being present check is enough.

ApekshaBhosale: JSObject will have ENTITY_TYPE

check this https://github.com/appsmithorg/appsmith/blob/9930224212cc393b1d4ee854f9a507e10958d90e/app/client/src/ce/entities/DataTree/dataTreeJSAction.ts#L79

ankitakinger: Yeah I checked that part, but the issue that I was trying to fix with this check works well with just ENTITY_TYPE check as well. I just wanted to make sure this section runs for JSACTION, ACTION and on EE, on a safer side, for MODULE_INPUTS and MODULE_INSTANCE. 

I'll have to unnecessarily split the code and inspite of doing that JSObject will still have the ENTITY_TYPE JSACTION which means it will enter this block anyway.

PR #41174, opened 2025-08-07
Fix problem with scrolling the table when pagination is disabled

morzan1001:
# Description
This pull request fixes a issue with TableWidgetV2 where not all table rows were accessible via scrolling when pagination is disabled.

# Problem
When pagination is completely disabled (showPagination: false and serverSidePaginationEnabled: false), the number of rows displayed was incorrectly limited to the visible height of the container. This meant that in a table with, for example, 20 rows, only 5-6 rows could be displayed and scrolled, even though all data had been loaded.

# Cause
The `getPageSize` function in `derived.js` calculated the pageSize based on the container height, even if paging was not used. This limited the number of rendered rows to the calculated pageSize.

# Solution
The logic was adjusted to return the full number of table rows (`props.tableData.length`) when pagination was disabled. This allows all data to be rendered and scrolled through the table.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved table behavior to display all data in a single scrollable view when pagination is fully disabled.
* **Style**
  * Cleaned up formatting in multi-line comments for better readability.

### DIFF
--- a/app/client/src/widgets/TableWidgetV2/widget/derived.js
+++ b/app/client/src/widgets/TableWidgetV2/widget/derived.js
@@ -185,6 +185,12 @@ export default {
     const componentHeight = props.componentHeight - 10;
     const tableSizes = TABLE_SIZES[compactMode];
 
+    // If pagination is completely disabled (no page navigation)
+    // and server-side pagination is also off, display all data with scroll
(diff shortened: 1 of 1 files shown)


morzan1001:
Hey, before the pull request is closed due to inactivity, I wanted to ask what you think of my change?

PR #41183, opened 2025-08-20
fix: update java version in the base Docker image

vhemery:
Use the latest 17 jdk.

## Summary by CodeRabbit

* **Chores**
  * Updated build to obtain the latest JDK 17 from the Adoptium API instead of a hardcoded download.
  * Streamlined installation by piping the downloaded archive directly into extraction, simplifying the build steps.
  * Uses secure HTTPS API access and reduces maintenance for JVM updates.
  * No user-facing functionality changes.

### DIFF
--- a/deploy/docker/base.dockerfile
+++ b/deploy/docker/base.dockerfile
@@ -46,12 +46,8 @@ ENV PATH="/usr/lib/postgresql/14/bin:${PATH}"
 # Install Java
 RUN set -o xtrace \
   && mkdir -p /opt/java \
-  # Assets from https://github.com/adoptium/temurin17-binaries/releases
-  # TODO: The release jdk-17.0.9+9.1 doesn't include Linux binaries, so this fails.
-  #       Temporarily using hardcoded version in URL until we figure out a more elaborate/smarter solution.
-  #&& version="$(curl --write-out '%{redirect_url}' 'https://github.com/adoptium/temurin17-binaries/releases/latest' | sed 's,.*jdk-,,')" \
-  && version="17.0.9+9" \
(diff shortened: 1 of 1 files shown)


deploy/docker/base.dockerfile
```diff
@@ -46,12 +46,8 @@ ENV PATH="/usr/lib/postgresql/14/bin:${PATH}"
...
 RUN set -o xtrace \
   && mkdir -p /opt/java \
-  # Assets from https://github.com/adoptium/temurin17-binaries/releases
-  # TODO: The release jdk-17.0.9+9.1 doesn't include Linux binaries, so this fails.
-  #       Temporarily using hardcoded version in URL until we figure out a more elaborate/smarter solution.
-  #&& version="$(curl --write-out '%{redirect_url}' 'https://github.com/adoptium/temurin17-binaries/releases/latest' | sed 's,.*jdk-,,')" \
-  && version="17.0.9+9" \
-  && curl --location "https://github.com/adoptium/temurin17-binaries/releases/download/jdk-$version/OpenJDK17U-jdk_$(uname -m | sed s/x86_64/x64/)_linux_hotspot_$(echo $version | tr + _).tar.gz" \
+  # Assets from https://github.com/adoptium/temurin17-binaries/releases using the Adoptium API
+  && curl --location "https://api.adoptium.net/v3/binary/latest/17/ga/linux/$(uname -m | sed s/x86_64/x64/)/jdk/hotspot/normal/eclipse" \
```
subrata71, line 50: Just wondering: by using latest, won’t this make the build non-deterministic? Every rebuild could potentially pull in a different JDK version. Please correct me if I am wrong here.

vhemery: That's right. It looked like the intention in the previous comment.
The question is: would there be a valid case where you want to build the docker image with an old JDK ?

If you want so for reproducibility, then the JDK version should most probably be an argument of the docker image.
From this point of view, the JDK version should be an optional argument in the github workflow, so we can decide whether we want to use the latest JDK ('latest/17/ga') or a specific version (e.g. 'version/17.0.9+9')

PR #41207, opened 2025-09-04
chore: Added hooks to remove dangling locks

sondermanish:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags="@tag.Git"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/17473362736>
> Commit: 9bbf40be38011df0829473545833739e11d7b743
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=17473362736&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Git`
> Spec:
> <hr>Thu, 04 Sep 2025 19:26:09 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

- Bug Fixes
  - Improved reliability of Git-connected workflows by automatically cleaning up dangling Git lock/index files before key operations, reducing intermittent errors and stuck states across checkouts, branch create/delete, commits, status, discard, and branch listing.
- Chores
  - Made Git-in-memory detection more robust to avoid false positives when the Git root path is missing or contains whitespace.

### DIFF
--- a/app/server/appsmith-git/src/main/java/com/appsmith/git/configurations/GitServiceConfig.java
+++ b/app/server/appsmith-git/src/main/java/com/appsmith/git/configurations/GitServiceConfig.java
@@ -20,6 +20,9 @@ public class GitServiceConfig {
             return Boolean.FALSE;
         }
 
-        return gitRootPath.startsWith("/dev/shm/");
+        final String trimmedRootPath = gitRootPath.strip();
+        return "/dev/shm".equals(trimmedRootPath)
+                || trimmedRootPath.startsWith("/dev/shm/")
+                || trimmedRootPath.startsWith("/tmp/shm/");
     }
 }
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/git/central/GitHandlingServiceCE.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/git/central/GitHandlingServiceCE.java
@@ -121,5 +121,15 @@ public interface GitHandlingServiceCE {
     Mono<MergeStatusDTO> pullArtifact(
             ArtifactJsonTransformationDTO jsonTransformationDTO, GitArtifactMetadata baseMetadata);
 
+    /**
+     * Removes leftover Git lock and index files for the repository referenced by the provided DTO
+     * to unblock subsequent Git operations.
+     *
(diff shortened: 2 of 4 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/git/central/CentralGitServiceCEImpl.java
```diff
@@ -2156,8 +2165,10 @@ protected Mono<? extends Artifact> discardChanges(
         Mono<? extends Artifact> artifactFromLastCommitMono = Mono.usingWhen(
                         gitRedisUtils.acquireGitLock(artifactType, baseArtifactId, GitCommandConstants.DISCARD, TRUE),
                         ignoreLockAcquisition -> {
-                            Mono<? extends ArtifactExchangeJson> artifactJsonFromLastCommitMono = gitHandlingService
-                                    .recreateArtifactJsonFromLastCommit(jsonTransformationDTO)
+                            Mono<Boolean> removeDanglingLock =
+                                    gitHandlingService.removeDanglingLocks(jsonTransformationDTO);
```
subrata71, line 2169: Could this end up removing valid Git locks as well? For example, if a previous Git operation is still in progress and the user triggers another one, this code might delete the existing lock and acquire its own. That could lead to unexpected results since the earlier Git operation may still be running.
However, I might be overthinking but do you think the solution is foolproof?

sondermanish: @subrata71 the premise over here is that each gitlock is followed by a redis lock, and it's TTL is set to 120 seconds. In 2 minutes these operations should be completed. We shouldn't have any operation running for 2 or more minutes. Because, if that is the case probe will likely kill the pod (due to no more event loop threads able to serve probe requests).

So in this case when we are trying to remove the dangling locks, we are doing only when we have acquired redis locks. based on the premise established. we can be sure that the older operation has either crashed in b/w or has been running for more than 2 minutes. Both of them is unideal

subrata71 (approved):
Looks good to me. Please ensure thorough tests on this

PR #41216, opened 2025-09-10
fix: Adding `responseMeta` in query object even when the query fails

ankitakinger:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags="@tag.All"

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

### DIFF
--- a/app/client/src/workers/Evaluation/fns/actionFns.ts
+++ b/app/client/src/workers/Evaluation/fns/actionFns.ts
@@ -68,6 +68,27 @@ export default async function run(
      * */
     return response[0];
   } catch (e) {
+    const error = {
+      // TODO: Fix this the next time the file is edited
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      message: (e as any).message,
+    };
+
+    // If error contains responseData, update action data and responseMeta before throwing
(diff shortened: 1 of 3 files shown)


app/client/src/workers/Evaluation/fns/actionFns.ts
```diff
@@ -68,6 +68,27 @@ export default async function run(
...
+    // @ts-expect-error: responseData is a custom property
+    if (e.responseData && e.responseData.length > 0) {
+      // @ts-expect-error: self type is not defined
+      const action = self[this.name] as ActionEntity;
+      // @ts-expect-error: responseData is array format
+      const responseData = e.responseData;
+
+      if (action && responseData.length >= 3) {
+        action.data = responseData[0]; // error response body
+        action.responseMeta = responseData[2]; // { isExecutionSuccess, statusCode, headers }
```
rahulbarwal, line 87: I see that index 0 and 2 are utilized, what is in index 1?

ankitakinger: params. That is not used in the success flow as well.

PR #41377, opened 2025-11-13
feat: display workspace logos in sidebar navigation

salevine:
- Modified Workspace.java to return null for logoUrl when no logo exists
- Updated workspace reducer to sync logo changes to search results
- Enhanced WorkspaceMenuItem to display uploaded logos with fallback to default icon
- Added proper error handling and styling for workspace logo display

Fixes #`41376`  

## Automation

/ok-to-test tags="@tag.Workspace"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/20007050511>
> Commit: 3dc1661ea7013f4c4b98167d90a070f9591b905f
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=20007050511&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Workspace`
> Spec:
> <hr>Sun, 07 Dec 2025 16:47:49 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

* **New Features**
  * Workspace menu items and left navigation now show workspace logos when available, with a compact row layout, improved hover/selected styling, and graceful image fallback.
  * Search results items use the same logo-aware row for consistent navigation and click behavior.

* **Bug Fixes**
  * Missing or failed workspace logos no longer trigger unnecessary asset requests and fall back to text/icon display.
  * Search results stay synchronized when workspace details are updated.

### DIFF
--- a/app/client/src/ce/pages/Applications/index.tsx
+++ b/app/client/src/ce/pages/Applications/index.tsx
@@ -484,7 +484,8 @@ export function WorkspaceMenuItem({
 
   // Use custom component when there's a logo, otherwise use ListItem
   if (hasLogo && !isFetchingWorkspaces) {
-    return (
+    const showTooltip = workspace?.name && workspace.name.length > 22;
+    const itemContent = (
       <WorkspaceItemRow
         className={selected ? "selected-workspace" : ""}
         onClick={handleWorkspaceClick}
@@ -501,6 +502,14 @@ export function WorkspaceMenuItem({
           </Text>
         </WorkspaceIconContainer>
       </WorkspaceItemRow>
+    );
+
+    return showTooltip ? (
+      <Tooltip content={workspace?.name} placement="bottomLeft">
+        {itemContent}
+      </Tooltip>
+    ) : (
+      itemContent
     );
   }
 
--- a/app/client/src/pages/common/SearchBar/WorkspaceSearchItems.tsx
+++ b/app/client/src/pages/common/SearchBar/WorkspaceSearchItems.tsx
@@ -1,6 +1,6 @@
 import type { Workspace } from "ee/constants/workspaceConstants";
 import { Icon, Text } from "@appsmith/ads";
-import React from "react";
+import React, { useState } from "react";
 import { useHistory } from "react-router";
 import styled from "styled-components";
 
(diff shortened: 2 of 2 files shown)


app/client/src/ce/pages/Applications/index.tsx
```diff
@@ -414,8 +469,43 @@ export function WorkspaceMenuItem({
...
 
+  const handleImageError = () => {
+    setImageError(true);
+  };
+
   if (!workspace.id) return null;
 
+  const hasLogo = workspace?.logoUrl && !imageError;
+  const displayText = isFetchingWorkspaces
+    ? workspace?.name
```
rahulbarwal, line 480: If the workspaces are still being fetched, where would you obtain the workspace name? Wouldn't this be undefined in this case?

ankitakinger:
@salevine Can we have the same size for the icons? 16x16? Else it doesn't look good:
Screenshot 2025-11-20 at 12 43 36 PM

Secondly, I'm seeing a tooltip on hover of the workspace name only when there is an image and even though the name is not long enough to show the tooltip:
Screenshot 2025-11-20 at 12 46 34 PM

Also, with larger icons, the icon is not even visible in the list, any reason we find this change useful?

app/client/src/ce/pages/Applications/index.tsx
salevine: @rahulbarwal It should not be undefined. Based on everything I can see, the name comes from the getFetchedWorkspaces. Initially, the  workspaces array is empty and isFetchingWorkspaces is true. Nothing would show up. While refreshing, it may contain older data - but then when done fetching, it will refresh over it.  In the initial state, it all should be safe

salevine:
@ankitakinger  - I made a change to make the new icons 16x16 . I had them as 20x20. Not a problem.  If the user picks something that doesn't look good to them - then that is up to them to pick something that looks better. Pretty common experience across all sorts of platforms. Overall, yes this is a useful feature. Several customers have asked for it.  It was partially implemented before.  I went ahead and removed the tooltip that I had added.

ankitakinger (dismissed):

ankitakinger:
I have 3 more points:

1. Search results also show workspace names and logos. You'll have to update the logo there as well.
Screenshot 2025-12-05 at 2 11 56 PM

2. Tooltip is not showing on longer names of workspaces with custom logos.

3. See if we can use the existing components which only updates the logo and nothing else, because that will take care of the tooltip part.

salevine:
@ankitakinger  - I fixed the issues.
image
image

Issue #24973, opened 2023-06-30
[Feature]: Allow providing of page number when using server-side pagination


narhzih:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

For a table widget, when using server-side pagination, I'm not allowed to specify a page number my user can jump to (the way it works with client-side pagination). 

### Why should this be worked on?

If we have a lot of data, it can be very hard to go from page 1 - page 100. Users will have to click a lot of times before eventually reaching their default page. If there's a way to specify a page number, then users will be able to navigate easily between pages of the table.

ame-appsmith:
[Another user](https://app.frontapp.com/open/cnv_189nsu8c?key=CaElYmiKThpBiJVxEiY4DHjtwr7WHSp9) requested this feature.

Mace111:
+1

Nikhil-Nandagopal:
The challenge with storing the value in the query is that the table still shows the wrong number on page load

enriquenalda:
+1. My team are demanding to me.

PR #41421, opened 2025-11-26
feat: added setPageNumber function for Table v2 widget

tomjose92:
## Description
Added a setPageNumber function for Table widget to programatically change pages on the table.

This change allows the user to set the pageNo to any number value. What this means is that
1. For Client side pagination, the user will not be restricted from changing to a page number that is greater than total number of pages.
2. For Server side pagination, the user will have to programatically call the query to get data after programatically changing page number.

The setPageNumber function will throw error if any value that does not evaluate to a postive integer greater than 0 is passed.

Fixes [24973](https://github.com/appsmithorg/appsmith/issues/24973)
Fixes [8440](https://github.com/appsmithorg/appsmith-ee/issues/8440)

## Automation

/ok-to-test tags="@tag.Table"

> [!WARNING]
> Tests have not run on the HEAD d52ab88ff9e4c930bb330a275ca13daf9138d5d6 yet
> <hr>Thu, 27 Nov 2025 03:59:59 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

* **New Features**
  * Table widgets expose a programmatic page-number setter to change the current page.
  * Pagination now includes a configurable page-number property with numeric validation and default.

* **Improvements**
  * Server-side pagination better handles programmatic vs user-driven page changes and avoids duplicate triggers.
  * Page number display switched to an editable page-number input in the header.

* **Tests**
  * Tests updated to exercise and validate the new page-number setter behavior.

### DIFF
--- a/app/client/src/widgets/TableWidgetV2/component/header/actions/index.tsx
+++ b/app/client/src/widgets/TableWidgetV2/component/header/actions/index.tsx
@@ -217,26 +217,26 @@ function Actions() {
             {totalRecordsCount ? (
               <TableHeaderContentWrapper>
                 Page&nbsp;
-                <PaginationItemWrapper
+                <PageNumberInput
                   accentColor={accentColor}
                   borderRadius={borderRadius}
-                  className="page-item"
-                  selected
-                >
-                  {pageNo + 1}
-                </PaginationItemWrapper>
+                  disabled={pageCount === 1}
+                  pageCount={pageCount}
+                  pageNo={pageNo + 1}
+                  updatePageNo={updatePageNo}
+                />
                 &nbsp;
                 <span>{`of ${pageCount}`}</span>
               </TableHeaderContentWrapper>
             ) : (
-              <PaginationItemWrapper
+              <PageNumberInput
                 accentColor={accentColor}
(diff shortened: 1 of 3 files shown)


rahulbarwal:
https://github.com/user-attachments/assets/5d899c48-d214-4474-ad03-13f4033c394e

This is not production-ready; I see multiple bugs.

-  When I start, I have the input, but pressing backspace initially shows zero. I don't understand why this happens; it shouldn't occur.
-  When I type any page number, it changes to one and disables the input. No matter how much I navigate backward or forward, the input remains disabled.
-  The selected row functionality behaves oddly with next and previous.
	
	-  The selected row might be an existing bug and may not be relevant, but I want to be 100% sure about it.
	-  When pressing previous or next, a selected row is visible in the update row form during the loading state.
	-  Once the page loads, the selected row disappears.

tomjose92:
@rahulbarwal , regarding points 2 and 3, could you check if this happens with a newly created Table widget set up with server side pagination.
Reason is because I was not able to reproduce the above issue, and seems like it happens only when using it with a template or building block. Need to check why could that be happening

PR #41444, opened 2025-12-03
fix: allow helm chart to support 0 replicas

wyattwalter:
## Description

The `| default ` is interpreting `0` as falsey, so it's setting to 1 when `replicas` is set to the value `0`. Added a unit test to cover this use case as well.

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!WARNING]
> Tests have not run on the HEAD 3f096f237c0c21ef3f9bc5c5f8612c66fb434dd0 yet
> <hr>Wed, 03 Dec 2025 21:40:23 UTC

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

* **Chores**
  * Updated Helm chart version to 3.6.7

* **Behavior**
  * Deployment replica handling no longer falls back to an implicit default — the configured replica value is used as-is when autoscaling is disabled

* **Tests**
  * Added tests verifying default-replica behavior and explicit zero-replica scenarios when autoscaling is disabled

### DIFF
--- a/deploy/helm/tests/workload_type_test.yaml
+++ b/deploy/helm/tests/workload_type_test.yaml
@@ -6,7 +6,7 @@ tests:
       - equal:
           path: kind
           value: StatefulSet
-  - name: workload type should be Deployment when set
+  - name: workload type should be Deployment when set and replicas default to 1
     set:
       workload:
         kind: Deployment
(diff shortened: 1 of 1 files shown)


wyattwalter:
The Coderabbit comment about the default value is not a problem. `values.yaml` already defaults this value, and I would argue this double-defaulting was an anti-pattern before. Adding an explicit unittest for that.

Issue #41468, opened 2025-12-16
[Feature]: Tooltip input for checkbox / Radio inputs


ricardodcc:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Summary

I see that only normal inputs have the tooltip option available, but I think that also radio and checkboxs inputs that having a tooltip options is an advantage for displaying information to final website users

### Why should this be worked on?

I think there are a lot of situations that having a tooltip in checbox and radio box inputs is a plus for explaining to the website's users what those options means.

thx :)

amelia-c0n:
The Radio Group widget already has a [Tooltip option](https://docs.appsmith.com/reference/widgets/radio-group#tooltip-string) implemented.

PR #41483, opened 2025-12-23
feat: added Tooltip property for Checkbox widget

amelia-c0n:
## Description
Added a Tooltip option for the Checkbox widget that enables developers to add hints or provide additional information to guide the user regarding the required input.

Fixes [#41468](https://github.com/appsmithorg/appsmith/issues/41468)

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/21445878024>
> Commit: 2bdb89028b27cbf56613596be11b12f99cd93c27
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=21445878024&attempt=3" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Wed, 28 Jan 2026 17:03:05 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

* **New Features**
  * Added label tooltip support for checkboxes and enabled tooltip display on checkbox labels.
  * Enhanced checkbox label alignment and layout controls (left/right alignment, wrapping, justification) and richer styling options (color, size, inline/dynamic height).

* **Tests**
  * Updated UI tests and test locators to reliably target checkbox labels after rendering changes.

### DIFF
--- a/app/client/cypress/e2e/Regression/ClientSide/Widgets/Input/Input_spec.js
+++ b/app/client/cypress/e2e/Regression/ClientSide/Widgets/Input/Input_spec.js
@@ -142,9 +142,9 @@ describe(
 
     it("7. Input label renders if label prop is not empty", () => {
       //Input label wrapper do not show if lable and tooltip is empty
-      cy.get(
-        `${publish.inputWidget} [data-testid='label-container']`,
-      ).should("not.exist");
+      cy.get(`${publish.inputWidget} [data-testid='label-container']`).should(
+        "not.exist",
+      );
 
       cy.openPropertyPane("inputwidgetv2");
       // enter label in property pan


.gitignore
```diff
@@ -1,3 +1,5 @@
+specs/quick-fixes/**
```
tomjose92: @amelia-c0n , can we remove this from .gitignore. We do not want fix/feature specific cursor files in the gitignore

app/client/src/widgets/CheckboxWidget/component/index.tsx
```diff
@@ -120,18 +144,26 @@ class CheckboxComponent extends React.Component<CheckboxComponentProps> {
           hasError={!isValid}
           inputRef={this.props.inputRef}
           labelElement={
-            <CheckboxLabel
-              alignment={this.props.alignWidget || AlignWidgetTypes.LEFT}
+            <LabelWithTooltip
+              alignment={
+                this.props.alignWidget === AlignWidgetTypes.RIGHT
```
tomjose92: Lets keep this simple by passing this.props.alignWidget || AlignWidgetTypes.LEFT.

app/client/src/widgets/CheckboxWidget/component/index.tsx
```diff
@@ -33,6 +37,25 @@ const CheckboxContainer = styled.div<StyledCheckboxContainerProps>`
...
+
+      label {
+        text-align: ${({ alignWidget }) =>
+          alignWidget === AlignWidgetTypes.RIGHT ? "right" : "left"};
+        white-space: normal;
+        overflow: auto;
+        text-overflow: unset;
+        word-break: break-word;
+      }
+    }
```
ashit-rath, line 61: Why are we adding more styles to label container? I believe we pass alignment prop to the <LabelWithTooltip> so as to adjust according to alignment settings

Also the style properties seems a bit off. Why do we need `max-width` `width` and `flex` in the same component with all of them maxed out to full width?

app/client/src/widgets/CheckboxWidget/component/index.tsx
```diff
@@ -120,18 +144,26 @@ class CheckboxComponent extends React.Component<CheckboxComponentProps> {
           hasError={!isValid}
           inputRef={this.props.inputRef}
           labelElement={
-            <CheckboxLabel
```
ashit-rath, line 123: Since CheckboxLabel is not used, can we remove the code for it?

app/client/src/widgets/CheckboxWidget/component/index.tsx
amelia-c0n: The reasons I added the additional styles for the label container are:
- `LabelWithTooltip` doesn’t apply alignment for the **Position = Right** case. The label stays left-aligned for the scenario **Position = Right + Alignment = Right**. To fix this, I introduced the `justify-content` and `label text-align`.
- Overrides such as `max-width` and `flex` are layout tweaks to ensure the label aligns to the edge of the widget with no offsets (e.g., if **Position = Left + Alignment = Left**). However, I can remove the `width` and `margin-right`, since they are redundant.
- Overrides such as `overflow` ensure the behavior matches the previous behavior of the Checkbox widget for cases such as Fixed Height + long labels (the label wraps, no ellipsis).
If we skip the LABEL_CONTAINER_CLASS overrides altogether, this is the behavior we get: [screen recording](https://app.screencast.com/VY6l3qlV7As76).
Let me know your thoughts.

app/client/src/widgets/CheckboxWidget/component/index.tsx
amelia-c0n: The `CheckboxLabel` is still used by the `CheckboxGroupWidget` for the labels of the individual options and for the Select All option, so removing it would break that dependency.

ashit-rath (approved):

PR #41506, opened 2026-01-08
fix: use appsmith-built mongodb image by default in helm chart

wyattwalter:
This PR updates the default image for a MongoDB cluster managed by the chart to use an image that has been patched for the recent MongoBleed vulnerability, since Bitnami stopped publishing updates to their images.

While not ideal, I made a few time-boxed attempts after the deprecation to use a non-Bitnami image while maintaining compatibility with the chart, but the coupling is too tight to patch over reliably. Ultimately, we need to remove the Bitnami chart from our stack, but that carries too much risk for a simple and critical security patch.

Since the Bitnami images are licensed under Apache 2.0, I’ve forked their builder and published a set of images for each major MongoDB version with the MongoBleed patch included: 6.0.27, 7.0.28, and 8.0.17. I’ve tested deployments using each of these versions, including an upgrade scenario. If anyone reading this decides to upgrade to 7.x or 8.x, please be sure to follow the upstream MongoDB documentation about feature compatibility [here](https://www.mongodb.com/docs/manual/reference/command/setFeatureCompatibilityVersion/), as we do not yet have an Appsmith-specific MongoDB upgrade guide and this is not automatically handled by MongoDB.

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

* **Chores**
  * Updated MongoDB to version 6.0.27
  * Bumped Helm chart version to 3.6.8

### DIFF
--- a/deploy/helm/Chart.yaml
+++ b/deploy/helm/Chart.yaml
@@ -11,7 +11,7 @@ sources:
   - https://github.com/appsmithorg/appsmith
 home: https://www.appsmith.com/
 icon: https://assets.appsmith.com/appsmith-icon.png
-version: 3.6.7
+version: 3.6.8
 dependencies:
   - condition: redis.enabled
     name: redis
@@ -21,7 +21,7 @@ dependencies:
   - condition: mongodb.enabled
     name: mongodb
     version: 12.1.16
-    appVersion: 6.0.10
+    appVersion: 6.0.27
     repository: https://charts.bitnami.com/bitnami
   - condition: postgresql.enabled
     name: postgresql
--- a/deploy/helm/values.yaml
+++ b/deploy/helm/values.yaml
@@ -34,8 +34,8 @@ mongodb:
   tolerations: []
   image:
     registry: docker.io
-    repository: bitnamilegacy/mongodb
-    tag: 6.0.13
+    repository: appsmith/mongodb
+    tag: 6.0.27
   arbiter:
     nodeSelector: {}
     affinity: {}
--- a/deploy/helm/tests/__snapshot__/defaults_snapshot_test.yaml.snap
+++ b/deploy/helm/tests/__snapshot__/defaults_snapshot_test.yaml.snap
@@ -25,7 +25,7 @@
         app.kubernetes.io/instance: RELEASE-NAME
         app.kubernetes.io/managed-by: Helm
         app.kubernetes.io/name: appsmith
-        appsmith.sh/chart: appsmith-3.6.7
+        appsmith.sh/chart: appsmith-3.6.8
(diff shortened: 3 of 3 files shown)


deploy/helm/Chart.yaml
```diff
@@ -21,7 +21,7 @@ dependencies:
   - condition: mongodb.enabled
     name: mongodb
     version: 12.1.16
-    appVersion: 6.0.10
+    appVersion: 6.0.27
     repository: https://charts.bitnami.com/bitnami
```
subrata71, line 25: Can you please help me understand why are we still relying on bitnami for chart dependencies?

deploy/helm/values.yaml
```diff
@@ -34,8 +34,8 @@ mongodb:
   tolerations: []
   image:
     registry: docker.io
-    repository: bitnamilegacy/mongodb
-    tag: 6.0.13
+    repository: appsmith/mongodb
+    tag: 6.0.27
```
subrata71, line 38: Thanks for the detailed context in the description. One clarification I wanted to check: are the MongoDB 7.x and 8.x images you mentioned (7.0.28, 8.0.17) published under the official appsmith/mongodb namespace and intended to be supported long term, or were they built mainly for validation and testing purposes?

This would help clarify whether users can safely start new installations on 7.x or 8.x using Appsmith maintained images, or if 6.0.x is the only supported default for now.

wyattwalter: Appsmith works with MongoDB 6, 7, or 8. Some customers, on their own, upgraded to 7 and 8 using the Bitnami image so I wanted to give them a path in the short term. The change was basically the same and it was very low effort to relieve the pain, so I did it.

I would like to start pushing installs to version 7 and then 8 (you can't leap major versions, which makes this more complicated), but the Bitnami images/chart do not automatically upgrade the compatibility version so it's difficult to do with the existing tooling.

I'd like to be able to offer a patch for this vulnerability without making those kinds of changes just yet.

deploy/helm/Chart.yaml
wyattwalter: Switching off the Bitnami chart is the most difficult step here. They offered a well-supported chart that supported replicasets and made it easy, then the free offering was removed. The MongoDB chart for replicasets has been deprecated in favor of a Kubernetes operator-based system that's a much larger and complicated change. There are other charts that have popped up recently, but have varying levels of support/popularity.

We will likely migrate to one of those, but there will likely be changes incompatible with the Bitnami chart that we'll need to account for and assist with the migration. This is not the moment for that.

subrata71 (approved):

PR #41515, opened 2026-01-14
feat: extraVolumes and extraVolumeMounts options in Helm chart

wyattwalter:
Adds a feature to allow extra volumes and volume mounts to be defined for the Appsmith Deployment/StatefulSet object.

Fixes https://linear.app/appsmith/issue/APP-14833/add-support-for-extravolumes-and-extravolumemounts-in-helm-chart

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

* **New Features**
  * Added optional configuration for additional pod volumes and volume mounts.

* **Tests**
  * Added validation tests for extra volumes and mounts in Deployment and StatefulSet configurations.

### DIFF
--- a/deploy/helm/values.yaml
+++ b/deploy/helm/values.yaml
@@ -162,6 +162,14 @@ securityContext: {}
   # runAsNonRoot: true
   # runAsUser: 1000
 
+## @param extraVolumes Additional volumes to add to the pod
+##
+extraVolumes: []
+
+## @param extraVolumeMounts Additional volume mounts to add to the appsmith container
+##
+extraVolumeMounts: []
+
 service:
   ## @param service.type Kubernetes Service type
   ##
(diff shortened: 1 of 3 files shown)


deploy/helm/values.yaml
```diff
@@ -162,6 +162,14 @@ securityContext: {}
   # runAsNonRoot: true
   # runAsUser: 1000
 
+## @param extraVolumes Additional volumes to add to the pod
+##
+extraVolumes: []
+
+## @param extraVolumeMounts Additional volume mounts to add to the appsmith container
+##
+extraVolumeMounts: []
```
subrata71, line 171: Just validating my understanding here. Initially, these values are empty. For our case especially for prod they should be populated from the prod `values.yaml` in the cloud-deployment repo. Is that correct?

wyattwalter: Yup you got it, the default is not set / empty but this allows it to be configurable.

PR #41532, opened 2026-01-27
feat: Add Betterbugs SDK support

sebastianiv21:
This PR integrates [Betterbugs](https://www.betterbugs.io/) bug reporting SDK into the Appsmith client application, providing users with an enhanced bug reporting experience directly within the application interface.

Motivation and Context
Currently, users need to manually report bugs through external channels, which can be cumbersome and may result in incomplete bug reports. By integrating Betterbugs, we enable:
- Streamlined bug reporting: Users can report bugs directly from the application
- Rich context capture: Automatically captures session metadata, user information, and application state
- Better debugging: Provides development team with comprehensive bug reports including screenshots, console logs, and user actions
- Improved user experience: Reduces friction in the bug reporting process

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/21915839106>
> Commit: f70f56393465e394dc03d3e868c16d0368e60793
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=21915839106&attempt=5" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Fri, 13 Feb 2026 01:50:46 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

* **New Features**
  * Adds a "Send support info" option in help menus and homepage header (visible when enabled) to let users send diagnostic/support info.

* **Chores**
  * Integrates a third‑party support widget and SDK, adds a feature‑flagged configuration with an API key toggle, wires build/deploy and runtime settings, and ensures widget metadata updates automatically on navigation.

### DIFF
--- a/app/client/src/RouteChangeListener.tsx
+++ b/app/client/src/RouteChangeListener.tsx
@@ -3,7 +3,6 @@ import { useEffect, useRef } from "react";
 import { useDispatch } from "react-redux";
 import { useLocation } from "react-router";
 import type { AppsmithLocationState } from "utils/history";
-import BetterbugsUtil from "utils/Analytics/betterbugs";
 
 export default function RouteChangeListener() {
   const location = useLocation<AppsmithLocationState>();
@@ -14,7 +13,6 @@ export default function RouteChangeListener() {
     const prevLocation = prevLocationRef;
 
     dispatch(routeChanged(location, prevLocation.current));
-    BetterbugsUtil.updateMetadata();
     prevLocation.current = location;
   }, [location.pathname, location.hash]);
 


tomjose92:
@sebastianiv21 , we need to have an option to allow the user to disable betterbugs. This is in scenarios where its airgapped, or the user has hosted in an isolated environment and does not want to outgoing traffic to betterbugs.
Or wouldnt whitelist it. It will throw up a lot of console errors or failed network

app/client/src/utils/Analytics/betterbugs.ts
```diff
@@ -0,0 +1,119 @@
...
+        mode:
+          process.env.NODE_ENV === "development" ? "development" : "production",
+        ...(user?.email ? { email: user.email } : {}),
+        showActionButton: false,
+        styles: this.getDefaultStyles(),
+        position: { bottom: "30px", right: "20px" },
+        metaData: {
+          ...(user?.email ? { user_email: user.email } : {}),
+          ...(user?.name ? { user_name: user.name } : {}),
+          instance_id: getInstanceId(store.getState()),
```
tomjose92: Betterbugs also has a function to update metadata. Its called setMetadata. 
We can use that to capture additional information like url path, or applicationid, etc

app/client/src/ce/configs/index.ts
```diff
@@ -172,6 +178,11 @@ export const getAppsmithConfigs = (): AppsmithUIConfigs => {
     APPSMITH_FEATURE_CONFIGS?.smartLook.id,
   );
 
+  const betterbugs = getConfig(
+    ENV_CONFIG.betterbugs.apiKey,
```
tomjose92, line 182: Should have a way to control this from env configuration, so that it can be disabled.
Or test and see what happens if we set it up an environment that does not have outgoing access. Would it keep throwing errors.

.github/workflows/ad-hoc-docker-image.yml
```diff
@@ -121,6 +121,7 @@ jobs:
           platforms: linux/arm64,linux/amd64
           build-args: |
             APPSMITH_SEGMENT_CE_KEY=${{ secrets.APPSMITH_SEGMENT_CE_KEY }}
+            APPSMITH_BETTERBUGS_API_KEY=${{ secrets.APPSMITH_BETTERBUGS_API_KEY }}
```
subrata71, line 124: Reminder: The secret would come from cloud-deployment repo from the respective environment's source.

subrata71: That means you also need to create a PR in cloud-deployment repo.

app/client/src/utils/Analytics/betterbugs.ts
```diff
@@ -0,0 +1,198 @@
...
+  public static async show(user?: User) {
+    if (isAirgapped() || !getAppsmithConfigs().betterbugs.enabled) {
+      log.warn("BetterBugs is disabled.");
+
+      return;
+    }
+
+    // Destroy existing instance to clear previous state
+    if (this.instance) {
+      this.destroy();
```
subrata71, line 151: What are all the scenarios when we clear the previous state? I mean what are all those trigger points?

app/client/src/utils/Analytics/betterbugs.ts
```diff
@@ -0,0 +1,198 @@
...
+      return {
+        url_path: null,
+        application_id: null,
+        page_id: null,
+      };
+    }
+
+    const state = store.getState();
+    const applicationId = getCurrentApplicationId(state);
+    const pageId = getCurrentPageId(state);
```
subrata71, line 84: Are we only capturing runtime metadata for application? What about packages and workflows? Do you plan to have overridden method in EE?

.github/workflows/ad-hoc-docker-image.yml
sebastianiv21: @subrata71 I've tagged you on the cloud-deployment PR

app/client/src/utils/Analytics/betterbugs.ts
sebastianiv21: We clear the previous state whenever a user clicks “Send support info” from either the Editor help menu or the Homepage header help menu. That ensures we re‑initialize the widget with fresh metadata from the current user context

app/client/src/pages/Editor/HelpButton.tsx
```diff
@@ -292,6 +303,11 @@ function HelpButton() {
                         }
                       }
                     }
+
+                    if (item.id === "betterbugs-trigger") {
```
ashit-rath, line 307: @subrata71 Right now this `HelpButton` is present in App and Workflows. Should be add it to packages as well?

app/client/src/utils/Analytics/betterbugs.ts
ashit-rath, line 84: What about workflows and packages?

app/client/src/utils/Analytics/betterbugs.ts
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    if (!apiKey) {
+      log.warn("BetterBugs API key not configured.");
+
+      return;
+    }
+
+    try {
+      const { default: Betterbugs } = await import("@betterbugs/web-sdk");
```
ashit-rath, line 121: Why are we importing it like this?

app/client/src/utils/Analytics/betterbugs.ts
```diff
@@ -0,0 +1,198 @@
...
+  destroy: () => void;
+  show: () => void;
+  hide: () => void;
+  openWidget: () => void;
+  closeWidget: () => void;
+  getMetadata?: () => Record<string | number, string | number | null>;
+  setMetadata?: (
+    metaData: Record<string | number, string | number | null>,
+  ) => void;
+}
```
ashit-rath: Instead of us defining the interface for the instance check if Betterbugs have exported an interface that we can use. Ask Cursor it might find it

app/client/src/utils/Analytics/betterbugs.ts
```diff
@@ -0,0 +1,198 @@
...
+      return;
+    }
+
+    try {
+      const { default: Betterbugs } = await import("@betterbugs/web-sdk");
+      const instance = new Betterbugs({
+        apiKey,
+        mode:
+          process.env.NODE_ENV === "development" ? "development" : "production",
+        ...(user?.email ? { email: user.email } : {}),
```
ashit-rath: This is not a good way to set optional variable to an object. Let's create a new function called getDefaultParams(user) and then in that function let's manipulate the object. 

The way to set this object would be
1. Define all the constant params 
```
const params = {
       showActionButton: false,
        styles: this.getDefaultStyles(),
        mainHeading: "Send support info",
        recordType: "recordVideo",
        position: { bottom: "30px", right: "20px" },
        successMessageHeaderText: "Information received",
        ...
}
```
2. Then add optionals
```
   if (email) {
      params.email = user.email
   }
```

app/client/src/RouteChangeListener.tsx
```diff
@@ -13,6 +14,7 @@ export default function RouteChangeListener() {
     const prevLocation = prevLocationRef;
 
     dispatch(routeChanged(location, prevLocation.current));
+    BetterbugsUtil.updateMetadata();
```
ashit-rath: Why do we need to keep capturing the appid and pageid? Shouldn't we just capture this when someone opens the better bug interface?

app/client/src/pages/Editor/HelpButton.tsx
subrata71: Yes, we should otherwise it will just remain as half-baked solution.

app/client/src/utils/Analytics/betterbugs.ts
sebastianiv21: Not in this version; we’re currently capturing runtime metadata only for applications. Packages and workflows aren’t included yet, but this is something we can add in upcoming versions.

sebastianiv21: Not in this version; we’re currently capturing runtime metadata only for applications. Packages and workflows aren’t included yet, but this is something we can add in upcoming versions.

app/client/src/utils/Analytics/betterbugs.ts
sebastianiv21: We’re using a dynamic import so the SDK is loaded only on the client when we initialize BetterBugs. This follows BetterBugs’ React guidance as well; they show the same pattern in their docs: [React app example](https://docs.betterbugs.io/developer-guide/betterbugs-web-sdk/installation/react-app-example).

app/client/src/ce/configs/index.ts
sebastianiv21: Users can disable the feature using the `APPSMITH_DISABLE_BETTERBUGS` env variable

app/client/src/RouteChangeListener.tsx
sebastianiv21: That makes sense. I've removed this logic.

PR #41534, opened 2026-02-02
fix: update mongodb fcv to 6

wyattwalter:
## Description

Looks like when the embedded MongoDB version was last updated, the feature compatibility version parameter was not updated. This means that any instances that started on 1.66 or less using the embedded MongoDB didn't have their compatibility version updated. This'll make the jump to 7.x harder as we need to ensure this is at least 6 in those scenarios. Getting this in place now.

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

* **Chores**
  * Updated MongoDB feature compatibility from version 5.x to 6.x in deployment configuration.

### DIFF
--- a/deploy/docker/fs/opt/appsmith/mongodb-fixer.sh
+++ b/deploy/docker/fs/opt/appsmith/mongodb-fixer.sh
@@ -17,10 +17,10 @@ tlog "MongoDB is RUNNING"
 
 for _ in {1..60}; do
   if mongosh --quiet "$APPSMITH_DB_URL" --eval '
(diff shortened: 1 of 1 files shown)


subrata71 (approved):

wyattwalter:
It'll just set the compatibility level to 6 (we're running 6 now) so that we can hop to 7. It does mean that you can't start up an older version of MongoDB using the data directory. But that would also not work once Appsmith starts anyway, so it doesn't matter.

PR #41593, opened 2026-03-06
feat: allow customization of HTML lang attribute 

salevine:
Users building apps in non-English languages (e.g. German) experience corrupted UI text because browsers and translation extensions see <html lang="en"> and auto-translate page content — sometimes German-to-German, sometimes misinterpreting English UI words like "Content" as the French adjective (translated to "Thrilled").
This PR fixes the problem at three layers:
Immediate defence: Adds translate="no" and <meta name="google" content="notranslate"> to the app shell HTML, editor UI containers (PropertyPane tabs, App Settings), and widget iframe templates. This stops browsers from offering or performing auto-translation.
Per-app language setting: Adds an "HTML Language" field to App Settings > General. App builders can enter a BCP 47 language code (e.g. de, fr, ja) that gets persisted as htmlLang on ApplicationDetail and applied to the published app's <html lang=""> attribute at runtime via react-helmet.
Instance-level default: Adds APPSMITH_DEFAULT_HTML_LANG environment variable for self-hosted admins who want all apps on their instance to default to a specific language. Falls back gracefully to "en" when the variable is not set.

Fixes #`6642`  

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/29444417195>
> Commit: 2e8764c9179ac69f75107669f45682badce8b816
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=29444417195&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Wed, 15 Jul 2026 20:40:33 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

* **New Features**
  * Added an **“HTML language”** setting to General Settings (BCP 47), with validation, persistence, and an environment-configurable default.

* **Behavior**
  * Published apps now apply the chosen (or default) HTML `lang` to the document metadata.
  * App pages, loading screens, and embedded/iframe content now opt out of automatic translation via `translate="no"` and Google `notranslate` where applicable.

* **Tests**
  * Added unit tests covering HTML language input validation and normalization.


app/client/public/index.html
```diff
@@ -1,7 +1,8 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" translate="no">
   <head>
     <meta charset="utf-8" />
+    <meta name="google" content="notranslate" />
```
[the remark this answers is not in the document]
salevine: Resolved

subrata71 (commented):
Review comments from Claude.

---

**`app/client/public/index.html` — fallback gap in the `<html lang>` attribute**

The `<html>` tag on line 2 uses:
```html
<html lang='{{env "APPSMITH_DEFAULT_HTML_LANG"}}' translate="no">
```

The template substitution in `caddy-reconfigure.mjs` replaces `{{env "APPSMITH_DEFAULT_HTML_LANG"}}` via:
```js
(_, name) => (process.env[name] || extraEnv[name] || "")
```

Since `extraEnv.APPSMITH_DEFAULT_HTML_LANG` is set to `process.env.APPSMITH_DEFAULT_HTML_LANG || "en"`, the `extraEnv` fallback should kick in when the env var isn't set — so in a **Docker-based deployment**, the rendered HTML will correctly get `lang="en"`. That part is fine.

However, in **non-Docker environments** (e.g. running the client dev server directly, or cloud deployments where `caddy-reconfigure` isn't the one serving `index.html`), the Caddy template expression may not be processed at all, which would leave `lang=''` verbatim in the HTML, or the browser receiving the unprocessed template string. It would be safer to set an explicit fallback at the template level itself. If Caddy's template syntax supports a default, something like:
```
{{env "APPSMITH_DEFAULT_HTML_LANG" | default "en"}}
```
would make the fallback self-contained and not dependent on the `extraEnv` JavaScript logic.

Similarly, in the JavaScript config injection block (line ~272), `parseConfig('{{env "APPSMITH_DEFAULT_HTML_LANG"}}') || ""` will resolve to `""` rather than `"en"` when the env var isn't set and the template isn't processed. The triple fallback in `AppViewerHtmlTitle` (`lang || defaultHtmlLang || "en"`) does save you at runtime, but it means the config carries an empty string that has to be silently corrected downstream.

---

**`AppViewerHtmlTitle.tsx` — module-level `getAppsmithConfigs()` call**

```ts
const { defaultHtmlLang } = getAppsmithConfigs();
```

This is called at module load time, outside any React component or hook. If `getAppsmithConfigs()` reads from a mutable config store (which it does — it merges `ENV_CONFIG` and `APPSMITH_FEATURE_CONFIGS`), this value will be captured once when the module is first imported and won't reflect any subsequent config updates. For an env-var-backed config that doesn't change at runtime this is probably fine in practice, but it's inconsistent with how the rest of the codebase typically calls `getAppsmithConfigs()` inside the component or a hook. Worth moving it inside the component function to be consistent and future-safe.

---

**`GeneralSettings.tsx` — no validation before saving**

The `saveHtmlLang` callback does a `trimmed === current` guard which is good, but there's no check on whether the trimmed value is a plausible BCP 47 code before dispatching the update. A user can type something like `"not a language"` or paste a long string with spaces, and it will be persisted and then applied to the `<html lang>` attribute. Invalid `lang` values don't break anything today, but they degrade accessibility tool behavior and are meaningless for the feature's stated goal.

Even a lightweight check — ensuring the value matches something like `/^[a-zA-Z]{2,3}(-[a-zA-Z0-9]+)*$/` before saving, with a small inline error message — would prevent accidental invalid saves. The `onBlur`/Enter flow means a user can type and walk away without realizing the input was invalid.

---

**`PropertyPaneTab.tsx` — `translate="no"` on tabs**

This adds `translate="no"` to the `StyledTabs` wrapper to prevent auto-translation of tab labels like "Content" and "Style". That's a targeted and reasonable fix for the editor. One thing worth confirming: does `StyledTabs` (from `@appsmith/ads`) forward arbitrary HTML attributes to the DOM, or could this silently get swallowed? If it's a styled-component or a custom component that doesn't spread props to the underlying DOM element, the attribute won't reach the browser and the protection won't work.

---

**`ApplicationDetailCE.java` — no server-side validation**

`htmlLang` is stored as a plain `String` with no length limit or format constraint. Since this flows from user input all the way into the `<html lang>` attribute of the served page, it's worth having at least a basic server-side sanitization (trim, max length, reject obviously invalid values) even if client-side validation is added too. Defense in depth applies here.

---

**Overall**

The feature direction is good and the code is generally clean. The main things I'd want addressed before merge are the `index.html` fallback gap (correctness risk for non-Docker deployments) and the missing input validation on the lang setting (both client and server side). The module-level config call in `AppViewerHtmlTitle` is a minor consistency issue.

app/client/public/index.html
```diff
@@ -1,7 +1,8 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang='{{env "APPSMITH_DEFAULT_HTML_LANG"}}' translate="no">
```
subrata71: The <html lang='{{env "APPSMITH_DEFAULT_HTML_LANG"}}'> template has no inline fallback value. The caddy-reconfigure.mjs provides the "en" default only during Docker deployment. Tagging @ashit-rath to validate if this concern is not valid.

ashit-rath: Let's add an inline fallback value if we can. Better to be safe if the option is available

subrata71 (approved):
The changes look good to me. Could you please verify the EE changes by running all CI checks and validating it manually both in App and UI package?

PR #41596, opened 2026-03-07
fix: use atomic setIfAbsent for Redis lock acquisition in RedisUtils

subrata71:
## Description

The `addFileLock` and `startAutoCommit` methods in `RedisUtils` used a non-atomic `hasKey` + `set` pattern that is susceptible to race conditions. Between the `hasKey` check and the `set` call, another concurrent caller could observe the key as absent and also proceed to acquire the lock, resulting in two callers believing they hold the lock simultaneously.

This replaces the two-step pattern with Redis's atomic `setIfAbsent` (SET NX) which checks and sets in a single atomic operation, consistent with how `GitRouteAspectCE` and `DistributedLockAspect` already handle lock acquisition in this codebase.

**Methods fixed:**
- `addFileLock(String key, String gitCommand)` — also fixes a pre-existing bug where the raw `gitCommand` parameter was stored instead of the normalized `command` value
- `addFileLock(String key, Duration, AppsmithException)` (deprecated)
- `startAutoCommit(String defaultApplicationId, String branchName)`

The method signatures and return types are unchanged, so all callers and mocked tests remain compatible.

## Automation

/ok-to-test tags=""

## Communication
- [ ] Yes
- [x] No

## Release Notes

* **Bug Fixes**
  * Improved file lock handling for more reliable concurrent file operations and clearer error messages when files are in use.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/RedisUtils.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/RedisUtils.java
@@ -37,16 +37,19 @@ public class RedisUtils {
 
     public Mono<Boolean> addFileLock(String key, String gitCommand) {
         String command = hasText(gitCommand) ? gitCommand : REDIS_FILE_LOCK_VALUE;
-        return redisOperations.hasKey(key).flatMap(isKeyPresent -> {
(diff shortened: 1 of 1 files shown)


subrata71:
Closing this as it may cause regression. We will revisit this once we start getting reports regarding dangling Git locks

PR #41598, opened 2026-03-09
fix(security): prevent unauthenticated disclosure of instance metadata [APP-14994]

subrata71:
## Summary

Fixes APP-14994: three API endpoints were accessible without authentication and exposed sensitive instance metadata to any unauthenticated caller.

- **`/api/v1/users/features`** — removed from `permitAll()` in `SecurityConfig`; now returns **401** for unauthenticated requests. Feature flags are per-user data and must not be disclosed to anonymous callers.
- **`/api/v1/tenants/current`** — kept accessible without authentication (the login page uses it to discover enabled auth providers), but `instanceId` and `adminEmailDomainHash` are now **suppressed** from the response when the caller is anonymous. This eliminates exposure of the unsalted SHA-256 admin-email-domain hash and the instance UUID.
- **`/api/v1/consolidated-api/view`** — kept accessible without authentication (required for public/published apps), but the embedded `organizationConfig` no longer contains `instanceId` or `adminEmailDomainHash` for anonymous callers (same service-layer fix as above).

## Changes

| File | Change |
|------|--------|
| `SecurityConfig.java` | Remove `GET /api/v1/users/features` from `permitAll()` |
| `OrganizationServiceCEImpl.java` | Check `ReactiveSecurityContextHolder` in `getOrganizationConfiguration()`; skip setting `instanceId` / `adminEmailDomainHash` for anonymous principals |
| `OrganizationServiceCETest.java` | Add two tests: anonymous caller gets `null` for sensitive fields; authenticated caller gets non-null values |
| `AuthGuardTest.java` | New controller-level test: `GET /api/v1/users/features` returns 401 for unauthenticated requests; `GET /api/v1/tenants/current` remains accessible |

## Test plan

- [x] `OrganizationServiceCETest#getOrganizationConfig_AnonymousUser_DoesNotExposeInstanceMetadata` — asserts `instanceId` and `adminEmailDomainHash` are `null` for anonymous callers
- [x] `OrganizationServiceCETest#getOrganizationConfig_AuthenticatedUser_ExposesInstanceMetadata` — asserts both fields are non-null for authenticated callers
- [x] `AuthGuardTest#featureFlagsEndpoint_unauthenticated_returns401` — HTTP 401 for unauthenticated `GET /api/v1/users/features`
- [x] `AuthGuardTest#tenantCurrentEndpoint_unauthenticated_isAccessible` — `GET /api/v1/tenants/current` still returns 200 without auth
- [x] All 15 existing `OrganizationServiceCETest` tests pass (1 pre-existing disabled test skipped)

## Summary by CodeRabbit

* **Bug Fixes**
  * The user features endpoint now requires authentication.
  * Organization instance metadata (instanceId, adminEmailDomainHash) is concealed for unauthenticated requests; tenant/current remains accessible but without sensitive fields.

* **Tests**
  * Added tests validating authentication for the features endpoint and that anonymous vs. authenticated users receive appropriately filtered organization metadata.

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/22959284307>
> Commit: 1bcf565adcf0e5e44c0bce6899ca7816473ffb25
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=22959284307&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Wed, 11 Mar 2026 16:17:45 UTC

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
@@ -214,7 +214,6 @@ public class SecurityConfig {
                                 ServerWebExchangeMatchers.pathMatchers(HttpMethod.PUT, USER_URL + "/invite/confirm"),
                                 ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, USER_URL + "/me"),
                                 ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, "/v3/**"),
-                                ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, USER_URL + "/features"),
                                 ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, ASSET_URL + "/*"),
                                 ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, ACTION_URL + "/**"),
                                 ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, ACTION_COLLECTION_URL + "/view"),
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/OrganizationServiceCEImpl.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/OrganizationServiceCEImpl.java
@@ -10,6 +10,7 @@ import com.appsmith.server.constants.MigrationStatus;
 import com.appsmith.server.constants.ce.FieldNameCE;
 import com.appsmith.server.domains.Organization;
 import com.appsmith.server.domains.OrganizationConfiguration;
+import com.appsmith.server.domains.User;
 import com.appsmith.server.exceptions.AppsmithError;
 import com.appsmith.server.exceptions.AppsmithException;
 import com.appsmith.server.helpers.CollectionUtils;
@@ -26,6 +27,7 @@ import io.micrometer.observation.ObservationRegistry;
 import jakarta.validation.Validator;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.context.annotation.Lazy;
+import org.springframework.security.core.context.ReactiveSecurityContextHolder;
 import org.springframework.util.StringUtils;
 import reactor.core.observability.micrometer.Micrometer;
 import reactor.core.publisher.Flux;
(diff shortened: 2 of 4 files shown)


app/server/appsmith-server/src/test/java/com/appsmith/server/configurations/AuthGuardTest.java
```diff
@@ -0,0 +1,57 @@
...
+     * OrganizationServiceCETest; this test just confirms the endpoint itself is still reachable.
+     */
+    @Test
+    void tenantCurrentEndpoint_unauthenticated_isAccessible() {
+        webTestClient
+                .get()
+                .uri("/api/v1/tenants/current")
+                .exchange()
+                .expectStatus()
+                .isOk();
```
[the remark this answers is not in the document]
subrata71: Fixed in 77141a92. Changed the chain to use `flatMap(ctx -> Mono.justOrEmpty(ctx.getAuthentication()))` + `.filter(authentication -> authentication.getPrincipal() instanceof User)` before casting, so a non-`User` anonymous principal no longer causes a `ClassCastException`. `defaultIfEmpty(true)` still covers a missing context.

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
```diff
@@ -214,7 +214,6 @@ public SecurityWebFilterChain securityWebFilterChain(ServerHttpSecurity http) {
                                 ServerWebExchangeMatchers.pathMatchers(HttpMethod.PUT, USER_URL + "/invite/confirm"),
                                 ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, USER_URL + "/me"),
                                 ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, "/v3/**"),
-                                ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, USER_URL + "/features"),
```
sondermanish, line 217: qq: why was it present here? did we have any usecase for this?

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/OrganizationServiceCEImpl.java
```diff
@@ -157,12 +159,23 @@ public Mono<Organization> findById(String organizationId, AclPermission permissi
...
+        // Determine if the current principal is an anonymous (unauthenticated) user so we can
+        // suppress instance-identifying metadata in the public response.
+        Mono<Boolean> isAnonymousMono = ReactiveSecurityContextHolder.getContext()
+                .flatMap(ctx -> Mono.justOrEmpty(ctx.getAuthentication()))
+                .filter(authentication -> authentication.getPrincipal() instanceof User)
+                .map(authentication -> ((User) authentication.getPrincipal()).getIsAnonymous())
+                .defaultIfEmpty(true);
+
+        Mono<Organization> clientOrganizationMono = isAnonymousMono.flatMap(
+                isAnonymous -> configService.getInstanceId().flatMap(instanceId -> {
```
sondermanish: Nit pick: for readability let's zip them together?

sondermanish (dismissed):

app/server/appsmith-server/src/main/java/com/appsmith/server/configurations/SecurityConfig.java
subrata71: The endpoint was originally added to `permitAll()` so the client could fetch feature flags before authentication — for example, to enable/disable UI features on the login page itself. However, this also lets any unauthenticated caller enumerate which paid/enterprise features are enabled on the instance, which is the information-disclosure issue in APP-14994. Since the login page doesn't actually depend on these flags pre-auth, removing it from `permitAll()` is safe.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/OrganizationServiceCEImpl.java
subrata71: Done in 47eb6e36 — switched to `Mono.zip(isAnonymousMono, configService.getInstanceId()).flatMap(tuple -> { ... })` so both monos resolve concurrently and the lambda is flat. Also took the opportunity to apply the CodeRabbit nitpick: replaced `getIsAnonymous()` (nullable `Boolean`) with the null-safe `isAnonymous()` (`boolean`) method on `User`.

sondermanish (approved):

PR #41616, opened 2026-03-12
chore: use redis from upstream image

wyattwalter:
Even with newer packages from the Redis project, container scanners are showing false positives for vulnerabilities. Switching to use the upstream Redis built docker images instead.

This bypasses scanners that look for dpkg metadata about packages, but I'm enabling docker dependency updates here via Dependabot to keep this up-to-date.

Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!WARNING]
> Tests have not run on the HEAD edb91ff00ee59af04f559081cb0f539aaa686f56 yet
> <hr>Fri, 13 Mar 2026 16:04:48 UTC

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

* **Chores**
  * Added automated dependency updates for Docker images.
  * Updated Docker build to source Redis binaries from an upstream image instead of installing them during build.
  * Adjusted Redis runtime configuration to remove an explicit working-directory setting.

### DIFF
--- a/deploy/docker/base.dockerfile
+++ b/deploy/docker/base.dockerfile
@@ -1,6 +1,4 @@
-ARG REDIS_VERSION=7.4.8
-
-FROM redis:${REDIS_VERSION} AS redis-source
+FROM redis:7.4.8 AS redis-source
 
 FROM caddy:builder-alpine AS caddybuilder
 
@@ -47,7 +45,6 @@ RUN set -o xtrace \
     /tmp/*
 
 # Install Redis from official image to avoid false positive CVE reports from dpkg-based scanners.
-# To upgrade, change the REDIS_VERSION ARG at the top of this file.
 COPY --from=redis-source /usr/local/bin/redis-server /usr/local/bin/redis-server
 COPY --from=redis-source /usr/local/bin/redis-cli /usr/local/bin/redis-cli
 ENV PATH="/usr/lib/postgresql/14/bin:${PATH}"
--- a/deploy/docker/base.dockerfile
+++ b/deploy/docker/base.dockerfile
@@ -1,3 +1,7 @@
+ARG REDIS_VERSION=7.4.8
+
+FROM redis:${REDIS_VERSION} AS redis-source
+
 FROM caddy:builder-alpine AS caddybuilder
 
 RUN xcaddy build \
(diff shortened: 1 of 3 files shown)


deploy/docker/fs/opt/appsmith/templates/supervisord/redis.conf
```diff
@@ -1,5 +1,4 @@
 [program:redis]
-directory=/etc/redis
```
wyattwalter, line 2: The Debian package dropped a default config file into /etc/redis and we switched directories into it. But all the config we pass aside from the defaults comes from the CLI args, so no need for this and the potential permission issues it creates.

deploy/docker/base.dockerfile
```diff
@@ -47,6 +46,10 @@ RUN set -o xtrace \
     /var/lib/apt/lists/* \
     /tmp/*
 
+# Install Redis from official image to avoid false positive CVE reports from dpkg-based scanners.
+# To upgrade, change the REDIS_VERSION ARG at the top of this file.
+COPY --from=redis-source /usr/local/bin/redis-server /usr/local/bin/redis-server
+COPY --from=redis-source /usr/local/bin/redis-cli /usr/local/bin/redis-cli
```
subrata71, line 49: Looks like this is a good pattern for avoiding false positive CVEs. Just for my clarity on this: Should we adopt this as the default approach for all deps, not just Redis?

wyattwalter: We should only do this on a case-by-case basis. There's two major downsides:

1. this software gets hidden from any automatic vulnerability scanners that use dpkg metadata, so we can hide valid vulnerabilities. This approach must be coupled with what I did in Dependabot (looking into the issue CodeRabbit raised there) so it stays up-to-date to mitigate. 
2. when we install via apt, we get a ton of files laid out that are important, and trying to side-load them like this is really fragile. Redis is quite simple, so we can get away with it. Caddy would be similar, but we already build that for other reasons.

The reason I chose this path for Redis is just because all scanners have a matcher that says ALL Redis packages installed via dpkg on Ubuntu are vulnerable, which is not strictly true.

subrata71 (approved):

PR #41650, opened 2026-03-24
chore: improve error git route aspect rules

sondermanish:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  

## Automation

/ok-to-test tags=""

> [!CAUTION]  

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

* **Chores**
  * Enhanced exception handling framework to preserve error context and root cause information throughout application layers, improving error diagnostics and enabling more informative error messages for better troubleshooting and issue resolution.

### DIFF
--- a/app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/ce/GitRouteAspectCE.java
+++ b/app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/ce/GitRouteAspectCE.java
@@ -289,6 +289,7 @@ public class GitRouteAspectCE {
                         ctx.setError(appsmithException);
                     } else {
                         ctx.setError(new AppsmithException(
+                                e,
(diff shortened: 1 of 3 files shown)


app/server/appsmith-server/src/main/java/com/appsmith/server/aspect/ce/GitRouteAspectCE.java
subrata71, line 305: @sondermanish Don't we need to put e as the last param to log the full stacktrace?

sondermanish: Yes, however I thought that we shouldn't log the stack error twice? I can either add the trace in log or I can  make a correct stack trace here.

subrata71 (approved):

PR #41668, opened 2026-03-28
fix: mssql readonly connections

sebastianiv21:
This PR wires up the MSSQL datasource’s Read-only connection mode end to end. When the datasource is set to read-only, the plugin adds ApplicationIntent=ReadOnly to the JDBC URL (so SQL Server can send the session to a readable secondary in Always On Availability Groups) and sets Hikari’s pool to read-only. Read/write mode is unchanged and does not add that parameter.

A short tooltip on the connection mode control explains that read-only uses ApplicationIntent=ReadOnly and that it mainly matters for Always On setups. New unit tests check that the built JDBC URL includes ApplicationIntent=ReadOnly for read-only mode and omits ApplicationIntent for read/write.

Fixes #`Issue Number`  
_or_  
Fixes https://github.com/appsmithorg/appsmith-ee/issues/8615
> [!WARNING]  

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/31190082454>
> Commit: c7119b71f4e215ee94c710e427376dfc104202c9
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=31190082454&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Fri, 07 Aug 2026 16:08:59 UTC

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

* **New Features**
  * Added a read-only connection mode for MSSQL integrations.
  * Read-only connections support read-intent routing and are marked as read-only where supported.
  * Added guidance about Always On routing and database permissions needed to prevent writes.

* **Bug Fixes**
  * Preserved standard behavior for read-write connections.
  * Clarified that standalone SQL Server instances may not enforce read-only restrictions.


sebastianiv21:
## ✅ Verified end to end against an Always On Availability Group

**Setup:** 2-node clusterless (read-scale) AG on SQL Server 2022 (Docker, `CLUSTER_TYPE = NONE`), secondary replica configured `SECONDARY_ROLE (ALLOW_CONNECTIONS = READ_ONLY)` — i.e. *read-intent only*, so it **rejects any connection that doesn't send `ApplicationIntent=ReadOnly`**. Deploy preview `ce-41668` connected to the local replicas through an ngrok TCP tunnel. Database field set to the AG database (`testdb`), which is required for the intent check to fire.

| # | Action | Connection mode | Target | Result |
|---|---|---|---|---|
| 1 | Datasource Test | Read / Write | read-intent secondary | ✅ Rejected by SQL Server with error 978 ("…accessible … when the application intent is set to read only") — the parameter is genuinely absent in Read/Write mode |
| 2 | `SELECT * FROM dbo.ag_check` | Read-only | read-intent secondary | ✅ Connects (~250 ms) and returns the replicated rows — the parameter is present and honored |
| 3 | `INSERT INTO dbo.ag_check …` | Read-only | read-intent secondary | ✅ Fails with Msg 3906 "Failed to update database … because the database is read-only" |
| 4 | Test + SELECT + INSERT | Read / Write | primary | ✅ Unchanged — connects fast, reads and writes succeed (regression check) |

Cross-checked outside Appsmith with `sqlcmd` (with/without `-K ReadOnly`) and with raw mssql-jdbc **11.2.4.jre11** (the exact version the plugin ships, per `mssqlPlugin/pom.xml`) — identical behavior in all four cases.

subrata71 (commented):
@sebastianiv21  Could you please add a test that attempts to modify one or more rows using a read-only connection? The test should assert that the operation fails with an error indicating that write operations are not allowed on a read-only connection.

sebastianiv21:
@subrata71 Tried adding this, but it can't pass on our test setup: the unit tests run against a standalone Testcontainers SQL Server, and read-only enforcement in MSSQL is entirely server-side. mssql-jdbc's `Connection.setReadOnly()` is a documented no-op, and a standalone instance accepts `ApplicationIntent=ReadOnly` but still takes writes — verified with our exact driver/HikariCP versions (`INSERT SUCCEEDED` through a READ_ONLY-mode pool). Writes only fail against a read-intent-only AG secondary (verified separately on a real AG: Msg 3906) — that's the caveat the tooltip documents.

Alternative: I can add a test asserting the INSERT *succeeds* on a standalone instance, pinning the documented behavior and acting as a canary if a future driver starts enforcing `setReadOnly()`. WDYT?

subrata71:
> @subrata71 Tried adding this, but it can't pass on our test setup: the unit tests run against a standalone Testcontainers SQL Server, and read-only enforcement in MSSQL is entirely server-side. mssql-jdbc's `Connection.setReadOnly()` is a documented no-op, and a standalone instance accepts `ApplicationIntent=ReadOnly` but still takes writes — verified with our exact driver/HikariCP versions (`INSERT SUCCEEDED` through a READ_ONLY-mode pool). Writes only fail against a read-intent-only AG secondary (verified separately on a real AG: Msg 3906) — that's the caveat the tooltip documents.
> 
> Alternative: I can add a test asserting the INSERT _succeeds_ on a standalone instance, pinning the documented behavior and acting as a canary if a future driver starts enforcing `setReadOnly()`. WDYT?

@sebastianiv21 Please go ahead adding that test if it's not already covered by existing tests. Please verify manually the above case that I mentioned.

sebastianiv21:
@subrata71 Done in `e404b320f9` — added `testReadOnlyModeDoesNotEnforceReadOnlyOnStandaloneInstance`, which attempts an INSERT through a READ_ONLY-mode pool. Since the test suite runs on a standalone instance where neither the driver nor the server blocks writes, the test asserts the write *succeeds* and documents why in a comment — it acts as a canary that fails if a future mssql-jdbc upgrade starts enforcing `setReadOnly()`. Manual verification of the failing case was done against a real Always On AG with a read-intent-only secondary (details in [my earlier comment](https://github.com/appsmithorg/appsmith/pull/41668#issuecomment-5185930967)): there the write fails with Msg 3906 "database is read-only".

subrata71 (approved):

Issue #32841, opened 2024-04-22
[Bug]: AST fails to refactor if moustache binding starts with a new line `\n`


subrata71:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Demonstrated in the video

https://github.com/appsmithorg/appsmith/assets/3524599/ce1453fc-138e-461f-bef9-c4d7b9375e16

### Steps To Reproduce

1. Create an Api e.g. Api1
2. Create another Api e.g. Api2
3. In the body of Api2 use Api1 like 
```
{{
   {
       "api1Data": Api1.data
   }
}}
```
4. Change the name of Api1 to ApiOne
5. Go to thte body of APi2 and notice that reference to Api1 still exists

### Environment

Production

### Severity

High (Blocker to building or releasing)

### Version

Cloud / Self Hosted

Ghanasree:
@Nikhil-Nandagopal Can I work on the issue please?

PR #41687, opened 2026-04-02
fix: Custom widget in UI Module not receiving model data in app

amelia-c0n:
### Problem
A Custom widget inside a UI Module renders correctly in the module editor but not in the app that uses the module. The widget does not the React content (e.g. title, key-value table). Other widgets in the same module (Text, Table) receive module inputs as expected.

### Root cause
The Custom widget’s defaultModel binding (e.g. `{{ { title: inputs.title, fields: inputs.fields } }}`) is not resolved when the widget is inside a module instance in the app. The evaluation pipeline does not provide inputs in scope for the Custom widget’s defaultModel evaluation, so the widget receives an empty model and the iframe never gets data.

### Solution (workaround)
A React context–based fallback was added so Custom widgets inside modules can use module inputs when the evaluated model is missing:

1. ModuleInputsContext – New context that holds the current module inputs.
2. Custom widget component – Reads context and, when model is undefined/null (binding failed), uses module inputs as the effective model. When model is present (including intentional {}), it is used as-is.
3. Reference stabilization – effectiveModel is memoized and its reference is stabilized via JSON.stringify comparison so that unchanged content does not trigger extra effect runs or postMessage calls when the provider passes new object references.

Fixes https://github.com/appsmithorg/appsmith-ee/issues/8713

## Automation

/ok-to-test tags=""

> [!WARNING]
> Tests have not run on the HEAD daba8b1e695b935c160eac495590da95e7fe69c5 yet
> <hr>Thu, 02 Apr 2026 13:12:15 UTC

## Communication
- [ ] Yes
- [x] No

## Release Notes

* **New Features**
  * Custom widgets can now access module-level inputs as backup when their directly bound model inputs are empty.

* **Refactor**
  * Custom widget model inputs parameter is now optional, improving flexibility and removing automatic empty object default behavior.

### DIFF
--- a/app/client/src/ce/contexts/ModuleInputsContext.tsx
+++ b/app/client/src/ce/contexts/ModuleInputsContext.tsx
@@ -1,0 +1,14 @@
+import { createContext } from "react";
+
+/**
+ * Context for module inputs. When a Custom widget is inside a ModuleContainer or
+ * ModuleWidget and its model (from defaultModel binding) is empty, it can fall
+ * back to these inputs. This fixes the issue where inputs/Commons1.inputs
+ * fails to resolve during evaluation.
+ */
+export const ModuleInputsContext = createContext<Record<string, unknown>>({});
+
+export const ModuleInputsProvider = ModuleInputsContext.Provider;
+
+/** Stable empty object to avoid unnecessary re-renders when inputs are undefined */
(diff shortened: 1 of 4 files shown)


amelia-c0n:
This is the actual root cause of the issue: https://github.com/appsmithorg/appsmith/issues/32841
The workaround is to remove the newlines (especially the first one after the moustache binding) and republish the package.

PR #41696, opened 2026-04-03
fix(helm): resolve PV schema mismatch for EFS

Arbab1308:
TL;DR: This PR resolves a Kubernetes schema validation error occurring during Appsmith deployments on EKS when EFS is enabled. It removes an incorrectly nested nfs field within the csi block of the PersistentVolume template.

Motivation and Context:
When persistence.efs.enabled is set to true, the current Helm chart attempts to define an nfs block inside the csi configuration. According to the Kubernetes v1.PersistentVolume specification, the csi source does not support a nested nfs field. This causes the EKS API server to reject the resource with the error: failed to create typed patch object: .spec.csi.nfs: field not declared in schema.

Changes:

Removed the redundant nfs: key from deploy/helm/templates/persistentVolume.yaml.

Ensured the csi block correctly identifies the driver and volumeHandle as top-level properties of the CSI spec.

Added quoting to values to ensure string safety during template rendering.

Fixes #38947

Automation
/ok-to-test tags="helm, eks, efs"

:mag: Cypress test results
[!CAUTION]

If you modify the content in this section, you are likely to disrupt the CI result for your PR.

Communication

[x] Yes (This unblocks AWS EFS users who were previously unable to install the chart).

[ ] No

## Release Notes

* **Bug Fixes**
  * iOS currency and numeric input fields now correctly trigger the numeric keypad instead of the full keyboard on Safari.

* **New Features**
  * Helm chart now supports multiple storage backends (EFS, NFS, EBS, GCP, local) with enhanced validation and improved metadata for Kubernetes compliance.

* **Documentation**
  * Added comprehensive iOS testing guide and Helm configuration documentation.

### DIFF
--- a/HELM_IMPROVEMENTS.md
+++ b/HELM_IMPROVEMENTS.md
@@ -1,0 +1,264 @@
+# Helm Chart Improvements - EFS Schema Fix & Production-Ready Enhancements
+
+## Overview
+This PR addresses a critical schema error in the Kubernetes PersistentVolume template when using AWS EFS, while also implementing comprehensive production-ready improvements across the entire Helm chart infrastructure.
+
+**Branch**: `fix/efs-pv-schema-error`  
+**Status**: Ready for Review  
+**Total Commits**: 13  
+**Files Modified**: 6 core Helm templates
+
+---
+
+## 🔴 Critical Issue Fixed
+
+### Problem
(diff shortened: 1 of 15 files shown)


wyattwalter:
Fix for this has been merged and released.

PR #41733, opened 2026-04-10
feat(helm): add support for mongodb operator

wyattwalter:
Adds preview support for the MongoDB operator to replace the old Bitnami chart. I've been unable to find a seamless alternative to the old chart and having to maintain our own templates is untenable. This is the direction that MongoDB is going, so it seems simpler to embrace it.

The Bitnami chart and MongoDB operator options are not mutually exclusive here to allow for transitioning between the two. However, in practice the time where both are enabled should be minimal to avoid confusion and extra resources.

Pros for this approach:
- official MongoDB support, will be better in the long run
- supports single node deployments and replicasets easily with little effort
- handles upgrades and downgrades (neither of which were supported by the old chart)
- leapfrog over MongoDB 7 in the Helm chart (?)

Cons:
- operator adds extra overhead in the case of a single node deployment of MongoDB, have to deal with CRDs, and all that goes with running an operator
- the transition is going to be awkward. once we are ready to cut out the old chart, we'll have to figure out a way to ensure users of the old chart migrate
- no built-in backup support in community operator, but that's not unique from the old chart
- needs a hook to generate a password, which adds RBAC and job overhead

Once we've gotten through some testing, intent is to share docs on how to use this for both new installs and how to transition. Then at some point in the future, create a new major release that forces the administrator to either migrate or bring in their own MongoDB from an external source.

Fixes https://linear.app/appsmith/issue/APP-14807/task-helm-chart-switch-for-mongodb-image-from-bitnami-to-mongodb

## Automation

/ok-to-test tags=""

> [!WARNING]
> Tests have not run on the HEAD 92c0eed3526f1064a0a2a632668f7f64357e706a yet
> <hr>Wed, 29 Apr 2026 20:05:39 UTC

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

* **New Features**
  * Optional MongoDB Operator deployment path, chart version bump, operator-managed MongoDB mode, transition support, init/init-container and DB connection handling, and automatic password-init Job.

* **Documentation**
  * New operator/community install guide, README updates, quickstart, verification, migration and troubleshooting steps.

* **Tests**
  * Expanded Helm tests for operator rendering, transition behavior, password-init hook, and validation errors.

* **Chores**
  * Ignore rule for local MongoDB artifacts updated; added values and JSON schema for operator/community.


deploy/helm/values.yaml
```diff
@@ -86,10 +163,15 @@ prometheus:
...
 
+## @param global.imageRegistry Global image registry. When set, overrides the registry used by
+##   this chart's configurable images (e.g. mongodbCommunity.passwordInit.image) and is honored
+##   by the Bitnami subcharts (redis, mongodb, postgresql, prometheus). Use this to redirect all
+##   image pulls to a mirror or private registry.
 ## @param global.storageClass Global StorageClass for Persistent Volume(s)
 ## @param global.namespaceOverride Override the namespace for resource deployed by the chart, but can itself be overridden by the local namespaceOverride
 ##
 global:
+  imageRegistry: ""
```
wyattwalter: This was an implicit option supported by Bitnami charts, but not documented.

deploy/helm/values.schema.json
```diff
@@ -0,0 +1,110 @@
+{
```
wyattwalter, line 1: It's in the doc already, but I added all the values that I'm adding to the chart in this pass into a new schema file. Will follow this PR / release with a more complete set.

wyattwalter:
The last thing from coderabbit is how the other places in the chart do it, so I'm ignoring that one for now.

sebastianiv21 (dismissed):
LGTM

subrata71 (commented):
Thanks for adding the operator-managed MongoDB path. The overall direction makes sense, but I have a concern around migration safety from the existing Bitnami chart-based MongoDB.
From what I understand, this PR enables a transition mode where both Bitnami MongoDB and `MongoDBCommunity` can run side by side, while Appsmith continues to use the Bitnami instance until `mongodb.enabled=false` is set. That helps stage the new MongoDB deployment, but I do not see anything that actually performs or validates data migration. Am I missing something here?

Can we clarify the expected production migration path before merging/recommending this?

wyattwalter:
@subrata71 Yes, that's intentional. Helm is not an orchestrator, so I did not attempt an automatic migration.

I added a few more explicit notes about the status of migration support (it works, but some edge cases), I'd like to merge this in while that works. It won't release until it gets merged to master.

sebastianiv21 (approved):

PR #41749, opened 2026-04-18
fix(Switch widget): sync isSwitchedOn meta on programmatic changes

### LINKED ISSUE #41566: [Bug]: .isSwitched of a Switch-widget doesn't get the correct (updates) value

### Description

Image
Image

Image

Image

### Steps To Reproduce

1. Click 2 (or more) times  on ToggleButton (note: one click seems mostly to work fine)
2. The Switch-widget is updates correctly and also the (depending) MyButton is updated correctly (with disabled true or false)
3. Click once on Switch-widget
4. The switch-widget changes of course correct, BUT the (depending) MyButton is NOT updated correctly (with disabled true or false)
5. I checked: the reason is on line 8: Switch.isSwitched doesn't have the correct value

### Public Sample App

_No response_

### Environment

Production

### Severity

Medium (Frustrating UX)

### Issue video log

_No response_

### Version

Docker self-hosted, Appsmith v1.95 2 jan 2026

itsmejay80:
## Description

Fixes a stale-binding bug in the Switch widget. Once a user interacts with the widget at least once, its `isSwitchedOn` meta takes precedence over the derived `defaultSwitchState`. Any subsequent programmatic change to `defaultSwitchState` (via `Switch1.setValue(...)`, a bound default state expression, or a ToggleButton driving the switch) updated `defaultSwitchState` but left the stale meta in place. Widgets bound to `Switch1.isSwitchedOn` kept reading the old value.

This PR adds a small `componentDidUpdate` block in `SwitchWidget` that syncs the meta value whenever `isSwitchedOn` changes, mirroring the existing behavior in `CheckboxWidget.componentDidUpdate` (`app/client/src/widgets/CheckboxWidget/widget/index.tsx`).

Fixes #41566

## Automation

/ok-to-test tags="@tag.Switch, @tag.Binding"

> [!CAUTION]

## Testing

Manual repro from the issue:
1. Drop a `Switch`, a `ToggleButton`, and a `Button` on the canvas.
2. Bind `Button.disabled` to `{{Switch1.isSwitchedOn}}`.
3. Wire `ToggleButton.onChange` to `{{Switch1.setValue(!Switch1.isSwitchedOn)}}`.
4. Toggle the `Switch` once directly, then toggle via `ToggleButton` — before the fix, `Button.disabled` goes stale; after the fix it follows the switch consistently.

## Communication
- [ ] Yes
- [x] No

## Summary by CodeRabbit

* **Bug Fixes**
  * Switch widget now correctly syncs its on/off state when changed programmatically, ensuring dependent widgets reflect the current state.
* **Tests**
  * Added a regression test to verify programmatic toggles update dependent bindings and prevent test interference.

### DIFF
--- a/app/client/cypress/e2e/Regression/ClientSide/Widgets/Switch/Switch_spec.js
+++ b/app/client/cypress/e2e/Regression/ClientSide/Widgets/Switch/Switch_spec.js
@@ -154,6 +154,41 @@ describe(
       _.deployMode.NavigateBacktoEditor();
     });
 
+    it("6a. Regression: programmatic defaultSwitchState changes sync isSwitchedOn for dependents (#41566)", function () {
+      // Bind dependent text widget to Switch1.isSwitchedOn
+      cy.openPropertyPane("textwidget");
+      cy.updateCodeInput(
+        ".t--property-control-text",
+        `{{Toggler.isSwitchedOn}}`,
+      );
+
+      // Seed default to true and assert dependent reflects it
+      cy.openPropertyPane("switchwidget");
+      cy.get(".t--property-control-defaultstate input").last().then(($el) => {
+        if (!$el.is(":checked")) cy.wrap($el).click({ force: true });
+      });
+      cy.get(".t--widget-textwidget")
+        .scrollIntoView()
+        .should("contain", "true");
+
+      // User toggles the Switch once so meta.isSwitchedOn starts overriding the default
+      cy.get(`${formWidgetsPage.switchWidget} label`).first().click();
+      cy.get(".t--widget-textwidget")
+        .scrollIntoView()
+        .should("contain", "false");
+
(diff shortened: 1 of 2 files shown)


itsmejay80:
Thanks for the review. Addressed both items in 7f655f5:

- Dropped the `&& this.props.onChange` guard on the meta-sync block you're right that the repro in #41566 doesn't require an onChange handler and the guard was preventing the fix from firing in exactly that case.
- Added a Cypress regression in `Switch_spec.js` that toggles the Switch once (to set meta), then changes `defaultSwitchState` programmatically and asserts a dependent bound to `{{Toggler.isSwitchedOn}}` reflects the new value.

Note: `CheckboxWidget.componentDidUpdate` has the same guard pattern. I left it alone in this PR to keep scope tight, but happy to open a follow-up if maintainers think it's worth fixing there too.

PR #41426, opened 2025-11-27
fix(auth): validate Origin header against APPSMITH_BASE_URL

subrata71:
## Description
Prevent account takeover by validating Origin header in password reset and email verification endpoints against configured APPSMITH_BASE_URL.

- Add validateBaseUrl() helper to check Origin matches APPSMITH_BASE_URL
- Add APPSMITH_BASE_URL field to admin settings UI
- Ensure backward compatibility when APPSMITH_BASE_URL is not set

EE PR: https://github.com/appsmithorg/appsmith-ee/pull/8448

Fixes https://github.com/appsmithorg/appsmith/security/advisories/GHSA-7hf5-mc28-xmcv

## Automation

/ok-to-test tags="@tag.All"

### :mag: Cypress test results

> [!TIP]
> 🟢 🟢 🟢 All cypress tests have passed! 🎉 🎉 🎉
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/19726406804>
> Commit: 3764386de800d51fb7d33071d059920897a6cd00
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=19726406804&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Thu, 27 Nov 2025 07:03:50 UTC

## Communication
Should the DevRel and Marketing teams inform users about this change?
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

* **New Features**
  * Added Appsmith Base URL configuration setting in Admin Settings, enabling administrators to specify their instance's base URL.

* **Bug Fixes**
  * Improved error handling and added URL validation in password reset and email verification flows for enhanced security.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #41766, opened 2026-04-28
fix(security): fail closed when APPSMITH_BASE_URL unset for token-bearing emails (GHSA-j9gf-vw2f-9hrw)

subrata71:
## Description

fix(security): fail closed when APPSMITH_BASE_URL unset for token-bearing emails (GHSA-j9gf-vw2f-9hrw)

- **Primary fix:** Extract a single canonical `SecureBaseUrlResolver` helper that gates the host portion of every emailed absolute link (forgot-password, email verification, workspace invite, instance-admin invite). The resolver no longer trusts the request `Origin` header as the canonical host when `APPSMITH_BASE_URL` is unset — it returns `Mono.empty()` instead, which propagates through the existing controller wiring (`defaultIfEmpty(true)` / `thenReturn`) so unauthenticated flows return the same generic 200 success without dispatching email (anti-enumeration preserved). Auth-gated invite flows surface the new `MISCONFIGURED_INSTANCE_BASE_URL` error.
- **Migration path:** New opt-in env var `APPSMITH_ALLOW_INSECURE_ORIGIN_BASED_LINKS` (defaults to `false`). When set to `true`, the legacy behaviour is restored. Documented as deprecated; intended only as a transition window for self-hosted deployments that have not yet configured `APPSMITH_BASE_URL`. The helper logs a clear WARN on every call so operators are aware they are running in an insecure mode.
- **Defense-in-depth:** Re-routes the workspace-invite and instance-admin-invite flows in `EmailServiceCEImpl` through the same helper. Previously these flows derived the invite URL host directly from `originHeader`, which was an auth-gated but identical attack surface — and was a latent bug in EE multi-org mode (where forgot-password used a server-controlled URL but invite emails still used `Origin`).
- **Test coverage:** New `SecureBaseUrlResolverCEImplTest` (7 tests, pure unit tests, no Spring context) pinning all four state permutations of the resolver. Existing `EmailServiceCEImplTest` updated with `@TestPropertySource(properties = "APPSMITH_BASE_URL=http://example.com")` so its invite/admin-invite tests exercise the strict-match path.
- **Strict-mode unchanged:** the protection added in PR #41426 (Origin must equal `APPSMITH_BASE_URL` when configured) is preserved exactly. The new compat flag does NOT weaken it — it only governs the unconfigured case.

### Vulnerability

| Field | Value |
|-------|-------|
| **GHSA** | [GHSA-j9gf-vw2f-9hrw](https://github.com/appsmithorg/appsmith/security/advisories/GHSA-j9gf-vw2f-9hrw) |
| **CVE** | Not yet assigned |
| **CVSS** | 8.1 (High) — `CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:U/C:H/I:H/A:N` |
| **CWE** | CWE-346 (Origin Validation Error), CWE-807 (Reliance on Untrusted Inputs in a Security Decision) |
| **Affected component** | Token-bearing email link generation (forgot-password, email verification, invite, instance-admin invite) |

### Exposure Analysis

**Who can exploit this?** Any unauthenticated network attacker who can reach `POST /api/v1/users/forgotPassword` or `POST /api/v1/users/resendEmailVerification`. No prior authentication, no recon.

**What can an attacker achieve?** When `APPSMITH_BASE_URL` is unset (the default state for self-hosted installs that haven't been hardened), an attacker can forge the `Origin` header to make Appsmith send password-reset and email-verification emails whose clickable host is attacker-controlled. The emails are otherwise legitimate (sent from the real Appsmith server, signed with SPF/DKIM, rendered with the real Appsmith branding). A victim who clicks the link delivers a valid token to the attacker's host — the attacker can then replay it against the legitimate Appsmith instance via `PUT /api/v1/users/resetPassword` to take over the account.

**Blast radius:** Per-user. Each successful exploit takes over one targeted account. There is no scope change to the wider workspace from this primitive alone, but a successful takeover of a workspace admin or instance admin escalates from there through normal application flows.

**Affected configurations:**
- Self-hosted CE/EE single-org with `APPSMITH_BASE_URL` unset and outbound email enabled — vulnerable.
- EE multi-org — already safe via the previous `OrganizationService.getOrganizationBaseUrl()` override that derived the host from `slug + deploymentDomain`. This PR relocates that override into the new helper so it now also covers invite flows (which were a latent bug under multi-org).
- Cloud-hosted Appsmith — has `APPSMITH_BASE_URL` configured, so the strict-mode check from PR #41426 applies. Not affected.

**Evidence of exploitation in the wild:** None known. Reported responsibly via GHSA by an external researcher.

### Fix

**Root cause.** `UserServiceCEImpl#resolveSecureBaseUrl` failed open when `APPSMITH_BASE_URL` was empty: it returned the caller-supplied value verbatim. That caller-supplied value originated from `@RequestHeader("Origin")` and was used to interpolate the host of token-bearing email links. The underlying coding pattern is using request metadata (a value the *client* claims) to populate an outbound, server-issued security artifact. Token-bearing artifacts must be bound to the server's *configured* identity, not the requester's claimed identity.

**Fix strategy.** Extract `SecureBaseUrlResolver` as a Spring `@Component` with a single resolver method. Make this the *only* path through which any emailed absolute link gets its host. The helper has three resolution branches:

1. **Configured base URL set** (existing strict-mode from PR #41426): require exact match with the supplied value, error on mismatch.
2. **Unconfigured + new compat flag off (default)**: log WARN, return `Mono.empty()`. Unauthenticated callers (forgot-password, resend-verification) translate this into a generic 200 success without dispatching email — anti-enumeration preserved by the existing `defaultIfEmpty(true).thenReturn(...)` controller wiring. Authenticated callers (invite, admin-invite) translate it into `MISCONFIGURED_INSTANCE_BASE_URL` so the admin caller sees an actionable configuration error.
3. **Unconfigured + new compat flag on (`APPSMITH_ALLOW_INSECURE_ORIGIN_BASED_LINKS=true`)**: log WARN, return the caller-supplied value. Migration-only, deprecated.

**What was intentionally NOT changed:**
- All HTTP endpoint paths and request/response shapes are unchanged.
- The strict-mode check from PR #41426 is preserved exactly. The new compat flag does NOT weaken it.
- `SecurityConfig` permit-all matchers stay as-is — the endpoints must remain reachable without auth; the fix is in URL generation, not auth gating.
- DTO shapes (`ResetUserPasswordDTO`, `ResendEmailVerificationDTO`) unchanged.

**Defense-in-depth.**
- Single canonical helper across all four token-bearing email flows, including the auth-gated invite flows that previously bypassed `resolveSecureBaseUrl` entirely.
- Helper unit tests pin every branch including a regression that the compat flag does not weaken strict-mode.
- WARN log on every fail-closed and fail-back call gives operators an immediate signal when emails aren't being dispatched and why.

**Reporter coordination.** This fix matches the design the reporter recommended in the GHSA discussion thread, point-for-point: single server-side helper, fail-closed by default, anti-enumeration preserved for unauth flows, opt-in insecure compatibility flag for migration. Reporter is `@0xmrma`.

### CE/EE sync

**Mostly CE-only safe with a small shadow EE PR.**

| File | Diff status (CE vs EE pre-sync) | Strategy |
|------|--------------------------------|----------|
| `UserServiceCEImpl.java` | Identical | CE edit syncs cleanly to EE |
| `EmailServiceCEImpl.java` | Identical | CE edit syncs cleanly to EE |
| `EmailServiceCEImplTest.java` | Identical | CE edit syncs cleanly to EE |
| `UserServiceCECompatibleImpl.java` | Identical | CE edit syncs cleanly to EE |
| `SecureBaseUrlResolver.java`, `SecureBaseUrlResolverCE.java`, `SecureBaseUrlResolverCEImpl.java` (new) | n/a | New files sync to EE |
| `SecureBaseUrlResolverImpl.java` (new) | CE no-op variant; **EE replaces with multi-org-aware variant** | Shadow EE PR layered on top |
| `UserServiceImpl.java` | EE-specific class with multi-org override | Shadow EE PR strips the now-redundant `resolveSecureBaseUrl` override and updates the constructor signature |
| `EmailServiceImpl.java` | EE-specific class | Shadow EE PR updates the constructor to inject `SecureBaseUrlResolver` |
| `AppsmithError.java` / `AppsmithErrorCode.java` | Differs (EE has additional codes) | Shadow EE PR adds the same `MISCONFIGURED_INSTANCE_BASE_URL` entry |

Shadow EE branch (`fix/origin-validation-email-links-ghsa-j9gf` on `appsmithorg/appsmith-ee`) is prepared and pushed; will be linked in a follow-up comment.

### Disclosure

> **Do not merge until advisory is ready for disclosure coordination.**
>
> After merge:
> 1. Confirm fix is in release branch
> 2. Coordinate with security team on disclosure timeline
> 3. Update advisory with patched version and publish
> 4. Notify reporter (`@0xmrma`)

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/25127579141>
> Commit: 306e492772c3b6f2692b91e40d68ee81ce6299e0
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=25127579141&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Wed, 29 Apr 2026 20:02:20 UTC

## Communication
- [x] Yes — operators of self-hosted instances that have not configured `APPSMITH_BASE_URL` will see password-reset / email-verification / invite emails stop being delivered after upgrading. Migration path: set `APPSMITH_BASE_URL`, or as a temporary opt-in, set `APPSMITH_ALLOW_INSECURE_ORIGIN_BASED_LINKS=true` (insecure, deprecated).
- [ ] No

## Follow-ups

- [ ] Admin UI tooltip update in `app/client/src/ce/pages/AdminSettings/config/configuration.tsx` strengthening the wording around `APPSMITH_BASE_URL` being required (deferred — local `lint-staged` requires a fresh `yarn install` in `app/client`; will land as a separate small PR).
- [ ] Plan to deprecate `APPSMITH_ALLOW_INSECURE_ORIGIN_BASED_LINKS` env var in a future major release once telemetry confirms self-hosted operators have migrated.
- [ ] Add Spring-context integration tests in `UserServiceTest` for `forgotPasswordTokenGenerate` / `resendEmailVerification` fail-closed paths (deferred — needs Mongo+Redis; helper unit tests already pin the security-critical semantics).

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced use of a trusted instance base URL for token-bearing email flows (password reset, verification, workspace invites). Added a clear internal error when the base URL is not configured; invite/email flows are disabled until set. Origin matching is stricter to prevent spoofing.

* **Tests**
  * Added comprehensive unit tests covering secure base URL resolution and compatibility modes.

* **Documentation / CI**
  * Updated CI, deploy and local scripts and added design docs to ensure the base URL is set during tests and previews.

### DIFF
--- a/.github/workflows/ci-test-custom-script.yml
+++ b/.github/workflows/ci-test-custom-script.yml
@@ -188,6 +188,7 @@ jobs:
             -e APPSMITH_CLOUD_SERVICES_BASE_URL=http://host.docker.internal:5001 \
             -e APPSMITH_CLOUD_SERVICES_SIGNATURE_BASE_URL=http://host.docker.internal:8090 \
             -e APPSMITH_RATE_LIMIT=1000 \
+            -e APPSMITH_BASE_URL=http://localhost \
             --add-host=host.docker.internal:host-gateway --add-host=api.segment.io:host-gateway --add-host=t.appsmith.com:host-gateway \
             cicontainer
 
--- a/.github/workflows/ci-test-limited-with-count.yml
+++ b/.github/workflows/ci-test-limited-with-count.yml
@@ -273,6 +273,7 @@ jobs:
             -e APPSMITH_DISABLE_TELEMETRY=true \
             -e APPSMITH_INTERCOM_APP_ID=DUMMY_VALUE \
             -e APPSMITH_CLOUD_SERVICES_BASE_URL=http://host.docker.internal:5001 \
+            -e APPSMITH_BASE_URL=http://localhost \
             --add-host=host.docker.internal:host-gateway --add-host=api.segment.io:host-gateway --add-host=t.appsmith.com:host-gateway \
             cicontainer
 
--- a/.github/workflows/ci-test-limited.yml
+++ b/.github/workflows/ci-test-limited.yml
@@ -184,6 +184,7 @@ jobs:
             -e APPSMITH_DISABLE_TELEMETRY=true \
             -e APPSMITH_INTERCOM_APP_ID=DUMMY_VALUE \
             -e APPSMITH_CLOUD_SERVICES_BASE_URL=http://host.docker.internal:5001 \
+            -e APPSMITH_BASE_URL=http://localhost \
             --add-host=host.docker.internal:host-gateway --add-host=api.segment.io:host-gateway --add-host=t.appsmith.com:host-gateway \
             cicontainer
 
--- a/.github/workflows/ci-test-playwright.yml
+++ b/.github/workflows/ci-test-playwright.yml
@@ -174,6 +174,7 @@ jobs:
             -e APPSMITH_CLOUD_SERVICES_BASE_URL=http://host.docker.internal:5001 \
             -e APPSMITH_CLOUD_SERVICES_SIGNATURE_BASE_URL=http://host.docker.internal:8090 \
             -e APPSMITH_RATE_LIMIT=1000 \
+            -e APPSMITH_BASE_URL=http://localhost \
             -e APPSMITH_CARBON_API_BASE_PATH=https://carbon.appsmith.com \
             -e APPSMITH_CARBON_API_KEY="$APPSMITH_CARBON_API_KEY" \
             -e APPSMITH_AI_SERVER_MANAGED_HOSTING=true \
--- a/app/client/cypress/support/commands.js
+++ b/app/client/cypress/support/commands.js
@@ -98,11 +98,15 @@ export const addIndexedDBKey = (key, value) => {
 };
 
 Cypress.Commands.add("stubPostHeaderReq", () => {
+  // GHSA-j9gf-vw2f-9hrw: server-side strict-mode now requires Origin to match
+  // APPSMITH_BASE_URL. Use the live Cypress baseUrl so the intercept produces a
+  // valid Origin (the literal string "Cypress" used to live here as a synthetic-
+  // traffic flag and no longer parses as a URL).
   cy.intercept("POST", "/api/v1/users/invite", (req) => {
-    req.headers["origin"] = "Cypress";
+    req.headers["origin"] = Cypress.config("baseUrl");
   }).as("mockPostInvite");
   cy.intercept("POST", "/api/v1/applications/invite", (req) => {
-    req.headers["origin"] = "Cypress";
+    req.headers["origin"] = Cypress.config("baseUrl");
   }).as("mockPostAppInvite");
 });
 
@@ -752,7 +756,8 @@ Cypress.Commands.add("startServerAndRoutes", () => {
       hostname: window.location.host,
     },
     (req) => {
-      req.headers["origin"] = "Cypress";
+      // GHSA-j9gf-vw2f-9hrw: send a real Origin matching APPSMITH_BASE_URL.
+      req.headers["origin"] = Cypress.config("baseUrl");
     },
   ).as("connectGitLocalRepo");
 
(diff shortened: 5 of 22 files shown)


subrata71:
Shadow EE PR: https://github.com/appsmithorg/appsmith-ee/pull/8997 (pre-staged for sync resolution)

PR #41550, opened 2026-02-10
fix: prevent open redirect in login and OAuth2 redirect flows

RaikaSurendra:
## Description
This PR fixes an **open redirect vulnerability** (CWE-601) in the login and OAuth2 authentication flows.

### Problem
The `RedirectHelper.fulfillRedirectUrl()` method accepted any URL as a redirect target after successful authentication, with no domain validation. An attacker could craft a malicious login link like:

```
https://app.example.com/api/v1/login\?redirectUrl\=https://evil.com/phish
```

After the user logs in successfully, they would be redirected to `evil.com` — enabling credential phishing, session hijacking, or social engineering attacks.

**Two attack vectors shared the same root cause:**
1. **Form login** — `redirectUrl` query parameter passed directly to `fulfillRedirectUrl()`
2. **OAuth2 login** — `redirectUrl` embedded in the OAuth2 `state` parameter, extracted in `handleOAuth2Redirect()`

### Fix
Added origin validation to ensure redirect URLs only point to the same host as the request:

- **`isSafeRedirectUrl()`** — validates that absolute redirect URLs match the request's `Origin` header (host + port comparison). Relative URLs are always allowed.
- **`sanitizeRedirectUrl()`** — public method that returns the default redirect URL (`/applications`) if the URL is unsafe. Used by the OAuth2 handler.
- Applied validation in **`fulfillRedirectUrl()`** (form login path)
- Applied `sanitizeRedirectUrl()` in **`handleOAuth2Redirect()`** (OAuth2 path)

### Behavior
| Redirect URL | Before | After |
|---|---|---|
| `/applications/123` | ✅ Allowed | ✅ Allowed (relative) |
| `https://app.appsmith.com/apps` | ✅ Allowed | ✅ Allowed (same origin) |
| `https://evil.com/phish` | ❌ **Allowed** | ✅ **Blocked** → `/applications` |
| `https://evil.com@app.appsmith.com` | ❌ **Allowed** | ✅ **Blocked** |

### Testing
- Added `RedirectHelperOpenRedirectTest.java` with 14 test cases covering:
  - Relative URLs (always safe)
  - Same-origin absolute URLs (safe)
  - Different host/port (blocked)
  - Subdomain mismatch (blocked)
  - No Origin header (absolute URLs blocked, relative allowed)
  - Common bypass attempts (userinfo injection, subdomain tricks, fragment injection)
  - Case-insensitive host comparison

### Files Changed
[...]

PR #41820, opened 2026-05-15
fix: SSO login redirect lands on /applications instead of original app URL

sebastianiv21:
## Description

Fixes a regression introduced in [#41550](https://github.com/appsmithorg/appsmith/pull/41550) where unauthenticated users opening a deep app link (e.g. `https://<host>/app/<app-name>/<page-id>?ssoTrigger=saml`) land on `/applications` after completing OIDC/SAML login instead of the originally requested URL.

### Root cause

The frontend triggers the SSO entry URL via `window.location.href = "/oauth2/authorization/keycloak?redirectUrl=<encoded absolute URL>"` — a top-level GET navigation. Browsers do not send the `Origin` header for top-level GET navigations, but the `redirectUrl` value is always absolute (it comes from `window.location.href`).

[`RedirectHelper.isSafeRedirectUrl`](https://github.com/appsmithorg/appsmith/blob/release/app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/RedirectHelper.java) reads `httpHeaders.getOrigin()`, finds it empty, and returns `false` for *any* absolute URL. `sanitizeRedirectUrl` then rewrites the URL to `<origin>/applications`, which gets baked into the OAuth2 `state` and used by the post-callback success handler — so the user always lands on `/applications`.

### Fix

Restructure `isSafeRedirectUrl` so the early bypass-blocking checks (relative path, http(s) prefix, URI parse, userInfo, null host) happen first, then split into two host-comparison branches:

- **Origin present:** existing logic — full host + port + scheme-aware port normalization. Unchanged semantics.
- **Origin absent:** derive the request host from `X-Forwarded-Host` (preferred, supports proxied envs and comma-separated lists) then `Host`. Compare hostnames only (these headers carry no scheme/port). If neither header is present, reject — preserves the old strict behavior in the truly no-info case.

The open-redirect protection from #41550 is preserved — only same-host URLs become permissible in the no-Origin path, and userinfo / protocol-relative / non-http(s) bypass attempts are still rejected.

Fixes [appsmith-ee#8815](https://github.com/appsmithorg/appsmith-ee/issues/8815)

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/26544767632>
> Commit: 045a32e11ddf2cc2519b4d4248cdd50371ba914d
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=26544767632&attempt=5" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Thu, 28 May 2026 20:55:26 UTC

## Communication
- [ ] Yes
- [x] No

## Test plan
- [x] Unit tests added in `RedirectHelperOpenRedirectTest` covering the new fallback path: Host / X-Forwarded-Host, XFH precedence, comma-separated XFH lists, port stripping, IPv6 brackets, userinfo + protocol-relative bypass attempts blocked.
- [ ] On a Keycloak/OIDC-enabled instance: sign out, open a deep app link in incognito, complete SSO, verify the browser lands on `/app/<app-name>/<page-id>?ssoTrigger=saml` (not `/applications`).
- [ ] Negative check: visit `/oauth2/authorization/keycloak?redirectUrl=https://evil.com/phish`, complete SSO, verify the browser lands on `/applications` and server logs contain `Blocked open redirect attempt to: https://evil.com/phish`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Hardened redirect validation to block malformed or userinfo-containing URLs, require a resolved request host, and enforce origin/host/port matching with scheme-aware default-port handling and IPv6/port parsing; improved logging sanitization for rejected redirects.

* **Tests**
  * Added tests covering Origin-absent fallbacks, X-Forwarded-Host precedence (first value), comma/IPv6/port parsing, X-Forwarded-Port behavior, and negative cases for unsafe redirects.

[Review Change Stack](https://app.coderabbit.ai/change-stack/appsmithorg/appsmith/pull/41820?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/RedirectHelper.java
```diff
@@ -181,29 +185,36 @@ static boolean isSafeRedirectUrl(String redirectUrl, HttpHeaders httpHeaders) {
...
 
-        // For absolute URLs, the host must match the request origin
-        String origin = httpHeaders.getOrigin();
-        if (StringUtils.isEmpty(origin)) {
-            // If there is no Origin header, we cannot validate — reject absolute URLs
-            // to be safe. Relative URLs were already allowed above.
+        final URI redirectUri;
+        try {
+            redirectUri = new URI(redirectUrl);
+        } catch (URISyntaxException e) {
```
subrata71, line 191: Nit: Let's add a log here with warn log level

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/RedirectHelper.java
```diff
@@ -181,29 +185,36 @@ static boolean isSafeRedirectUrl(String redirectUrl, HttpHeaders httpHeaders) {
...
+        }
+
+        String origin = httpHeaders.getOrigin();
+        if (!StringUtils.isEmpty(origin)) {
+            // Origin-present path: full host + port + scheme-aware port normalization.
+            final URI originUri;
+            try {
+                originUri = new URI(origin);
+            } catch (URISyntaxException e) {
                 return false;
```
subrata71, line 216: Nit: Let's add a log here with warn log level

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/RedirectHelper.java
```diff
@@ -224,9 +235,54 @@ static boolean isSafeRedirectUrl(String redirectUrl, HttpHeaders httpHeaders) {
...
+        return normalizedRedirectHost.equalsIgnoreCase(requestHost);
+    }
+
+    /**
+     * Derives the request's host for same-origin comparison when the Origin header
+     * is absent. Prefers X-Forwarded-Host (handles proxied deployments and may carry
+     * a comma-separated list — the first entry is the outermost client-facing host),
+     * then falls back to the Host header. Returns null when neither is set.
+     */
+    private static String extractRequestHost(HttpHeaders headers) {
```
subrata71: Could you please check if there are already such util method available which does the same thing? Just trying to make sure we are not keeping duplicate logic scattered in different places.

subrata71 (commented):
The changes look good to me. If you want to address those nitpicks then it would be great otherwise I can approve. But before you merge this can you please create a shadow EE PR and make sure all tests pass there as well?

app/server/appsmith-server/src/main/java/com/appsmith/server/helpers/RedirectHelper.java
sebastianiv21: [`CustomServerOAuth2AuthorizationRequestResolverCE.extractHostFromRequest`](https://github.com/appsmithorg/appsmith/blob/release/app/server/appsmith-server/src/main/java/com/appsmith/server/authentication/handlers/ce/CustomServerOAuth2AuthorizationRequestResolverCE.java#L291) was doing essentially the same job (XFH comma-parsing + Host fallback) in the same SSO entry flow. 

I promoted `RedirectHelper.extractRequestHost(HttpHeaders)` from `private` to `public static` and refactored the OAuth2 resolver to delegate. 

The resolver still keeps `request.getURI().getHost()` as a last-resort fallback, so the only behavior change is the order between `Host` header and URI host when XFH is unset (which is a no-op in proxied envs since XFH is always set there). Let me know if the changes are ok.

sebastianiv21:
@subrata71 I've addressed the nitpicks + a coderabbit suggestion. Could you please take a look at the changes? Tests are passing here and [in EE](https://github.com/appsmithorg/appsmith-ee/pull/9165)

subrata71 (approved):

PR #41851, opened 2026-05-27
fix(jdbc): add Hikari keepalive and bounded socket timeout

wyattwalter:
## Summary

Resolves [APP-15267](https://linear.app/appsmith/issue/APP-15267/intermittent-pql-connector-timeouts).

A production pod recently hung for ~66 hours with every Postgres query timing out at the 10s action timeout, recovering only on pod replacement. Root cause: HikariCP's connection-adder thread blocks indefinitely inside pgjdbc's auth read because `socketTimeout` defaults to `0` (infinite). Since `addConnectionExecutor` is single-threaded, one stuck creator wedges the entire pool — every subsequent borrow fails at the action timeout because no replacement connection is ever produced.

Two complementary changes across 4 JDBC plugins (Postgres, MSSQL, Oracle, Redshift):

- **Hikari `setKeepaliveTime(150_000)`** — `Connection.isValid()` probe on idle connections every 150s, comfortably under AWS NAT's 350s idle-eviction default. Prevents the most common trigger that forces the pool to create new connections.
- **Driver-level socket read timeout** — bounds every socket read including the auth-time read where the connection-adder was observed hung. Default **600s (10 min)**, configurable via `appsmith.plugin.jdbc.socket-timeout-seconds` (env `APPSMITH_PLUGIN_JDBC_SOCKET_TIMEOUT_SECONDS`). 600s leaves plenty of headroom for legitimately slow queries (default action timeout is 10s) while ensuring the pool can always recover.

Driver-specific property mapping:

| Plugin | Property | Unit |
|---|---|---|
| Postgres / Redshift | `socketTimeout` | seconds |
| MSSQL | `socketTimeout` | milliseconds |
| Oracle | `oracle.net.READ_TIMEOUT` | milliseconds |
| Snowflake | keepalive only | (driver uses a different property mechanism; full coverage deferred) |

Oracle and Redshift previously had no `ConnectionPoolConfig` injection; constructor-injected here following the existing Postgres/MSSQL pattern.

## Reproduction and validation

Reproduced the wedge end-to-end with toxiproxy + iptables against a local Appsmith v2.0 container. Thread dump confirmed the hang in `HikariPool-N connection adder` at `doAuthentication.receiveChar`, with byte-identical CPU time across snapshots ~7 minutes apart — definitive proof the thread was parked indefinitely. With the fix, the pool recovers automatically once the network path is restored.

## Test plan

- [x] Deploy preview comes up cleanly and Postgres datasources connect normally
- [x] Verify default 600s applies when `APPSMITH_PLUGIN_JDBC_SOCKET_TIMEOUT_SECONDS` is unset
- [x] Verify env-var override is honored (set to a different value, observe behavior)
- [x] Confirm a normal Postgres query under 600s server-side latency executes successfully
- [x] (Manual repro) Simulate connection-adder wedge via toxiproxy + iptables; confirm pool recovers within ~600s instead of hanging indefinitely

## Follow-ups (not in this PR)

- Snowflake `networkTimeout` plumbing (its driver uses a different property mechanism that lives in upstream Properties)
- Consider per-query `Connection.setNetworkTimeout(actionTimeout + buffer)` in `executeCommon` for more precise per-query bounding (defense in depth)
- Production observability: periodic Hikari pool-stats logging (not just on successful query) so the next recurrence — if any — surfaces faster

## Summary by CodeRabbit

* **New Features**
  * Configurable socket timeout for database connections with a 600-second default, exposed via configuration.

* **Improvements**
  * Connection pools across multiple database plugins now respect the configured socket timeout and enable TCP keepalive probes to reduce idle connection drops.

* **Tests**
  * Updated tests to provide and validate the new socket timeout and pool-size parameters.

[Review Change Stack](https://app.coderabbit.ai/change-stack/appsmithorg/appsmith/pull/41851?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/30005563482>
> Commit: 2f8c1689c0cce58399b5e7419c9e40bbba3d3c13
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=30005563482&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Thu, 23 Jul 2026 13:17:24 UTC

## Automation
/ok-to-test tags="@tag.All"


subrata71 (approved):
The changes look good to me. No Cypress tests were run as part of this PR, but I’ve just triggered them.

Please keep an eye on things once these changes reach the EE repo. Alternatively, you can create a shadow EE PR to catch any failures early.

PR #41889, opened 2026-06-10
ci: distribute SUT image via GHCR instead of Actions cache

wyattwalter:
## Description

The `build-docker-image` workflow currently distributes the system-under-test image to test jobs by saving it to a gzipped tarball (single-threaded, ~4 minutes) and storing it in the Actions cache, which every test shard then restores as a single blob (~3 minutes each, worse under fan-out contention since the restores can't parallelize).

This switches distribution to GHCR: the build job pushes the image to `ghcr.io/<repo>-citest:run-<run_id>`, and test jobs pull it from there. Registry pulls are layer-parallel and CDN-backed. Measured on the EE repo across a 20-shard run: median image acquisition dropped from ~218s to 74s per shard, and the build job no longer spends ~4 minutes on `docker save | gzip`.

Changes:

- `build-docker-image.yml`: push the built `cicontainer` image to GHCR using `GITHUB_TOKEN`; remove the tar/gzip/cache-save steps.
- `ci-test-custom-script.yml`, `ci-test-playwright.yml`: pull the image from GHCR and tag it back to `cicontainer`, so downstream `docker run` references are unchanged.
- `ci-test-limited.yml`, `ci-test-limited-with-count.yml`: same, honoring `previous-workflow-run-id` for the image-reuse path (`run-<previous-id>` tag). Registry tags outlive Actions cache eviction, so reuse becomes more reliable, bounded by the retention window below.
- `ghcr-citest-retention.yml` (new): daily prune of `run-*` package versions older than 7 days (each is ~2 GB). Supports manual dispatch with a dry-run flag.

Notes:

- The GHCR package is created on first push and defaults to private, linked to this repository. Pulls inside Actions use `GITHUB_TOKEN`; no new secrets or permissions are required.
- Fork-PR paths are unaffected: fork-triggered `pull_request` events never reach the build job (existing gate), and the `/ok-to-test` flow runs via `repository_dispatch` in base-repo context.
- A `/ci-test-limit` rerun referencing a run that predates this change will not find a registry tag for it; this resolves itself as new runs accumulate.

Ref: APP-15259

## Validation

- After merge, run `/ci-test-limit` and `/ci-test-limit-count` on a test PR, including one rerun that exercises `previous-workflow-run-id`.
- Dispatch `ghcr-citest-retention.yml` with `dry-run: true` to confirm the token can enumerate and would prune package versions before the first scheduled run.

## Summary by CodeRabbit

* **Chores**
  * CI now pushes and pulls test container images via the GitHub Container Registry instead of caching/loading tarball artifacts.
  * Test jobs retrieve run-specific image tags so downstream steps pull images by tag.
  * Added a scheduled/manual cleanup workflow to prune old test image versions in the registry with configurable retention and dry-run options.

> [!WARNING]
> Tests have not run on the HEAD a55f67e897d439a9e5a02d611f0b33bbbc6ccc86 yet
> <hr>Wed, 10 Jun 2026 20:11:30 UTC

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -120,21 +120,16 @@ jobs:
           args+=(--label "org.opencontainers.image.version=${{ steps.info_json.outputs.version }}")
           docker build -t cicontainer "${args[@]}" .
 
-      # Saving the docker image to tar file
-      - name: Save Docker image to tar file
+      # Push the SUT image to GHCR so test jobs can pull it with layer-level
+      # parallelism instead of restoring a single-blob tarball from the Actions
+      # cache. Old run-<id> tags are pruned by ghcr-citest-retention.yml.
+      - name: Push docker image to GHCR
+        env:
+          GHCR_IMAGE: ghcr.io/${{ github.repository }}-citest:run-${{ github.run_id }}
         run: |
-          docker image ls --all --no-trunc --format '{{.Repository}},{{.ID}}' \
-            | grep -v cicontainer \
-            | cut -d, -f2 \
-            | xargs -r docker rmi || true
-          docker save cicontainer -o cicontainer.tar
-          gzip cicontainer.tar
-
-      - name: Cache docker image
-        uses: actions/cache/save@v4
-        with:
-          path: cicontainer.tar.gz
-          key: docker-image-${{github.run_id}}
+          echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io --username "${{ github.actor }}" --password-stdin
+          docker tag cicontainer "$GHCR_IMAGE"
+          docker push "$GHCR_IMAGE"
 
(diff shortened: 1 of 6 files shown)


wyattwalter:
Re: the outside-diff-range comment on `ci-test-limited-with-count.yml` ("Allow reusable workflow invocations through the ci-test-limited job gate") — skipping this one. The gate is pre-existing code, and the finding misreads how reusable workflows behave: jobs in a called workflow inherit the **caller's** triggering event, so `github.event_name` here is `repository_dispatch` / `workflow_dispatch` / `push`, all already in the allow list. `github.event_name == 'workflow_call'` is never true at runtime, so the suggested condition would be dead code. The identical gate pattern in `ci-test-custom-script.yml` runs via `workflow_call` on every scheduled TBP run today.

subrata71 (approved):

PR #41893, opened 2026-06-12
fix(test): wait for email HTML body in Email_settings spec

wyattwalter:
## Summary

`Email_settings_Spec` fails intermittently in CI (~25-35% of runs) when it shares a duration-balanced shard with heavy specs. It's a null-handling bug in the spec's email flow, exposed under maildev/SMTP contention.

## Root cause

`waitForEmail` matches an email by subject + recipient, then hands it to the test. Under contention maildev can surface an email's metadata before its HTML part is parsed, so the matched email's `.html` is `undefined`. Tests #3–#6 then do:

```js
const emailHtml = email.html;          // undefined
const match = emailHtml.match(/.../);  // TypeError, thrown inside .then()
```

That throw is an unhandled promise rejection that aborts the test (the wrapping `try/catch` is inert against async Cypress failures). Captured CI stack: `Email_settings_Spec.ts` test #4 (`To verify invite workspace email`); the same `email.html` → `.match` pattern is in tests #3, #5 and #6.

## Fix

Fix it at the source rather than masking it. Add an opt-in `requireHtml` flag to `waitForEmail`: when set, a matched email whose HTML body isn't populated yet is treated as **not yet delivered**, so polling continues until the body arrives (or the existing timeout). Tests #3–#6 opt in; their email-body assertions then run for real on a complete email.

(An earlier revision of this PR used `email.html || ""`, but that only stopped the crash by silently skipping the invite-link verification when the body was missing — masking the flake. Reverted in favor of the helper-level fix.)

`waitForEmail` is identical in CE and EE, so this lands in CE and the hourly sync carries it to EE. The new flag is optional and defaults to current behavior, so other callers (including this spec's test #2, which reads `email.text`) are unaffected.

## Validation

Attribution and the fix were verified on appsmith-ee via amplified `/ci-test-limit` runs (multiple copies of the spec downstream of a heavy git spec in one shard): the unfixed baseline reproduced the failure at the exact line above.

Ref APP-15291

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/27509234477>
> Commit: 56deaca5230eec74bbe33359d8436569c2afe061
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=27509234477&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Email`
> Spec:
> <hr>Sun, 14 Jun 2026 19:36:07 UTC

## Summary by CodeRabbit

* **Tests**
  * Improved email-related test robustness: four end-to-end flows (password reset, workspace invite, application invite — developer role, and application invite — view role) now explicitly require the HTML portion of delivered emails, reducing flakiness and better validating email content.
* **Chores**
  * Updated limited-test configuration to run the client-side email regression spec during focused CI runs.

## Automation
/ok-to-test tags="@tag.Email"

### DIFF
--- a/app/client/cypress/e2e/Regression/ClientSide/AdminSettings/Email_settings_Spec.ts
+++ b/app/client/cypress/e2e/Regression/ClientSide/AdminSettings/Email_settings_Spec.ts
@@ -202,6 +202,7 @@ describe(
             timeout: TIMEOUT,
             targetSubject: resetPassSubject,
             targetEmail: emailOne,
+            requireHtml: true,
           })
           .then((email) => {
             if (email) {
@@ -209,7 +210,7 @@ describe(
               expect(email.headers.subject).to.equal(resetPassSubject);
               expect(email.headers.to).to.equal(emailOne);
 
-              const emailHtml = email.html || ""; // Store the email HTML content
+              const emailHtml = email.html; // Store the email HTML content
               const resetPasswordLinkMatch = emailHtml.match(
                 /href="([^"]*resetPassword[^"]*)"/,
               );
@@ -282,6 +283,7 @@ describe(
             timeout: TIMEOUT,
             targetSubject: inviteEmailSubject,
             targetEmail: emailTwo,
+            requireHtml: true,
           })
           .then((email) => {
             if (email) {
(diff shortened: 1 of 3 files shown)


wyattwalter:
Resurrecting this flake fix. It was closed 2026-06-30 by the inactivity bot with zero human review, not because anything was wrong with it.

Context: `Email_settings_Spec` was one of the worst hidden-flake offenders in the June baseline (4 retried / 3 recovered / 1 persistent). This change makes `waitForEmail` optionally require the HTML body (`requireHtml`) so the spec stops matching a partially-delivered email and then failing on the missing body.

Evidence: the targeted `/ci-test-limit` repro passed (run 27428547269) and the full ci-test was green (run 27509234477). Reopening re-triggered the standard client checks; the Cypress side is already green.

Small, test-only, self-contained. Requesting a review. Tracking under the flake-reduction effort: https://linear.app/appsmith/issue/APP-15705

subrata71 (approved):

PR #41895, opened 2026-06-12
feat(client): shared tabOrder property for keyboard Tab navigation across standard widgets

### LINKED ISSUE #37947: configure the order of the inputs that are focused when the “Tab” key is pressed

sebastianiv21:
## Description

**TL;DR:** Adds a single platform-level `tabOrder` property, exposed in a shared **Accessibility** property-pane section for every standard (non-Anvil) widget, letting builders optionally control keyboard Tab order in fixed layout. Existing apps keep their exact current Tab behavior unless `tabOrder` is explicitly set.

### What changed

- **Shared property pane plumbing** — `WidgetFactory` centrally appends one shared `Accessibility > Tab order` section to both property-pane paths (legacy `getPropertyPaneConfig` and content config) via `WidgetProvider/factory/tabOrderPropertyConfig.ts`. No per-widget config edits, no per-widget property names or validation rules. Anvil-only (`WDS_*`, `SECTION_WIDGET`, `ZONE_WIDGET`) and internal widgets (`CANVAS_WIDGET`, `SKELETON_WIDGET`, `TABS_MIGRATOR_WIDGET`) are excluded; widgets without a property pane stay untouched.
- **New `TAB_ORDER_INPUT` control** (`TabOrderControl`) reusing the ADS `NumberInput`: persists only valid non-negative integers as numbers, removes the property from the DSL when cleared (blank always means Auto), never persists `""` or invalid input. Not JS-convertible (`isJSConvertible/isBindProperty/isTriggerProperty: false`), validated as `ValidationTypes.NUMBER { min: 0, natural: true }`.
- **Runtime** — `PositionedContainer` renders a sanitized `data-tab-order` attribute only for valid explicit values (`0` is valid and earliest). No positive native `tabIndex` is ever used, so non-focusable widgets stay non-focusable.
- **Fixed-layout tabbing** (`useWidgetFocus/tabbable.ts`) — new `sortTabbableWidgets()`: when no widget in the current scope has a valid explicit order, it delegates to the unchanged `sortWidgetsByPosition` (exact current behavior). Otherwise, explicitly ordered widgets come first ascending (duplicates tie-break by position), Auto/invalid widgets follow in existing position order, and Shift+Tab reverses the same sequence. Modal, container, JSONForm, CheckboxGroup, SwitchGroup and ButtonGroup behaviors are preserved.

### Compatibility

- No DSL migration, no `LATEST_PAGE_VERSION` bump, no widget default pollution — `tabOrder` only exists in a DSL after a builder sets it.
- Auto layout is intentionally unchanged: `useWidgetFocus` opts out of auto layout, so `tabOrder` is currently enforced only by the fixed-layout custom tabbing (`FlexComponent` and `layoutSystems/anvil/*` untouched).

### Security notes

- No JS/eval support for the field; numeric-only non-negative integer validation at the control, and re-sanitization at every runtime read (`sanitizeTabOrder`).
- Only a sanitized integer is reflected into the DOM attribute; no user-controlled selectors or HTML; invalid values are ignored and treated as Auto.

### Tests

- `utils/widgetTabOrder.test.ts` — sanitizer value rules (0 valid; null/blank/decimals/negatives/NaN/Infinity/non-numeric → Auto).
- `utils/hooks/useWidgetFocus/tabbable.test.ts` — all-Auto preserves position order; explicit overrides position; 0 before 1; explicit widget above/left can be next; duplicate tie-break; blank/invalid → Auto; Shift+Tab; nested container scope; modal scope; composite widgets keep native internal tabbing.
- `WidgetProvider/factory/tabOrderPropertyConfig.test.ts` — unit tests plus an integration test over all registered widgets proving standard non-Anvil widgets expose exactly one identical shared property and Anvil/internal widgets expose none.
- `components/propertyControls/TabOrderControl.test.tsx` — persistence semantics incl. regression that clearing returns to Auto without persisting an invalid value.
- `components/designSystems/appsmith/PositionedContainer.test.tsx` — `data-tab-order` renders only for valid explicit values (including `0`), never a native `tabindex`.

Fixes #37947 

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/29929171193>
> Commit: aba0b0f91ce24b5a64aa5a28d9c771b705f8c775
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=29929171193&attempt=2" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Wed, 22 Jul 2026 20:30:36 UTC

## Communication
- [x] Yes
- [ ] No

## Summary by CodeRabbit

* **New Features**
  * Added explicit `tabOrder` support for eligible widgets via a shared Accessibility property-pane section (clearable numeric input).
  * Introduced a tab-order overlay badge in the editor with an **Alt+T** hotkey toggle.
* **Bug Fixes**
  * Improved keyboard navigation and reverse traversal to respect explicit/automatic tab order.
  * Prevented focus handling from getting stuck when no focusable target is available.
* **Tests**
  * Expanded coverage for property-pane integration, tab-order rendering, overlay badge behavior, keyboard traversal, and numeric input interactions.


sondermanish:
I've reviewed the code, The changes which I see is looking fine. I've two concerns:
- I'm assuming that this property would be saved on git serialization, so if A discard happens, what is the automatic value being set?
- Not sure if in case we would have missed something

app/client/src/components/propertyControls/TabOrderControl.tsx
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Numeric input for the shared `tabOrder` property.
+ *
+ * Unlike NUMERIC_INPUT, this control:
+ * - persists valid values as numbers (non-negative integers only)
+ * - removes the property from the DSL when the field is cleared, so a blank
+ *   field always means "Auto"
+ * - never persists invalid placeholder strings
+ */
+class TabOrderControl extends BaseControl<TabOrderControlProps> {
```
ashit-rath: Ideally we try to make these controls generic so that it can be re-used and in this case it looks we are aren't doing anything specific so can we not call it `TabOrderControl` and call is something generic.

app/client/src/utils/hooks/useWidgetFocus/tabbable.ts
```diff
@@ -202,6 +205,113 @@ function getWidgetSiblingsOfNode(node: HTMLElement) {
...
+ * followed by the remaining widgets in automatic position order. Shift+Tab
+ * walks the same sequence in reverse.
+ *
+ * @param boundingClientRect top/left of the element we are tabbing from
+ * @param tabbableDescendants candidate widgets of the current scope
+ * @param shiftKey
+ * @param currentWidget the widget we are tabbing from, when it is part of the scope
+ * @returns
+ */
+export function sortTabbableWidgets(
```
ashit-rath, line 260: I don't get why this is needed; we are already passing `tabOrder` in the PositionedComponentLayer.tsx which helps the tabbing by default. What is re-sorting based on tabOrder helping us with?

sebastianiv21: `data-tab-order` is just a custom data attribute. The browser doesn't do anything with it on its own. PositionedComponentLayer only renders it, so it doesn't affect the focus order. In fixed layout, Tab navigation is handled entirely in our code. `useWidgetFocus` intercepts the keydown event, handleTab calls `preventDefault()`, and then manually moves focus to the next widget with `.focus()`. `sortTabbableWidgets` is the only place that actually reads data-tab-order, so without it the attribute is effectively ignored. We also can't rely on native `tabindex` because positive tabindex values are considered an accessibility anti-pattern, and they would also conflict with our custom Tab handling.

sebastianiv21:
> I've reviewed the code, The changes which I see is looking fine. I've two concerns:
> 
> * I'm assuming that this property would be saved on git serialization, so if A discard happens, what is the automatic value being set?
> * Not sure if in case we would have missed something

@sondermanish On serialization: tabOrder isn't in widget defaults and doesn't have a defaultValue, so it only gets written to the DSL when a builder actually sets a number. Auto isn't a stored value, it's just the absence of the key. So on discard, a widget goes back to its committed state (no key means Auto, or whatever number was committed). There's no migration, and any absent/null/invalid value is treated as Auto at runtime anyway, so there's nothing for discard to corrupt.

app/client/src/utils/hooks/useWidgetFocus/tabbable.ts
```diff
@@ -13,8 +15,22 @@ export const MODAL_WIDGET = ".t--modal-widget";
...
+  "select",
+  "textarea",
+  "button",
+  "object",
+  "audio",
+  "video",
+  "[tabindex='-1']",
+]
+  .map((selector) => `${selector}:not([data-tabbable='false'])`)
+  .join(", ");
```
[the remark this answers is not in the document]
sebastianiv21: False positive — the source template is `${selector}:not([data-tabbable='false'])` with the closing paren present. The generated selector is a:not([data-tabbable='false']), input:not([data-tabbable='false']), ... (balanced parens; verified by evaluating the expression). The verification script in the analysis chain dropped the ) when re-typing the template, which is what produced the invalid string. The handleTab unit tests exercising this selector pass in CI.

sondermanish (approved):

PR #41587, opened 2026-03-02
feat: Redis datasource: add TLS (SSL mode) support in backend + datasource UI

mohanarpit:
## Summary
This PR enables TLS connections for the Redis datasource and exposes the setting in the datasource configuration UI.

## What changed
- Added Redis TLS mode handling in datasource creation:
  - `ENABLED` => use TLS
  - `DISABLED` or missing SSL config => no TLS (backward compatible default)
  - any other SSL auth type => fail fast with datasource argument error
- Updated Redis URI construction to support `rediss://` when TLS is enabled.
- Added Redis datasource form control:
  - Section: `SSL (optional)`
  - Field: `SSL mode`
  - Options: `Disabled (DISABLED)` and `Enabled (ENABLED)`
  - Default: `DISABLED`
- Added/updated tests for TLS behavior and URI generation.

## Behavior / compatibility
- Existing Redis datasources without SSL config continue to work as non-TLS.
- TLS is opt-in via datasource `SSL mode = Enabled`.

## Test plan
- Added unit tests for:
  - Redis datasource creation with TLS `ENABLED`
  - Redis datasource creation with TLS `DISABLED`
  - Unsupported SSL auth type returns error
  - Redis URI uses `redis://` for non-TLS and `rediss://` for TLS

## Verification
- Pre-commit Spotless checks were applied successfully during commit.
- Full plugin tests were not run in this step.

## Automation

/ok-to-test tags="@tag.All"

### :mag: Cypress test results

> [!IMPORTANT]
> 🟣 🟣 🟣 Your tests are running.
> Tests running at: <https://github.com/appsmithorg/appsmith/actions/runs/22610750972>
> Commit: 4a9a492d99ebeca70320ec86cf4c4cfbb2bf513a
> Workflow: `PR Automation test suite`
> Tags: `@tag.All`
> Spec: ``
> <hr>Tue, 03 Mar 2026 06:10:33 UTC

PR #41921, opened 2026-06-23
fix(security): block Redis datasource from internal Appsmith Redis (GHSA-qhfj-g87x-m39w)

wyattwalter:
## Summary

Closes [GHSA-qhfj-g87x-m39w](https://github.com/appsmithorg/appsmith/security/advisories/GHSA-qhfj-g87x-m39w). The Redis datasource plugin opened a Jedis connection to a user-controlled host/port with no IP-class filtering, so a user with datasource-create permission could point a Redis datasource at the internal Appsmith Redis. The existing HTTP-side filter doesn't cover Redis — Jedis uses the JVM's default resolver and bypasses the Reactor Netty hook.

This PR also closes a DNS-rebinding TOCTOU on the same path and enforces TLS certificate hostname verification for `rediss://` datasources (items 4 and 5 below).

## Changes

1. **Block Redis datasources from reaching the internal Appsmith Redis.** `RedisPlugin.datasourceCreate` now consults a shared `RestrictedHostFilter`. Blocks literal hostname / IP matches against `APPSMITH_REDIS_URL` and `APPSMITH_REDIS_GIT_URL`, plus loopback / link-local / cloud-metadata literals. Enforcement is connection-time only — `PluginExecutor.validateDatasource` is documented format-only, so host-policy belongs on the connection path. UX: Save succeeds, "Test Datasource" rejects with `Host not allowed.`.

2. **Make SSRF protections always-on; rename `IN_DOCKER` to `APPSMITH_DISABLE_SSRF_FILTER`.** The previous filter gated the loopback / link-local block on `IN_DOCKER=1`, which Appsmith's official Docker image always sets — so users on documented deployment paths were already getting this protection. This change closes the gap for unofficial distributions (self-built images, bare-metal, etc.) and replaces a poorly-named variable with one that actually describes what it does. The most likely place to hit this is a development machine running the server directly against `http://localhost:*` — set `APPSMITH_DISABLE_SSRF_FILTER=true` to allow it.

3. **Extract `RestrictedHostFilter` from `WebClientUtils`.** Pure refactor — the host-filter logic had already grown several non-HTTP call sites and the class name had outgrown its scope.

4. **Close the DNS-rebinding TOCTOU on the Redis path (upgrade Jedis 3.3.0 → 5.2.0 + pin the resolved IP).** The create-time pre-check in (1) and Jedis's own connection-time lookup resolved the hostname independently, so a hostile resolver could return an allowed IP to the check and the internal Redis IP to the driver. Jedis 5.2.0 (up from 3.3.0, which predated the API) allows injecting a `JedisSocketFactory`; the new `RestrictedHostJedisSocketFactory` resolves once at connect time, validates every resolved address, and connects directly to the pinned IP — so the driver never re-resolves and the window is closed. The create-time check stays as fast Save/Test UX and now logs the block at WARN, tagged with the requesting user/org/trace (a useful reconnaissance signal). Jedis is plugin-isolated (PF4J + shaded); the server's own Redis runs on Lettuce and is untouched.

5. **Enforce TLS certificate hostname verification for `rediss://` datasources.** Jedis 5.2.0 validates the certificate trust chain but does not verify that the cert's SAN/CN matches the host (it only does so by default in 7.5.0+), so a man-in-the-middle presenting any CA-trusted-but-mismatched cert on the pinned IP would have been accepted. The socket factory now sets the endpoint identification algorithm to `HTTPS`, verifying against the configured hostname (we connect to the pinned IP but hand the original hostname to the TLS layer for SNI and verification). This was a pre-existing gap, not introduced here; closing it now is low-risk because Redis TLS support is recent (#41587, ~2026-03) and barely adopted, and chain validation was already on — so only CA-trusted-but-mismatched certs and IP-addressed `rediss://` (certs rarely carry IP SANs) are newly rejected.

6. **Housekeeping: drop dead `IN_DOCKER` references from Dockerfile / deploy scripts / CI workflow.** No consumer left after the rename in (2).

/ok-to-test tags="@tag.All"

## Test plan

- [x] CI green across all commits (Jedis 5.2.0 upgrade, socket factory, TLS verification, logging).
- [x] Deploy preview: Redis datasource at \`localhost\` / in-cluster Redis service DNS — Save accepted, Test rejected with \`Host not allowed.\`. Legitimate external Redis works.
- [x] Connect-time SSRF rejection (the rebinding gate) covered by `RedisPluginTest` (`socketFactory_blockedHost_failsAtConnectTime`). A live end-to-end rebinding flip needs a changing resolver and was not exercised on the DP (rbndr.us is unreachable from AWS).
- [x] Confirm the one existing `rediss://` user (Appsmith internal) connects via a hostname matching its cert, now that hostname verification is enforced.
- [x] Dev environment: confirm \`APPSMITH_DISABLE_SSRF_FILTER=true\` unblocks \`http://localhost:*\` datasources during local development.

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/28840292682>
> Commit: c795905e5e2ace43f06797277939f7467e04f638
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=28840292682&attempt=2" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Tue, 07 Jul 2026 13:31:52 UTC

## Release Notes

* **New Features**
  * Added centralized SSRF/restricted host filtering with denylisting and normalization.
  * Enabled consistent enforcement across outbound web requests, Elasticsearch DNS resolution, SMTP test-email, and Redis datasource connections.
  * Mitigated Redis DNS rebinding by validating once and pinning the resolved address.

* **Bug Fixes**
  * Blocked hosts now fail with a consistent “Host not allowed” error across execution paths.

* **Tests**
  * Added dedicated host/IP blocking test suite.
  * Updated integration tests to explicitly control SSRF filtering enablement and state restoration during runs.


app/server/appsmith-interfaces/src/main/java/com/appsmith/util/RestrictedHostFilter.java
```diff
@@ -0,0 +1,556 @@
...
+                hosts.add(normalizeHostForComparison(hostCandidate));
+            } catch (UnknownHostException e) {
+                throw new IllegalStateException("Invalid disallowed host configured: " + hostCandidate, e);
+            }
+        }
+    }
+
+    private static Set<String> computeInternalRedisHosts() {
+        final Set<String> hosts = new HashSet<>();
+        addInternalRedisHostFromEnv(hosts, "APPSMITH_REDIS_URL");
```
subrata71, line 167: What if Redis is configured using application.properties e.g.  `appsmith.redis.url`. In that case it remains fail-open. Please correct me if I am wrong.

app/server/appsmith-interfaces/src/main/java/com/appsmith/util/RestrictedHostFilter.java
```diff
@@ -0,0 +1,556 @@
...
+     * then connects directly to the returned {@link InetAddress} — so the check and the connect
+     * share one resolution and the driver never re-resolves the hostname.
+     *
+     * <p>The caller owns DNS resolution deliberately: passing addresses in (rather than resolving
+     * here) lets the caller distinguish an unresolvable host (a genuine connection error) from a
+     * policy block (this method returning empty), and guarantees the validated address is the one
+     * connected to. Honors the {@code APPSMITH_DISABLE_SSRF_FILTER} kill-switch and the test
+     * allowlist, in which case the first resolved address is returned without policy checks.
+     */
+    public static Optional<InetAddress> firstAllowedRedisAddress(String host, InetAddress[] resolvedAddresses) {
```
subrata71, line 430: QQ: Do we need to loop through all the addresses? Could it be possible if the first resolved IP is IPV6 but the actual redis is listening on IPV4 address?

app/server/appsmith-interfaces/src/main/java/com/appsmith/util/RestrictedHostFilter.java
wyattwalter: Excellent point. I think this is addressed now.

app/server/appsmith-interfaces/src/main/java/com/appsmith/util/RestrictedHostFilter.java
wyattwalter: Yep, adjusted that. Good catch.

subrata71 (approved):
The changes look good to me. The release notes should mention that SSRF protections are now always-on (previously gated on IN_DOCKER=1) or do you have any other plan regarding this communication?

wyattwalter:
Yeah, we'll have to add this to the release notes.

PR #41966, opened 2026-07-08
chore(docker): bind bundled internal services to loopback

wyattwalter:
**Linear:** [APP-15350](https://linear.app/appsmith/issue/APP-15350)

## Summary

Defense-in-depth hardening of the Docker image. The bundled internal services are bound to the loopback interface (`127.0.0.1`) by default so they only accept connections from within the container and are not reachable over the container's bridge network interface.

## Changes

- **RTS** (`server.ts`): bind the RTS listener to `127.0.0.1`.
- **Backend server** (`application-ce.properties`): default `server.address` to `127.0.0.1` (overridable via `APPSMITH_SERVER_ADDRESS`).
- **Redis** (`entrypoint.sh`): add `bind 127.0.0.1 -::1` to the generated Redis config.

These services are internal to the image and are fronted by the in-container Caddy proxy, so they do not need to listen on the container's external interfaces.

## Testing

All services continue to be reached through the existing in-container proxy paths.

## Summary by CodeRabbit

* **Bug Fixes**
  * RTS server now binds to an explicit configured host by defaulting to loopback (`127.0.0.1`) and using the configured RTS port.
  * App server now defaults to loopback (`127.0.0.1`) unless overridden by `APPSMITH_SERVER_ADDRESS`.
  * Redis now listens only on loopback (`127.0.0.1`/`::1`) by default for safer local-only access.
* **Documentation**
  * Updated Docker documentation to reflect **Caddy** supervision and restart examples instead of **Nginx**.

> [!WARNING]
> Tests have not run on the HEAD d685989def1f13d725b71e95a076020674889382 yet
> <hr>Wed, 08 Jul 2026 14:59:26 UTC

### DIFF
--- a/app/client/packages/rts/src/server.ts
+++ b/app/client/packages/rts/src/server.ts
@@ -2,7 +2,7 @@ import server from "./ee/server";
 import log from "loglevel";
 import { VERSION as buildVersion } from "./version"; // release version of the api
 
-const APPSMITH_RTS_PORT = process.env.APPSMITH_RTS_PORT || 8091;
+const APPSMITH_RTS_PORT = Number(process.env.APPSMITH_RTS_PORT) || 8091;
 
 server.listen(APPSMITH_RTS_PORT, "127.0.0.1", () => {
   log.info(
--- a/app/client/packages/rts/src/server.ts
+++ b/app/client/packages/rts/src/server.ts
@@ -3,8 +3,10 @@ import log from "loglevel";
 import { VERSION as buildVersion } from "./version"; // release version of the api
 
 const APPSMITH_RTS_PORT = Number(process.env.APPSMITH_RTS_PORT) || 8091;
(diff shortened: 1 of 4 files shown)


salevine (approved):
Assessment: This is a small, well-scoped, low-risk change. The pattern is consistent across all three services — sane loopback default, explicit env-var override for anyone who needs it — and the Number() fix for the RTS port is a genuine bug fix picked up along the way. A couple of things worth being aware of, not blockers: anyone who was previously connecting directly to RTS/backend/Redis over the container's bridge network (bypassing Caddy) rather than through the proxied paths will lose that access — this seems intentional per the PR description, but worth confirming it's flagged in release notes if any users rely on that. Also, there's no override env var for the Redis bind address like there is for the other two services, which is probably fine since Redis is purely internal, but it's a minor asymmetry

PR #41967, opened 2026-07-08
fix(plugins): parameterize query execution for Redshift, Snowflake and Databricks

wyattwalter:
**Linear:** [APP-15378](https://linear.app/appsmith/issue/APP-15378)

## Summary

The Redshift, Snowflake and Databricks plugins previously executed queries by substituting dynamic bindings directly into the SQL text. This PR adds prepared-statement support to all three so query execution is parameterized, bringing them in line with the other SQL plugins (Postgres, MySQL, MSSQL).

## Changes

- **Parameterized execution** — dynamic values (`{{...}}`) are bound as `PreparedStatement` parameters instead of being substituted into the SQL string.
- **"Use prepared statement" toggle** — added to the plugin editor/settings for each plugin, enabled by default, with an opt-out that matches the Postgres/MySQL/MSSQL behavior. When the setting is absent or unrecognized, execution defaults to prepared; only an explicit opt-out selects the non-prepared path (which still supports binding identifiers/fragments a prepared statement cannot parameterize).
- **`dependency.json`** — added for each plugin so the editor links the query body to the prepared-statement toggle (matching the existing SQL plugins).
- **Array-typed parameters** are rejected with a clear error rather than being bound incorrectly, and prepared statements are closed on binding failure.

## Tests

Adds unit tests for all three plugins covering:

- the prepared path (`prepareStatement` used, values bound, `createStatement` never called),
- default-on behavior when the setting is absent or malformed,
- the explicit opt-out selecting the non-prepared path, and
- resource cleanup (statements closed on binding failure).

Focused module suites are green: `redshiftPlugin` 19, `snowflakePlugin` 16, `databricksPlugin` 8 — 0 failures. `spotless:check` passes for all three modules.

## Summary by CodeRabbit

* **New Features**
  * Added “Use prepared statements” execution mode for Databricks, Redshift, and Snowflake with dynamic binding support.
  * Updated editors to show separate prepared vs non-prepared query inputs, controlled by the prepared-statement toggle.
  * Added run behavior, run-confirmation, and query-timeout settings for these plugins.

* **Bug Fixes**
  * Improved prepared execution by binding dynamic values as JDBC parameters (with plugin-specific SQL handling) and reliably closing JDBC resources.
  * Added clearer errors for query-preparation failures and unsupported array parameters.
  * Improved write-style/no-result handling and safer catalog/schema usage.

* **Tests**
  * Added regression tests covering prepared vs non-prepared routing, defaults, array rejection, error surfacing, and dependency wiring.

> [!WARNING]
> Tests have not run on the HEAD 2da557361d9bcf56a71f7d5babd5ea4a01ff8394 yet
> <hr>Wed, 08 Jul 2026 21:35:06 UTC

### DIFF
--- a/app/server/appsmith-plugins/databricksPlugin/src/main/java/com/external/plugins/DatabricksPlugin.java
+++ b/app/server/appsmith-plugins/databricksPlugin/src/main/java/com/external/plugins/DatabricksPlugin.java
@@ -1,16 +1,29 @@
 package com.external.plugins;
 
+import com.appsmith.external.constants.DataType;
+import com.appsmith.external.datatypes.AppsmithType;
+import com.appsmith.external.dtos.ExecuteActionDTO;
 import com.appsmith.external.exceptions.pluginExceptions.AppsmithPluginError;
 import com.appsmith.external.exceptions.pluginExceptions.AppsmithPluginException;
 import com.appsmith.external.exceptions.pluginExceptions.StaleConnectionException;
+import com.appsmith.external.helpers.DataTypeServiceUtils;
+import com.appsmith.external.helpers.MustacheHelper;
 import com.appsmith.external.models.ActionConfiguration;
+import com.appsmith.external.models.ActionExecutionRequest;
 import com.appsmith.external.models.ActionExecutionResult;
 import com.appsmith.external.models.BearerTokenAuth;
 import com.appsmith.external.models.DatasourceConfiguration;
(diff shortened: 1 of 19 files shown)


subrata71 (commented):
There's a good chance of regressions. It needs manual verification since there are not enough automated test coverage for these areas. The changes look good by the way.

PR #41972, opened 2026-07-09
fix(git): project only required fields when loading sibling-branch entities during import

wyattwalter:
## Problem

Creating a git branch (`create-ref`) imports the exported artifact, and the import reconciles entities against **all sibling branches** via `getExistingResourcesInOtherBranchesFlux`. Both the `ActionCollection` and `NewAction` implementations fetched **full documents** across every branch `applicationId` with no projection, pulling the heavy `unpublished*/published*` action bodies for thousands of entities in a single transaction.

For git-connected apps with many branches this can exceed Mongo's `transactionLifetimeLimitSeconds`, aborting the transaction with `TransactionExceededLifetimeLimitSeconds` and making branch operations fail. Observed in a customer environment: an `applicationId $in [~120 ids]` read on `actionCollection` returning ~15k documents (~400 MB, ~27 KB/doc) inside the import transaction, killed at 60s.

## Fix

Pass a field projection (`id, baseId, applicationId, gitSyncId, policies`) instead of `null` so only the fields the import actually consumes are loaded. This mirrors the existing `NewPage` import path, which already projects exactly these fields for the same other-branches lookup.

## Why this is safe

The sibling-branch objects are **read-only** during import: only `gitSyncId` (the lookup map key) and `baseId`/`id` (base-id resolution) are consumed downstream, and these objects are never mutated or written back to Mongo. Projecting away the heavy body fields cannot cause data loss.

## Automation

/ok-to-test tags="@tag.ImportExport"

## Scope

- `ActionCollectionApplicationImportableServiceCEImpl.getExistingResourcesInOtherBranchesFlux`
- `NewActionApplicationImportableServiceCEImpl.getExistingResourcesInOtherBranchesFlux`

`ModuleInstanceApplicationImportableServiceImpl` has the same `null`-projection pattern and will be addressed separately.

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/29083663032>
> Commit: 162f73666b74514cac624ddd024c507150f4cbd5
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=29083663032&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.ImportExport`
> Spec:
> <hr>Fri, 10 Jul 2026 10:03:20 UTC

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved branch import reliability by restricting lookups to only the fields required for matching and resolution, instead of loading full documents.
  * This reduces unnecessary data retrieval and processing during cross-branch comparisons, making imports more efficient and consistent.


app/server/appsmith-server/src/main/java/com/appsmith/server/actioncollections/importable/applications/ActionCollectionApplicationImportableServiceCEImpl.java
```diff
@@ -58,7 +58,17 @@ public Flux<ActionCollection> getExistingResourcesInCurrentArtifactFlux(Artifact
...
+        // Only project the fields consumed downstream during import (gitSyncId for the lookup map, and
+        // id/baseId for base-id resolution). Fetching full documents here loads the heavy
+        // unpublishedCollection/publishedCollection bodies across every branch, which for apps with many
+        // branches can exceed the Mongo transaction lifetime limit during branch creation.
+        final List<String> projectionForOtherBranches = List.of(
+                ActionCollection.Fields.id,
+                ActionCollection.Fields.baseId,
+                ActionCollection.Fields.applicationId,
+                ActionCollection.Fields.gitSyncId,
+                ActionCollection.Fields.policies);
```
sondermanish: Please add policyMap as well

subrata71 (approved):

Issue #15380, opened 2022-07-22
[Task]: audit logs file structure and basic wireframe


IAmAnubhavSaini:
### Is there an existing issue for this?

- [X] I have searched the existing issues

IAmAnubhavSaini:
https://github.com/appsmithorg/appsmith/issues/15373

IAmAnubhavSaini:
This is done. Have a visual prototype.

PR #40936, opened 2025-06-13
chore: Disable anonymous user tracking when feature flag turned on

trishaanand:
Fixes #`Issue Number`  
_or_  
Fixes `Issue URL`
> [!WARNING]  
> _If no issue exists, please create an issue first, and check with the maintainers if the issue is valid._

## Automation

/test sanity

### :mag: Cypress test results

> [!TIP]
> 🟢 🟢 🟢 All cypress tests have passed! 🎉 🎉 🎉
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/15996654359>
> Commit: ba524fe2f769b6f8d2c72e0332560dd5e0c0465e
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=15996654359&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.Sanity`
> Spec:
> <hr>Tue, 01 Jul 2025 11:08:01 UTC

## Communication
Should the DevRel and Marketing teams inform users about this change?
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

* **New Features**
  * Introduced a feature flag to control event tracking for anonymous users.
  * Added the ability to block analytics event tracking for anonymous users when the feature flag is enabled.

* **Bug Fixes**
  * Improved logic to ensure analytics events are not sent for anonymous users if the feature flag is active.

* **Chores**
  * Updated analytics initialization to respect the new tracking preference for anonymous users.
  * Enhanced tracking initialization flow to conditionally enable or disable analytics based on user status and feature flag.
  * Added tests to verify analytics initialization respects user tracking preferences.

PR #41973, opened 2026-07-09
fix: fully honor block-anonymous-tracking flag on client and server

salevine:
## Description

Makes the `configure_block_event_tracking_for_anonymous_users` feature flag **completely** turn off anonymous-user event tracking. Two gaps let anonymous events through when the flag was on:

### Gap 1 — client license bypass (`app/client/src/ce/sagas/userSagas.tsx`)
`shouldTrackUser` short-circuited on an active license:
```
isAnonymous && (licenseActive || (telemetryOn && !featureFlag))
```
so licensed EE/cloud instances tracked anonymous users regardless of the flag. Removed the `licenseActive` bypass — the flag is now honored on **all** instances.

### Gap 2 — ungated direct `sendEvent` path (`AnalyticsServiceCEImpl`)
The flag was only checked in `sendObjectEvent`. Direct callers of `sendEvent(event, userId, props)` with an anonymous userId bypassed it (and that path resolves the anonymous user to the client `x-anonymous-user-id`). Added a gate on the public `sendEvent` that blocks anonymous userIds **before** id resolution, and it fails closed if the flag state can't be resolved (drops the event, completes the chain). The existing `sendObjectEvent` check is kept (it governs the session-user/signup path, where `sendEvent` receives the new user's email rather than `"anonymousUser"`). `sendObjectEvent` now routes through the extracted private `sendEventInternal` to avoid a redundant (Redis-backed) flag check on the hot anonymous path (published-app page views / action executions).

### Scope
- Behavior is **opt-in**: the flag defaults to `false`, so existing instances are unchanged until an admin enables it.
- **Out of scope (intentional):** the usage-pulse channel (`/api/v1/usage-pulse`) still emits the anonymous ID — that feeds billing/seat metering, not Segment analytics, and is treated separately.

### Tests
- **Client** — new `userSagas.test.ts` covering the `shouldTrackUser` matrix, including the regression guard that anonymous + telemetry-on + flag-on returns `false` on a licensed instance.
- **Server** — new `AnalyticsServiceCEImplTest` cases: a direct anonymous `sendEvent` enqueues nothing when the flag is on, and fails closed (no enqueue, no error) when the flag check errors.

Ticket: #15380

## Automation

/ok-to-test tags="@tag.All"

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/30322009616>
> Commit: 6b500e20ef3677ab381f6d04cca576cfb8d7d490
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=30322009616&attempt=1" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Tue, 28 Jul 2026 03:13:55 UTC

## Communication
- [ ] Yes
- [ ] No

## Summary by CodeRabbit

* **New Features**
  * Enhanced anonymous-user analytics controls using the `configure_block_event_tracking_for_anonymous_users` feature flag.
  * Usage pulse now uses a stable anonymous identifier with resilient local fallback behavior.

* **Bug Fixes**
  * Anonymous event tracking is now correctly suppressed when the flag blocks tracking, including fail-closed behavior when the flag can’t be evaluated.
  * Improved anonymous detection using the shared anonymous username constant.

* **Tests**
  * Added/extended tests covering anonymous tracking decisions, event suppression, and anonymous identifier fallback persistence.


subrata71 (commented):
This blocks Segment initialization for anonymous users when `configure_block_event_tracking_for_anonymous_users` is enabled, which looks correct for analytics tracking.

But `usage-pulse` still depends on `AnalyticsUtil.getAnonymousId()` when `enableTelemetry && segment.enabled` is true. In this new state, `segmentAnalytics` exists but has no Segment user, so `getAnonymousId()` can return `undefined`, and `/api/v1/usage-pulse` may stop sending `anonymousUserId`.

Can we also update usage-pulse to fall back to its local `FALLBACK_KEY`/`nanoid()` ID when `AnalyticsUtil.getAnonymousId()` is unavailable? That would preserve the PR’s stated scope that usage-pulse remains separate from Segment analytics blocking. Also could you please verify how does the PR behave in the EE repo?

salevine:
**Hardening: guard `localStorage` in the usage-pulse fallback id** (commit b23ab45479)

Follow-up to the usage-pulse fallback change. `getOrCreateFallbackAnonymousId()` in `app/client/src/usagePulse/utils.ts` writes to `localStorage`, which can throw when storage is unavailable (private/incognito mode, quota exceeded, or storage disabled by policy). Because this helper is now reached on the **telemetry-enabled** anonymous path (not just the telemetry-disabled one), an unguarded throw would propagate up through `getUsagePulsePayload` → `UsagePulse.sendPulse` and **drop the pulse entirely**.

Fix: wrap the read/write in a `try/catch`. On failure we return a non-persisted per-call `nanoid()` so the pulse still carries an `anonymousUserId` instead of failing. The trade-off is that the id isn't stable across pulses in that degraded environment — acceptable, since the alternative is losing the pulse.

```js
const getOrCreateFallbackAnonymousId = (): string => {
  try {
    let fallback = localStorage.getItem(FALLBACK_KEY);
    if (!fallback) {
      fallback = nanoid();
      localStorage.setItem(FALLBACK_KEY, fallback);
    }
    return fallback;
  } catch {
    // localStorage unavailable (private mode, quota, disabled) — return a
    // non-persisted per-call id so the pulse still carries an anonymousUserId.
    return nanoid();
  }
};
```

Covered by a new unit test (`utils.test.ts`) that mocks `Storage.prototype.setItem` to throw and asserts a non-empty `anonymousUserId` is still produced.

Verification: `yarn jest src/usagePulse/utils.test.ts src/usagePulse/usagePulse.test.ts` → 11/11 pass; `yarn run check-types` and `yarn eslint` on the changed files both clean.

salevine:
> This blocks Segment initialization for anonymous users when `configure_block_event_tracking_for_anonymous_users` is enabled, which looks correct for analytics tracking.
> 
> But `usage-pulse` still depends on `AnalyticsUtil.getAnonymousId()` when `enableTelemetry && segment.enabled` is true. In this new state, `segmentAnalytics` exists but has no Segment user, so `getAnonymousId()` can return `undefined`, and `/api/v1/usage-pulse` may stop sending `anonymousUserId`.
> 
> Can we also update usage-pulse to fall back to its local `FALLBACK_KEY`/`nanoid()` ID when `AnalyticsUtil.getAnonymousId()` is unavailable? That would preserve the PR’s stated scope that usage-pulse remains separate from Segment analytics blocking. Also could you please verify how does the PR behave in the EE repo?

@subrata71  - I updated it to address this

subrata71:
> > This blocks Segment initialization for anonymous users when `configure_block_event_tracking_for_anonymous_users` is enabled, which looks correct for analytics tracking.
> > But `usage-pulse` still depends on `AnalyticsUtil.getAnonymousId()` when `enableTelemetry && segment.enabled` is true. In this new state, `segmentAnalytics` exists but has no Segment user, so `getAnonymousId()` can return `undefined`, and `/api/v1/usage-pulse` may stop sending `anonymousUserId`.
> > Can we also update usage-pulse to fall back to its local `FALLBACK_KEY`/`nanoid()` ID when `AnalyticsUtil.getAnonymousId()` is unavailable? That would preserve the PR’s stated scope that usage-pulse remains separate from Segment analytics blocking. Also could you please verify how does the PR behave in the EE repo?
> 
> @subrata71 - I updated it to address this

@salevine I don't see any recent changes which addresses this. Forgot to push maybe?

subrata71:
@salevine just checking in here. I still don’t see the usage-pulse fallback commit on this PR as mentioned above.

salevine:
@subrata71 I re-checked this against the current PR head.

Your usage-pulse concern was valid before the last push: when anonymous Segment tracking is blocked, `AnalyticsUtil.getAnonymousId()` can be `undefined`, so `/api/v1/usage-pulse` needed to fall back to usage-pulse's own `FALLBACK_KEY` / `nanoid()` id.

The missing commit is now pushed on this PR: `b23ab45479` (`fix: keep usage-pulse anonymousUserId when Segment is blocked for anonymous users`). It updates `app/client/src/usagePulse/utils.ts` to prefer Segment's id when available and fall back to the locally persisted usage-pulse id when it is not. It also guards localStorage failures so the pulse still carries an `anonymousUserId`.

EE verification: I checked the EE client shape as well. There is no separate EE `usagePulse` implementation; `app/client/src/usagePulse/index.ts` calls the shared `getUsagePulsePayload`, and `ee/constants/UsagePulse` re-exports the CE constants. So the same fallback path applies in EE once this client change is present there.

Local verification:
- `yarn jest src/usagePulse/utils.test.ts src/usagePulse/usagePulse.test.ts` -> 11/11 passing
- `yarn eslint src/usagePulse/utils.ts src/usagePulse/utils.test.ts` -> passing

subrata71 (dismissed):
The changes look good to me. It would be great if you can confirm the following before merging this PR:
1. Anonymous page views stop sending Segment events
2. Usage-pulse /api/v1/usage-pulse still sends an anonymousUserId (the fallback path)
3. Logged-in user analytics remain unaffected
4. configure_block_event_tracking_for_anonymous_users feature flag creation

salevine:
Confirmed all four before merge:

Anonymous page views stop sending Segment events ✅ — gated in two places: the client no longer initializes Segment tracking for anonymous users when the flag is on (removed the licenseActive bypass, so licensed instances are covered too), and the server's sendEvent now blocks anonymous userIds before id resolution, closing the direct-caller path (published-app page views / action executions). Server gate fails closed.
Usage-pulse still sends an anonymousUserId ✅ — /api/v1/usage-pulse continues to carry an id. When Segment's id is unavailable it falls back to a locally-persisted id (and a per-call id if localStorage is unavailable), so billing/seat metering is unaffected. This channel is intentionally out of scope of the Segment block.
Logged-in user analytics unaffected ✅ — both client and server gates key strictly on the anonymous user; named-user events route through the normal path unchanged.
Feature flag — one clarification: configure_block_event_tracking_for_anonymous_users is not created in this PR; it already exists (introduced in [#40936](https://github.com/appsmithorg/appsmith/pull/40936)). This PR makes the existing flag fully honored. It remains opt-in (default false), so no behavior change until an admin enables it.
Test coverage looks good: userSagas.test.ts (shouldTrackUser matrix incl. the licensed-instance regression guard), usagePulse/utils.test.ts (fallback-id behavior), and new AnalyticsServiceCEImplTest cases (anonymous sendEvent enqueues nothing when the flag is on; fails closed on flag-resolution error).

subrata71 (approved):

PR #41980, opened 2026-07-10
feat: embedded MCP server with user-scoped token authentication (#15383)

salevine:
## What & why

Adds an **opt-out, same-instance Streamable HTTP MCP server** that lets an MCP client (e.g. an AI agent) act as a specific Appsmith user. The client authenticates with a **user-scoped bearer token**; the Node service forwards that token to the existing `/api/v1` endpoints, so Spring Security reconstructs the real user and the existing workspace/app/page ACLs authorize every operation. **No privileged or instance-wide credential is used.**

## How it works

```
MCP client --(Bearer mcp_… token)--> Caddy /mcp --> Node MCP service --(forwards same token)--> /api/v1 --> Spring Security (real user) --> existing ACLs
```

### Server (CE, EE-overridable)
- `UserMcpToken` domain + repository + service + `McpTokenController` for **create / list / revoke** of user-scoped tokens. Every layer follows the CE-base + thin concrete-subclass split (`*CE` / `*CEImpl`) so EE can override.
- Tokens are **SHA-256 pre-hashed then bcrypt-hashed at rest** (avoids bcrypt's 72-byte truncation), plaintext returned exactly once, **max 10 active tokens/user**.
- A bearer `AuthenticationWebFilter` (only engages for the `mcp_` prefix) reconstructs the token owner. Invalid / revoked / disabled-user tokens return **401**.
- `Migration076` creates the `userMcpToken` indexes (the instance runs with `auto-index-creation=false`, so `@Indexed` alone is inert).

### Node service (`app/client/packages/mcp`)
- Streamable HTTP transport bound to **loopback only**, `/health` endpoint, request-body size cap, **per-request token revalidation**, **per-session token binding** (constant-time compare), and **per-user + global session caps**.
- Tools: `list_workspaces`, `list_applications`, `get_application_context`, and `import_application_artifact` / `import_partial_application_artifact` — writes go through the **existing validated import / partial-import APIs** (no raw DSL/Mongo writes).

### Client
- MCP token management UI in the user profile (create / copy-once / revoke).

### Rollout / deploy
- **Disabled by default — an admin opts in.** `APPSMITH_MCP_ENABLED`, `APPSMITH_MCP_DATA_ENABLED` and `APPSMITH_MCP_JS_ENABLED` all default to **off**, using allow-list semantics (`true|1|yes|on`): absent, blank, and unrecognized values all mean disabled, so a typo in `docker.env` fails safe. **Existing instances are unaffected by an upgrade** — the entrypoint backfills the gates as `false`. Enable from **Admin Settings → MCP Server**.
- **While MCP is off:** `/mcp` returns a 404 `{"code":"mcp_disabled"}` JSON (not the SPA's HTML); an already-issued `mcp_…` token is rejected server-side with **401**, evaluated per request so the kill switch is immediate and needs no restart; users **cannot create or rotate** tokens (`AE-MCP-5049`), while `list`/`revoke` stay available so existing tokens can still be cleaned up; and the Profile → MCP tokens section is hidden.
- MCP is never fatal to container health unless explicitly enabled, its logs reach `docker logs`/`kubectl logs`, and they are size-capped on the stacks volume.
- Dockerfile copy, dedicated `mcp-build` CI workflow, and a `/mcp/health` route test.

## Testing
- **Java**: unit tests for token issuance/auth/revocation (real BCrypt), the converter/manager, and a WebFilter test asserting invalid MCP bearer → 401 and non-MCP bearer → passthrough.
- **Node**: 13 tests — token forwarding, auth (missing / non-`mcp_` / anonymous → 401), session binding & revalidation, TTL/expiry, per-user (429) and global (503) caps, artifact validation, malformed/oversized body handling. All pass.
- Ran Spotless, client `check-types`, ESLint (0 errors), and the MCP package typecheck locally.

## Security review
Reviewed by a multi-agent council (architecture, security, QA, data-migration, DX, UX, product). Key hardening landed from that review: 401 (not 500) on bad tokens, the CE/EE split, the index migration, the `mcp_`-prefix + non-anonymous `/me` gate (closes an unauthenticated-session DoS), per-user session cap, and the opt-in deploy gate.

## Follow-ups (tracked, not blocking an opt-in ship)
- Gate the client "MCP Tokens" profile tab on the same enablement signal (needs the flag surfaced to the client).
- End-user connection snippet (endpoint URL + client config) near the token UI.
- Minor UX polish (revoke-failure toast, monospace token field, destructive-button styling).

## Automation
/ok-to-test tags="@tag.All"

## Summary by CodeRabbit

* **New Features**
  * Added an MCP server for authenticated app building, editing, data access, publishing, and Git workflows.
  * Added secure token creation, listing, rotation, revocation, naming, expiry, and client configuration copying.
  * Added admin controls for MCP, data tools, restricted JavaScript, and token lifetime.
  * Added structured layouts, widgets, themes, events, presets, queries, governance safeguards, confirmations, audit history, rate limits, and session controls.
  * Integrated MCP artifacts into Docker images and release builds.

* **Documentation**
  * Added MCP setup, configuration, capabilities, and workflow documentation.

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/31097538595>
> Commit: ffc74086e2d242e38e6a3003d1ce075de4e1a1d3
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=31097538595&attempt=2" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Thu, 06 Aug 2026 12:56:37 UTC


app/client/packages/mcp/src/app.ts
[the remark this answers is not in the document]
salevine: addressed

deploy/docker/fs/opt/appsmith/healthcheck.sh
```diff
@@ -1,5 +1,11 @@
 #!/usr/bin/env bash
 healthy=true
+# MCP is opt-in (APPSMITH_MCP_ENABLED); its supervisord program is absent when disabled. Only health-check it when
+# it is actually a configured program, so a container with MCP off (the default) is not reported unhealthy.
+processes="editor rts backend"
+if supervisorctl status | grep -q '^mcp'; then
+  processes="$processes mcp"
+fi
```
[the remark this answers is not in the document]
salevine: addressed

app/client/packages/mcp/src/app.ts
```diff
@@ -0,0 +1,480 @@
...
+          sessionIdGenerator: randomUUID,
+          onsessioninitialized: (id) => {
+            sessions.set(id, {
+              expiresAt: now() + sessionTtlMs,
+              token,
+              username,
+              transport: transport!,
+            });
+          },
+        });
```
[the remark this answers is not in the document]
salevine: fixed

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/McpTokenControllerCE.java
```diff
@@ -0,0 +1,78 @@
...
+    public Mono<ResponseDTO<McpTokenResponseDTO>> rotate(
+            @AuthenticationPrincipal User user, @PathVariable String tokenId) {
+        // Rotation returns a replacement bearer secret, so MCP-authenticated calls must not be able to use a leaked
+        // token to perpetuate access beyond its intended revocation window.
+        return requireSessionAuthentication()
+                .then(userMcpTokenService.rotate(user, tokenId))
+                .map(token -> new ResponseDTO<>(HttpStatus.OK, token));
+    }
+
+    private static boolean isMcpAuthenticated(Authentication authentication) {
```
sondermanish: This should go to some utilsMethod and should be imported as static methods from there, instead of staying in controller

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/McpTokenControllerCE.java
```diff
@@ -0,0 +1,78 @@
...
+        return authentication != null
+                && authentication.getAuthorities().stream()
+                        .anyMatch(authority -> McpTokenAuthentication.MCP_AUTHORITY.equals(authority.getAuthority()));
+    }
+
+    private Mono<Void> requireSessionAuthentication() {
```
sondermanish, line 96: would this make more sense from webfilter override?

app/server/appsmith-server/src/main/java/com/appsmith/server/ratelimiting/ce/RateLimitServiceCEImpl.java
```diff
@@ -59,6 +59,18 @@ public Mono<Boolean> tryIncreaseCounter(String apiIdentifier, String userIdentif
                 .subscribeOn(LoadShifter.elasticScheduler);
     }
 
+    @Override
+    public Mono<Boolean> isRateLimitExceeded(String apiIdentifier, String userIdentifier) {
+        return sanitizeInput(apiIdentifier, userIdentifier)
+                .map(isInputValid -> rateLimitConfig
+                                .getOrCreateAPIUserSpecificBucket(apiIdentifier, userIdentifier)
+                                .getAvailableTokens()
+                        == 0)
```
sondermanish: this check should be a bit more stronger, i.e. gte zero

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration076AddUserMcpTokenIndexes.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    @RollbackExecution
+    public void rollbackExecution() {
+        // Indexes are retained to avoid degrading token authentication or token-management queries.
+    }
+
+    @Execution
+    public void addUserMcpTokenIndexes() {
+        Index tokenIdIndex = makeIndex(UserMcpTokenCE.Fields.tokenId)
```
sondermanish, line 30: let's also add a combined tokenId and userId index as well, since that would be used?

sondermanish (changes requested):
@salevine I believe that it would be beneficial if we could break this pr into 2-3 pieces as follows
- server changes which introduces changes for config, web filters and exchanges
- client changes along with token config screen
- workflows, and docker changes

otherwise it's getting difficult to review and key insights would be lost

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/McpTokenControllerCE.java
salevine: Done — moved the check into a shared static `McpTokenAuthentication.isMcpPrincipal(Authentication)` and both the controller and `McpAllowlistWebFilter` now call it. The filter had a near-identical private copy, so this also collapses the duplication (and the drift risk — the controller copy wasn't checking `isAuthenticated()`, the filter's was; the shared one does).

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/McpTokenControllerCE.java
salevine: The webfilter already enforces this: `McpAllowlistWebFilter` deliberately omits every `/api/v1/users/mcp-tokens` route from its allowlist, so an MCP-token principal is 403'd at the filter before it ever reaches this controller. The `requireSessionAuthentication()` check here is intentional defense-in-depth (double coverage — it's called out in the filter's comment), and it now runs through the same shared `isMcpPrincipal` predicate.

Happy to drop the controller-level check and rely solely on the filter if you'd rather it live in one place — I kept both because the security review valued the redundancy on the token-minting path.

app/server/appsmith-server/src/main/java/com/appsmith/server/ratelimiting/ce/RateLimitServiceCEImpl.java
salevine: Done — changed to `<= 0` so an exhausted bucket is still reported as exhausted even if `getAvailableTokens()` ever returns a non-positive value.

app/server/appsmith-server/src/main/java/com/appsmith/server/migrations/db/ce/Migration076AddUserMcpTokenIndexes.java
salevine: I dug into this and don't think the compound index earns its keep. `tokenId` already has a unique index (`user_mcp_token_token_id`), so `findByTokenIdAndUserIdAndDeletedAtIsNull` resolves through it to at most one document and the `userId`/`deletedAt` predicates just filter that single doc — a `(tokenId, userId)` index can't beat a unique single-key lookup and would only add write/storage overhead. The userId-only queries (`findAllByUserIdAndDeletedAtIsNull`, `countByUserIdAndDeletedAtIsNull`) are already served by `user_mcp_token_user_deleted_at`.

Left it out for that reason — but if you're seeing a query pattern I'm missing, happy to add it.

app/client/packages/mcp/src/app.ts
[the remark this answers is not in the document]
salevine: !fixed

`isReadOnlyAction` no longer trusts `actionConfiguration.httpMethod` on a host-restricted (DB) action. Because the backend does not strip an injected `httpMethod`, a mutating SQL action stamped with `"httpMethod":"GET"` previously classified as read-only and auto-ran through `run_action` without the `prepare_run_action` → `confirm_run_action` gate. It now returns false for host-restricted actions (and for external-egress actions), so the only remaining read auto-run door is the Google Sheets path — admitted only after strict server-authoritative plugin **and** datasource resolution, never the action document's own strings.

The test that had asserted a DB+GET action auto-runs is flipped into a regression test asserting `confirmation_required` (a `DROP TABLE` body with an injected `httpMethod:"GET"` now routes through the confirm gate and `executeAction` is never called).

Fixed in 4f833b59e4.

.github/workflows/mcp-build.yml
```diff
@@ -0,0 +1,142 @@
...
+      github.event_name == 'push' ||
+      github.event_name == 'workflow_dispatch' ||
+      github.event_name == 'repository_dispatch' ||
+      github.event_name == 'schedule' ||
+      github.event_name == 'release'
+    # Mongo + Redis back the governance store's distributed-lock / confirmation-atomicity safety net. Without these
+    # service containers store.integration.test.ts stays describe.skip and that safety net never runs in CI.
+    services:
+      mongodb:
+        image: mongo:6.0
```
sondermanish: nit: could we use mongo similar to version which we are running on appsmith, does not make a huge diff though

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserMcpTokenServiceCEImpl.java
```diff
@@ -0,0 +1,236 @@
...
+    public Mono<User> authenticate(String token) {
+        String tokenId = extractTokenId(token);
+        if (tokenId == null) {
+            return Mono.empty();
+        }
+
+        return userMcpTokenRepository
+                .findByTokenIdAndDeletedAtIsNull(tokenId)
+                .filter(storedToken -> matchesStoredHash(token, storedToken.getTokenHash()))
+                .filter(UserMcpTokenServiceCEImpl::isNotExpired)
```
sondermanish: we should not directly call the CE services, always, route through the final interface

.github/workflows/mcp-build.yml
```diff
@@ -0,0 +1,108 @@
...
+        run: yarn check-types
+
+      - name: Lint
+        run: yarn lint
+
+      - name: Check formatting
+        run: yarn prettier
+
+      - name: Lint
+        run: yarn lint
```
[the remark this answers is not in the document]
salevine: Fixed in b55ba0a — `mcp-build.yml` now runs a single `Lint` step followed by `Check formatting`. Verified against the current file: only one `yarn lint` invocation remains (line 109).

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/McpTokenControllerCE.java
salevine: Confirming this one is landed: the check now lives in `McpTokenAuthentication.isMcpPrincipal(Authentication)` and is called from both `McpTokenControllerCE` (line 63) and `McpAllowlistWebFilter` (line 102), so there's no duplicated private copy left to drift.

app/server/appsmith-server/src/main/java/com/appsmith/server/controllers/ce/McpTokenControllerCE.java
salevine: Closing this out with the current design: keeping both layers.

`McpAllowlistWebFilter` is the primary gate — it deliberately omits every `/api/v1/users/mcp-tokens` route from its allowlist, so an MCP-token principal is rejected there before it ever reaches this controller. The `requireSessionAuthentication()` check here is intentional defense-in-depth on the token-minting path specifically, and both sides now go through the same shared `McpTokenAuthentication.isMcpPrincipal` predicate, so the two can't drift apart.

Still happy to drop the controller-level check and rely solely on the filter if you'd prefer it in one place — just say the word and I'll do it as a follow-up.

app/server/appsmith-server/src/main/java/com/appsmith/server/ratelimiting/ce/RateLimitServiceCEImpl.java
```diff
@@ -59,6 +59,21 @@ public Mono<Boolean> tryIncreaseCounter(String apiIdentifier, String userIdentif
...
+                // correct even if getAvailableTokens() ever reports a non-positive value, so we never mis-report an
+                // exhausted bucket as having capacity.
+                .map(isInputValid -> rateLimitConfig
+                                .getOrCreateAPIUserSpecificBucket(apiIdentifier, userIdentifier)
+                                .getAvailableTokens()
+                        <= 0)
+                // Since we are interacting with redis, we want to make sure that the operation is done on a separate
+                // thread pool
+                .subscribeOn(LoadShifter.elasticScheduler);
+    }
```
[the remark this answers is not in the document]
salevine: !fixed — removed `isRateLimitExceeded` from `RateLimitServiceCE` and `RateLimitServiceCEImpl` entirely.

The finding is correct about the primitive. The MCP auth path it points at was already reworked in 40ecdd8ca2: `McpTokenAuthenticationManager` now spends a token via the atomic `tryIncreaseCounter` (a single `BucketProxy.tryConsume`) **before** the credential check, so N concurrent requests get at most `capacity` credential verifications no matter how they interleave.

That left `isRateLimitExceeded` with zero production callers. Rather than leave an observe-without-consume primitive sitting on a shared interface for a future caller to misuse the same way, it's deleted. It was introduced on this branch and never shipped on `release`, and there are zero remaining references in either the CE or EE repo, so nothing external depends on it.

On the regression-test criterion: a deleted method can't carry one, but the ordering that actually fixes the reported behavior is pinned by two tests that shipped with the fix — `McpTokenAuthenticationManagerTest.authenticate_spendsRateLimitTokenBeforeVerifyingCredential` and `authenticate_exhaustedBucketPerformsNoCredentialVerification` (the latter asserts an exhausted bucket performs no token lookup at all).

.github/workflows/mcp-build.yml
salevine: Good catch — bumped to `mongo:7.0` to match what we actually ship: `deploy/docker/base.dockerfile` installs `mongodb-org` 7.0. Worth having regardless, since 6.0 is past end-of-life. Redis is already on major 7 in both places (`redis:7` here vs `redis:7.4.8` in the base image).

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserMcpTokenServiceCEImpl.java
salevine: Fixed — `.filter(UserMcpTokenServiceCEImpl::isNotExpired)` is now `.filter(this::isNotExpired)`, with `isNotExpired` changed from `private static` to a `protected` instance method.

`protected` rather than `private` is deliberate: private methods aren't virtual in Java, so `this::isNotExpired` on a private method would still bind statically to the CE implementation and an EE override would be silently ignored — which would defeat the point. `protected` is what actually makes it a real CE extension point. Since that makes an auth-critical predicate overridable, I've added a comment on it stating that any override must remain fail-closed (a token with a null expiry must still be rejected).

Worth noting this was the only `CEImpl::` static reference anywhere in the server source, so it brings the file in line with the rest of the codebase.

app/server/appsmith-server/src/main/java/com/appsmith/server/ratelimiting/ce/RateLimitServiceCEImpl.java
salevine: Updating this thread — my reply above is now out of date. `isRateLimitExceeded` has since been removed entirely (from both `RateLimitServiceCE` and `RateLimitServiceCEImpl`), so the `<= 0` check no longer exists.

Context: a separate review flagged the method as a TOCTOU-prone rate-limit gate, and by then its only production caller (`McpTokenAuthenticationManager`) had already been reworked to use the atomic `tryIncreaseCounter` before the credential check. With zero callers left, deleting it beat leaving an observe-without-consume primitive on a shared interface. I've added a note on `tryIncreaseCounter` recording why a non-consuming availability check isn't a valid gate, so it doesn't get reinvented.

deploy/docker/fs/opt/appsmith/healthcheck.sh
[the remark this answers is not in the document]
salevine: Re-checked this against the baseline, and it isn't a regression — `mongo` and `redis` were already excluded before this PR.

`origin/release` line 35 reads:

```bash
done <<< $(supervisorctl status editor rts backend)
```

So the `mongo` and `redis` branches in that loop were already unreachable on `release`; this PR changed `editor rts backend` to `$processes`, which is `editor rts backend` plus `mcp` when the program is present. The set of monitored programs is therefore a strict superset of what it was — nothing stopped being health-checked here.

Worth fixing on its own merits (embedded mongo/redis genuinely aren't probed), but it's pre-existing and orthogonal to MCP, so I'm not folding it into this PR. Flagging it separately rather than silently.

deploy/docker/fs/opt/appsmith/healthcheck.sh
```diff
@@ -1,5 +1,11 @@
 #!/usr/bin/env bash
 healthy=true
+# MCP is opt-in (APPSMITH_MCP_ENABLED); its supervisord program is absent when disabled. Only health-check it when
+# it is actually a configured program, so a container with MCP off (the default) is not reported unhealthy.
+processes="editor rts backend"
```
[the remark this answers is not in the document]
salevine: This one is **correct and still live** — the branch really is dead. `processes` lists `backend`, the branch tests `"$process" == "server"`, so the `/api/v1/health` probe never runs and the backend only gets the RUNNING check. A second reviewer reached the same conclusion independently this week.

I'm deliberately **not** fixing it in this PR, for a reason worth stating rather than leaving implicit.

It's pre-existing: `origin/release` line 18 carries the identical `elif [[ "$process" == "server" ]]` branch, so the probe has been dormant for as long as the file has looked like this. Renaming it to `backend` doesn't just correct a typo — it *activates* an HTTP health probe on every existing Docker install. From that upgrade onward, a container whose backend is briefly slow to answer `/api/v1/health` reports unhealthy and gets restarted by Compose/Swarm/ECS/Nomad. That may well be the behaviour we want, but it's a health-semantics change for the whole self-hosted install base, it needs its own release note, and it has nothing to do with MCP. Shipping it inside an MCP feature PR is exactly how that kind of change reaches operators unannounced.

For what it's worth, this PR took the opposite care in the same file: MCP contributes to `healthy=false` only when an operator explicitly enabled MCP, precisely so a new subsystem can't take down an otherwise-working instance.

Happy to send it as a standalone PR with a release note — say the word.

salevine:
## CodeRabbit findings — re-audited against current HEAD

CodeRabbit auto-paused reviews on this branch ("under active development"), so its findings predate a lot of what has since landed and all its threads were sitting resolved. Rather than trust the resolved flag, I re-verified every finding against the current code — prompted by finding one today that was marked resolved but only half-fixed (see the `runId` row).

| Finding | Status at HEAD |
|---|---|
| esbuild `target` built `node20.11.0` | **Fixed** — `node${process.versions.node.split(".")[0]}` |
| No Origin/Host validation on `/mcp` | **Fixed, more strictly than suggested** — any request carrying a browser `Origin` header is rejected 403 outright, plus optional `APPSMITH_MCP_ALLOWED_HOSTS` |
| `extractTokenId` threw on a null token | **Fixed** — `token == null \|\| !token.startsWith(...)` |
| `process.exitCode` didn't terminate the process | **Fixed** — `process.exit(1)` in both `uncaughtException` and `unhandledRejection` |
| Outbound backend requests had no timeout | **Fixed** — one shared `request()` helper wraps every upstream call in an `AbortController`, 30s default (longer budget passed for git commit) |
| `placementSchema` allowed `after` + `inside` together | **Fixed** — `.refine` enforces mutual exclusivity |
| Container height not synced for `inside` edits | **Fixed** — `cascadeFit` grows the enclosing container/form/tabs chain recursively |
| Shared `unknown` stub tripped strict Mockito | **Fixed** — `lenient()` |
| `not.toEqual(arrayContaining([...]))` passed if *any* item was absent | **Fixed** — individual `not.toContain` per tool, with a comment explaining why |
| Prettier/ESLint failures across `app.ts` / `app.test.ts` | **Fixed** — `prettier --check .` and `eslint .` both clean |
| Duplicate `yarn lint` step | **Fixed** |
| `runId == 0` unquoted | **Was only half-fixed.** Corrected in `build-client-server.yml` but the identical `mcp-build` job in `build-client-server-count.yml` still had it — the only unquoted comparison of 14 in the repo. Fixed today in `5bb59a795c` |
| `mongo`/`redis` dropped from health monitoring | **Not a regression** — `origin/release` already had `supervisorctl status editor rts backend`; those branches were dead before this PR. Replied on the thread |
| Health branch tests `server` but the program is `backend` | **Correct and still live, deliberately deferred** — pre-existing on `release`; fixing it activates a dormant HTTP probe for every existing Docker install, which is a health-semantics change needing its own release note. Replied on the thread |

Two of these needed no code change but did need a correction to the record, so I've replied on those threads rather than leaving them silently resolved.

Since CodeRabbit hasn't seen the last three commits (`5bb59a795c`, `5a498c408e`, `c16f06c92e` — council fixes, and flipping every MCP gate to off-by-default), I'm asking it for a fresh pass below.

salevine:
@coderabbitai review

Reviews were auto-paused on this branch while it was under active development. It's settled now — the last three commits (`5bb59a795c`, `5a498c408e`, `c16f06c92e`) address a full nine-seat review and flip every MCP gate to off-by-default, and none of them have been looked at. Please take a fresh pass.

app/server/appsmith-server/src/main/java/com/appsmith/server/filters/McpAllowlistWebFilter.java
```diff
@@ -0,0 +1,166 @@
...
+    private static boolean hasReservedFinalSegment(String path) {
+        String trimmed = path.endsWith("/") ? path.substring(0, path.length() - 1) : path;
+        int lastSlash = trimmed.lastIndexOf('/');
+        if (lastSlash < 0 || lastSlash == trimmed.length() - 1) {
+            return false;
+        }
+        // Case-insensitive: the allowlist must not be evadable by casing, even though WebFlux routing is
+        // case-sensitive and a mis-cased literal would 404 rather than reach the sibling handler.
+        return RESERVED_ROUTE_SEGMENTS.contains(trimmed.substring(lastSlash + 1).toLowerCase(Locale.ROOT));
+    }
```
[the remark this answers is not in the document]
salevine: !fixed — real bypass, reproduced before fixing. Closed in `2d97b041a7`.

Confirmed the mechanism exactly as described: WebFlux splits matrix parameters off a path segment, so `PathPattern` matched `/api/v1/actions/move;bypass=true` against the `/api/v1/actions/{actionId}` rule as `move`, while the guard's raw-string scan saw `move;bypass=true`, didn't recognize the reserved literal, and allowed the request. The guard and the matcher were reading different values off the same request — the string scan was the bug, not the reserved-segment approach.

I wrote the failing test first and watched it fail on the unpatched code (`PUT /api/v1/actions/move;bypass=true` was allowed through), rather than fixing on the strength of the report.

**Fix:** read the parsed `PathContainer` and compare `PathContainer.PathSegment.valueToMatch()` — the very value `PathPattern` matches on — instead of doing string surgery on the raw path. The guard and the matcher now consume the same parsed value, so this class of divergence is impossible by construction rather than patched case by case. That also covers variants the original scan would have needed separate handling for (multiple matrix params, `;` combined with casing evasion).

**Regression tests** in `McpAllowlistWebFilterTest`: `move;bypass=true`, `refactor;a=1;b=2`, and `MoVe;x=1` (matrix + casing together, since they're independent knobs). All fail on the unpatched guard. The existing coverage still holds — a legitimate ObjectId-shaped `actionId` passes through, and bare `move`/`refactor` stay denied.

Worth noting the guard is defence-in-depth rather than the only control: every one of these routes is still ACL-enforced under the caller's own user, so the bypass widened the reachable surface but did not grant anything the token's owner couldn't already do through the UI. Fixed regardless — the allowlist should mean what its comment says.

app/client/packages/mcp/.env.example
```diff
@@ -0,0 +1,73 @@
...
+# Note what APPSMITH_MCP_ENABLED does NOT do here: standalone, `start-server.sh` execs the bundle unconditionally
+# and `server.ts` reads only the DATA and JS gates below — the top-level gate is enforced by run-mcp.sh, Caddy, and
+# SecurityConfig, i.e. only inside the container. So on this path it is a documentation marker; the two gates that
+# actually change standalone behaviour are the DATA and JS ones.
+#
+# For a local dev loop against the data/JS tools (including live-harness.mjs, which now exits non-zero on a failed
+# check), set APPSMITH_MCP_DATA_ENABLED=1 and APPSMITH_MCP_JS_ENABLED=1 below.
+APPSMITH_MCP_ENABLED=0
+APPSMITH_MCP_PORT=8092
+APPSMITH_API_BASE_URL=http://127.0.0.1:8080
```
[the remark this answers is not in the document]
salevine: Valid, and trivial — `dotenv-linter`'s `UnorderedKey` wants `APPSMITH_API_BASE_URL` (currently line 18) above `APPSMITH_MCP_ENABLED` (line 16), values unchanged.

Not applying it in this pass, because `.env.example` has uncommitted changes in flight from the work that's addressing the rest of this review round (the gate defaults in this same file are already flipped to `0` locally). Editing it from a second place would risk clobbering that. It'll be folded into that change rather than landed separately.

Flagging rather than silently skipping, since a lint warning left alone tends to stay alone.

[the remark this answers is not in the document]
salevine: Fixed in `1c0822cab1` — `APPSMITH_API_BASE_URL` now sits above `APPSMITH_MCP_ENABLED`, values and comments unchanged.

Worth noting why only that one line moved: `dotenv-linter` evaluates ordering within blank-line-separated groups, so the first block was the only one out of order. `APPSMITH_MCP_DATA_ENABLED` and `APPSMITH_MCP_JS_ENABLED` each sit alone in their own commented group and were already clean — reordering them into one alphabetical list would have separated each gate from the comment explaining it, for no linting gain.

I'd held this back on the earlier pass because the file had uncommitted changes in flight from the work addressing the rest of this review round; that work has since landed, so this is now a clean one-line follow-up rather than a risk of clobbering it.

app/server/appsmith-server/src/main/java/com/appsmith/server/services/ce/UserServiceCEImpl.java
```diff
@@ -114,6 +116,22 @@ public class UserServiceCEImpl extends BaseService<UserRepository, User, String>
     private final SecureBaseUrlResolver secureBaseUrlResolver;
     private final CacheablePylonHelper cacheablePylonHelper;
 
+    // Same allow-list gate and default (OFF) as SecurityConfig, the token controller, and the deploy scripts.
+    // Setter-injected so the existing constructor wiring (and every subclass/test that calls it) stays unchanged.
+    private String mcpEnabledFlag;
+
+    @Value("${APPSMITH_MCP_ENABLED:false}")
```
sondermanish, line 123: Ideally, this should have come from config files instead of direct landing on the service file

sondermanish (approved):

sondermanish:
https://github.com/appsmithorg/appsmith/pull/42098/ created a revert pr as this has EE failures.

PR #41985, opened 2026-07-13
fix(security): restrict datasource access to the instance's own network address

wyattwalter:
## What

Adds a defense-in-depth SSRF protection so a user-defined datasource cannot connect
to the Appsmith instance's own network address.

## Why

The SSRF host filter (`RestrictedHostFilter`) intentionally allows private-network
(RFC 1918 / site-local) addresses so that legitimate customer datasources on private
networks keep working. In a typical Docker/Kubernetes deployment the instance's own
routable address is itself a private/site-local address, so it falls inside that
allowed range. As a hardening measure, this change blocks that one target — the
instance's own address — without restricting the rest of the private network.

## How

- Registers the instance's own hostname(s) at startup (from the local hostname and the
  `HOSTNAME` environment variable) and resolves them to their current address(es),
  mirroring the existing internal-Redis host registration. Resolution is live per check,
  so an instance whose address changes stays covered with no cached-address bookkeeping.
- Blocks a datasource whose target is a registered own hostname, or resolves to one of
  the instance's own addresses. Enforced consistently across the datasource egress paths
  — WebClient (REST/GraphQL/SaaS), Elasticsearch, Redis, and datasource validation —
  evaluating the address already resolved by each path, so no additional lookup of the
  user-supplied host is introduced.
- Only the instance's own address(es) are affected; other private-network datasources
  continue to work unchanged.
- Coverage is best-effort for the address(es) the registered hostname(s) resolve to
  (the primary, hostname-mapped address), not a full network-interface enumeration.

## Testing

- Unit tests cover rejection of the own address through each enforcement entry point
  (raw IPv4 literal, IPv4-mapped IPv6, and hostname), confirm other private-network
  addresses stay allowed, and confirm the existing SSRF kill-switch still applies.
  Own-address resolution is injected in tests, so the suite has no external DNS
  dependency.
- `appsmith-interfaces` module tests pass.

## Summary by CodeRabbit

* **Security Enhancements**
  * Strengthened SSRF protection by blocking outbound requests that target the application’s own hostname or any resolved IP addresses associated with it.
  * Expanded detection to cover multiple address representations (including IPv4 literals and IPv4-mapped IPv6).
  * Added startup registration of the instance’s own hostname(s) and environment hostname with graceful fallback if resolution fails.
  * Preserved the existing allow/deny behavior for other private network targets and the SSRF bypass setting.
* **Tests**
  * Added deterministic test coverage for own-host blocking, caching behavior, and SSRF bypass across all relevant enforcement paths.

> [!WARNING]
> Tests have not run on the HEAD 1d3d64dc953d7fdf48e1959200396688e9335d35 yet
> <hr>Mon, 13 Jul 2026 19:13:41 UTC

### DIFF
--- a/app/server/appsmith-interfaces/src/main/java/com/appsmith/util/RestrictedHostFilter.java
+++ b/app/server/appsmith-interfaces/src/main/java/com/appsmith/util/RestrictedHostFilter.java
@@ -97,6 +97,28 @@ public final class RestrictedHostFilter {
     private static volatile Set<String> internalRedisHosts = computeInternalRedisHosts();
 
     /**
+     * Hostnames that identify the Appsmith instance itself — seeded at JVM start from the local
+     * hostname ({@link InetAddress#getLocalHost()}) and the {@code HOSTNAME} environment variable.
+     *
+     * <p>Defense-in-depth: the instance's own routable address is always an RFC 1918 / site-local
+     * IP (Docker bridge {@code 172.17.x}, k8s pod {@code 10.x}, etc.), which the filter otherwise
+     * intentionally allows so that legitimate customer datasources on private networks keep
+     * working (see {@link #resolveIfAllowed(String)}). That carve-out means a user-defined
+     * datasource could point at the Appsmith instance itself. Registering the instance's own
+     * hostname(s) here lets the filter resolve and block just its own address(es) — via the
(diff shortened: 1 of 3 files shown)


subrata71 (approved):
LGTM

PR #41986, opened 2026-07-13
fix(workflows): use continue instead of break in close-labeler loop

### LINKED ISSUE #41983: close-labeler workflow uses break instead of continue, dropping the QA label on multi-issue PRs
### Summary

`.github/workflows/close-labeler.yml` uses `break` where it needs `continue`, so a PR that closes more than one issue can silently miss getting the `QA` label.

The loop over the PR's closing issues is:

```js
for (const node of result.repository.pullRequest.closingIssuesReferences.nodes) {
  const shouldQA = node.labels.edges.find(edge => edge.node.name === "Enhancement" || edge.node.name === "Bug") != null
  if (!shouldQA) {
    break        // <-- aborts the whole loop on the first non-Bug/Enhancement issue
  }
  // ... add "QA" label
}
```

Because it `break`s (rather than `continue`s) on the first issue that isn't a `Bug`/`Enhancement`, any `Bug`/`Enhancement` issues that come **after** it in the list are never reached, so the `QA` label isn't added.

### Steps to reproduce

A PR whose `Closes #…` references resolve, in order, to `[Documentation, Bug, Enhancement]`:
- current (`break`): QA label added to **none**
- expected (`continue`): QA label added to the Bug and the Enhancement

I ran the exact loop body in Node to confirm the order-dependent difference.

### Expected behavior

Skip the non-qualifying issue and keep checking the rest.

### Fix

Change the one word `break` → `continue`. Happy to open the PR if useful.

<sub>AI-assisted; I verified the workflow logic myself.</sub>

moosairfaan:
Fixes #41983
The close-labeler loop used break instead of continue, so any qualifying (Bug/Enhancement) issue that came after a non-qualifying one in a PR's closing issues list never got the QA label. Changed break to continue so the loop skips non-qualifying issues instead of aborting.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automated labeling so all eligible referenced issues are processed.
  * Issues meeting the required criteria now receive the appropriate QA label without being skipped due to earlier non-matching issues.


moosairfaan:
Bumping this — the fix is small and ready for review whenever a maintainer has a chance. Happy to make any changes if needed!

PR #42013, opened 2026-07-17
fix(client): show wildcard frame-ancestors value as allow-everywhere in embed settings

wyattwalter:
## What & why

The **Admin → Configuration** embed setting picks which radio to show for the stored `APPSMITH_ALLOWED_FRAME_ANCESTORS` value. Previously only the exact string `*` mapped to **"Allow embedding everywhere"**; any other value fell through to **"Limit embedding to certain URLs"** and rendered its tokens as chips.

In a CSP `frame-ancestors` policy a **bare `*` token matches every origin and overrides the other sources**, so a value like the shipped default `'self' *` was functionally allow-everywhere while the UI claimed embedding was limited. An admin could add `*` to the limit list and silently reopen the instance to all origins while the UI reassured them it was restricted.

## Changes

- **Honest display** — `formatEmbedSettings` now treats any value containing a standalone `*` token (whitespace-split) as **allow-everywhere**, so an existing `'self' *` shows the truthful "Allow embedding everywhere" radio instead of a contradictory "Limit" state. A host wildcard like `https://*.example.com` only matches subdomains of a specific host and is still treated as a limit-list entry.
- **Input prevention** — in "Limit embedding to certain URLs" mode, adding a bare `*` chip is rejected with an inline message steering the admin to the "Allow embedding everywhere" radio. The check is whitespace-aware per chip, so a **pasted** value that arrives as a single chip (e.g. `'self' *` or `* https://a.com`) is caught too. Legitimate host wildcards such as `https://*.example.com` are preserved.
- **Round-trip** — the `parse` localStorage guard uses the same bare-`*` check, so an allow-all value is never remembered as a limit list.

## Tests

Unit tests for `formatEmbedSettings` cover: exact `*`, `'self' *` and `* 'self'` → allow-everywhere; `'none'` → disable; `'self'` and `'self' https://a.com` → limit; `https://*.example.com` → limit (host wildcard preserved); empty → limit/empty. Additional tests cover the input guard, including pasted single-chip `'self' *` and `* https://a.com`.

## Scope

Client-only change. No server, Caddy, or `docker.env.sh` changes.

## Automation

/ok-to-test tags="@tag.All"

## Issue

https://linear.app/appsmith/issue/APP-15353

## Summary by CodeRabbit

* **New Features**
  * Enhanced the embedded-page “Embed settings” control for CSP `frame-ancestors` with dedicated formatting and validation, including clearer guidance for bare `*` and `none`/disable cases.
* **Bug Fixes**
  * Correctly treats a standalone bare `*` token anywhere as “Allow embedding everywhere”.
  * Sanitizes restricted-mode values so contradictory allow-all/disable keywords are not persisted, while keeping valid host wildcard entries.
  * Normalizes `self`/`none` keyword formatting for safer, consistent saving.
* **Tests**
  * Expanded Jest coverage for token parsing/formatting/sanitization and Admin Settings radio behavior.

> [!TIP]
> Workflow run: <https://github.com/appsmithorg/appsmith/actions/runs/31499794770>
> Commit: fd85c3f6cb8635d74270a0dcb4752aa426710712
> <a href="https://internal.appsmith.com/app/cypress-dashboard/rundetails-65890b3c81d7400d08fa9ee5?branch=master&workflowId=31499794770&attempt=2" target="_blank">Cypress dashboard</a>.
> Tags: `@tag.All`
> Spec:
> <hr>Tue, 11 Aug 2026 15:18:18 UTC


wyattwalter:
@coderabbitai review

wyattwalter:
Addressed in 499cf3e: the parse step now strips any stale bare `*` from the limited list via `stripAllowAllFrameAncestorTokens` and clears the stale localStorage entry, so a limited list can never round-trip back to allow-all. Host wildcards are preserved. Added unit tests.

wyattwalter:
@coderabbitai review

wyattwalter:
Addressed in 37c2e1b: added `sanitizeLimitedFrameAncestors`, which strips bare `*` and normalizes a resulting `'none'` to empty, so LIMIT mode can never emit the disable-everywhere sentinel (e.g. from a stale `* 'none'`). Added a regression test for `* 'none'`. Host wildcards preserved.

wyattwalter:
@coderabbitai review

wyattwalter:
@coderabbitai review

wyattwalter:
@coderabbitai review

wyattwalter:
Addressed in the latest commit: `normalizeFrameAncestorToken` now canonicalizes the quoted forms too (`'SELF'`/`'NONE'`, case-insensitive), so a quoted-uppercase `'NONE'` is recognized as the disable keyword and stripped/rejected like a bare `none`. Extended the predicate/contains/removal/sanitizer tests with `'NONE'` coverage.

wyattwalter:
@coderabbitai review

